PR #1, opened 2021-08-26
Roadmap to MVP

da-kami:
This is a rough MVP roadmap that we can iterate on.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -1,0 +1,60 @@
+# Hermes Roadmap
+
+## MVP
+
+The Minimal Viable Product's goal is to showcase that non-custidial CFD trading on Bitcoin is possible.
+
+## Scope
+
+### Roles
+
+- **Taker**: Cannot create and publish orders, but just takes what is published by a Maker
+- **Maker**: Can create and publish orders
+- **Oracle**: Signs and publishes attested prices at specific points in time
+
+### In scope
+
+For the MVP there is only one Maker that takes the selling side and creates sell orders.
+The maker does not do any automation.
+The maker dictates the price.
+
+A user is always in the role of a taker.
+The user has a simple user interface and can take the maker's order there.
+The taker can specify a quantity, the leverage is fixed to `x5`.
+For the MVP the leverage is fixed to `x5` for both sell and buy orders.
+The oracle is needed for attestation of prices at a certain point in time.
+The oracle is to be run by a separate party that is neither the taker nor the maker.
+
+Overview sequence diagram:
+
+![MVP sequence diagram](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/comit-network/hermes/b7778f0556d1cbfd578aa28042beb24c7f13d5a3/docs/asset/mvp_sequence_diagram.puml)
+
+Constraints:
+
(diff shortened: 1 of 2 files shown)


docs/roadmap.md
```diff
@@ -0,0 +1,59 @@
...
+A user is always in the role of a taker.
+The user has a simple user interface and can take the maker's order there.
+The taker can specify a quantity, the leverage is fixed to `x5`.
+For the MVP the leverage is fixed to `x5` for both sell and buy orders.
+The oracle is needed for attestation of prices at a certain point in time.
+The oracle is to be run by a separate party that is neither the taker nor the maker.
+
+Overview sequence diagram:
+
+![MVP sequence diagram](http://www.plantuml.com/plantuml/proxy?cache=no&src=https://raw.githubusercontent.com/comit-network/hermes/d12e04d4954deb2ee9ebdc96f5a3ab59b7de370c/docs/asset/mvp_sequence_diagram.puml)
```
luckysori: :question: 

On the sequence diagram, why and how are both the buyer and seller publishing a CET? AFAIK if either party decides to publish it the other just has to wait for the money to arrive in their wallet.

da-kami: I thought that there would be two execution paths for the buyer and the seller to claim their funds. But I see that this might not make sense, as the CET published would split up the output to both the seller and buyer no matter who publishes the transaction. I can re-model (or abstract over) that to avoid confusion. 

Questions on top: Who is expected to publish the contract? (in the current model the seller), and who is expected to publish the settlement transaction? (either? - I could also model that...)

luckysori: >  the CET published would split up the output to both the seller and buyer no matter who publishes the transaction

:+1: 

> Who is expected to publish the contract?

From the protocol's perspective I think either can publish. See: https://github.com/coblox/coblox-research/blob/cfd-protocol/cfds/protocol.org#both-parties-should-end-up-with-full-lock-transaction.

> who is expected to publish the settlement transaction?

I think the simplest thing is to allow all parties to publish all CETs. But, in practice, a transaction which is less beneficial for a certain party than the fallback refund transaction may never be published by that particular party.

da-kami: I updated the diagram. Note that the assumption is that the maker (seller) publishes the contract initially, but the taker (buyer) to close the position by publishing a CET. I think only that makes sense - because as long as the position is open the buyer will have to pay interest. Hence, it is in the buyer's interest to close (=settle on chain atm) the position eventually.

da-kami: From the seller's point of view initially I would not care about closing the position. Eventually this is of course needed. Does that make sense @bonomat ?

docs/roadmap.md
```diff
@@ -0,0 +1,60 @@
...
+  - Maker
+    - Can create a sell order (represents the sell side)
+    - Sell order publication is done manually
+    - Take requests are accepted manually
+  - ♻️ Oracle
+
+Architecture recommendations
+
+- Maker and taker have different purpose / scope and hence should be separate binaries. 
+  - Code can be shared in e.g. UI components and rust libs
```
thomaseizinger: Can remove that now given the open PR for `ARCHITECTURE.md`.

thomaseizinger (approved):

PR #13, opened 2021-08-31
Initial taker UI+daemon prototype

da-kami:
- taker receives static balance from taker daemon (sse)
- taker receives static CFD offer from the taker daemon (sse)
- taker can post CFD take request to taker daemon
- taker receives CFD list from taker daemon (sse)

Upon posting a CFD to the daemon a CFD will be created in the daemon and pushed into the sse stream.
There is one sse stream with three events (balance, offer, cfds).


da-kami:
The taker is "functional" using static dummy data. We decided to get this work in so we can build on top of it.

daemon/src/routes.rs
```diff
@@ -0,0 +1,134 @@
+use crate::cfd::{static_cfd_offer, Cfd, CfdOffer, CfdState, CfdTakeRequest, Usd};
+
```
thomaseizinger: This newline prevents the formatter from merging all use statements.

daemon/src/main.rs
```diff
@@ -0,0 +1,14 @@
+mod cfd;
```
thomaseizinger: Should this file just live in `bin/taker.rs` perhaps so we can have two different binaries?

dprint.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "https://dprint.dev/schemas/v0.json",
+  "projectType": "openSource",
+  "incremental": true,
+  "markdown": {
+  },
+  "rustfmt": {
+    "merge_imports": true
```
thomaseizinger: This has been deprecated already, I can recommend you update to the rust plugin `0.4.0`.

Probably best to re-initialize the repo from scratch, as that will pull the latest plugin versions in.

thomaseizinger: It is superseded by `import_granularity`.

frontend/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "frontend",
```
thomaseizinger, line 2: Isn't this just the taker frontend perhaps?

frontend/src/App.test.tsx
```diff
@@ -0,0 +1,9 @@
+import { render, screen } from "@testing-library/react";
+import React from "react";
+import App from "./App";
+
+test("renders learn react link", () => {
+    render(<App />);
+    const linkElement = screen.getByText(/learn react/i);
+    expect(linkElement).toBeInTheDocument();
+});
```
thomaseizinger: Does this test even pass?

daemon/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "daemon"
```
thomaseizinger, line 2: I guess only one daemon means you will be having two binaries within that crate?

daemon/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+edition = "2018"
+
+[dependencies]
+anyhow = "1"
+bitcoin = { version = "0.27", features = ["use-serde"] }
```
thomaseizinger: Maybe depend on `bdk` right away? That is what we will need eventually.

In `xmr-btc-swap`, it was a pain to try and unify dependencies from all `bitcoin`, `bdk` and `miniscript`. Would be nice to avoid this here.

Also, dependabot has an easier time to update things if you access everything through `bdk`.

thomaseizinger (commented):
Looks good so far.

Don't forget to add a CI config eventually :)

daemon/src/main.rs
da-kami: yes, @klochowicz is working on that. We can also decide to merge at a later point (together with the maker ramp-up), but it keeps growing so I felt why not make a cut here

frontend/package.json
da-kami: yes, but we are not entirely sure how to do the split yet

daemon/Cargo.toml
da-kami: that was the idea so far, yes

daemon/Cargo.toml
da-kami: Ok! Then I'll switch that for a `bdk` dependency and remove `bitcoin`

da-kami:
@thomaseizinger given your comments on the frontend/daemon maker/taker split it's probably better to wait a bit longer with merging this. Hence, converted to Draft :)

Thanks for the review! I'd say we decide tomorrow if we want to add stuff and then merge or just merge as is and add another PR on top :)

dprint.json
da-kami: I pushed a commit that changes that and re-formats - is that good enough? Why do I have to re-initialize the repo?

frontend/src/App.tsx
```diff
@@ -0,0 +1,237 @@
...
+import axios from "axios";
+import React, { useState } from "react";
+import { useAsync } from "react-async";
+import { Link as RouteLink, Route, Switch } from "react-router-dom";
+import { useEventSource, useEventSourceListener } from "react-sse-hooks";
+import "./App.css";
+import CFD from "./components/CFD";
+
+/* TODO: Change from localhost:8000 */
+const BASE_URL = "http://localhost:8000";
```
bonomat: I guess that's what you want to replace with the proxy? 

https://create-react-app.dev/docs/proxying-api-requests-in-development/

frontend/src/App.tsx
```diff
@@ -0,0 +1,237 @@
...
+
+function App() {
+    let source = useEventSource({ source: BASE_URL + "/feed" });
+
+    const cfds = useLatestEvent<Cfd[]>(source, "cfds");
+    const offer = useLatestEvent<Offer>(source, "offer");
+    const balance = useLatestEvent<number>(source, "balance");
+
+    const toast = useToast();
+    let [quantity, setQuantity] = useState<string>("10000");
```
bonomat: I think the type is inferred here and you don't need `<string>`.

daemon/src/cfd.rs
```diff
@@ -0,0 +1,113 @@
...
+
+pub fn static_cfd_offer() -> CfdOffer {
+    CfdOffer {
+        id: Uuid::new_v4(),
+        price: Usd(49_000),
+        min_amount: Usd(1_000),
+        max_amount: Usd(10_000),
+        leverage: Leverage(5),
+        trading_pair: TradingPair::BtcUsd,
+        liquidation_price: Usd(42_000),
```
bonomat: Just so that this is not hit you later on: the `liquidation_price` depends on the taker's input (amount + leverage), e.g. if he wants to buy 1 BTC at $49k with a leverage of 5 the needed collateral will be 0.2BTC and liquidation_price depends on this amount.

frontend/src/App.tsx
da-kami: yes, we already tried but could not getting working properly so just ignored it for now 😬

frontend/src/App.tsx
da-kami: Yeah, initially it was `<string | null>`, but we just set it now so the type is always string and can be inferred.

daemon/src/cfd.rs
da-kami: As far as our formulas in the model go, the liquidation price is purely based on the `leverage` factor (i.e. `x5`) and the initial price (and a maintenance margin that we currently don't depict yet). 
The amount the taker enters is irrelevant as far as I understand our model!

Given, that we have a fixed leverage at the moment the liquidation price is fixed for each offer, because the offer includes price and leverage.

bonomat: The taker can enter an amount between `min_amount` and `max_amount` right? If you look at our spread sheet: that's the quantity the taker wants to buy. That's what the maker (as the lender or the person who wants to go short) then has to come up with. 
This amount + the leverage (`5x`) influences the `liquidation_price`. 

Differently said: 
- the maker has to put up an amount X in BTC which the taker wants to buy (defined in USD aka Z). 
- the taker has to put up an amount Y in BTC which depends on the `X / leverage`
- the `liquidation_price` depends then on the combination of `X` and `Y`, i.e. roughy `(X+Y)*liquidation_price=Z)`

I hope that helps.

bonomat: Don't mind me then :D As discussed via call.

da-kami: Note for completion: the model is currently based the formula described as `isolated margin mode` described [here](
https://help.bybit.com/hc/en-us/articles/360039261334-How-to-calculate-Liquidation-Price-Inverse-Contract-).

daemon/src/routes_taker.rs
```diff
@@ -108,16 +111,52 @@ pub async fn start_http() -> Result<()> {
...
-            loop {
-                let cfd = take_cfd_receiver.recv().await.unwrap();
+            let frame_read = FramedRead::new(read, LengthDelimitedCodec::new());
 
-                // TODO: Communicate with maker to initiate taking CFD...
-                //  How would we achieve long running async stuff in here?
+            let mut messages = frame_read.map(|result| {
+                let message = serde_json::from_slice::<Message>(&result?)?;
+                anyhow::Result::<_>::Ok(message)
+            });
```
thomaseizinger, line 161: If you want to abstract this away, you can create an implementation of `Encoder` and `Decoder` that wraps the `LengthDelimitedCodec` and converts to and from the provided chunks of bytes to messages via `serde_json`.

daemon/src/socket.rs
```diff
@@ -9,8 +11,9 @@ use tokio_util::codec::{FramedWrite, LengthDelimitedCodec};
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(tag = "type", content = "payload")]
 pub enum Message {
-    Ping,
-    Pong,
+    SendCurrentOffer(CfdOffer),
+    TakeOffer(CfdTakeRequest),
+    ConfirmOffer(Cfd),
 }
```
thomaseizinger: Perhaps we can split this enum into two:

```rust
enum MakerToTaker {

}

enum TakerToMaker {

}
```

If you make `spawn_sender` generic over the message, this should work.

frontend/taker.html
```diff
@@ -8,6 +8,6 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="module" src="../src/main_taker.tsx"></script>
+    <script type="module" src="./src/main_taker.tsx"></script>
```
thomaseizinger, line 11: I think you could try and inline the content of `main_taker` here as well. Not sure if `rollup` is powerful enough for that but that would be kind of cool!

Then we don't need to have these `main_taker.tsx` and `main_maker.tsx` files ...

PR #16, opened 2021-09-01
Add cfd_protocol crate to workspace

luckysori:
We've added a test which should prove that the spend transactions we generate in the protocol are valid. Some further assertions may be missing at this stage.

Currently it is all test code, so defining useful library APIs will come next. That will also include off-chain renewal of the CFD.

cc @da-kami @klochowicz


cfd_protocol/src/lib.rs
```diff
@@ -1,31 +1,154 @@
...
+use rand::RngCore;
+use rand::{CryptoRng, SeedableRng};
+use rand_chacha::ChaChaRng;
+use secp256k1_zkp::{self, schnorrsig, Signature};
+use secp256k1_zkp::{bitcoin_hashes::sha256t_hash_newtype, SECP256K1};
+use secp256k1_zkp::{bitcoin_hashes::*, EcdsaAdaptorSignature};
+use secp256k1_zkp::{Secp256k1, SecretKey};
+use std::collections::HashMap;
+
+// TODO: Consider extension trait
```
thomaseizinger: +1

cfd_protocol/src/lib.rs
```diff
@@ -1,31 +1,154 @@
...
+
+// NOTE: We have decided to not offer any verification utility because
+// the APIs would be incredibly thin
+
+pub struct Party {
+    pub lock_psbt: PartiallySignedTransaction,
+    pub dlc_pk: PublicKey,
+    pub dlc_amount: Amount,
+    pub address: Address,
+}
```
thomaseizinger: Creating one of these requires funding the psbt correct?

Maybe a extension trait above should already return you a `Party` struct?

thomaseizinger: The wallet could fund the psbt and automatically fill in the address!

cfd_protocol/src/lib.rs
```diff
@@ -646,11 +648,110 @@ mod tests {
...
+                maker_own_pk_hash, taker_publish_pk_hash, taker_rev_pk_hash
+            );
+            let taker_punish_condition = format!(
+                "and_v(v:pkh({}),and_v(v:pkh({}),pk_h({})))",
+                taker_own_pk_hash, maker_publish_pk_hash, maker_rev_pk_hash
+            );
+            let descriptor_str = format!(
+                "wsh(c:or_i(or_i({},{}),{}))",
+                maker_punish_condition, taker_punish_condition, cet_or_refund_condition
+            );
```
thomaseizinger: Have you considered using the miniscript compiler to create a miniscript from a policy instead of handrolling it? Might yield a more optimal version!

cfd_protocol/src/lib.rs
```diff
@@ -558,31 +621,161 @@ mod tests {
...
+            address: &Address,
+            amount: Amount,
+            encisg: EcdsaAdaptorSignature,
+            sk: SecretKey,
+            (revocation_them_sk, revocation_them_pk): (SecretKey, PublicKey),
+            publishing_them_pk: PublicKey,
+            revoked_commit_tx: Transaction,
+        ) -> Result<Self> {
+            // CommitTransaction has only one input
+            let input = revoked_commit_tx.input[0].clone();
```
thomaseizinger: With `Itertools`, you could encode this invariant using `into_iter` and `exactly_one`: https://docs.rs/itertools/0.10.1/itertools/trait.Itertools.html#method.exactly_one

cfd_protocol/src/lib.rs
```diff
@@ -558,31 +621,161 @@ mod tests {
...
+                let witness = witness.as_slice();
+
+                let res = bitcoin::secp256k1::Signature::from_der(&witness[..witness.len() - 1]);
+                match res {
+                    Ok(sig) => sigs.push(sig),
+                    Err(_) => {
+                        continue;
+                    }
+                }
+            }
```
thomaseizinger: Such kind of stuff can be nicely extracted into extension traits so the actual flow of this function is less noisy and composed of more powerful low-level functionality.

cfd_protocol/src/lib.rs
```diff
@@ -558,31 +621,161 @@ mod tests {
...
+                }
+            };
+
+            let digest = Self::compute_digest(&punish_tx, &commit_tx);
+
+            let satisfier = {
+                let mut satisfier = HashMap::with_capacity(3);
+
+                let pk = bitcoin::secp256k1::PublicKey::from_secret_key(SECP256K1, &sk);
+                let pk_hash = hash160::Hash::hash(&pk.serialize()[..]);
```
thomaseizinger: https://docs.rs/bitcoin/0.27.0/bitcoin/util/ecdsa/struct.PublicKey.html#method.pubkey_hash

thomaseizinger (commented):
Some thoughts, feel free to ignore for now :)

cfd_protocol/src/lib.rs
luckysori: Great idea! The only thing that's less obvious is who should generate the corresponding `dlc_sk`? I'm leaning towards generating before and passing in `dlc_pk` to the trait method.

cfd_protocol/src/lib.rs
luckysori: Good point! We pretty much copied what we already had in `thor`, but we did leave a TODO about optimizing the miniscript. We'll try to use the miniscript compiler once we tackle that TODO.

cfd_protocol/src/lib.rs
luckysori: Interesting idea. I rarely consider using the extension trait pattern, so it's very useful to get feedback on where they can be used!

In this case, would your idea be to extend TxIn with (e.g.) TxInExt and define a extract_signatures trait method? I guess in this particular instance since this all remains private API of the library we could just extract a free function too.

thomaseizinger: Yeah, the only advantage of an extension trait over a free function is that it is easier to discover because the IDE will give you auto-completion for it.

In this case, yes probably put it on `TxIn`.

cfd_protocol/src/lib.rs
```diff
@@ -246,7 +242,7 @@ pub fn punish_transaction(
     revoked_commit_tx: &Transaction,
 ) -> Result<Transaction> {
     // CommitTransaction has only one input
-    let input = revoked_commit_tx.input.clone().into_iter().exactly_one()?;
```
bonomat: Unfortunately this breaks the build with: 

```
error[E0621]: explicit lifetime required in the type of `revoked_commit_tx`
   --> cfd_protocol/src/lib.rs:246:61
    |
243 |     revoked_commit_tx: &Transaction,
    |                        ------------ help: add explicit lifetime `'static` to the type of `revoked_commit_tx`: `&'static bitcoin::Transaction`
...
246 |     let input = revoked_commit_tx.input.iter().exactly_one()?;
    |                                                             ^ lifetime `'static` required

```

thomaseizinger:
Is this ready to review / merge?

Some commits have the exact same commit message, I guess those could be squashed? Apart from that, I wouldn't worry too much. Let's focus on getting a foundation in that we can iterate on!

luckysori:
> Is this ready to review / merge?

I think so. It's only a starting point, but I think we can merge this and start working on smaller PRs. Perhaps we could wait for your review and then decide what should be done immediately and what can be done in separate PRs.

> Some commits have the exact same commit message, I guess those could be squashed?

I've done an interactive rebase now. Let me know if there's anything else that we could squash or improve in terms of commits!

cfd_protocol/src/lib.rs
luckysori: I've created https://github.com/comit-network/hermes/issues/19 to track this.

cfd_protocol/src/lib.rs
luckysori: We haven't done this yet because it's just refactoring, but I would like to try this out: https://github.com/comit-network/hermes/issues/20. 

I could have probably just done it, but we can use that issue to remember that there might be other opportunities to use extension traits.

cfd_protocol/src/lib.rs
bonomat: I used the online compiler for the updated script.

luckysori: Ah, I didn't know that! It would still be good to have the code generate the miniscript from the policy, so I think we can leave the issue open.

cfd_protocol/src/lib.rs
```diff
@@ -162,13 +520,375 @@ mod tests {
...
+
+    fn encverify(
+        &self,
+        encsig: &EcdsaAdaptorSignature,
+        signing_pk: &PublicKey,
+        encryption_pk: &PublicKey,
+    ) -> Result<()> {
+        encsig.verify(
+            SECP256K1,
+            &secp256k1_zkp::Message::from_slice(&self.sighash).expect("sighash is valid message"),
```
thomaseizinger: I think all the hash types of the `rust-bitcoin` ecosystem have fail-safe `From` implementations via this `ThirtyTwoByteHash` marker trait?

https://docs.rs/secp256k1/0.20.2/src/secp256k1/lib.rs.html#504-509

cfd_protocol/src/lib.rs
```diff
@@ -162,13 +520,375 @@ mod tests {
...
+            .unsigned_tx
+            .output
+            .into_iter()
+            .filter(|out| !out.script_pubkey.is_empty())
+            .collect();
+
+        let lock_output = TxOut {
+            value: amount.as_sat(),
+            script_pubkey: lock_descriptor
+                .address(Network::Regtest)
```
thomaseizinger, line 813: Don't hardcode any networks please :)

cfd_protocol/src/lib.rs
```diff
@@ -22,6 +22,9 @@ use std::collections::HashMap;
 /// In satoshi per vbyte.
 const MIN_RELAY_FEE: u64 = 1;
 
+// TODO use Amount type
+const P2PKH_DUST_LIMIT: u64 = 546;
```
thomaseizinger: I thin the recent rust-bitcoin version has a constant for this!

cfd_protocol/src/lib.rs
```diff
@@ -543,6 +544,14 @@ impl PunishTransaction {
                 &publish_them_sk,
             );
 
+            let revocation_them_pk = PublicKey::from_private_key(
+                SECP256K1,
+                &PrivateKey {
+                    compressed: true,
+                    network: Network::Regtest,
```
thomaseizinger: Same here, no hard-coded network please!

cfd_protocol/src/lib.rs
```diff
@@ -13,6 +13,7 @@ use bdk::{
     descriptor::Descriptor,
     miniscript::{descriptor::Wsh, DescriptorTrait},
 };
+use bitcoin::PrivateKey;
```
thomaseizinger, line 16: Not sure if better but maybe the protocol should work with `secp256k1::SecretKey`s instead of the `bitcoin` ones?

cfd_protocol/src/lib.rs
```diff
@@ -28,12 +30,27 @@ const MIN_RELAY_FEE: u64 = 1;
...
-        todo!()
+impl<B, D> WalletExt for bdk::Wallet<B, D>
+where
+    D: BatchDatabase,
+{
+    fn build_party_params(&self, amount: Amount, identity_pk: PublicKey) -> Result<PartyParams> {
+        let mut builder = self.build_tx();
+        builder
+            .ordering(bdk::wallet::tx_builder::TxOrdering::Bip69Lexicographic)
+            .add_recipient(Script::new(), amount.as_sat());
```
thomaseizinger: Can we have a comment here why it is okay to pass an empty script? Does the size of the script here affect fee calculation?

cfd_protocol/src/lib.rs
```diff
@@ -209,12 +208,157 @@ pub fn spending_tx_sighash(
...
+    (taker_pk, taker_sig): (PublicKey, Signature),
+) -> Result<Transaction> {
+    let satisfier = {
+        let mut satisfier = HashMap::with_capacity(2);
+
+        satisfier.insert(maker_pk, (maker_sig, SigHashType::All));
+        satisfier.insert(taker_pk, (taker_sig, SigHashType::All));
+
+        satisfier
+    };
```
thomaseizinger, line 219: Might be able to condense this using the `from_iter` function on `HashMap`: https://doc.rust-lang.org/std/collections/struct.HashMap.html#impl-FromIterator%3C(K%2C%20V)%3E

```rust
HashMap::from_iter(vec![
	(maker_pk, (maker_sig, SigHashType::All)),
	(taker_pk, (taker_sig, SigHashType::All))
]);
```

cfd_protocol/src/lib.rs
```diff
@@ -209,12 +208,157 @@ pub fn spending_tx_sighash(
...
+    if sigs.is_empty() {
+        // No signature found, this should fail
+        unimplemented!()
+    }
+
+    // Attempt to extract y_other from every signature
+    let publish_them_sk = sigs
+        .into_iter()
+        .find_map(|sig| encsig.recover(SECP256K1, &sig, &publish_them_pk.key).ok())
+        .context("Could not recover secret key from revoked transaction")?;
```
thomaseizinger: Wild idea: To condense this down, could extract a custom `Iterator` extension trait that works on `Iterator<Item = Signature>`.

cfd_protocol/src/lib.rs
```diff
@@ -255,7 +399,7 @@ pub struct Payout {
 }
 
 impl Payout {
-    fn to_txouts(&self, maker_address: &Address, taker_address: &Address) -> Vec<TxOut> {
+    fn as_txouts(&self, maker_address: &Address, taker_address: &Address) -> Vec<TxOut> {
```
thomaseizinger, line 399: `as_` should be a cheap conversion by convention. You are returning an owned instance from a reference (`&self`) which implies an allocation. I think this should be named `to_`.

Did clippy suggest this?

thomaseizinger (approved):
Nice work, thank you!

Should be a great foundation to build on top of!
It would be nice if we can gradually move towards extracting a test harness for Bitcoin transactions that combines consensus-based tests with all kinds of other checks that we learn over time are needed :)

cfd_protocol/src/lib.rs
luckysori: I blame Clippy! (It was Clippy)

cfd_protocol/src/lib.rs
bonomat: This could be a source of errors as we are underpaying for the fees. 
@luckysori : I think we didn't notice this because I added  `.fee_rate(FeeRate::from_sat_per_vb(2.0))`. 

Did we run the docker-test with `1.0 sat/vbyte` as well?

bonomat: Can confirm, this is an issue if fees are set to 1.0
` RpcError { code: -26, message: "min relay fee not met, 238 < 252", data: None }`

bonomat: Fixed.

bonomat:
Shall we get this baby merged?

cfd_protocol/src/lib.rs
luckysori: Good catch. This just came in because we copied over something from `thor` (which didn't need this type anyway :stuck_out_tongue:).

cfd_protocol/src/lib.rs
luckysori: I think Philipp tackled this in a1cc6badb21de442403f4f8fb046b139076d498e.

PR #22, opened 2021-09-08
Sighash to message

luckysori:
This is for https://github.com/comit-network/hermes/pull/16#discussion_r703115601.

What do you think? I tried to just use the trait you suggested, but there is a mismatch between the types because we're using `secp256k1` and `secp256k1-zkp`. There might be a better solution than what I've done though.

### DIFF
--- a/cfd_protocol/src/lib.rs
+++ b/cfd_protocol/src/lib.rs
@@ -112,8 +112,7 @@ pub fn build_cfd_transactions(
             taker.lock_amount,
         );
 
-        let sighash =
-            secp256k1_zkp::Message::from_slice(&tx.sighash()).expect("sighash is valid message");
+        let sighash = tx.sighash().to_message();
         let sig = SECP256K1.sign(&sighash, &identity_sk);
 
         (tx.inner, sig)
@@ -200,7 +199,7 @@ pub fn spending_tx_sighash(
         spent_amount.as_sat(),
         SigHashType::All,
     );
-    secp256k1_zkp::Message::from_slice(&sighash).expect("sighash is valid message")
+    sighash.to_message()
 }
 
 pub fn finalize_spend_transaction(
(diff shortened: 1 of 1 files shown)


thomaseizinger:
If we have a type mismatch then we need to bump the version of -zkp and release. Or where is the mismatch?
I assume bitcoin depends on a newer version of secp than -zkp does?

cfd_protocol/src/lib.rs
```diff
@@ -875,6 +868,19 @@ impl TransactionExt for bitcoin::Transaction {
     }
 }
 
+trait SigHashExt {
+    fn as_message(&self) -> secp256k1_zkp::Message;
+}
+
+impl SigHashExt for SigHash {
+    fn as_message(&self) -> secp256k1_zkp::Message {
```
thomaseizinger: Should be `to_message` I think.

thomaseizinger (approved):
An extension trait is a good workaround until we can pass it along natively.

luckysori:
> If we have a type mismatch then we need to bump the version of -zkp and release. Or where is the mismatch?
> I assume bitcoin depends on a newer version of secp than -zkp does?

I haven't managed to get to the bottom of the dependency mismatch problem, but I think you're on the right track. I'll merge this and record and issue.

thomaseizinger:
> > If we have a type mismatch then we need to bump the version of -zkp and release. Or where is the mismatch?
> > I assume bitcoin depends on a newer version of secp than -zkp does?
> 
> I haven't managed to get to the bottom of the dependency mismatch problem, but I think you're on the right track. I'll merge this and record and issue.

I can't find any version mismatch in the latest `bitcoin` version. Weird ...

luckysori:
> I can't find any version mismatch in the latest `bitcoin` version. Weird ...

I was just getting a headache trying to figure it out lol. Maybe we can take a look together or you can give it a go yourself.

PR #23, opened 2021-09-08
Remove hard-coded `Network`s

luckysori:
And other stuff related to PR #16.

It looks like the protocol doesn't need to know about what network it runs on? Is that weird?


thomaseizinger (approved):
I don't think that is weird at all. Bitcoin's networks are a surface level thing so users don't mix up addresses :)

PR #24, opened 2021-09-08
Construct satisfiers using HashMap::from_iter

luckysori:
As suggested in https://github.com/comit-network/hermes/pull/16#discussion_r703118707.


thomaseizinger (approved):
Less mutability 🥳

PR #28, opened 2021-09-08
Get rid of unnecessary and unwanted `Message` type

luckysori:
Is `String` the best choice for modelling this type? This is just an option, but at least we can move away from the silly `Message` type I introduced for no reason.


thomaseizinger (approved):
String is better than the message, maybe loosen it to bytes even? Not sure why it would need to be necessary human-readable!

luckysori:
> String is better than the message, maybe loosen it to bytes even? Not sure why it would need to be necessary human-readable!

Let's do bytes, yep!

PR #31, opened 2021-09-09
Extract script from descriptor directly

thomaseizinger:
No need to do a round-trip to an address and back!


luckysori (approved):
Thank you, I forgot to do this in the tests!

bonomat (approved):
😅

PR #32, opened 2021-09-09
Add `renew_cfd_transactions` API

luckysori:
As expected, renewing the CFD is the same as creating a CFD without a new lock transaction.

The tests are very repetitive, so I would be happy to start thinking about refactoring things there.

### DIFF
--- a/cfd_protocol/src/lib.rs
+++ b/cfd_protocol/src/lib.rs
@@ -70,17 +70,16 @@ pub fn build_cfd_transactions(
     /// TODO: Should this be an argument to this function?
     const CET_TIMELOCK: u32 = 12;
 
-    let lock_tx = LockTransaction::new(
+    let lock_tx = lock_transaction(
         maker.lock_psbt.clone(),
         taker.lock_psbt.clone(),
         maker.identity_pk,
         taker.identity_pk,
         maker.lock_amount + taker.lock_amount,
-    )
-    .context("cannot build lock tx")?;
+    );
 
     let commit_tx = CommitTransaction::new(
-        &lock_tx,
+        &lock_tx.global.unsigned_tx,
         (
             maker.identity_pk,
             maker_punish_params.revocation_pk,
@@ -91,7 +90,8 @@ pub fn build_cfd_transactions(
             taker_punish_params.revocation_pk,
             taker_punish_params.publish_pk,
         ),
-    );
+    )
+    .context("cannot build commit tx")?;
 
     let identity_pk = secp256k1_zkp::PublicKey::from_secret_key(SECP256K1, &identity_sk);
     let commit_encsig = if identity_pk == maker.identity_pk.key {
@@ -138,7 +138,7 @@ pub fn build_cfd_transactions(
         .context("cannot build and sign all cets")?;
 
     Ok(CfdTransactions {
-        lock: lock_tx.inner,
+        lock: lock_tx,
         commit: (commit_tx.inner, commit_encsig),
         cets,
         refund,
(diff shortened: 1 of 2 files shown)


cfd_protocol/src/lib.rs
```diff
@@ -731,95 +745,55 @@ impl CommitTransaction {
...
-    }
+    let lock_tx = Transaction {
+        version: 2,
+        lock_time: 0,
+        input: vec![
+            maker_psbt.global.unsigned_tx.input,
+            taker_psbt.global.unsigned_tx.input,
+        ]
+        .concat(),
+        output: vec![vec![lock_output], maker_change, taker_change].concat(),
```
thomaseizinger: I think this double `Vec` is a bit confusing. Instead of `collect`ing all the individual iterators, you could concatenate them and use `iter::once` for the `lock_output`.

cfd_protocol/src/lib.rs
```diff
@@ -645,24 +645,38 @@ impl CommitTransaction {
...
+        let (lock_outpoint, lock_amount) = {
+            let vout = lock_tx
+                .output
+                .iter()
+                .position(|out| out.script_pubkey == lock_descriptor.script_pubkey())
+                .context("lock script not found in lock tx")?;
+            let outpoint = OutPoint {
+                txid: lock_tx.txid(),
+                vout: vout as u32,
+            };
```
thomaseizinger: Perhaps a candidate for an extension trait?

cfd_protocol/src/lib.rs
```diff
@@ -78,25 +71,112 @@ pub fn build_cfd_transactions(
...
+}
+
+pub fn renew_cfd_transactions(
+    lock_tx: PartiallySignedTransaction,
+    (maker_pk, maker_lock_amount, maker_address, maker_punish_params): (
+        PublicKey,
+        Amount,
+        Address,
+        PunishParams,
+    ),
```
thomaseizinger, line 102: Why does `create_cfd_transactions` use `PartyParams` but `renew_cfd_transactions` doesn't?

thomaseizinger (approved):
Nice work! Condensing down the tests sounds good to me. Should be useful for the integration later as well!

cfd_protocol/src/lib.rs
luckysori: I thought about it but it would be a bit redundant to pass in the lock PSBT twice.

thomaseizinger: Didn't see that, fair enough :)

cfd_protocol/src/lib.rs
```diff
@@ -70,17 +70,16 @@ pub fn build_cfd_transactions(
...
         maker.identity_pk,
         taker.identity_pk,
         maker.lock_amount + taker.lock_amount,
-    )
-    .context("cannot build lock tx")?;
+    );
 
     let commit_tx = CommitTransaction::new(
-        &lock_tx,
+        &lock_tx.global.unsigned_tx,
```
bonomat, line 162: Should one prefer this over `lock_tx.extract_tx()`?

cfd_protocol/tests/cfds.rs
```diff
@@ -372,6 +372,443 @@ fn create_cfd() {
         .expect("valid punish transaction signed by taker");
 }
 
+#[test]
+fn renew_cfd() {
```
bonomat, line 103: I guess this asks for some refactoring and code-reuse between the two different tests :)

bonomat (approved):
Look good, nice work 👋👋

cfd_protocol/src/lib.rs
thomaseizinger: `extract_tx` takes [`self`](https://docs.rs/bitcoin/0.27.0/bitcoin/util/psbt/struct.PartiallySignedTransaction.html#method.extract_tx) which consumes the entire PSBT. I guess in this case, @luckysori just wanted a reference to taking ownership of the PSBT itself was wasteful and would have probably resulted in a clone.

cfd_protocol/tests/cfds.rs
```diff
@@ -46,290 +47,77 @@ fn create_cfd() {
...
+            pub_pk: maker_pub_pk,
+        },
+        CfdKeys {
+            sk: taker_sk,
+            pk: taker_pk,
+            rev_sk: taker_rev_sk,
+            rev_pk: taker_rev_pk,
+            pub_sk: taker_pub_sk,
+            pub_pk: taker_pub_pk,
+        },
```
thomaseizinger: Just call this `taker` and refer to fields as `taker.sk`.

thomaseizinger (approved):

PR #48, opened 2021-09-14
Calculate refund timelock based on CFD term.

bonomat:
A threshold is added to the contract term. 

This is needed because a user might "change their mind" and publish transactions at any time trying to get out of the contract. The refund timeout should ensure that the refund transaction can only be published if the oracle does not publish a signature.

When implementing this I came across a small problem which I documented here: https://github.com/comit-network/hermes/issues/47

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -299,6 +299,20 @@ fn calculate_profit(
     Ok((Amount::ZERO, Usd::ZERO))
 }
 
+pub trait AsBlocks {
+    /// Calculates the duration in Bitcoin blocks.
+    ///
+    /// On Bitcoin there is a block every 10 minutes/600 seconds on average.
+    /// It's the caller's responsibility to round the resulting floating point number.
+    fn as_blocks(&self) -> f32;
+}
+
+impl AsBlocks for Duration {
+    fn as_blocks(&self) -> f32 {
+        self.as_secs_f32() / 60.0 / 10.0
+    }
+}
+
 /// Calculates the buyer's margin in BTC
 ///
 /// The margin is the initial margin and represents the collateral the buyer has to come up with to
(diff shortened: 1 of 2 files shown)


daemon/src/model/cfd.rs
```diff
@@ -510,6 +524,23 @@ mod tests {
             r#"{"type":"Error","payload":{"common":{"transition_timestamp":{"secs_since_epoch":0,"nanos_since_epoch":0}}}}"#
         );
     }
+
+    #[test]
+    fn test_secs_into_blocks() {
+        let error_margin = f32::EPSILON;
```
thomaseizinger, line 530: :+1:

daemon/src/model/cfd.rs
```diff
@@ -331,6 +331,20 @@ fn calculate_sell_margin(price: Usd, quantity: Usd) -> Result<Amount> {
     Ok(margin)
 }
 
+pub trait AsBlocks {
+    /// Calculates the duration in Bitcoin blocks.
+    ///
+    /// On Bitcoin there is a block every 10 minutes/600 seconds on average.
+    /// It's the caller's responsibility to round the resulting floating point number.
```
thomaseizinger: Nice doc comment!

daemon/src/taker_cfd_actor.rs
```diff
@@ -192,7 +210,7 @@ fn setup_contract(
             (maker.clone(), maker_punish),
             (taker.clone(), taker_punish),
             oracle_pk,
-            0, // TODO: Calculate refund timelock based on CFD term
+            offer.term.mul_f32(REFUND_THRESHOLD).as_blocks().ceil() as u32,
```
thomaseizinger, line 215: Nit: If we move this onto the `Offer`, it would be testable!

thomaseizinger (approved):
Good work! Thanks for writing such nice documentation!

PR #50, opened 2021-09-14
Margin calculation & refactoring

da-kami:
Sry, should have cut more contained commits, will see to do that next time 😬

The buy margin is calculated from `initial_price`, `quantity` and `leverage`.
The sell margin is calculated from `initial_price` and `quantity` (no leverage trading for the seller at the moment).

Includes several refactors:
- Separate API interface for `Cfd` and `CfdOffer` when mapping `toSseEvent`. This allows internal modelling to be different than the exposed API.
  - Remove the profit calculation internally, it is only relevant on the UI feed API level.
  - Move code that is specific to taker/maker http API from `model` to the respective `routes` module.
- Some simplification of the calculation model of `Usd` (could be further improved by deriving calculation traits)

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1,5 +1,5 @@
 use crate::model::{Leverage, Position, TradingPair, Usd};
-use anyhow::{Context, Result};
+use anyhow::Result;
 use bdk::bitcoin::secp256k1::{SecretKey, Signature};
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::{Amount, Transaction};
@@ -88,46 +88,18 @@ fn calculate_liquidation_price(
     price: &Usd,
     maintenance_margin_rate: &Decimal,
 ) -> Result<Usd> {
-    let leverage = Decimal::from(leverage.0);
-    let price = price.0;
+    let leverage = Decimal::from(leverage.0).into();
+    let maintenance_margin_rate: Usd = (*maintenance_margin_rate).into();
 
     // liquidation price calc in isolated margin mode
     // currently based on: https://help.bybit.com/hc/en-us/articles/360039261334-How-to-calculate-Liquidation-Price-Inverse-Contract-
-    let liquidation_price = price
-        .checked_mul(leverage)
-        .context("multiplication error")?
-        .checked_div(
-            leverage
-                .checked_add(Decimal::ONE)
-                .context("addition error")?
-                .checked_sub(
-                    maintenance_margin_rate
-                        .checked_mul(leverage)
-                        .context("multiplication error")?,
-                )
-                .context("subtraction error")?,
-        )
-        .context("division error")?;
+    let liquidation_price = price.checked_mul(leverage)?.checked_div(
+        leverage
+            .checked_add(Decimal::ONE.into())?
+            .checked_sub(maintenance_margin_rate.checked_mul(leverage)?)?,
+    )?;
 
-    Ok(Usd(liquidation_price))
-}
-
-/// The taker POSTs this to create a Cfd
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct CfdTakeRequest {
-    pub offer_id: CfdOfferId,
-    pub quantity: Usd,
-}
-
-/// The maker POSTs this to create a new CfdOffer
-// TODO: Use Rocket form?
-#[derive(Debug, Clone, Deserialize)]
-pub struct CfdNewOfferRequest {
-    pub price: Usd,
-    // TODO: [post-MVP] Representation of the contract size; at the moment the contract size is
-    // always 1 USD
-    pub min_quantity: Usd,
-    pub max_quantity: Usd,
+    Ok(liquidation_price)
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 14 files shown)


da-kami:
Just noticed that the UI integration is not complete, still have to add the `margin` into the taker's `buy` view and trigger re-calculation upon quantity change. Will keep it as draft till that is done and then open again.

daemon/src/model/cfd.rs
```diff
@@ -313,6 +294,36 @@ fn calculate_profit(
     Ok((Amount::ZERO, Usd::ZERO))
 }
 
+/// Calculates the buyer's margin in BTC
+///
+/// The margin is the initial margin and represents the collateral that both parties have to come up with to satisfy the contract.
```
bonomat: Isn't it only the buyer who has to come up with this margin?

daemon/src/model/cfd.rs
```diff
@@ -331,9 +342,61 @@ mod tests {
...
 
+    #[test]
+    fn given_leverage_of_one_and_equal_price_and_quantity_then_buy_margin_is_one_btc() {
+        let price = Usd(dec!(40000));
+        let quantity = Usd(dec![40000]);
+        let leverage = Leverage(1);
+
+        let buy_margin = calculate_buy_margin(price, quantity, leverage).unwrap();
+
+        assert_eq!(buy_margin, Amount::from_btc(1.0).unwrap());
```
bonomat: There is a constant for 1 BTC: `Amount::ONE_BTC`

daemon/src/model/cfd.rs
```diff
@@ -84,45 +84,18 @@ fn calculate_liquidation_price(
...
-        )
-        .context("division error")?;
-
-    Ok(Usd(liquidation_price))
-}
+    let liquidation_price = price.checked_mul(leverage)?.checked_div(
+        leverage
+            .checked_add(Decimal::ONE.into())?
+            .checked_sub(maintenance_margin_rate.checked_mul(leverage)?)?,
+    )?;
```
bonomat, line 100: looks good 👍

daemon/src/model/cfd.rs
```diff
@@ -313,6 +294,36 @@ fn calculate_profit(
...
 }
 
+/// Calculates the buyer's margin in BTC
+///
+/// The margin is the initial margin and represents the collateral that both parties have to come up with to satisfy the contract.
+/// Here we calculate the initial buy margin as: quantity / (initial_price * leverage)
+pub fn calculate_buy_margin(price: Usd, quantity: Usd, leverage: Leverage) -> Result<Amount> {
+    let leverage = Decimal::from(leverage.0).into();
+
+    let margin = quantity.checked_div(price.checked_mul(leverage)?)?;
```
bonomat, line 310: looks good 👍

daemon/src/model/cfd.rs
```diff
@@ -313,6 +294,36 @@ fn calculate_profit(
...
+
+    Ok(margin)
+}
+
+/// Calculates the seller's margin in BTC
+///
+/// The seller margin is represented as the quantity of the contract given the initial price.
+/// The seller can currently not leverage the position but always has to cover the complete quantity.
+fn calculate_sell_margin(price: Usd, quantity: Usd) -> Result<Amount> {
+    let margin = quantity.checked_div(price)?;
```
bonomat, line 325: looks good 👍

bonomat (approved):
Only checked the calculations which look good 👍

daemon/src/model/cfd.rs
da-kami: yeah, sorry for the confusion. 
It should probably not be called `margin` for the seller at the moment, because the seller cannot margin trade at the moment. For the maker (being the seller) it is always just the BTC amount of the quantity given the price atm. Weirdly, the leverage is also set to `5` for the seller at the moment.

Would it be more correct to also apply the leverage for the seller? (we don't have that in our spreadsheet, but I think it would be more correct)

I still used the same terminology as for the taker because at some point I would expect the maker to be able to use a margin (through the leverage) as well. But I am not sure that makes sense.

frontend/src/components/Types.tsx
```diff
@@ -45,8 +21,12 @@ export interface Cfd {
...
     quantity_usd: number;
+
+    margin: number;
+
     profit_btc: number;
     profit_usd: number;
 
-    state: CfdState;
+    state: string;
+    state_transition_timestamp_secs_since_epoch: number;
```
klochowicz: perhaps `state_transition_unix_timestamp` ? A shorter name meaning the same thing

klochowicz (approved):

daemon/src/model/cfd.rs
bonomat: I was actually only referring to the sentence: 

```
/// Calculates the buyer's margin in BTC
///
/// The margin is the initial margin and represents the collateral that both parties have to come up with to satisfy the contract.
```

It says first: `calculate the buyer's margin` and then further down: `.. that both parties have to come up with`. 

Imho everything in this function is correct, the only thing which I'd change is this: 

```suggestion
/// The margin is the initial margin and represents the collateral that the buyer has to come up with to satisfy the contract.
```


--------


Regarding the topic of allowing both sides to leverage their position: I'm still of the opinion that this does not make sense....

da-kami:
I'll squash and rebase this work and then see to get it merged :)

PR #51, opened 2021-09-14
Taker verifies all transactions

bonomat:
@luckysori : this needs some trained eyes... I'm not sure I got this right.

@thomaseizinger : shall we write tests for this ?  If, how? 😅

### DIFF
--- a/daemon/src/taker_cfd_actor.rs
+++ b/daemon/src/taker_cfd_actor.rs
@@ -5,14 +5,15 @@ use crate::db::{
 use crate::model::cfd::{Cfd, CfdOffer, CfdOfferId, CfdState, CfdStateCommon, FinalizedCfd};
 use crate::model::Usd;
 use crate::wire;
-use crate::wire::{Msg0, Msg1, SetupMsg};
-use bdk::bitcoin::secp256k1::{schnorrsig, SecretKey};
-
-use bdk::bitcoin::{self, Amount};
+use crate::wire::{AdaptorSignature, Msg0, Msg1, SetupMsg};
+use anyhow::{Context, Result};
+use bdk::bitcoin::secp256k1::{schnorrsig, SecretKey, Signature, SECP256K1};
+use bdk::bitcoin::{self, Amount, PublicKey, Transaction, Txid};
 use bdk::database::BatchDatabase;
+use bdk::descriptor::Descriptor;
 use cfd_protocol::{
-    commit_descriptor, create_cfd_transactions, lock_descriptor, PartyParams, PunishParams,
-    WalletExt,
+    commit_descriptor, compute_signature_point, create_cfd_transactions, lock_descriptor,
+    spending_tx_sighash, EcdsaAdaptorSignature, PartyParams, PunishParams, WalletExt,
 };
 use core::panic;
 use futures::Future;
@@ -201,10 +202,11 @@ fn setup_contract(
         send_to_maker(SetupMsg::Msg1(Msg1::from(taker_cfd_txs.clone())));
         let msg1 = receiver.recv().await.unwrap().try_into_msg1().unwrap();
 
-        let _lock_desc = lock_descriptor(maker.identity_pk, taker.identity_pk);
-        // let lock_amount = maker_lock_amount + taker_lock_amount;
+        let lock_desc = lock_descriptor(maker.identity_pk, taker.identity_pk);
 
-        let _commit_desc = commit_descriptor(
+        let lock_amount = maker.lock_amount + taker.lock_amount;
+
+        let commit_desc = commit_descriptor(
             (
                 maker.identity_pk,
                 maker_punish.revocation_pk,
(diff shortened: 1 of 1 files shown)


daemon/src/taker_cfd_actor.rs
```diff
@@ -253,3 +287,88 @@ fn setup_contract(
...
+        .context("could not calculate signature point")?;
+    verify_adaptor_signature(
+        tx,
+        spent_descriptor,
+        spent_amount,
+        encsig,
+        &PublicKey {
+            compressed: true,
+            key: sig_point,
+        },
```
thomaseizinger: Nit: There is a `new` constructor that sets `compresssed` to `true` for you!

thomaseizinger (approved):
I think for now this should do.

@luckysori I am getting the feeling that we probably want newtypes in the library for this kind of stuff. We don't need to enforce any invariants using these newtypes though.

In other words, it would be fine to have newtypes which can be constructed  / destructed from and to the inner types but they woudl just a provide natural place for putting all of this verification logic. I think one of the biggest source of bugs in here might be that we mix up different keys for example.

We don't have to put these types into `cfd_protocol` right away, they can also live within `daemon`. @bonomat are you interested in attempting this? The basic idea is to have things like:

```rust
pub struct CETs(pub Vec<...>);

impl CETs {
	pub fn verify(&self, ...) -> bool {}
}
```

Having such types would make the call-site (the `setup_contract` fn) more concise and we can re-use them for the maker as well. If we move them into the lib, we can even use them as part of the test and have everything tested as well.

thomaseizinger:
@bonomat We will squash and merge the other PR so hold back in merging this in please.

daemon/src/taker_cfd_actor.rs
```diff
@@ -253,3 +287,88 @@ fn setup_contract(
...
+fn verify_cets(
+    oracle_pk: &schnorrsig::PublicKey,
+    maker: &PartyParams,
+    taker_cets: &[(
+        Transaction,
+        EcdsaAdaptorSignature,
+        Vec<u8>,
+        schnorrsig::PublicKey,
+    )],
+    msg1: &Msg1,
```
luckysori: :upside_down_face: I would just pass in the `Vec<(Txid, EcdsaAdaptorSignature)>`.

daemon/src/taker_cfd_actor.rs
```diff
@@ -206,28 +208,60 @@ fn setup_contract(
...
+            &commit_tx,
+            &lock_desc,
+            lock_amount,
+            &msg1.commit,
+            &maker_punish.publish_pk,
+            &taker.identity_pk,
+        )
+        .unwrap();
+
+        // verify cet tx signatures
```
luckysori: :upside_down_face: I think we can remove these comments (particularly this one since it's a bit redundant with the name of the function you call just below it).

daemon/src/taker_cfd_actor.rs
```diff
@@ -206,28 +208,60 @@ fn setup_contract(
...
 
-        // TODO: Verify all signatures from the maker here
+        // verify commit tx signature
+        verify_adaptor_signature(
+            &commit_tx,
+            &lock_desc,
+            lock_amount,
+            &msg1.commit,
+            &maker_punish.publish_pk,
+            &taker.identity_pk,
```
luckysori: :wrench: We're verifying the `maker_encsig` on the `commit_tx`. That means we need to check it against the `maker_pk` and with the `taker_publish_pk` as the encryption key (it will leak the `taker_publish_sk` if the taker ever uses it). 

```suggestion
            &taker_punish.publish_pk,
            &maker.identity_pk,
```

luckysori (commented):
Just one mistake and a couple of nits.

luckysori:
> @luckysori I am getting the feeling that we probably want newtypes in the library for this kind of stuff. We don't need to enforce any invariants using these newtypes though.

Would the idea be to still send around "raw" types, and then allow users to construct one such newtype to call a method on it?

thomaseizinger:
> > @luckysori I am getting the feeling that we probably want newtypes in the library for this kind of stuff. We don't need to enforce any invariants using these newtypes though.
> 
> Would the idea be to still send around "raw" types, and then allow users to construct one such newtype to call a method on it?

I think we may just want to construct these new-types directly, just to make things a bit more ergonomic but have them "open" to not restrict the user in what they can do with it.

bonomat:
@thomaseizinger 

> We don't have to put these types into cfd_protocol right away, they can also live within daemon. @bonomat are you interested in attempting this? The basic idea is to have things like:

I think I follow your idea but I'm not sure where to start and which types to replace. At this point I don't feel confident enough to start refactoring things. 

One thing I noticed though is that the type of `CfdTransactions.cets` is different to `Msg1.cets` :)

thomaseizinger (approved):
LGTM!

https://www.youtube.com/watch?v=rR4n-0KYeKQ

thomaseizinger:
> @luckysori I am getting the feeling that we probably want newtypes in the library for this kind of stuff. We don't need to enforce any invariants using these newtypes though.

I just tried this and couldn't come up with anything satisfactory :(

PR #53, opened 2021-09-14
Add wallet to maker and taker

thomaseizinger:
For now we have different, static descriptors with a hardcoded DB
path. We also only operate on testnet and use sled as a database.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -1,12 +1,14 @@
+use crate::seed::Seed;
 use anyhow::Result;
 use bdk::bitcoin::secp256k1::{schnorrsig, SECP256K1};
-use bdk::bitcoin::{self, Amount};
+use bdk::bitcoin::{Amount, Network};
 use bdk::blockchain::{ElectrumBlockchain, NoopProgress};
+use bdk::KeychainKind;
+use clap::Clap;
 use model::cfd::{Cfd, Order};
 use rocket::fairing::AdHoc;
-use rocket::figment::util::map;
-use rocket::figment::value::{Map, Value};
 use rocket_db_pools::Database;
+use std::path::PathBuf;
 use tokio::sync::{mpsc, watch};
 
 mod db;
@@ -15,6 +17,7 @@ mod maker_cfd_actor;
 mod maker_inc_connections_actor;
 mod model;
 mod routes_maker;
+mod seed;
 mod send_wire_message_actor;
 mod to_sse_event;
 mod wire;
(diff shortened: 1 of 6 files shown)


daemon/src/seed.rs
```diff
@@ -0,0 +1,73 @@
...
+pub struct Seed([u8; 256]);
+
+impl Seed {
+    /// Initialize a [`Seed`] from a path.
+    ///
+    /// Fails if the file does not exist or it exists but would be overwritten.
+    pub async fn initialize(seed_file: &Path, generate: bool) -> Result<Seed> {
+        let exists = seed_file.exists();
+
+        let seed = match (exists, generate) {
```
klochowicz, line 19: out of curiosity: would placing `seed_file.exists()` here directly cause any difference, e.g. would that we evaluated multiple times when trying to match different arms?

daemon/src/maker.rs
```diff
@@ -23,19 +26,54 @@ mod wire;
 #[database("maker")]
 pub struct Db(sqlx::SqlitePool);
 
+#[derive(Clap)]
+struct Opts {
+    /// The port to listen on for p2p connections.
```
klochowicz, line 31: wow, I've just realised that Clap parses these comments into the CLI interface. super cool! 😎

daemon/src/taker.rs
```diff
@@ -23,19 +27,55 @@ mod wire;
...
+        tokio::fs::create_dir_all(&data_dir).await?;
+    }
+
+    let seed = Seed::initialize(&data_dir.join("seed"), opts.generate_seed).await?;
+
+    let client = bdk::electrum_client::Client::new(&opts.electrum).unwrap();
 
     // TODO: Replace with sqlite once https://github.com/bitcoindevkit/bdk/pull/376 is merged.
-    let db = bdk::sled::open("/tmp/taker.db")?;
+    let db = bdk::sled::open(data_dir.join("wallet_db"))?;
```
klochowicz: I believe you should maintain unique wallet db names for both sides - there's no enforcement of specifying different data dirs for maker and the taker at the moment, and after this changeset you might end up trying to open the same wallet db on both ends.
```suggestion
    let db = bdk::sled::open(data_dir.join("taker_wallet_db"))?;
```

daemon/src/maker.rs
```diff
@@ -23,19 +26,54 @@ mod wire;
...
+        tokio::fs::create_dir_all(&data_dir).await?;
+    }
+
+    let seed = Seed::initialize(&data_dir.join("seed"), opts.generate_seed).await?;
+
+    let client = bdk::electrum_client::Client::new(&opts.electrum).unwrap();
 
     // TODO: Replace with sqlite once https://github.com/bitcoindevkit/bdk/pull/376 is merged.
-    let db = bdk::sled::open("/tmp/maker.db")?;
+    let db = bdk::sled::open(data_dir.join("wallet_db"))?;
```
klochowicz: as per other comment:

```suggestion
    let db = bdk::sled::open(data_dir.join("taker_wallet_db"))?;
```

daemon/src/taker.rs
thomaseizinger: Fair enough, also need to change the name of the seed file then!

daemon/src/seed.rs
thomaseizinger: Nope, would only be evaluated once AFAIK!

It just liked the symmetry of matching against two variables instead of one variable and one expression.

klochowicz (approved):
I can't commend on wallet initialisation intricacies (e.g. choice of template), but other that that, it looks good!

thomaseizinger:
> I can't commend on wallet initialisation intricacies (e.g. choice of template), but other that that, it looks good!

[BIP84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) is the derivation scheme for native segwit accounts. Using BIP84 allows us to produce the most space-efficient and cost-effective Bitcoin transactions.

klochowicz:
> > I can't commend on wallet initialisation intricacies (e.g. choice of template), but other that that, it looks good!
> 
> [BIP84](https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki) is the derivation scheme for native segwit accounts. Using BIP84 allows us to produce the most space-efficient and cost-effective Bitcoin transactions.

thank you for the clarification and the link,  I'll have a read!

PR #57, opened 2021-09-15
Add contract setup and signature verification on the maker side

klochowicz:
Extract the functionality into its own module, as the logic was exactly the same
on both sides.

I've introduced 2 new concepts:
     - "own" - the caller of the function, it's either taker or maker,
        depending on the OwnParams parameter,
     - "other" - the other party, the opposite role.

XXX: Should I rather follow the naming convention in Origin enum (Ours/Theirs?)

### DIFF
--- a/daemon/src/setup_contract_actor.rs
+++ b/daemon/src/setup_contract_actor.rs
@@ -1,0 +1,310 @@
+use crate::model::cfd::{AsBlocks, FinalizedCfd, Order};
+use crate::wire::{AdaptorSignature, Msg0, Msg1, SetupMsg};
+use anyhow::{Context, Result};
+use bdk::bitcoin::secp256k1::{schnorrsig, SecretKey, Signature, SECP256K1};
+use bdk::bitcoin::{Amount, PublicKey, Transaction, Txid};
+use bdk::descriptor::Descriptor;
+use cfd_protocol::{
+    commit_descriptor, compute_signature_point, create_cfd_transactions, lock_descriptor,
+    spending_tx_sighash, EcdsaAdaptorSignature, PartyParams, PunishParams,
+};
+use futures::Future;
+use std::collections::HashMap;
+use tokio::sync::mpsc;
+
+/// A factor to be added to the CFD order term for calculating the refund timelock.
+///
+/// The refund timelock is important in case the oracle disappears or never publishes a signature.
+/// Ideally, both users collaboratively settle in the refund scenario. This factor is important if
+/// the users do not settle collaboratively.
+/// `1.5` times the term as defined in CFD order should be safe in the extreme case where a user
+/// publishes the commit transaction right after the contract was initialized. In this case, the
+/// oracle still has `1.0 * cfdorder.term` time to attest and no one can publish the refund
+/// transaction.
+/// The downside is that if the oracle disappears: the users would only notice at the end
+/// of the cfd term. In this case the users has to wait for another `1.5` times of the
+/// term to get his funds back.
+pub const REFUND_THRESHOLD: f32 = 1.5;
+
+/// Given an initial set of parameters, sets up the CFD contract with the other party.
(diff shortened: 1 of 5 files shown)


daemon/src/contract.rs
```diff
@@ -0,0 +1,309 @@
...
+    Taker(PartyParams),
+}
+
+enum Role {
+    Maker,
+    Taker,
+}
+
+/// A convenience struct for storing PartyParams and PunishParams for both
+/// parties and the
```
thomaseizinger: Only half a sentence :)

daemon/src/contract.rs
```diff
@@ -0,0 +1,309 @@
...
+pub fn setup_contract(
+    send_to_other: impl Fn(SetupMsg),
+    own_params: OwnParams,
+    sk: SecretKey,
+    oracle_pk: schnorrsig::PublicKey,
+    order: Order,
+) -> (
+    impl Future<Output = FinalizedCfd>,
+    mpsc::UnboundedSender<SetupMsg>,
+) {
```
thomaseizinger: As a general theme, try to put the public API of a module at the very top and order things from highest to lowest level of importance. That way, as you start reading the code, you know what this module does and by continuing to read, you will learn about the details how it is implemented. Assuming descriptive names, just reading the top-level public function may give you enough information about what this is doing without having to scan down further!

daemon/src/contract.rs
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+/// Given an initial set of parameters, sets up the CFD contract with the other party.
+/// Passing OwnParams identifies whether caller is the maker or the taker.
+///
+/// Returns the [`FinalizedCfd`] which contains the lock transaction, ready to be signed and sent to
+/// the other party. Signing of the lock transaction is not included in this function because we
+/// want the Cfd actor to own the wallet.
+pub fn setup_contract(
```
thomaseizinger: To align this with our naming, we could call this `new` and the file `setup_contract_actor`.

thomaseizinger (approved):
Nice! I was tempted to do this yesterday but wanted to wait until everything is merged :D

Thanks for unifying this!

A few nits but otherwise LGTM!

daemon/src/contract.rs
klochowicz: oops! 😳 thanks for catching that!

daemon/src/contract.rs
klochowicz: that makes a lot of sense! will change things around to reflect this.

klochowicz: @thomaseizinger I've read somewhere (in the rustbook?) that it's customary for `const` to  be at the top of the file - shall I keep the `REFUND_TRESHOLD` above the `fn new()` ?

thomaseizinger: Yeah consts can stay at the top :)

thomaseizinger (approved):
Thanks!

daemon/src/setup_contract_actor.rs
```diff
@@ -0,0 +1,310 @@
...
+
+fn verify_cets(
+    oracle_pk: &schnorrsig::PublicKey,
+    maker: &PartyParams,
+    taker_cets: &[(
+        Transaction,
+        EcdsaAdaptorSignature,
+        Vec<u8>,
+        schnorrsig::PublicKey,
+    )],
```
bonomat: Nit: since this function is called by both parties now, it's not `maker: PartyParams` and `taker_cets` but `other: PartyParams` and `own_cets...`
Same goes for the usage of `maker_encsig` in the function body.

bonomat (approved):
Looks good. Nit on some parameter naming which could be changed, but otherwise, it looks identical to what's been before :)

daemon/src/setup_contract_actor.rs
klochowicz: indeed, I missed that one - fixed!

PR #59, opened 2021-09-15
Run ci smoke tests on armv7

bonomat:
Imho it's a common assumption that our tool will run on a RPi and we should ensure that it keeps building for this platform.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,0 +1,4 @@
+[toolchain]
+channel = "1.55"
+components = ["clippy"]
+targets = ["armv7-unknown-linux-gnueabihf"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.armv7-unknown-linux-gnueabihf]
+linker = "arm-linux-gnueabihf-gcc"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,25 +53,36 @@ jobs:
             os: ubuntu-latest
           - target: x86_64-apple-darwin
             os: macos-latest
+          - target: armv7-unknown-linux-gnueabihf
+            os: ubuntu-latest
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
-        with:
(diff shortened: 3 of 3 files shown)


thomaseizinger (changes requested):
This patch doesn't actually do anything useful because I've optimized things away in https://github.com/comit-network/hermes/commit/b5fe043ca85eada2a712dc5ec67a9dce28dab9c7 :sweat_smile: 

We will need to add usage of ${{ matrix.target }} again!

bonomat:
Converted to draft because something is not working: 
- the project builds on arm
- the tests run on arm
- executing the binaries fails (see last commit)

Any idea?

thomaseizinger:
> Any idea?

You can't run an ARM binary on a x64 processor. You would need to start up an emulator like QEMU to do that. In the past we simply skipped the smoke test for certain targets.

bonomat:
> > Any idea?
> 
> You can't run an ARM binary on a x64 processor. You would need to start up an emulator like QEMU to do that. In the past we simply skipped the smoke test for certain targets.


Oh makes sense, I didn't know that all targets run on `x86_64` architectures.

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.54"
```
thomaseizinger: 1.55 is the latest release, any reason why you want 1.54 here?

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
```
thomaseizinger, line 1: If we have a `rust-toolchain.toml` file, it would be good to do an explicit install step instead of "implying" this with the first call to cargo. Reason being that the cache action works out the cache key based on the [installed rust version](https://github.com/Swatinem/rust-cache#cache-details) and that would happen before the first cargo invocation, effectively creating the wrong cache key for what is actually being used.

See https://github.com/actions-rs/toolchain/issues/126 + recommended solution.

thomaseizinger (changes requested):
Thanks!

A question + suggestion!

rust-toolchain.toml
bonomat: No particular reason. Changed to 1.55 and updated locally :)

thomaseizinger (approved):

PR #61, opened 2021-09-16
Wait for the maker if it is not online

klochowicz:
Taker now prints a message if maker was not online, and waits for preconfigured
interval before trying again.


thomaseizinger (approved):
Nice! Once we have xtra actors this might get even better :)

da-kami (approved):

PR #78, opened 2021-09-16
Add script for preparing the DB for convenience

da-kami:


daemon/prepare_db.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+DAEMON_DIR=daemon
+TEMPDB=tempdb
+
+# crated temporary DB
```
klochowicz: nit: typo `create'

klochowicz: nit: typo `create'

daemon/prepare_db.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
klochowicz, line 1: nit: I think you forgot to `chmod +x` this script

klochowicz (approved):
nice, it's always easier to maintain and use a script 👍 

shall we point to the script in the README so devs could use that instead of following the manual process?

daemon/prepare_db.sh
```diff
@@ -0,0 +1,15 @@
...
+# create temporary DB
+DATABASE_URL=sqlite:$TEMPDB cargo sqlx database create
+
+# run the migration scripts to create the tables
+DATABASE_URL=sqlite:$TEMPDB cargo sqlx migrate run
+
+# prepare the sqlx-data.json rust mappings
+DATABASE_URL=sqlite:./$DAEMON_DIR/$TEMPDB cargo sqlx prepare -- --bin taker
+
+rm $TEMPDB
```
thomaseizinger: You may want to do is in a `trap` to make sure it always runs on exit.

thomaseizinger (approved):
Thanks!

Should we specify `set -e` to bail early if something fails?

PR #79, opened 2021-09-16
 Remove hardcoded absolute URLs from the frontend source code 

### LINKED ISSUE #6: Configure proxy to allow relative API paths in frontend source code
This is important to fix because it currently forces us to hardcode `localhost` and port in the frontend soure-code!

In production, the frontend will be served from the same port as the API, hence the code needs to use relative paths to hit the correct endpoints.

thomaseizinger:
See commit message for details on the solution.

Fixes #6.


bonomat (approved):
And the award for longest commit message ever goes to 🥁🥁🥁 📣@thomaseizinger 📣

da-kami (changes requested):
I ran both daemons and `APP=taker yarn run dev` but the UI does not come up. It is just an empty white screen when navigating to `localhost:3000`

frontend/vite.config.ts
```diff
@@ -11,40 +22,17 @@ export default defineConfig({
...
         rollupOptions: {
-            input: {
-                maker: resolve(__dirname, "maker.html"),
-                taker: resolve(__dirname, "taker.html"),
-            },
+            input: resolve(__dirname, `index.html`),
         },
+        outDir: `dist/${app}`,
     },
     server: {
```
klochowicz, line 33:
```suggestion
    server: {
      open: "/trade",
```

klochowicz: if you don't mind, the `open` call was nice as it gave us the same behavior as we had before introducing vite (open the browser when starting the app). 
It could be either `/` or `/trade` (I'd vote the latter, as we don't forward to `/trade` by default - maybe we could?)

dprint.json
```diff
@@ -7,7 +7,7 @@
     "wrap_comments": true,
     "comment_width": 120
   },
-  "includes": ["**/*.{md,rs,toml}"],
+  "includes": ["**/*.{md,rs,toml,ts,tsx,js}"],
```
klochowicz, line 10: good catch!

klochowicz (approved):

frontend/vite.config.ts
thomaseizinger: I will do `/` to encourage you to fix the app such that it also has a convenient homepage when the user is going to see it :)

klochowicz:
> I ran both daemons and `APP=taker yarn run dev` but the UI does not come up. It is just an empty white screen when navigating to `localhost:3000`

I think taker is on port `3001` (if @thomaseizinger applies my other suggestion, it would get started automatically, preventing the confusion)

FWIW I ran both UIs concurrently with the daemons and it worked.

frontend/vite.config.ts
klochowicz: fair enough :)

da-kami (approved):
I am not sure why it did not load properly. After restarting Clion and Firefox it worked smoothly (also with using my Run configurations). Maybe there was some caching issues. 
🤷‍♂️

Thanks for figuring this setup out @thomaseizinger - bit tricky indeed :)

da-kami:
> > I ran both daemons and `APP=taker yarn run dev` but the UI does not come up. It is just an empty white screen when navigating to `localhost:3000`
> 
> I think taker is on port `3001` (if @thomaseizinger applies my other suggestion, it would get started automatically, preventing the confusion)
> 
> FWIW I ran both UIs concurrently with the daemons and it worked.

Thanks for the suggestion, but the port was definitely not the problem. I'm not sure why it did not load properly, but I fixed it by restarting both IDE and Browser. I reckon it might have been some browser cache issues.

PR #80, opened 2021-09-16
Use the actual margin

### LINKED ISSUE #62: Use margin to make `build_party_params` pass
Currently the taker panics because we pass a `ZERO` Bitcoin amount when building params.

Make construction of party params work.

- Load the margin from DB
- Update the CFD state after successful contract setup
- Make building the params pass on both sides

da-kami:
fixes #62

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1,8 +1,7 @@
 use crate::model::cfd::{Cfd, CfdState, Order, OrderId, Origin};
 use crate::model::{Leverage, Position};
-use anyhow::Context;
+use anyhow::{Context, Result};
 use rocket_db_pools::sqlx;
-
 use sqlx::pool::PoolConnection;
 use sqlx::{Acquire, Sqlite, SqlitePool};
 use std::convert::TryInto;
@@ -244,7 +243,7 @@ async fn load_latest_cfd_state(
 pub async fn load_cfd_by_order_id(
     order_id: OrderId,
     conn: &mut PoolConnection<Sqlite>,
-) -> anyhow::Result<Cfd> {
+) -> Result<Cfd> {
     let order_uuid = serde_json::to_string(&order_id)?;
 
     let row = sqlx::query!(
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -441,7 +441,6 @@ mod tests {
             },
         );
 
-        // the order ahs to exist in the db in order to be able to insert the cfd
         insert_order(&order, &mut conn).await.unwrap();
         insert_cfd(cfd.clone(), &mut conn).await.unwrap();
 
@@ -468,7 +467,6 @@ mod tests {
 
         let order_id = order.id;
 
-        // the order ahs to exist in the db in order to be able to insert the cfd
         insert_order(&order, &mut conn).await.unwrap();
         insert_cfd(cfd.clone(), &mut conn).await.unwrap();
 
@@ -492,7 +490,6 @@ mod tests {
             },
         );
 
-        // the order ahs to exist in the db in order to be able to insert the cfd
         insert_order(&order, &mut conn).await.unwrap();
         insert_cfd(cfd.clone(), &mut conn).await.unwrap();
 
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -241,6 +241,82 @@ async fn load_latest_cfd_state(
     Ok(latest_cfd_state_in_db)
 }
 
+pub async fn load_cfd_by_order_id(
+    order_id: OrderId,
+    conn: &mut PoolConnection<Sqlite>,
+) -> anyhow::Result<Cfd> {
+    let order_uuid = serde_json::to_string(&order_id)?;
+
+    let row = sqlx::query!(
+        r#"
+        select
+            cfds.id as cfd_id,
+            orders.uuid as order_id,
+            orders.initial_price as price,
+            orders.min_quantity as min_quantity,
+            orders.max_quantity as max_quantity,
+            orders.leverage as leverage,
+            orders.trading_pair as trading_pair,
+            orders.position as position,
+            orders.origin as origin,
+            orders.liquidation_price as liquidation_price,
+            orders.creation_timestamp as creation_timestamp,
+            orders.term as term,
+            cfds.quantity_usd as quantity_usd,
+            cfd_states.state as state
+        from cfds as cfds
+        inner join orders as orders on cfds.order_uuid = ?
+        inner join cfd_states as cfd_states on cfd_states.cfd_id = cfds.id
+        where cfd_states.state in (
+            select
+              state
+              from cfd_states
+            where cfd_id = cfds.id
+            order by id desc
+            limit 1
+        )
+        "#,
+        order_uuid
+    )
+    .fetch_one(conn)
+    .await?;
+
+    let order_id = serde_json::from_str(row.order_id.as_str()).unwrap();
+    let trading_pair = serde_json::from_str(row.trading_pair.as_str()).unwrap();
+    let position: Position = serde_json::from_str(row.position.as_str()).unwrap();
+    let price = serde_json::from_str(row.price.as_str()).unwrap();
+    let min_quantity = serde_json::from_str(row.min_quantity.as_str()).unwrap();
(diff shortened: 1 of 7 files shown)


da-kami:
@bonomat @thomaseizinger when implementing https://github.com/comit-network/hermes/commit/ef773e62a297f5064621bb5abf355dd8aa7e6150 I came to the conclusion that we could actually remove the amounts from the data that is exchanged between the two parties. 
The lock amount is `margin_taker` + `margin_maker` - which can be deterministically calculated on both sides. Is there a specific reason why the margin amounts are exchanged? Should I refactor this and remove them (and just use the calculation)?

daemon/src/db.rs
```diff
@@ -378,6 +445,35 @@ mod tests {
...
+                common: CfdStateCommon {
+                    transition_timestamp: SystemTime::now(),
+                },
+            },
+            Position::Buy,
+        );
+
+        let order_id = order.id;
+
+        // the order ahs to exist in the db in order to be able to insert the cfd
```
bonomat:
```suggestion
        // the order has to exist in the db in order to be able to insert the cfd
```

daemon/src/db.rs
```diff
@@ -249,6 +249,73 @@ async fn load_latest_cfd_state(
...
+    let latest_state = serde_json::from_str(row.state.as_str()).unwrap();
+    let term = serde_json::from_str(row.term.as_str()).unwrap();
+
+    let origin: Origin = serde_json::from_str(row.origin.as_str()).unwrap();
+    let position: Position = serde_json::from_str(row.position.as_str()).unwrap();
+
+    let position = match origin {
+        Origin::Ours => position,
+        Origin::Theirs => position.counter_position(),
+    };
```
bonomat: This fucks with my head and I had to think this through twice 😅

We store an `order` with `position::sell` and load it (still has `position::sell`) and then return a `cfd` with `position::buy` if we are the taker. 

Imho the struct `Cfd` deserves some rust doc on this field or rename it to `own_position`, otherwise I can guarantee you that I will confuse things sometimes in the future.

bonomat:
> @bonomat @thomaseizinger when implementing [ef773e6](https://github.com/comit-network/hermes/commit/ef773e62a297f5064621bb5abf355dd8aa7e6150) I came to the conclusion that we could actually remove the amounts from the data that is exchanged between the two parties.
> The lock amount is `margin_taker` + `margin_maker` - which can be deterministically calculated on both sides. Is there a specific reason why the margin amounts are exchanged? Should I refactor this and remove them (and just use the calculation)?

I'm not sure I fully understand your question: 
Are you asking if you can remove `Msg0::lock_amount` which represents the other party's amount the promise to _lock up_? Because both sides can deterministically compute these amounts?  

If so, I'm not sure this simplifies things.

da-kami:
> > @bonomat @thomaseizinger when implementing [ef773e6](https://github.com/comit-network/hermes/commit/ef773e62a297f5064621bb5abf355dd8aa7e6150) I came to the conclusion that we could actually remove the amounts from the data that is exchanged between the two parties.
> > The lock amount is `margin_taker` + `margin_maker` - which can be deterministically calculated on both sides. Is there a specific reason why the margin amounts are exchanged? Should I refactor this and remove them (and just use the calculation)?
> 
> I'm not sure I fully understand your question:
> Are you asking if you can remove `Msg0::lock_amount` which represents the other party's amount the promise to _lock up_? Because both sides can deterministically compute these amounts?
> 
> If so, I'm not sure this simplifies things.

I think if both parties can deterministically calculate it, it does not have to be sent back and forth and verified. 
If both calculate they should come up with the same result - so when signing it should work as long as the calculation is deterministic (and neither party is cheating). There is no extra verification and sending the amount over needed.
That would be my reasoning.

daemon/src/db.rs
da-kami: The `Cfd` could also be refactored to just contain the `Order` - that might make it more clear. At the moment the `Order` is flattened out in the `Cfd` which is confusing because we have to set it when loading. If we would know the complete `Order` (including `Origin`) in the `Cfd` then the `position` would always just be determined in the `Cfd` and it would be clear I think.
I will see to get that together later.

da-kami: https://github.com/comit-network/hermes/pull/82

da-kami:
> > > @bonomat @thomaseizinger when implementing [ef773e6](https://github.com/comit-network/hermes/commit/ef773e62a297f5064621bb5abf355dd8aa7e6150) I came to the conclusion that we could actually remove the amounts from the data that is exchanged between the two parties.
> > > The lock amount is `margin_taker` + `margin_maker` - which can be deterministically calculated on both sides. Is there a specific reason why the margin amounts are exchanged? Should I refactor this and remove them (and just use the calculation)?
> > 
> > 
> > I'm not sure I fully understand your question:
> > Are you asking if you can remove `Msg0::lock_amount` which represents the other party's amount the promise to _lock up_? Because both sides can deterministically compute these amounts?
> > If so, I'm not sure this simplifies things.
> 
> I think if both parties can deterministically calculate it, it does not have to be sent back and forth and verified.
> If both calculate they should come up with the same result - so when signing it should work as long as the calculation is deterministic (and neither party is cheating). There is no extra verification and sending the amount over needed.
> That would be my reasoning.

Since this change is not critical I will ignore it for now and move on to funding accounts to get the punch through further. 

For me this PR is ready to merge.

daemon/src/db.rs
```diff
@@ -241,6 +241,82 @@ async fn load_latest_cfd_state(
     Ok(latest_cfd_state_in_db)
 }
 
+pub async fn load_cfd_by_order_id(
+    order_id: OrderId,
+    conn: &mut PoolConnection<Sqlite>,
+) -> anyhow::Result<Cfd> {
```
thomaseizinger: Nit: We usually _import_ `anyhow::Result` these days.

daemon/src/db.rs
```diff
@@ -241,6 +241,82 @@ async fn load_latest_cfd_state(
...
+            order by id desc
+            limit 1
+        )
+        "#,
+        order_uuid
+    )
+    .fetch_one(conn)
+    .await?;
+
+    let order_id = serde_json::from_str(row.order_id.as_str()).unwrap();
```
thomaseizinger, line 283: We may want to implement `Encode` & `Decode` for some of these types to remove these conversions.

https://docs.rs/sqlx/0.5.7/sqlx/trait.Decode.html

There is also a custom-derive that one can use! https://docs.rs/sqlx/0.5.7/sqlx/derive.Decode.html

thomaseizinger: If all fields implement decode, we may even be able to use `query_as` directly!

daemon/src/model/cfd.rs
```diff
@@ -277,6 +277,17 @@ impl Cfd {
         Ok(margin)
     }
 
+    pub fn calc_counterparty_margin(&self) -> Result<Amount> {
```
thomaseizinger: Not sure if the `calc` prefix adds much. The fact that this is a function already sort of implies that it is a computed property.

daemon/src/setup_contract_actor.rs
```diff
@@ -66,11 +66,15 @@ pub fn new(
...
 
+        if params.other.lock_amount != cfd.calc_counterparty_margin().unwrap() {
+            panic!("Sorry, have to panic 😬 - the amounts that the counterparty sent were wrong, expected {} actual {}", cfd.calc_counterparty_margin().unwrap(), params.other.lock_amount)
+        }
+
         let own_cfd_txs = create_cfd_transactions(
             (params.maker().clone(), *params.maker_punish()),
             (params.taker().clone(), *params.taker_punish()),
             oracle_pk,
-            order.term.mul_f32(REFUND_THRESHOLD).as_blocks().ceil() as u32,
```
thomaseizinger, line 73: This math may be better defined on `Cfd` itself?

thomaseizinger (approved):

daemon/src/db.rs
da-kami: I recorded this one in a refactoring ticket: https://github.com/comit-network/hermes/issues/88
I did not add this ticket to any epic so far as I don't think it's critical. We could still add it to the project though and just do it at some point.

PR #81, opened 2021-09-17
Construct `Payout` based on interval

### LINKED ISSUE #68: Define and implement compression/mapping of prices into signature points
Having the oracle attest to each individual price point results in an explosion of signatures. We need to define and implement a more clever solution to this, likely along the lines of what is specified by dlc-specs.

luckysori:
Fixes #68.

This is now closer to what I think we should do when adding the feature of supporting an oracle/protocol combination with payouts based on price intervals (rather than concrete prices).

I will very likely end up squashing a lot of this, but I think this can still be reviewed in its current state **patch by patch** (probably too much going on overall). Particularly to @thomaseizinger, who reviewed it the first time, https://github.com/comit-network/hermes/pull/81/commits/4d9727f79b7fdc467be7bda7b7dc248a80b470e2 is the one patch which may be confusing, as I got carried away and it was getting late.


cfd_protocol/src/interval.rs
```diff
@@ -0,0 +1,124 @@
...
+use std::ops::RangeInclusive;
+
+mod digit_decomposition;
+
+const BASE: usize = 2;
+
+/// Maximum number of binary digits for BTC price in whole USD.
+const MAX_DIGITS: usize = 20;
+
+const MAX_PRICE_DEC: u64 = (BASE as u64).pow(MAX_DIGITS as u32);
```
thomaseizinger: `const` :heart_eyes:

cfd_protocol/src/interval.rs
```diff
@@ -0,0 +1,124 @@
...
+mod tests {
+    use super::*;
+    use proptest::prelude::*;
+
+    impl PartialEq<Vec<Digits>> for Interval {
+        fn eq(&self, other: &Vec<Digits>) -> bool {
+            let sub_intervals = other.iter().flat_map(|i| i.range());
+            sub_intervals.eq(self.0.clone())
+        }
+    }
```
thomaseizinger, line 142: Wouldn't it be clearer if that were part of the actual test?

cfd_protocol/src/interval.rs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        if start > end {
+            bail!("invalid interval: start > end")
+        }
+
+        Ok(Self(start..=end))
+    }
+
+    pub fn as_digits(&self) -> Vec<Digits> {
```
thomaseizinger: This allocates :upside_down_face: 

```suggestion
    pub fn to_digits(&self) -> Vec<Digits> {
```

cfd_protocol/src/interval.rs
```diff
@@ -0,0 +1,124 @@
...
+            MAX_DIGITS,
+        )
+        .iter()
+        .map(|digits| {
+            let digits = digits.iter().map(|n| *n != 0).collect::<BitVec>();
+            Digits(digits)
+        })
+        .collect()
+    }
+}
```
thomaseizinger: It looks like the public API of this module is essentially: `(start, end) -> Vec<Digits>`

You could expose that as a free function and make `Interval` private (or even inline it if deemed unnecessary).

cfd_protocol/src/interval.rs
```diff
@@ -0,0 +1,124 @@
...
+use bit_vec::BitVec;
+use std::fmt::Display;
+use std::ops::RangeInclusive;
+
+mod digit_decomposition;
+
+const BASE: usize = 2;
+
+/// Maximum number of binary digits for BTC price in whole USD.
+const MAX_DIGITS: usize = 20;
```
thomaseizinger, line 11: Bit more optimism Lucas!

cfd_protocol/tests/cfds.rs
```diff
@@ -833,50 +862,50 @@ where
...
+/// Represents the oracle's commitment to a set of nonces that will be used to
+/// sign each digit of the price of BTC in USD at a specific time in the future.
 struct Event {
-    /// Nonce.
+    /// Nonces.
     ///
     /// Must remain secret.
-    nonce: SecretKey,
-    nonce_pk: schnorrsig::PublicKey,
+    nonces: Vec<SecretKey>,
```
thomaseizinger: Should we make this a fixed-size array?

cfd_protocol/src/interval.rs
luckysori: I promise I will not mess this up one day.

cfd_protocol/tests/cfds.rs
```diff
@@ -877,16 +873,7 @@ impl Event {
...
-            .map(|_| {
-                let nonce = SecretKey::new(rng);
-
-                let key_pair = schnorrsig::KeyPair::from_secret_key(SECP256K1, nonce);
-                let nonce_pk = schnorrsig::PublicKey::from_keypair(SECP256K1, &key_pair);
-
-                (nonce, nonce_pk)
-            })
-            .unzip();
+        let (nonces, nonce_pks) = (0..20).map(|_| oracle::nonce(rng)).unzip();
```
thomaseizinger: Pure elegance.

cfd_protocol/tests/cfds.rs
```diff
@@ -265,8 +264,8 @@ fn renew_cfd() {
     )
 }
 
-fn create_cfd_txs(
-    rng: &mut ChaChaRng,
+fn create_cfd_txs<R>(
+    rng: &mut R,
```
thomaseizinger: Could use `impl RngCore + CryptoRng` to save on the `where` clause.

cfd_protocol/tests/cfds.rs
```diff
@@ -843,6 +920,20 @@ impl Oracle {
...
         oracle::attest(&self.key_pair, nonce, msg)
     }
+
+    fn attest_price(
+        &self,
+        price: u64,
+        nonces: &[SecretKey; Self::MAX_DIGITS],
+    ) -> Vec<schnorrsig::Signature> {
+        let mut bits = BitVec::from_bytes(&price.to_be_bytes());
+        let bits = bits.split_off(bits.len() - 20);
```
thomaseizinger: Is this `20` related to `MAX_DIGITS`? Should use the constant for documentation purposes at least.

cfd_protocol/src/protocol/sighash_ext.rs
```diff
@@ -0,0 +1,15 @@
+use bdk::bitcoin::hashes::Hash;
+use bdk::bitcoin::SigHash;
+
+pub(super) trait SigHashExt {
```
thomaseizinger: Why `pub(super)` specifically?

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -0,0 +1,441 @@
+use crate::protocol::sighash_ext::SigHashExt;
+use crate::protocol::transaction_ext::TransactionExt;
+use crate::protocol::{
+    commit_descriptor, compute_adaptor_point, lock_descriptor, Payout, DUMMY_2OF2_MULTISIG,
+};
+
```
thomaseizinger: Would prefer no empty lines in imports. Prevents rustfmt from re-arranging all of these.

cfd_protocol/src/protocol/sighash_ext.rs
thomaseizinger: I think `pub(crate)` would be sufficient no? (And is stable for moving things around).

cfd_protocol/src/protocol.rs
```diff
@@ -344,12 +345,14 @@ pub struct Payout {
 
 impl Payout {
     pub fn new(
-        interval: Interval,
+        start: u64,
+        end: u64,
```
thomaseizinger: You could take a `Range` maybe?

cfd_protocol/src/protocol.rs
```diff
@@ -0,0 +1,532 @@
...
+            lock_amount: amount,
+            address,
+        })
+    }
+}
+
+pub fn create_cfd_transactions(
+    (maker, maker_punish_params): (PartyParams, PunishParams),
+    (taker, taker_punish_params): (PartyParams, PunishParams),
+    (oracle_pk, nonce_pks): (schnorrsig::PublicKey, &[schnorrsig::PublicKey]),
```
thomaseizinger, line 83: The slice here represents the nonces per digit yes? I think we will at least need to documentation on whether that is in big or little endian order or create a data structure that encodes this somehow.

thomaseizinger (approved):
Only nits really. Thank you, was nice to review!

The implementation looks sound, I haven't checked it in super high detail but I think the APIs makes sense and I trust that the tests are good :)

cfd_protocol/tests/cfds.rs
luckysori: Did I never know that this was equivalent? Thanks!

cfd_protocol/src/protocol/sighash_ext.rs
luckysori: Good point. I don't know what possessed me to suddenly prefer `pub(super)` over `pub(crate)`.

luckysori:
Instead of squashing which would have been either really hard and good or really easy and destructive in terms of history, I decided to ensure that all patches pass the following checks:

```
cargo clippy -- -D warnings
dprint check
cargo test -p cfd_protocol
```

I only ran `cfd_protocol` tests because I think all the tests in `daemon` are unaffected by these changes anyway.

PR #82, opened 2021-09-17
Refactor Cfd to be based on Order

da-kami:
Flattening the Cfd out in the internal mode has caused duplication and confusion.
This refactor puts the Order into the Cfd to make relations clearer.

Note that once we have multiple leverage the leverage and liquidation price will have to move into the Cfd because they depend on the user's choice then.
This should be trivial to do.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -26,6 +26,12 @@ impl Display for OrderId {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum Origin {
+    Ours,
+    Theirs,
+}
+
 /// A concrete order created by a maker for a taker
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Order {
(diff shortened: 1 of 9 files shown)


daemon/src/taker_inc_message_actor.rs
```diff
@@ -19,6 +20,14 @@ pub fn new(
         while let Some(message) = messages.next().await {
             match message {
                 Ok(wire::MakerToTaker::CurrentOrder(order)) => {
+                    let order = match order {
+                        None => None,
+                        Some(mut order) => {
+                            order.origin = Origin::Theirs;
+                            Some(order)
+                        }
+                    };
```
thomaseizinger: This can be an `if let` with `as_mut`.

thomaseizinger (approved):

daemon/src/model/cfd.rs
```diff
@@ -240,53 +250,45 @@ impl Display for CfdState {
...
+        let profit =
+            calculate_profit(self.order.price, current_price, dec!(0.005), Usd(dec!(0.1)))?;
         Ok(profit)
     }
+
+    pub fn position(&self) -> Position {
+        match self.order.origin {
+            Origin::Ours => self.order.position.clone(),
+            Origin::Theirs => self.order.position.counter_position(),
+        }
```
klochowicz: with the `position` logic now nicely encapsulated, we probably don't need to leak `counter_position` in its public interface - or are you planning to use it anywhere?

klochowicz (approved):
nice cleanup!

daemon/src/model/cfd.rs
da-kami: not for the moment, good catch, I'll remove it

PR #85, opened 2021-09-17
Wallet updates in the UI 🎉

### LINKED ISSUE #83: Testnet setup for maker and taker for building transactions
Building the transactions involves the need for funding (`InsufficientFunds`). 
The idea is to fund maker and taker and have two known seed that result in funded BTC wallets that will be shared.

da-kami:
fixes #83 

- wallet feed that sends balance + current address
- display in UI (balance + address) in separate wallet component (shared for taker and maker for now)

Includes two seed files that are already funded with some testnet coins. We can share those for testing for now. If more funds are needed I'm happy to top them up.

Not made beautiful yet, but functional:

image


daemon/src/maker.rs
```diff
@@ -80,13 +85,19 @@ async fn main() -> Result<()> {
...
     let (order_feed_sender, order_feed_receiver) = watch::channel::<Option<Order>>(None);
-    let (_balance_feed_sender, balance_feed_receiver) = watch::channel::<Amount>(Amount::ZERO);
+    let (wallet_feed_sender, wallet_feed_receiver) =
+        watch::channel::<WalletInfo>(WalletInfo::Syncing);
+
+    // initial wallet sync (blocking, so we do it before kicking things off)
+    wallet.sync(NoopProgress, None).unwrap(); // TODO: Use LogProgress once we have logging.
+
+    // TODO: We should get addresses on demand from a POST endpoint
+    let _address = wallet.get_address(AddressIndex::New).unwrap().address;
```
thomaseizinger: Should we do [`LastUnused`](https://docs.rs/bdk/0.11.0/bdk/wallet/enum.AddressIndex.html#variant.LastUnused) instead?

daemon/src/wallet.rs
```diff
@@ -0,0 +1,50 @@
...
+    wallet: &bdk::Wallet<B, D>,
+    previous: WalletInfo,
+) -> Result<Option<WalletInfo>, WalletError>
+where
+    B: Blockchain,
+    D: BatchDatabase,
+{
+    wallet
+        .sync(NoopProgress, None)
+        .map_err(|_| WalletError::Sync)?;
```
thomaseizinger: Technically, this could take a very long time and be blocking. We might want to put this on tokio's blocking task pool!

https://docs.rs/tokio/1.11.0/tokio/task/fn.spawn_blocking.html

daemon/src/wallet.rs
```diff
@@ -0,0 +1,50 @@
...
+        WalletInfo::Update {
+            balance: prev_balance,
+            ..
+        } => balance != prev_balance.as_sat(),
+        WalletInfo::Syncing | WalletInfo::Error(_) => true,
+    };
+
+    let wallet_info = if update {
+        let address = wallet
+            .get_address(AddressIndex::New)
```
thomaseizinger: Why not `LastUnused`?

daemon/src/wallet.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    let balance = wallet.get_balance().map_err(|_| WalletError::Balance)?;
+
+    let update = match previous {
+        WalletInfo::Update {
+            balance: prev_balance,
+            ..
+        } => balance != prev_balance.as_sat(),
+        WalletInfo::Syncing | WalletInfo::Error(_) => true,
+    };
```
thomaseizinger: Do we need this?

frontend/src/components/Wallet.tsx
```diff
@@ -0,0 +1,47 @@
...
+        walletInfo,
+    }: WalletProps,
+) {
+    let balance = <Spinner />;
+    let address = <Spinner />;
+
+    console.log(walletInfo);
+
+    if (walletInfo) {
+        if (walletInfo.type === "Error") {
```
thomaseizinger: Not sure but maybe we want a dedicated error feed actually.

daemon/src/model.rs
```diff
@@ -83,3 +84,21 @@ impl Display for TakerId {
...
+    Sync,
+    #[error("Wallet balance retrieval failure")]
+    Balance,
+    #[error("Wallet address retrieval failure")]
+    Address,
+}
+
+#[derive(Debug, Clone)]
+pub enum WalletInfo {
+    Syncing,
```
thomaseizinger: Does it bring much benefit to tell the user that we are syncing? What if we output a timestamp next to the wallet with "Last updated at"?

thomaseizinger (commented):
Thanks!

Some thoughts.

daemon/src/wallet.rs
da-kami: I am actually not sure in what scenario it would take long. I think it would only take long if someone uses an existing wallet. If we assume that the tool will always generate a wallet and one has to transfer funds into it it won't make much time. 
Additionally there is the initial sync when the daemon starts up, which can take longer (the user will see that it takes long...)
Still happy to give tokio blocking task pools a try, but I think it's actually not a must atm.

daemon/src/wallet.rs
da-kami: We only trigger this code path if either the balance changed or we ran into an Error or were syncing before.
We might want to distinguish those cases, but I felt we should not run into the Error and Syncing path too often (initially and upon error...), so I thought New would be find. Did not give it much extra though.

I we use `LastUsed`, wouldn't this result in address reuse?

daemon/src/wallet.rs
da-kami: That's the logic that determines if a wallet update should be sent or not. I thought this is what we had discussed :D
If the balance did not change we don't send and update into the feed (unless we were previously in an Error or Syncing state.
I think Syncing is actually only used initially now, so we could rename it (previously I had used it more, but that did not make sense).

frontend/src/components/Wallet.tsx
da-kami: Yeah, probably - not sure about that yet, but wanted to give this a try. Generally, dealing with the enum values is a bit of a pain (checking the type of the enum as string and then casting to payload is not optimal... - I did not have the brainpower to come up with something more elegant today ;)

daemon/src/model.rs
da-kami: I think it does somewhat help, but it is actually only used initially when we set it to be a `Spinner`. But given that the `Spinner` is actually not based on state but just an initial value we might as well simplify this.

daemon/src/maker.rs
da-kami: This line is actually a relic that should be removed 😅

daemon/src/wallet.rs
thomaseizinger: It is Last::**Un**used!

daemon/src/wallet.rs
thomaseizinger: Your code here is generic over the wallet backend. If it is a p2p backend, you need to download all blocks since you were last online, that can take hours!

If you want to make the assumption that sync doesn't take long because we are using esplora, then don't be generic over the backend :)

If you want to be generic over the backend, write the code in a way so it works with all backends!

daemon/src/model.rs
thomaseizinger: Our model of updates on the feed works best if you don't depend on the updates being sent in any particular order. Each update should map to a clear UI state.

We can do a "syncing" update but I find it a bit weird because the UI doesn't actually know about the state.
I would probably just send out the final update and add an animation everytime it changes, plus update the timestamp so the user knows it is up2date.

da-kami: One problem I see with that approach is, that if the wallet takes significant time for syncing the user just won't know and might assume that the balance is up to date (even though it might be outdated).
I remember we had problems with that in Ambrosia at some point - not sure that scenario applies here though. 
I can go with the timestamp, but I think the timestamp does not tell the user anything because there is no periodic balance change. So, the timestamp is no indicator for the balance being up to date or not. It just shows when the last change happened. I don't see that as something very useful.

thomaseizinger: If we remove the optimisation with sending out updates, the timestamp will refresh with your interval and hence accurately reflect, how often you are actually checking the balance internally.

daemon/src/maker_cfd_actor.rs
```diff
@@ -13,6 +14,7 @@ use tokio::sync::{mpsc, watch};
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 pub enum Command {
+    WalletSync,
```
thomaseizinger: Nit: `SyncWallet` sounds more like a command.

daemon/src/maker_cfd_actor.rs
```diff
@@ -37,11 +39,13 @@ pub fn new<B, D>(
...
     cfd_feed_actor_inbox: watch::Sender<Vec<Cfd>>,
     order_feed_sender: watch::Sender<Option<Order>>,
+    wallet_feed_sender: watch::Sender<WalletInfo>,
 ) -> (
     impl Future<Output = ()>,
     mpsc::UnboundedSender<maker_cfd_actor::Command>,
 )
 where
+    B: Blockchain,
     D: BatchDatabase,
```
thomaseizinger: Needs a rebase on top of @klochowicz work, otherwise master will fail once you merge this.

brb activating bors ...

daemon/src/wallet.rs
```diff
@@ -0,0 +1,27 @@
+use crate::model::WalletInfo;
+use anyhow::Result;
+use bdk::bitcoin::Amount;
+use bdk::blockchain::{Blockchain, NoopProgress};
+use bdk::database::BatchDatabase;
+use bdk::wallet::AddressIndex;
+use std::time::SystemTime;
+
+pub fn wallet_info<B, D>(wallet: &bdk::Wallet<B, D>) -> Result<WalletInfo>
```
thomaseizinger: Call this function `sync`, that describes better what actually happens in here. With the rebase on top of @klochowicz's work, perhaps make it a function on `Wallet` right away.

daemon/src/to_sse_event.rs
```diff
@@ -114,8 +114,26 @@ impl ToSseEvent for Option<model::cfd::Order> {
     }
 }
 
-impl ToSseEvent for Amount {
+#[derive(Debug, Clone, Serialize)]
+pub struct WalletInfo {
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    balance: Amount,
+    address: String,
+    unix_timestamp: u64,
```
thomaseizinger:
```suggestion
    last_updated_at: u64,
```

daemon/src/model.rs
```diff
@@ -83,3 +85,10 @@ impl Display for TakerId {
         self.0.fmt(f)
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct WalletInfo {
+    pub balance: Amount,
+    pub address: Address,
+    pub timestamp: SystemTime,
```
thomaseizinger:
```suggestion
    pub last_updated_at: SystemTime,
```

frontend/src/components/Wallet.tsx
```diff
@@ -0,0 +1,51 @@
...
+        walletInfo,
+    }: WalletProps,
+) {
+    const { hasCopied, onCopy } = useClipboard(walletInfo ? walletInfo.address : "");
+
+    let balance = <Skeleton height="20px" />;
+    let addressWidget = <Skeleton height="20px" />;
+    let timestamp = <Skeleton height="20px" />;
+
+    console.log(walletInfo);
```
thomaseizinger: Remove this?

frontend/src/components/Wallet.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+export default function Wallet(
+    {
+        walletInfo,
+    }: WalletProps,
+) {
+    const { hasCopied, onCopy } = useClipboard(walletInfo ? walletInfo.address : "");
+
+    let balance = <Skeleton height="20px" />;
```
thomaseizinger, line 17: I'd prefer conditional definitions over mutable variables:

```suggestion
    const balance = walletInfo ? <Text>{walletInfo.balance} BTC</Text> : <Skeleton height="20px" />;
```

daemon/util/testnet_seeds/maker_seed
```diff
@@ -0,0 +1 @@
+%�_�U���]�v@cz���S�RJ�����o��o{���T�&h���Ƭu_�֬�h*(��BBm^�[6�Nm�I�+댵[cm���G�j�uC���>�x�A��L�Ccx��ZE��X#�z��}��ڀ�lA*�}"ڬ�I�p�`��lr=B�ٳ��T7_)D�p�,k�T�R�r��T��A�$����Ψ���^���VSA�+ѓ���^F�\=��U^H{�K�/ ,��è(��E�#8KO��ɨ��׹
```
thomaseizinger, line 1: At the moment, we can only specify data-directory so in order to use these, we need to point the daemons to this via `--data-dir=./util/testnet_seeds`. That will also dump the database into here.

You didn't extend the `.gitignore` which means the DB  will always show up as part of the diff yeah?
Can we just define two data-dirs and only allow the seeds to be committed to Git? The naming of the dirs currently suggests that it only stores the seeds which is not true.

thomaseizinger: Also, can we extend the README to document how to start the daemons to use these seeds?

thomaseizinger (dismissed):
Couple of more comments. Needs a rebase to avoid conflicts in master.

daemon/src/wallet.rs
da-kami: removed

daemon/util/testnet_seeds/maker_seed
da-kami: I like the idea, but I did **not** see those keys as ready to use in this dir for now. I expected everyone to copy them where needed. They are not supposed to change. I don't see a reason for optimising this at the moment. If you feel the need for this please create a follow up PR.

frontend/src/components/Wallet.tsx
da-kami: In this case I disagree. I think the code reads way nicer the way it is. You see that all the different widgets are initialized as `Skeleton` and only if `walletInfo` is present they change. Additionally less code duplication this way, because otherwise I have to check `walletInfo` 3 times...

PR #89, opened 2021-09-20
Secure the maker's API routes with basic authentication

### LINKED ISSUE #76: Guard the maker's interface and API with authentication
We will initially host the broker with a publicly accessible IP. We need to guard the API and frontend with some authentication to make sure no one can muck around with our instance.

thomaseizinger:
Resolves #76.

We choose basic authentication because it is simple.
Basic auth allows us to not code any particular login API and instead just trigger the browser's built-in authentication dialog via the `WWW-Authenticate` header of the 401 Unauthorized response. It looks like this:

image

The password is derived from the maker's seed.
The username is always `maker`.

The implementation has one ceveat: Basic authentication is vulnerable to MITM attacks unless some form of transport encryption is in place. I suggest all securing the maker's API additionally with TLS. However, I think we can easily ship this as a feature on top!

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -108,9 +108,6 @@ pub async fn post_sell_order(
 #[rocket::get("/alive")]
 pub fn get_health_check() {}
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-struct RetrieveCurrentOrder;
-
 #[derive(RustEmbed)]
 #[folder = "../frontend/dist/maker"]
 struct Asset;
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -1,10 +1,11 @@
+use crate::auth::Authenticated;
 use crate::maker_cfd_actor;
 use crate::model::cfd::{Cfd, Order, Origin};
 use crate::model::{Usd, WalletInfo};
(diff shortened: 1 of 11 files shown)


daemon/src/routes_maker.rs
```diff
@@ -103,3 +124,71 @@ pub async fn post_sell_order(
...
+            .manage(state2)
+            .manage(state3)
+            .manage(state4)
+            .manage(Password::from(*b"Now I'm feelin' so fly like a G6"))
+    }
+
+    /// Creates an "Authorization" header that matches the password above,
+    /// in particular it has been created through:
+    /// ```
+    /// base64(maker:hex("Now I'm feelin' so fly like a G6"))
```
bonomat, line 228: 😅

bonomat (approved):
Would it not make sense to also have authentication for the taker?

daemon/src/routes_maker.rs
thomaseizinger: I was listening to music and needed some random 32 bytes ...

thomaseizinger:
> Would it not make sense to also have authentication for the taker? 

We can trivially reuse this for the taker as well. I thought the taker is most likely run locally and on-demand whereas the maker is probably on a server somewhere due to the need for an open port!

da-kami (approved):

klochowicz (approved):
nice one!

PR #91, opened 2021-09-20
Remove duplicated refund timelock block calculation

da-kami:
I moved this code into the `Cfd` in #85 but unfortunately die not pull that through completely, hence this cleanup :)


thomaseizinger (approved):
Thanks!

PR #93, opened 2021-09-20
Successful `ContractSetup` phase

### LINKED ISSUE #84: Let both parties try to publish the transaction and deal with "already present" errors on both sides.
See discussion for details:
https://github.com/comit-network/hermes/discussions/63

Should be ready to go, but in order to be properly tested in click through it depends on:
https://github.com/comit-network/hermes/issues/62
https://github.com/comit-network/hermes/issues/83

da-kami:
Both parties have the finalized CFD and are in `PendingOpen` ready to publish on chain.

Next I will work on #84 to let both parties try to publish the transaction.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -497,7 +497,7 @@ mod tests {
                 transition_timestamp: SystemTime::now(),
             },
         };
-        insert_new_cfd_state_by_order_id(cfd.order.id, cfd.state, &mut conn)
+        insert_new_cfd_state_by_order_id(cfd.order.id, cfd.state.clone(), &mut conn)
             .await
             .unwrap();
 
--- a/daemon/src/maker_cfd_actor.rs
+++ b/daemon/src/maker_cfd_actor.rs
@@ -1,5 +1,8 @@
-use crate::db::{insert_cfd, insert_order, load_all_cfds, load_cfd_by_order_id, load_order_by_id};
-use crate::model::cfd::{Cfd, CfdState, CfdStateCommon, FinalizedCfd, Order, OrderId};
+use crate::db::{
+    insert_cfd, insert_new_cfd_state_by_order_id, insert_order, load_all_cfds,
+    load_cfd_by_order_id, load_order_by_id,
+};
+use crate::model::cfd::{Cfd, CfdState, CfdStateCommon, Dlc, Order, OrderId};
 use crate::model::{TakerId, Usd, WalletInfo};
 use crate::wallet::Wallet;
 use crate::wire::SetupMsg;
@@ -27,7 +30,10 @@ pub enum Command {
         id: TakerId,
     },
     IncProtocolMsg(SetupMsg),
-    CfdSetupCompleted(FinalizedCfd),
+    CfdSetupCompleted {
+        order_id: OrderId,
+        dlc: Dlc,
+    },
 }
 
 pub fn new(
@@ -43,7 +49,11 @@ pub fn new(
     mpsc::UnboundedSender<maker_cfd_actor::Command>,
 ) {
     let (sender, mut receiver) = mpsc::unbounded_channel();
+
+    // TODO: Move the contract setup into a dedicated actor and send messages to that actor that
+    // manages the state instead of this ugly buffer
     let mut current_contract_setup = None;
+    let mut contract_setup_message_buffer = vec![];
 
     let mut current_order_id = None;
 
@@ -157,6 +167,8 @@ pub fn new(
                             .unwrap();
                     }
                     maker_cfd_actor::Command::StartContractSetup { taker_id, order_id } => {
+                        println!("CONTRACT SETUP");
+
                         // Kick-off the CFD protocol
                         let (sk, pk) = crate::keypair::new(&mut rand::thread_rng());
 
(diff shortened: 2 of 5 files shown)


daemon/src/maker_cfd_actor.rs
```diff
@@ -27,7 +30,10 @@ pub enum Command {
         id: TakerId,
     },
     IncProtocolMsg(SetupMsg),
-    CfdSetupCompleted(FinalizedCfd),
+    CfdSetupCompleted {
+        order_id: OrderId,
+        dlc: FinalizedCfd,
```
thomaseizinger: Why `dlc`?

daemon/src/maker_cfd_actor.rs
```diff
@@ -185,27 +197,77 @@ pub fn new(
                             cfd,
                         );
 
+                        current_contract_setup = Some(inbox.clone());
+                        for msg in contract_setup_message_buffer.clone() {
```
thomaseizinger: You could drain the buffer here, then you don't need to clone and reset the buffer.

daemon/src/maker_cfd_actor.rs
```diff
@@ -185,27 +197,77 @@ pub fn new(
...
+                        .await
+                        .unwrap();
+
+                        cfd_feed_actor_inbox
+                            .send(load_all_cfds(&mut conn).await.unwrap())
+                            .unwrap();
+
+                        // TODO: Publish on chain and only then transition to open - this might
+                        // require saving some internal state to make sure we are able to monitor
+                        // the publication after a restart
```
thomaseizinger, line 271: Just try to publish on start up and handle error response accordingly would be my suggestion.

daemon/src/setup_contract_actor.rs
```diff
@@ -12,20 +12,6 @@ use futures::Future;
...
-/// Ideally, both users collaboratively settle in the refund scenario. This factor is important if
-/// the users do not settle collaboratively.
-/// `1.5` times the term as defined in CFD order should be safe in the extreme case where a user
-/// publishes the commit transaction right after the contract was initialized. In this case, the
-/// oracle still has `1.0 * cfdorder.term` time to attest and no one can publish the refund
-/// transaction.
-/// The downside is that if the oracle disappears: the users would only notice at the end
-/// of the cfd term. In this case the users has to wait for another `1.5` times of the
-/// term to get his funds back.
-pub const REFUND_THRESHOLD: f32 = 1.5;
```
thomaseizinger: Isn't deleting this already part of a different PR?

daemon/src/taker_cfd_actor.rs
```diff
@@ -17,11 +17,17 @@ use tokio::sync::{mpsc, watch};
 #[allow(clippy::large_enum_variant)]
 pub enum Command {
     SyncWallet,
-    TakeOrder { order_id: OrderId, quantity: Usd },
+    TakeOrder {
+        order_id: OrderId,
+        quantity: Usd,
+    },
```
thomaseizinger: Reformatting? 🤔

daemon/src/wire.rs
```diff
@@ -29,7 +29,7 @@ pub enum MakerToTaker {
     Protocol(SetupMsg),
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
thomaseizinger: Should not be necessary.

daemon/src/maker_cfd_actor.rs
```diff
@@ -157,6 +167,8 @@ pub fn new(
                             .unwrap();
                     }
                     maker_cfd_actor::Command::StartContractSetup { taker_id, order_id } => {
+                        println!("CONTRACT SETUP");
```
thomaseizinger, line 170: We need a logger.

daemon/src/maker_cfd_actor.rs
da-kami: Because it is the parameters that define the DLC. I think that is what it is. I can also name it `finalized_cfd`, but I found that name a bit arbitrary.

daemon/src/maker_cfd_actor.rs
da-kami: Yeah, handling the state upon startup should be trivial. If we are in `PendingOpen we just use the same function to try to publish as here. Because we have to handle the case where it is already on chain anyway.

daemon/src/maker_cfd_actor.rs
da-kami: #94 (I converted the ToDo that was still there to an issue)

daemon/src/taker_cfd_actor.rs
da-kami: 😅 how? I don't understand how this happened and I could not find a `reformat` option on dprint.

daemon/src/maker_cfd_actor.rs
thomaseizinger: Yeah that name is totally up for change haha. It was just the first thing that come to my mind when we prototyped the protocol integration :)

daemon/src/taker_cfd_actor.rs
thomaseizinger: Yeah I meant whether or not this is really just reformatting and why it is happening :D

daemon/src/maker_cfd_actor.rs
thomaseizinger: Naming it `Dlc` sounds good to me!

daemon/src/taker_cfd_actor.rs
```diff
@@ -142,10 +155,34 @@ pub fn new(
 
                         inbox.send(msg).unwrap();
                     }
-                    Command::CfdSetupCompleted(_finalized_cfd) => {
-                        todo!("but what?")
+                    Command::CfdSetupCompleted { order_id, dlc } => {
+                        dbg!("Setup complete, publishing on chain now...");
```
thomaseizinger: Nit: With `dbg` this will essentially print _twice_. String literals in `dbg` are basically always pointless.

daemon/src/db.rs
```diff
@@ -497,7 +497,7 @@ mod tests {
                 transition_timestamp: SystemTime::now(),
             },
         };
-        insert_new_cfd_state_by_order_id(cfd.order.id, cfd.state, &mut conn)
+        insert_new_cfd_state_by_order_id(cfd.order.id, cfd.state.clone(), &mut conn)
```
thomaseizinger, line 500: This should be squashed into some other commit, whichever one caused it to fail. You can find out with `git rebase master -x 'cargo test --workspace'`

daemon/src/maker_cfd_actor.rs
```diff
@@ -198,7 +198,8 @@ pub fn new(
                         );
 
                         current_contract_setup = Some(inbox.clone());
-                        for msg in contract_setup_message_buffer.clone() {
+
+                        for msg in contract_setup_message_buffer.drain(..) {
```
thomaseizinger, line 202: `git absorb` would have likely picked the right commit here to squash this into :)

da-kami: 😬

PR #95, opened 2021-09-21
Embed maker/taker frontend into binary

### LINKED ISSUE #64: Embed frontend in binaries
Automate embedding of frontend artifacts in the respective binaries and serve them on the same port as the HTTP API.

Outcome should be a self-contained binary that serves its own frontend once started.

bonomat:
Note: we need to check-in `frontend/dist` directories as otherwise the build will fail.
If you want to embed the files into the binaries you will need to build them first (`yarn run build`).

resolves #64

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -1,9 +1,9 @@
 use crate::maker_cfd_actor;
 use crate::model::cfd::{Cfd, Order, Origin};
 use crate::model::{Usd, WalletInfo};
+use crate::routes::EmbeddedFileExt;
 use crate::to_sse_event::ToSseEvent;
 use anyhow::Result;
-
 use rocket::http::{ContentType, Status};
 use rocket::response::stream::EventStream;
 use rocket::response::{status, Responder};
@@ -11,7 +11,7 @@ use rocket::serde::json::Json;
 use rocket::State;
 use rust_embed::RustEmbed;
 use serde::Deserialize;
-use std::ffi::OsStr;
+use std::borrow::Cow;
 use std::path::PathBuf;
 use tokio::select;
 use tokio::sync::{mpsc, watch};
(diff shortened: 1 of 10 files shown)


daemon/src/routes_maker.rs
```diff
@@ -4,11 +4,15 @@ use crate::model::{Usd, WalletInfo};
 use crate::to_sse_event::ToSseEvent;
 use anyhow::Result;
```
thomaseizinger: Can we remove this newline? rustfmt doesn't format import blocks across empty lines.

daemon/src/routes_maker.rs
```diff
@@ -106,3 +110,32 @@ pub fn get_health_check() {}
...
+                .extension()
+                .and_then(OsStr::to_str)
+                .ok_or_else(|| Status::new(400))?;
+            let content_type = ContentType::from_extension(ext).ok_or_else(|| Status::new(400))?;
+            Ok((content_type, d.data))
+        },
+    )
+}
+
+#[rocket::get("/<_paths..>")]
```
thomaseizinger: Can we add `format = "text/html"` here?

That would make it more obvious that we only want to respond with `index.html` for all requests that ask for a HTML page.

I think  that should be possible: https://docs.rs/rocket/0.5.0-rc.1/rocket/attr.get.html#grammar

daemon/src/routes_maker.rs
```diff
@@ -106,3 +110,32 @@ pub fn get_health_check() {}
...
+pub fn index<'r>(_paths: PathBuf) -> impl Responder<'r, 'static> {
+    Asset::get("index.html").map_or_else(
+        || Err(Status::NotFound),
+        |d| Ok((ContentType::HTML, d.data)),
```
thomaseizinger: I'd find this easier to read (untested but I think should work):

```suggestion
    let asset = Asset::get("index.html").ok_or(Status::NotFound)?;
    
    Ok((ContentType::HTML, asset.data))
```

Or if you wanted to use one of rocket's built-in responders:

```rust
Ok(Html(asset.data))
```

daemon/src/routes_maker.rs
```diff
@@ -106,3 +110,32 @@ pub fn get_health_check() {}
...
+            let ext = file
+                .as_path()
+                .extension()
+                .and_then(OsStr::to_str)
+                .ok_or_else(|| Status::new(400))?;
+            let content_type = ContentType::from_extension(ext).ok_or_else(|| Status::new(400))?;
+            Ok((content_type, d.data))
+        },
+    )
+}
```
thomaseizinger: These are exactly the same for both right?

Can we have a shared implementation that is generic over the `RustEmbed` trait and hence allows to plugin in the respective `Asset` struct?

thomaseizinger (approved):
Thanks!

Couple of nits :)

daemon/src/routes_maker.rs
bonomat: I gave it a try in : [`28cc157` (#95)](https://github.com/comit-network/hermes/pull/95/commits/28cc157cee7da73cbe9a17ca2bbe59b876d019dc)

thomaseizinger (approved):
Nice, thank you for making it more DRY!

PR #98, opened 2021-09-21
Roll-out xtra framework usage in maker

klochowicz:
It's not rebased on master, I plan to do this after getting initial feedback (as I am likely to run into new merge conflicts anyway due to the size of this PR...)

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -1,5 +1,6 @@
 use crate::auth::MAKER_USERNAME;
-use crate::maker_cfd_actor::Command;
+use crate::maker_inc_connections_actor::in_taker_messages;
+use crate::model::TakerId;
 use crate::seed::Seed;
 use crate::wallet::Wallet;
 use anyhow::{Context, Result};
@@ -12,8 +13,10 @@ use rocket::fairing::AdHoc;
 use rocket_db_pools::Database;
 use std::path::PathBuf;
 use std::time::Duration;
-use tokio::sync::{mpsc, watch};
+use tokio::sync::watch;
 use tracing_subscriber::filter::LevelFilter;
+use xtra::prelude::*;
+use xtra::spawn::TokioGlobalSpawnExt;
 
 mod auth;
 mod db;
@@ -137,23 +140,56 @@ async fn main() -> Result<()> {
                     None => return Err(rocket),
                 };
 
-                let (connections_actor_inbox_sender, connections_actor_inbox_recv) =
-                    mpsc::unbounded_channel();
-
-                let (cfd_maker_actor, cfd_maker_actor_inbox) = maker_cfd_actor::new(
+                let cfd_maker_actor_inbox = maker_cfd_actor::MakerCfdActor::new(
                     db,
                     wallet,
                     schnorrsig::PublicKey::from_keypair(SECP256K1, &oracle),
-                    connections_actor_inbox_sender,
                     cfd_feed_sender,
                     order_feed_sender,
                     wallet_feed_sender,
-                );
-                let connections_actor = maker_inc_connections_actor::new(
-                    listener,
-                    cfd_maker_actor_inbox.clone(),
-                    connections_actor_inbox_recv,
-                );
+                )
+                .await
+                .create(None)
+                .spawn_global();
+
+                let maker_inc_connections_address =
+                    maker_inc_connections_actor::MakerIncConnectionsActor::new(
+                        cfd_maker_actor_inbox.clone(),
+                    )
+                    .create(None)
+                    .spawn_global();
+
+                tokio::spawn({
+                    let cfd_maker_actor_inbox = cfd_maker_actor_inbox.clone();
+                    let maker_inc_connections_address = maker_inc_connections_address.clone();
+                    async move {
+                        loop {
+                            if let Ok((socket, remote_addr)) = listener.accept().await {
+                                tracing::info!("Connected to {}", remote_addr);
+                                let taker_id = TakerId::default();
+                                let (read, write) = socket.into_split();
+
+                                let in_taker_actor = in_taker_messages(
+                                    read,
+                                    cfd_maker_actor_inbox.clone(),
+                                    taker_id,
+                                );
+                                let (out_msg_actor, out_msg_actor_inbox) =
+                                    send_wire_message_actor::new::<wire::MakerToTaker>(write);
+
+                                tokio::spawn(in_taker_actor);
+                                tokio::spawn(out_msg_actor);
+
+                                maker_inc_connections_address.do_send_async(
+                                    maker_inc_connections_actor::NewTakerOnline {
+                                        taker_id,
+                                        out_msg_actor_inbox,
+                                    },
+                                );
+                            };
+                        }
+                    }
+                });
 
                 // consecutive wallet syncs handled by task that triggers sync
                 let wallet_sync_interval = Duration::from_secs(10);
@@ -161,14 +197,21 @@ async fn main() -> Result<()> {
                     let cfd_actor_inbox = cfd_maker_actor_inbox.clone();
                     async move {
                         loop {
-                            cfd_actor_inbox.send(Command::SyncWallet).unwrap();
+                            cfd_actor_inbox
+                                .do_send_async(maker_cfd_actor::SyncWallet)
+                                .await
+                                .unwrap();
                             tokio::time::sleep(wallet_sync_interval).await;
                         }
                     }
                 });
 
-                tokio::spawn(cfd_maker_actor);
-                tokio::spawn(connections_actor);
+                cfd_maker_actor_inbox
+                    .do_send_async(maker_cfd_actor::Initialized(
+                        maker_inc_connections_address.clone(),
+                    ))
+                    .await
+                    .expect("not to fail after actors were initialized");
 
                 Ok(rocket.manage(cfd_maker_actor_inbox))
             },
--- a/daemon/src/maker_cfd_actor.rs
+++ b/daemon/src/maker_cfd_actor.rs
@@ -2,281 +2,412 @@ use crate::db::{
     insert_cfd, insert_new_cfd_state_by_order_id, insert_order, load_all_cfds,
     load_cfd_by_order_id, load_order_by_id,
 };
+use crate::maker_inc_connections_actor::{MakerIncConnectionsActor, TakerCommand};
 use crate::model::cfd::{Cfd, CfdState, CfdStateCommon, Dlc, Order, OrderId};
 use crate::model::{TakerId, Usd, WalletInfo};
 use crate::wallet::Wallet;
 use crate::wire::SetupMsg;
-use crate::{maker_cfd_actor, maker_inc_connections_actor, setup_contract_actor};
+use crate::{maker_inc_connections_actor, setup_contract_actor};
+use anyhow::{Context as AnyhowContext, Result};
+use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
-use futures::Future;
 use std::time::SystemTime;
 use tokio::sync::{mpsc, watch};
+use xtra::prelude::*;
 
-#[allow(clippy::large_enum_variant)]
-#[derive(Debug)]
-pub enum Command {
-    SyncWallet,
-    TakeOrder {
-        taker_id: TakerId,
-        order_id: OrderId,
-        quantity: Usd,
-    },
-    NewOrder(Order),
-    StartContractSetup {
-        taker_id: TakerId,
-        order_id: OrderId,
-    },
-    NewTakerOnline {
-        id: TakerId,
-    },
-    IncProtocolMsg(SetupMsg),
-    CfdSetupCompleted {
-        order_id: OrderId,
-        dlc: Dlc,
-    },
+pub struct Initialized(pub Address<MakerIncConnectionsActor>);
+
+impl Message for Initialized {
+    type Result = Result<()>;
 }
 
-pub fn new(
+pub struct TakeOrder {
+    pub taker_id: TakerId,
+    pub order_id: OrderId,
+    pub quantity: Usd,
+}
+
+impl Message for TakeOrder {
+    type Result = Result<()>;
+}
+
+pub struct NewOrder(pub Order);
+
+impl Message for NewOrder {
+    type Result = Result<()>;
+}
+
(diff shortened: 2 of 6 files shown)


klochowicz:
I've rebased on master & added the missing error handling wrappers.

daemon/src/maker_cfd_actor.rs
```diff
@@ -1,216 +1,344 @@
...
-    CfdSetupCompleted(FinalizedCfd),
-}
-
-pub fn new(
+use async_trait::async_trait;
+use xtra::prelude::*;
+
+pub struct Initialized(pub Address<MakerIncConnectionsActor>);
+impl Message for Initialized {
+    type Result = anyhow::Result<()>;
```
thomaseizinger: Would prefer to import `anyhow::Result`.

daemon/src/maker_cfd_actor.rs
```diff
@@ -1,216 +1,344 @@
...
+        // 2. Update actor state to current order
+        self.current_order_id.replace(order.id);
+
+        // 3. Notify UI via feed
+        self.order_feed_sender.send(Some(order.clone()))?;
+
+        // 4. Inform connected takers
+        self.takers
+            .as_ref()
+            .expect("actor to be initialized")
```
thomaseizinger: You can use `.context` to go from `Option` -> `anyhow::Result` which you can then `?`.

daemon/src/maker_inc_connections_actor.rs
```diff
@@ -1,95 +1,225 @@
...
+    type Result = anyhow::Result<()>;
+}
+
+pub struct OutProtocolMsg {
+    pub taker_id: TakerId,
+    pub msg: SetupMsg,
+}
+impl Message for OutProtocolMsg {
+    type Result = anyhow::Result<()>;
+}
```
thomaseizinger: I reckon these should be grouped into one enum for sending to a specific taker and one for sending to all. That is way too much boilerplate :D

daemon/src/maker_inc_connections_actor.rs
```diff
@@ -1,95 +1,225 @@
...
+                            .unwrap();
+
+                        write_connections
+                            .lock()
+                            .await
+                            .insert(taker_id, out_message_actor_inbox);
+                    };
                 }
-            }
+            });
```
thomaseizinger: We could do this outside of this actor and pass _this_ actor's `Address` into the loop, then we could have an incoming message for passing a new socket to it. That would allow you to avoid the `Arc<Mutex>` because inside `Handler` you have a mutable reference.

daemon/src/maker_cfd_actor.rs
```diff
@@ -1,216 +1,344 @@
...
+            Some(inbox) => inbox,
+        };
+        inbox.send(msg.0)?;
+        Ok(())
+    }
+
+    async fn handle_sync_wallet(&mut self) -> anyhow::Result<()> {
+        let wallet_info = self.wallet.sync().unwrap();
+        self.wallet_feed_sender.send(wallet_info).unwrap();
+        Ok(())
```
thomaseizinger: https://doc.rust-lang.org/unstable-book/language-features/try-blocks.html would be nice to have here but we can abuse `async` blocks:

```suggestion
		if let Err(e) = async {
	          let wallet_info = self.wallet.sync()?;
	          self.wallet_feed_sender.send(wallet_info)?;
	          
	          Ok(())
		}.await;
```

thomaseizinger: That should allow us to do stuff directly inside the handler functions without having to wrap them all!

thomaseizinger (commented):
Nice! Thanks for taking this on.

There will be quite a bit of merge conflicts. Can we trim this PR down to just a single actor and translate them one at a time? Otherwise you will never be done rebasing I am afraid.

klochowicz:
> Nice! Thanks for taking this on.
> 
> There will be quite a bit of merge conflicts. Can we trim this PR down to just a single actor and translate them one at a time? Otherwise you will never be done rebasing I am afraid.

challenge accepted 💪 however, in all seriousness, most merge conflicts are likely to happen in the maker cfd actor anyway, so I'm not sure if it would be any easier for me to split it (especially that at this stage splitting would not be trivial). 

so far I've solved the merge conflicts and rebased on master again :) if you don't mind, I'll try to keep it fresh and address the review comments promptly, aiming to merge PR this size. One good thing is that we can translate the taker in a separate PR.

thomaseizinger:
> > Nice! Thanks for taking this on.
> > There will be quite a bit of merge conflicts. Can we trim this PR down to just a single actor and translate them one at a time? Otherwise you will never be done rebasing I am afraid.
> 
> challenge accepted muscle however, in all seriousness, most merge conflicts are likely to happen in the maker cfd actor anyway, so I'm not sure if it would be any easier for me to split it (especially that at this stage splitting would not be trivial).
> 
> so far I've solved the merge conflicts and rebased on master again :) if you don't mind, I'll try to keep it fresh and address the review comments promptly, aiming to merge PR this size. One good thing is that we can translate the taker in a separate PR.

Okay, go for it then :)

daemon/src/maker_inc_connections_actor.rs
klochowicz: I ended up having 3 different messages: `BroadcastOrder` (sending to all), `TakerMessage` (sending to one taker, with enum inside), and `OutProtocolMsg`, as it's unrelated to both.

daemon/src/maker_inc_connections_actor.rs
klochowicz: not sure if I did it exactly the way you suggested, but it did remove the need for `Arc<Mutex>` :)

thanks for the suggestion, I wasn't happy with the previous solution!

daemon/src/maker_cfd_actor.rs
klochowicz: I tried using it, and it was complaining about the need of using fully qualified types.... if you don't mind, I'd be happy to add a small PR on top of these changes that cleans the handlers a bit - now I'm trying to merge before new rebase conflicts :)

daemon/src/maker.rs
```diff
@@ -132,31 +134,67 @@ async fn main() -> Result<()> {
...
-                );
-                let connections_actor = maker_inc_connections_actor::new(
-                    listener,
-                    cfd_maker_actor_inbox.clone(),
-                    connections_actor_inbox_recv,
-                );
+                )
+                .await
+                .create(None)
+                .run();
```
thomaseizinger: Use [`spawn_global`](https://docs.rs/xtra/0.5.1/xtra/spawn/trait.TokioGlobalSpawnExt.html#tymethod.spawn_global) instead :)

daemon/src/maker_inc_connections_actor.rs
```diff
@@ -1,95 +1,164 @@
...
+    SendOrder { order: Option<Order> },
+    NotifyInvalidOrderId { id: OrderId },
+    NotifyOrderAccepted { id: OrderId },
+}
+
+pub struct TakerMessage {
+    pub taker_id: TakerId,
+    pub command: TakerCommand,
+}
+impl Message for TakerMessage {
```
thomaseizinger: Would prefer to have newlines here!

daemon/src/maker_inc_connections_actor.rs
```diff
@@ -1,95 +1,164 @@
...
+
+        self.write_connections
+            .insert(msg.taker_id, msg.out_msg_actor_inbox);
+        Ok(())
+    }
+
+    async fn handle_out_protocol_msg(&mut self, msg: OutProtocolMsg) -> Result<()> {
+        self.send_to_taker(msg.taker_id, wire::MakerToTaker::Protocol(msg.msg))?;
+        Ok(())
+    }
```
thomaseizinger: Any particular reason this is not part of `TakerCommand`?

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
 use crate::model::{TakerId, Usd, WalletInfo};
 use crate::wallet::Wallet;
 use crate::wire::SetupMsg;
-use crate::{maker_cfd_actor, maker_inc_connections_actor, setup_contract_actor};
+use crate::{maker_inc_connections_actor, setup_contract_actor};
+use anyhow::{Context as AnyhowContext, Result};
 use bdk::bitcoin::secp256k1::schnorrsig;
-use futures::Future;
 use std::time::SystemTime;
 use tokio::sync::{mpsc, watch};
```
thomaseizinger: No newlines between import blocks please, otherwise the formatter doesn't pick them up!

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
-        order_id: OrderId,
-    },
-    NewTakerOnline {
-        id: TakerId,
-    },
-    IncProtocolMsg(SetupMsg),
-    CfdSetupCompleted {
-        order_id: OrderId,
-        dlc: Dlc,
-    },
```
thomaseizinger, line 36: Nit:

Something that might be worth trying: To keep the diff small, we could structure the new code to just list the messages as structs one after the other and have all the `impl Message for XYZ` at the very end of the file. They are just noise anyway because we don't actually use it.

That may reduce the diff because the fields and names of the variants didn't actually change.

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
+    ) -> Self {
+        let mut conn = db.acquire().await.unwrap();
+        // populate the CFD feed with existing CFDs
+        cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await.unwrap())
+            .unwrap();
+
+        // TODO: Move the contract setup into a dedicated actor and send messages to that actor that
+        // manages the state instead of this ugly buffer
+        let contract_setup_message_buffer: Vec<SetupMsg> = vec![];
```
thomaseizinger: Initialize the field further down and move the comment to the field of the struct.

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
+        wallet.sync().await.unwrap(); // XXX: Not sure if needed?
```
thomaseizinger: Not on startup no.

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
+            &mut conn,
+        )
+        .await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        ctx.notify(CfdSetupCompleted {
+            order_id,
+            dlc: actor.await,
+        });
```
thomaseizinger: This is blocking until the contract is successfully set up. I think we will run into a dead-lock here because the `actor` future cannot complete until the messages have been exchanged which only works if this actor can receive message for which it would have to acquire a mutable reference and it cannot do that because the current function is holding a mutable reference.

Instead of doing this, you should call `.address()` on the context and spawn this into the runtime again so it can complete in the background. At a later point, we should probably just combine this directly in the `contract_setup` actor.

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
+impl Handler<TakeOrder> for MakerCfdActor {
+    async fn handle(&mut self, msg: TakeOrder, _ctx: &mut Context<Self>) -> Result<()> {
+        tracing::debug!(%msg.taker_id, %msg.quantity, %msg.order_id,
+                        "Taker wants to take an order"
+        );
+
+        let takers = self
+            .takers
+            .as_ref()
+            .expect("Maker inc connections actor to be initialised");
```
thomaseizinger: This is duplicated across a couple of functions, maybe extract it?

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,402 @@ use crate::db::{
...
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        let txid = self
+            .wallet
+            .try_broadcast_transaction(msg.dlc.lock)
+            .await
+            .unwrap();
+
+        println!("Lock transaction published with txid {}", txid);
```
thomaseizinger: The tracing PR has been merged so these should all be using the `tracing` macros.

daemon/src/maker_inc_connections_actor.rs
klochowicz: oops, indeed, I overlooked that! fixed

daemon/src/maker_cfd_actor.rs
klochowicz: sure! extracted with`context()` call.

daemon/src/maker_cfd_actor.rs
klochowicz: I'll try to create a macro to remove this boilerplate altogether after this PR (or see whether `spaad` can be used in targeted way to remove it

daemon/src/maker_cfd_actor.rs
```diff
@@ -2,281 +2,410 @@ use crate::db::{
...
+
+        let address = ctx
+            .address()
+            .expect("actor to be able to give address to itself");
+        address
+            .do_send_async(CfdSetupCompleted {
+                order_id,
+                dlc: actor.await,
+            })
+            .await?;
```
thomaseizinger: The `actor.await` needs to be in a separate `tokio::spawn`, otherwise we have previously mentioned deadlock! :)

thomaseizinger (changes requested):
Just one minor changed required, otherwise good to be merged!

daemon/src/maker_cfd_actor.rs
klochowicz: like this ? 

```
        tokio::spawn(address.do_send_async(CfdSetupCompleted {
            order_id,
            dlc: actor.await,
        }));
```

thomaseizinger:
```rust
tokio::spawn(async move {
	address.do_send_async(CfdSetupCompleted {
       order_id,
       dlc: actor.await,
    })
});
```

thomaseizinger (approved):

PR #99, opened 2021-09-21
Add logging library and initialize by default

### LINKED ISSUE #94: Proper logging in the daemons
Configure `tracing` for both maker and taking so we have proper logs and not just prints.

bonomat:
resolves #94 

I used a similar approach as in xmr_btc.

### DIFF
--- a/daemon/README.md
+++ b/daemon/README.md
@@ -25,3 +25,16 @@ Currently the database for taker and maker is the same.
 The `taker` binary is used as an example to run the `prepare` command above, but it is irrelevant if you run it for taker or maker.
 The `tempdb` created can be deleted, it should not be checked into the repo.
 You can keep it around and just run the `prepare` statement multiple times when working on the database.
+
+## Logging
+
+`maker` and `taker` both make use of a log library which can be controlled using the environment variable `RUST_LOG`.
+By default, logging is set to `debug` for application specific logs only.
+If you want to see logs by third party libraries, you need to set `RUST_LOG=`.
+For example, to enable logs for rocket, use:
+
+```bash
+RUST_LOG=rocket=debug cargo run --bin taker
+```
+
+Additionally, logs can be formatted in json format by passing `--json` or `-j`.
--- a/daemon/src/logger.rs
+++ b/daemon/src/logger.rs
@@ -1,0 +1,33 @@
+use anyhow::Result;
+use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::fmt::time::ChronoLocal;
+use tracing_subscriber::{EnvFilter, FmtSubscriber};
+
+pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
+    if level == LevelFilter::OFF {
+        return Ok(());
+    }
+
+    let is_terminal = atty::is(atty::Stream::Stderr);
+
+    let filter = EnvFilter::from_default_env()
+        .add_directive(format!("taker={}", level).parse()?)
+        .add_directive(format!("maker={}", level).parse()?);
+
+    let builder = FmtSubscriber::builder()
+        .with_env_filter(filter)
+        .with_writer(std::io::stderr)
+        .with_ansi(is_terminal)
+        .with_timer(ChronoLocal::with_format("%F %T".to_owned()))
+        .with_target(false);
+
+    if json_format {
+        builder.json().init();
+    } else {
+        builder.init();
+    }
+
+    tracing::info!("Initialized logger");
+
+    Ok(())
+}
--- a/daemon/src/logger.rs
+++ b/daemon/src/logger.rs
@@ -12,7 +12,8 @@ pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
 
     let filter = EnvFilter::from_default_env()
         .add_directive(format!("taker={}", level).parse()?)
-        .add_directive(format!("maker={}", level).parse()?);
+        .add_directive(format!("maker={}", level).parse()?)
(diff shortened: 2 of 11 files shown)


daemon/src/logger.rs
```diff
@@ -0,0 +1,36 @@
...
+        .with_ansi(is_terminal)
+        .with_timer(ChronoLocal::with_format("%F %T".to_owned()))
+        .with_target(false);
+
+    if json_format {
+        builder.json().init();
+    } else if is_terminal {
+        builder.init();
+    } else {
+        builder.without_time().init();
```
thomaseizinger: This was a bad idea back then. We should expose this as a dedicated switch from the CLI (or not mess with the setting at all).

See https://github.com/comit-network/xmr-btc-swap/pull/765.

daemon/src/logger.rs
```diff
@@ -0,0 +1,36 @@
...
+use tracing_subscriber::{EnvFilter, FmtSubscriber};
+
+pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
+    if level == LevelFilter::OFF {
+        return Ok(());
+    }
+
+    let is_terminal = atty::is(atty::Stream::Stderr);
+
+    // match EnvFilter::try_from_default_env() {}
```
thomaseizinger: Stale comment?

daemon/src/maker_inc_connections_actor.rs
```diff
@@ -43,7 +43,7 @@ pub fn new(
         loop {
             tokio::select! {
                 Ok((socket, remote_addr)) = listener.accept() => {
-                    println!("Connected to {}", remote_addr);
+                    tracing::info!(%remote_addr, "Connected to");
```
thomaseizinger: Can we have complete log messages please.

The default formatter puts the fields next to the message so it more or less reads like a sentence. But not every formatter behaves this way. For example the JSON format puts these fields as separate fields in a JSON so the event reads something like:

```json
{
	"message": "Connected to"
}
```

Which is not a particularly nice log message when considered in isolation.

daemon/src/maker.rs
```diff
@@ -90,7 +99,7 @@ async fn main() -> Result<()> {
     let listener = tokio::net::TcpListener::bind(&format!("0.0.0.0:{}", opts.p2p_port)).await?;
     let local_addr = listener.local_addr().unwrap();
 
-    println!("Listening on {}", local_addr);
+    tracing::info!(%local_addr, "Listening on");
```
thomaseizinger: Same here.

daemon/src/maker.rs
```diff
@@ -51,12 +53,19 @@ struct Opts {
     /// Generate a seed file within the data directory.
     #[clap(long)]
     generate_seed: bool,
+
+    /// If enabled logs will be in json format
+    /// Generate a seed file within the data directory.
```
thomaseizinger: ?

daemon/src/maker.rs
```diff
@@ -51,12 +53,19 @@ struct Opts {
...
+    #[clap(short, long)]
+    json: bool,
 }
 
 #[rocket::main]
 async fn main() -> Result<()> {
     let opts = Opts::parse();
 
+    logger::init(LevelFilter::DEBUG, opts.json).expect("initialize logger");
```
thomaseizinger: We could `.context()?` here, no need to panic.

daemon/src/taker.rs
```diff
@@ -55,12 +57,19 @@ struct Opts {
     /// Generate a seed file within the data directory.
     #[clap(long)]
     generate_seed: bool,
+
+    /// If enabled logs will be in json format
+    /// Generate a seed file within the data directory.
```
thomaseizinger: Copy-pasta error.

daemon/README.md
```diff
@@ -25,3 +25,16 @@ Currently the database for taker and maker is the same.
 The `taker` binary is used as an example to run the `prepare` command above, but it is irrelevant if you run it for taker or maker.
 The `tempdb` created can be deleted, it should not be checked into the repo.
 You can keep it around and just run the `prepare` statement multiple times when working on the database.
+
+## Logging
+
+`maker` and `taker` both make use of a log library which can be controlled using the environment variable `RUST_LOG`.
+By default, logging is set to `debug` for application specific logs only.
+If you want to see logs by third party libraries, you need to set `RUST_LOG=`.
+For example, to enable logs for rocket, use:
```
thomaseizinger, line 34: I think we should retain rocket logs by default.

thomaseizinger (changes requested):
Thanks!

Couple of comments!

daemon/README.md
bonomat: Do you want that hardcoded `tracing.rs`?

daemon/src/logger.rs
bonomat: I opt for leaving the tiemstamp :)

daemon/src/maker_inc_connections_actor.rs
bonomat: I'm happy to change it back. 
For the other interested parties, this is how it would look like in json format: 

```json
{
  "timestamp": "....",
  "level": "INFO",
  "fields": {
    "message": "Listening on",
    "local_addr": "0.0.0.0:9999"
  }
}
```

daemon/README.md
thomaseizinger: I would have just included `rocket=` the same way we have `taker=` and `maker=`.

bonomat: done

thomaseizinger (approved):

Issue #71, opened 2021-09-16
Allow trader to propose collaborative settlement to the maker


thomaseizinger:
This should be the default user interaction.

thomaseizinger:
@klochowicz closing an issue will move it to "Done" but not the other way round :)

klochowicz:
> @klochowicz closing an issue will move it to "Done" but not the other way round :)

why not? AFAICT Jira works exactly like that! ;D

PR #100, opened 2021-09-21
Add close_transaction API to cfd_protocol lib

luckysori:
Work towards #71.

Notes:
- We use a flat fee like with all the other chain transactions. Both punish and close could take a `feerate` argument since they are both meant to be published soon after constructing them. We don't need to solve this now, but I remembered it.
- Both parties must know `lock_descriptor`, `lock_amount` and `lock_outpoint` in order to build the close transaction. Maybe it would be best to remember the descriptor and the txid and calculate the other two when necessary.

---

What are the next steps? Should I try to use this new API in the binaries?

### DIFF
--- a/cfd_protocol/src/lib.rs
+++ b/cfd_protocol/src/lib.rs
@@ -5,8 +5,9 @@ pub mod interval;
 
 pub use oracle::{attest, nonce};
 pub use protocol::{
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):
Cool, that will do for now! :)

Don't think we need to integrate it right away!

PR #109, opened 2021-09-22
Use a dedicated test route for testing maker's authentication

klochowicz:
Upcoming rollout of xtra will make instantiating an actor for test purposes much
more cumbersome. These tests don't need the actor, therefore a new route that
does not depend on instantiating one has been created.

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -86,6 +86,23 @@ pub async fn post_sell_order(
     Ok(status::Accepted(None))
 }
 
+/// Test route solely for the purposes of exercising authentication.
+/// It validates whether the posted order request was correct, but does not do
+/// anything else with it.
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):
I would have gone even more bare-bones and have the simplest GET endpoint that is authenticated.

I think you can even reuse the `index` endpoint for that so you don't even have to add a dedicated one (and trim down the test to not test any other endpoints).

klochowicz:
> I would have gone even more bare-bones and have the simplest GET endpoint that is authenticated.
> 
> I think you can even reuse the `index` endpoint for that so you don't even have to add a dedicated one (and trim down the test to not test any other endpoints).

thanks for the feedback! I'll look into this after the other PR lands, and might post a small follow-up cleanup.

PR #112, opened 2021-09-22
Add missing await for the message announcing NewTakerOnline

klochowicz:
Without the await, the message was never sent and the taker could not see new offers.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -180,12 +180,13 @@ async fn main() -> Result<()> {
                                 tokio::spawn(in_taker_actor);
                                 tokio::spawn(out_msg_actor);
 
-                                maker_inc_connections_address.do_send_async(
-                                    maker_inc_connections_actor::NewTakerOnline {
+                                maker_inc_connections_address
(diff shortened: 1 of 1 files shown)


daemon/src/maker.rs
```diff
@@ -180,12 +180,13 @@ async fn main() -> Result<()> {
...
-                                    maker_inc_connections_actor::NewTakerOnline {
+                                maker_inc_connections_address
+                                    .do_send_async(maker_inc_connections_actor::NewTakerOnline {
                                         taker_id,
                                         out_msg_actor_inbox,
-                                    },
-                                );
+                                    })
+                                    .await
+                                    .unwrap();
```
thomaseizinger, line 189: Why is this not flagged as a "future must be used" by clippy?

da-kami: nobody knows. we should probably report that. I think it should be reported...

da-kami (approved):

daemon/src/maker.rs
klochowicz: I'll report it tomorrow. it took us a while to find that problem ;/

PR #113, opened 2021-09-22
Add BitMex price feed

### LINKED ISSUE #96: Price feed for taker and maker
In order to be able to price the CFD as the maker we will need the current price feed. Given, that the maker might want to open counterpositions on Bitmex we should potentially use the Bitmex price feed.

This will enable: https://github.com/comit-network/hermes/issues/33

thomaseizinger:
Fixes #96.

### DIFF
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -1,0 +1,107 @@
+use crate::model::Usd;
+use anyhow::Result;
+use futures::{StreamExt, TryStreamExt};
+use rust_decimal::Decimal;
+use std::convert::TryFrom;
+use std::future::Future;
+use std::time::SystemTime;
+use tokio::sync::watch;
+use tokio_tungstenite::tungstenite;
+
+/// Connects to the BitMex price feed, returning the polling task and a watch channel that will
+/// always hold the last quote.
+pub async fn new() -> Result<(impl Future<Output = ()>, watch::Receiver<Quote>)> {
+    let (connection, _) = tokio_tungstenite::connect_async(
+        "wss://www.bitmex.com/realtime?subscribe=quoteBin1m:XBTUSD",
+    )
+    .await?;
+    let mut quotes = connection
+        .inspect_err(|e| tracing::warn!("Error on websocket stream: {}", e))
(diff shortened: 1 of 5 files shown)


da-kami:
Would you mind going one step further and wiring the price feed into the UIs as well? 
We might want to add a separate price feed for that.
Alternatively could also create a follow up ticket / PR:)

thomaseizinger:
> Would you mind going one step further and wiring the price feed into the UIs as well? 
> We might want to add a separate price feed for that.
> Alternatively could also create a follow up ticket / PR:)

Would prefer a follow-up as this integration is functional it itself!

thomaseizinger:
f.e. I think the integration will require to also let the CfdActor now about the price change so it might not just be an integration into the feed.

daemon/src/bitmex_price_feed.rs
```diff
@@ -0,0 +1,107 @@
...
+
+        let table_message = match serde_json::from_str::<wire::TableMessage>(&text_message) {
+            Ok(table_message) => table_message,
+            Err(_) => return Ok(None),
+        };
+
+        let [quote] = table_message.data;
+
+        Ok(Some(Self {
+            timestamp: SystemTime::now(),
```
bonomat, line 65: Why did you go for `SystemTime::now()` and not the `timestamp` from the quote?

thomaseizinger: Because parsing the timestamp from the quote was a pain using `serde` and the `time` library and I didn't want to add in more dependencies just for that. The timestamp here is only approximate anyway.

In the end, the oracle has authority over the timestamp of the signed price so I deemed it to not be important.

da-kami (approved):

daemon/src/bitmex_price_feed.rs
thomaseizinger: We could take it as `String` in the serde struct and then parse it here. Do you want me to do that?

bonomat: I think it's fine. The only reason why this could be relevant is if we want to check if the rate we received is current. We don't do any math with the rate (yet) so I guess that is not important? 🤷‍♂️

bonomat (approved):

PR #114, opened 2021-09-22
Remove `StartContractSetup` message

thomaseizinger:


klochowicz (approved):
thanks!

have you been able to fully complete the contract setup?

klochowicz (approved):
thanks, this makes things less confusing (at least we don't give hope to a casual reader that xtra would do anything with these Results - as it turned out, it just ignores them silently...

thomaseizinger:
> have you been able to fully complete the contract setup?

I haven't tried that but now that I am thinking about it, this change may also require to do some buffering on the taker side. Let me convert this to draft.

da-kami:
Note: You will want to rebase on master to get the `await` + `async` send fix in, otherwise the taker does not get an order from the maker initially and the maker's protocol messages won't go over the wire. 

Note: I just gave the code on master a try but ran into a panic on this unwrap (independent of the changes in this PR I think):

https://github.com/comit-network/hermes/blob/de66e85d730e018cc6abb692546b45e018f1f8c0/daemon/src/setup_contract_actor.rs#L50

Seems that's because the messages don't arrive in the right order 😅

```
thread 'rocket-worker-thread' panicked at 'called `Result::unwrap()` on an `Err` value: Msg1(Msg1 { commit: EcdsaAdaptorSignature(032ddbe8e3d1f61aa19e87c6c8fe912ea4b88443628b8fed7c5509bfef09370238030715448aa817e4b06f1207927170768d238d748a9b2611e3d8131f25291bc982c924d6d0eb9c1dd1e2d5397532ddb55507709ca74e4eeccd95cba1e191e5d4e840e9090cf6025a27a8c1167bb45772d28c02c5ef43093cc8f34eb795c7bd44d9095abb4f0ad0924dff3fa5578b088024030c511aa98489415a1dfe472cbc5be1), cets: [], refund: 3044022034bc74e8bef9f8ef916c79b38f03ed44e44a43fa041b000529ab8fef29dc68f50220413f8f78c23bd3285f56ff11a53f9fda259319e57024c1b1f76a595896557fd6 })', daemon/src/setup_contract_actor.rs:50:71
stack backtrace:
```

da-kami:
I tested (with adding the `await` in maker.rs) and the contract setup currently does not finish. Both parties get stuck in/after `ContractSetup` similar to the behavior on master. But I think that is expected behaviour because we use `send` here (which would have to be polled). I think we will have to change this to be async and await it. That's what @klochowicz and me found earlier.

https://github.com/comit-network/hermes/blob/9cb765cc0de47611cd537f38fb239518d6d07cd5/daemon/src/maker_cfd_actor.rs#L251-L266

PR #115, opened 2021-09-22
Send the protocol messages to the actor asynchronously

klochowicz:
`send()` does not actually send anything unless its polled; convert to
`do_send_async()` like elsewhere to send the message.

Thanks to this the contract setup advances a bit further (ie. the behaviour is restored, at least partially).

Now I am getting an error:

```
thread 'rocket-worker-thread' panicked at 'Sorry, have to panic 😬 - the amounts that the counterparty sent were wrong, expected 0.00006000 BTC actual 0.00060000 BTC', daemon/src/setup_contract_actor.rs:57:13
```

@da-kami  do you have any idea how this error could have originated?

### DIFF
--- a/daemon/src/maker_cfd_actor.rs
+++ b/daemon/src/maker_cfd_actor.rs
@@ -169,11 +169,19 @@ impl MakerCfdActor {
 
         let (actor, inbox) = setup_contract_actor::new(
             {
-                let inbox = self.takers()?.clone();
+                let takers = self.takers()?.clone();
                 move |msg| {
-                    inbox.send(maker_inc_connections_actor::TakerMessage {
-                        taker_id,
-                        command: TakerCommand::OutProtocolMsg { setup_msg: msg },
+                    tokio::spawn({
+                        let inbox = takers.clone();
+                        async move {
+                            inbox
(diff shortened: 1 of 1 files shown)


daemon/src/maker_cfd_actor.rs
```diff
@@ -169,11 +169,19 @@ impl MakerCfdActor {
...
+                        async move {
+                            inbox
+                                .do_send_async(maker_inc_connections_actor::TakerMessage {
+                                    taker_id,
+                                    command: TakerCommand::OutProtocolMsg { setup_msg: msg },
+                                })
+                                .await
+                                .unwrap();
+                        }
                     });
```
thomaseizinger, line 185: There is no need to spawn this into the runtime, sending the message doesn't take long so you can easily to this in the parent async context. Just append an `.await` to the `send` function.

thomaseizinger (changes requested):

thomaseizinger (approved):
All good!

daemon/src/maker_cfd_actor.rs
klochowicz: hm, this is a closure, so there's no parent async context unfortunately :(

PR #117, opened 2021-09-23
Cleanup MakerCfdActor

thomaseizinger:
- Have all handlers return `()`
- Move boilerplate to bottom of file
- Replace unwraps with `?`


klochowicz (approved):
Thanks!

PR #119, opened 2021-09-23
Add accept/reject feature

### LINKED ISSUE #87: Maker should either accept or reject a TakeRequest from the taker
Currently the maker auto-accepts it (a shortcut to allow testing other features)

bonomat:
Resolves #87

### DIFF
--- a/daemon/src/maker_cfd_actor.rs
+++ b/daemon/src/maker_cfd_actor.rs
@@ -23,6 +23,14 @@ pub struct TakeOrder {
     pub quantity: Usd,
 }
 
+pub struct AcceptOrder {
+    pub order_id: OrderId,
+}
+
+pub struct RejectOrder {
+    pub order_id: OrderId,
+}
+
 pub struct NewOrder(pub Order);
 
 pub struct NewTakerOnline {
@@ -178,7 +186,7 @@ impl MakerCfdActor {
         Ok(())
     }
 
-    async fn handle_take_order(&mut self, msg: TakeOrder, ctx: &mut Context<Self>) -> Result<()> {
+    async fn handle_take_order(&mut self, msg: TakeOrder) -> Result<()> {
         let TakeOrder {
             taker_id,
             order_id,
@@ -207,24 +215,18 @@ impl MakerCfdActor {
         };
 
         // 2. Insert CFD in DB
-        // TODO: Don't auto-accept, present to user in UI instead
         let cfd = Cfd::new(
             current_order.clone(),
-            quantity,
-            CfdState::Accepted {
+            msg.quantity,
+            CfdState::IncomingOrderRequest {
                 common: CfdStateCommon {
                     transition_timestamp: SystemTime::now(),
                 },
+                taker_id,
             },
         );
         insert_cfd(cfd, &mut conn).await?;
 
-        self.takers()?
-            .do_send_async(maker_inc_connections_actor::TakerMessage {
-                taker_id,
-                command: TakerCommand::NotifyOrderAccepted { id: order_id },
-            })
-            .await?;
         self.cfd_feed_actor_inbox
             .send(load_all_cfds(&mut conn).await?)?;
 
@@ -233,6 +235,60 @@ impl MakerCfdActor {
         self.takers()?
             .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
             .await?;
+        self.order_feed_sender.send(None)?;
+
+        Ok(())
+    }
+
+    async fn handle_accept_order(
+        &mut self,
+        msg: AcceptOrder,
+        ctx: &mut Context<Self>,
+    ) -> Result<()> {
+        tracing::debug!(%msg.order_id, "Maker accepts an order" );
+
+        let mut conn = self.db.acquire().await?;
+
+        // Validate if order is still valid
+        let cfd = load_cfd_by_order_id(msg.order_id, &mut conn).await?;
+
+        let taker_id = match cfd {
+            Cfd {
+                state: CfdState::IncomingOrderRequest { taker_id, .. },
+                ..
+            } => taker_id,
+            _ => {
+                anyhow::bail!("Order is in invalid state. Ignoring trying to accept it.")
+            }
+        };
+
+        // Update order in db
+        let order_id = cfd.order.id;
+        insert_new_cfd_state_by_order_id(
+            order_id,
+            CfdState::Accepted {
+                common: CfdStateCommon {
+                    transition_timestamp: SystemTime::now(),
+                },
+            },
+            &mut conn,
+        )
+        .await
+        .unwrap();
+
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::TakerMessage {
+                taker_id,
+                command: TakerCommand::NotifyOrderAccepted { id: msg.order_id },
+            })
+            .await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
+            .await?;
+        self.current_order_id = None;
         self.order_feed_sender.send(None)?;
 
         // 4. Start contract setup
@@ -300,7 +356,56 @@ impl MakerCfdActor {
 
         Ok(())
     }
+
+    async fn handle_reject_order(&mut self, msg: RejectOrder) -> Result<()> {
+        tracing::debug!(%msg.order_id, "Maker rejects an order" );
+
+        let mut conn = self.db.acquire().await?;
+        let cfd = load_cfd_by_order_id(msg.order_id, &mut conn).await?;
+
+        let taker_id = match cfd {
+            Cfd {
+                state: CfdState::IncomingOrderRequest { taker_id, .. },
+                ..
+            } => taker_id,
+            _ => {
+                anyhow::bail!("Order is in invalid state. Ignoring trying to accept it.")
+            }
+        };
+
+        // Update order in db
+        insert_new_cfd_state_by_order_id(
+            msg.order_id,
+            CfdState::Rejected {
+                common: CfdStateCommon {
+                    transition_timestamp: SystemTime::now(),
+                },
+            },
+            &mut conn,
+        )
+        .await
+        .unwrap();
+
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::TakerMessage {
+                taker_id,
+                command: TakerCommand::NotifyOrderRejected { id: msg.order_id },
+            })
+            .await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        // Remove order for all
+        self.current_order_id = None;
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
+            .await?;
+        self.order_feed_sender.send(None)?;
+
+        Ok(())
+    }
 }
+
 #[async_trait]
 impl Handler<Initialized> for MakerCfdActor {
     async fn handle(&mut self, msg: Initialized, _ctx: &mut Context<Self>) {
@@ -318,8 +423,22 @@ macro_rules! log_error {
 
 #[async_trait]
 impl Handler<TakeOrder> for MakerCfdActor {
-    async fn handle(&mut self, msg: TakeOrder, ctx: &mut Context<Self>) {
-        log_error!(self.handle_take_order(msg, ctx))
+    async fn handle(&mut self, msg: TakeOrder, _ctx: &mut Context<Self>) {
+        log_error!(self.handle_take_order(msg))
+    }
+}
+
+#[async_trait]
+impl Handler<AcceptOrder> for MakerCfdActor {
+    async fn handle(&mut self, msg: AcceptOrder, ctx: &mut Context<Self>) {
+        log_error!(self.handle_accept_order(msg, ctx))
+    }
+}
+
+#[async_trait]
+impl Handler<RejectOrder> for MakerCfdActor {
+    async fn handle(&mut self, msg: RejectOrder, _ctx: &mut Context<Self>) {
+        log_error!(self.handle_reject_order(msg))
     }
 }
 
@@ -383,6 +502,14 @@ impl Message for CfdSetupCompleted {
 }
 
 impl Message for SyncWallet {
+    type Result = ();
+}
+
+impl Message for AcceptOrder {
+    type Result = ();
+}
+
+impl Message for RejectOrder {
     type Result = ();
 }
 
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -137,14 +137,15 @@ pub enum CfdState {
     /// The taker has requested to take a CFD, but has not messaged the maker yet.
     ///
     /// This state only applies to the taker.
+    /// TODO: unused?
     TakeRequested {
         common: CfdStateCommon,
     },
-    /// The taker sent an open request to the maker to open the CFD but don't have a response yet.
+    /// The taker sent an order to the maker to open the CFD but doesn't have a response yet.
     ///
     /// This state applies to taker and maker.
     /// Initial state for the maker.
-    PendingTakeRequest {
+    PendingOrderRequest {
         common: CfdStateCommon,
     },
     /// The maker has accepted the CFD take request, but the contract is not set up on chain yet.
@@ -154,7 +155,7 @@ pub enum CfdState {
         common: CfdStateCommon,
     },
 
-    /// The maker rejected the CFD take request.
+    /// The maker rejected the CFD order.
     ///
     /// This state applies to taker and maker.
     Rejected {
@@ -213,7 +214,7 @@ impl CfdState {
     fn get_common(&self) -> CfdStateCommon {
         let common = match self {
             CfdState::TakeRequested { common } => common,
-            CfdState::PendingTakeRequest { common } => common,
+            CfdState::PendingOrderRequest { common } => common,
             CfdState::Accepted { common } => common,
             CfdState::Rejected { common } => common,
             CfdState::ContractSetup { common } => common,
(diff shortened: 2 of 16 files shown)


daemon/src/model/cfd.rs
```diff
@@ -237,10 +238,10 @@ impl Display for CfdState {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         match self {
             CfdState::TakeRequested { .. } => {
-                write!(f, "Take Requested")
+                write!(f, "Order created")
```
thomaseizinger: Why do we have a mismatch here between the enum name and what is printed?

daemon/src/taker_cfd_actor.rs
```diff
@@ -17,6 +17,7 @@ use tokio::sync::{mpsc, watch};
 #[allow(clippy::large_enum_variant)]
 pub enum Command {
     SyncWallet,
+    // TODO: should this be TakeOffer ?
```
thomaseizinger: I'd vote for either `TakeOffer` or `PlaceOrder`.

daemon/src/maker_cfd_actor.rs
```diff
@@ -307,14 +323,62 @@ impl Handler<TakeOrder> for MakerCfdActor {
...
+
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl Handler<AcceptOrder> for MakerCfdActor {
+    async fn handle(&mut self, msg: AcceptOrder, _ctx: &mut Context<Self>) -> Result<()> {
+        tracing::debug!(%msg.taker_id, %msg.order_id,
+                        "Taker accepts an order."
```
thomaseizinger: Is this a typo and it should say "Maker"?

daemon/src/model/cfd.rs
```diff
@@ -137,14 +137,15 @@ pub enum CfdState {
     /// The taker has requested to take a CFD, but has not messaged the maker yet.
     ///
     /// This state only applies to the taker.
+    /// TODO: unused?
```
bonomat: Removed in follow-up commit.

daemon/src/maker_cfd_actor.rs
```diff
@@ -307,14 +323,62 @@ impl Handler<TakeOrder> for MakerCfdActor {
...
+        );
+
+        let mut conn = self.db.acquire().await?;
+
+        // Validate if order is still valid
+        // TODO: shall we also reject if the order was already requested by someone else?
```
bonomat: Created a ticket here: https://github.com/comit-network/hermes/issues/120

daemon/src/wire.rs
```diff
@@ -25,8 +25,8 @@ pub enum TakerToMaker {
 #[allow(clippy::large_enum_variant)]
 pub enum MakerToTaker {
     CurrentOrder(Option<Order>),
-    // TODO: Needs RejectOrder as well
     ConfirmTakeOrder(OrderId), // TODO: Include payout curve in "accept" message from maker
+    RejectTakeOrder(OrderId),
```
thomaseizinger: Along with the other naming above, I think dropping the `Take` here would be better.

The taker is placing an order yes? Then saying `ConfirmOrder` and `RejectOrder` should be correct?

frontend/src/components/CfdTile.tsx
```diff
@@ -1,18 +1,63 @@
-import { Box, Button, SimpleGrid, Text, VStack } from "@chakra-ui/react";
+import { Box, Button, HStack, SimpleGrid, Text, VStack } from "@chakra-ui/react";
 import React from "react";
+import { useAsync } from "react-async";
 import { Cfd, unixTimestampToDate } from "./Types";
 
 interface CfdTileProps {
     index: number;
     cfd: Cfd;
+    onConfirm: ((args: any) => void) | undefined;
```
thomaseizinger: I think we should have two callbacks here, one for accept and one for reject.

daemon/src/routes_maker.rs
```diff
@@ -122,27 +122,45 @@ pub struct PromptAuthentication {
...
-//         .send(order)
-//         .await
-//         .map_err(|_| status::BadRequest(Some("internal server error".to_string())))?;
-
-//     Ok(status::Accepted(None))
-// }
+/// The maker POSTs this to accept an order
+#[derive(Debug, Clone, Deserialize)]
+pub struct AcceptOrRejectOrderRequest {
+    pub taker_id: TakerId,
```
thomaseizinger: Why do we need to send the `TakerId` along here? We should be able to know from the `OrderId`, where we need to send the information yes?

Having to compare that this `TakerId` is indeed the one that we have stored internally looks like an unnecessary error case.

daemon/src/model/cfd.rs
```diff
@@ -134,27 +134,29 @@ pub struct CfdStateCommon {
...
+    /// The maker received an order from the taker to open the CFD but doesn't have a response yet.
     ///
-    /// This state applies to taker and maker.
-    /// Initial state for the maker.
-    PendingTakeRequest {
+    /// This state applies to the maker only.
+    IncomingOrderRequest {
         common: CfdStateCommon,
+        taker_id: TakerId,
     },
```
thomaseizinger, line 150: This is where we might want to split the states into one per party. @da-kami what do you think?

daemon/src/maker_cfd_actor.rs
```diff
@@ -307,14 +323,62 @@ impl Handler<TakeOrder> for MakerCfdActor {
...
+                },
+            },
+            &mut conn,
+        )
+        .await
+        .unwrap();
+
         self.takers()?
             .do_send_async(maker_inc_connections_actor::TakerMessage {
                 taker_id: msg.taker_id,
```
thomaseizinger: Instead of `msg.taker_id`, we should be grabbing the `TakerId` from the state of the `Cfd`. This will also verify that the `Cfd` is in fact in the correct state.

daemon/src/maker_cfd_actor.rs
```diff
@@ -324,7 +388,49 @@ impl Handler<TakeOrder> for MakerCfdActor {
...
+                },
+            },
+            &mut conn,
+        )
+        .await
+        .unwrap();
+
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::TakerMessage {
+                taker_id: msg.taker_id,
```
thomaseizinger: Same here, grab the `TakerId` from the Cfd's state instead!

thomaseizinger (changes requested):
Thanks! Few ideas in comments inline.

The only really pressing things IMO are:

- Don't expose the `TakerId` in the accept/reject endpoints
- Have two callbacks in the UI component instead of a generic one

daemon/src/routes_maker.rs
bonomat: There is a potential racecondition here. I opted for keeping the taker id in this request so that we know which order we are accepting. 

I shortly summarized it in here: https://github.com/comit-network/hermes/issues/120

thomaseizinger: We can only have a race condition if we don't update the state properly. Actors process messages in sequence. If the state update within an incoming message like `TakeOrder` is sound, there can be no race conditions.

At the moment, we remove the currently active order as soon as it is taken. With your PR, we update the state to `IncomingOrderRequest` upon `TakeOrder`.

The only thing missing to prevent a race condition is to check that the order is not already in that state when another `TakeOrder` comes in.

daemon/src/taker_cfd_actor.rs
klochowicz: +1 for `PlaceOrder`

daemon/src/model/cfd.rs
```diff
@@ -137,14 +137,15 @@ pub enum CfdState {
     /// The taker has requested to take a CFD, but has not messaged the maker yet.
     ///
     /// This state only applies to the taker.
+    /// TODO: unused?
     TakeRequested {
         common: CfdStateCommon,
     },
-    /// The taker sent an open request to the maker to open the CFD but don't have a response yet.
+    /// The taker sent an order to the maker to open the CFD but doesn't have a response yet.
```
klochowicz: it seems that we'll need to consolidate naming throughout the daemon code. This meaning of `order` makes a lot of sense, but in many places we use "order" when it should be "offer" instead (ie. what maker is offering to the taker).

daemon/src/routes_maker.rs
bonomat: I'll remove it for now as it's not really consistent. The ticket remains though and we should takle it sooner or later.

bonomat:
Rebased and updated :)

daemon/src/model/cfd.rs
bonomat: Resolved in a later commit.

daemon/src/model/cfd.rs
bonomat: I agree, we should invest some efforts in unifying this.

daemon/src/maker_cfd_actor.rs
```diff
@@ -292,24 +301,136 @@ impl MakerCfdActor {
...
+
+        let mut conn = self.db.acquire().await?;
+
+        // Validate if order is still valid
+        let cfd = match self.current_order_id {
+            Some(current_order_id) if current_order_id == msg.order_id => {
+                load_cfd_by_order_id(msg.order_id, &mut conn).await?
+            }
+            _ => {
+                // TODO: Return an error here?
```
thomaseizinger: Our latest idea is that we have a dedicated feed where we can send error messages to the UI. We can add that later!

thomaseizinger: For now, we can just return an error here using `bail!` if you want to abort.

daemon/src/maker_cfd_actor.rs
```diff
@@ -416,4 +537,12 @@ impl Message for SyncWallet {
     type Result = ();
 }
 
+impl Message for AcceptOrder {
+    type Result = Result<()>;
```
thomaseizinger: Use `()` please and wrap all the functions in handlers that use the `log_error!` macro, otherwise the errors will be silently dropped.

daemon/src/routes_maker.rs
```diff
@@ -122,27 +122,42 @@ pub struct PromptAuthentication {
...
+pub struct AcceptOrRejectOrderRequest {
+    pub order_id: OrderId,
+}
+
+#[rocket::post("/order/accept", data = "<cfd_accept_order_request>")]
+pub async fn post_accept_order(
+    cfd_accept_order_request: Json<AcceptOrRejectOrderRequest>,
+    cfd_actor_address: &State<Address<MakerCfdActor>>,
+    _auth: Authenticated,
+) -> Result<status::Accepted<()>, status::BadRequest<String>> {
```
thomaseizinger: We never return an `Err` in here so this doesn't have to be a `Result`.

frontend/src/MakerClient.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+export async function postAcceptOrder(payload: AcceptOrderRequestPayload) {
+    let res = await fetch(
+        `/api/order/accept`,
+        { method: "POST", body: JSON.stringify(payload), credentials: "include" },
+    );
+
+    if (!res.status.toString().startsWith("2")) {
+        console.log("Status: " + res.status + ", " + res.statusText);
+        throw new Error("failed to publish new order");
```
thomaseizinger: This error text seems to be off.

thomaseizinger: Also, looking at the Rust code, this route will never return an error so I think we don't need to do any of this?

daemon/src/maker_cfd_actor.rs
```diff
@@ -292,24 +301,136 @@ impl MakerCfdActor {
...
                     transition_timestamp: SystemTime::now(),
                 },
+                taker_id: msg.taker_id,
             },
         );
         insert_cfd(cfd, &mut conn).await?;
 
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
```
thomaseizinger, line 232: I think we should set `current_order_id` to `None` here otherwise a 2nd taker can override someone else's order.

thomaseizinger (commented):
Thanks! Some more things to discuss.

frontend/src/MakerClient.tsx
bonomat: It can fail if you lose connection to your application.

thomaseizinger: You are checking the HTTP status code here. That implies that you have a connection. If you don't have a connection, the call to `fetch` would have already thrown an exception.

bonomat:
Squashed some commits and rebased

daemon/src/maker_cfd_actor.rs
```diff
@@ -207,33 +215,23 @@ impl MakerCfdActor {
...
 
         // 3. Remove current order
         self.current_order_id = None;
-        self.takers()?
-            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
-            .await?;
-        self.order_feed_sender.send(None)?;
 
         // 4. Start contract setup
         tracing::info!("Starting contract setup");
```
thomaseizinger: I think there might have been an error during the rebase. Contract setup shouldn't happen directly after the `TakeOrder` should it? I think we need to move this to `AcceptOrder`.

daemon/src/maker_cfd_actor.rs
```diff
@@ -300,7 +298,107 @@ impl MakerCfdActor {
...
+        let cfd = load_cfd_by_order_id(msg.order_id, &mut conn).await?;
+
+        let taker_id = match cfd {
+            Cfd {
+                state: CfdState::IncomingOrderRequest { taker_id, .. },
+                ..
+            } => taker_id,
+            _ => {
+                tracing::warn!("Order is in invalid state. Ignoring trying to accept it.");
+                anyhow::bail!("Order is in invalid state. Ignoring trying to accept it.")
```
thomaseizinger: Just the `bail!` should be sufficient because of the `log_error` macro. This way we are going to see the message twice!

daemon/src/maker_cfd_actor.rs
```diff
@@ -207,33 +215,23 @@ impl MakerCfdActor {
...
-            .await?;
         self.cfd_feed_actor_inbox
             .send(load_all_cfds(&mut conn).await?)?;
 
         // 3. Remove current order
         self.current_order_id = None;
-        self.takers()?
-            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
-            .await?;
-        self.order_feed_sender.send(None)?;
```
thomaseizinger: I think these have to stay as well, otherwise the UI and all other takers will be out of sync.

daemon/src/maker_cfd_actor.rs
```diff
@@ -300,7 +298,107 @@ impl MakerCfdActor {
...
+                command: TakerCommand::NotifyOrderAccepted { id: msg.order_id },
+            })
+            .await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
+            .await?;
+        self.order_feed_sender.send(None)?;
```
thomaseizinger: As per the other comment, this needs to happen together with setting `current_order_id` to `None`.

daemon/src/maker_cfd_actor.rs
```diff
@@ -300,7 +298,107 @@ impl MakerCfdActor {
...
+            .await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
+
+        // Remove current order
+        self.current_order_id = None;
+        self.takers()?
+            .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
+            .await?;
+        self.order_feed_sender.send(None)?;
```
thomaseizinger: We could just put it back instead? Otherwise the maker has to create a new one every time they reject an order?

thomaseizinger (changes requested):
Sorry for yet another round of requesting changes but I think a bug creeped in during the rebase.

daemon/src/maker_cfd_actor.rs
bonomat: Good catch.. I guess I should stop for today.

daemon/src/maker_cfd_actor.rs
bonomat: This does not seem to work because we save the `CFD` in the state `rejected` also for the maker.

thomaseizinger: Ah missed that! I guess in some way it is "cleaner" that although it might be a bit annoying.

daemon/src/maker_cfd_actor.rs
```diff
@@ -207,32 +215,80 @@ impl MakerCfdActor {
...
 
-        // 3. Remove current order
-        self.current_order_id = None;
         self.takers()?
             .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
             .await?;
+        self.current_order_id = None;
         self.order_feed_sender.send(None)?;
 
         // 4. Start contract setup
```
thomaseizinger, line 294: We are loading things again from the DB down here which would not be necessary but it also not harmful so we can ignore it for now!

daemon/src/maker_cfd_actor.rs
```diff
@@ -207,32 +215,80 @@ impl MakerCfdActor {
...
         self.cfd_feed_actor_inbox
             .send(load_all_cfds(&mut conn).await?)?;
 
-        // 3. Remove current order
-        self.current_order_id = None;
         self.takers()?
             .do_send_async(maker_inc_connections_actor::BroadcastOrder(None))
             .await?;
+        self.current_order_id = None;
         self.order_feed_sender.send(None)?;
```
thomaseizinger, line 292: Doesn't hurt to do this but I think it may never actually do anything because we set all this to `None` upon `TakeOrder`.

thomaseizinger (approved):
Thanks, looks good to me!

thomaseizinger:
I will merge this so you don't have to rebase again :)

bonomat:
> I will merge this so you don't have to rebase again :)

thank you 🙏 :)

PR #123, opened 2021-09-23
Simplify UI

thomaseizinger:
UI now looks like this when it doesn't have a lot of space:

image

But also looks decent when it does have a lot of space:

Screenshot from 2021-09-23 13-43-20


thomaseizinger:
Replaced by https://github.com/comit-network/hermes/pull/125.

PR #125, opened 2021-09-23
Simplify UI

thomaseizinger:
UI now looks like this when it doesn't have a lot of space:

image

But also looks decent when it does have a lot of space:

Screenshot from 2021-09-23 13-43-20

### DIFF
--- a/frontend/src/Maker.tsx
+++ b/frontend/src/Maker.tsx
@@ -1,25 +1,11 @@
-import {
-    Box,
-    Button,
-    Center,
-    Divider,
-    Flex,
-    HStack,
-    SimpleGrid,
-    StackDivider,
-    Text,
-    useToast,
-    VStack,
-} from "@chakra-ui/react";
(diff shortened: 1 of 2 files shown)


klochowicz:
I like getting rid of the 3 tabs - they weren't used at all! 👍 

Is there a reason why the tiles don't look like tiles anymore - perhaps you could return to the 2 columns in the grid approach? They seem way too big now.

thomaseizinger:
> Is there a reason why the tiles don't look like tiles anymore - perhaps you could return to the 2 columns in the grid approach? They seem way too big now.

Scaling things horizontally is difficult in my experience if you want things to be somewhat responsive.
Also I find lists easier to understand because my eyes don't have to go left and right :)

da-kami (approved):
Good enough as foundation for now.
I expect the UI to change more radical once we get user feedback.

thomaseizinger:
> Good enough as foundation for now.
> I expect the UI to change more radical once we get user feedback.

Agreed which is why I wanted to make it simpler.

PR #128, opened 2021-09-23
Misc improvements

thomaseizinger:
Some minor improvements that fell out of trying to extract a dedicated actor for handling the contract setup.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -12,7 +12,6 @@ use model::WalletInfo;
 use rocket::fairing::AdHoc;
 use rocket_db_pools::Database;
 use std::path::PathBuf;
-use std::time::Duration;
 use tokio::sync::watch;
 use tracing_subscriber::filter::LevelFilter;
 use xtra::prelude::*;
@@ -34,6 +33,7 @@ mod send_wire_message_actor;
 mod setup_contract_actor;
 mod to_sse_event;
 mod wallet;
+mod wallet_sync;
 mod wire;
 
 #[derive(Database)]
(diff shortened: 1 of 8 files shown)


daemon/src/maker.rs
```diff
@@ -157,23 +157,28 @@ async fn main() -> Result<()> {
                     None => return Err(rocket),
                 };
 
+                let (maker_inc_connections_address, maker_inc_connections_context) =
+                    xtra::Context::new(None);
```
klochowicz, line 161: cool!

daemon/src/wallet_sync.rs
```diff
@@ -0,0 +1,24 @@
+use crate::wallet::Wallet;
```
klochowicz, line 1: nit: rename file to `wallet_sync_actor.rs`?

klochowicz (approved):
Thanks for the cleanup! I'll post a small one doing similar cleanup for the `TakerCfdActor` in a sec 😃

luckysori (approved):
LGTM

daemon/src/wallet_sync.rs
da-kami: I don't think this is an actor

da-kami (approved):
It starts up and I see balances in the UI, merge it!

daemon/src/wallet_sync.rs
klochowicz: fair enough, but it's a long-running task, and it is a bit misleading when it just starts with `new()` - maybe it could start with `spawn()` or something like that? 

or can we reason about "actors" without inboxes?

thomaseizinger: I think that is the main thing right there, if something doesn't have an inbox, it is not an actor :)

PR #130, opened 2021-09-23
2 minor things as follow up from #119

bonomat:


daemon/src/maker_cfd_actor.rs
```diff
@@ -285,13 +285,12 @@ impl MakerCfdActor {
...
             .send(load_all_cfds(&mut conn).await?)?;
 
-        // 4. Start contract setup
+        // Start contract setup
         tracing::info!("Starting contract setup");
 
         // Kick-off the CFD protocol
         let (sk, pk) = crate::keypair::new(&mut rand::thread_rng());
 
-        let cfd = load_cfd_by_order_id(order_id, &mut conn).await?;
```
thomaseizinger: So actually, we should do `cfd.state = CfdState::Accepted` here.

bonomat: I guess that would be cleaner...
 
It's not crucial though because the state is not used and a few lines below we update the db with the new state `CfdState::ContractSetup`. 

Shall I add it nevertheless?

thomaseizinger: Not necessary then :)

thomaseizinger (approved):

PR #139, opened 2021-09-24
Put sources under src/

thomaseizinger:
The only thing that has to stay in the root is the `index.html` file,
otherwise the dev server doesn't pick it up.
Also make sure our frontend tests actually run and pass in CI.


klochowicz (approved):
Thanks!

PR #142, opened 2021-09-24
Simplify `EmbeddedFileExt`

thomaseizinger:
- Put sources under src/
- Refactor `EmbeddedFileExt` to use combinators and ?


klochowicz (approved):
it indeed reads easier!

PR #144, opened 2021-09-24
Add standard release workflows

### LINKED ISSUE #65: One click release
When the time comes, one click / action should be all we need to do create a "release" that provides us with download-able production binaries.

thomaseizinger:
These are mostly copy-pasted from xmr-btc-swap modulo the changelog
handling (because we don't have a changelog).

The gist is:

- We attach release binaries to every release that is created.
- Every merge to master creates a `preview` release.
- A merge from a branch that is named `release/x.y.z` creates a
release `x.y.z`.
- A new release branch can be created via `draft-new-release.yml`.

Fixes #65.

Example runs of these workflows can be found here: https://github.com/thomaseizinger/hermes/actions

### DIFF
--- a/.github/workflows/build-release-binary.yml
+++ b/.github/workflows/build-release-binary.yml
@@ -1,0 +1,129 @@
+name: "Build release binary"
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build_binaries:
+    name: Build release binary
+    strategy:
+      matrix:
+        include:
+          - bin: taker
+            target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            archive_ext: tar
+          - bin: taker
+            target: armv7-unknown-linux-gnueabihf
+            os: ubuntu-latest
+            archive_ext: tar
+          - bin: taker
+            target: x86_64-apple-darwin
+            os: macos-latest
+            archive_ext: tar
+          - bin: taker
+            target: x86_64-pc-windows-msvc
+            os: windows-latest
+            archive_ext: zip
+          - bin: maker
+            target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            archive_ext: tar
+          - bin: maker
+            target: armv7-unknown-linux-gnueabihf
+            os: ubuntu-latest
+            archive_ext: tar
+          - bin: maker
+            target: x86_64-apple-darwin
+            os: macos-latest
+            archive_ext: tar
+          - bin: maker
+            target: x86_64-pc-windows-msvc
+            os: windows-latest
+            archive_ext: zip
+    runs-on: ${{ matrix.os }}
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-release-binary.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Install compiler for armhf arch
+        if: matrix.target == 'armv7-unknown-linux-gnueabihf'
+        run: |
+          sudo apt-get update
+          sudo apt-get install gcc-arm-linux-gnueabihf
+
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '14'
```
bonomat: Any particular reason why version 14? 
I'm on 16.2

bonomat (approved):
Thank you. 
The actions still mention `changelog` here and there but that's not an issue.

.github/workflows/build-release-binary.yml
thomaseizinger: Nope, just didn't know there was already 16!

da-kami (approved):

bonomat (approved):
Looks good

.github/workflows/build-release-binary.yml
```diff
@@ -0,0 +1,129 @@
...
+jobs:
+  build_binaries:
+    name: Build release binary
+    strategy:
+      matrix:
+        include:
+          - bin: taker
+            target: x86_64-unknown-linux-gnu
+            os: ubuntu-latest
+            archive_ext: tar
```
klochowicz, line 16: out of curiosity: why `.tar` and not more conventional `.tar.gz`?

.github/workflows/build-release-binary.yml
```diff
@@ -0,0 +1,129 @@
...
+          - bin: taker
+            target: armv7-unknown-linux-gnueabihf
+            os: ubuntu-latest
+            archive_ext: tar
+          - bin: taker
+            target: x86_64-apple-darwin
+            os: macos-latest
+            archive_ext: tar
```
klochowicz, line 24: on macos `zip` is probably more convenient choice, as it automatically unpacked by Safari when extracted

klochowicz (approved):

.github/workflows/build-release-binary.yml
thomaseizinger: We are not compressing things so it is just an archive!

.github/workflows/build-release-binary.yml
thomaseizinger: Do you mind sending a follow-up patch if you care enough? :D

Testing Github workflows is a pain ...

klochowicz: I'll look into it at some point, when all the MVP tasks are done :)

PR #146, opened 2021-09-24
Convert `send_wire_message_actor` to xtra::Actor

thomaseizinger:
This will give us a generalized `Sink` interface for writing to the
connection.


klochowicz:
I just finished looking at it in my local checkout, and was just about to post "approve", but couldn't find the PR anymore :D thanks for fixing this!

PR #147, opened 2021-09-24
Fix sql

da-kami:
That was a tricky one 😅

This should fix all issues where we get an error complaining that we try to transition to `Accepted` even though the CFD was already accepted (which it was not...).
This problem might be the source of several weird side effects.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -263,7 +263,7 @@ pub async fn load_cfd_by_order_id(
             cfds.quantity_usd as quantity_usd,
             cfd_states.state as state
         from cfds as cfds
-        inner join orders as orders on cfds.order_uuid = ?
+        inner join orders as orders on cfds.order_id = orders.id
         inner join cfd_states as cfd_states on cfd_states.cfd_id = cfds.id
         where cfd_states.state in (
             select
@@ -273,6 +273,7 @@ pub async fn load_cfd_by_order_id(
             order by id desc
             limit 1
         )
+        and orders.uuid = ?
         "#,
         order_uuid
     )
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -473,6 +473,52 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_insert_and_load_cfd_by_order_id_multiple() {
+        let pool = setup_test_db().await;
+        let mut conn = pool.acquire().await.unwrap();
+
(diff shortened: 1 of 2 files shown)


daemon/src/db.rs
```diff
@@ -472,6 +473,56 @@ mod tests {
...
+
+        let order_id = order.id;
+
+        insert_order(&order, &mut conn).await.unwrap();
+        insert_cfd(cfd.clone(), &mut conn).await.unwrap();
+
+        let cfd_from_db = load_cfd_by_order_id(order_id, &mut conn).await.unwrap();
+        assert_eq!(cfd, cfd_from_db);
+
+        // insert second cfd with order-id that is higher in alphabet
```
thomaseizinger: You gotta be joking ...

thomaseizinger (approved):
Thank you for the elaborate commit message, that made things really easy to understand :)

daemon/src/db.rs
da-kami: when I saw your message I actually thought - yeah... that one needs more validation. 
It is actually a more general error. The UUID does not matter (I just had used the failing ones to reproduce it). I just flipped the UUIDs and it still works. Will see to generalize the test on Monday and remove the comment!

da-kami: One thing that I find surprising: sqlx does not complain when `fetch_one` returns more than one.

bonomat: Given `fetch_one` does not fail, shall we better not use it then?

thomaseizinger: The docs are somewhat ambiguous here. It sort of suggests that it fails if there are more than 2 results but also doesn't say that clearly anywhere.

Best to open an issue!

Not sure if we need to entirely avoid it but it is good to keep the behaviour in mind. If we do decide that we don't want it, I want it enforced via https://rust-lang.github.io/rust-clippy/rust-1.55.0/index.html#disallowed_method.

bonomat (approved):

daemon/src/db.rs
da-kami: I had another look, and to be fair the documentation states:

> Errors if no rows were returned. **Extra rows are ignored.** Aggregate queries, use this.

https://docs.rs/sqlx/0.5.1/sqlx/macro.query.html

This means we should probably avoid using it. I'd still go for merge now - we can re-visit this for refactoring later. Better get the fix in :)

klochowicz: @da-kami perhaps we could go in for merge now and create an issue to remove `fetch_one()` from the code?

klochowicz (approved):

PR #148, opened 2021-09-26
Beautify UI :) 

bonomat:
An attempt to make the UI a bit nicer :) 
Biggest change is using a table for the cfds. 

Note: there is an "unsorted" table. I opted for keeping it so that we immediately see if we changed states and a cfd does not fit into the existing states anymore :) 

Maker: 
image

image

Taker: 
image

image


frontend/src/components/cfdtables/CfdTable.tsx
```diff
@@ -0,0 +1,271 @@
...
+                            ? <IconButton
+                                aria-label="Reduce"
+                                icon={<ChevronUpIcon />}
+                                onClick={() => {
+                                    row.toggleRowExpanded();
+                                }}
+                            />
+                            : <IconButton
+                                aria-label="Expand"
+                                icon={<ChevronRightIcon />}
```
da-kami, line 36: nit: I think this should be a `ChevronDownIcon` to indicate that this can be "unfolded downwards".

frontend/src/Maker.tsx
```diff
@@ -111,11 +127,36 @@ export default function App() {
...
+                <GridItem colStart={3} colSpan={2}>
+                    {order && <OrderTile order={order} />}
+                </GridItem>
             </Grid>
+            <Tabs>
+                <TabList>
+                    <Tab>Running [{running.length}]</Tab>
+                    <Tab>Open [{open.length}]</Tab>
+                    <Tab>Closed [{closed.length}]</Tab>
+                    <Tab>Unsorted [{unsorted.length}] (should be empty)</Tab>
```
da-kami, line 142: Why does this column exist, what is it for? Can we just remove it if it's always empty?

frontend/src/Taker.tsx
```diff
@@ -93,81 +108,98 @@ export default function App() {
         },
     });
 
+    const runningStates = ["Request sent", "Requested", "Contract Setup", "Pending Open"];
```
da-kami, line 111: Creating a mapping / state handling within the UI just makes our life harder and I would do it. 
We might as well just display what the daemon sends us (and do the mapping there, so we can enforce the mapping through rust code).

We already map the internal state of the application in `to_sse_event.rs`. I would just adapt the mapping to whatever we need and then the UI will just display what is sent. We might want to extend this mapping to separate label from state, so we can change the label easily while the underlying state does not have to change.

da-kami (commented):
Thanks for working on improving the UI!

Overall, I like the idea of creating data tables. Besides my few comments on specific code parts here are some overall comments: 

- For me the table header looks odd, as my eyes cannot make out consistent text alignment there. Is it possible to to change the header text-align to left instead of center? (I assume that this might improve things, but I am not sure.)
- We might want to invest another round of moving things around to layout them better. I feel we are not using the space efficiently. The CFD list should be more present and not at the bottom. (Happy to get this in and re-arrange things on top of the current work)

frontend/src/components/cfdtables/CfdTable.tsx
bonomat: I wasn't sure, it looked weird when I looked at it :D

frontend/src/Maker.tsx
bonomat: It's only _temporarily_ to detect if backend and frontend get out of sync in regards to the CFD states. E.g. if we introduce a new _running_ state in the backend but not in the frontend we might wonder why the CFD does not show up under `running`. It will then show up in `unsorted` and requires adaptation. Eventually this tab should disappear.

frontend/src/Taker.tsx
bonomat: Definitely worth to reconsider this. My goal was only on creating a _nicer_ ui.

bonomat:
> Thanks for working on improving the UI!
> 
> Overall, I like the idea of creating data tables. Besides my few comments on specific code parts here are some overall comments:
> 
>     * For me the table header looks odd, as my eyes cannot make out consistent text alignment there. Is it possible to to change the header text-align to left instead of center? (I assume that this might improve things, but I am not sure.)
> 

I did not play around with this much. It is definitely possible. The approach for this would be adding a css to the table. 

>     * We might want to invest another round of moving things around to layout them better. I feel we are not using the space efficiently. The CFD list should be more present and not at the bottom. (Happy to get this in and re-arrange things on top of the current work)

I liked this layout and would move a chart into the middle once we have one. I'm thinking of either a price chart or a payout curve for the current offer.

da-kami (approved):
Overall ack - feel free to do improvements as follow ups.

frontend/src/components/cfdtables/CfdTable.tsx
da-kami: For me it feels weirder at the moment, because for me the right chevron indicates a link that takes me somewhere else. 
I think down is correct, but we can tackle that later :)

bonomat: I'll leave that up to you ok? No strong feeling on this one :)

PR #149, opened 2021-09-27
Add a script that can start all processes in a single terminal

klochowicz:
Not suitable for all use-cases as the logs are jumbled up, but it makes
regression testing more approachable.

Hitting a single `ctrl+c` stops all processes.


thomaseizinger (approved):
Nice!

PR #150, opened 2021-09-27
Share code for handling monitoring events to enable taker

da-kami:
Given the symmetry of the protocol all the state transitions should be exactly the same on the taker and maker side.

State transitions logic moved into the `Cfd` - I propose to do that for all other transitions as well (eventually).

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -4,7 +4,8 @@ use anyhow::{bail, Result};
 use bdk::bitcoin::secp256k1::{SecretKey, Signature};
 use bdk::bitcoin::{Address, Amount, PublicKey, Transaction};
 use bdk::descriptor::Descriptor;
-use cfd_protocol::secp256k1_zkp::EcdsaAdaptorSignature;
+use cfd_protocol::secp256k1_zkp::{EcdsaAdaptorSignature, SECP256K1};
+use cfd_protocol::{finalize_spend_transaction, spending_tx_sighash};
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use serde::{Deserialize, Serialize};
@@ -542,6 +543,35 @@ impl Cfd {
         };
 
         Ok(new_state)
+    }
+
+    pub fn refund_tx(&self) -> Result<Transaction> {
+        let dlc = if let CfdState::MustRefund { dlc, .. } = self.state.clone() {
+            dlc
+        } else {
+            bail!("Refund transaction can only be constructed when in state MustRefund, but we are currently in {}", self.state.clone())
+        };
+
+        let sig_hash = spending_tx_sighash(
+            &dlc.refund.0,
+            &dlc.commit.2,
+            Amount::from_sat(dlc.commit.0.output[0].value),
+        );
+        let our_sig = SECP256K1.sign(&sig_hash, &dlc.identity);
+        let our_pubkey = PublicKey::new(bdk::bitcoin::secp256k1::PublicKey::from_secret_key(
+            SECP256K1,
+            &dlc.identity,
+        ));
+        let counterparty_sig = dlc.refund.1;
+        let counterparty_pubkey = dlc.identity_counterparty;
+        let signed_refund_tx = finalize_spend_transaction(
+            dlc.refund.0,
+            &dlc.commit.2,
+            (our_pubkey, our_sig),
+            (counterparty_pubkey, counterparty_sig),
+        )?;
+
+        Ok(signed_refund_tx)
     }
 }
 
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1,5 +1,5 @@
 use crate::model::{Leverage, Position, TakerId, TradingPair, Usd};
-use crate::monitor::CfdMonitoringEvent;
+use crate::monitor;
 use anyhow::{bail, Result};
 use bdk::bitcoin::secp256k1::{SecretKey, Signature};
 use bdk::bitcoin::{Address, Amount, PublicKey, Transaction};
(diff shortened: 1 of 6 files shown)


thomaseizinger (commented):
What do you think of putting these transitions onto the `Cfd` itself? That would allow us to keep the actual transition functions side-effect free.

da-kami:
> What do you think of putting these transitions onto the `Cfd` itself? That would allow us to keep the actual transition functions side-effect free.

I am not sure I can follow, how do you imagine this?
1. A `transtition_to` function that takes the connection and saves to the db?
2. A `transition_to` function that returns the new state the Cfd should transition to (or an error)?
3. A `transition_to` function that mutates the Cfd?

I guess 2.

I agree that it might be nice to have the state transitions logic bundled, so state machine becomes more obvious. We might want to tackle this as a refactor on top though, because I think all transitions should move there then (not only the newly introduced ones).

thomaseizinger:
> because I think all transitions should move there then (not only the newly introduced ones).

Doesn't have to be all at once though :)
We can merge this as is but I overall I it would be good to transition to that when it comes to sharing code between the daemons. How and when things are saved to the DB is independent of the transitions.

da-kami:
> What do you think of putting these transitions onto the `Cfd` itself? That would allow us to keep the actual transition functions side-effect free.

done. Not sure that is exactly what you had in mind but I think it goes in the right direction.

daemon/src/model/cfd.rs
```diff
@@ -363,6 +364,192 @@ impl Cfd {
...
+
+    pub fn transition_to(&self, event: CfdStateChangeEvent) -> Result<CfdState> {
+        use CfdState::*;
+
+        // TODO: Display impl
+        tracing::info!("Cfd state change event {:?}", event);
+
+        let order_id = self.order.id;
+
+        let new_state = match event {
```
thomaseizinger, line 377: If you wouldn't bundle all the events in an enum but have `handle_lock_finality` etc then you would get rid of one level of indentation within all these functions, potentially making things more readable.

daemon/src/model/cfd.rs
```diff
@@ -363,6 +364,192 @@ impl Cfd {
 
     #[allow(dead_code)]
     pub const CET_TIMELOCK: u32 = 12;
+
+    pub fn transition_to(&self, event: CfdStateChangeEvent) -> Result<CfdState> {
```
thomaseizinger: I think there might be a better name for this function. `transition_to` implies that you are passing another state but actually you are passing an `event`. Events are usually "handled".

daemon/src/monitor.rs
```diff
@@ -147,33 +142,30 @@ impl xtra::Message for StartMonitoring {
...
-impl xtra::Message for CommitFinality {
-    type Result = ();
-}
-
-pub struct CetTimelockExpired(pub OrderId);
-
-impl xtra::Message for CetTimelockExpired {
-    type Result = ();
+#[derive(Debug, Clone)]
+pub enum CfdMonitoringEvent {
```
thomaseizinger: When naming things, try to make use of the modules name you are currently in.

This module is called `monitor`, so referring to this type as `monitor::Event` would give enough context.

daemon/src/taker_cfd.rs
```diff
@@ -228,12 +232,62 @@ impl Actor {
...
+        //  Consider sending a message to ourselves to trigger broadcasting refund?
+        if let CfdState::MustRefund { .. } = new_state {
+            let signed_refund_tx = cfd.refund_tx()?;
+            let txid = self
+                .wallet
+                .try_broadcast_transaction(signed_refund_tx)
+                .await?;
+
+            tracing::info!("Refund transaction published on chain: {}", txid);
+        }
```
thomaseizinger: Didn't we want triggering of these things up to the user?

thomaseizinger (commented):
Thanks! Definitely an improvement :)

thomaseizinger (approved):
Meant to approve!

daemon/src/taker_cfd.rs
da-kami: When implementing this with @luckysori we went by the convention: If the user has a choice we bubble it up, if the user does not have a choice we act. Refund is no choice. To my understanding, if the refund timelock expired you have to refund. Or is there a scenario where it is save to do anything but refund then? 

We can still give the user a choice based on the state and only trigger the publication upon user request. Shall I record it as a ticket?

daemon/src/model/cfd.rs
da-kami: The idea was to prepare this so that **all** state transitions would be handle in one function (eventually). 
Granted, there is only one type of events at the moment, but I think it's OK to outline this. 
If the indentation is annoying we can also match on the complete states instead of indenting. Will keep that for now if that's OK.

daemon/src/taker_cfd.rs
luckysori: The assumption is that you only reach this state if the oracle never attests a price and the refund timelock has expired. At this stage I don't think the application should wait for anything: it should refund automatically.

daemon/src/model/cfd.rs
```diff
@@ -363,6 +365,221 @@ impl Cfd {
...
+            (counterparty_pubkey, counterparty_sig),
+        )?;
+
+        Ok(signed_refund_tx)
+    }
+}
+
+#[derive(Debug, Clone)]
+pub enum CfdStateChangeEvent {
+    // TODO: groupd other events by actors into enums and add them here so we can bundle all
```
luckysori, line 580: :question: Is this TODO up-to-date?

luckysori (approved):
Thanks for cleaning up, looks great!

daemon/src/model/cfd.rs
da-kami: It is meant for all other state transitions (outside of monitoring). the idea is to bundle *all* state transitions of a Cfd in this function. 
That said, it is still valid (but has a typo...)

PR #152, opened 2021-09-27
Add meta_state to cfd for simplifying ui

bonomat:
followup from #148 

On a granular level, a CFD can be in multiple states. On an abstract level a CFD is either open (hasn't been accepted/rejected yet), running (it's ongoing) or closed (it has been terminated).
By introducing this meta state into our cfd model we simplify the UI logic.

### DIFF
--- a/daemon/src/to_sse_event.rs
+++ b/daemon/src/to_sse_event.rs
@@ -1,6 +1,6 @@
 use crate::model;
 use crate::model::cfd::OrderId;
-use crate::model::{Leverage, Position, TradingPair, Usd};
+use crate::model::{Leverage, TradingPair, Usd};
 use bdk::bitcoin::Amount;
 use rocket::response::stream::Event;
 use serde::Serialize;
@@ -25,8 +25,49 @@ pub struct Cfd {
     pub profit_btc: Amount,
     pub profit_usd: Usd,
 
-    pub state: String,
+    pub state: CfdState,
     pub state_transition_timestamp: u64,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct Position {
+    label: model::Position,
+    colorScheme: ColorScheme,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct CfdState {
+    pub label: String,
+    pub meta_state: CfdMetaState,
+    pub color_scheme: ColorScheme,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum CfdMetaState {
+    /// A CFD which is still being set up (not on chain yet)
+    Opening,
+    AcceptReject,
+    /// A CFD that is an ongoing open position (on chain)
+    Open,
+    /// A CFD that has been successfully or not-successfully terminated
+    Closed,
+}
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum ColorScheme {
+    Gray,
+    Green,
+    Red,
+    Orange,
+    Purple,
+}
+
+impl Default for ColorScheme {
+    fn default() -> Self {
+        ColorScheme::Gray
+    }
 }
 
 #[derive(Debug, Clone, Serialize)]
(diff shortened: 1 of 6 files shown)


daemon/src/to_sse_event.rs
```diff
@@ -25,10 +25,26 @@ pub struct Cfd {
...
+    pub label: String,
+    pub meta_state: CfdMetaSate,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub enum CfdMetaSate {
+    /// A CFD which has not been accepted or rejected
+    Open,
+    /// A CFD is which is on going
+    Running,
```
da-kami: I find the `Open` vs. `Running` confusing. I think `Running` should be `Open`. 
For me `Open` means that your position is open, and in our context the position is open once we are on chain. Anything before is not `Open` but `ActiveOrders` (or `Opening` or `Setup` ...).

da-kami (changes requested):

daemon/src/to_sse_event.rs
bonomat: I'm happy to go for different terminology. 

On Bitmex they have: 

`Positions`: these are the current running positions
`Active Orders`: these are the orders which are not yet filled
`Order History`: passed positions
(I ignored the other states on Bitmex).

da-kami: I'm happy to just go with the same model as Bitmex when mapping the internal states to meta states for the UI! Go for it.

---

Terminology wise: `Open` as we defined it in the daemon means the Cfd is actually an open position which in our context means it is on chain. I would still keep that internally in the daemon. (https://github.com/comit-network/hermes/blob/master/daemon/src/model/cfd.rs#L182-L188)

bonomat: How about these concrete enum variants:

`Opening`: has not been accepted or rejected yet
`Open`: has been accepted and is running (independent of transactions being included on-chain or not
`Closed`: has been rejected or it's over

da-kami: > How about these concrete enum variants:
> 
> `Opening`: has not been accepted or rejected yet
> `Open`: has been accepted and is running (independent of transactions being included on-chain or not
> `Closed`: has been rejected or it's over

I don't think `Closed` is enough, another suggestion:

`Opening`: has not been accepted or rejected yet
`Open`: has been accepted and is running (independent of transactions being included on-chain or not
`Closed`: position that was once open and then was closed (currently maps final state `Refunded` only, other final states are still missing)
`Other`: `Rejected`, `SafelyAborted` (see #160)

Feel free to chose a different name than `Other` - e.g. `Aborted`, `Failed` (...).

bonomat: Why would you not combine `Rejected`/`SafelyAborted` with `Closed`?

klochowicz: just throwing in another suggestion, perhaps a mid-way one: how about `Closed { Settled, Rejected, SafelyAborted}` - this would keep the main state from ballooning up, and at the same time allow us to keep the information how it ended?

da-kami: > Why would you not combine `Rejected`/`SafelyAborted` with `Closed`?

My reasoning woudl have been, that the user will likely not care about `SafelyAborted` / `Rejected` at all because those never made it on chain - no actual money was involved.  
But the user *might* want to see all the `Closed` ones (as in, he closed the position) because that determines wins/losses. That's why I would separate. Bitmex does not have `Rejected`/`SafelyAborted` because that cannot happen there I think :D 

Probably not important though, we can group them for now.

da-kami: > just throwing in another suggestion, perhaps a mid-way one: how about `Closed { Settled, Rejected, SafelyAborted}` - this would keep the main state from ballooning up, and at the same time allow us to keep the information how it ended?

We are only discussing how to group in the UI and not the internal state.

da-kami:
I'll take this over, will adapt this PR and then build the cfd actions on top of it.

da-kami:
After taking a step back and re-thinking the problem:

The actual problem is, that we pass the internal `model::CfdState` out. We should not do that, but create a proper mapping to a `to_sse_event::CfdState` that soul be it's own `enum` that represents the state the UI needs.
Currently we just map the internal state to a string, which results in leaking all the internal state to the UI.

Actions can then live on the `to_sse_event::CfdState` enum.

Any color / group (...) mapping can then be in the UI, because we have a "stable" API anyway (stable in the way that it won't break if we change the `model::CfdState`.

Closing this for now, will rework :)

PR #153, opened 2021-09-27
Refactor setup_contract to be just an async function 

thomaseizinger:


daemon/src/maker_inc_connections.rs
```diff
@@ -61,7 +61,7 @@ impl Actor {
             .write_connections
             .get(&taker_id)
             .context("no connection to taker_id")?;
-        conn.do_send_async(msg).await?;
+        conn.send(msg).await?;
```
da-kami, line 51: Could you maybe add a comment to this `send` so we are aware that it should not be changed to be async?

daemon/src/maker_cfd.rs
```diff
@@ -290,7 +290,7 @@ impl Actor {
         .unwrap();
 
         self.takers
-            .do_send_async(maker_inc_connections::TakerMessage {
+            .send(maker_inc_connections::TakerMessage {
```
da-kami, line 306: Could you maybe add a comment to this `send` so we are aware that it should not be changed to be async?

daemon/src/maker_cfd.rs
```diff
@@ -50,11 +51,16 @@ pub struct Actor {
     order_feed_sender: watch::Sender<Option<Order>>,
     takers: Address<maker_inc_connections::Actor>,
     current_order_id: Option<OrderId>,
-    current_contract_setup: Option<mpsc::UnboundedSender<SetupMsg>>,
-    // TODO: Move the contract setup into a dedicated actor and send messages to that actor that
-    // manages the state instead of this ugly buffer
-    contract_setup_message_buffer: Vec<SetupMsg>,
```
da-kami, line 58: sweet, I was hoping we could get rid of those buffers :)

daemon/src/maker_cfd.rs
```diff
@@ -275,19 +284,22 @@ impl Actor {
             }
         };
 
-        // Update order in db
-        let order_id = cfd.order.id;
+        let (sender, receiver) = mpsc::unbounded();
+
+        if let SetupState::Active { .. } = self.setup_state {
```
da-kami: 🛠 I think this early exit should be at the top of the function, we should potentially not even print the `Maker accepts an order` if we can't.

daemon/src/maker_cfd.rs
```diff
@@ -298,66 +310,36 @@ impl Actor {
...
 
         tokio::spawn(async move {
-            address
-                .do_send_async(CfdSetupCompleted {
-                    order_id,
-                    dlc: actor.await,
-                })
+            let dlc = contract_future.await;
+
+            this.do_send_async(CfdSetupCompleted { order_id, dlc })
```
da-kami, line 336: ❓ Is this one OK to be async or should it also be `send`? 
I think we want to ensure that we handle the completion, but it is not critical to be async. Should be fine being async, but I'm not 100% about side effects.

daemon/src/setup_contract.rs
```diff
@@ -0,0 +1,328 @@
...
+    commit_descriptor, compute_adaptor_point, create_cfd_transactions, interval, lock_descriptor,
+    spending_tx_sighash, PartyParams, PunishParams,
+};
+use futures::stream::FusedStream;
+use futures::{Sink, SinkExt, StreamExt};
+use std::collections::HashMap;
+use std::ops::RangeInclusive;
+
+/// Given an initial set of parameters, sets up the CFD contract with the other party.
+/// Passing OwnParams identifies whether caller is the maker or the taker.
```
da-kami: 🧹 We are not passing `OwnParams` anymore, I'd remove that comment.

da-kami (approved):

daemon/src/maker_inc_connections.rs
thomaseizinger: Would extending the commit message be acceptable?

daemon/src/maker_cfd.rs
thomaseizinger: Should be fine because nothing happens after this function, i.e. the current task just dies :)

daemon/src/setup_contract.rs
thomaseizinger: Well spotted, thank you!

bonomat (commented):
I think it looks good, but I can't say much about these changes unfortunately.

da-kami (approved):
Let's merge it! :)

Issue #108, opened 2021-09-22
Price recommendation for the maker's sell offer creation


da-kami:
What we currently have:

![image](https://user-images.githubusercontent.com/5557790/134262042-b992cb2d-629f-4516-9ee0-c679e7968e94.png)

The maker currently has to enter a price manually when pricing the CFD. 
This is quite cumbersome and can be optimized.

In an initial iteration I propose that we change the fields to:

1. Instead of an editable field `OrderPrice` have an editable field `Spread` (in percent) that is defaulted to `3%`
2. Maker the `OrderPrice` field non-editable
3. Calculate the `OrderPrice` from current price and spread.

(3.) can easily be achieved with a calculation endpoint (similar to the taker's `/calculate/margin`) that is triggered on either a price change or a spread change. 

Note: This ticket is depending on having a price feed as described in #96 

Note: This could eventually be extended with automated offer re-publication,  but that is out of scope for this ticket.

da-kami:
One thing we should definitely do once we have a price feed is change the `Order Price` initial value upon startup to be in relation to the price feed's price.

da-kami:
I see this as a `ready-to-play` ticket. For the `happy-path-clickthrough` it might as well just be the editable price field.

thomaseizinger:
The idea of `ready-to-play` was to be a lot more _bare-bones_ than `happy-path-clickthrough`! It should only contain things that stand inbetween us and doing a release and using it whatever "using" means at that point in time from a feature-set perspective.

bonomat:
This seems to be done: we pre-load the order price with a number and add a spread.

PR #154, opened 2021-09-27
Display price feed in the UI and use it for profit margin calculations in the CFDs.

klochowicz:


frontend/src/TakerApp.tsx
```diff
@@ -123,12 +125,9 @@ export default function App() {
                 <Grid templateColumns="repeat(6, 1fr)" gap={4}>
                     <GridItem colStart={1} colSpan={2}>
                         <Wallet walletInfo={walletInfo} />
+                        {/*TODO: Do we need this? does it make sense to only display the price from the order?*/}
```
thomaseizinger: Is this still up2date?

frontend/src/components/CurrentPrice.tsx
```diff
@@ -0,0 +1,41 @@
+import { Box, Center, Divider, HStack, Skeleton, Text } from "@chakra-ui/react";
+import React from "react";
+import TimeStamp from "./TimeStamp";
+import { PriceInfo } from "./Types";
+
+interface PriceProps {
+    priceInfo: PriceInfo | null;
+}
```
thomaseizinger: Would just call this `Props`. If we stick to one component per file we don't need to qualify.

frontend/src/components/CurrentPrice.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export default function Wallet(
```
thomaseizinger: Copy pasta error.

frontend/src/components/TimeStamp.tsx
```diff
@@ -0,0 +1,26 @@
+import { Text } from "@chakra-ui/react";
+import React from "react";
+import { unixTimestampToDate } from "./Types";
+
+interface TimeStampProps {
+    timestamp: number;
+}
+
+export default function TimeStamp(
```
thomaseizinger: I think `Timestamp` is one word, no need to capitalize.

daemon/src/routes_taker.rs
```diff
@@ -37,11 +36,11 @@ pub async fn feed(
         let order = rx_order.borrow().clone();
         yield order.to_sse_event();
 
-        let cfds = rx_cfds.borrow().clone();
-        yield cfds.to_sse_event();
         let quote = rx_quote.borrow().clone();
         yield quote.to_sse_event();
 
+        let cfds_with_price = CfdsWithCurrentPrice{cfds: rx_cfds.borrow().clone(), current_price: quote.for_taker()};
+        yield cfds_with_price.to_sse_event();
```
thomaseizinger: Instead of doing this, the `maker_cfd::Actor` should also listen for price updates and update its internal state accordingly. It can then send out updates to the frontend.

This will ensure that we are consistent and for example send the correct offer to all connected takers.

thomaseizinger (changes requested):

frontend/src/TakerApp.tsx
klochowicz: removed, as we're likely to have the price feed.

daemon/src/routes_taker.rs
klochowicz: I've reworked it to make the updates inside  the CFD actors... I don't think it looks much better TBH (perhaps I misunderstood your intention?).

The updates are just for the feed itself, the augmented CFDs were not used within the actor for anything other than for sending the updates to the UI.

Sending the correct offer to the takers is not part of this review, but it indeed would require the maker to listen for the price updates (which is possible with the latest update).

Happy to discuss it tomorrow if I misundestood your comment ;D

daemon/src/bitmex_price_feed.rs
```diff
@@ -67,6 +68,25 @@ impl Quote {
...
+
+    #[allow(dead_code)] // Not used by all binaries.
+    pub fn for_maker(&self) -> Usd {
+        // XXX: is this the correct assumption?
+        self.ask
+    }
+
+    #[allow(dead_code)] // Not used by all binaries.
+    pub fn for_taker(&self) -> Usd {
+        // XXX: is this the correct assumption?
```
da-kami: I think it should be the `bid` price, if we assume an arbitrage model where the seller sells what he gets back on the market. The seller sells at the `bid` price, so the seller is covered with the `bid`. 
Note: In the end mid_range is fine too - we can always make this part of the interest calc (which is currently not implemented yet). 
Feel free to just go with mid_range for now. 
Please either convert the `XXX` to be a `TODO` (if we should track this further), or just remove the comment.

daemon/src/maker.rs
```diff
@@ -204,6 +195,31 @@ async fn main() -> Result<()> {
                 });
 
                 tokio::spawn(maker_inc_connections_address.attach_stream(listener_stream));
+
+                // XXX: Is there a way to not repeat this in both maker and taker?
```
da-kami: I would not bother too much, duplicate for now. As long as it is gluecode that's fine. 

Please remove the `XXX` comments before merging - or convert them to `TODO`. I would avoid having multiple different types of actionable comments in there. Will make it hard to find them.

daemon/src/taker.rs
```diff
@@ -192,6 +183,29 @@ async fn main() -> Result<()> {
                     cfd_actor_inbox.clone(),
                 )));
                 tokio::spawn(wallet_sync::new(wallet, wallet_feed_sender));
+                // XXX: Is there a way to not repeat this in both maker and taker?
```
da-kami: `XXX` -> `TODO` or remove

daemon/src/bitmex_price_feed.rs
```diff
@@ -67,6 +68,25 @@ impl Quote {
             ask: Usd::from(Decimal::try_from(quote.ask_price)?),
         }))
     }
+
+    #[allow(dead_code)] // Not used by all binaries.
+    pub fn for_maker(&self) -> Usd {
+        // XXX: is this the correct assumption?
```
da-kami: `XXX` -> `TODO` or remove

da-kami (approved):
From my side this is good for merge, can do improvements on top. 
Would be cool if you could provide a screenshot of how it looks in the UI :)

daemon/src/maker.rs
klochowicz: I can leave it as-is, the question was more to find out what sort of Rust wizardry I still have to learn :D 

I typically put `XXX` when I ask questions that are meant for the review, but which should stay inside when merged. So, don't worry, I'm onboard with sticking to `TODO`s in the codebase 💯  happy to use `TODO` for that as well, but I didn't want to accidentally push some transient comments in.

daemon/src/cfd_feed.rs
```diff
@@ -0,0 +1,120 @@
+use std::time::UNIX_EPOCH;
+
+use anyhow::{Context as _, Result};
+use bdk::bitcoin::Amount;
+use serde::Serialize;
+use sqlx::pool::PoolConnection;
+use sqlx::Sqlite;
+use tokio::sync::watch;
+
```
thomaseizinger: Get rid of these empty lines please.

daemon/src/cfd_feed.rs
```diff
@@ -0,0 +1,120 @@
...
+        )
+    }
+
+    /// Update price from BitMex. It is recommended to call `update()` afterwards.
+    pub fn set_current_price(&mut self, quote: bitmex_price_feed::Quote) {
+        self.current_price = Some(quote);
+    }
+
+    /// Updates the CFD feed in the frontend
+    pub async fn update(&self, conn: &mut PoolConnection<Sqlite>) -> Result<()> {
```
thomaseizinger, line 68: I don't think this should take the database. A list of CFDs is sufficient and avoids a dependency between these modules.

daemon/src/cfd_feed.rs
```diff
@@ -0,0 +1,120 @@
...
+            .context("Could not update CFD feed")
+    }
+
+    fn calculate_profit(&self, cfd: &model::cfd::Cfd) -> (Amount, Usd) {
+        if let Some(quote) = &self.current_price {
+            cfd.profit(match self.role {
+                Role::Taker => quote.for_taker(),
+                Role::Maker => quote.for_maker(),
+            })
+            .unwrap()
```
thomaseizinger, line 112: Does the `Cfd` not know whether or not we are on the `maker` or `taker` side? That seems odd.

daemon/src/maker.rs
thomaseizinger: The thing to use here is `Address::attach_stream`. You need to compose a `Stream` out of `quote_updates` which you can do via `futures::stream::unfold`.

thomaseizinger: This `quotes_updates` -> `Stream` function can be shared between both binaries and then it is just a single function call for both.

daemon/src/maker_cfd.rs
```diff
@@ -68,21 +71,21 @@ impl Actor {
...
+        cfd_feed: CfdFeed,
         order_feed_sender: watch::Sender<Option<Order>>,
         takers: Address<maker_inc_connections::Actor>,
         monitor_actor: Address<monitor::Actor<Actor>>,
     ) -> Result<Self> {
         let mut conn = db.acquire().await?;
 
         // populate the CFD feed with existing CFDs
-        cfd_feed_actor_inbox.send(load_all_cfds(&mut conn).await?)?;
+        cfd_feed.update(&mut conn).await?;
```
thomaseizinger, line 82: I think this should happen outside of the `Actor` actually, there is no reason why we can't initialize the feed already with the list of CFDs.

daemon/src/maker_cfd.rs
```diff
@@ -418,6 +417,14 @@ impl Actor {
 
         Ok(())
     }
+
+    async fn handle_price_update(&mut self, quote: bitmex_price_feed::Quote) -> Result<()> {
+        self.cfd_feed.set_current_price(quote);
+
+        let mut conn = self.db.acquire().await?;
+        self.cfd_feed.update(&mut conn).await?;
```
thomaseizinger, line 425: I think a cleaner way of doing this is to calculate the profit of every CFD here and then send a tuple of `(CFD, Profit)` to the cfd_feed.

The feed should be "dumb" and all the crucial calculation should be within the actor. Conversion the the actual feed model should also happen outside the actor though which is why I would just send over a tuple. Alternatively you could create another domain model that combines CFD + profit.

daemon/src/to_sse_event.rs
```diff
@@ -127,13 +71,34 @@ impl ToSseEvent for model::WalletInfo {
...
-                .as_secs(),
+            last_updated_at: into_unix_secs(self.last_updated_at),
         };
 
         Event::json(&wallet_info).event("wallet")
     }
 }
+
+#[derive(Debug, Clone, Serialize)]
+pub struct PriceInfo {
```
thomaseizinger: Isn't this also just a `Quote`? Why invent new names?

thomaseizinger (approved):
We can merge this as is to move forward but on the whole, I think we could still make improvements to this design.

I don't think that the feed should be mutable for example.

(Approval means all comments optional)

daemon/src/maker_cfd.rs
klochowicz: What you described sounds a lot like the previous iteration of the PR, where the feed was "dumb" and I had the `CFD + profit` domain model. Wasn't that better in retrospect?

I struggle to see why the cfd actor needs to calculate the profit itself, as it does not make any use of it - it's just for the purpose of updating the feed. Doing it inside the CFD actors would mean:
a) need to store the current price (ie. quote) around, reducing encapsulation
b) need to duplicate that logic between 2 actors, as both maker and taker need to perform the calculations.

thomaseizinger: In the previous iteration, the feed still did the calculation of the profit yeah?

At the moment, the `maker_cfd::Actor` is responsible for updating the feed. I just don't like that being "by-passed" by other actors. Are we really never going to need to current price for anything else inside the `maker_cfd::Actor`? It just seems worthwhile to centralize things within the `cfd::Actor`.

daemon/src/cfd_feed.rs
klochowicz: Currently, it does not. It only tells "Origin", ie. "ours/theirs". That's not enough information for me to pick here.

daemon/src/maker_cfd.rs
klochowicz: I'll leave this optimisation out for now, as it's the same behaviour as before. can come back to it later.

daemon/src/maker.rs
klochowicz: thank you @thomaseizinger ! I'll give it a shot, but as a separate cleanup PR, to not to drag this one.

daemon/src/cfd_feed.rs
klochowicz: Without this, `cfd_feed` module makes no sense and we can go back to a simple `watch` channel + storing the last quote in both actors. Perhaps that would be better in the end? We could then a free function that calculates the cfd with profit, taking current quote and the database connection as the arguments. However, the function will need to live somewhere, so dependency on the database is still likely to happen.

(but at least we'd get rid of the mutability you mentioned elsewhere!)

daemon/src/bitmex_price_feed.rs
klochowicz: I had some chats with @bonomat  about this, and we decided to go with mid_range for now. As I'm not sure whether this is correct, I'll convert it to a TODO.

klochowicz:
> From my side this is good for merge, can do improvements on top. Would be cool if you could provide a screenshot of how it looks in the UI :)

I'll keep it in my mind for the future, that I should have sent a screenshot to showcase the UI change!

daemon/src/maker_cfd.rs
klochowicz: The feed was just a watch channel - the calculation was done in the `to_sse_event` module, as the current quote was available there (it's managed by Rocket after all).

I can't see the need for the current price in cfd actors as for now (even in #108, the offer price will be calculated with the spread outside of the actor, and then `CfdNewOrderRequest` will take the offer price in).
 I have no idea whether it might be needed for CFD rollover or interacting with oracle? 

My initial motivation for keeping it out was that display of profit in the frontend is an implementation detail from the perspective of this Actor, whilst CfdActor takes care of things like ensuring the database records are up to date and new cfds are created/settled with the other party.

daemon/src/bitmex_price_feed.rs
bonomat: For the taker it is just "informative" and `mid_market` makes sense. For the `maker` it should be the `ask` price, because that's the same rate he would get on Bitmex if he'd hedge his position. 
Note: in our hedging model, the `maker` has to `buy` on Bitmex as he is `selling` on Hermes - he is literally mirroring the position.

PR #157, opened 2021-09-28
Fix `MAX_PRICE_DEC`

luckysori:
2^20 is 21 bits long, _not_ 20 bits long. If we allow a bound equal to 2^20 we would need 21 bits to represent it, past our limit of 20.

Thanks @bonomat for making me test what `proptest` didn't.

### DIFF
--- a/cfd_protocol/src/interval.rs
+++ b/cfd_protocol/src/interval.rs
@@ -5,7 +5,7 @@ use std::ops::RangeInclusive;
 mod digit_decomposition;
 
 /// Maximum supported BTC price in whole USD.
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Your change subtracts `1` from the final price. Is that what you are intending to do?

luckysori:
> Your change subtracts `1` from the final price. Is that what you are intending to do?

Yep. `2^MAX_DIGITS` cannot be the max price because we would need `MAX_DIGITS + 1` digits to represent it. The highest number representable with `MAX_DIGITS` digits is `2^MAX_DIGITS - 1`.

thomaseizinger (approved):

PR #158, opened 2021-09-28
Use `Address::attach_stream` for incoming connections

thomaseizinger:


daemon/src/maker_inc_connections.rs
```diff
@@ -124,9 +144,22 @@ impl Handler<TakerMessage> for Actor {
...
+            ListenerMessage::NewConnection { stream, address } => {
+                self.handle_new_connection(stream, address).await;
+
+                KeepRunning::Yes
+            }
+            ListenerMessage::Error { source } => {
+                tracing::warn!("TCP listener produced an error: {}", source);
+
+                // Maybe we should move the actual listening on the socket into here and restart the
+                // actor upon an error?
```
klochowicz, line 159: an interesting idea, definitely worth exploring - is there a way of recovering from this otherwise? However, not sure whether we need it for the MVP?

klochowicz (approved):
Looks good!

daemon/src/maker_inc_connections.rs
thomaseizinger: > we need it for the MVP?

Probably not but something to keep in mind :)

Issue #69, opened 2021-09-16
Allow trader and broker to trigger non-collaborative settlement


thomaseizinger:
Following our decision on no-automation-empower-the-user, both parties should be able to trigger the non-collaborative settlement path by broadcasting the necessary transactions.

PR #160, opened 2021-09-28
Handle existing cfds on daemon startup

### LINKED ISSUE #137: Cfd handling upon daemon restart
Upon restarting a deamon the application has to be able to handle already existing Cfds.

This means, when starting the Cfd actor and the monitoring actor we have to reason about the Cfd state and might want to trigger certain interaction so the Cfd is properly loaded and displayed, as well as monitoring is picked up to trigger state changes. 

When implementing the monitor actor we had planned to just load all Cfds from the db and pass them to the monitor actor so it could trigger monitoring as needed based on the Cfd state. 
i.e.: match statement in the constructor of monitor actor that triggers the remaining part of the transaction monitoring based on the state.

For the Cfd actor we did not invest too much time in what a restart would require. The assumption is that it does no have to do anything, because it always gets messages from another actor to transition the Cfd into a new state. 
One thing we will have to think about is how to deal with Cfds that upon restart are in a state that is prior to `PendingOpen`. I think such Cfds should be transitioned to a state `SavelyAborted` by the Cfd actor upon a message from **whom?**
i.e. we don't go on with any Cfd that was not completely set up. 

Note: I would still not delete `SafelyAborted` Cfds for now, including all states (since we always append the state it will keep it unless that logic is changed). We can filter them out upon initial load (...). Once we have more confidence in the implementation we could add removing them.

da-kami:
fixes #137 

Cfd actor handles upon startup:

*  "cleaning up" cfds that were left in a state that cannot be continued (any state before `PendingOpen` will be transitioned to `SafelyAborted`)
* transaction (re-)publication for states that publish a transaction (at the moment: `PendingOpen` and `MustRefund`)
  * note: publishing commit transaction through user interaction is not done yet, see #69 (I will see to work on that next)

Monitor actor handles upon start:

* trigger necessary monitoring based on state.

### DIFF
--- a/daemon/src/cleanup.rs
+++ b/daemon/src/cleanup.rs
@@ -1,0 +1,27 @@
+use crate::db::{insert_new_cfd_state_by_order_id, load_all_cfds};
+use crate::model::cfd::{Cfd, CfdState, CfdStateCommon};
+use anyhow::Result;
+use sqlx::SqlitePool;
+use std::time::SystemTime;
+
+pub async fn transition_non_continue_cfds_to_setup_failed(db: SqlitePool) -> Result<()> {
+    let mut conn = db.acquire().await?;
+
+    let cfds = load_all_cfds(&mut conn).await?;
+
+    for cfd in cfds.iter().filter(|cfd| Cfd::is_cleanup(cfd)) {
+        insert_new_cfd_state_by_order_id(
+            cfd.order.id,
+            CfdState::SetupFailed {
+                common: CfdStateCommon {
+                    transition_timestamp: SystemTime::now(),
+                },
+                info: format!("Was in state {} which cannot be continued.", cfd.state),
+            },
+            &mut conn,
+        )
+        .await?;
+    }
+
+    Ok(())
+}
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -67,6 +67,7 @@ enum SetupState {
 }
 
 impl Actor {
+    #[allow(clippy::too_many_arguments)]
     pub async fn new(
         db: sqlx::SqlitePool,
         wallet: Wallet,
@@ -75,11 +76,25 @@ impl Actor {
         order_feed_sender: watch::Sender<Option<Order>>,
         takers: Address<maker_inc_connections::Actor>,
         monitor_actor: Address<monitor::Actor<Actor>>,
+        cfds: Vec<Cfd>,
     ) -> Result<Self> {
         let mut conn = db.acquire().await?;
 
         // populate the CFD feed with existing CFDs
         cfd_feed.update(&mut conn).await?;
+
+        for dlc in cfds.iter().filter_map(|cfd| Cfd::pending_open_dlc(cfd)) {
+            let txid = wallet.try_broadcast_transaction(dlc.lock.0.clone()).await?;
+
+            tracing::info!("Lock transaction published with txid {}", txid);
+        }
+
+        for cfd in cfds.iter().filter(|cfd| Cfd::is_must_refund(cfd)) {
+            let signed_refund_tx = cfd.refund_tx()?;
+            let txid = wallet.try_broadcast_transaction(signed_refund_tx).await?;
+
+            tracing::info!("Refund transaction published on chain: {}", txid);
+        }
 
         Ok(Self {
             db,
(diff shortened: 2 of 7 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -75,8 +75,57 @@ impl Actor {
     ) -> Result<Self> {
         let mut conn = db.acquire().await?;
 
+        let cfds = load_all_cfds(&mut conn).await?;
+
         // populate the CFD feed with existing CFDs
-        cfd_feed_actor_inbox.send(load_all_cfds(&mut conn).await?)?;
+        cfd_feed_actor_inbox.send(cfds.clone())?;
```
da-kami: ah, that is bogus - I will have to do that after the loop, because there might be potential state changes 😅
That is not optimal.

da-kami: Maybe it's better to move those state changes out of this actor after all - has to many side effects in here. I will see to move it to a `cleanup` module. I had that initially but decided it would be better to move this into the actor, because the cfd actor should be responsible for all state transitions...

daemon/src/monitor.rs
```diff
@@ -15,35 +16,144 @@ const FINALITY_CONFIRMATIONS: u32 = 1;
...
 
+    async fn monitor_all(&self, params: &MonitorParams, order_id: OrderId) {
+        self.monitor_lock_finality(params, order_id).await;
+
+        self.monitor_commit_finality_and_timelocks(params, order_id)
+            .await;
+
+        self.monitor_refund_finality(params.clone(), order_id).await;
+
+        // TODO: CET subscription => Requires information from Oracle
```
da-kami: This todo is misplaced, the CET monitoring has to be started upon message from the oracle. Better remove this.

daemon/src/cleanup.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    // Handle about existing cfds on startup
+    for cfd in cfds {
+        match cfd.state {
+
+            // Transition all Cfds in early states that cannot be continued to SafelyAborted
+            CfdState::OutgoingOrderRequest { .. }
+            | CfdState::IncomingOrderRequest { .. }
+            | CfdState::Accepted { .. }
+            | CfdState::ContractSetup { .. } => {
```
thomaseizinger: Doesn't this imply that `SafelyAborted` is actually just `SetupFailed`? `SafelyAborted` sounds like it was a conscious decision to abort but really we didn't have a choice here.

daemon/src/maker_cfd.rs
```diff
@@ -75,8 +75,52 @@ impl Actor {
...
+                // Responsibility of the monitor actor to start monitoring on startup and ping us
+                CfdState::Open { .. }
+                | CfdState::OpenCommitted { .. }
+
+                // Final states
+                | CfdState::Rejected { .. }
+                | CfdState::Refunded { .. }
+                | CfdState::SafelyAborted { .. } => ()
+            }
+        }
```
thomaseizinger: If you'd create a function on `Cfd` that does:

```rust
fn pending_open_dlc(&self) -> Option<Dlc> {
	if let CfdState::PendingOpen { dlc, .. } => {
		Some(dlc)
	} else {
		None
	}
}
```

Then you could write:

```rust

for dlc in cfds.iter().filter_map(Cfd::pending_open_dlc) {
	let txid = wallet.try_broadcast_transaction(dlc.lock.0.clone()).await?;

	tracing::info!("Lock transaction published with txid {}", txid);
}
```

daemon/src/cleanup.rs
thomaseizinger: Similar to the `filter_map`, you could shorted this a lot by moving the selection of states into a predicate function on the CFD.

daemon/src/taker.rs
```diff
@@ -172,8 +173,10 @@ async fn main() -> Result<()> {
 
                 let (monitor_actor_address, monitor_actor_context) = xtra::Context::new(None);
 
+                let mut conn = db.acquire().await.unwrap();
+                let cfds = load_all_cfds(&mut conn).await.unwrap();
```
thomaseizinger, line 168: If we are loading these out here already, I think we should also pass them into the `taker_cfd::Actor` instead of loading them inside the constructor again!

thomaseizinger (approved):
Thanks!

Some ideas of improvement.

daemon/src/taker.rs
da-kami: Not sure about this one, the cfd actor explicitly has access to the db and should always reason about the cfds that are in the db. 
If any actor does changes to the cfds in the db this list of cfds will not be up to date anymore. Currently this would work, because we know that there are no changes in between, but I would rather not optimize that. I think the cfd actor, given that it has access to the db, should always load from the db.

daemon/src/model/cfd.rs
```diff
@@ -593,6 +593,28 @@ impl Cfd {
...
+        } else {
+            None
+        }
+    }
+
+    pub fn filter_must_refund(&self) -> bool {
+        matches!(self.state.clone(), CfdState::MustRefund { .. })
+    }
+
+    pub fn filter_cleanup(&self) -> bool {
```
thomaseizinger: I don't think these would necessarily have to be named `filter_`. They could be used in other contexts as well.

For the latter two, a naming with `is_` would be more idiomatic.

The first one is simply a getter for the the dlc in the `pending_open` state (which is why I proposed `pending_open_dlc` originally). Getters are named without `get_` in Rust.

thomaseizinger (approved):

daemon/src/taker.rs
thomaseizinger: The thing is, multiple actors (like the monitoring one) need access to the list of CFDs that are initially present in the database upon startup. It seems unnecessary to load these multiple times. Additionally, to reach a consistent system, this should actually be the exact same list for every actor. Passing in a list is also synchronous which is a plus for a constructor IMO.

da-kami: fair enough, after Mariusz' changes merged I changed it because now I don't have the cfds anymore because we have a separate task for populating the feed. 
I still think we have to be careful with this approach though - the only thing that should ever change cfds in the db prior to the actors starting should be cleanup. And that is always run before anything starts up. We should avoid a situation where we pass an outdated list of cfds around.

da-kami:
squashed to make rebase easier

thomaseizinger (approved):
Thanks :)

PR #161, opened 2021-09-28
Allow more than one attestation event per CFD

bonomat:
Allow multiple payout attestation points

### DIFF
--- a/cfd_protocol/src/protocol.rs
+++ b/cfd_protocol/src/protocol.rs
@@ -21,6 +21,7 @@ use bdk::FeeRate;
 use itertools::Itertools;
 use secp256k1_zkp::{self, schnorrsig, EcdsaAdaptorSignature, SecretKey, Signature, SECP256K1};
 use std::collections::HashMap;
+use std::hash::Hasher;
 use std::iter::FromIterator;
 use std::num::NonZeroU8;
 use std::ops::RangeInclusive;
@@ -72,19 +73,19 @@ where
 /// * `taker` - The initial parameters of the taker.
 /// * `taker_punish_params` - The punish parameters of the taker.
 /// * `oracle_pk` - The public key of the oracle.
-/// * `nonce_pks` - One R-value per price digit signed by the oracle. Their order matches a big
-///   endian encoding of the price.
+/// * `cet_timelock` - Relative timelock of the CET transaction with respect to the commit
+///   transaction.
 /// * `refund_timelock` - Relative timelock of the refund transaction with respect to the commit
 ///   transaction.
-/// * `payouts` - All the possible ways in which the contract can be settled, according to the
-///   conditions of the bet.
+/// * `payouts_per_event` - All the possible ways in which the contract can be settled, according to
+///   the conditions of the bet. The key is the event at which the oracle will attest the price.
 /// * `identity_sk` - The secret key of the caller, used to sign and encsign different transactions.
 pub fn create_cfd_transactions(
     (maker, maker_punish_params): (PartyParams, PunishParams),
     (taker, taker_punish_params): (PartyParams, PunishParams),
-    (oracle_pk, nonce_pks): (schnorrsig::PublicKey, &[schnorrsig::PublicKey]),
+    oracle_pk: schnorrsig::PublicKey,
     (cet_timelock, refund_timelock): (u32, u32),
-    payouts: Vec<Payout>,
+    payouts_per_event: HashMap<Announcement, Vec<Payout>>,
     identity_sk: SecretKey,
 ) -> Result<CfdTransactions> {
     let lock_tx = lock_transaction(
@@ -109,9 +110,9 @@ pub fn create_cfd_transactions(
             taker.address,
             taker_punish_params,
         ),
-        (oracle_pk, nonce_pks),
+        oracle_pk,
         (cet_timelock, refund_timelock),
-        payouts,
+        payouts_per_event,
         identity_sk,
     )
 }
@@ -130,9 +131,9 @@ pub fn renew_cfd_transactions(
         Address,
         PunishParams,
     ),
-    (oracle_pk, nonce_pks): (schnorrsig::PublicKey, &[schnorrsig::PublicKey]),
+    oracle_pk: schnorrsig::PublicKey,
     (cet_timelock, refund_timelock): (u32, u32),
-    payouts: Vec<Payout>,
+    payouts_per_event: HashMap<Announcement, Vec<Payout>>,
     identity_sk: SecretKey,
 ) -> Result<CfdTransactions> {
     build_cfds(
@@ -149,9 +150,9 @@ pub fn renew_cfd_transactions(
             taker_address,
             taker_punish_params,
         ),
-        (oracle_pk, nonce_pks),
+        oracle_pk,
         (cet_timelock, refund_timelock),
-        payouts,
+        payouts_per_event,
         identity_sk,
     )
 }
@@ -170,9 +171,9 @@ fn build_cfds(
         Address,
         PunishParams,
     ),
-    (oracle_pk, nonce_pks): (schnorrsig::PublicKey, &[schnorrsig::PublicKey]),
+    oracle_pk: schnorrsig::PublicKey,
     (cet_timelock, refund_timelock): (u32, u32),
-    payouts: Vec<Payout>,
+    payouts_per_event: HashMap<Announcement, Vec<Payout>>,
     identity_sk: SecretKey,
 ) -> Result<CfdTransactions> {
     let commit_tx = CommitTransaction::new(
@@ -215,24 +216,31 @@ fn build_cfds(
         (tx.into_inner(), sig)
     };
 
-    let cets = payouts
-        .into_iter()
-        .map(|payout| {
-            let cet = ContractExecutionTx::new(
-                &commit_tx,
-                payout.clone(),
-                &maker_address,
-                &taker_address,
-                nonce_pks,
-                cet_timelock,
-            )?;
+    let mut cets = vec![];
+    for (event, payouts) in payouts_per_event.iter() {
+        let cets_tmp = payouts
+            .iter()
+            .map(|payout| {
+                let cet = ContractExecutionTx::new(
+                    &commit_tx,
+                    payout.clone(),
+                    &maker_address,
+                    &taker_address,
+                    event.nonce_pks.as_slice(),
+                    cet_timelock,
+                )?;
 
-            let encsig = cet.encsign(identity_sk, &oracle_pk)?;
+                let encsig = cet.encsign(identity_sk, &oracle_pk)?;
 
-            Ok((cet.into_inner(), encsig, payout.digits))
-        })
-        .collect::<Result<Vec<_>>>()
-        .context("cannot build and sign all cets")?;
+                Ok((cet.into_inner(), encsig, payout.digits.clone()))
+            })
+            .collect::<Result<Vec<_>>>()
+            .context("cannot build and sign all cets")?;
+        cets.push(Cets {
+            event: event.clone(),
+            cets: cets_tmp,
+        });
+    }
 
     Ok(CfdTransactions {
         lock: lock_tx,
@@ -341,8 +349,40 @@ pub struct PunishParams {
 pub struct CfdTransactions {
     pub lock: PartiallySignedTransaction,
     pub commit: (Transaction, EcdsaAdaptorSignature),
-    pub cets: Vec<(Transaction, EcdsaAdaptorSignature, interval::Digits)>,
+    pub cets: Vec<Cets>,
     pub refund: (Transaction, Signature),
+}
+
+/// Group of CETs associated with a particular oracle announcement.
+///
+/// All of the adaptor signatures included will be _possibly_ unlocked
+/// by the attestation corresponding to the announcement. In practice,
+/// only one of the adaptor signatures should be unlocked if the
+/// payout intervals are constructed correctly. To check if an adaptor
+/// signature can be unlocked by a price attestation, verify whether
+/// the price attested to lies within its interval.
+#[derive(Debug, Clone)]
+pub struct Cets {
+    pub event: Announcement,
+    pub cets: Vec<(Transaction, EcdsaAdaptorSignature, interval::Digits)>,
+}
+
+#[derive(Debug, Clone, Eq)]
+pub struct Announcement {
+    pub id: String,
+    pub nonce_pks: Vec<schnorrsig::PublicKey>,
+}
+
+impl std::hash::Hash for Announcement {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.id.hash(state)
+    }
+}
+
+impl PartialEq for Announcement {
+    fn eq(&self, other: &Self) -> bool {
+        self.id.eq(&other.id)
+    }
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


da-kami:
On first glance: A `Vec<schnorrsig::PublicKey>` as key of a `HashMap` seems a bit odd. Is that really what you want to depict? 

> everything blows up

Could you be a bit more concrete? 

Happy to have a look together if you want.

bonomat:
The second commit (https://github.com/comit-network/hermes/pull/161/commits/16c65979d99037ace99761f1bfe4c5cbfee1b548) was painful. I'm sure there is a better approach.

cfd_protocol/tests/cfds.rs
```diff
@@ -460,7 +460,8 @@ struct CfdKeys {
 fn verify_cfd_sigs(
     (maker_cfd_txs, maker_pk, maker_publish_pk): (&CfdTransactions, PublicKey, PublicKey),
     (taker_cfd_txs, taker_pk, taker_publish_pk): (&CfdTransactions, PublicKey, PublicKey),
-    (oracle_pk, nonce_pks): (schnorrsig::PublicKey, &[schnorrsig::PublicKey]),
+    // TODO remove unused field
```
bonomat: Removed in fixup commit.

daemon/src/model/cfd.rs
```diff
@@ -740,6 +741,11 @@ pub struct Dlc {
...
     pub commit: (Transaction, EcdsaAdaptorSignature, Descriptor<PublicKey>),
-    pub cets: Vec<(Transaction, EcdsaAdaptorSignature, RangeInclusive<u64>)>,
+    pub cets: CETs,
     pub refund: (Transaction, Signature),
 }
+
+type CETs = HashMap<
+    Vec<schnorrsig::PublicKey>,
+    Vec<(Transaction, EcdsaAdaptorSignature, RangeInclusive<u64>)>,
+>;
```
thomaseizinger: If we have a hashmap here, why did we need a different structure within the protocol? :thinking:

thomaseizinger (commented):
Will give this a proper review tomorrow.

daemon/src/model/cfd.rs
bonomat: I first thought so as well, but there is a slight difference.
In the protocol we have 
`Vec<(Transaction, EcdsaAdaptorSignature, interval::Digits)>,`
while here we have
`Vec<(Transaction, EcdsaAdaptorSignature, RangeInclusive<u64>)>,`

daemon/src/setup_contract.rs
```diff
@@ -140,11 +141,33 @@ pub async fn new(
...
-    let mut cet_by_digits = own_cets
+    let cets_by_payout_point = own_cets
         .into_iter()
-        .map(|(tx, _, digits)| (digits.range(), (tx, digits)))
+        .map(|pp| {
+            (
+                pp.nonce_pks,
+                pp.cets
+                    .into_iter()
+                    .map(|(tx, ecdsa_adaptor_sig, digits)| (tx, ecdsa_adaptor_sig, digits.range()))
```
bonomat: Note: this is a bug. We need to take the other party's adaptor sig.

cfd_protocol/tests/cfds.rs
```diff
@@ -116,16 +120,21 @@ fn renew_cfd() {
...
+            Payout::new(0..=10_000, Amount::from_btc(2.0).unwrap(), Amount::ZERO).unwrap(),
+            Payout::new(
+                10_001..=20_000,
+                Amount::ZERO,
+                Amount::from_btc(2.0).unwrap(),
+            )
+            .unwrap(),
+        ]
+        .concat(),
+    );
```
thomaseizinger: HashMaps support being constructed from anything that implements `IntoIterator`.
Arrays recently (1.53) got an `IntoIterator` implementation, meaning you can now initialize HashMaps without mutability from an array!

```rust
HashMap::from_iter([...])
```

daemon/src/setup_contract.rs
```diff
@@ -61,15 +61,18 @@ pub async fn new(
         )
     }
 
+    // TODO: insert payouts and oracle nonces ;)
+    let payouts_per_nonce: HashMap<&[schnorrsig::PublicKey], Vec<Payout>> = HashMap::new();
```
thomaseizinger: I think you shouldn't need the type hint if we are passing it straight on.

cfd_protocol/src/protocol.rs
```diff
@@ -2,7 +2,7 @@ pub use transaction_ext::TransactionExt;
 pub use transactions::{close_transaction, punish_transaction};
 
 use crate::interval;
-use crate::interval::Digits;
+use crate::oracle::bitcoin_hashes::_export::_core::hash::Hasher;
```
thomaseizinger: That is the wrong import, please use `std::hash::Hasher`.

cfd_protocol/src/protocol.rs
```diff
@@ -345,22 +348,48 @@ pub struct PunishParams {
...
 }
 
+#[derive(Debug, Clone)]
+pub struct CetsByEvent {
+    pub event: Event,
+    pub cets: Vec<(Transaction, EcdsaAdaptorSignature, interval::Digits)>,
+}
+
+#[derive(Debug, Clone, Eq)]
+pub struct Event {
```
thomaseizinger: Can we name this `AttestationEvent`?

cfd_protocol/src/protocol.rs
```diff
@@ -70,20 +71,19 @@ where
 /// * `maker_punish_params` - The punish parameters of the maker.
 /// * `taker` - The initial parameters of the taker.
 /// * `taker_punish_params` - The punish parameters of the taker.
-/// * `oracle_pk` - The public key of the oracle.
-/// * `nonce_pks` - One R-value per price digit signed by the oracle. Their order matches a big
-///   endian encoding of the price.
+/// * `oracle_pk` - The public key of the oracle. endian encoding of the price.
```
thomaseizinger: The 2nd part of these docs seem stale?

cfd_protocol/src/protocol.rs
```diff
@@ -70,20 +71,19 @@ where
...
 /// * `refund_timelock` - Relative timelock of the refund transaction with respect to the commit
 ///   transaction.
-/// * `payouts` - All the possible ways in which the contract can be settled, according to the
-///   conditions of the bet.
+/// * `payouts_per_nonce` - All the possible ways in which the contract can be settled, according to
+///   the conditions of the bet. The key is ane R-value per price digit signed by the oracle. Their
+///   order matches a big  endian encoding of the price.
```
thomaseizinger: Some typos:

```suggestion
///   the conditions of the bet. The key is an R-value per price digit signed by the oracle. Their
///   order matches a big endian encoding of the price.
```

cfd_protocol/tests/cfds.rs
```diff
@@ -149,21 +158,26 @@ fn renew_cfd() {
...
+            )
+            .unwrap(),
+            Payout::new(
+                10_001..=20_000,
+                Amount::from_btc(0.5).unwrap(),
+                Amount::from_btc(1.5).unwrap(),
+            )
+            .unwrap(),
+        ]
+        .concat(),
```
thomaseizinger, line 193: It was there before but I believe this `concat` call is useless?

cfd_protocol/src/protocol.rs
```diff
@@ -369,6 +398,7 @@ impl Payout {
                 digits,
                 maker_amount,
                 taker_amount,
+                event_id: event_id.to_string(),
```
thomaseizinger: Nit: It is not ideal to accept a `&str` if you are going to unconditionally call `to_string` on it because it means you are forcing an allocation even if the caller would already have an owned instance they could give you ownership of.

If you want flexibility in passing `&str` and `String`, use `impl Into<String>` as a type, that will evaluate to a no-op for `String`.

cfd_protocol/tests/cfds.rs
```diff
@@ -625,33 +680,50 @@ fn check_cfd_txs(
...
-        build_and_check_cet(
-            tx,
-            &taker_cfd_txs.cets,
-            (&maker_sk, &maker_pk),
-            &taker_pk,
-            (price, &oracle_sigs),
-            (&signed_commit_tx_maker, &commit_desc, commit_amount),
-        )
-        .expect("valid maker cet");
+    for maker_cets_by_event in maker_cfd_txs.cets.clone().into_iter() {
```
thomaseizinger: You seem to be cloning individual items as part of the for-loop as well. Just `.iter()` would be more idiomatic then.

daemon/src/setup_contract.rs
```diff
@@ -72,15 +72,17 @@ pub async fn new(
...
         sk,
     )
     .context("Failed to create CFD transactions")?;
 
     sink.send(SetupMsg::Msg1(Msg1::from(own_cfd_txs.clone())))
         .await
         .context("Failed to send Msg1")?;
-    let msg1 = stream
+
+    // TODO: remove Msg1
```
thomaseizinger: What is this TODO telling us? Is that still up to date?

cfd_protocol/tests/cfds.rs
```diff
@@ -29,32 +29,41 @@ fn create_cfd() {
...
+        nonce_pks: nonce_pks_1,
+    };
+    payouts_per_event.insert(
+        event_0.clone(),
         Payout::new(
             0..=10_000,
             Amount::from_btc(1.5).unwrap(),
             Amount::from_btc(0.5).unwrap(),
-            "id",
+            "id_0",
```
thomaseizinger: Shouldn't this line up with the one on line 39?

cfd_protocol/tests/cfds.rs
```diff
@@ -29,32 +29,41 @@ fn create_cfd() {
...
+    let nonce_pks_0 = announcement_0.nonce_pks();
 
-    let payouts = vec![
+    let event_0 = Event {
+        id: "id".to_string(),
+        nonce_pks: nonce_pks_0,
+    };
+    let nonce_pks_1 = announcement_1.nonce_pks();
+    let event_1 = Event {
+        id: "id".to_string(),
```
thomaseizinger: Why are these IDs not different?

thomaseizinger (commented):
Thanks! This is a good changeset but I am not sure if we should get it in today, I don't think it is that critical.

Couple of questions though. Feel free to not address any of the nits but please squash your commits before merging :)

cfd_protocol/src/protocol.rs
bonomat: I thought we need to allocate anyway because we have an iterator over digits.

daemon/src/setup_contract.rs
bonomat: Fixed. The typehint was not necessary.

cfd_protocol/tests/cfds.rs
bonomat: It's needed because otherwise we endup with a `Vec<Vec<Payouts>>`.

cfd_protocol/src/protocol.rs
thomaseizinger: Perhaps but that iterator has nothing to do with this string. Allocations are not batched together when they concern different variables.

cfd_protocol/tests/cfds.rs
bonomat: Good catch 🤦‍♂️

Because of this we ended up not having any CETs in this test.

cfd_protocol/tests/cfds.rs
bonomat: I failed at this exercise 🙈

cfd_protocol/tests/cfds.rs
bonomat: But then I would need more clones further down.

cfd_protocol/tests/cfds.rs
thomaseizinger: Ah right, thanks I didn't realize that!

cfd_protocol/tests/cfds.rs
luckysori: Should be done now :D

luckysori:
@bonomat have a look if you want to make sure that I haven't undone something crucial. I should have pushed fixup commits after rebasing, but I got carried away during the rebase and started making changes, so it may be hard to review the differences. Sorry.

daemon/src/setup_contract.rs
```diff
@@ -168,6 +183,37 @@ pub async fn new(
...
+                .map(|(tx, _, digits)| {
+                    let other_encsig = other_cets
+                        .iter()
+                        .find_map(|(other_range, other_encsig)| {
+                            (other_range == &digits.range()).then(|| other_encsig)
+                        })
+                        .with_context(|| {
+                            format!(
+                                "Missing counterparty adaptor signature for CET corresponding to
+                                 price range {:?}",
```
bonomat, line 206: I think there shouldn't be a line break here. 
```suggestion
                                "Missing counterparty adaptor signature for CET corresponding to price range {:?}",
```

bonomat (commented):
Imho it looks good. Thanks @luckysori

cfd_protocol/src/protocol.rs
thomaseizinger: This is still stale I believe.

cfd_protocol/tests/cfds.rs
```diff
@@ -112,20 +132,23 @@ fn renew_cfd() {
...
+    let payouts_per_event = HashMap::from_iter([(
+        event,
+        vec![
+            Payout::new(0..=10_000, Amount::from_btc(2.0).unwrap(), Amount::ZERO).unwrap(),
+            Payout::new(
+                10_001..=50_000,
+                Amount::ZERO,
+                Amount::from_btc(2.0).unwrap(),
+            )
+            .unwrap(),
```
thomaseizinger: I find it a bit odd to return `Vec<Self>` from a `new` function but it will do for now :)

cfd_protocol/tests/cfds.rs
```diff
@@ -930,81 +1017,171 @@ fn make_keypair(rng: &mut (impl RngCore + CryptoRng)) -> (SecretKey, PublicKey)
...
+    }
+
+    fn announcement(&self) -> Announcement {
+        Announcement {
+            id: self.id.clone(),
+            nonce_pks: self.nonce_pks.clone(),
         }
     }
+
+    const OLIVIA_PK: &'static str =
```
thomaseizinger, line 1082: Clippy doesn't complain that const references imply `'static`?

thomaseizinger (approved):
Thanks!

I pushed a follow-up commit, feel free to drop it if you don't like it :)

cfd_protocol/src/protocol.rs
luckysori: Oh, I didn't notice the second sentence.

cfd_protocol/tests/cfds.rs
luckysori: We can rename it if it helps, but I think the behaviour of generating multiple `Payout`s per interval is still desirable/necessary.

cfd_protocol/tests/cfds.rs
luckysori: I think it complained about it being absent, which surprised me. I'll check again.

luckysori: Compiler asks for it. Perhaps it has to do with the constant being defined within an `impl` block.

thomaseizinger (approved):

PR #163, opened 2021-09-28
Layout UI with flexes

da-kami:
Was in the mood for a bit late night coding.

The Grid was restrictive and caused problems overlapping things.
We can achieve better layouting with flex components (for now mostly HStack and VStack flexes).
We can eventually also add space bounds to the flexes for more layourting (i.e. graph should take 70% on right...), but for now I kept it rather simple.

I experimented with Grids for the tiles where we group information (i.e. Buy, Order) but it did not look good, so I amd using flexes with fixed label width for better looks.

Before:

image

After:

image

image

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -1,9 +1,8 @@
 import {
+    Box,
     Button,
     Container,
-    Flex,
-    Grid,
-    GridItem,
+    Divider,
     HStack,
     Tab,
     TabList,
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Can you add a screenshot of what this looks like when there is less space, like the browser is half or 1/3 of your screen?

thomaseizinger:
I think on the whole, this sell order tile is misplaced anyway. What if we integrate this into the table somehow?

Putting things next to each other never scales nicely in my experience.

thomaseizinger:
From what I know (could very much be wrong), grids are meant to be superior to flex boxes, at least when it comes to the general layout of a page, which is why I transitioned it to a grid in the first place.

I don't care if we go back to flexboxes just wanted to mentioned that the decision wasn't arbtirary :)

Most design systems are oriented around a 12 column grid for example, with bootstrap being among the most famous ones.

frontend/src/MakerApp.tsx
```diff
@@ -74,67 +73,70 @@ export default function App() {
         !runningStates.includes(value.state) && !closedStates.includes(value.state) && !openStates.includes(value.state)
     );
 
+    const labelWidth = 110;
```
thomaseizinger, line 76: I _think_ grids would solve this problem because the column would size itself to the longest row or at least you wouldn't need to set the length at every row.

da-kami: as discussed: I could not easily figure out how to create a grid with a fixed sized column. It was always a relation between the columns (e.g. 1:3), which is not what I want I'm fairly sure it *can* be done, but I would not invest too much time into it atm.

da-kami:
as discussed: merging this for now, because it is an improvement to what was there before (and the flexes should not cause any pain). we can do another iteration in the next few days and try more things.

PR #164, opened 2021-09-28
Remove the span that does not align the table header properly

da-kami:
When sorting the triangle appears next to the header text - that is expected and standard behavior.
The span was always rendered (also with no sorting), and message up the alignment of the text.

Before:

image

After:

image

when sorting (triangle appears for a few secs then disappears again):

image


thomaseizinger (approved):
Nice fix!

PR #169, opened 2021-09-30
Restore first version of price feed integration

klochowicz:
Calculating profit for the needs of UI inside `ToSseEvent` turned out to be much simpler and easier to build upon.


daemon/src/routes_maker.rs
```diff
@@ -39,11 +38,11 @@ pub async fn maker_feed(
         let order = rx_order.borrow().clone();
         yield order.to_sse_event();
 
-        let cfds = rx_cfds.borrow().clone();
-        yield cfds.to_sse_event();
         let quote = rx_quote.borrow().clone();
         yield quote.to_sse_event();
 
+        let cfds_with_price = CfdsWithCurrentPrice{cfds: rx_cfds.borrow().clone(), current_price: quote.for_maker()};
```
thomaseizinger: This formatting is a bit off but not being picked up by dprint due to the macro, can you format it manually please?

thomaseizinger (approved):

daemon/src/taker_cfd.rs
```diff
@@ -62,16 +59,14 @@ impl Actor {
...
         order_feed_actor_inbox: watch::Sender<Option<Order>>,
         send_to_maker: Address<send_to_socket::Actor<wire::TakerToMaker>>,
         monitor_actor: Address<monitor::Actor<Actor>>,
         cfds: Vec<Cfd>,
     ) -> Result<Self> {
         let mut conn = db.acquire().await?;
-
-        // populate the CFD feed with existing CFDs
-        cfd_feed.update(&mut conn).await?;
+        cfd_feed_actor_inbox.send(load_all_cfds(&mut conn).await?)?;
```
da-kami: Given, that after the restart changes we now pass in the initial list of cfds you don't have to lead them from the db here, but could just clone them into the feed. (I don't mind much though, I think loading is not expensive)

daemon/src/maker_cfd.rs
```diff
@@ -81,7 +78,7 @@ impl Actor {
         let mut conn = db.acquire().await?;
 
         // populate the CFD feed with existing CFDs
-        cfd_feed.update(&mut conn).await?;
+        cfd_feed_actor_inbox.send(load_all_cfds(&mut conn).await?)?;
```
da-kami: same here, since restart we have a list of `cfds` in here, and the constructor should not have side effects in changing cfd state, so we could just use the list we get and send it out.

da-kami (approved):
I like this simplified version better. 
I'd role with this for now and only make it more complex if needed.

daemon/src/routes_maker.rs
klochowicz: thanks for spotting! I sometimes still forget about these macro problems. fixed in all occurrences & pushed a fixup commit.

PR #172, opened 2021-09-30
Directly initialize feed with list of current CFDs

thomaseizinger:


daemon/src/taker_cfd.rs
```diff
@@ -65,9 +65,6 @@ impl Actor {
         monitor_actor: Address<monitor::Actor<Actor>>,
         cfds: Vec<Cfd>,
     ) -> Result<Self> {
-        // populate the CFD feed with existing CFDs
-        cfd_feed_actor_inbox.send(cfds.clone())?;
-
         for dlc in cfds.iter().filter_map(|cfd| Cfd::pending_open_dlc(cfd)) {
```
klochowicz, line 76: have you considered moving these broadcasting transations out of the actors too? seems like generic code that could have been done before the actor gets initialised (it only needs the wallet, which is available outside too). Then we could get rid of passing the `Vec<Cfd>` inside the actor, which feels like a kludge.

klochowicz (approved):
nice one!

daemon/src/taker_cfd.rs
thomaseizinger: @da-kami what do you think about that? I actually like the idea of having all the setup code visible outside of the actor instead of inside.

PR #173, opened 2021-09-30
Clean up message handlers in maker_cfd actor

klochowicz:
No need to pass the message structs into the handlers as they are mostly
superfluous boilerplate and are destructured afterwards anyway.


thomaseizinger (approved):
Nice, thank you!

PR #174, opened 2021-09-30
Clear UI model separation and dynamic actions

da-kami:


da-kami:
will open once properly rebased

daemon/src/to_sse_event.rs
```diff
@@ -26,9 +27,26 @@ pub struct Cfd {
...
+pub enum CfdAction {
+    Accept,
+    Reject,
+}
+
+impl<'v> FromParam<'v> for CfdAction {
+    type Error = serde_plain::Error;
+
+    fn from_param(param: &'v str) -> Result<Self, Self::Error> {
+        let action = serde_plain::from_str(param)?;
```
thomaseizinger, line 45: Nice, didn't know about `serde_plain`!

thomaseizinger (commented):
Coming along well! Thanks for introducing this!

The commit history needs a bit of a cleanup but otherwise this looks really promising!

da-kami:
> Coming along well! Thanks for introducing this!
> 
> The commit history needs a bit of a cleanup but otherwise this looks really promising!

In that case I should have just squashed upon rebase... 
I tried to keep the histroy at first, but there were too many moving parts. Had to fix several things that actually did not work at the end - just did not spot them before.

It's not such a big change in lines, should be fine as one commit :)

daemon/src/to_sse_event.rs
da-kami: :)

daemon/src/routes_maker.rs
```diff
@@ -127,40 +127,33 @@ pub struct PromptAuthentication {
 
 /// The maker POSTs this to accept an order
 #[derive(Debug, Clone, Deserialize)]
-pub struct AcceptOrRejectOrderRequest {
+pub struct AcceptOrRejectTakeRequest {
```
thomaseizinger: Is this still used?

frontend/src/components/Types.tsx
```diff
@@ -46,3 +184,23 @@ export interface PriceInfo {
 export function unixTimestampToDate(unixTimestamp: number): Date {
     return new Date(unixTimestamp * 1000);
 }
+
+export function intoCfd(key: string, value: any): any {
```
thomaseizinger, line 180: If you want to be fancy, you could add a unit test for using these together with `JSON.parse`.

frontend/src/components/cfdtables/CfdTable.tsx
```diff
@@ -104,18 +141,28 @@ export function CfdTable(
...
+            {
+                Header: "Action",
+                accessor: ({ actions, order_id }) => {
+                    const actionIcons = actions.map((action) => {
+                        return (<IconButton
+                            key={action}
+                            colorScheme={colorSchemaForAction(action)}
+                            aria-label={action}
+                            icon={iconForAction(action)}
+                            onClick={async () => postAction(order_id, action)}
```
thomaseizinger: Unnecessary `async`.

(Do we not have the lint enabled that protects against this?)

daemon/src/maker_cfd.rs
```diff
@@ -19,11 +19,11 @@ use tokio::sync::watch;
 use xtra::prelude::*;
 use xtra::KeepRunning;
 
-pub struct AcceptOrder {
+pub struct Accept {
```
thomaseizinger: Not sure I agree with this. We will have another accept / reject pair for collab. settlement so I think we will need to differentiate here.

thomaseizinger (approved):
Nicely done, thanks for tackling this!

frontend/src/MakerApp.tsx
```diff
@@ -149,24 +149,24 @@ export default function App() {
 
             <Tabs marginTop={5}>
                 <TabList>
-                    <Tab>Running [{running.length}]</Tab>
                     <Tab>Open [{open.length}]</Tab>
+                    <Tab>Accept / Reject [{acceptOrReject.length}]</Tab>
```
bonomat, line 143: I was first a bit confused and thought this would be a `Accepted`/`Rejected` category. I can't come up with a better term though.

frontend/src/MakerApp.tsx
```diff
@@ -66,16 +72,10 @@ export default function App() {
...
-    const openStates = ["Requested"];
-    const open = cfds.filter((value) => openStates.includes(value.state));
-    const closedStates = ["Rejected", "Closed"];
-    const closed = cfds.filter((value) => closedStates.includes(value.state));
-    // TODO: remove this. It just helps to detect immediately if we missed a state.
-    const unsorted = cfds.filter((value) =>
-        !runningStates.includes(value.state) && !closedStates.includes(value.state) && !openStates.includes(value.state)
-    );
+    const acceptOrReject = cfds.filter((value) => value.state.getGroup() === StateGroupKey.ACCEPT_OR_REJECT);
+    const opening = cfds.filter((value) => value.state.getGroup() === StateGroupKey.OPENING);
```
bonomat, line 66: I'm not sold on having an extra state/tab for `opening`. Imho the user does not care about the difference between `opening` and `open`.

bonomat (approved):
Nice. I'm not convinced on the sates but happy to leave that for now.

frontend/src/components/cfdtables/CfdTableMaker.tsx
```diff
@@ -1,236 +0,0 @@
-import { CheckIcon, ChevronRightIcon, ChevronUpIcon, CloseIcon } from "@chakra-ui/icons";
```
bonomat, line 1: The reason why I introduced a separate file (`CfdTableMaker`) for the creating a table for the maker was because the table to accept/reject orders should contain different information than the other tables.

Now (with this PR) the maker sees values like `Unrealized P/L` when he does not care about it as it will always be 0 at this time. Also, he sees `margin` which is the taker's margin but does not know about his own values (because we don't have these values in our model yet).

What I want to say is: imho these tables cannot be merged.

frontend/src/MakerApp.tsx
da-kami: Yeah, I am not so sure about it neither - I think there are good arguments for and against it. 
I think we will have to see what actual users say :)

frontend/src/components/cfdtables/CfdTableMaker.tsx
da-kami: I am not sure I can follow your arguments, let's have a closer look at the values needed so the maker can take a decision if he should accept / reject (because that's what we need).

> Unrealized P/L

Yes, this one is unnecessarily in the accept/reject table now. But, it was actually part of the maker table and if it simplifies things I would not care to much about that.

> sees margin which is the taker's margin

No, each role sees their margin correctly. The margin currently always represents the own margin - we can also send the counterparty margin to the UI if needed. Should the maker also see the one of the taker? Is that relevant for decision making? (I think not...)

---

Things that were additionally displayed in the maker's table were:

* OrderId
* Leverage
* Timestamp

Does that maker need that for decision making? 
Given that we are currently always at a fixed leverage, I think we can introduce this table at a later point. 
Are the timestamp and the order id necessary for decision making?

---

Sidnote: The taker takes by offer-id + quantity. So the taker cannot game the maker by sending e.g. wrong leverage (...). The maker will always just use the order values and that's it.  This model was introduced to keep validation simple...

One thing that we could change is, that we remove the current order (offer) from the maker once a take request came in. The maker still wants to see it! I think we should remove it "publicly" only - i.e. introduce something like a "public" badge for the current offer (or similar). That way the maker can see the offer values when dealing with take requests.

frontend/src/MakerApp.tsx
da-kami: The maker has to make a decision there, so it should be easy for the maker to see those. That's why I opted for it. But yeah, it is a bit weird. Would change this once we have used it and find it bad.

daemon/src/maker_cfd.rs
da-kami: I see, rename to `AcceptTakeRequest` / `RejectTakeRequest` ? 
I just find the order confusing here - the action is on the cfd level now, and both are actually in the context of a cfd already.
Don't have a strong opinion though, can also just remove the commit.

thomaseizinger: Didn't we settle on the terminology that the taker makes an order with the maker? From that perspective, the maker IS accepting / rejecting an order.

frontend/src/components/cfdtables/CfdTableMaker.tsx
bonomat: The `Unrealized P/L` was hidden (same for orderId, leverage and timestamp). Look further down below on line 190.

> No, each role sees their margin correctly. [..]

I wasn't aware of it. The maker definitely cares about both margins. Especially in a case where the maker wants to open a hedged position. Imho these values are crucial: 

- opening price
- taker's leverage and margins and liquidation price: to be able to make an informed decision if this position is too risky and to be able to open a hedged position
- maker's margin: to know how much the maker has to put up. 

-----

I'm fine with removing this if you think it simplifies things. Imho it's easier to have two different tables which can evolve separately.

thomaseizinger: Removing this duplication simplifies code a lot and we did say initially that the maker's UI is not as important so I am happy to accept that certain values are there unnecessarily. If it just comes to filtering data, we can also easily build that on top of a single table component.

daemon/src/maker_cfd.rs
klochowicz: +1, the terminology does not need to change now, especially that as @thomaseizinger mentioned, i'madding another pair of accept/reject today :)

klochowicz (approved):
nice, looking forward to rebasing my collab work on top of this!

frontend/src/components/cfdtables/CfdTableMaker.tsx
da-kami: I understand @bonomat's agruments, and I think we will need the second table eventually. I would still merge it with one for now, and eventually just duplicate that one and adapt as needed, once it is used. Since this table is fully functional it should be easy to copy and adapt as needed.

daemon/src/maker_cfd.rs
da-kami: I'll just revert the commit, don't have strong feelings about it.

PR #175, opened 2021-09-30
Add a taker endpoint for initiating collaborative settlement

klochowicz:
For now, only logs the proposal in the logs on the maker side

I've copied getting order id via `/<id>/` in the endpoint from #174

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -10,7 +10,7 @@ use crate::model::Usd;
 use crate::monitor::{self, MonitorParams};
 use crate::wallet::Wallet;
 use crate::wire::SetupMsg;
-use crate::{bitmex_price_feed, send_to_socket, setup_contract, wire};
+use crate::{send_to_socket, setup_contract, wire};
 use anyhow::{Context as _, Result};
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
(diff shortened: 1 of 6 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -26,6 +26,11 @@ pub struct TakeOffer {
     pub quantity: Usd,
 }
 
+pub struct ProposeSettlement {
+    pub order_id: OrderId,
+    pub quote: bitmex_price_feed::Quote,
```
thomaseizinger: Would prefer to just have the price here rather than the entire quote.

daemon/src/wire.rs
```diff
@@ -20,13 +20,15 @@ use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
 #[allow(clippy::large_enum_variant)]
 pub enum TakerToMaker {
     TakeOrder { order_id: OrderId, quantity: Usd },
+    ProposeSettlement(SettlementProposal),
```
thomaseizinger: Instead of embedding this, can we just have fields in the enum? Would like to not couple the domain model too much to wire message format!

thomaseizinger (commented):
Looks good, some minor nits :)

thomaseizinger (approved):
Nice!

Squash and merge!

PR #176, opened 2021-09-30
Fix electrum parsing error and refactor its detection

thomaseizinger:
We currently cannot boot up the daemons because we actually cannot successfully detect that the transaction is already in the blockchain.


da-kami (approved):
Sorry for copy pasting and introducing this one 😬

klochowicz (approved):
clean code; and with a test-case covering the bugfix. good one!

PR #177, opened 2021-09-30
Fix parsing of electrum error response

thomaseizinger:
This is actually a security bug as it would allow remote triggers
of panics!


klochowicz (approved):
have you been fuzzing binary or just randomly ran into this issue? ;D 
luckily we found it now.

thomaseizinger:
> have you been fuzzing binary or just randomly ran into this issue? ;D 
> luckily we found it now.

Ran into it due to a insufficient fund error :)

PR #178, opened 2021-09-30
Rework monitoring actor to use local state instead of tasks

thomaseizinger:
This should greatly improve resilience as we no longer have any
tasks within `tokio` that can fail without recovering.

Instead, we periodically ping the actor with a `Sync` message which
updates the local state of all scripts and sends out messages accordingly.


thomaseizinger:
Will rebase on top of https://github.com/comit-network/hermes/pull/145 once merged.

daemon/src/monitor.rs
```diff
@@ -23,41 +29,46 @@ pub struct MonitorParams {
...
-                .map(|(tx, _, range)| (tx.txid(), script_pubkey.clone(), range))
-                .collect(),
-            refund: (
-                dlc.refund.0.txid(),
-                script_pubkey,
-                refund_timelock_in_blocks,
-            ),
-        }
-    }
+pub struct Sync;
```
da-kami, line 32: ❓ So this message is used to trigger syncing from the outside and configured when instantiating the actor?

(just validating I got it correctly 😬)

da-kami (commented):
Looking good, but needs are rebase on latest master - can't test

The log messages changed, I don't see the transaction monitoring being picked up for existing Cfds Open anymore, but I suppose that is expected.

daemon/src/monitor.rs
da-kami: read the f***ing PR description Daniel!

daemon/src/monitor.rs
```diff
@@ -279,30 +224,200 @@ where
...
+        // 4. check for finished monitoring tasks
+        for ((txid, script), status) in self.current_status.iter() {
+            match self.awaiting_status.entry((*txid, script.clone())) {
+                Entry::Vacant(_) => {
+                    unreachable!("we are only fetching the status of scripts we are waiting for")
+                }
+                Entry::Occupied(mut occupied) => {
+                    let targets = occupied.insert(Vec::new());
+
+                    // Split vec into two lists, all the once for which we reached the target and
```
da-kami:
```suggestion
                    // Split vec into two lists, all the ones for which we reached the target and
```

daemon/src/monitor.rs
```diff
@@ -279,30 +224,200 @@ where
...
+                    // the once which we need to still monitor
```
da-kami:
```suggestion
                    // the ones which we need to still monitor
```

daemon/src/monitor.rs
```diff
@@ -279,30 +224,200 @@ where
...
+
+impl Confirmed {
+    fn with_confirmations(blocks: u32) -> Self {
+        Self { depth: blocks - 1 }
+    }
+
+    /// Compute the depth of a transaction based on its inclusion height and the
+    /// latest known block.
+    ///
+    /// Our information about the latest block might be outdated. To avoid an
```
da-kami, line 359: thanks for the comment, I started thinking about the saturating sub again...

daemon/src/monitor.rs
```diff
@@ -525,3 +529,275 @@ where
...
+                            height: *included_at,
+                            tx_hash: *tx,
+                            fee: None,
+                        }]
+                    })
+                    .collect())
+            }
+
+            fn batch_call(&self, _batch: &Batch) -> Result<Vec<serde_json::Value>, Error> {
+                todo!()
```
da-kami: nit: maybe use `unreachable` instead? I always feel `todo` indicates something that should be actioned at some point

da-kami (approved):
I gave it a more thorough review and did not spot any issues. However, I think only using this for some time will reveal bugs. 
The test is cool, interesting concepts in there.

daemon/src/monitor.rs
thomaseizinger: This was just autofilled by rust-analyzer but yeah, I can change it to unreachable!

daemon/src/monitor.rs
thomaseizinger: This comment is a copy paste from the old code lol :D

PR #179, opened 2021-09-30
Non collaborative close

### LINKED ISSUE #73: Epic: Non-collaborative settlement
All tasks around providing the backup path of non-collaborative settlement. The oracle plays a key role in this.

- [x] https://github.com/comit-network/hermes/issues/66
- [x] https://github.com/comit-network/hermes/issues/68
- [x] https://github.com/comit-network/hermes/issues/69
- [x] https://github.com/comit-network/hermes/issues/70
- [x] https://github.com/comit-network/hermes/issues/67
- [x] https://github.com/comit-network/hermes/issues/110
- [x] https://github.com/comit-network/hermes/issues/111
- [x] https://github.com/comit-network/hermes/issues/170
- [x] https://github.com/comit-network/hermes/issues/193

da-kami:
fixes #73 

Looks can be improved, but should be enough to test if we get till refund.

Ideas that can be done on top of this work:
- [should] add tooltip or similar to all the actions so the user gets more info. 
- [could] the user if he/she know what they do when pressing the commit button. 
- [could] move the commit action button into the dropdown and not show it in actions. Not sure about that.

(would fill them in tickets once this goes in)

Maker screenshot:

image

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -97,6 +97,13 @@ impl Actor {
             tracing::info!("Refund transaction published on chain: {}", txid);
         }
 
+        for cfd in cfds.iter().filter(|cfd| Cfd::is_pending_commit(cfd)) {
+            let signed_commit_tx = cfd.commit_tx()?;
+            let txid = wallet.try_broadcast_transaction(signed_commit_tx).await?;
+
+            tracing::info!("Commit transaction published on chain: {}", txid);
+        }
+
         Ok(Self {
             db,
             wallet,
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -428,9 +428,9 @@ impl Actor {
         Ok(())
     }
 
-    async fn handle_commit(&mut self, msg: Commit) -> Result<()> {
+    async fn handle_commit(&mut self, order_id: OrderId) -> Result<()> {
         let mut conn = self.db.acquire().await?;
-        let cfd = load_cfd_by_order_id(msg.order_id, &mut conn).await?;
+        let cfd = load_cfd_by_order_id(order_id, &mut conn).await?;
 
         let signed_commit_tx = cfd.commit_tx()?;
 
@@ -512,7 +512,7 @@ impl Handler<RejectOrder> for Actor {
 #[async_trait]
 impl Handler<Commit> for Actor {
     async fn handle(&mut self, msg: Commit, _ctx: &mut Context<Self>) {
-        log_error!(self.handle_commit(msg))
+        log_error!(self.handle_commit(msg.order_id))
     }
 }
 
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -5,7 +5,8 @@ use crate::db::{
 };
 use crate::maker_inc_connections::TakerCommand;
 use crate::model::cfd::{
-    Cfd, CfdState, CfdStateChangeEvent, CfdStateCommon, Dlc, Order, OrderId, SettlementProposal,
+    Cfd, CfdState, CfdStateChangeEvent, CfdStateCommon, Dlc, Order, OrderId, Role,
+    SettlementProposal,
 };
 use crate::model::{TakerId, Usd};
 use crate::monitor::MonitorParams;
@@ -358,7 +359,7 @@ impl Actor {
             (self.oracle_pk, nonce_pks),
             cfd,
             self.wallet.clone(),
-            setup_contract::Role::Maker,
+            Role::Maker,
         );
 
         let this = ctx
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -29,6 +29,10 @@ pub struct RejectOrder {
     pub order_id: OrderId,
 }
 
+pub struct Commit {
+    pub order_id: OrderId,
+}
+
 pub struct NewOrder(pub Order);
 
 pub struct NewTakerOnline {
@@ -417,6 +421,25 @@ impl Actor {
         Ok(())
     }
 
+    async fn handle_commit(&mut self, msg: Commit) -> Result<()> {
+        let mut conn = self.db.acquire().await?;
+        let cfd = load_cfd_by_order_id(msg.order_id, &mut conn).await?;
+
+        let signed_commit_tx = cfd.commit_tx()?;
+
+        let txid = self
+            .wallet
+            .try_broadcast_transaction(signed_commit_tx)
+            .await?;
+
+        tracing::info!("Commit transaction published on chain: {}", txid);
+
+        let new_state = cfd.handle(CfdStateChangeEvent::CommitTxSent)?;
+        insert_new_cfd_state_by_order_id(cfd.order.id, new_state, &mut conn).await?;
+
+        Ok(())
+    }
+
     async fn handle_monitoring_event(&mut self, event: monitor::Event) -> Result<()> {
         let order_id = event.order_id();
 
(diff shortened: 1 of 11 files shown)


daemon/src/routes_taker.rs
```diff
@@ -96,6 +96,27 @@ pub async fn post_order_request(
...
 
+#[rocket::post("/cfd/<id>/<action>")]
+pub async fn post_cfd_action(
+    id: OrderId,
+    action: CfdAction,
+    cfd_actor_address: &State<Address<taker_cfd::Actor>>,
+) -> Result<status::Accepted<()>, status::BadRequest<String>> {
+    match action {
+        CfdAction::Accept | CfdAction::Reject => {
+            unreachable!("The taker does not accept and reject");
```
thomaseizinger: Should just be bad request I think, not a panic.

daemon/src/model/cfd.rs
```diff
@@ -405,41 +405,32 @@ impl Cfd {
...
-                    OutgoingOrderRequest { .. } => unreachable!("taker-only state"),
-                    IncomingOrderRequest { .. } | Accepted { .. } | ContractSetup { .. } => {
-                        bail!("Did not expect lock finality yet: ignoring")
-                    }
-                    Open { .. } | OpenCommitted { .. } | MustRefund { .. } => {
-                        bail!("State already assumes lock finality: ignoring")
-                    }
-                    Rejected { .. } | Refunded { .. } | SetupFailed { .. } => {
-                        bail!("The cfd is already in final state {}", self.state)
+                monitor::Event::LockFinality(_) => {
```
thomaseizinger, line 462: Could potentially collapse the two `match` patterns into one.

daemon/src/model/cfd.rs
```diff
@@ -549,6 +579,43 @@ impl Cfd {
         Ok(signed_refund_tx)
     }
 
+    pub fn commit_tx(&self) -> Result<Transaction> {
+        let dlc = if let CfdState::Open { dlc, .. } | CfdState::PendingOpen { dlc, .. } =
```
thomaseizinger, line 627: Could have another accessor or that returns `Option<Dlc>`, then you could just use `context` to bail out of the case where it is `None`.

daemon/src/routes_taker.rs
```diff
@@ -96,6 +96,27 @@ pub async fn post_order_request(
...
+        }
+        CfdAction::Commit => {
+            cfd_actor_address
+                .do_send_async(taker_cfd::Commit { order_id: id })
+                .await
+                .expect("actor to always be available");
```
thomaseizinger, line 125: We should actually also not panic here but return a 500. No network API ever should make it possible to panic an internal process as that can be a security issue.

daemon/src/taker_cfd.rs
```diff
@@ -85,6 +85,13 @@ impl Actor {
             tracing::info!("Refund transaction published on chain: {}", txid);
         }
 
+        for cfd in cfds.iter().filter(|cfd| Cfd::is_pending_commit(cfd)) {
+            let signed_commit_tx = cfd.commit_tx()?;
+            let txid = wallet.try_broadcast_transaction(signed_commit_tx).await?;
+
+            tracing::info!("Commit transaction published on chain: {}", txid);
```
thomaseizinger, line 97: Should these be moved into the `Wallet` perhaps?

thomaseizinger (approved):
Thank you!

Some minor comments bother otherwise good to go in!

daemon/src/maker_cfd.rs
```diff
@@ -19,11 +19,15 @@ use tokio::sync::watch;
...
 use xtra::KeepRunning;
 
-pub struct AcceptOrder {
+pub struct Accept {
     pub order_id: OrderId,
 }
 
-pub struct RejectOrder {
+pub struct Reject {
+    pub order_id: OrderId,
```
klochowicz, line 30: in the same line as commented on other review, I'll be adding another pair of Accept/Reject structs, so it would be easier if this stayed the same.

daemon/src/taker_cfd.rs
```diff
@@ -301,6 +312,26 @@ impl Actor {
 
         Ok(())
     }
+
+    // TODO: Duplicated with maker
```
klochowicz, line 356: if you'd like to reduce duplication, you could extract a function that would look like this when invoked:

`handle_commit(&mut conn, self.wallet, cfd).await?`

daemon/src/taker_cfd.rs
```diff
@@ -301,6 +312,26 @@ impl Actor {
...
+    async fn handle_commit(&mut self, msg: Commit) -> Result<()> {
```
klochowicz: in the `handle_***` functions, you don't need to pass the msg, but just `order_id` in this case (like in other handlers)

klochowicz (approved):
just a bunch of nits :) nice one!

daemon/src/routes_taker.rs
da-kami: I agree, should fix this in the other API calls at some point, we have several expects there. I only fixed it in the ones concerned by this PR for now, to not cause potential unwanted side effects.

daemon/src/maker_cfd.rs
da-kami: thanks, missed dropping that one here

daemon/src/routes_taker.rs
```diff
@@ -104,13 +104,14 @@ pub async fn post_cfd_action(
...
-            unreachable!("The taker does not accept and reject");
+            return Err(status::BadRequest(None));
         }
+
         CfdAction::Commit => {
             cfd_actor_address
                 .do_send_async(taker_cfd::Commit { order_id: id })
                 .await
-                .expect("actor to always be available");
+                .map_err(|e| status::BadRequest(Some(e.to_string())))?;
```
thomaseizinger, line 115: Nit: I don't think we should leak this error, I will just say "Disconnected".

thomaseizinger (approved):
Thanks!

I think in this case, `git absorb` would have done a good job here.

da-kami:
Taker:

image

Maker:

image

daemon/src/model/cfd.rs
```diff
@@ -39,12 +39,29 @@ impl<'v> FromParam<'v> for OrderId {
...
 }
 
+/// Role in the Cfd
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq)]
+pub enum Role {
+    Maker,
+    Taker,
+}
+
+impl From<Origin> for Role {
```
klochowicz, line 57: Perhaps add a comment that this assumption only holds until we only allow the maker to be the only one creating the offers?

daemon/src/to_sse_event.rs
```diff
@@ -221,12 +222,15 @@ fn into_unix_secs(time: SystemTime) -> u64 {
         .as_secs()
 }
 
-fn actions_for_state(state: model::cfd::CfdState) -> Vec<CfdAction> {
-    match state {
-        model::cfd::CfdState::IncomingOrderRequest { .. } => {
+fn actions_for_state(state: model::cfd::CfdState, role: Role) -> Vec<CfdAction> {
+    match (state, role) {
```
klochowicz, line 226: nice! I'll use that really soon for accepting/rejecting the settlement.

klochowicz (approved):
coolness! thanks for reworking the taker settlement into the action :)

daemon/src/model/cfd.rs
da-kami: I think it always holds, because if you are in the role that created the offer, then you are always the maker. 

Sidenote: I think the order is not actually and order. We will have to refactor that eventually. We might have an order at some point, but I don't think it will represent what it currently is.

da-kami:
I ammended a minor cleanup, will merge once CI is done :)

PR #181, opened 2021-10-01
Add dummy payouts

thomaseizinger:


daemon/src/payout_curve.rs
```diff
@@ -1,13 +1,30 @@
...
-    _price: Usd,
+    price: Usd,
     _quantity: Usd,
-    _maker_payin: bitcoin::Amount,
-    (_taker_payin, _leverage): (bitcoin::Amount, Leverage),
+    maker_payin: bitcoin::Amount,
+    (taker_payin, _leverage): (bitcoin::Amount, Leverage),
 ) -> Result<Vec<Payout>> {
-    Ok(vec![])
+    let dollars = price.try_into_u64()?;
```
klochowicz, line 13: perhaps add a `TODO` that this should be substituted by something better than a dummy payout in due time?

klochowicz (commented):
Can't really approve as I have no knowledge how a valid vector of payouts should look like...

da-kami (approved):

daemon/src/payout_curve.rs
```diff
@@ -1,13 +1,30 @@
...
+    let payouts = vec![
+        Payout::new(
+            0..=(dollars - 10),
+            maker_payin + taker_payin,
+            bitcoin::Amount::ZERO,
+        )?,
+        Payout::new((dollars - 10)..=(dollars + 10), maker_payin, taker_payin)?,
```
luckysori, line 20: :wrench: It may not matter for the dummy payouts, but they ranges are inclusive. Here there would be slight overlaps between some CETs, which is bad.

daemon/src/payout_curve.rs
```diff
@@ -1,13 +1,30 @@
...
     _quantity: Usd,
-    _maker_payin: bitcoin::Amount,
-    (_taker_payin, _leverage): (bitcoin::Amount, Leverage),
+    maker_payin: bitcoin::Amount,
+    (taker_payin, _leverage): (bitcoin::Amount, Leverage),
 ) -> Result<Vec<Payout>> {
-    Ok(vec![])
+    let dollars = price.try_into_u64()?;
+    let payouts = vec![
+        Payout::new(
```
luckysori, line 15: :mega: This API has changed, but you're probably not getting any compiler errors because this file is not included in any module.

daemon/src/payout_curve.rs
thomaseizinger: A right, didn't realize this.

PR #190, opened 2021-10-04
Calculate P/L

### LINKED ISSUE #33: Calculate profit in daemon and display in UI
The CFD profit is dependent on the current price (calculated). We should take the current price into account for the CFD feed.

- upon CFD feed update re-calculate the profit every time
- upon price update trigger a CFD feed update

This requires some kind of price feed (from the Oracle or a component that publishes a reliable price related to the Oracle's price).

bonomat:
Resolves #33

For calculating the Profit/Loss in BTC and in Percent

image

Note: Profit in USD does not make sense. Think about this scenario. 
- opening price: $50k
- closing price: $25k
- margin provided: 1BTC

The trader loses everything at $25k, i.e. he lost 1BTC. 1BTC at the closing price is worth $25k while at the opening it was $50k. Did he lose $25k or $50k? Neither, he lost 1BTC.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1,17 +1,18 @@
-use crate::model::{Leverage, Position, TakerId, TradingPair, Usd};
+use crate::model::{Leverage, Percent, Position, TakerId, TradingPair, Usd};
 use crate::monitor;
-use anyhow::{bail, Result};
+use anyhow::{bail, Context, Result};
 use bdk::bitcoin::secp256k1::{SecretKey, Signature};
-use bdk::bitcoin::{Address, Amount, PublicKey, Transaction};
+use bdk::bitcoin::{Address, Amount, PublicKey, SignedAmount, Transaction};
 use bdk::descriptor::Descriptor;
 use cfd_protocol::secp256k1_zkp::{EcdsaAdaptorSignature, SECP256K1};
 use cfd_protocol::{finalize_spend_transaction, spending_tx_sighash};
 use rocket::request::FromParam;
+use rust_decimal::prelude::{FromPrimitive, ToPrimitive};
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use serde::{Deserialize, Serialize};
 use std::fmt;
-use std::ops::RangeInclusive;
+use std::ops::{Neg, RangeInclusive};
 use std::time::{Duration, SystemTime};
 use uuid::Uuid;
 
@@ -390,10 +391,16 @@ impl Cfd {
         Ok(margin)
     }
 
-    pub fn profit(&self, current_price: Usd) -> Result<(Amount, Usd)> {
-        let profit =
-            calculate_profit(self.order.price, current_price, dec!(0.005), Usd(dec!(0.1)))?;
-        Ok(profit)
+    pub fn profit(&self, current_price: Usd) -> Result<(SignedAmount, Percent)> {
+        let (p_n_l, p_n_l_percent) = calculate_profit(
+            self.order.price,
+            current_price,
+            self.quantity_usd,
+            self.margin()?,
+            self.position(),
+        )?;
+
+        Ok((p_n_l, p_n_l_percent))
     }
 
     #[allow(dead_code)] // Not used by all binaries.
@@ -702,14 +709,90 @@ pub enum CfdStateChangeEvent {
     CommitTxSent,
 }
 
+/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
 fn calculate_profit(
-    _intial_price: Usd,
-    _current_price: Usd,
-    _interest_per_day: Decimal,
-    _fee: Usd,
-) -> Result<(Amount, Usd)> {
-    // TODO: profit calculation
-    Ok((Amount::ZERO, Usd::ZERO))
+    initial_price: Usd,
+    current_price: Usd,
+    quantity: Usd,
+    margin: Amount,
+    position: Position,
+) -> Result<(SignedAmount, Percent)> {
+    let margin_as_sat =
+        Decimal::from_u64(margin.as_sat()).context("Expect to be a valid decimal")?;
+
+    let initial_price_inverse = dec!(1)
+        .checked_div(initial_price.0)
+        .context("Calculating inverse of initial_price resulted in overflow")?;
+    let current_price_inverse = dec!(1)
+        .checked_div(current_price.0)
+        .context("Calculating inverse of current_price resulted in overflow")?;
+
+    // calculate profit/loss (P and L) in BTC
+    let profit_btc = match position {
+        Position::Buy => {
+            // for long: profit_btc = quantity_usd * ((1/initial_price)-(1/current_price))
+            quantity
+                .0
+                .checked_mul(
+                    initial_price_inverse
+                        .checked_sub(current_price_inverse)
+                        .context("Subtracting current_price_inverse from initial_price_inverse resulted in an overflow")?,
+                )
+        }
+        Position::Sell => {
+            // for short: profit_btc = quantity_usd * ((1/current_price)-(1/initial_price))
+            quantity
+                .0
+                .checked_mul(
+                    current_price_inverse
+                        .checked_sub(initial_price_inverse)
+                        .context("Subtracting initial_price_inverse from current_price_inverse resulted in an overflow")?,
+                )
+        }
+    }
+        .context("Calculating profit/loss resulted in an overflow")?;
+
+    let sat_adjust = Decimal::from(Amount::ONE_BTC.as_sat());
+    let profit_btc_as_sat = profit_btc
+        .checked_mul(sat_adjust)
+        .context("Could not adjust profit to satoshi")?;
+
+    // loss cannot be more than provided margin
+    let margin_plus_profit_btc = margin_as_sat
+        .checked_add(profit_btc_as_sat)
+        .context("Adding up margin and profit_btc resulted in an overflow")?;
+
+    let in_percent = if profit_btc_as_sat.is_zero() {
+        Decimal::ZERO
+    } else {
+        profit_btc_as_sat
+            .checked_div(margin_as_sat)
+            .context("Profit divided by margin resulted in overflow")?
+    };
+
+    if margin_plus_profit_btc < Decimal::ZERO {
+        return Ok((
+            SignedAmount::from_sat(
+                margin_as_sat
+                    .neg()
+                    .to_i64()
+                    .context("Could not convert margin to i64")?,
+            ),
+            dec!(-100).into(),
+        ));
+    }
+
+    Ok((
+        SignedAmount::from_sat(
+            profit_btc_as_sat
+                .to_i64()
+                .context("Could not convert profit to i64")?,
+        ),
+        in_percent
+            .checked_mul(dec!(100.0))
+            .context("Converting to percent resulted in an overflow")?
+            .into(),
+    ))
 }
 
 pub trait AsBlocks {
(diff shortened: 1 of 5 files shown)


daemon/src/model/cfd.rs
```diff
@@ -390,10 +391,15 @@ impl Cfd {
         Ok(margin)
     }
 
-    pub fn profit(&self, current_price: Usd) -> Result<(Amount, Usd)> {
-        let profit =
-            calculate_profit(self.order.price, current_price, dec!(0.005), Usd(dec!(0.1)))?;
-        Ok(profit)
+    pub fn profit(&self, current_price: Usd) -> Result<(Decimal, Decimal)> {
```
da-kami: I think the profit calculation has to be different for taker and maker - at the moment we use the same formula, does that make sense?

bonomat: Why should it be different? As long as we fill in the correct values (maker's margin), it should be correct. 

// Edit: 
you are right. It should be different.

daemon/src/model/cfd.rs
```diff
@@ -700,14 +706,59 @@ pub enum CfdStateChangeEvent {
...
-    _fee: Usd,
-) -> Result<(Amount, Usd)> {
-    // TODO: profit calculation
-    Ok((Amount::ZERO, Usd::ZERO))
+    initial_price: Usd,
+    current_price: Usd,
+    quantity: Usd,
+    margin: Amount,
+) -> Result<(Decimal, Decimal)> {
+    let margin = Decimal::from_f64(margin.as_btc()).context("Expect to be a valid decimal")?;
```
da-kami: I would still do the conversion from the `u64` sat and avoid float here. You can always shift the Decimal.
I see that this would result in the same thing, but I would keep it consistent in not using `from_f64` at all if possible.

bonomat: I opted for using Bitcoin for the UI and hence I use `as_btc` here and avoid back and forth conversion.

daemon/src/model/cfd.rs
da-kami: But they have different positions, and the position is not taken into account in the calculation. Wouldn't this result in always calculating a positive profit or negative profit for both?

daemon/src/model/cfd.rs
```diff
@@ -700,14 +706,59 @@ pub enum CfdStateChangeEvent {
...
+
+    let one_through_initial_price = dec!(1)
+        .checked_div(initial_price.0)
+        .context("Division to work")?;
```
thomaseizinger: The way `.context` works is that it adds a "layer" of error message to the chain of errors. For these error message to be the most helpful, they need to represent causes that get more detailed with every layer. When `anyhow` prints out such a chain of causes, it looks like this:

```
Error: Failed to read instrs from ./path/to/instrs.json

Caused by:
    No such file or directory (os error 2)
 ```

"Division to work" is not a cause for why this code is failing. Imagine how this error message will look like:

```
Error: Failed to calculate profit

Caused by:
    Division to work
```

Whenever we use `context`, we should think in _causes_ that explain why the underlying code failed. To be even more precise, we should add **context** the implementation of the failing function does not have available. For example, the implementation does not know the domain or potentially hard-coded parameter values.

When [examining](https://docs.rs/rust_decimal/1.16.0/src/rust_decimal/decimal.rs.html#1550-1556) why `checked_div` can fail, we can see that it can return `None` for two reasons:

1. Overflow
2. Division by Zero

In our concrete case, division by zero would only apply if the price was zero. That is something we can likely rule out, meaning a more useful cause here would be say something like:

> Calculating inverse of price resulted in overflow

Not only does this add context by including the domain of the parameter (the price), it also tells us what exactly failed!

Please apply the same reasoning to the rest of the function and error handling in general! Thank you! :)

File: daemon/src/model/cfd.rs (hunk 1 above)
Comment (bonomat) [reply daemon/src/model/cfd.rs]: Fixed in the last commit.

File: daemon/src/model/cfd.rs [hunk 4]
```diff
@@ -700,14 +706,59 @@ pub enum CfdStateChangeEvent {
...
-    // TODO: profit calculation
-    Ok((Amount::ZERO, Usd::ZERO))
+    initial_price: Usd,
+    current_price: Usd,
+    quantity: Usd,
+    margin: Amount,
+) -> Result<(Decimal, Decimal)> {
+    let margin = Decimal::from_f64(margin.as_btc()).context("Expect to be a valid decimal")?;
+
+    let one_through_initial_price = dec!(1)
```
thomaseizinger: Nit: Perhaps `initial_price_inverse` or `one_over_initial_price` is slightly more idiomatic English?

daemon/src/model/cfd.rs
thomaseizinger: Please read this: https://stackoverflow.com/a/3730040/2489334

Using float for monetary calculations ALWAYS includes loss of precision.

There is no "back and forth here". `Amount`s are already in sats internally, the only conversation to BTC should happen in the UI / API layer.

daemon/src/model/cfd.rs
```diff
@@ -700,14 +707,75 @@ pub enum CfdStateChangeEvent {
...
+    } else {
+        profit_btc.checked_div(margin).context("Division to work")?
+    };
+    if margin_plus_profit_btc < Decimal::ZERO {
+        Ok((
+            margin.neg(),
+            Decimal::ONE
+                .neg()
+                .checked_mul(dec!(100.0))
+                .context("Multiplication to work")?,
```
thomaseizinger: This is a static value, can we declare that via `dec!` please?

daemon/src/model/cfd.rs
```diff
@@ -700,14 +707,75 @@ pub enum CfdStateChangeEvent {
...
+        .checked_add(profit_btc)
+        .context("Expect addition to work")?;
+
+    let in_percent = if profit_btc.is_zero() {
+        Decimal::ZERO
+    } else {
+        profit_btc.checked_div(margin).context("Division to work")?
+    };
+    if margin_plus_profit_btc < Decimal::ZERO {
+        Ok((
```
thomaseizinger: An early return for either one or the other would be more readable in my opinion.

daemon/src/model/cfd.rs
```diff
@@ -700,14 +707,75 @@ pub enum CfdStateChangeEvent {
...
-    _fee: Usd,
-) -> Result<(Amount, Usd)> {
-    // TODO: profit calculation
-    Ok((Amount::ZERO, Usd::ZERO))
+    initial_price: Usd,
+    current_price: Usd,
+    quantity: Usd,
+    margin: Amount,
+    position: Position,
+) -> Result<(Decimal, Decimal)> {
```
thomaseizinger: Returning a tuple with the same type is asking for trouble to mix them up.

The first value is in Bitcoin yes? Can we return that as `bitcoin::Amount` please?

daemon/src/model/cfd.rs
```diff
@@ -841,6 +909,169 @@ mod tests {
...
         assert!(blocks - error_margin < 0.1 && blocks + error_margin > 0.1);
     }
+
+    #[test]
+    fn test_profit_calculation_no_profit() {
+        let initial_price = Usd::from(dec!(10_000));
+        let closing_price = Usd::from(dec!(10_000));
+        let quantity = Usd::from(dec!(10_000));
+        let margin = Amount::ONE_BTC;
+        let (profit, in_percent) = calculate_profit(
```
thomaseizinger: Nit: A newline here would separate "arrange" - "act" - "assert" better.

daemon/src/model/cfd.rs
```diff
@@ -841,6 +909,169 @@ mod tests {
         let blocks = duration.as_blocks();
         assert!(blocks - error_margin < 0.1 && blocks + error_margin > 0.1);
     }
+
+    #[test]
+    fn test_profit_calculation_no_profit() {
```
thomaseizinger: These tests are very repetitive. We could declare a single test function and run a list of values through it to make this less verbose.

daemon/src/to_sse_event.rs
```diff
@@ -103,7 +103,10 @@ impl ToSseEvent for CfdsWithCurrentPrice {
             .cfds
             .iter()
             .map(|cfd| {
-                let (profit_btc, profit_usd) = cfd.profit(current_price).unwrap();
+                let (profit_btc, profit_in_percent) = cfd
+                    .profit(current_price)
+                    .map(|(p_l, in_percent)| (p_l.round_dp(6), in_percent.round_dp(2)))
+                    .unwrap_or((Decimal::ZERO, Decimal::ZERO));
```
thomaseizinger: Masking failures here with a default value will be a debugging nightmare. Can we at least add a log statement here?

daemon/src/to_sse_event.rs
```diff
@@ -103,7 +103,10 @@ impl ToSseEvent for CfdsWithCurrentPrice {
             .cfds
             .iter()
             .map(|cfd| {
-                let (profit_btc, profit_usd) = cfd.profit(current_price).unwrap();
+                let (profit_btc, profit_in_percent) = cfd
+                    .profit(current_price)
+                    .map(|(p_l, in_percent)| (p_l.round_dp(6), in_percent.round_dp(2)))
```
thomaseizinger: I'd prefer if we output nice domain types from our domain layer. (This is essentially the same feedback as above with not returning decimals from `profit`).

These domain types could then have more descriptive functions for what you are doing here. (Plus be tested in isolation).

I'd prefer the API layer to be rather stupid in regards to its mapping logic.

thomaseizinger (changes requested):
Thanks!

I don't want to return `Decimal`s from our domain layer, please use proper descriptive types there.

The other feedback is not strictly blocking.

daemon/src/model/cfd.rs
bonomat: > The first value is in Bitcoin yes? Can we return that as bitcoin::Amount please?

`bitcoin::Amount` uses a u64 internally so unfortunately we can't because the value can be negative.

daemon/src/to_sse_event.rs
bonomat: What types do you want this domain type to wrap around? Is Decimal fine?

daemon/src/model/cfd.rs
bonomat: With back and fourth I mean something what we do in `margin()`: 

```
    let sat_adjust = Decimal::from(Amount::ONE_BTC.as_sat()).into();
    let margin = margin.checked_mul(sat_adjust)?;
    let margin = Amount::from_sat(margin.try_into_u64()?);
```

I'm happy to use `from_u64` instead.

daemon/src/model/cfd.rs
thomaseizinger: Use https://docs.rs/bitcoin/0.27.1/bitcoin/util/amount/struct.SignedAmount.html then.

daemon/src/to_sse_event.rs
thomaseizinger: Yeah it is fine if `Decimal` is the internal type as long as it is an implementation detail!

daemon/src/model/cfd.rs
bonomat: Actually I change my mind. Imho in this case it does not make much sense to use sats: `profit_btc` is in _Bitcoin_ (resulting from the formula we use) and I only need the `margin` for subtracting it from the `profit_btc`. 

The other time where `margin` is used is for calculating the `profit_in_percent` which is not a crucial value and should be rounded anyways. 

Any more thoughts on that?

da-kami: I still think we should do the "back-and-forth" because every time I would look at this code I'd have to think about if the `f64` conversion should not be removed here. I would change it for consistency reasons, even if it results in more "unnecessary" adjustment steps for this particular instance.

bonomat: No worries, I'll do it ;) It turned out to not be the worst idea after all.

bonomat:
I guess I squash the commits once approved.

da-kami (approved):
thanks for the nice cleanup!

thomaseizinger (approved):
Thank you!

PR #195, opened 2021-10-05
Payout curve

### LINKED ISSUE #60: Add ability to define payout curves generically
Initial plan is to use a curve that corresponds to a clipped version of the function $f(x) = 1/x$, but no need to limit ourselves to only this function.

DeliciousHair:
Resolves #60.

### DIFF
--- a/daemon/src/payout_curve.rs
+++ b/daemon/src/payout_curve.rs
@@ -1,30 +1,744 @@
-use crate::model::{Leverage, Usd};
-use anyhow::Result;
-use bdk::bitcoin;
-use cfd_protocol::interval::MAX_PRICE_DEC;
-use cfd_protocol::{generate_payouts, Payout};
+use std::fmt;
 
-pub fn calculate(
-    price: Usd,
-    _quantity: Usd,
-    maker_payin: bitcoin::Amount,
-    (taker_payin, _leverage): (bitcoin::Amount, Leverage),
-) -> Result<Vec<Payout>> {
-    let dollars = price.try_into_u64()?;
-    let payouts = vec![
-        generate_payouts(
-            0..=(dollars - 10),
-            maker_payin + taker_payin,
-            bitcoin::Amount::ZERO,
-        )?,
-        generate_payouts((dollars - 9)..=(dollars + 10), maker_payin, taker_payin)?,
-        generate_payouts(
-            (dollars + 11)..=MAX_PRICE_DEC,
-            bitcoin::Amount::ZERO,
-            maker_payin + taker_payin,
-        )?,
-    ]
-    .concat();
+use crate::model::{Leverage, Usd};
+use crate::payout_curve::curve::Curve;
+use anyhow::{Context, Result};
+use bdk::bitcoin;
+use cfd_protocol::{generate_payouts, Payout};
+use itertools::Itertools;
+use ndarray::prelude::*;
+use num::{FromPrimitive, ToPrimitive};
+use rust_decimal::Decimal;
+
+mod basis;
+mod basis_eval;
+mod compat;
+mod csr_tools;
+mod curve;
+mod curve_factory;
+mod splineobject;
+mod utils;
+
+/// function to generate an iterator of values, heuristically viewed as:
+///
+///     `[left_price_boundary, right_price_boundary], maker_payout_value`
+///
+/// with units
+///
+///     `[Usd, Usd], bitcoin::Amount`
+///
+/// A key item to note is that although the POC logic has been to imposed
+/// that maker goes short every time, there is no reason to make the math
+/// have this imposition as well. As such, the `long_position` parameter
+/// is used to indicate which party (Maker or Taker) has the long position,
+/// and everything else is handled internally.
+///
+/// As well, the POC has also demanded that the Maker always has unity
+/// leverage, hence why the ability to to specify this amount has been
+/// omitted from the parameters. Internally, it is hard-coded to unity
+/// in the call to PayoutCurve::new(), so this behaviour can be changed in
+/// the future trivially.
+///
+/// ### Paramters
+///
+/// * price: BTC-USD exchange rate used to create CFD contract
+/// * quantity: Interger number of one-dollar USD contracts contained in the
+/// CFD; expressed as a Usd amount
+/// * leverage: Leveraging used by the taker
+///
+/// ### Returns
+///
+/// The list of [`Payout`]s for the given price, quantity and leverage.
+pub fn calculate(price: Usd, quantity: Usd, leverage: Leverage) -> Result<Vec<Payout>> {
+    let payouts = calculate_payout_parameters(price, quantity, leverage)?
+        .into_iter()
+        .map(PayoutParameter::into_payouts)
+        .flatten_ok()
+        .collect::<Result<Vec<_>>>()?;
 
     Ok(payouts)
+}
+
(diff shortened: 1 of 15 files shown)


cfd_protocol/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 mod oracle;
 mod protocol;
 
+pub mod curve_handlers;
```
thomaseizinger: I would like all of this code to move into the `daemon` crate, it doesn't really belong to the protocol as such. It is an implementation detail of our application which lives inside the `daemon`.

As an added benefit, defining the code within `daemon` should correctly flag any kind of unused code once it is integrated.

cfd_protocol/src/curve_handlers/utils.rs
```diff
@@ -0,0 +1,370 @@
...
+    let u = get_unique_f64(&t.slice(s![k - 1..k + 1]).into_dyn().to_owned())?;
+    let s = u.shape().iter().product::<usize>() - 1;
+    let d = p.shape()[0];
+
+    // Find multiplicity of the knot vector t
+    let b = BSplineBasis::new(Some(k), Some(t.clone()), None)?;
+    let z = b
+        .knot_spans(false)
+        .iter()
+        .map(|x| (k as isize) - 1 - b.continuity(*x).unwrap())
```
thomaseizinger: Is this guaranteed to not panic?

thomaseizinger (changes requested):
I can't really review any of the math :man_shrugging: 

What I'd like to see is one or more tests that only use the `payout_curve::calculate` function. Something like:

```rust
#[test]
fn basic_test() {
	let price = Usd(dec!(49000));
	let quanitity = Usd(dec!(300));
	let maker_buyin = bitcoin::Amount::from_btc(0.5).unwrap();
	let taker_buyin = bitcoin::Amount::from_btc(0.01).unwrap();
	let leverage = Leverage(5);

	let payouts = calculate(price, quantity, maker_buying, (taker_buyin, leverage)).unwrap();

	assert_eq!(
		payouts,
		vec![
			Payout::new(
				0..30000,
				bitcoin::Amount::from_sat(51000000),
				bitcoin::Amount::ZERO
			),
			// TODO: More payouts here
		].flatten()
	);
}
```

We don't need to calculate these payouts upfront, it is fine if we run the test and then copy the results in. I just want at least one test that sets the functionality of all of these modules in stone.

That will give us piece of mind in regards to refactoring / changing it, plus we get an idea on whether or not it performs as expected.

cfd_protocol/src/lib.rs
DeliciousHair: Can do! Lemme get an API sorted and I'll move it all over.

cfd_protocol/src/curve_handlers/utils.rs
DeliciousHair: Not 100% sure TBH, but I cannot think of a way it could off the top of my head.

daemon/src/payout_curve.rs
```diff
@@ -1,30 +1,740 @@
...
+
+    let n_payouts = 200;
+    let payouts_arr = maker_payout_curve.generate_payout_scheme(n_payouts)?;
+    let mut payouts_vec = Vec::<_>::with_capacity(n_payouts);
+
+    for (i, e) in payouts_arr.slice(s![..n_payouts, 0]).iter().enumerate() {
+        let bnd_left = *e as u64;
+        let bnd_right = payouts_arr[[i, 1]] as u64;
+
+        let maker_amount = payouts_arr[[i, 2]];
```
thomaseizinger: It is fixed here ...

thomaseizinger:
@da-kami @luckysori @klochowicz @bonomat @rishflab 

Can you guys please checkout this PR and run all tests to see if they pass on your machine? In particular `calculate_snapshot` is important.

bonomat:
> @da-kami @luckysori @klochowicz @bonomat @rishflab
> 
> Can you guys please checkout this PR and run all tests to see if they pass on your machine? In particular `calculate_snapshot` is important.

ran it a couple of times without failures 👍

klochowicz:
> @da-kami @luckysori @klochowicz @bonomat @rishflab
> 
> Can you guys please checkout this PR and run all tests to see if they pass on your machine? In particular `calculate_snapshot` is important.

same here, ran it like 5 times in a row - no failures 🏋️‍♀️

luckysori:
> @da-kami @luckysori @klochowicz @bonomat @rishflab
> 
> Can you guys please checkout this PR and run all tests to see if they pass on your machine? In particular `calculate_snapshot` is important.

Test passed like 10/10.

daemon/src/payout_curve.rs
```diff
@@ -1,30 +1,794 @@
...
+    let quantity = quantity
+        .try_into_u64()
+        .context("Cannot convert quantity to u64")? as usize;
+    let maker_leverage = 1;
+
+    let maker_payout_curve = match long_position {
+        LongPosition::Maker => PayoutCurve::new(
+            initial_rate as f64,
+            maker_leverage,
+            taker_leverage.0 as usize,
```
bonomat: This seems to be switched: 

Second argument says: `leverage_long` but `maker_leverage` is provided who is short. and third argument says `leverage_short` but `taker_leverage` is provided who is long.

Or am I misreading something?

bonomat:
> @bonomat I think these values make sense now but can you also have another look?

The numbers look better now. Thanks for looking into this. 

Would a different step size make a difference (e.g. of 10)? 

(Looking at the example in `calculate_snapshot`) 
Right now the difference to jump onto the next step of payout is `1`$, which has the effect that 
for `$45315` the short would get `7750759` sats. 
If the price moves up by `$1` he would receive `7697244` sats which is a difference `0.70%` while the `$1` was a difference of `0.0022%`.

DeliciousHair:
> @bonomat
> Would a different step size make a difference (e.g. of 10)?
> 
> (Looking at the example in `calculate_snapshot`) Right now the difference to jump onto the next step of payout is `1`$, which has the effect that for `$45315` the short would get `7750759` sats. If the price moves up by `$1` he would receive `7697244` sats which is a difference `0.70%` while the `$1` was a difference of `0.0022%`.

Step size is not (directly) controllable, instead it falls out as a result of setting the number of steps to take. Which is also not controllable per-se at this point as the value is set at (I think) 200 for the moment. Regardless, changing the number of steps is easy, even making it a user-defined parameter. Fixing the step size, on the other hand, is not so straightforward. It can be done, but requires a change at the logic level with regards to how things are computed.

Also, with regards to your example numbers, recall that this curve follows a function of the form `payout ~ 1/exchange_rate`. So for some values, a small price difference will lead to a relatively large change in price, for others it will lead to a relatively small value.

da-kami (commented):
Thanks for all the work! 
I'm having a hard time approving this, given all the complex math that was added. 

Thanks for adding the test, it helps validating the sanity of the output.
Given, that the values that can go into the calculations can be manyfold, and we only test with one set: Can we add proptests so we can at least ensure, that we don't panic on a specific price?

da-kami (approved):
Can actually add more tests in follow up, we can merge this from my point of view. 
I would recommend proptests to ensure that we don't panic on certain combinations, where some values can be fixed, but the price should be variable.
Not sure how time intensive that would be though.

thomaseizinger (approved):
bors r+

PR #196, opened 2021-10-05
Run CI on windows

thomaseizinger:


da-kami (approved):

da-kami:
shall we add to to building the release binaries as well?

thomaseizinger:
> shall we add to to building the release binaries as well?

Releases are already built on windows!

PR #197, opened 2021-10-05
Transition cfd to new state when handling an event

da-kami:
Previously we only returned the new state, but we also have to set it internally, otherwise the cfd might be out of date.

This was a problem when getting the refund transaction from the cfd directly after a state transition to `MustRefund`.
The cfd instance had not actually transitioned to the new state, even though it was saved in the database.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -449,7 +449,7 @@ impl Cfd {
     #[allow(dead_code)]
     pub const CET_TIMELOCK: u32 = 12;
 
-    pub fn handle(&self, event: CfdStateChangeEvent) -> Result<CfdState> {
+    pub fn handle(&mut self, event: CfdStateChangeEvent) -> Result<CfdState> {
         use CfdState::*;
 
         // TODO: Display impl
@@ -590,6 +590,8 @@ impl Cfd {
                 }
             }
         };
+
+        self.state = new_state.clone();
 
         Ok(new_state)
     }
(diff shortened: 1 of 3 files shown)


daemon/src/model/cfd.rs
```diff
@@ -449,7 +449,7 @@ impl Cfd {
     #[allow(dead_code)]
     pub const CET_TIMELOCK: u32 = 12;
 
-    pub fn handle(&self, event: CfdStateChangeEvent) -> Result<CfdState> {
+    pub fn handle(&mut self, event: CfdStateChangeEvent) -> Result<CfdState> {
```
thomaseizinger, line 452: I think it might even be easier to understand to consume `self` and return an entirely new `Cfd`?

thomaseizinger (approved):

daemon/src/model/cfd.rs
da-kami: I agree, but this is a quite intrusive change and I would do it in a separate refactoring. I'll create a ticket.

There are several considerations we *might* want to on top of that: 

- Not save "state by id" because it is error prone. Instead always change a cfd and then save that cfd.
- Move **all** state transitions onto the cfd's `handle` function. At the moment there are still several early stage states that are just transitioned in the cfd actor without the cfd `handle` function being involved.

da-kami: https://github.com/comit-network/hermes/issues/200

PR #201, opened 2021-10-05
Use olivia's attestation scheme instead of dlcspecs'

luckysori:
We also stopped generating random oracle-related values for the protocol tests, greatly simplifying the test suite. This has the
side-effect that we no longer verify that _all_ CETs can be unlocked.

Re-introducing random oracle-related data in the future should be considered.


cfd_protocol/tests/cfds.rs
```diff
@@ -982,21 +928,82 @@ fn make_keypair(rng: &mut (impl RngCore + CryptoRng)) -> (SecretKey, PublicKey)
...
-        start
-    } else {
-        rng.gen_range(start, end)
+impl OliviaData {
+    /// An example of all the data necessary from `olivia` to test the
+    /// CFD protocol.
+    ///
+    /// Data comes from this event:
+    /// https://outcome.observer/h00.ooo/x/BitMEX/BXBT/2021-10-05T02:00:00.price[n:20].
+    pub fn example() -> Self {
```
luckysori, line 945: @thomaseizinger: I opted for this instead of using the code in `daemon/oracle.rs` because it's not trivial to use that code in several places:

- can't depend on the application from the library.
- moving it into the library is tricky because some of the types still belong in `daemon`, so it's worth designing the sub-module/sub-crate properly.
- copying the `oracle_api` sub-module into the test folder is overkill and also doesn't work out of the box.

And more broadly, I think we should reinstate tests in which we have full control of the oracle, because then we can test each CET.

cfd_protocol/src/oracle.rs
```diff
@@ -1,75 +1,32 @@
...
+    index: NonZeroU8,
+) -> Result<secp256k1_zkp::PublicKey> {
+    let nonce_pk = schnorr_pubkey_to_pubkey(nonce_pk);
+    let nonce_pk_sum = (0..index.get()).map(|_| &nonce_pk).collect::<Vec<_>>();
 
-    fn verify(sig: &schnorrsig::Signature, msg: &[u8], pk: &schnorrsig::PublicKey) -> bool {
-        let msg = secp256k1_zkp::Message::from_hashed_data::<sha256::Hash>(msg);
-        SECP256K1.schnorrsig_verify(sig, &msg, pk).is_ok()
-    }
+    let nonce_pk_sum = PublicKey::combine_keys(&nonce_pk_sum)?;
```
thomaseizinger: I think this could also be solved by https://docs.rs/secp256k1/0.20.3/secp256k1/key/struct.PublicKey.html#method.mul_assign, not sure what is more convenient.

thomaseizinger (approved):

PR #203, opened 2021-10-05
Offer price based on current price

da-kami:
To keep it simple we set the offer price to current price + 3% spread if we have a current price and the offer's initial value is 0.
This. means we cannot set the offer price to 0 anymore, but that is OK.
Additionally, there is a refresh button so the maker can refresh the offer price (no auto-refresh for now).

(did not invest the time for a reducer, but eventually we might want to invest the time...)

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -29,6 +29,8 @@ import { Cfd, intoCfd, intoOrder, Order, PriceInfo, StateGroupKey, WalletInfo } 
 import Wallet from "./components/Wallet";
 import { CfdSellOrderPayload, postCfdSellOrderRequest } from "./MakerClient";
 
+const SPREAD = 1.03;
+
 export default function App() {
     let source = useEventSource({ source: "/api/feed", options: { withCredentials: true } });
 
@@ -42,14 +44,7 @@ export default function App() {
     let [minQuantity, setMinQuantity] = useState<string>("10");
     let [maxQuantity, setMaxQuantity] = useState<string>("100");
     let [orderPrice, setOrderPrice] = useState<string>("0");
-
-    const spread = 1.03;
-    if (priceInfo && orderPrice === "0") {
-        setOrderPrice((priceInfo.bid * spread).toString());
-    }
-
-    const format = (val: any) => `$` + val;
-    const parse = (val: any) => val.replace(/^\$/, "");
+    let [hasEnteredPrice, setHasEnteredPrice] = useState(false);
 
     let { run: makeNewCfdSellOrder, isLoading: isCreatingNewCfdOrder } = useAsync({
         deferFn: async ([payload]: any[]) => {
@@ -107,15 +102,18 @@ export default function App() {
                         <Text>Order Price:</Text>
                         <HStack>
                             <CurrencyInputField
-                                onChange={(valueString: string) => setOrderPrice(parse(valueString))}
-                                value={format(orderPrice)}
+                                onChange={(valueString: string) => {
+                                    setOrderPrice(parse(valueString));
+                                    setHasEnteredPrice(true);
+                                }}
+                                value={priceToDisplay(hasEnteredPrice, orderPrice, priceInfo)}
                             />
                             <IconButton
                                 aria-label="Reduce"
                                 icon={<RepeatIcon />}
                                 onClick={() => {
                                     if (priceInfo) {
-                                        setOrderPrice((priceInfo.bid * spread).toString());
+                                        setOrderPrice((priceInfo.bid * SPREAD).toString());
                                     }
                                 }}
                             />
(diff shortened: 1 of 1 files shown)


frontend/src/MakerApp.tsx
```diff
@@ -37,9 +39,14 @@ export default function App() {
...
-    let [minQuantity, setMinQuantity] = useState<string>("100");
-    let [maxQuantity, setMaxQuantity] = useState<string>("1000");
-    let [orderPrice, setOrderPrice] = useState<string>("10000");
+    let [minQuantity, setMinQuantity] = useState<string>("10");
+    let [maxQuantity, setMaxQuantity] = useState<string>("100");
+    let [orderPrice, setOrderPrice] = useState<string>("0");
+
+    const spread = 1.03;
+    if (priceInfo && orderPrice === "0") {
+        setOrderPrice((priceInfo.bid * spread).toString());
```
bonomat: He should be using the `ask` price. 

The `ask` price is the `lowest` price the seller is willing to sell the price. 
In our case, the `maker` is `selling` to the `taker`.

Maybe it's clearer if you look at the values we get from BitMEX. `ask>bid`: you buy (pay the `ask` price) at a higher value then you sell (pay the `bid` price).

frontend/src/MakerApp.tsx
```diff
@@ -98,10 +105,21 @@ export default function App() {
...
+                            <CurrencyInputField
+                                onChange={(valueString: string) => setOrderPrice(parse(valueString))}
+                                value={format(orderPrice)}
+                            />
+                            <IconButton
+                                aria-label="Reduce"
+                                icon={<RepeatIcon />}
+                                onClick={() => {
+                                    if (priceInfo) {
+                                        setOrderPrice((priceInfo.bid * spread).toString());
```
bonomat: Same as above, please use `ask`.

frontend/src/MakerApp.tsx
```diff
@@ -1,3 +1,4 @@
+import { ChevronUpIcon, RepeatIcon } from "@chakra-ui/icons";
```
bonomat: Nice, github action is pointing you out what's wrong :)

bonomat (changes requested):
Please use the `ask` price instead of `bid`.

frontend/src/MakerApp.tsx
```diff
@@ -37,9 +39,14 @@ export default function App() {
...
     const toast = useToast();
-    let [minQuantity, setMinQuantity] = useState<string>("100");
-    let [maxQuantity, setMaxQuantity] = useState<string>("1000");
-    let [orderPrice, setOrderPrice] = useState<string>("10000");
+    let [minQuantity, setMinQuantity] = useState<string>("10");
+    let [maxQuantity, setMaxQuantity] = useState<string>("100");
+    let [orderPrice, setOrderPrice] = useState<string>("0");
+
+    const spread = 1.03;
+    if (priceInfo && orderPrice === "0") {
```
thomaseizinger: This is not ideal, you shouldn't call these hook functions as part of the regular render code because they cause side-effects and rendering should be pure.

I think we should solve this by tracking the _real state_ that we care about which seems to be whether or not the user has entered a price themselves yet. Based on that, we can derive the price to display from the `priceInfo` or from whatever the user entered.

thomaseizinger (approved):
One comment about the initialization being impure, that is asking for trouble down the line ...

Trying to solve this locally atm.

thomaseizinger:
@da-kami Pushed a commit that addresses my comment.

frontend/src/MakerApp.tsx
da-kami: My thinking process: given an arbitrage model, the seller wants to be covered for the bid (because that's what the seller goes to market with. Not correct?

frontend/src/MakerApp.tsx
da-kami: mean though, I ran `yarn run eslint` locally and did not get that error 😅
I'm running the same yarn version... potentially different node version problem :/ ?

frontend/src/MakerApp.tsx
bonomat: I'm not sure I follow your line of thinking. Can you elaborate more please?

In case of hedging, the `maker` needs to mirror his trade and hence take the counter position on a different exchange. _Hedging_  means the he tries to offset losses on one trade by taking an opposite position on another trade. 

In our case, the `maker` goes `short` on Hermes and needs to go `long` on Bitmex. Since going long means he gets the `ask` price on Bitmex, he needs to offer the `ask` price on Hermes as well. To make some profit, he might want to add a spread.

da-kami: I think my thinking was just wrong :)

bonomat (approved):
Sweet :)

PR #204, opened 2021-10-05
Allow accepting/rejecting settlement proposals in the maker daemon

klochowicz:
Keep track of outgoing/incoming settlement proposals in a hashmap inside the CFD
rocket that is available as a Rocket state in order to derive correct CfdState
for the UI.

Add placeholders for actions for accepting/rejecting settlement offers in the maker.

Also:
    - rename Accept/Reject in few places to reduce ambiguity between interacting
      with an order and a settlement proposal.
    - allow only one in-flight settlement proposal from the taker

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -11,6 +11,7 @@ use rust_decimal::prelude::{FromPrimitive, ToPrimitive};
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
 use std::fmt;
 use std::ops::{Neg, RangeInclusive};
 use std::time::{Duration, SystemTime};
@@ -49,7 +50,7 @@ pub enum Origin {
 }
 
 /// Role in the Cfd
-#[derive(Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum Role {
     Maker,
     Taker,
@@ -348,6 +349,13 @@ pub struct SettlementProposal {
     pub timestamp: SystemTime,
     pub taker: Amount,
     pub maker: Amount,
+}
+
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // Variants used by different binaries
+pub enum SettlementProposals {
+    Incoming(HashMap<OrderId, SettlementProposal>),
+    Outgoing(HashMap<OrderId, SettlementProposal>),
 }
 
 /// Represents a cfd (including state)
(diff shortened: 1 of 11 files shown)


klochowicz:
I tested whether I can still accept an offer after renames, and the Cfd setup passes that point. I'm currently working on the UI changes that I'll post as a separate commit.

klochowicz:
> I tested whether I can still accept an offer after renames, and the Cfd setup passes that point. I'm currently working on the UI changes that I'll post as a separate commit.

basic UI got added in a separate commit (tried to be consistent with the way how we approve/reject orders). Actions availability and invocation seem to work fine - I was able to observe maker approve settlement proposal handler's output in the logs.

daemon/src/taker_cfd.rs
```diff
@@ -126,14 +141,27 @@ impl Actor {
...
+        if self
+            .current_settlement_proposals
+            .insert(proposal.order_id, proposal.clone())
+            .is_some()
+        {
+            anyhow::bail!(
+                "Settlement proposal for order id {} already present",
+                order_id
+            )
+        }
```
thomaseizinger: This will replace and then error, is that what you want?

daemon/src/model/cfd.rs
```diff
@@ -349,6 +350,13 @@ pub struct SettlementProposal {
     pub maker: Amount,
 }
 
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // Variants used by different binaries
+pub enum SettlementProposals {
+    Incoming(HashMap<OrderId, SettlementProposal>),
+    Outgoing(HashMap<OrderId, SettlementProposal>),
+}
```
thomaseizinger, line 359: Does this enum actually bring any advantages? I think we could split this into a single hashmap per binary no?

thomaseizinger: I think a better data structure would be:

```rust
HashMap<OrderId, (SettlementProposal, IncomingOrOutgoing)>
```

thomaseizinger (approved):
I think the data structure could be optimized but that is not blocking.

In the long run, I think the accept / reject order should be a similarly transient state. What do you think @da-kami?

daemon/src/model/cfd.rs
klochowicz: The difference would be that it means that we *could* have both incoming and outgoing settlement proposals in the taker and maker. the way as it is currently you can only setup the channel with either, and it can catch some programmer errors (e.g. as it did when I was developing :D ) 

Of course, if we even decide to handle both incoming and outgoing proposals the code will change.

daemon/src/taker_cfd.rs
klochowicz: oops, indeed that would happen! I'll post a quick fix for this one in a sec (bailing if `contains_key()` returns `true)

daemon/src/model/cfd.rs
thomaseizinger: Hmm, I think eventually we want both parties to propose collaborative settlement.

klochowicz: ok, I'll look into simplifying the data structure as a separate PR.

klochowicz: https://github.com/comit-network/hermes/pull/215

PR #205, opened 2021-10-05
Offer with oracle event

### LINKED ISSUE #193: Include oracle announcement ID in maker's offer
The oracle announcement that is "to be used" needs to be included in the offer. It is part of the contract terms the taker has to agree to.

When a taker's order is accepted by the maker, the taker needs to obtain the nonces of the particular event ID from the oracle.

da-kami:
fixes #193

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -5,7 +5,7 @@ use crate::db::{
 };
 use crate::maker_inc_connections::TakerCommand;
 use crate::model::cfd::{
-    Cfd, CfdState, CfdStateChangeEvent, CfdStateCommon, Dlc, Order, OrderId, Role,
+    Cfd, CfdState, CfdStateChangeEvent, CfdStateCommon, Dlc, Order, OrderId, Origin, Role,
     SettlementProposal, SettlementProposals,
 };
 use crate::model::{TakerId, Usd};
@@ -43,7 +43,11 @@ pub struct RejectSettlement {
     pub order_id: OrderId,
 }
 
-pub struct NewOrder(pub Order);
+pub struct NewOrder {
+    pub price: Usd,
+    pub min_quantity: Usd,
+    pub max_quantity: Usd,
+}
 
 pub struct NewTakerOnline {
     pub id: TakerId,
@@ -121,7 +125,26 @@ impl Actor {
             ))?)
     }
 
-    async fn handle_new_order(&mut self, order: Order) -> Result<()> {
+    async fn handle_new_order(
+        &mut self,
+        price: Usd,
+        min_quantity: Usd,
+        max_quantity: Usd,
+    ) -> Result<()> {
+        let oracle_event_id = if let Some(latest_announcement) = self.latest_announcement.clone() {
+            latest_announcement.id
+        } else {
+            bail!("Cannot create order because no announcement from oracle")
+        };
+
+        let order = Order::new(
+            price,
+            min_quantity,
+            max_quantity,
+            Origin::Ours,
+            oracle_event_id,
+        )?;
+
         // 1. Save to DB
         let mut conn = self.db.acquire().await?;
         insert_order(&order, &mut conn).await?;
@@ -565,7 +588,7 @@ impl Handler<Commit> for Actor {
 #[async_trait]
 impl Handler<NewOrder> for Actor {
     async fn handle(&mut self, msg: NewOrder, _ctx: &mut Context<Self>) {
-        log_error!(self.handle_new_order(msg.0));
+        log_error!(self.handle_new_order(msg.price, msg.min_quantity, msg.max_quantity));
     }
 }
 
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -12,7 +12,7 @@ use crate::model::{TakerId, Usd};
 use crate::monitor::MonitorParams;
 use crate::wallet::Wallet;
 use crate::{maker_inc_connections, monitor, oracle, setup_contract, wire};
-use anyhow::{Context as _, Result};
+use anyhow::{bail, Context as _, Result};
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
 use futures::channel::mpsc;
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -539,8 +539,6 @@ impl Actor {
         &mut self,
         announcements: oracle::Announcements,
     ) -> Result<()> {
-        tracing::debug!("Updating latest oracle announcements");
-
         self.latest_announcements.replace(
             announcements
                 .0
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -74,8 +74,8 @@ pub struct Actor {
     current_order_id: Option<OrderId>,
     monitor_actor: Address<monitor::Actor<Actor>>,
     setup_state: SetupState,
-    latest_announcement: Option<oracle::Announcement>,
-    _oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
+    latest_announcements: Option<oracle::Announcements>,
+    oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
     current_settlement_proposals: HashMap<OrderId, SettlementProposal>,
 }
 
@@ -111,8 +111,8 @@ impl Actor {
             current_order_id: None,
             monitor_actor,
             setup_state: SetupState::None,
-            latest_announcement: None,
-            _oracle_actor: oracle_actor,
+            latest_announcements: None,
+            oracle_actor,
             current_settlement_proposals: HashMap::new(),
         }
     }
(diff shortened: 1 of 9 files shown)


daemon/src/oracle.rs
```diff
@@ -200,9 +201,9 @@ fn event_url(datetime: OffsetDateTime) -> String {
 pub struct Announcement {
     /// Identifier for an oracle event.
     ///
-    /// Doubles up as the path of the URL for this event i.e.
+    /// Bubbles up as the path of the URL for this event i.e.
```
thomaseizinger: I think this is meant to say _doubles_.

thomaseizinger: The expression "To double up as something" means it serves two purposes :)

daemon/src/maker_cfd.rs
```diff
@@ -99,7 +103,26 @@ impl Actor {
...
+        &mut self,
+        price: Usd,
+        min_quantity: Usd,
+        max_quantity: Usd,
+    ) -> Result<()> {
+        let oracle_event_id = if let Some(latest_announcement) = self.latest_announcement.clone() {
+            latest_announcement.id
+        } else {
+            bail!("Cannot create order because no announcement from oracle")
+        };
```
thomaseizinger: `.context` on an `Option` does exactly this.

daemon/src/routes_maker.rs
```diff
@@ -95,15 +95,14 @@ pub async fn post_sell_order(
...
     cfd_actor_address
-        .do_send_async(maker_cfd::NewOrder(order))
+        .do_send_async(maker_cfd::NewOrder {
+            price: order.price,
+            min_quantity: order.min_quantity,
+            max_quantity: order.max_quantity,
+        })
         .await
-        .expect("actor to always be available");
+        .map_err(|e| status::BadRequest(Some(e.to_string())))?;
```
thomaseizinger: I know you just took it over but I really think this should be a 500. This failing has nothing to do with the user's request.

daemon/src/taker_cfd.rs
```diff
@@ -60,8 +60,8 @@ pub struct Actor {
     send_to_maker: Address<send_to_socket::Actor<wire::TakerToMaker>>,
     monitor_actor: Address<monitor::Actor<Actor>>,
     setup_state: SetupState,
-    latest_announcement: Option<oracle::Announcement>,
-    _oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
+    latest_announcements: Option<oracle::Announcements>,
```
thomaseizinger: Can we make this a `HashMap` instead? Sounds like the better data structure if we are going to look-up by event ID.

thomaseizinger: Just make sure to actively clear it on every update, otherwise we have a memory leak here.

daemon/src/maker_cfd.rs
```diff
@@ -64,8 +64,8 @@ pub struct Actor {
     current_order_id: Option<OrderId>,
     monitor_actor: Address<monitor::Actor<Actor>>,
     setup_state: SetupState,
-    latest_announcement: Option<oracle::Announcement>,
-    _oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
+    latest_announcements: Option<oracle::Announcements>,
```
thomaseizinger: You can use a `BTreeMap` here, that is a sorted `HashMap` so you can grab the last one but still look up by ID.

thomaseizinger (approved):
Thanks!

Some suggestions for using better data structures but nothing blocking.

daemon/src/model.rs
```diff
@@ -105,3 +105,6 @@ pub struct WalletInfo {
     pub address: Address,
     pub last_updated_at: SystemTime,
 }
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct OracleEventId(pub String);
```
luckysori, line 110: ❓ Would it make sense to put this type in `oracle` and name it `Id`?

daemon/src/model/cfd.rs
```diff
@@ -90,11 +90,22 @@ pub struct Order {
     pub term: Duration,
 
     pub origin: Origin,
+
+    /// The id of the event to be used for price attestation
+    ///
+    /// The maker includes this into the Order based on the Oracle's next event.
```
luckysori: 🔧 Technically it's not the _next_ event, but the event 24 hours from now IIRC.

daemon/src/taker_cfd.rs
```diff
@@ -184,18 +184,23 @@ impl Actor {
...
-        //     .await?;
+        let offer_announcements = self
+            .latest_announcements
+            .clone()
+            .context("No oracle announcements available")?;
+        let offer_announcement = offer_announcements
+            .0
+            .iter()
+            .find(|announcement| announcement.id == cfd.order.oracle_event_id)
+            .context("Order's announcement not found in list of current oracle announcements")?;
```
luckysori: 👍 So instead of asking the oracle actor for a specific announcement we assume that the announcement in the order we get from the maker is within our `latest_announcements` state? It should work and it's definitely simpler to implement it this way. 

Perhaps we can go back an introduce a dedicated message later on.

daemon/src/taker_cfd.rs
```diff
@@ -363,8 +368,7 @@ impl Actor {
         &mut self,
         announcements: oracle::Announcements,
     ) -> Result<()> {
-        tracing::debug!("Updating latest oracle announcements");
```
luckysori, line 395: 😭 Why don't you like my arbitrary, useless debug statements?

daemon/src/maker_cfd.rs
```diff
@@ -338,18 +344,23 @@ impl Actor {
...
-        //     .await?;
+        let offer_announcements = self
+            .latest_announcements
+            .clone()
+            .context("No oracle announcements available")?;
+        let offer_announcement = offer_announcements
+            .0
+            .iter()
+            .find(|announcement| announcement.id == cfd.order.oracle_event_id)
+            .context("Order's announcement not found in list of current oracle announcements")?;
```
luckysori: ❓ I can't see the types here, so maybe I'm missing something, but can't you just use the `cfd.order.oracle_event_id` directly when sending the `oracle::MonitorEvent` message to the oracle actor? The maker shouldn't need to verify anything about an offer _they_ created.

daemon/src/maker_cfd.rs
```diff
@@ -486,7 +497,7 @@ impl Actor {
         announcements: oracle::Announcements,
     ) -> Result<()> {
         tracing::debug!("Updating latest oracle announcements");
```
luckysori: 🔧 To be consistent you could also delete this useless debug statement.

luckysori (approved):
Looks good to me, just some little fixes.

daemon/src/maker_cfd.rs
luckysori: Okay, ignore this. You need to get the full announcement to get the `nonce_pks`.

daemon/src/taker_cfd.rs
da-kami: Went for keeping the array in the oracle and map + replace on every update in the other actors - ok?

daemon/src/taker_cfd.rs
da-kami: hahaha, I think I actually removed that one by accident - in the maker I kept it 🤦‍♂️

daemon/src/maker_cfd.rs
da-kami: hahaha, and there it is :D

daemon/src/model.rs
da-kami: This type is now used in various modules - it is also serialized/deserialized as part of the `Order`. We could be cleaner here, and map it to a specific type for serialization in `to_sse_event`. I admit I was too lazy for that. But I would keep it in `model` and treat it similar to e.g. `Usd` which is also ussed throughout the codebase but directly serialized for the UI.

daemon/src/maker_cfd.rs
da-kami: We need the actual `Announcement` to retrieve the `nonce_pks` after we send the message to the monitor actor. 
In the `Order` we only have the `OracleEventId`. I think it is safer to only trigger the monitoring once we know we also have the announcement (containing the `nonce_pks`), that's why I first retrieve the announcement and only if that succeeded go on with triggering monitoring.

daemon/src/model.rs
thomaseizinger: Would be nice to parse this one into its semantic parts at some point :)

daemon/src/maker_cfd.rs
```diff
@@ -543,7 +540,14 @@ impl Actor {
...
-        self.latest_announcements = Some(announcements);
+
+        self.latest_announcements.replace(
+            announcements
+                .0
+                .iter()
+                .map(|announcement| (announcement.id.clone(), announcement.clone()))
+                .collect(),
```
thomaseizinger, line 547: Nice use of `collect`!

daemon/src/taker_cfd.rs
```diff
@@ -63,7 +63,7 @@ pub struct Actor {
     send_to_maker: Address<send_to_socket::Actor<wire::TakerToMaker>>,
     monitor_actor: Address<monitor::Actor<Actor>>,
     setup_state: SetupState,
-    latest_announcements: Option<oracle::Announcements>,
+    latest_announcements: Option<BTreeMap<OracleEventId, oracle::Announcement>>,
```
thomaseizinger, line 66: Not sure if we need to semantically differentiate between `Option::None` and an empty map, so I think we could get rid of the `Option`.

thomaseizinger (approved):
Thanks!

PR #211, opened 2021-10-05
Rename "Accept / Reject" to "Pending" in CfdTables

klochowicz:
Pending is a more succinct nudge to the user that they should react with an
action to the CFDs inside these tables.

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -69,13 +69,9 @@ export default function App() {
         },
     });
 
-    const acceptOrRejectOrder = cfds.filter((value) => value.state.getGroup() === StateGroupKey.ACCEPT_OR_REJECT_ORDER);
-    const acceptOrRejectSettlement = cfds.filter((value) =>
-        value.state.getGroup() === StateGroupKey.ACCEPT_OR_REJECT_SETTLEMENT
-    );
-    const acceptOrRejectRollOvers = cfds.filter((value) =>
-        value.state.getGroup() === StateGroupKey.ACCEPT_OR_REJECT_ROLL_OVER
-    );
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):

da-kami:
I am not sure I agree. This group was created so the maker sees all `Accept / Reject` CFDs that have to be actioned. `Pending` is verbose and can mean a lot of things (which is reflected in the state). 
I could not come up with a better name than `Accept / Reject`, but I don't think `Pending` is better.

bonomat:
> I am not sure I agree. This group was created so the maker sees all `Accept / Reject` CFDs that have to be actioned. `Pending` is verbose and can mean a lot of things (which is reflected in the state). I could not come up with a better name than `Accept / Reject`, but I don't think `Pending` is better.

I talked to the person who wants to run the maker at the beginning and he said either way is fine.

klochowicz:
> I am not sure I agree. This group was created so the maker sees all `Accept / Reject` CFDs that have to be actioned. `Pending` is verbose and can mean a lot of things (which is reflected in the state). I could not come up with a better name than `Accept / Reject`, but I don't think `Pending` is better.

The way I see it is that tabs indicate a "property" of CFDs (for grouping purposes etc), so it's more natural to use adjectives than verbs. 
Verbs, such as "accept" or "reject" would be wonderful materials for buttons that actually do the action. 

Pending is definitely less verbose (it's a whole word less on 2 occasions), but I take your point that it might be a bit ambiguous at the first glance, before you start using the app.
I think  eventually we can add some tooltips that prevent any initial ambiguities.

da-kami:
🤷‍♂️

thomaseizinger (approved):
I prefer the _pending_ wording.

PR #212, opened 2021-10-05
Make expand row explicit

bonomat:
By removing the onclick handler on the table row, clicking any button in the row does not fire the onclick handler of the row as well.
This has the effect that for expanding the row you need to click the expand button.


thomaseizinger (approved):
Nice! Good fix.

PR #213, opened 2021-10-05
PriceInfo UI improvement

bonomat:
Remove price info field from taker and move price info into create order box for maker.

Imho this looks much nicer. 

image


klochowicz:
the maker side looks nicer indeed, but why remove it on the taker side?

bonomat:
> the maker side looks nicer indeed, but why remove it on the taker side?

The taker should only be concerned about the rate he receives from the maker. 

image

da-kami (approved):

PR #215, opened 2021-10-05
Optimise data structure used for settlement proposals

klochowicz:
New HashMap can store both incoming and outgoing settlement proposals at the
same time and resulted in some code simplification, as there are less types
needed to convey the intention.

Provide a type alias for convenience (this also ensures that both binaries store
settlement proposals in the same way)

### DIFF
--- a/daemon/src/to_sse_event.rs
+++ b/daemon/src/to_sse_event.rs
@@ -1,4 +1,4 @@
-use crate::model::cfd::{OrderId, Role, SettlementProposals};
+use crate::model::cfd::{OrderId, Role, SettlementKind, SettlementProposals};
 use crate::model::{Leverage, Position, TradingPair, Usd};
 use crate::{bitmex_price_feed, model};
 use bdk::bitcoin::{Amount, SignedAmount};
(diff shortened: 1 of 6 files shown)


daemon/src/to_sse_event.rs
```diff
@@ -146,20 +126,11 @@ impl CfdsWithAuxData {
...
-                if proposals.contains_key(&cfd.order.id) {
-                    return SettlementProposalStatus::Outgoing;
-                }
-            }
+    fn settlement_proposal_status(&self, cfd: &model::cfd::Cfd) -> Option<SettlementKind> {
+        if let Some((_, kind)) = self.settlement_proposals.get(&cfd.order.id) {
+            return Some(kind.clone());
         }
-        SettlementProposalStatus::None
+        None
```
thomaseizinger: This could be replaced by a `.map` on the `Option`.

thomaseizinger (approved):

daemon/src/to_sse_event.rs
klochowicz: thanks!  I still sometimes forget about these cool features ;)

changed to:
```rust
       self.settlement_proposals
           .get(&cfd.order.id)
           .map(|(_, kind)| kind.clone())
```

thomaseizinger (approved):

PR #217, opened 2021-10-06
Allow application to run on mainnet or testnet

thomaseizinger:


daemon/src/maker.rs
```diff
@@ -72,6 +68,48 @@ struct Opts {
     /// If enabled logs will be in json format
     #[clap(short, long)]
     json: bool,
+
+    #[clap(subcommand)]
+    network: Network,
```
bonomat, line 73: If there is no default, you might need to adapt the CI and want to update: `start_all.sh`

bonomat (approved):
Nice, we are getting closer 🥳

daemon/src/maker.rs
thomaseizinger: Good catch!

PR #221, opened 2021-10-06
Adds boiler plate for rolling over 

bonomat:
The actual roll over is not yet implemented but I'm afraid of the code changing below me too fast hence a PR adding the boiler plate :)

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -1,6 +1,6 @@
 use crate::auth::MAKER_USERNAME;
 use crate::db::load_all_cfds;
-use crate::model::cfd::SettlementProposals;
+use crate::model::cfd::UpdateCfdProposals;
 use crate::seed::Seed;
 use crate::wallet::Wallet;
 use anyhow::{Context, Result};
@@ -113,8 +113,8 @@ async fn main() -> Result<()> {
 
     let (order_feed_sender, order_feed_receiver) = watch::channel::<Option<Order>>(None);
     let (wallet_feed_sender, wallet_feed_receiver) = watch::channel::<WalletInfo>(wallet_info);
-    let (settlement_feed_sender, settlement_feed_receiver) =
-        watch::channel::<SettlementProposals>(HashMap::new());
+    let (update_cfd_feed_sender, update_cfd_feed_receiver) =
+        watch::channel::<UpdateCfdProposals>(HashMap::new());
 
     let figment = rocket::Config::figment()
         .merge(("databases.maker.url", data_dir.join("maker.sqlite")))
@@ -131,7 +131,7 @@ async fn main() -> Result<()> {
     rocket::custom(figment)
         .manage(order_feed_receiver)
         .manage(wallet_feed_receiver)
-        .manage(settlement_feed_receiver)
+        .manage(update_cfd_feed_receiver)
         .manage(auth_password)
         .manage(quote_updates)
         .attach(Db::init())
@@ -180,7 +180,7 @@ async fn main() -> Result<()> {
                     schnorrsig::PublicKey::from_keypair(SECP256K1, &oracle),
                     cfd_feed_sender,
                     order_feed_sender,
-                    settlement_feed_sender,
+                    update_cfd_feed_sender,
                     maker_inc_connections_address.clone(),
                     monitor_actor_address.clone(),
                     oracle_actor_address,
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -345,6 +345,18 @@ impl fmt::Display for CfdState {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum UpdateCfdProposal {
+    Settlement {
+        proposal: SettlementProposal,
+        direction: SettlementKind,
+    },
+    RollOverProposal {
+        proposal: RollOverProposal,
+        direction: SettlementKind,
+    },
+}
+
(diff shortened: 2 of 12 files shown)


frontend/src/components/cfdtables/CfdTable.tsx
```diff
@@ -215,6 +215,10 @@ function colorSchemaForAction(action: Action): string {
         case Action.ACCEPT_SETTLEMENT:
             return "green";
         case Action.REJECT_SETTLEMENT:
+            return "green";
```
thomaseizinger: Shouldn't this be `red`?

daemon/src/maker.rs
```diff
@@ -113,7 +113,7 @@ async fn main() -> Result<()> {
 
     let (order_feed_sender, order_feed_receiver) = watch::channel::<Option<Order>>(None);
     let (wallet_feed_sender, wallet_feed_receiver) = watch::channel::<WalletInfo>(wallet_info);
-    let (settlement_feed_sender, settlement_feed_receiver) =
+    let (update_cfd_feed_sender, update_cfd_feed_receiver) =
```
thomaseizinger, line 116: These renames better fit into the previous commit if you want to keep them separate.

thomaseizinger (changes requested):
Looks good. The renames are scattered over both commits, would be nice to have that contained in one commit, can also squash both if you don't want to tear it apart.

What I usually do in these situations is: As I notice that I forgot to rename something in an earlier commit, stash everything, amend the rename, pop stash and continue!
You can use the same strategy to fix these:

- `git reset --soft` the last commit
- `git add -p` the rename changes
- `git commit --amend`

Minor bug in the choice of colors I believe.

frontend/src/components/cfdtables/CfdTable.tsx
bonomat: Good catch, that should be green.

daemon/src/maker.rs
bonomat: If you are fine with it then I'm happy to just squash both commits.

bonomat:
We found one issue with this approach (same applies for propose_settlement): anyone can ask for rollover or settlement. 
I'll leave this out of scope and will create a ticket for this later on.

frontend/src/components/cfdtables/CfdTable.tsx
thomaseizinger: It is already green :upside_down_face:

daemon/src/maker.rs
thomaseizinger: Sure!

frontend/src/components/cfdtables/CfdTable.tsx
bonomat: LoL 🙈 I meant it should be red.

daemon/src/model/cfd.rs
```diff
@@ -354,14 +366,27 @@ pub struct SettlementProposal {
     pub maker: Amount,
 }
 
+/// Proposed collaborative settlement
```
thomaseizinger: Copy-paste error.

daemon/src/model/cfd.rs
```diff
@@ -345,6 +345,18 @@ impl fmt::Display for CfdState {
...
 
+#[derive(Debug, Clone)]
+pub enum UpdateCfdProposal {
+    Settlement {
+        proposal: SettlementProposal,
+        direction: SettlementKind,
+    },
+    RollOverProposal {
+        proposal: RollOverProposal,
+        direction: SettlementKind,
```
thomaseizinger, line 356: If you reuse `SettlementKind` here, I guess it should be renamed to something more fitting?

daemon/src/model/cfd.rs
```diff
@@ -354,14 +366,27 @@ pub struct SettlementProposal {
...
+pub struct RollOverProposal {
+    pub order_id: OrderId,
+    pub timestamp: SystemTime,
+    pub term_length: TermLength,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum TermLength {
+    TwentyFourHours,
+}
```
thomaseizinger: Do we have to have this or could we also just imply that we are rolling over by 24h? Having an enumeration here suggests that we introduce different term lengths soon. Is that the intention?

Extending the message with a field is not hard so I think I'd rather keep things lean here and not have unnecessary abstractions.

thomaseizinger (approved):
Thanks!

Few more comments, nothing blocking.

daemon/src/model/cfd.rs
bonomat: It's probably too early for having this configurable.  
I think we need to play around with different strategies but this can still be an implicit strategy.

daemon/src/model/cfd.rs
bonomat: This will change and I'll use an own enum for the RollOverProposal direction. I'll leave it for now.

daemon/src/model/cfd.rs
thomaseizinger: > It's probably too early for having this configurable.
> I think we need to play around with different strategies but this can still be an implicit strategy.

It sounds like you agree! Are we deleting this enum then so the roll-over term is implicit?

bonomat: 😂
Iremoved it in the force push.

PR #222, opened 2021-10-06
Prevent zero-priced CFD offers

klochowicz:
Simplify the UI code and disable the button for posting a new offer if there is
no price set.

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -1,4 +1,3 @@
-import { RepeatIcon } from "@chakra-ui/icons";
 import {
     Box,
     Button,
@@ -7,7 +6,7 @@ import {
     Grid,
     GridItem,
     HStack,
-    IconButton,
+    Switch,
     Tab,
     TabList,
     TabPanel,
@@ -17,7 +16,7 @@ import {
     useToast,
     VStack,
 } from "@chakra-ui/react";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useAsync } from "react-async";
 import { useEventSource } from "react-sse-hooks";
 import { CfdTable } from "./components/cfdtables/CfdTable";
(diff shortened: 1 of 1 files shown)


frontend/src/MakerApp.tsx
```diff
@@ -187,18 +185,6 @@ export default function App() {
     );
 }
 
-function priceToDisplay(hasEnteredPrice: boolean, orderPrice: string, priceInfo: PriceInfo | null) {
-    if (!priceInfo) {
-        return format("0");
-    }
-
-    if (!hasEnteredPrice) {
-        return format((priceInfo.ask * SPREAD).toString());
```
thomaseizinger, line 196: If you delete this, we are no longer calculating a default price using the spread.

frontend/src/MakerApp.tsx
```diff
@@ -187,18 +185,6 @@ export default function App() {
...
-    }
-
-    return format(orderPrice);
```
thomaseizinger, line 199: We are no longer calling `format` on the price now, is that intentional?

thomaseizinger (changes requested):
Two questions.

frontend/src/MakerApp.tsx
klochowicz: there was no default price using the spread. It was displayed, but not used anywhere... that was the bug in question.

klochowicz: is there a way of triggering the "refresh" button when the `priceInfo` becomes available? That would solve the actual problem in a much nicer way IMHO.

frontend/src/MakerApp.tsx
klochowicz: fixed (as per the fixup commit) - added format when the button refreshes.

frontend/src/MakerApp.tsx
```diff
@@ -107,19 +112,18 @@ export default function App() {
...
-                                    if (priceInfo) {
-                                        setOrderPrice((priceInfo.ask * SPREAD).toString());
-                                    }
+                                    setAutoRefresh(false);
                                 }}
+                                value={format(orderPrice)}
                             />
+                            <HStack>
+                                <Switch
+                                    id="auto-refresh"
```
thomaseizinger, line 121: This can go away if you are not using it for attaching a label to the component.

thomaseizinger (approved):

thomaseizinger:
At some point it would be nice to invest some time into UI component tests as well :)

da-kami (approved):
noice!

frontend/src/MakerApp.tsx
klochowicz: I find it still useful for conveying the purpose of the switch (it's self-explanatory now, without going into the details of the callback)

PR #223, opened 2021-10-06
Publish cet

da-kami:
- Handles the Oracle's attestation in the cfd actor, transition according to the knowledge we have about the timelock expiry already.
- Tries to publish the CET if we are in `CetStatus::Ready`, i.e. both the attestation and timelock expiry happened. We try publishing if either event happened, and just print a log in case it's not ready yet.
- Tries to re-publish CET if we are in `PendingCet` upon restart.
- Re-triggers CET monitoring upon startup.
- Extends any state after `ContractSetup` to be able to store the attestation. (see below for ideas to change that)

Things that could be done different:
Currently we are carrying on the attestation through a lot of states - because we cannot just transition the user to `OpenCommitted` because it is a user decision to go for commit, but we have to keep the attestation around once it happened.
To reduce the state complexity, we could store the attestation independent of the state, but associated with the cfd.
This would make things a lot simpler, but we would then always have to go to the database to check if the attestation is already around (which might make other parts more complex).

---

TODO:

- [ ] Test it 😬


da-kami:
@luckysori I pulled the `Attestation` through to all the other states past `ContractSetup` but before `OpenCommitted` - that tackles the problem we discussed earlier, where we might "lose" an attestation if we are in an earlier state. 

I think that should tie everything together, including the "user interaction" to publish the CET - we won't need it, because OpenCommitted will just handle that for us. Not totally sure about that though, would be great if we can think it through together tomorrow.


Not super happy with having the `Attestation` creeping through the states, but I think it should be complete and should work 😬 - we should be able to refactor this to store relevant `Attestation`s separately (related to cfds in the db). But that might have other downsides. Did not open that rabbit hole yet :)

da-kami:
And of course we need a rebase... will do that later/tomorrow :D

daemon/src/db.rs
```diff
@@ -411,6 +411,92 @@ pub async fn load_all_cfds(conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<
...
 
+/// Loads all CFDs with the latest state as the CFD state
+pub async fn load_cfds_by_oracle_event_id(
+    oracle_event_id: OracleEventId,
+    conn: &mut PoolConnection<Sqlite>,
+) -> anyhow::Result<Vec<Cfd>> {
+    let rows = sqlx::query!(
+        r#"
+        select
+            cfds.id as cfd_id,
```
bonomat: I was a bit confused by this `cfd_id` as on line 444 there is also a `cfd_id` but it's a different one (on line 444 `cfd_id` is coming from `cfd_states`). Imho this one here is not used as on line 439 you use `cfd_states.cfd_id = cfds.id` . Can you remove it please if I'm right?

daemon/src/db.rs
```diff
@@ -411,6 +411,92 @@ pub async fn load_all_cfds(conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<
...
+        from cfds as cfds
+        inner join orders as orders on cfds.order_id = orders.id
+        inner join cfd_states as cfd_states on cfd_states.cfd_id = cfds.id
+        where cfd_states.state in (
+            select
+              state
+              from cfd_states
+            where cfd_id = cfds.id
+            order by id desc
+            limit 1
```
bonomat, line 443: So the latest state is the one with the highest id? We might want to fix this as this asks for troubles 😬

daemon/src/model/cfd.rs
```diff
@@ -243,6 +247,17 @@ pub enum CfdState {
...
 
+    PendingCet {
+        common: CfdStateCommon,
+        dlc: Dlc,
+        cet_status: CetStatus,
+    },
+
+    Closed {
+        common: CfdStateCommon,
+    },
```
bonomat: Can you add a comment above the state what it represents please?

daemon/src/monitor.rs
```diff
@@ -21,11 +22,19 @@ pub struct StartMonitoring {
     pub params: MonitorParams,
 }
 
+#[derive(Clone)]
+pub struct Cet {
```
bonomat, line 27: @luckysori : look, I found a Cet 😅

daemon/src/maker_cfd.rs
```diff
@@ -551,7 +551,7 @@ impl Actor {
     async fn handle_oracle_attestation(&mut self, attestation: oracle::Attestation) -> Result<()> {
         tracing::debug!(
             "Learnt latest oracle attestation for event: {}",
-            attestation.id
+            attestation.id.0
```
thomaseizinger: In these scenarios, I'd favor implementing `fmt::Display` instead of reaching into the newtype. Takes 10 seconds and moves into the right direction. If you don't want to write it by hand, also happy to use something like https://jeltef.github.io/derive_more/derive_more/display.html although more macros means longer compile times in the long run.

bonomat (commented):
I only had a quick glance over it so far.

daemon/src/housekeeping.rs
```diff
@@ -54,5 +54,13 @@ pub async fn rebroadcast_transactions(
         tracing::info!("Commit transaction published on chain: {}", txid);
     }
 
+    for cfd in cfds.iter().filter(|cfd| Cfd::is_pending_cet(cfd)) {
+        // Double question-mark OK because if we are in PendingCet we must have been Ready before
```
thomaseizinger, line 58: Take a note of https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/ when you have a couple of minutes.

You can avoid such "problems" by combining the check with a type transition. If `is_pending_cet` would not just return you a `bool` but an `Option<>`, you could use `filter_map` here and automatically iterate over only the pending CETs.

The blogpost is relevant because what you are doing here is essentially _parsing_ the pending CETs out of a list of CFDs.

daemon/src/model/cfd.rs
```diff
@@ -607,8 +650,66 @@ impl Cfd {
...
                 }
             }
+            CfdStateChangeEvent::OracleAttestation(attestation) => match self.state.clone() {
+                CfdState::Open { dlc, .. } => CfdState::Open {
+                    common: CfdStateCommon {
+                        transition_timestamp: SystemTime::now(),
+                    },
+                    dlc,
+                    attestation: Some(attestation),
+                },
```
thomaseizinger, line 679: We desperately need constructor functions for these states that automatically initialize `common`. This boilerplate makes it really hard to review what is going on.

daemon/src/model/cfd.rs
```diff
@@ -683,6 +783,72 @@ impl Cfd {
...
+            | CfdState::PendingCommit { .. } => {
+                return Ok(Err(NotReadyYet));
+            }
+            _ => bail!("Cannot publish CET in state {}", self.state.clone()),
+        };
+
+        let cets = dlc
+            .cets
+            .get(&attestation.id.0)
+            .context("Unable to find oracle event id within the cets of the dlc")?;
```
thomaseizinger: Can we push such low-level code down into the respective types like `Dlc` please?

daemon/src/model/cfd.rs
```diff
@@ -683,6 +783,72 @@ impl Cfd {
...
+            .iter()
+            .find(|Cet { range, .. }| range.contains(&attestation.price))
+            .context("Price out of range of cets")?;
+
+        let oracle_attestations = attestation.scalars;
+
+        let mut decryption_sk = oracle_attestations[0];
+        for oracle_attestation in oracle_attestations[1..*n_bits].iter() {
+            decryption_sk.add_assign(oracle_attestation.as_ref())?;
+        }
```
thomaseizinger, line 842: This sort of low-level code should be on `Attestation`.

daemon/src/model/cfd.rs
```diff
@@ -683,6 +783,72 @@ impl Cfd {
...
+
+        let counterparty_sig = encsig.decrypt(&decryption_sk)?;
+        let counterparty_pubkey = dlc.identity_counterparty;
+
+        let signed_cet = finalize_spend_transaction(
+            cet.clone(),
+            &dlc.commit.2,
+            (our_pubkey, our_sig),
+            (counterparty_pubkey, counterparty_sig),
+        )?;
```
thomaseizinger, line 863: This can easily be on `Dlc` unless I am mistaken.

daemon/src/model/cfd.rs
```diff
@@ -714,12 +884,18 @@ impl Cfd {
     }
 }
 
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+#[error("The cfd is not committed yet")]
+pub struct NotReadyYet;
```
thomaseizinger, line 905:
```suggestion
pub struct NotCommittedYet;
```

daemon/src/model/cfd.rs
```diff
@@ -1073,6 +1257,8 @@ pub struct Dlc {
     /// The fully signed lock transaction ready to be published on chain
     pub lock: (Transaction, Descriptor<PublicKey>),
     pub commit: (Transaction, EcdsaAdaptorSignature, Descriptor<PublicKey>),
-    pub cets: HashMap<String, Vec<(Transaction, EcdsaAdaptorSignature, RangeInclusive<u64>)>>,
+    // TODO: Range + number of digits (usize) could be represented as Digits similar to what we do
+    // in the protocol lib
+    pub cets: HashMap<String, Vec<Cet>>,
```
thomaseizinger: `s/String/OracleEventId`?

daemon/src/monitor.rs
```diff
@@ -21,11 +22,19 @@ pub struct StartMonitoring {
...
+    range: RangeInclusive<u64>,
+    n_bits: usize,
+}
+
 #[derive(Clone)]
 pub struct MonitorParams {
     lock: (Txid, Descriptor<PublicKey>),
     commit: (Txid, Descriptor<PublicKey>),
-    cets: HashMap<String, Vec<(Txid, Script, RangeInclusive<u64>)>>,
+    cets: HashMap<String, Vec<Cet>>,
```
thomaseizinger: `s/String/OracleEventId`?

daemon/src/monitor.rs
```diff
@@ -490,6 +506,31 @@ impl MonitorParams {
...
+                            txid: tx.txid(),
+                            script: script_pubkey.clone(),
+                            range: range.clone(),
+                            n_bits: *n_bits,
+                        },
+                    )
+                    .collect::<Vec<_>>(),
+            )
+        })
+        .collect::<HashMap<_, _>>()
```
thomaseizinger: Nit: Could omit this because it will be inferred by the return type.

daemon/src/monitor.rs
```diff
@@ -490,6 +506,31 @@ impl MonitorParams {
...
+                cets.iter()
+                    .map(
+                        |model::cfd::Cet {
+                             tx, range, n_bits, ..
+                         }| Cet {
+                            txid: tx.txid(),
+                            script: script_pubkey.clone(),
+                            range: range.clone(),
+                            n_bits: *n_bits,
+                        },
```
thomaseizinger, line 527: Could implement `from` or a different constructor to remove the low-level bits from this.

thomaseizinger (approved):
Nice work, thanks for tackling this!

Couldn't spot any mistakes, left a few comments on how we could improve things (and reduce some of the duplication). In a nutshell: We need to push more code down onto domain objects. I wouldn't worry too much about duplication of "glue-code" as it is happening within the actor.

In the long run, we are going to merge both binaries so most of this duplication will go away.

daemon/src/db.rs
thomaseizinger: What do you want to fix about this? The ID is auto-incremented so that highest one IS always the latest state.

daemon/src/db.rs
```diff
@@ -411,6 +411,92 @@ pub async fn load_all_cfds(conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<
...
 }
 
+/// Loads all CFDs with the latest state as the CFD state
+pub async fn load_cfds_by_oracle_event_id(
+    oracle_event_id: OracleEventId,
+    conn: &mut PoolConnection<Sqlite>,
+) -> anyhow::Result<Vec<Cfd>> {
+    let rows = sqlx::query!(
+        r#"
+        select
```
thomaseizinger, line 419: Why no capslock ...

image

daemon/src/monitor.rs
da-kami: Yeah, I introduced that because clippy started complaining about type complexity, and the mapping between the model's "Cet" represenation and the monitor actor's "Cet" representation just became too cumbersome. Can give it a different name as well, but I think it fits.

daemon/src/model/cfd.rs
da-kami: That is not entirely  true... We had named it `NotCommittedYet` - but this variant may also be used if it *is* committed, but either the attestation or the timelock expiry is still missing.

da-kami:
It's funny most of the discussions are about the SQL statement where we already have exactly this statement two times in the code and I just keep copying and adapting it :D

daemon/src/model/cfd.rs
thomaseizinger: Then the error message needs to change!

daemon/src/db.rs
da-kami: I'm trying to leave the Matrix!!!

daemon/src/model/cfd.rs
da-kami: Oh yes, can we tackle that together later?
I looked into how to achieve this in a generic way (given that we have variants with different data inside...) but could not figure it out yet.

daemon/src/model.rs
```diff
@@ -108,3 +108,12 @@ pub struct WalletInfo {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct OracleEventId(pub String);
+
+impl Display for OracleEventId {
+    fn fmt(
+        &self,
+        f: &mut Formatter<'_>,
+    ) -> cfd_protocol::secp256k1_zkp::bitcoin_hashes::_export::_core::fmt::Result {
```
thomaseizinger: Wrong import :grimacing: 

Only import this from `std` please.

da-kami: stupid Clion trying to be smart 😅

daemon/src/monitor.rs
da-kami: Not totally trivial because of the `script_pubkey`. We can try to untangle this in a follow up.

daemon/src/monitor.rs
da-kami: https://github.com/comit-network/hermes/pull/223/commits/71ac42a97e8835bdf90564c49389dcfa4ad83b0b

daemon/src/model.rs
```diff
@@ -108,3 +109,9 @@ pub struct WalletInfo {
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct OracleEventId(pub String);
+
+impl Display for OracleEventId {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
thomaseizinger, line 117: I usually refer to all 3 symbols (`Display`, `Formatter` and `Result`) via `fmt::`. That results in only one import (`use fmt`) and is unambiguous in what you are referring to.

daemon/src/model/cfd.rs
```diff
@@ -243,19 +239,34 @@ pub enum CfdState {
         cet_status: CetStatus,
     },
 
-    /// The CFD contract's refund transaction was published but it not final yet
-    MustRefund {
+    /// The CET was published on chain but is not final yet
+    ///
+    /// This state applies to taker and maker.
+    /// This state is needed, because otherwise the user does not get any feedback.
```
thomaseizinger, line 245: Can we not extend `CetStatus` and get rid of this one?

thomaseizinger (approved):

daemon/src/model/cfd.rs
da-kami: I went with the `Pending` state because that's what we do for other transactions. Would at least keep some consistency until we have a better idea of how we want to refactor the model. Happy to tackle this in follow-up(s).

PR #227, opened 2021-10-07
Handle collab settlement response from the maker

klochowicz:
- store TakerId for proposals to be able to respond to specific taker about
  settlement proposal when the user takes action
- send the collab settlement response (accept or reject) from the maker
- handle rejection (remove from the map), CFD state in the UI returns to "Open"

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -85,7 +85,8 @@ pub struct Actor {
     setup_state: SetupState,
     latest_announcements: Option<BTreeMap<OracleEventId, oracle::Announcement>>,
     oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
-    current_pending_proposals: UpdateCfdProposals,
+    // Maker needs to also store TakerId to be able to send a reply back
+    current_pending_proposals: HashMap<OrderId, (UpdateCfdProposal, TakerId)>,
 }
 
 enum SetupState {
@@ -126,10 +127,33 @@ impl Actor {
         }
     }
 
+    /// Send pending proposals for the purposes of UI updates.
+    /// Filters out the TakerIds, as they are an implementation detail inside of
+    /// the actor
     fn send_pending_proposals(&self) -> Result<()> {
-        Ok(self
-            .update_cfd_feed_sender
-            .send(self.current_pending_proposals.clone())?)
+        Ok(self.update_cfd_feed_sender.send(
+            self.current_pending_proposals
+                .iter()
+                .map(|(order_id, (update_cfd, _))| (*order_id, (update_cfd.clone())))
+                .collect(),
+        )?)
+    }
+
+    fn get_taker_id_of_proposal(&self, order_id: &OrderId) -> Result<TakerId> {
+        let (_, taker_id) = self
+            .current_pending_proposals
+            .get(order_id)
+            .context("Could not find proposal for given order id")?;
+        Ok(*taker_id)
+    }
+
+    /// Removes a proposal and updates the update cfd proposals' feed
+    fn remove_pending_proposal(&mut self, order_id: &OrderId) -> Result<()> {
+        if self.current_pending_proposals.remove(order_id).is_none() {
+            anyhow::bail!("Could not find proposal with order id: {}", &order_id)
+        }
+        self.send_pending_proposals()?;
+        Ok(())
     }
 
     async fn handle_new_order(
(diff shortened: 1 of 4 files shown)


klochowicz:
After a discussion with @thomaseizinger and @da-kami today, we decided that the taker will instead infer the response of the maker by monitoring for txid inside a time window (if txid was seen, then the maker must have accepted the settlement proposal, if it was not, it means that it didn't).

klochowicz:
reopened the pull request (still need to rebase), as the settlement will continue after the initial agreement from the maker. Without this initial agreement, we would need to proactively send the signature to the maker, and then either always fallback to non-collab closure after a time window, or *trust* the maker that they will not use the signature when it's convenient for them at any later point before the Cfd lapses.

daemon/src/maker_cfd.rs
```diff
@@ -117,10 +118,34 @@ impl Actor {
...
     }
 
+    /// Send settlement proposals for the purposes of UI updates.
+    /// Filters out the TakerIds, as they are an implementation detail inside of
+    /// the actor
     fn send_current_settlement_proposals(&self) -> Result<()> {
+        Ok(self.settlements_feed_sender.send(
+            self.current_settlement_proposals
+                .iter()
+                .map(|x| (*x.0, (x.1 .0.clone(), x.1 .1.clone())))
```
thomaseizinger: This would be more readable if you would destruct the `x` tuple directly!

daemon/src/maker_cfd.rs
```diff
@@ -117,10 +118,34 @@ impl Actor {
...
+
+    fn get_taker_id_of_proposal(&self, order_id: &OrderId) -> Result<TakerId> {
         Ok(self
-            .settlements_feed_sender
-            .send(self.current_settlement_proposals.clone())?)
+            .current_settlement_proposals
+            .get_key_value(order_id)
+            .context("Could not find proposal for given order id")?
+            .1
+             .2)
```
thomaseizinger: Assigning and destructing would make this also more readable.

daemon/src/maker_cfd.rs
```diff
@@ -117,10 +118,34 @@ impl Actor {
...
+                .collect(),
+        )?)
+    }
+
+    fn get_taker_id_of_proposal(&self, order_id: &OrderId) -> Result<TakerId> {
         Ok(self
-            .settlements_feed_sender
-            .send(self.current_settlement_proposals.clone())?)
+            .current_settlement_proposals
+            .get_key_value(order_id)
```
thomaseizinger: Why not just `get` if you are throwing away the key anyway?

thomaseizinger (approved):
Few nits in regards to readability, otherwise good!

daemon/src/maker_cfd.rs
klochowicz: thank you, this made things sooo much nicer!!!

daemon/src/maker_cfd.rs
klochowicz: indeed, changed to `get`

daemon/src/maker_cfd.rs
klochowicz: thank you! indeed much more readable.

daemon/src/maker_cfd.rs
```diff
@@ -516,47 +555,57 @@ impl Actor {
...
-        // TODO: Initiate the settlement
 
-        self.current_pending_proposals
-            .remove(&order_id)
-            .context("Could not find proposal for given order id")?;
-        self.send_pending_proposals()?;
+        let taker_id = self.get_taker_id_of_proposal(&order_id)?;
+
+        // TODO: Initiate the settlement - should we start calculating the
+        // signature here?
```
thomaseizinger, line 562: I don't think so. The maker can reactively do this upon receiving the signature from the taker. I would do that in a completely different message and simply assert that upon receiving that message, we have actually agreed to a settlement.

thomaseizinger (approved):

daemon/src/maker_cfd.rs
```diff
@@ -193,35 +217,50 @@ impl Actor {
...
-            UpdateCfdProposal::Settlement {
-                proposal,
-                direction: SettlementKind::Incoming,
-            },
+            (
+                UpdateCfdProposal::Settlement {
+                    proposal,
+                    direction: SettlementKind::Incoming,
+                },
+                taker_id,
```
bonomat, line 236: We solved it slightly different and stored the `taker_id` inside the `SettlementKind::Incoming`. But I guess your approach is also nice :)

bonomat (approved):
👍

daemon/src/maker_cfd.rs
klochowicz: I'll get rid of the comment in the subsequent PR, where I'll be actually sending the signature.

daemon/src/maker_cfd.rs
klochowicz: I should have predicted I'll need this, it would have been easier for us both ;D

PR #234, opened 2021-10-07
Fix contract setup

da-kami:


da-kami:
https://github.com/comit-network/hermes/issues/235

daemon/src/maker.rs
```diff
@@ -149,7 +150,10 @@ async fn main() -> Result<()> {
         auth_password
     );
 
-    let oracle = schnorrsig::KeyPair::new(SECP256K1, &mut rand::thread_rng()); // TODO: Fetch oracle public key from oracle.
+    // TODO: Actually fetch it from Olivia
```
thomaseizinger, line 153: I think hard-coding is completely fine. If we fetch it from Olivia, we make ourselves vulnerable to MITM attacks.

thomaseizinger (approved):

daemon/src/maker.rs
da-kami: ok, then I close #235 for now

PR #237, opened 2021-10-07
Oracle actor startup behavior

da-kami:


daemon/src/maker.rs
```diff
@@ -265,6 +265,8 @@ async fn main() -> Result<()> {
                     cfds,
                 )));
 
+                oracle_actor_address.do_send(oracle::Sync).unwrap();
```
thomaseizinger: `do_send` will block the entire executor thread. Better use `send().await`.

thomaseizinger (changes requested):
Thanks!

daemon/src/maker.rs
da-kami: woops, thanks! will change :)

thomaseizinger (approved):

PR #240, opened 2021-10-07
Implement rollover protocol in daemons

### LINKED ISSUE #136: Taker to initiate CFD rollover
The taker should be able to say that he wants to roll over the contract for another iterations. 

The maker should only accept this if the taker should not have gotten liquidated already.

Proposal: 
Initially we can make this a manual step, e.g. the user has to click a button to say: `roll over`. Eventually I'd see this as an automated action. Whenever the user comes online, he rolls over to the next iteration.

luckysori:
Fix #136.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -37,6 +37,7 @@ pub struct MonitorParams {
     commit: (Txid, Descriptor<PublicKey>),
     cets: HashMap<OracleEventId, Vec<Cet>>,
     refund: (Txid, Script, u32),
+    revoked_commits: Vec<(Txid, Script)>,
 }
 
 pub struct Sync;
@@ -110,7 +111,7 @@ where
                             actor.monitor_refund_finality(&params,cfd.order.id);
                         }
                         CetStatus::OracleSigned(attestation) => {
-                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.address.script_pubkey()), attestation, cfd.order.id)?;
+                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.maker_address.script_pubkey()), attestation, cfd.order.id)?;
                             actor.monitor_commit_cet_timelock(&params, cfd.order.id);
                             actor.monitor_commit_refund_timelock(&params, cfd.order.id);
                             actor.monitor_refund_finality(&params,cfd.order.id);
@@ -120,7 +121,7 @@ where
                             actor.monitor_refund_finality(&params,cfd.order.id);
                         }
                         CetStatus::Ready(attestation) => {
-                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.address.script_pubkey()), attestation, cfd.order.id)?;
+                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.maker_address.script_pubkey()), attestation, cfd.order.id)?;
                             actor.monitor_commit_refund_timelock(&params, cfd.order.id);
                             actor.monitor_refund_finality(&params,cfd.order.id);
                         }
@@ -163,6 +164,7 @@ where
         self.monitor_commit_cet_timelock(params, order_id);
         self.monitor_commit_refund_timelock(params, order_id);
         self.monitor_refund_finality(params, order_id);
+        self.monitor_revoked_commit_transactions(params, order_id);
     }
 
     fn monitor_lock_finality(&mut self, params: &MonitorParams, order_id: OrderId) {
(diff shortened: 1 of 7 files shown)


daemon/src/monitor.rs
```diff
@@ -110,7 +111,7 @@ where
                             actor.monitor_refund_finality(&params,cfd.order.id);
                         }
                         CetStatus::OracleSigned(attestation) => {
-                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.address.script_pubkey()), attestation, cfd.order.id)?;
+                            actor.monitor_cet_finality(map_cets(dlc.cets, dlc.maker_address.script_pubkey()), attestation, cfd.order.id)?;
```
thomaseizinger, line 114: Is this line what https://github.com/comit-network/hermes/issues/238 is referring to? Can we just fix this by monitoring for both here?

daemon/src/monitor.rs
```diff
@@ -236,6 +238,18 @@ where
         Ok(())
     }
 
+    fn monitor_revoked_commit_transactions(&mut self, params: &MonitorParams, order_id: OrderId) {
+        for revoked_commit_tx in params.revoked_commits.iter() {
+            self.awaiting_status
+                .entry((revoked_commit_tx.0, revoked_commit_tx.1.clone()))
+                .or_default()
+                .push((
+                    ScriptStatus::with_confirmations(0),
```
thomaseizinger, line 247: I think you may want to monitor for mem-pool instead.

daemon/src/setup_contract.rs
```diff
@@ -221,12 +226,291 @@ pub async fn new(
...
+            .iter()
+            .map(|txin| TxIn {
+                previous_output: txin.previous_output,
+                script_sig: txin.script_sig.clone(),
+                sequence: txin.sequence,
+                witness: vec![],
+            })
+            .collect(),
+        output: dlc.lock.0.output.clone(),
+    };
```
thomaseizinger: I think this would be easier to understand if you would make a copy of `dlc.lock` and iterate over the inputs to clear the `witness`.

daemon/src/wire.rs
```diff
@@ -270,3 +281,78 @@ impl From<CfdTransactions> for Msg1 {
...
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct RollOverMsg1 {
+    pub commit: EcdsaAdaptorSignature,
+    pub cets: HashMap<String, Vec<(RangeInclusive<u64>, EcdsaAdaptorSignature)>>,
+    pub refund: Signature,
+}
+#[derive(Debug, Serialize, Deserialize)]
```
thomaseizinger, line 332: Nit: Missing newline.

thomaseizinger (approved):
Nice work, looks solid.

I haven't tested it :)

daemon/src/monitor.rs
bonomat: Yes it is. I'd think this should work.

daemon/src/monitor.rs
bonomat: Would we miss it if an "attacker" manages to get the transaction included in the block without sending it to the mempool?

bonomat:
@luckysori : the honor of merging is all yours ;)

daemon/src/monitor.rs
thomaseizinger: That is a good question but the answer is luckily of course not!

We compare desired and actual script status using `>=`. I am fairly certain that it works but you are welcome to add a test that compares two `ScriptStatus` against each other.

I implemented it via `PartialOrd` but it would probably actually be better to have a custom method on `ScriptStatus` to make sure it cannot be used wrongly by the application code, thus reducing the risk of a big if all we have is the comparison unit test.

thomaseizinger:
@luckysori @bonomat 

The first commit in this PR does not pass the formatting checks. Please make sure at least formatting and clippy passes on each commit, otherwise rebases and git bisect are unnecessarily hard.

Perhaps set yourself some pre-commit hook if you find that it is otherwise too hard to keep track of these things.

luckysori:
> The first commit in this PR does not pass the formatting checks. Please make sure at least formatting and clippy passes on each commit, otherwise rebases and git bisect are unnecessarily hard.
> 
> Perhaps set yourself some pre-commit hook if you find that it is otherwise too hard to keep track of these things.

Oops, my bad. I didn't realise we had more than 1 commit in this PR when I pushed it in.

I was gonna suggest to maybe run CI on every commit, but CI is probably too slow for that. I'll set up a pre-commit hook like the good old days of `comit-rs`.

PR #243, opened 2021-10-07
Remove the spread

da-kami:
It does not make sense to put a spread on top of the price; this just results in a shitty UX, because the user always starts with a loss.
We need proper interest on top of the amount, i.e. a fixed rate that the maker always gets that is independent of the price.

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -28,8 +28,6 @@ import { Cfd, intoCfd, intoOrder, Order, PriceInfo, StateGroupKey, WalletInfo } 
 import Wallet from "./components/Wallet";
 import { CfdSellOrderPayload, postCfdSellOrderRequest } from "./MakerClient";
 
-const SPREAD = 1.03;
-
 export default function App() {
     let source = useEventSource({ source: "/api/feed", options: { withCredentials: true } });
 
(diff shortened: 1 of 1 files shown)


frontend/src/MakerApp.tsx
```diff
@@ -47,7 +45,7 @@ export default function App() {
 
     useEffect(() => {
         if (autoRefresh && priceInfo) {
-            setOrderPrice((priceInfo.ask * SPREAD).toFixed(2).toString());
+            setOrderPrice((priceInfo.ask).toFixed(2).toString());
```
thomaseizinger, line 48: Unnecessary params.

thomaseizinger (approved):

bonomat (approved):
Interesting though. The loss should be ~3% and not more. Maybe we should check double check if our profit calculation is using the right numbers.

The better approach would be to add an explicit _origination fee_ which the taker has to pay to the maker.

da-kami:
> Interesting though. The loss should be ~3% and not more. Maybe we should check double check if our profit calculation is using the right numbers.
> 
> The better approach would be to add an explicit _origination fee_ which the taker has to pay to the maker.

The thing is: The origination fee seems to be part of the protocol (i.e. contract setup, payout curve, ...). This might be more complicated than it has to be.

The idea of the spread was to just pack this on top of the price to reduce the risk for the maker. 
The multiplicator also makes sense, i.e. if the user has a leverage of 5 then the price "change" will have a way higher impact. So, the calculation is fine, we pack the spread on top of the price the oracle report (because essentially there is the same price source behind). And the leverage is related to the price...

After some more thought: We might actually just want to keep the spread for now to keep it simple; but I would reduce it to `0.3%` instead of `3%` :)

da-kami:
replaced by https://github.com/comit-network/hermes/pull/245

bonomat:
Ah right, 3% spread and 5x makes a big loss 🙈

PR #245, opened 2021-10-07
Spread of `1%` instead of `3%`

da-kami:


frontend/src/MakerApp.tsx
```diff
@@ -28,7 +28,7 @@ import { Cfd, intoCfd, intoOrder, Order, PriceInfo, StateGroupKey, WalletInfo }
 import Wallet from "./components/Wallet";
 import { CfdSellOrderPayload, postCfdSellOrderRequest } from "./MakerClient";
 
-const SPREAD = 1.03;
+const SPREAD = 1.003;
```
bonomat:
```suggestion
const SPREAD = 1.01;
```

bonomat (approved):
Counter proposal: 1% and let's change the default leverage to 2 to reduce the capital requirements for the maker :) 

(Note: 0.3% is a guaranteed loss for a maker who wants to hedge with our current setup as on other platforms there is a broker fee and interest rate)

PR #249, opened 2021-10-07
First cet trial fixes

da-kami:
Note: Just discussed with @thomaseizinger that the Oracle actor should be changed to be able to handle multiple terms.
So, instead of the oracle actor sending out the current 24 hours to the cfd actor, the cfd actor should be able to request events for a time interval (i.e. different terms).

Changing that should also allow us to test more easily, because we can use a very short term.


thomaseizinger (approved):
ACK the first commit, drop the 2nd one please to avoid conflicts.

PR #252, opened 2021-10-08
Rework oracle actor communication

da-kami:


daemon/src/taker_cfd.rs
```diff
@@ -271,13 +269,11 @@ impl Actor {
...
-            .clone()
-            .context("No oracle announcements available")?;
-        let offer_announcement = offer_announcements
-            .get(&cfd.order.oracle_event_id)
-            .context("Order's announcement not found in current oracle announcements")?;
+        let offer_announcement = self
+            .oracle_actor
+            .send(oracle::GetAnnouncement(cfd.order.oracle_event_id.clone()))
+            .await?
+            .context("Announcement not found")?;
```
thomaseizinger: Perhaps we should do `with_context` here and print out the actual ID.

thomaseizinger (approved):

da-kami:
@bonomat do we want to have a `8` hour or `24` hour `term`?
With the changes we introduce the oracle actor still keeps a 24 hour window, but that is not exposed to the cfd actor. The cfd actor can decide to use a shorter term.

thomaseizinger (approved):

daemon/src/oracle.rs
```diff
@@ -69,7 +68,7 @@ where
         }
 
         Self {
-            latest_announcements: None,
+            latest_announcements: HashMap::new(),
```
luckysori, line 68: :question: Is it still worth regularly caching latest announcements now that we are not pushing them out to the CFD actor? We could simply go to the oracle with the `event_id`(s) that we constructed/received?

luckysori (approved):
Makes sense, nice.

daemon/src/oracle.rs
thomaseizinger: The idea was to decouple IO errors from the actual contract-setup.

luckysori: Okay, but keep in mind that this results in making 24 HTTP requests to `olivia` every minute as opposed to 1 per CFD.

If we want caching perhaps we should only update this internal state every hour and load a wider range of events to cover our bases.

da-kami:
added commit to cleanup/fix URL and event_id handling

thomaseizinger (approved):

bonomat:
> @bonomat do we want to have a `8` hour or `24` hour `term`? With the changes we introduce the oracle actor still keeps a 24 hour window, but that is not exposed to the cfd actor. The cfd actor can decide to use a shorter term.

24 for now.

PR #254, opened 2021-10-08
Minor cleanups in `oracle::Actor`

thomaseizinger:


daemon/src/oracle.rs
```diff
@@ -16,22 +16,14 @@ use std::collections::{HashMap, HashSet};
 const OLIVIA_EVENT_TIME_FORMAT: &[FormatItem] =
     format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]");
 
-pub struct Actor<CFD, M>
-where
-    CFD: xtra::Handler<Attestation>,
-    M: xtra::Handler<Attestation>,
```
luckysori, line 22: :+1: I'm surprised to see that we don't need these trait bounds everywhere. I swear we tried to elide them before and it didn't work.

luckysori (approved):

daemon/src/oracle.rs
thomaseizinger: I moved the methods into dedicated impl blocks.

luckysori: We also tried that, but we clearly did something wrong.

da-kami (approved):

PR #256, opened 2021-10-08
2nd cet trial fixes

da-kami:
I wish I could have had the brainpower and dicipline to do the last commit in multiple atomic commits. 
Quite a nightmare "fixing" all these things. Too many moving parts and complex code. 

@thomaseizinger maybe you can help me split it up?

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -172,10 +172,9 @@ pub async fn insert_new_cfd_state_by_order_id(
         .await
         .context("loading latest state failed")?;
 
-    // make sure that the new state is different than the current one to avoid that we save the same
-    // state twice
     if mem::discriminant(&latest_cfd_state_in_db) == mem::discriminant(&new_state) {
-        tracing::warn!(
+        // Since we have states where we add information this happens quite frequently
+        tracing::trace!(
             "Same state transition for cfd with order_id {}: {}",
             order_id,
             latest_cfd_state_in_db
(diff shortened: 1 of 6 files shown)


da-kami:
I'm cleaning the last commit up atm to make it more clear what has been changed and why.

da-kami:
I noticed that there are sveral things around `cfd.handle(...)` that are not optimal:

- We distinguish three cases: The cfd transitions to a new state, the cfd remains in the same state and the state transition fails. This now results in returning `Result<Option<CfdState>>`. The code around these transitions is fairly complex and hard to understand/reason about. This might need an overall refactor. 
  - Sidenote: The "Early exit" was needed for "early" final states. We had cases where we would keep transitioning from final to final state, which caused weird side effects and unwanted state appending in the db. 

- `cfd.handle(...)` in combination with the `cfd.[...]_tx(...)` functions: We need guards for the transaction construction, but those guards are basically a duplication of the transition logic in `cfd.handle`. We could try to combine the logic so `handle` returns transaction to be published. (For the `housekeeping` this might require a bit more, because that does not use `handle` at the moment.)

thomaseizinger (approved):
Thanks for the fixes and structuring them into nice patches!

thomaseizinger:
> cfd.handle(...) in combination with the cfd.[...]_tx(...) functions: We need guards for the transaction construction, but those guards are basically a duplication of the transition logic in cfd.handle. We could try to combine the logic so handle returns transaction to be published. (For the housekeeping this might require a bit more, because that does not use handle at the moment.)

I think the long run, something like this is a good way forward. Might be able to generalize from "Transaction" to just "Action" that needs to be taken.

luckysori (approved):

PR #257, opened 2021-10-08
Sign and broadcast collaborative settlement

klochowicz:
Taker sends his signature to the maker, which allows the maker to sign and
broadcast the settled transaction.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -15,6 +15,7 @@ use crate::{maker_inc_connections, monitor, oracle, setup_contract, wire};
 use anyhow::{bail, Context as _, Result};
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
+use cfd_protocol::secp256k1_zkp::Signature;
 use futures::channel::mpsc;
 use futures::{future, SinkExt};
 use std::collections::HashMap;
@@ -91,6 +92,10 @@ pub struct Actor {
     oracle_actor: Address<oracle::Actor<Actor, monitor::Actor<Actor>>>,
     // Maker needs to also store TakerId to be able to send a reply back
     current_pending_proposals: HashMap<OrderId, (UpdateCfdProposal, TakerId)>,
+    // TODO: Persist instead of using an in-memory hashmap for resiliency?
+    // (not a huge deal, as in the worst case taker will fall back to
+    // noncollaborative settlement)
+    current_agreed_proposals: HashMap<OrderId, (SettlementProposal, TakerId)>,
 }
 
 enum SetupState {
@@ -136,6 +141,7 @@ impl Actor {
             roll_over_state: RollOverState::None,
             oracle_actor,
             current_pending_proposals: HashMap::new(),
+            current_agreed_proposals: HashMap::new(),
         }
     }
 
@@ -166,6 +172,21 @@ impl Actor {
         }
         self.send_pending_proposals()?;
         Ok(())
+    }
+
+    fn get_settlement_proposal(&self, order_id: OrderId) -> Result<(SettlementProposal, TakerId)> {
+        let (update_proposal, taker_id) = self
+            .current_pending_proposals
+            .get(&order_id)
+            .context("have a proposal that is about to be accepted")?;
+
+        let proposal = match update_proposal {
+            UpdateCfdProposal::Settlement { proposal, .. } => proposal,
+            UpdateCfdProposal::RollOverProposal { .. } => {
+                anyhow::bail!("did not expect a rollover proposal");
+            }
+        };
+        Ok((proposal.clone(), *taker_id))
     }
 
     async fn handle_new_order(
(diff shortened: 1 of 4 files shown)


daemon/src/setup_contract.rs
```diff
@@ -241,6 +241,29 @@ pub async fn new(
     })
 }
 
+// Create a close transaction based on the current contract and a settlement proposals
+pub fn close_transaction(
```
thomaseizinger: I don't think this should be in `contract_setup`. It has nothing to do with setting up a contract :)

I would suggest making this a function on `Dlc`.

daemon/src/taker_cfd.rs
```diff
@@ -337,7 +351,25 @@ impl Actor {
...
+        let mut conn = self.db.acquire().await?;
+
+        let cfd = load_cfd_by_order_id(order_id, &mut conn).await?;
+        let dlc = cfd.open_dlc().context("CFD was in wrong state")?;
+
+        let proposal = self.get_settlement_proposal(order_id)?;
+        let (_tx, sighash) = setup_contract::close_transaction(&dlc, proposal)
+            .context("Unable to collaborative close transaction")?;
+
+        let sig_taker = SECP256K1.sign(&sighash, &dlc.identity);
```
thomaseizinger: This would fit well into `close_transaction`.

daemon/src/maker_cfd.rs
```diff
@@ -246,6 +268,67 @@ impl Actor {
...
+            &dlc.identity,
+        ));
+
+        let (_, lock_desc) = dlc.lock;
+        let spend_tx = cfd_protocol::finalize_spend_transaction(
+            tx,
+            &lock_desc,
+            (maker_pk, sig_maker),
+            (dlc.identity_counterparty, sig_taker),
+        )?;
```
thomaseizinger: All of this onto `Dlc` please.

thomaseizinger: `Dlc` + taker's `Signature` -> `Transaction`

daemon/src/maker_cfd.rs
```diff
@@ -168,6 +175,21 @@ impl Actor {
...
     }
 
+    fn get_settlement_proposal(&self, order_id: OrderId) -> Result<(SettlementProposal, TakerId)> {
+        let (update_proposal, taker_id) = self
+            .current_pending_proposals
+            .get(&order_id)
+            .context("have a proposal that is about to be accepted")?;
+
+        let proposal = match update_proposal {
+            UpdateCfdProposal::Settlement { proposal, .. } => proposal,
```
thomaseizinger, line 184: Can we just store the `proposal` in the hashmap instead? Then I think we can avoid this boilerplate.

thomaseizinger (changes requested):
Thanks!

A few suggestions that should make things a bit more succinct :)

daemon/src/setup_contract.rs
klochowicz: great idea.

daemon/src/maker_cfd.rs
klochowicz: that was the case before rollover support was added - shall we just have 2 separate hashmaps here instead and zip them only for the feed purposes? that would probably make it easier to see what's going on inside the cfd actors.

daemon/src/maker_cfd.rs
klochowicz: I moved it to the `impl Dlc` block, but opted for the following signature:

```rust
    pub fn finalize_spend_transaction(
        &self,
        (close_tx, own_sig): (Transaction, Signature),
        counterparty_sig: Signature,
    ) -> Result<Transaction> {
```
Slightly more verbose, but the steps are more clear, as the function does only one thing.

klochowicz:
> Thanks!
> 
> A few suggestions that should make things a bit more succinct :)

Thanks for the suggestions, they made it not only succinct, but allowed to hide protocol details from the actor itself, which is great! 👍

daemon/src/maker_cfd.rs
klochowicz: @thomaseizinger there are basically 3 choices:
- this boilerplate
- have 2 hashmaps internally (one for each enum variant) and zip them in `send_pending_proposals()`
- manage hashmap per variant in Rocket (one for proposals, one for rollovers)

None of solutions is perfect. Happy to change this if we have a clear decision, preferably in a separate cleanup PR as it's not the core of this change.

thomaseizinger: > that was the case before rollover support was added - shall we just have 2 separate hashmaps here instead and zip them only for the feed purposes? that would probably make it easier to see what's going on inside the cfd actors.

I don't quite understand. You are introducing this hashmap here, yeah? Why can you not design it exactly for what you need?

Even outside of this, I think we should have two data structures. If we can avoid error cases by modelling the problem differently, then I think that is the superior solution.

Happy for this to be tackled in a follow-up!

daemon/src/maker_cfd.rs
```diff
@@ -630,16 +698,15 @@ impl Actor {
...
             .do_send_async(maker_inc_connections::TakerMessage {
                 taker_id,
                 command: TakerCommand::NotifySettlementAccepted { id: order_id },
             })
             .await?;
 
+        self.current_agreed_proposals
+            .insert(order_id, self.get_settlement_proposal(order_id)?);
         self.remove_pending_proposal(&order_id)
             .context("accepted settlement")?;
```
thomaseizinger, line 712: Removing the proposal from the hashmap would return you the owned instance in case it is present. I think this would be cleaner if we would just:

1. Always try to remove it
2. Fail if it wasn't there
3. If it was there, move it into the other `HashMap`.

thomaseizinger: Alternatively, we could have a single `HashMap` with a data type that can internally transition from Proposed -> Accepted.

thomaseizinger (approved):

daemon/src/maker_cfd.rs
klochowicz: again, the other hashmap has a different type so we can't really use that without a boilerplate :/ non-trivial to change, as I can't use a single HashMap easily (the first one is used for the rollovers as well).
I'll have a go at it in the follow-up.

daemon/src/maker_cfd.rs
klochowicz: I'm happy to separate this HashMap into a separate ones for settlements and for rollovers. I think it will be a cleaner solution - I'll post a follow-up tomorrow :)

PR #263, opened 2021-10-08
Signet support and bug fix

bonomat:
Valid hours are only 0-23, which means we need to roll over to the next day.

Unfortunately I could not find a public electrum server for Signet so maybe we want to run one ourselves. I tested with a node locally and it works.


thomaseizinger (approved):
Thanks for adding a test!

Running a signet blockexplorer sounds like a good idea if we want to continue supporting our stuff on signet!

PR #267, opened 2021-10-11
Make HTTP listen address configurable

bonomat:
Maker and taker by default only listened on localhost:XXXX. Given our architecture it is very likely that the taker and maker might run on remote machine.


thomaseizinger (approved):
Thanks!

PR #269, opened 2021-10-11
Fix monitoring

### LINKED ISSUE #260: `Refunded` instead of `Closed`
Once a CET is confirmed we always get `CetFinality` and `RefundFinality`, this is very strange:

```
2021-10-08 21:41:37  INFO Bitcoin transaction reached monitoring target txid=677d5e79318aa6e8fe8fa3c3de299bb25a8ce530bb1a27fabb67108d81facf28 target=confirmed with 1 blocks current=confirmed with 1 blocks
2021-10-08 21:41:37  INFO Cfd state change event Monitor(CetFinality(OrderId(5d0d71aa-efbc-48a1-baf8-853f288fe14d)))
2021-10-08 21:41:37  INFO Bitcoin transaction reached monitoring target txid=543662c6d0de53ed7a3f71cb016e53daf3c1bee7a3fcb6c525b8917aa8189c6a target=confirmed with 1 blocks current=confirmed with 1 blocks
2021-10-08 21:41:37  INFO Cfd state change event Monitor(RefundFinality(OrderId(5d0d71aa-efbc-48a1-baf8-853f288fe14d)))
```

depending on which message from monitor actor is processed in cfd actor first we transition either to `Closed` or `Refunded` - but actually always are in a `Close` scenario (i.e. CET published and confirmed, Refund tx is never sent on chain).

da-kami:
In this protocol we *can* actually have transactions with the same script, hence, we have to handle the `GetHistoryRes` by `Txid`.
To make this easier to understand we group the `Vec<Vec<GetHistoryRes>>` by `Txid` into `HashMap<Txid, GetHistroyRes>`.
Note: The assumption is, that only `GetHistroyRes` is returned for the `Txid`+`Script` combination.

fixes #260

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -284,7 +284,7 @@ where
                 let txid = response.tx_hash;
                 let script = match txid_to_script.get(&txid) {
                     None => {
-                        tracing::error!(
+                        tracing::trace!(
                             "Could not find script in own state for txid {}, ignoring",
                             txid
                         );
@@ -382,6 +382,12 @@ where
                         });
 
                     tracing::trace!("{} subscriptions reached their monitoring target, {} remaining for this script", reached_monitoring_target.len(), remaining.len());
+
+                    // TODO: When reaching finality of a final tx (CET, refund_tx,
+                    // collaborate_close_tx) we have to remove the remaining "competing"
+                    // transactions. This is not critical, but when fetching
+                    // `GetHistoryRes` by script we can have entries that we don't care about
+                    // anymore.
 
                     if remaining.is_empty() {
                         occupied.remove();
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -383,7 +383,11 @@ where
 
                     tracing::trace!("{} subscriptions reached their monitoring target, {} remaining for this script", reached_monitoring_target.len(), remaining.len());
 
-                    occupied.insert(remaining);
+                    if remaining.is_empty() {
+                        occupied.remove();
+                    } else {
+                        occupied.insert(remaining);
+                    }
 
                     for (target_status, event) in reached_monitoring_target {
                         tracing::info!(%txid, target = %target_status, current = %status, "Bitcoin transaction reached monitoring target");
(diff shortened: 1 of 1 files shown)


daemon/src/monitor.rs
```diff
@@ -267,12 +267,37 @@ where
...
             .context("Failed to get script histories")?;
 
-        self.update_state(latest_block_height, histories).await?;
+        let mut histories_grouped_by_txid = HashMap::new();
+        for history in histories {
+            for response in history {
+                let txid = response.tx_hash;
+                let script = match txid_to_script.get(&txid) {
+                    None => {
+                        tracing::error!(
```
thomaseizinger: Use `warn` here? `error` seems a bit drastic.

da-kami:
Tested, but no test yet 🤪
I'd say we can still merge the fix and add the test on top of it (without logic changes).

daemon/src/monitor.rs
```diff
@@ -298,30 +323,29 @@ where
...
-                                u32::try_from(last.height).expect("we checked that height is > 0"),
+        // 1. Decide new status based on script history
+        let new_status = self
+            .awaiting_status
+            .iter()
+            .map(|(key, _old_status)| {
+                let new_script_status = match histories.get(key) {
+                    None => ScriptStatus::Unseen,
+                    Some(history_entry) => {
+                        if history_entry.height <= 0 {
```
thomaseizinger, line 334: This could be a match arm guard.

daemon/src/monitor.rs
```diff
@@ -298,30 +323,29 @@ where
...
+                            ScriptStatus::InMempool
+                        } else {
+                            ScriptStatus::Confirmed(Confirmed::from_inclusion_and_latest_block(
```
thomaseizinger, line 337: As a clean-up, I think it would make sense to get rid of this `Confirmed` struct and just inline the entire logic into `ScriptStatus`.

thomaseizinger (approved):
Thanks for investigating and fixing this!

thomaseizinger (approved):
Good work!

daemon/src/monitor.rs
```diff
@@ -383,6 +383,12 @@ where
 
                     tracing::trace!("{} subscriptions reached their monitoring target, {} remaining for this script", reached_monitoring_target.len(), remaining.len());
 
+                    // TODO: When reaching finality of a final tx (CET, refund_tx,
+                    // collaborate_close_tx) we have to remove the remaining "competing"
+                    // transactions. This is not critical, but when fetching
+                    // `GetHistoryRes` by script we can have entries that we don't care about
+                    // anymore.
```
thomaseizinger, line 390: Make this an issue please.

thomaseizinger (approved):

daemon/src/monitor.rs
da-kami: https://github.com/comit-network/hermes/issues/276

PR #270, opened 2021-10-11
Some oracle fixes

thomaseizinger:
- Only fetch announcements we don't already have
- Await futures in random order
- Use the same code for computing the next attestation event

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -4,7 +4,7 @@ use crate::model::OracleEventId;
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use cfd_protocol::secp256k1_zkp::{schnorrsig, SecretKey};
-use futures::stream::FuturesOrdered;
+use futures::stream::FuturesUnordered;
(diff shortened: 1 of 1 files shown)


daemon/src/oracle.rs
```diff
@@ -103,9 +107,11 @@ where
                 Result::<_, anyhow::Error>::Ok((event_id, announcement))
             })
             .collect::<FuturesOrdered<_>>()
-            .try_collect()
+            .try_collect::<HashMap<_, _>>()
             .await?;
 
+        self.latest_announcements.extend(new_announcements); // FIXME: This results in linear memory growth.
```
luckysori, line 113: :sob:

luckysori (approved):

daemon/src/oracle.rs
thomaseizinger: I proposed a fix for this [here](https://github.com/comit-network/hermes/issues/275).

da-kami (approved):
Thanks for the fix and for writing up #275

PR #272, opened 2021-10-11
Transaction IDs and payout in the UI

da-kami:
- [x] Profit based on attested price.
- [x] Show the payout and transaction ids in the UI
image

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -271,7 +271,7 @@ pub enum CfdState {
     ///
     /// This state applies to taker and maker.
     /// This is a final state.
-    Refunded { common: CfdStateCommon },
+    Refunded { common: CfdStateCommon, dlc: Dlc },
 
     /// The Cfd was in a state that could not be continued after the application got interrupted
     ///
@@ -310,17 +310,16 @@ impl Attestation {
 
         let txid = cet.tx.txid();
 
+        let our_script_pubkey = match role {
+            Role::Maker => dlc.maker_address.script_pubkey(),
+            Role::Taker => dlc.taker_address.script_pubkey(),
+        };
         let payout = cet
             .tx
             .output
             .iter()
             .find_map(|output| {
-                (output.script_pubkey
-                    == match role {
-                        Role::Maker => dlc.maker_address.script_pubkey(),
-                        Role::Taker => dlc.taker_address.script_pubkey(),
-                    })
-                .then(|| Amount::from_sat(output.value))
+                (output.script_pubkey == our_script_pubkey).then(|| Amount::from_sat(output.value))
             })
             .unwrap_or_default();
 
@@ -335,6 +334,14 @@ impl Attestation {
 
     pub fn price(&self) -> Usd {
         Usd(Decimal::from(self.price))
+    }
+
+    pub fn txid(&self) -> Txid {
+        self.txid
+    }
+
+    pub fn payout(&self) -> Amount {
+        self.payout
     }
 }
 
@@ -742,18 +749,15 @@ impl Cfd {
                     }
                 }
                 monitor::Event::RefundFinality(_) => {
-                    if let MustRefund { .. } = self.state.clone() {
-                    } else {
-                        tracing::debug!(
-                            "Was in unexpected state {}, jumping ahead to Refunded",
-                            self.state
-                        );
-                    }
+                    let dlc = self
+                        .dlc()
+                        .context("No dlc available when reaching refund finality")?;
 
                     Refunded {
                         common: CfdStateCommon {
                             transition_timestamp: SystemTime::now(),
                         },
+                        dlc,
                     }
                 }
                 monitor::Event::CetFinality(_) => {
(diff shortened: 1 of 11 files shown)


daemon/src/model/cfd.rs
```diff
@@ -284,17 +287,50 @@ pub enum CfdState {
...
+            .tx
+            .output
+            .iter()
+            .find_map(|output| {
+                (output.script_pubkey
+                    == match role {
+                        Role::Maker => dlc.maker_address.script_pubkey(),
+                        Role::Taker => dlc.taker_address.script_pubkey(),
+                    })
+                .then(|| Amount::from_sat(output.value))
```
thomaseizinger: Can we break this up a bit so it is easier to read? I have to turn my head side ways for this :joy:

daemon/src/model/cfd.rs
```diff
@@ -284,17 +287,50 @@ pub enum CfdState {
...
+            })
+            .unwrap_or_default();
```
thomaseizinger, line 324: `or_default`? What is the reasoning here?

daemon/src/to_sse_event.rs
```diff
@@ -272,6 +308,111 @@ fn to_cfd_state(
...
+    let refund_label = "refund".to_string();
+
+    let (txs, payout) = match state {
+        PendingOpen {
+            dlc, attestation, ..
+        }
+        | Open {
+            dlc, attestation, ..
+        } => (
+            vec![TxUrl::new(dlc.lock.0.txid(), network, lock_label)],
```
thomaseizinger: Perhaps make 4 constructor functions that set the label to a specific value:

```rust
TxUrl::lock(....)
```

Then you can also pass the entire `dlc` in because the function knows which transaction to pull out.

daemon/src/to_sse_event.rs
```diff
@@ -272,6 +308,111 @@ fn to_cfd_state(
...
+        .find(|output| {
+            output.script_pubkey
+                == match role {
+                    Role::Taker => dlc.taker_address.script_pubkey(),
+                    Role::Maker => dlc.maker_address.script_pubkey(),
+                }
+        })
+        .map(|output| Amount::from_sat(output.value))
+        .unwrap_or_default()
+}
```
thomaseizinger: Can we have this on `Dlc` please.

daemon/src/to_sse_event.rs
```diff
@@ -32,6 +34,35 @@ pub struct Cfd {
...
+#[derive(Debug, Clone, Serialize)]
+pub struct CfdDetails {
+    tx_url_list: Vec<TxUrl>,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc::opt")]
+    payout: Option<Amount>,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct TxUrl {
+    pub label: String,
```
thomaseizinger: Why not an `enum`? This is not Python.

daemon/src/to_sse_event.rs
```diff
@@ -32,6 +34,35 @@ pub struct Cfd {
...
+    pub url: String,
+}
+
+impl TxUrl {
+    pub fn new(txid: Txid, network: Network, label: String) -> Self {
+        Self {
+            label,
+            url: match network {
+                Network::Bitcoin => format!("https://mempool.space/tx/{}", txid),
```
thomaseizinger, line 59: Could use the `Url` type here.

daemon/src/to_sse_event.rs
```diff
@@ -32,6 +34,35 @@ pub struct Cfd {
...
+
+    pub details: CfdDetails,
+}
+
+#[derive(Debug, Clone, Serialize)]
+pub struct CfdDetails {
+    tx_url_list: Vec<TxUrl>,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc::opt")]
+    payout: Option<Amount>,
+}
```
thomaseizinger, line 46: Is this type necessary? Can we not just have `tx_urls` and `payout` in the `Cfd` type?

frontend/src/components/cfdtables/CfdTable.tsx
```diff
@@ -89,6 +92,27 @@ export function CfdTable(
...
+                    return (
+                        <Box>
+                            <VStack>
+                                {txs}
+                                {details.payout && <Box>Payout: {details.payout}</Box>}
+                            </VStack>
+                        </Box>
+                    );
+                },
+                isNumeric: true,
```
thomaseizinger: What does this do?

frontend/src/components/cfdtables/CfdTable.tsx
```diff
@@ -236,19 +260,18 @@ function renderRowSubComponent(row: Row<Cfd>) {
     // TODO: I would show additional information here such as txids, timestamps, actions
```
thomaseizinger: Resolved now? :)

thomaseizinger (approved):
Thanks!

Couple of suggestions but overall mergable!

daemon/src/to_sse_event.rs
da-kami: will use a builder, that's even neater

daemon/src/model/cfd.rs
da-kami: I'll add a comment - reasoning is: if there is no output then we were liquidated, hence we default to `ZERO` which is the default.

daemon/src/model/cfd.rs
da-kami: :D

frontend/src/components/cfdtables/CfdTable.tsx
da-kami: it copies my paste!

daemon/src/to_sse_event.rs
da-kami: That just makes the constructions unnecessarily clunky - we have `expects` there then. I think the `String` is good enough for now, can tackle in a follow up if we think it's important.

daemon/src/to_sse_event.rs
da-kami: Tried to remove it, but due to the data table being quite a mess it is hard, so I would push this in a follow up :/

PR #273, opened 2021-10-11
Build for aarch64 which is needed for RPi4

bonomat:
Adds more build targets and publishes docker containers. 

This was tested on my personal fork and resulted in the following container which can run on an RPi4: 

```
docker run ghcr.io/bonomat/hermes-maker-aarch64-unknown-linux-gnu:latest
```

A few notes which might come up: 
- The naming of the containers can be improved, please someone with sed knowledge give it a try :)
- I was not able to use the github action `docker/build-push-action`. This moves the image build into a tmp folder which I was not able to escape. Hence, I was not able to copy the binaries into the image
- I was not able to combine the containers into a single one because we have parallel builds. Apparently this should be possible though (reference: [manifest](https://www.docker.com/blog/multi-platform-docker-builds/)). 
- I was also not able to use alpine because of the different libc6 version


bonomat:
Ignore the last commit...

.github/workflows/build-release-binary.yml
```diff
@@ -63,6 +71,12 @@ jobs:
           sudo apt-get update
           sudo apt-get install gcc-arm-linux-gnueabihf
 
+      - name: Install compiler for armhf arch
```
klochowicz: IMHO better wording would be: "Install compiler for aarch64 arch (armv8)"

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:bullseye-slim
```
klochowicz: perhaps add a comment that we need a relatively new image as we depend on libc6 2.29?

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ARG DIRECTORY
+ARG BINARY
+RUN echo "Copying $DIRECTORY/$BINARY into container"
+
+COPY target/$DIRECTORY/release/$BINARY $BINARY
```
klochowicz: I think you can easily get away with a single argument, just name the binary arbitrarily, like
`program` or `binary`:

`COPY ${BINARY_PATH} program`

`RUN chmod +x program`

etc

this way it would be actually easier to use, as we're just copying the path of the binary from cargo root to include it, without any special tricks.

docker/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM debian:bullseye-slim
+
+ARG DIRECTORY
+ARG BINARY
+RUN echo "Copying $DIRECTORY/$BINARY into container"
```
klochowicz: nit: prefer wrapping variables inside `{}` , like: `${PATH}` (otherwise it would break on a space).

klochowicz (approved):
good work! 🥇

.github/workflows/build-release-binary.yml
```diff
@@ -129,3 +154,29 @@ jobs:
...
+
+      - name: Build and publish docker image
+        if: matrix.os != 'windows-latest' && matrix.os != 'macos-latest'
+        run: |
+          docker buildx build \
+            --push \
+            --platform=${{ matrix.docker_platforms }} \
+            --tag ghcr.io/${{ github.actor }}/hermes-${{ matrix.bin }}-${{ matrix.target }} \
+            --build-arg BINARY_PATH=target/${{ matrix.target }}/release/${{ matrix.bin }} \
+            --file docker/Dockerfile .
```
thomaseizinger: I think this `.` here means the entire directory will be sent as context to the docker daemon. Should we add a `.dockerignore` to ignore all of that?

I've never worked with `build-arg` before so don't know how they interact.

docker/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM debian:bullseye-slim
+
```
thomaseizinger: Any benefit of this file being in a sub-directory? Can we not just have this in the root of the repo? That would follow the convention which means not having to specify the file within the action on line 182.

.github/workflows/build-release-binary.yml
```diff
@@ -129,3 +154,29 @@ jobs:
           asset_path: ./${{ steps.create-archive-name.outputs.archive }}
           asset_name: ${{ steps.create-archive-name.outputs.archive }}
           asset_content_type: application/gzip
+
+      - name: Set up QEMU
+        if: matrix.os != 'windows-latest' && matrix.os != 'macos-latest'
+        uses: docker/setup-qemu-action@v1
```
thomaseizinger, line 160: What do we need QEMU for?

.github/workflows/build-release-binary.yml
```diff
@@ -129,3 +154,29 @@ jobs:
...
+        uses: docker/login-action@v1.10.0
+        if: matrix.os != 'windows-latest' && matrix.os != 'macos-latest'
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and publish docker image
+        if: matrix.os != 'windows-latest' && matrix.os != 'macos-latest'
+        run: |
```
thomaseizinger, line 176: Instead of making all of these conditional, would it work to:

1. Have a separate job after this
2. With a dependency on the first one
3. Download the artifacts from the above job

The benefits I see is that:

- Build matrix gets simpler because all the new variables are only concerned with docker, which would be in a new job.
- The job declaration gets easier because we don't need to exclude it for macos and windows.

thomaseizinger (commented):
Good effort!

Couple of questions :)

.github/workflows/build-release-binary.yml
bonomat: `.dockerignore` is probably a good idea :)

docker/Dockerfile
bonomat: definitely doable.

.github/workflows/build-release-binary.yml
bonomat: We need to build images for different architectures (armv7 and aarch64).

.github/workflows/build-release-binary.yml
bonomat: I would assume so but I haven't tried this. 
Sounds like an "easy first ticket" @klochowicz ?

docker/Dockerfile
bonomat: done

.github/workflows/build-release-binary.yml
bonomat: I gave it a try with ignoring everything but target. Can probably be more optimized.

.github/workflows/build-release-binary.yml
bonomat: I'll postpone this to the future #285

thomaseizinger: > I'll postpone this to the future #285

Thanks for creating a ticket. It is a bit hard to understand out of context so some re-wording would be nice :)

thomaseizinger (approved):
Thanks!

That will do for now.

bonomat:
bors r+

🤞that it works here as well...

bonomat:
bors retry

PR #279, opened 2021-10-11
More oracle actor fixes

thomaseizinger:
- Re-organize oracle module a bit
- Parallelize fetching of announcements via self messages
- Add error message to JSON deserializer
- Don't fail attestation sync just because it is not attested yet

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -276,8 +276,10 @@ async fn main() -> Result<()> {
                     cfds,
                 )));
 
-                // use `.send` here to ensure we only continue once the update was processed
-                oracle_actor_address.send(oracle::Sync).await.unwrap();
+                oracle_actor_address
+                    .do_send_async(oracle::Sync)
+                    .await
+                    .unwrap();
 
                 let listener_stream = futures::stream::poll_fn(move |ctx| {
                     let message = match futures::ready!(listener.poll_accept(ctx)) {
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -165,7 +165,10 @@ where
                     }
                 };
 
-                let attestation = res.json::<Attestation>().await?;
+                let attestation = res
+                    .json::<Attestation>()
+                    .await
+                    .with_context(|| format!("Failed to decode body for event {}", event_id))?;
 
                 self.cfd_actor_address
                     .clone()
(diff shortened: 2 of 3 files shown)


daemon/src/maker.rs
```diff
@@ -276,8 +276,10 @@ async fn main() -> Result<()> {
                     cfds,
                 )));
 
-                // use `.send` here to ensure we only continue once the update was processed
-                oracle_actor_address.send(oracle::Sync).await.unwrap();
+                oracle_actor_address
```
luckysori, line 279: :question: Should we get ahead of the likely problem caused by this and create an issue? I would expect it to cause an error when trying to create an order before the attestations are loaded.

daemon/src/oracle.rs
```diff
@@ -192,6 +204,13 @@ impl<CFD: 'static, M: 'static> xtra::Handler<GetAnnouncement> for Actor<CFD, M>
     }
 }
 
+#[async_trait]
+impl<CFD: 'static, M: 'static> xtra::Handler<NewAnnouncementFetched> for Actor<CFD, M> {
+    async fn handle(&mut self, msg: NewAnnouncementFetched, _ctx: &mut xtra::Context<Self>) {
+        self.latest_announcements.insert(msg.id, msg.announcement);
```
luckysori, line 220: :question: I know this works, but why do we need to send a message to ourselves instead of just directly inserting the new announcements after completing the for loop in `update_latest_announcements`? This seems a bit too indirect.

luckysori (approved):
Nice, thank you!

daemon/src/maker.rs
thomaseizinger: Announcements are still loaded on startup. A user would have to be really quick for them to not be available and even then, they can still retry and it will likely succeed.

daemon/src/oracle.rs
thomaseizinger: I can only hold one mutable reference at a time, meaning if I want to run all fetching tasks in parallel, I need to centralize the state update in a message handler.

luckysori: Okay, that makes sense. I didn't realise we were spawning tasks to fetch the announcements in parallel, which is what the title of the commit says :facepalm:

daemon/src/maker.rs
luckysori: Yeah, it likely won't be a problem for users. I would expect manual testing to be influenced by this though, but if we are happy to just retry then it's okay.

PR #281, opened 2021-10-12
Monitor CETs based on a script pubkey in the transaction

### LINKED ISSUE #238: Monitor for CET publishment for both addresses
When monitoring for CET finality we only watch out for one party's address: 

https://github.com/comit-network/hermes/blob/159240cc9fc4e424c71cc57c70231dbc2ab8a47c/daemon/src/monitor.rs#L113

We should however monitor for both addresses as some CETs may only have one address (e.g. those which pay all the funds to one of the parties and _liquidate_ the other one). 

This is a bug: we might miss the finality of some CETs and stagnate.

luckysori:
Fixes #238.

Instead of passing in a script which may or may not be part of a CET (because some CETs only pay to one party), we take the script pubkey from an output of the transaction itself.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -24,7 +24,7 @@ pub struct StartMonitoring {
 }
 
 #[derive(Clone)]
-pub struct Cet {
+struct Cet {
     txid: Txid,
     script: Script,
     range: RangeInclusive<u64>,
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -24,14 +24,6 @@ pub struct StartMonitoring {
 }
 
 #[derive(Clone)]
-struct Cet {
-    txid: Txid,
-    script: Script,
-    range: RangeInclusive<u64>,
-    n_bits: usize,
-}
-
-#[derive(Clone)]
 pub struct MonitorParams {
     lock: (Txid, Descriptor<PublicKey>),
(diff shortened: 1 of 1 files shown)


daemon/src/monitor.rs
```diff
@@ -576,7 +575,7 @@ fn map_cets(
                              tx, range, n_bits, ..
                          }| Cet {
                             txid: tx.txid(),
-                            script: script_pubkey.clone(),
+                            script: tx.output[0].script_pubkey.clone(),
                             range: range.clone(),
                             n_bits: *n_bits,
```
thomaseizinger: Can we have a `From` for this now?

thomaseizinger: What is `tx.output[0]`?

thomaseizinger (approved):

daemon/src/monitor.rs
luckysori: > What is tx.output[0]?

Either the maker's or the taker's output in the CET.

> Can we have a From for this now?

I'll take a look.

da-kami (approved):

daemon/src/monitor.rs
luckysori: > Can we have a From for this now?

Done with https://github.com/comit-network/hermes/pull/281/commits/70107f4d8131663a665b7fcc01f8bf74a44f4383.

PR #282, opened 2021-10-12
Only fetch announcements when needed

### LINKED ISSUE #275: Refine how the oracle actor fetches announcements
Currently, we just fetch 24h of announcements ahead of time and sync in a regular interval. This is quite arbitrary.

The `cfd::Actor`s actually know, which attestations they will need:

- The maker knows at the time they create the offer
- The taker knows at the time they receive an offer

In both cases, we could ping the oracle actor asynchronously with the relevant event ID which would trigger the actor to fetch the announcement in the background (and perhaps persist permanently?).

The announcement is only needed at the time of contract setup which has to be manually triggered by the user. By that time, the announcement will already have been fetched and is available for the actor to retrieve. If it is not, the user can retry up until it is.

This solves multiple problems:

- Reduces bandwidth because we don't arbitrarily fetch announcements from the oracle
- Reduces memory footprint as we only fetch those announcements that we will care about (or at least very likely)
- Reduces complexity by not having to regularly "sync" with the oracle for future announcements (we will still have to check for attested events

luckysori:
Fixes #275.

I think this is what you meant with #275, @thomaseizinger. It does result in immediately deleting some code you wrote in #279. I do think we'll end up reintroducing it once we actually use more than one oracle event per CFD, but it seems unnecessary to keep it around until then. Let me know if you disagree.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -590,7 +590,7 @@ impl Actor {
             .with_context(|| format!("Announcement {} not found", cfd.order.oracle_event_id))?;
 
         self.oracle_actor
-            .do_send_async(oracle::MonitorEvent {
+            .do_send_async(oracle::MonitorAttestation {
                 event_id: offer_announcement.id.clone(),
             })
             .await?;
@@ -781,7 +781,7 @@ impl Actor {
             .await?;
 
         self.oracle_actor
-            .do_send_async(oracle::MonitorEvent {
+            .do_send_async(oracle::MonitorAttestation {
                 event_id: announcement.id.clone(),
             })
             .await?;
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -198,6 +198,10 @@ impl Actor {
         let oracle_event_id =
             oracle::next_announcement_after(time::OffsetDateTime::now_utc() + Order::TERM)?;
 
+        self.oracle_actor
+            .do_send_async(oracle::FetchAnnouncement(oracle_event_id.clone()))
+            .await?;
+
         let order = Order::new(
             price,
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):
Yeah nice!

Thank you! There is a slight caveat that we may want to consider:

It is possible that fetching the announcement fails at the time we are told to do so which means we will never have it later and the entire situation cannot be fixed without receiving / making another order.

We could work around that by making processing of the `FetchAnnouncement` message infallible (only store it in a hashmap) and fetch it on the next `Sync` trigger.

luckysori:
> It is possible that fetching the announcement fails at the time we are told to do so which means we will never have it later and the entire situation cannot be fixed without receiving / making another order.
> 
> We could work around that by making processing of the `FetchAnnouncement` message infallible (only store it in a hashmap) and fetch it on the next `Sync` trigger.

@thomaseizinger: does https://github.com/comit-network/hermes/pull/282/commits/d0df9f00618f62ce301545321772a6dfc4fec14e do the trick? We get to reintroduce your code too, which means that it'll be simpler to support multiple oracle events per CFD too.

thomaseizinger:
> > It is possible that fetching the announcement fails at the time we are told to do so which means we will never have it later and the entire situation cannot be fixed without receiving / making another order.
> > We could work around that by making processing of the `FetchAnnouncement` message infallible (only store it in a hashmap) and fetch it on the next `Sync` trigger.
> 
> @thomaseizinger: does [d0df9f0](https://github.com/comit-network/hermes/commit/d0df9f00618f62ce301545321772a6dfc4fec14e) do the trick? We get to reintroduce your code too, which means that it'll be simpler to support multiple oracle events per CFD too.

Yeah that will do, thanks!

PR #286, opened 2021-10-12
Monitor for close finality

klochowicz:


klochowicz:
Note: I ran into the following issue when trying to broadcast transaction:

```
2021-10-12 17:11:42 DEBUG Maker accepts a settlement proposal order_id=0d3195d5-04de-485c-b99a-429792d98b25
2021-10-12 17:11:42  INFO Settlement proposal got accepted order_id=0d3195d5-04de-485c-b99a-429792d98b25
2021-10-12 17:11:42  INFO Cfd state change event ProposalSigned(Transaction { version: 2, lock_time: 0, input: [TxIn { previous_output: OutPoint { txid: 6297abb2c5c1bc1b4334c437404568ce889b94fb66c16c6621ca6f12665dd30f, vout: 0 }, script_sig: Script(), sequence: 4294967295, witness: [] }], output: [TxOut { value: 17336, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 d826dce8a3e1a502c23c0bcd3bf348a1a8a53cfd) }, TxOut { value: 8365, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 3acff941a0d04bca71816cd1a3261218a64d3417) }] })
2021-10-12 17:11:42  INFO Taker 73d8780a-4c81-437e-b450-6ad25b559611 initiated collab settlement for order 0d3195d5-04de-485c-b99a-429792d98b25 by sending their signature
2021-10-12 17:11:43 ERROR Message handler failed: Broadcasting spend transaction: Electrum(Protocol(String("sendrawtransaction RPC error: {\"code\":-26,\"message\":\"min relay fee not met, 166 < 168\"}")))
```

@luckysori , @bonomat , perhaps I'm not using the protocol right?

bonomat:
> @luckysori , @bonomat , perhaps I'm not using the protocol right?

No, it's not your fault 😅

Do you still have the data? If so, we should try to get the raw transaction and inspect it.

klochowicz:
> > @luckysori , @bonomat , perhaps I'm not using the protocol right?
> 
> No, it's not your fault 😅
> 
> Do you still have the data? If so, we should try to get the raw transaction and inspect it.

It happened yesterday arvo, so I should still have the data. I'll PM you how to extract the things that might be useful.

cfd_protocol/src/protocol.rs
```diff
@@ -387,9 +387,9 @@ impl PartialEq for Announcement {
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct Payout {
-    digits: interval::Digits,
-    maker_amount: Amount,
-    taker_amount: Amount,
+    pub digits: interval::Digits,
+    pub maker_amount: Amount,
+    pub taker_amount: Amount,
```
luckysori: :wrench: I would prefer it if we kept these private and just added getters where needed. That way we can ensure that the `struct` can only be built through `generate_payouts` and that it can never be mutated (outside of this module).

daemon/src/maker_cfd.rs
```diff
@@ -294,16 +294,21 @@ impl Actor {
...
 
-        self.wallet
+        let txid = self
+            .wallet
             .try_broadcast_transaction(spend_tx)
             .await
             .context("Broadcasting spend transaction")?;
+        tracing::info!("Close transaction published with txid {}", txid);
+
+        cfd.handle(CfdStateChangeEvent::CloseSent)?;
```
luckysori: :wrench: We should add another wrapper around `handle` specific to `CfdStateChangeEvent::CloseSent` like we've done with `handle_commit` and others.

daemon/src/model/cfd.rs
```diff
@@ -226,6 +227,8 @@ pub enum CfdState {
         common: CfdStateCommon,
         dlc: Dlc,
         attestation: Option<Attestation>,
+        // TODO: Unsure about whether we need to keep this if we're pending commit
+        collaborative_close: Option<(Transaction, SystemTime)>,
```
luckysori: :wrench: I think we should remove this field. If we ever transition back to `Open` from this state collaborative close is then possible again. And we should never be able to transition from `Open` to `PendingCommit` with this field set to `Some(_)`, so no information should be lost.

daemon/src/model/cfd.rs
```diff
@@ -251,6 +254,17 @@ pub enum CfdState {
         cet_status: CetStatus,
     },
 
+    /// The collaborative close transaction was published but it not final yet
```
luckysori: :wrench: Little typo.

```suggestion
    /// The collaborative close transaction was published but is not final yet.
```

daemon/src/model/cfd.rs
```diff
@@ -805,6 +851,51 @@ impl Cfd {
...
+                    attestation,
+                    collaborative_close: Some(collaborative_close),
+                } => CfdState::PendingClose {
+                    common,
+                    dlc,
+                    attestation,
+                    collaborative_close,
+                },
+                _ => bail!(
+                    "Cannot transition to PendingCet because of unexpected state {}",
```
luckysori: :wrench: Little copy-paste error.

```suggestion
                    "Cannot transition to PendingClose because of unexpected state {}",
```

daemon/src/monitor.rs
```diff
@@ -82,14 +82,20 @@ where
                     actor.cfds.insert(cfd.order.id, params.clone());
                     actor.monitor_all(&params, cfd.order.id);
                 }
-                CfdState::Open { dlc, .. } | CfdState::PendingCommit { dlc, .. } => {
+                CfdState::Open { dlc, collaborative_close, .. } | CfdState::PendingCommit { dlc, collaborative_close, .. } => {
```
luckysori: :upside_down_face: We could add a method `collaborative_close` on `CfdState` that returns an `Option` of the `collaborative_close` field and check that after the `match` statement. We could reduce some code duplication that way.

luckysori (approved):
LGTM, just some suggestions.

daemon/src/monitor.rs
```diff
@@ -173,6 +185,13 @@ where
             .push((ScriptStatus::finality(), Event::CommitFinality(order_id)));
     }
 
+    fn monitor_close_finality(&mut self, close_params: (Txid, Script), order_id: OrderId) {
+        self.awaiting_status
+            .entry((close_params.0, close_params.1))
```
thomaseizinger: You can just pass `close_params` here, no need to reach into the tuple just to assemble a tuple again.

thomaseizinger (approved):
Thanks!

Looks good. Would like to see @luckysori suggestions but not blocking IMO.

daemon/src/model/cfd.rs
```diff
@@ -624,6 +655,11 @@ impl Cfd {
                         },
                     }
                 }
+                monitor::Event::CloseFinality(_) => CfdState::Closed {
```
da-kami, line 726: As discussed briefly yesterday: 

With the `Closed` state now containing the payout information we will have to extend `Closed` to have a enum that lets us distinguish if we closed collaboratively or un-collaboratively and then come up with the payout accordingly. For the collaborative close we will have to pass on the `Transaction` and extract the output similar to what we do on the cet transaction outputs.

daemon/src/model/cfd.rs
```diff
@@ -755,13 +795,19 @@ impl Cfd {
                     },
                     dlc,
                     attestation: Some(attestation),
+                    collaborative_close: None,
                 },
-                CfdState::PendingCommit { dlc, .. } => CfdState::PendingCommit {
+                CfdState::PendingCommit {
```
da-kami, line 871: I don't think it will be critical, but I think we should also attach the `attestation` to `PendingClose` for completion. It should never really matter, but since the `PendingClose` state records it we should also record it here.

daemon/src/model/cfd.rs
```diff
@@ -805,6 +851,51 @@ impl Cfd {
...
+                    dlc,
+                    attestation,
+                    ..
+                } => CfdState::Open {
+                    common,
+                    dlc,
+                    attestation,
+                    collaborative_close: Some((transaction, SystemTime::now())),
+                },
+                CfdState::PendingCommit {
```
da-kami: That is an interesting one - this transition is also mostly for completion, but in a real world scenario is very weird :D
Something along the lines of "we proposed collaborative settlement but then hit commit anyway before the maker came back to us to actually confirm the collaborative settlement". I think such a scenario will result in the maker  failing to send the collaboratively constructed transaction and we will have to use the cet.

klochowicz:
> > > @luckysori , @bonomat , perhaps I'm not using the protocol right?
> > 
> > 
> > No, it's not your fault 😅
> > Do you still have the data? If so, we should try to get the raw transaction and inspect it.
> 
> It happened yesterday arvo, so I should still have the data. I'll PM you how to extract the things that might be useful.

The following PR should deal with the issue with min relay fee not met: 
https://github.com/comit-network/hermes/pull/294

daemon/src/maker_cfd.rs
klochowicz: 😕 I'm afraid I don't understand.... `handle_commit()` etc are xtra actor message handlers, just like this function is already: `handle_initiate_settlement()`

thomaseizinger:
> > > > @luckysori , @bonomat , perhaps I'm not using the protocol right?
> > > 
> > > 
> > > No, it's not your fault sweat_smile
> > > Do you still have the data? If so, we should try to get the raw transaction and inspect it.
> > 
> > 
> > It happened yesterday arvo, so I should still have the data. I'll PM you how to extract the things that might be useful.
> 
> The following PR should deal with the issue with min relay fee not met: #294

Can rebase on top of that and target your PR against it?

klochowicz:
> > > > > @luckysori , @bonomat , perhaps I'm not using the protocol right?
> > > > 
> > > > 
> > > > No, it's not your fault sweat_smile
> > > > Do you still have the data? If so, we should try to get the raw transaction and inspect it.
> > > 
> > > 
> > > It happened yesterday arvo, so I should still have the data. I'll PM you how to extract the things that might be useful.
> > 
> > 
> > The following PR should deal with the issue with min relay fee not met: #294
> 
> Can rebase on top of that and target your PR against it?

it got merged, I need to rebase on master anyway, so it will be dealt with :)

daemon/src/model/cfd.rs
klochowicz: 👍  removed it!

daemon/src/model/cfd.rs
klochowicz: After rebase, I don't think `PendingClose` should have attestation, as it always transition to `Closed`without attestation value.

daemon/src/model/cfd.rs
klochowicz: This has been overtaken by changes, `PendingCommit` does not have `collaborative_settlement` anymore :)

daemon/src/model/cfd.rs
klochowicz: I aim to tackle that in a separate issue (so that it's easier for everyone, as this one was approved):

https://github.com/comit-network/hermes/issues/297

klochowicz:
in order to facilitate the painful rebase, I have autosquashed the fixups from the reviews (sorry!), but I have added a fixup commit that adds additional code that was needed after recent @da-kami 's changes.

daemon/src/model/cfd.rs
```diff
@@ -1577,3 +1682,19 @@ pub struct RevokedCommit {
     pub txid: Txid,
     pub script_pubkey: Script,
 }
+
+/// In some cases, when
```
thomaseizinger: Must have fallen asleep here :)

daemon/src/taker_cfd.rs
```diff
@@ -370,7 +370,10 @@ impl Actor {
             })
             .await?;
 
-        // TODO: Monitor for the transaction
+        cfd.handle(CfdStateChangeEvent::ProposalSigned(
+            TimestampedTransaction::new(tx),
+        ))?;
+        insert_new_cfd_state_by_order_id(cfd.order.id, cfd.state, &mut conn).await?;
```
thomaseizinger, line 375: If we are inserting a new state here, couldn't we use the transition timestamp instead of attaching another one onto the transaction?

thomaseizinger (approved):
LGTM!

daemon/src/taker_cfd.rs
klochowicz: we're adding an event, that doesn't change the state - it still stays in 'Open' - and as transition timestamps can happen due to variety of reasons (main ones being settlement proposals and attestations), we can't reliably tell how long did it take since the proposal signature...

daemon/src/model/cfd.rs
klochowicz: LOL :D that unfinished comment was meant to clear any doubts why we need the type (and would have likely avoided your second comment). adding now :)

PR #287, opened 2021-10-12
Fetch attestations only when it is time

thomaseizinger:
- Rename `OracleEventId`
- Refer to types within `fmt` module via module qualifier
- fixup! Refer to types within `fmt` module via module qualifier
- Prefer `.parse` over `FromStr` to avoid an import
- Properly model event IDs
- Fix typo
- Only fetch attestation that are likely to have occured
- Introduce drop-in replacement for tokio::spawn that logs errors

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -200,7 +200,7 @@ impl Actor {
             oracle::next_announcement_after(time::OffsetDateTime::now_utc() + Order::TERM)?;
 
         self.oracle_actor
-            .do_send_async(oracle::FetchAnnouncement(oracle_event_id.clone()))
+            .do_send_async(oracle::FetchAnnouncement(oracle_event_id))
             .await?;
 
         let order = Order::new(
@@ -586,13 +586,13 @@ impl Actor {
 
         let offer_announcement = self
             .oracle_actor
-            .send(oracle::GetAnnouncement(cfd.order.oracle_event_id.clone()))
+            .send(oracle::GetAnnouncement(cfd.order.oracle_event_id))
             .await?
             .with_context(|| format!("Announcement {} not found", cfd.order.oracle_event_id))?;
 
         self.oracle_actor
             .do_send_async(oracle::MonitorAttestation {
-                event_id: offer_announcement.id.clone(),
+                event_id: offer_announcement.id,
             })
             .await?;
 
@@ -767,7 +767,7 @@ impl Actor {
             oracle::next_announcement_after(time::OffsetDateTime::now_utc() + Order::TERM)?;
         let announcement = self
             .oracle_actor
-            .send(oracle::GetAnnouncement(oracle_event_id.clone()))
+            .send(oracle::GetAnnouncement(oracle_event_id))
             .await?
             .with_context(|| format!("Announcement {} not found", oracle_event_id))?;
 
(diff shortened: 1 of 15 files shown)


daemon/src/oracle.rs
```diff
@@ -258,11 +253,7 @@ fn ceil_to_next_hour(original: OffsetDateTime) -> Result<OffsetDateTime, anyhow:
 /// Construct the URL of `olivia`'s `BitMEX/BXBT` event to be attested
 /// for at the time indicated by the argument `datetime`.
 fn event_id(datetime: OffsetDateTime) -> BitMexPriceEventId {
```
luckysori: :wrench: I think we can remove `event_id`.

daemon/src/model.rs
```diff
@@ -127,6 +127,15 @@ impl BitMexPriceEventId {
         Self::new(timestamp, 20)
     }
 
+    /// Checks whether this event has likely already occurred.
+    ///
+    /// We can't be sure about it because our local clock might be off from the oracle's clock.
+    pub fn has_likely_occured(&self) -> bool {
+        let now = OffsetDateTime::now_utc();
+
+        self.timestamp > now
```
luckysori: :mega: Isn't this the wrong way round? We want the event's timestamp to be in the past.

luckysori (changes requested):
Looks good except for what I think is a logic error.

daemon/src/model.rs
thomaseizinger: Good catch!

luckysori (approved):

daemon/src/maker_cfd.rs
```diff
@@ -200,7 +200,7 @@ impl Actor {
             oracle::next_announcement_after(time::OffsetDateTime::now_utc() + Order::TERM)?;
 
         self.oracle_actor
-            .do_send_async(oracle::FetchAnnouncement(oracle_event_id.clone()))
+            .do_send_async(oracle::FetchAnnouncement(oracle_event_id))
```
da-kami, line 203: ❓ Just noticed that now - not related to this PR - what is the difference between `FetchAnnouncement` and `GetAnnouncement` ? Is this in the literal sense - i.e. `GetAnnouncement` (tries to) get it from the locally known state and `FetchAnnouncement` fetches it from the remote?

daemon/src/tokio_ext.rs
```diff
@@ -0,0 +1,14 @@
+use std::fmt;
+use std::future::Future;
+
+pub fn spawn_fallible<F, E>(future: F)
```
da-kami, line 4: interesting concept. Question on top - again unrelated to the PR - what was the motivation for spawning this task inside the oracle? isn't the whole sync triggered asynchronously?

da-kami (approved):

daemon/src/maker_cfd.rs
thomaseizinger: I will add documentation.

thomaseizinger: Oh, there is documentation. Can you check that one and see if it answers your question?

daemon/src/tokio_ext.rs
thomaseizinger: Parallelization.

The task cannot hold references meaning it cannot update the state once it is done, hence we do the loop inside the handler, spawn of a number of tasks which all send a message at the end. Thus, all the tasks are independent, one failing doesn't impact the others and we get to update the state only for the successful ones.

PR #288, opened 2021-10-12
Rename order price to offer price

bonomat:
I want to trigger a new `preview` build and wasn't sure if there is another way. Looks like a last release build failed: https://github.com/comit-network/hermes/runs/3868563726?check_suite_focus=true
And I want to see if this was a glitch or if something else went wrong.


DeliciousHair (approved):
Request granted😀

PR #291, opened 2021-10-13
Delete tag in addition to release

thomaseizinger:


bonomat (approved):
Nice, thanks for looking into it.

PR #293, opened 2021-10-13
We need to log in as the organisation

bonomat:
Untested - just a WAG at the moment. 

The [build](https://github.com/comit-network/hermes/runs/3877728910?check_suite_focus=true) failed because it tried to push the container to comit-botty-mc-botface. We need to push to comit-network.

### DIFF
--- a/.github/workflows/build-release-binary.yml
+++ b/.github/workflows/build-release-binary.yml
@@ -168,7 +168,7 @@ jobs:
         if: matrix.os != 'windows-latest' && matrix.os != 'macos-latest'
         with:
           registry: ghcr.io
-          username: ${{ github.actor }}
+          username: comit-network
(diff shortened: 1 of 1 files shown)


thomaseizinger (approved):

thomaseizinger:
Are you sure we need to log in as an organization? We might also just have to update the permissions of the token that is stored for botty.

bonomat:
> Are you sure we need to log in as an organization? We might also just have to update the permissions of the token that is stored for botty.

Not 100% but I used the docker cli to login locally and the docker image landed here: https://github.com/orgs/comit-network/packages/container/package/hermes-maker 

I'm still trying on my fork...

bonomat:
Replaced by 

image

PR #294, opened 2021-10-13
Fix min relay not met error

klochowicz:


klochowicz:
thanks @bonomat for guidance! 🏄

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -131,9 +131,9 @@ impl CommitTransaction {
             input: vec![lock_input],
             output: vec![output],
         };
-        let fee = (Self::SIGNED_VBYTES * SATS_PER_VBYTE as f64) as u64;
+        let fee = Fee::new(Self::SIGNED_VBYTES);
 
-        let commit_tx_amount = lock_amount - fee as u64;
+        let commit_tx_amount = lock_amount - fee.rounded();
```
da-kami, line 136: Is rounded really the way to go, not `floor` ?
Will this be correct if it rounds up?

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -21,6 +21,45 @@ use std::num::NonZeroU8;
...
+    fee: f64,
+}
+
+impl Fee {
+    pub fn new(signed_vbytes: f64) -> Self {
+        let fee = signed_vbytes * SATS_PER_VBYTE;
+        Self { fee }
+    }
+
+    /// Note: does not mutate, but creates another Fee object
```
thomaseizinger: You could put `#[must_use]` here so the compiler generates a warning if you don't use the result.

Alternatively, consuming `self` will also make this harder to misuse.

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -21,6 +21,45 @@ use std::num::NonZeroU8;
...
+    pub fn add(&self, number: f64) -> Fee {
+        Fee {
+            fee: self.fee + number,
+        }
+    }
+
+    pub fn rounded(&self) -> u64 {
+        self.fee.ceil() as u64
```
thomaseizinger: Ceiling is not rounding?

thomaseizinger: I think I would prefer `as_u64`.

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -496,3 +494,42 @@ pub fn punish_transaction(
...
```
da-kami, line 517: Interesting that we `ceil` and not `floor` - would be cool to have a comment that outlines why. 
I would have expected that we always `floor` the value to be on the save side.

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -21,6 +21,45 @@ use std::num::NonZeroU8;
...
+#[test]
+fn test_splitting_fees() {
+    const SIGNED_VBYTES_TEST: f64 = 148.5 + (3.0 * 2.0) / 4.0;
+
+    let fee = Fee::new(SIGNED_VBYTES_TEST);
+
+    let (maker_fee, taker_fee) = fee.split();
+
+    assert_eq!(fee.rounded(), maker_fee + taker_fee);
+}
```
thomaseizinger: Needs a `cfg(test)`.

thomaseizinger: Also, would be nice to make this a proptest.

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -21,6 +21,45 @@ use std::num::NonZeroU8;
 /// In satoshi per vbyte.
 const SATS_PER_VBYTE: f64 = 1.0;
 
+#[derive(Clone, Debug)]
+struct Fee {
+    fee: f64,
+}
+
+impl Fee {
+    pub fn new(signed_vbytes: f64) -> Self {
```
thomaseizinger: If the type is not public, no need for `pub`!

cfd_protocol/src/protocol/transactions.rs
da-kami, line 518: ❓ This `ceil` is for the minimum relay fee, so with `ceil`ing we ensure that the minimum is always met, correct?
Just wanna make sure my understanding is correct :)

thomaseizinger (approved):

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -496,3 +494,42 @@ pub fn punish_transaction(
...
+        Fee {
+            fee: self.fee + number,
+        }
+    }
+
+    pub fn rounded(&self) -> u64 {
+        self.fee.ceil() as u64
+    }
+
+    pub fn split(&self) -> (u64, u64) {
```
da-kami: I am a bit confused by this, can we maybe add some more simple tests to show the behavior?

let's say we have `3.0` as `f64` self:

> `let half = self.rounded() / 2;` 

`self.rounded()` = 3
`half = 3 / 2` = 1.5

> `(half as u64, self.rounded() - half)`

`half as u64` = 2 // I assume? Does this always round?
`self.rounded() - half` = 3 - 2 = 1

so the output would be `(2, 1)`
Is that the expected behavior?

Would be cool to see behavior for odd and even number, and potentially also the dot ranges.

da-kami (approved):

cfd_protocol/src/protocol/transactions.rs
klochowicz: changed to `as_u64`

cfd_protocol/src/protocol/transactions.rs
klochowicz: great suggestions. added both :)

cfd_protocol/src/protocol/transactions.rs
klochowicz: removed `pub` from all the methods.

cfd_protocol/src/protocol/transactions.rs
klochowicz: yes, I'll add a comment reflecting that.

cfd_protocol/src/protocol/transactions.rs
klochowicz: Added `cfd(test)` and converted into a proptest.

BTW: Never heard of proptest before - it's really cool!

cfd_protocol/src/protocol/transactions.rs
klochowicz: I added a few test cases in addition to the proptest to illustrate the behaviour, for `1.0`, `2.0` and `2.6` - it should be much more clear now :)

cfd_protocol/src/protocol/transactions.rs
```diff
@@ -501,35 +501,83 @@ struct Fee {
...
+        assert_eq!(fee.as_u64(), 1);
+        assert_eq!(maker_fee, 0);
+        assert_eq!(taker_fee, 1);
+        assert!((maker_fee + taker_fee) as f64 >= SIGNED_VBYTES_TEST);
+    }
 
-    assert_eq!(fee.rounded(), maker_fee + taker_fee);
+    #[test]
+    fn test_splitting_fee_2_0() {
+        const SIGNED_VBYTES_TEST: f64 = 1.0;
```
da-kami: shouldn't this be `2.0` here?

da-kami (approved):

luckysori (approved):

Issue #88, opened 2021-09-20
Refactor the database loading/saving for readability


da-kami:
At the moment we load and save types by using `serde_json::to_string` and `serde_json::from_str` directly in code.

sqlx offers en- and de-coding traits that would make this code more readable and easier to use:
https://docs.rs/sqlx/0.5.7/sqlx/trait.Decode.html
https://docs.rs/sqlx/0.5.7/sqlx/derive.Decode.html

da-kami:
I think with #295 merged we can close this

PR #295, opened 2021-10-13
Db refactor

DeliciousHair:
DB code refactoring for #88

The primary goal was to remove all of the calls to `serde_json::to_string()` for the data-handling, thus enabling us to do (more or less):

```rust
Order {
    row.column,
    ...
}
```

as well as clean up the SQL for easier reading. This has mostly been accomplished, with further refinements easily accomplished once the upstream issues in `sqlx` are addressed. See #314 for issues we are tracking.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2616,6 +2616,7 @@ dependencies = [
  "thiserror",
  "tokio-stream",
  "url",
+ "uuid",
  "webpki",
  "webpki-roots 0.21.1",
  "whoami",
--- a/daemon/migrations/20210903050345_create_cfd_and_order_tables.sql
+++ b/daemon/migrations/20210903050345_create_cfd_and_order_tables.sql
@@ -1,19 +1,21 @@
 -- todo: Decimal is had to deserialize as number so we use text
 create table if not exists orders
 (
-    id                 integer primary key autoincrement,
-    uuid               text unique not null,
-    trading_pair       text        not null,
-    position           text        not null,
-    initial_price      text        not null,
-    min_quantity       text        not null,
-    max_quantity       text        not null,
-    leverage           integer     not null,
-    liquidation_price  text        not null,
-    creation_timestamp text        not null,
-    term               text        not null,
-    origin             text        not null,
-    oracle_event_id    text        not null
+    id                              integer primary key autoincrement,
+    uuid                            text unique not null,
+    trading_pair                    text        not null,
+    position                        text        not null,
+    initial_price                   text        not null,
+    min_quantity                    text        not null,
+    max_quantity                    text        not null,
+    leverage                        integer     not null,
+    liquidation_price               text        not null,
+    creation_timestamp_seconds      integer     not null,
+    creation_timestamp_nanoseconds  integer     not null,
+    term_seconds                    integer     not null,
+    term_nanoseconds                integer     not null,
+    origin                          text        not null,
+    oracle_event_id                 text        not null
 );
 
 create unique index if not exists orders_uuid
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1,11 +1,14 @@
-use crate::model::cfd::{Cfd, CfdState, Order, OrderId, Origin};
-use crate::model::{BitMexPriceEventId, Leverage, Position};
+use crate::model::cfd::{Cfd, CfdState, Order, OrderId};
+use crate::model::{BitMexPriceEventId, Usd};
 use anyhow::{Context, Result};
 use rocket_db_pools::sqlx;
+use rust_decimal::Decimal;
 use sqlx::pool::PoolConnection;
-use sqlx::{Acquire, Sqlite, SqlitePool};
-use std::convert::TryInto;
+use sqlx::{Sqlite, SqlitePool};
 use std::mem;
+use std::str::FromStr;
+use std::time::SystemTime;
+use time::{Duration, OffsetDateTime};
 
 pub async fn run_migrations(pool: &SqlitePool) -> anyhow::Result<()> {
     sqlx::migrate!("./migrations").run(pool).await?;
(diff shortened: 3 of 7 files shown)


daemon/migrations/20210903050345_create_cfd_and_order_tables.sql
```diff
@@ -1,19 +1,21 @@
...
+    uuid                            text unique not null,
+    trading_pair                    text        not null,
+    position                        text        not null,
+    initial_price                   text        not null,
+    min_quantity                    text        not null,
+    max_quantity                    text        not null,
+    leverage                        integer     not null,
+    liquidation_price               text        not null,
+    creation_timestamp_seconds      integer     not null,
+    creation_timestamp_nanoseconds  integer     not null,
```
thomaseizinger, line 14: I think I'd rather drop the nanoseconds. What do you think @da-kami ?

daemon/src/db.rs
```diff
@@ -296,8 +306,8 @@ pub async fn load_cfd_by_order_id(
     let max_quantity = serde_json::from_str(row.max_quantity.as_str()).unwrap();
```
thomaseizinger: Parsing `max_quantity` via `serde` here is not going to work is it? We are only `to_string`ing it on the insert.

daemon/src/db.rs
```diff
@@ -493,6 +513,19 @@ pub async fn load_cfds_by_oracle_event_id(
...
+    let nanos = row_nanos as i128;
+    let offset_dt =
+        OffsetDateTime::from_unix_timestamp_nanos(secs * 1_000_000_000 + nanos).unwrap();
+
+    SystemTime::from(offset_dt)
+}
+
+fn convert_to_duration(row_secs: i64, row_nanos: i32) -> Duration {
+    Duration::new(row_secs, row_nanos)
+}
```
thomaseizinger: Don't think this function is providing much benefit?

daemon/src/model.rs
```diff
@@ -14,7 +14,8 @@ use uuid::Uuid;
 
 pub mod cfd;
 
-#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, sqlx::Type)]
+#[sqlx(type_name = "String")]
```
thomaseizinger: IIRC, this only does something for postgres so we can remove it.

daemon/src/db_sql/query_load_cfd_by_order_id.sql
```diff
@@ -0,0 +1,70 @@
...
+    order.creation_timestamp_nanoseconds,
+    order.term_seconds,
+    order.term_nanoseconds,
+    order.origin,
+    order.oracle_event_id,
+    state.quantity_usd,
+    state.state
+
+from order
+    inner join state on state.order_id = order.order_id
```
thomaseizinger: Missing newlines on these.

daemon/src/db_sql/query_load_all_cfds.sql
```diff
@@ -0,0 +1,69 @@
+with order as (
```
thomaseizinger: I'd prefer if the changes were made in two steps:

1. Moving queries to external files
2. Changing up the query to use `with`

thomaseizinger (changes requested):
Thanks!

At the moment, there are multiple changes mangled up here. I can already see that this will prove tricky to untangle into multiple, distinct commits.

I would suggest the following:

Make a new branch off master where you just change the SQL queries to use the `with` syntax and move them into separate files without any of the improvements around `sqlx::Type` etc.

Given that you already have the queries, that should be reasonably easy. On top of that, we can then introduce `sqlx::Type` etc and get rid of all the `serde_json` stuff!

daemon/src/db.rs
DeliciousHair: FYI, all things that involve `serde_json::from_str()` will go away with one exception:

```rust
    // TODO:
    // still have the use of serde_json::from_str() here, which will be dealt with
    // via https://github.com/comit-network/hermes/issues/290
    Ok(Cfd {
        order,
        quantity_usd: Usd(Decimal::from_str(&row.quantity_usd).unwrap()),
        state: serde_json::from_str(row.state.as_str()).unwrap(),
    })
```

daemon/src/db.rs
DeliciousHair: Benefit is visual only; if nanoseconds are dropped then even that benefit is gone. :-)

thomaseizinger:
Actually, all things considered it is not that important to split things up into multiple commits so feel free to squash everything together!

Overall, these changes are a major improvement, thank you!

daemon/migrations/20210903050345_create_cfd_and_order_tables.sql
DeliciousHair: @thomaseizinger @da-kami I've left the ns in for the moment as they are now handled without issue. Can drop them, but it looks like that may lead to a bit more work upstream as the timestamps are generated by `SystemTime::new()` so if the ns are dropped, things may not line up.

Not a big deal either way, this just seemed the path of least resistance for the moment.

daemon/src/db_sql/query_load_all_cfds.sql
```diff
@@ -0,0 +1,71 @@
...
+  ord.term_secs as "term_secs: i64",
+  ord.term_nanos as "term_nanos: i32",
+  ord.origin as "origin: Origin",
+  ord.oracle_event_id as "oracle_event_id: OracleEventId",
+  state.quantity_usd,
+  state.state
+
+from ord
+  inner join state on state.order_id = ord.order_id
+
```
thomaseizinger: Missing newline!

daemon/src/model.rs
thomaseizinger: This is unaddressed !

daemon/src/db.rs
```diff
@@ -66,98 +71,39 @@ pub async fn load_order_by_id(
...
-    let initial_price = serde_json::from_str(row.initial_price.as_str()).unwrap();
-    let min_quantity = serde_json::from_str(row.min_quantity.as_str()).unwrap();
-    let max_quantity = serde_json::from_str(row.max_quantity.as_str()).unwrap();
-    let leverage = Leverage(row.leverage.try_into().unwrap());
-    let liquidation_price = serde_json::from_str(row.liquidation_price.as_str()).unwrap();
-    let creation_timestamp = serde_json::from_str(row.creation_timestamp.as_str()).unwrap();
-    let term = serde_json::from_str(row.term.as_str()).unwrap();
-    let origin = serde_json::from_str(row.origin.as_str()).unwrap();
-    let oracle_event_id = OracleEventId(row.oracle_event_id);
+    let row = sqlx::query_file!("src/db_sql/query_load_order_by_id.sql", id)
```
thomaseizinger: I'd prefer if all the SQL files are next to the rust files and not in their own directory.

daemon/src/db.rs
```diff
@@ -66,98 +71,39 @@ pub async fn load_order_by_id(
...
-        leverage,
-        liquidation_price,
-        creation_timestamp,
-        term,
-        origin,
-        oracle_event_id,
+        id: row.uuid,
+        trading_pair: row.trading_pair,
+        position: row.position,
+        price: Usd(Decimal::from_str(&row.initial_price).unwrap()),
```
thomaseizinger: I know we `unwrap()`ed here before but really we shouldn't. Can you replace those with `?` please?

daemon/src/db.rs
```diff
@@ -637,7 +453,7 @@ mod tests {
             .unwrap();
 
         let cfds_from_db = load_all_cfds(&mut conn).await.unwrap();
-        let cfd_from_db = cfds_from_db.first().unwrap().clone();
+        let cfd_from_db = cfds_from_db.last().unwrap().clone();
```
thomaseizinger, line 723: Why did we need to change this test?

This looks like a regression!

thomaseizinger (changes requested):
Thank you! Massive improvement over what we currently have!

- Found what I think is a regression.
- A few unaddressed comments and commits need to be squashed together.

daemon/src/model/cfd.rs
```diff
@@ -67,6 +68,7 @@ impl From<Origin> for Role {
 }
 
 /// A concrete order created by a maker for a taker
+// #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
thomaseizinger: This doesn't seem necessary :)

daemon/src/db.rs
DeliciousHair: This was changed because the the SQL had an unnecessary reverse-ordering step:
```
        where cfd_states.state in (
            select
              state
              from cfd_states
            where cfd_id = cfds.id
            order by id desc
            limit 1
        )
```

With the reversal step removed, the test item moves from `.first()` to `.last()` is all.

Cargo.lock
```diff
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ahash"
-version = "0.7.4"
+version = "0.7.5"
```
thomaseizinger: These update must have been accidental. Can you undo that please?

`git checkout master -- Cargo.lock` should do the trick.

daemon/src/db.rs
```diff
@@ -66,97 +68,83 @@ pub async fn load_order_by_id(
...
+        liquidation_price: Usd(Decimal::from_str(&row.liquidation_price)?),
+        creation_timestamp: convert_to_system_time(row.ts_secs, row.ts_nanos)?,
+        term: Duration::new(row.term_secs, row.term_nanos),
+        origin: row.origin,
+        oracle_event_id: row.oracle_event_id.parse::<BitMexPriceEventId>()?,
     })
 }
 
 pub async fn insert_cfd(cfd: Cfd, conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<()> {
-    let mut tx = conn.begin().await?;
```
thomaseizinger, line 109: Cool! No more transaction for this!

daemon/src/model.rs
```diff
@@ -13,7 +13,7 @@ use uuid::Uuid;
 
 pub mod cfd;
 
-#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, sqlx::Type)]
```
thomaseizinger: I think this is not actually doing anything because we are not specifying `#[sqlx(transparent)]`.

Can you remove it to avoid confusion about what this is doing?

thomaseizinger (changes requested):
Thanks!

2 more tiny changes and this is good to go in. Just amend the now only commit for that (`git commit --amend`). Also, can you reword the commit message to something more descriptive please?

daemon/src/model.rs
DeliciousHair: Oops! Yes, removed now.

thomaseizinger (approved):
Thanks!

Feel free to `bors r+` once CI is green.

PR #296, opened 2021-10-13
Use tokio_ext::spawn_fallible instead of log_error!

thomaseizinger:
Just an idea. What do you think @klochowicz ?

I think we can replace our `log_error!` usage with a combination of `send` and `spawn_fallible`.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -212,8 +212,7 @@ async fn main() -> Result<()> {
                 let (maker_inc_connections_address, maker_inc_connections_context) =
                     xtra::Context::new(None);
 
-                let (monitor_actor_address, mut monitor_actor_context) = xtra::Context::new(None);
-
+                let (monitor_actor_address, monitor_actor_context) = xtra::Context::new(None);
                 let (oracle_actor_address, mut oracle_actor_context) = xtra::Context::new(None);
 
                 let mut conn = db.acquire().await.unwrap();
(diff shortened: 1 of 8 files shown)


klochowicz:
I'm not sure if the change is worth it - it makes it harder to see *why* we're doing the thing what we're doing:
- it hides the details that the logging is going to happen
- it makes it harder to seethe fact that we're sending an async task
- it does not fix the core problem (users could still spawn a future with `Result<>` and never see the quashed error)

I believe that ultimately we should extend `xtra` to log an error instead of ignoring it (maybe behind a feature flag or a handler that library clients could implement if the author wouldn't want a dependency on `tracing`)? Note: it's not an original thought, AFAICT it was suggested by you some time ago 😸

thomaseizinger:
One of the things I want to achieve is being able to get the error on the sending side. For example, I think we should actually synchronously send messages like `AcceptOrder` on the HTTP handler so that we can give a direct error back to the user. That will be much easier than sending an error on some kind of feed.

But for this to work, we cannot swallow all results in the handler. Hence, I think the same way it is a decision of the sending side to choose between sync / async send, it should also be a decision on what happens to the error in case you want to send it synchronously.

I don't actually understand why `do_send_async` is a thing as it can always be built as `tokio::spawn(actor.send(...))`.

klochowicz:
that is indeed a valid use-case, thanks for clarifying this!

do we want to expand it to returning error from async sends? AFAICT actix uses some oneshot channels to return back the `Result<>` to the caller

thomaseizinger:
> do we want to expand it to returning error from async sends? AFAICT actix uses some oneshot channels to return back the `Result<>` to the caller

Well, for me an async send implies that I don't care about the outcome, so why would I send it back? Most of our handlers return `Result<(), Error>` so all I really want is to log  the error which I can do with `spawn_fallible`! :)

PR #300, opened 2021-10-13
Fix docker image build

### LINKED ISSUE #285: Cleanup release workflow
#273 introduced a docker image build during release time. 

There are a few things to clean up: 

1. The docker image is created as part of the main release build. This is messy and should be cleaned up: For example, we could have a separate job after the release job only taking care of creating and publishing the docker image. This will need to have a dependency on the first job and download the required release binaries. This could also simplify the setup. 
2. add more files to `.dockerignore` : right now the whole target directory is included in the context which could be improved.

bonomat:
The new task does a few things:
1. Having an extra build target makes the workflow file cleaner
2. it downloads the release targets which we want to build docker images for
3. it extracts the tar files into sub folders. We need to extract into sub folders to so that our image can fine the binaries. By providing `platform` to `docker buildx` the variable `TARGETPLATFORM` will be available when building the image. This variable is either `linux/amd64/`, `linux/arm64/` or `linux/arm/v7`. Hence we need to extract into subfolders.
4. We create 2 images, one for the maker and one for the taker. The images are then pushed to our organisation's docker registry.

### DIFF
--- a/.github/workflows/build-release-binary.yml
+++ b/.github/workflows/build-release-binary.yml
@@ -14,52 +14,42 @@ jobs:
             target: x86_64-unknown-linux-gnu
             os: ubuntu-latest
             archive_ext: tar
-            docker_platforms: linux/amd64
           - bin: taker
             target: armv7-unknown-linux-gnueabihf
             os: ubuntu-latest
             archive_ext: tar
-            docker_platforms: linux/arm/v7
           - bin: taker
             target: aarch64-unknown-linux-gnu
             os: ubuntu-latest
             archive_ext: tar
-            docker_platforms: linux/arm64
           - bin: taker
             target: x86_64-apple-darwin
             os: macos-latest
             archive_ext: tar
-            docker_platforms: None
           - bin: taker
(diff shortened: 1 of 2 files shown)


bonomat:
Let's hope this works. 
It did on my fork: 
https://github.com/bonomat/hermes/runs/3888576210?check_suite_focus=true
and published the images there: 
https://github.com/bonomat?tab=packages&repo_name=hermes

thomaseizinger (approved):
Great work, much better! Thank you!

One question: Why do we have to publish to the organization's packages? I would have expected the images to appear here: https://github.com/comit-network/hermes/packages

bonomat:
> Great work, much better! Thank you!
> 
> One question: Why do we have to publish to the organization's packages? I would have expected the images to appear here: [comit-network/hermes/packages](https://github.com/comit-network/hermes/packages)

Good question. I don't know unfortunately. Let's see if this works and then try to publish somewhere else.

bonomat:
Testing to publish under the repo here: https://github.com/bonomat/hermes/actions/runs/1339699306

PR #302, opened 2021-10-14
Collaborative close payout

da-kami:
Distinguish between collaborative close and cet closing.
Moved the payout on the `Cfd` which decides what the payout of the `Cfd` is, where

- `None` means the payout is not decided yet (we can optimize that and potentially combine this with `profit` in another iteration)
- `Some(payout)` can be a refund, collaborative or cet payout

Refund is handled in an early exit guard based on the state.
Collaborative settlement has priority over cet payout (attestation). In case there is a collaborative settlement present we always return payout based on that.

Note: I am not totally sure if collaborative close payout should always have priority over attestation payout. We could do a more complex reasoning about the state as well, but I felt for now this is probably good enough and we can adapt based on experience. Happy for other suggestions :)

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -273,7 +273,7 @@ pub enum CfdState {
     Closed {
         common: CfdStateCommon,
         // TODO: Use an enum of either Attestation or CollaborativeSettlement
-        attestation: Option<Attestation>,
+        payout: Payout,
     },
 
     // TODO: Can be extended with CetStatus
@@ -295,6 +295,12 @@ pub enum CfdState {
         common: CfdStateCommon,
         info: String,
     },
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub enum Payout {
+    CollaborativeClose(TimestampedTransaction),
+    Cet(Attestation),
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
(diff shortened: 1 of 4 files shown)


daemon/src/model/cfd.rs
```diff
@@ -1664,6 +1726,13 @@ impl Dlc {
             .map(|output| Amount::from_sat(output.value))
             .unwrap_or_default()
     }
+
+    pub fn own_script_pubkey(&self, role: Role) -> Script {
```
thomaseizinger: I don't think `own` is entirely correct here. If I am passing in the other party's role, it is not true. Perhaps this is better?

```suggestion
    pub fn script_pubkey_for(&self, role: Role) -> Script {
```

daemon/src/model/cfd.rs
```diff
@@ -1688,13 +1757,35 @@ pub struct RevokedCommit {
...
+            .iter()
+            .find(|output| output.script_pubkey == own_script_pubkey)
+            .map(|output| Amount::from_sat(output.value))
+            .unwrap_or_default();
+
+        if payout == Amount::ZERO {
+            tracing::error!(
+                "Collaborative settlement with a zero amount, this should really not happen!"
+            );
+        }
```
thomaseizinger: In this case, I think matching on the `Option` returned from `find` and logging as part of returning `Amount::ZERO` is easier to understand:

```suggestion
        let payout = match tx
            .output
            .iter()
            .find(|output| output.script_pubkey == own_script_pubkey)
            .map(|output| Amount::from_sat(output.value)) {
      	    Some(payout) => payout,
      	    None => {
      	  	tracing::error!(
	            "Collaborative settlement with a zero amount, this should really not happen!");
	        Amount::ZERO
	    }    
        };
```

thomaseizinger (approved):

PR #307, opened 2021-10-14
Setting boundaries

bonomat:
This improves the UX slightly as the user knows what is the min and max amount. 

The UI will default to min amount.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -525,7 +525,18 @@ impl Actor {
             }
         };
 
-        // 2. Insert CFD in DB
+        // 2. check if order has acceptable amounts
+        if quantity < current_order.min_quantity || quantity > current_order.max_quantity {
+            self.takers
+                .do_send_async(maker_inc_connections::TakerMessage {
+                    taker_id,
+                    command: TakerCommand::NotifyOrderRejected { id: order_id },
+                })
+                .await?;
+            return Ok(());
+        }
+
+        // 3. Insert CFD in DB
         let cfd = Cfd::new(
             current_order.clone(),
             quantity,
@@ -541,7 +552,7 @@ impl Actor {
         self.cfd_feed_actor_inbox
             .send(load_all_cfds(&mut conn).await?)?;
 
-        // 3. Remove current order
+        // 4. Remove current order
         self.current_order_id = None;
         self.takers
             .do_send_async(maker_inc_connections::BroadcastOrder(None))
--- a/frontend/src/TakerApp.tsx
+++ b/frontend/src/TakerApp.tsx
@@ -15,7 +15,7 @@ import {
     useToast,
     VStack,
 } from "@chakra-ui/react";
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useAsync } from "react-async";
 import { useEventSource } from "react-sse-hooks";
 import { CfdTable } from "./components/cfdtables/CfdTable";
@@ -58,6 +58,8 @@ async function getMargin(payload: MarginRequestPayload): Promise<MarginResponse>
 }
 
 export default function App() {
+    const toast = useToast();
+
     document.title = "Hermes Taker";
 
     let source = useEventSource({ source: "/api/feed" });
@@ -67,9 +69,11 @@ export default function App() {
     const order = useLatestEvent<Order>(source, "order", intoOrder);
     const walletInfo = useLatestEvent<WalletInfo>(source, "wallet");
 
-    const toast = useToast();
     let [quantity, setQuantity] = useState("0");
     let [margin, setMargin] = useState("0");
+    let [userHasEdited, setUserHasEdited] = useState(false);
+
+    let quantityToShow = userHasEdited ? quantity : (order?.min_quantity.toString() || "0");
 
     let { run: calculateMargin } = useAsync({
         deferFn: async ([payload]: any[]) => {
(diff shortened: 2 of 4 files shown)


frontend/src/TakerApp.tsx
```diff
@@ -115,6 +115,24 @@ export default function App() {
...
+            return;
+        }
+        let quantity = valueString ? Number.parseFloat(valueString) : 0;
+        let payload: MarginRequestPayload = {
+            leverage: order.leverage,
+            price: order.price,
+            quantity,
+        };
+        calculateMargin(payload);
+    }
```
thomaseizinger: Side effects should always happen within `useEffect`!

thomaseizinger: Hang on, this is not a side-effect at all. This should be changed to just use derivation.

thomaseizinger (changes requested):

da-kami (commented):
Concept ack but requires some fixes (see complaints of CI and Thomas 🤪)

frontend/src/TakerApp.tsx
bonomat: > This should be changed to just use derivation.

Can you clarify please, I don't follow :)

thomaseizinger: Sure!

- Components in React are pure-ish functions from Data -> View
- To make state changes easier, React provides hooks that allow some form of "escape" out of the pure nature, i.e. `useState` allows you to persist state across the invocations of the render function

The combination of the above allows React to determine when the render function (your component) needs to be called again to correctly update the UI based on hooks being called and / or the inputs of your render function (props) changing.

The consequence of that is that the view of your component (the JSX you return) should be a pure function of your props + local state (which may come from hooks).

"Derivation" here just means that you only do pure function calls / computations:

```ts
let foo = bar + 2;
let dividable_by_three = foo % 3 == 0 ? "Dividable by 3" : "Not dividable by 3";
```

You must never call `setState` as part of the normal render function as that may cause infinite render loops because React things it needs to re-render again.

You are doing exactly that on line 123 though. Instead of updating the state, you should derive new state based on the current one.

bonomat: oh, thanks for the clear explanation. 

That means I cannot update the `margin` unless I would calculate it here as well.

I'll leave the default value of `quantity` to `0` then. This will make the field invalid as `0` won't be within the bounds but I think that is better than having `margin=0` and `quantity=100`.

thomaseizinger: > That means I cannot update the `margin` unless I would calculate it here as well.

You can, you just need to do it in a controlled way via `useEffect` where you declare all dependencies upon which you wish to cause a re-render.

thomaseizinger: Alternatively, and that is often the better solution, these kind of situations make it visible that there is actually state we care about, but are not tracking at the moment, for example whether or not the use has edited the field already. We had a similar situation in the past in regards to calculating the initial price.

daemon/src/maker_cfd.rs
```diff
@@ -521,7 +521,18 @@ impl Actor {
             }
         };
 
-        // 2. Insert CFD in DB
+        // 2. check if order has acceptable amounts
+        if quantity.0 < current_order.min_quantity.0 || quantity.0 > current_order.max_quantity.0 {
```
thomaseizinger: Instead of reaching into new types via `.0`, it would be better if we could establish a habit of implementing the required traits on it.

frontend/src/components/CurrencyInputField.tsx
```diff
@@ -11,18 +11,27 @@ import React from "react";
...
 export default function CurrencyInputField(
     {
         onChange,
         value,
+        min,
+        max,
     }: CurrencyInputFieldProps,
 ) {
+    let minAmount = min ? min! : 0;
+    let maxAmount = max ? max! : Number.MAX_SAFE_INTEGER;
```
thomaseizinger: I _think_ there is a dedicated operator in TS for what you are trying to achieve here.

```suggestion
    let minAmount = min || 0;
    let maxAmount = max || Number.MAX_SAFE_INTEGER;
```

thomaseizinger (approved):

frontend/src/components/CurrencyInputField.tsx
bonomat: nice,

frontend/src/TakerApp.tsx
bonomat: I'm interested in the second option. What do you have in mind? Indicating that the user hasn't edited the field and don't show the margin?

da-kami (approved):

frontend/src/TakerApp.tsx
thomaseizinger: If I understand correctly, what you are trying to achieve is to set a useful default value yes? To model a default value, we have to keep track of the state where the value came from (initial vs user-filled in). So we can do something like:

```ts
let [userHasEdited, setUserHasEdited] = useState(false);	
let quantityToShow = userHasEdited ? quantity : order.minQuantity;
```

And do `setUserHasEdited(true)` in the onChange handler of the input field.

thomaseizinger: You can then go on and calculate the `margin` based off `quantityToShow` without caring about where it came from.

bonomat: Like this [`8450ef6` (#307)](https://github.com/comit-network/hermes/pull/307/commits/8450ef6fb0cb50397590a3fa843b4a6563e20f55)?

thomaseizinger (approved):
Yep, that looks good! Well done :)

PR #308, opened 2021-10-14
Close bad connections to takers

### LINKED ISSUE #303: Don't open p2p port via browser
If one connects to the maker's p2p port via the browser one will crash the node. 

The maker will print the following and becomes unresponsive. 

```
2021-10-14 11:17:18  WARN Error while receiving message from taker 145d3ae8-a8b9-4763-9acc-d60692d7ad35: frame size too big
2021-10-14 11:17:18  WARN Error while receiving message from taker fc3909fc-6df2-46a6-822c-97b13f999543: frame size too big
2021-10-14 11:17:18  WARN Error while receiving message from taker fc3909fc-6df2-46a6-822c-97b13f999543: frame size too big
2021-10-14 11:17:18  WARN Error while receiving message from taker 145d3ae8-a8b9-4763-9acc-d60692d7ad35: frame size too big
2021-10-14 11:17:18  WARN Error while receiving message from taker 145d3ae8-a8b9-4763-9acc-d60692d7ad35: frame size too big
```

thomaseizinger:
- Introduce dedicated actor for forwarding only `Ok` messages
- Close connections with takers that don't adhere to the message format

### DIFF
--- a/daemon/src/forward_only_ok.rs
+++ b/daemon/src/forward_only_ok.rs
@@ -1,0 +1,58 @@
+use std::fmt;
+
+use xtra::prelude::MessageChannel;
+use xtra::{Handler, KeepRunning};
+
+/// A forwarding actor that only forwards [`Result::Ok`] values and shuts itself down upon the first
+/// error.
+pub struct Actor<M> {
+    forward: Box<dyn MessageChannel<M>>,
+}
+
+impl<M> Actor<M> {
+    pub fn new(forward: Box<dyn MessageChannel<M>>) -> Self {
+        Self { forward }
+    }
+}
+
+pub struct Message<TOk, TErr>(pub Result<TOk, TErr>);
+
+impl<TOk, TErr> xtra::Message for Message<TOk, TErr>
+where
+    TOk: Send + 'static,
+    TErr: Send + 'static,
+{
+    type Result = KeepRunning;
+}
+
+#[async_trait::async_trait]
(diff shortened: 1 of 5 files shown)


da-kami (approved):
nifty, but a bit hard to understand

daemon/src/forward_only_ok.rs
```diff
@@ -0,0 +1,58 @@
...
+{
+    async fn handle(
+        &mut self,
+        Message(result): Message<TOk, TErr>,
+        _: &mut xtra::Context<Self>,
+    ) -> KeepRunning {
+        let ok = match result {
+            Ok(ok) => ok,
+            Err(e) => {
+                tracing::error!("Stopping forwarding due to error: {}", e);
```
bonomat, line 42: Are you happy with losing the cause? 

```suggestion
                tracing::error!("Stopping forwarding due to error: {:#}", e);
```

bonomat (approved):
Nice, thanks.

luckysori (approved):

daemon/src/forward_only_ok.rs
thomaseizinger: I don't mind using alternate display, will not make a difference for most error types. Rendering causes of errors using the _alternate_ sigil is a feature of `anyhow::Error`, not of the `std::error::Error` type.

So unless `TErr` is `anyhow::Error`, it won't make a difference.

Issue #231, opened 2021-10-07
Actor-based integration tests


thomaseizinger:
Currently, we have to manually test the entire application if we want to verify that it still works.

We need automated tests that operate on an actor level.

Some ideas:

- Create an abstraction called an "actor system" that initializes and wires a set of actors together
- Said actor system should be generic over blockchain communication
- Communication between maker and taker could be generic over IO (by directly wiring the two actors together) or also happen on a random port (listen on 0)
- Tests **must** be fast, no communication with blockchain nodes. This will likely require the wallet to be generic over the client. Possible introducing a dedicated `Wallet` actor might also help as it would allow us to swap out the actor that handles blockchain communication for the tests

thomaseizinger:
In order to resolve this we may need to choose, which actors we want to "stub" in the tests and thus, have only those addresses be generic over the actor type.

In other words, the `monitor::Actor` can likely know that it will always send message to the `cfd::Actor` whereas for the tests, we probably want to stub the `monitor::Actor` within `cfd::Actor` and thus have that address be generic.

I guess this will require two monitoring actors instead of one (one per daemon) but we can just share the code on a lower-level and have the actor just be glue code.

da-kami:
Everything concerning the test framework as described in this ticket is done with merging #428 - however, we did not add a whole lot of tests yet. 

@klochowicz should we close this ticket or just keep it open until we have some more test coverage?

klochowicz:
I would be in favour of closing this ticket and creating a few dedicated, actionable ones for increasing the test coverage in the areas we care about (e.g. collab settlement, non-collab settlement, restarting during various stages)

PR #313, opened 2021-10-14
Replace `Address` with `MessageChannel` in some key places

luckysori:
We may need to use `MessageChannel` instead of `Address` elsewhere (perhaps everywhere), but these definitely need to change.

---

Work towards #231.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -14,6 +14,7 @@ use std::convert::{TryFrom, TryInto};
 use std::fmt;
 use std::marker::Send;
 use std::ops::{Add, RangeInclusive};
+use xtra::prelude::StrongMessageChannel;
 
 const FINALITY_CONFIRMATIONS: u32 = 1;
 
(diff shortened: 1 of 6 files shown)


daemon/src/monitor.rs
```diff
@@ -33,26 +34,19 @@ pub struct MonitorParams {
...
 
-impl<T> Actor<T, bdk::electrum_client::Client>
-where
-    T: xtra::Actor + xtra::Handler<Event>,
-{
+impl Actor<bdk::electrum_client::Client> {
     pub async fn new(
         electrum_rpc_url: &str,
-        cfd_actor_addr: xtra::Address<T>,
+        event_channel: Box<dyn StrongMessageChannel<Event>>,
```
thomaseizinger: Perhaps change this to `impl StrongMessageChannel<Event>`, then you can pass in an address and do the boxing inside.

daemon/src/oracle.rs
```diff
@@ -9,13 +9,13 @@ use serde::Deserialize;
...
+use xtra::prelude::StrongMessageChannel;
 
-pub struct Actor<CFD, M> {
+pub struct Actor {
     announcements: HashMap<BitMexPriceEventId, (OffsetDateTime, Vec<schnorrsig::PublicKey>)>,
     pending_announcements: HashSet<BitMexPriceEventId>,
     pending_attestations: HashSet<BitMexPriceEventId>,
-    cfd_actor_address: xtra::Address<CFD>,
-    monitor_actor_address: xtra::Address<M>,
+    attestation_channels: [Box<dyn StrongMessageChannel<Attestation>>; 2],
```
thomaseizinger, line 18: There is an actor design pattern called "fan-out" where you have one actor that forwards to N other actors. We could make use of that here.

thomaseizinger: I'll send a follow-up.

thomaseizinger (approved):

daemon/src/monitor.rs
luckysori: I'll keep the `Box`es in the `oracle::Actor` constructor since there's two of them and you're probably touching that in your follow-up PR.

PR #316, opened 2021-10-14
Provide more feedback to the user during collaborative settlement

### LINKED ISSUE #309: Provide feedback in the frontend after agreeing on collaborative settlement
After agreeing on a collaborative settlement we should change the Cfd state visible to the user (and disable the approve/ reject settlement actions in the maker). 

After the transaction gets published, the Cfd correctly moves to 'closed', but before that user can get pretty confused.

klochowicz:
Fixes #309

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -307,6 +307,8 @@ impl Actor {
             TimestampedTransaction::new(tx.clone(), dlc.script_pubkey_for(cfd.role())),
         ))?;
         insert_new_cfd_state_by_order_id(cfd.order.id, &cfd.state, &mut conn).await?;
+        self.cfd_feed_actor_inbox
+            .send(load_all_cfds(&mut conn).await?)?;
 
         let spend_tx = dlc.finalize_spend_transaction((tx, sig_maker), sig_taker)?;
 
@@ -316,12 +318,6 @@ impl Actor {
             .await
             .context("Broadcasting spend transaction")?;
         tracing::info!("Close transaction published with txid {}", txid);
-
-        cfd.handle(CfdStateChangeEvent::CloseSent(TimestampedTransaction::new(
-            spend_tx,
-            dlc.script_pubkey_for(cfd.role()),
-        )))?;
-        insert_new_cfd_state_by_order_id(cfd.order.id, &cfd.state, &mut conn).await?;
 
         self.current_agreed_proposals
             .remove(&order_id)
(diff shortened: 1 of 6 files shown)


klochowicz:
there will be one more commit with TypeScript code, showing the state in the frontend.

thomaseizinger (approved):
Nice!

The 2nd commit goes towards https://github.com/comit-network/hermes/issues/290. Good work!

da-kami (changes requested):
I noticed that we are missing trying to re-broadcast the collaborative settlement tx upon startup in `housekeeping::rebroadcast_transactions`. I think this can be cleanly added on top, but am not sure about the removal about `PendingClose` for that - we might need that state to be able to know we have to re-boradcast!

Could you maybe add the re-braodcast (for the maker only...?) and see how that influences removing `PendingClose`?

da-kami (approved):
> I noticed that we are missing trying to re-broadcast the collaborative settlement tx upon startup in housekeeping::rebroadcast_transactions. I think this can be cleanly added on top, but am not sure about the removal about PendingClose for that - we might need that state to be able to know we have to re-boradcast!
>
> Could you maybe add the re-braodcast (for the maker only...?) and see how that influences removing PendingClose?

After some discussion: We actually don't need re-broadcast here. Reasoning goes like this:
If we start up and see `PendingClose` the user will have to check on the blockchain explorer using the txid link. If the transaction is there then just wait for finality. Otherwise publish commit (both should have the commit action still).
We cannot decide to just force a re-publication of the transaction in housekeeping because the collaborative close is related to a specific time where they agree. If we start up at a later point we don't know if the maker still wants to allow it.

klochowicz:
> > I noticed that we are missing trying to re-broadcast the collaborative settlement tx upon startup in housekeeping::rebroadcast_transactions. I think this can be cleanly added on top, but am not sure about the removal about PendingClose for that - we might need that state to be able to know we have to re-boradcast!
> > Could you maybe add the re-braodcast (for the maker only...?) and see how that influences removing PendingClose?
> 
> After some discussion: We actually don't need re-broadcast here. Reasoning goes like this: If we start up and see `PendingClose` the user will have to check on the blockchain explorer using the txid link. If the transaction is there then just wait for finality. Otherwise publish commit (both should have the commit action still). We cannot decide to just force a re-publication of the transaction in housekeeping because the collaborative close is related to a specific time where they agree. If we start up at a later point we don't know if the maker still wants to allow it.

I agree that we don't need to re-broadcast, but if somehow we changed our mind it is still possible without `PendingClose` state in the model. It would be sufficient to derive it by checking whether we're in `Open` state with `Some(collaborative_close)`.

da-kami:
related: https://github.com/comit-network/hermes/issues/320

PR #318, opened 2021-10-14
Introduce a `fan_out::Actor` instead of iterating in the `oracle::Actor`

thomaseizinger:
A fan-out actor takes an incoming message and forwards it to a set
of actors. This allows our `oracle::Actor` to be unaware, how many
other actors are interested in the attestation, thus simplifying it.


luckysori (approved):
Nice improvement!

PR #321, opened 2021-10-14
Use `MessageChannels` in `maker_inc_connections::Actor`

luckysori:
We would have liked to attach the stream of `TakerStreamMessage`s directly to the `MessageChannel`, but we failed and instead decided to let the `maker_inc_connections::Actor` send these messages to itself and then forward them to the `maker_cfd::Actor`.

---

I wasted a lot of time trying to send the message directly to the `maker_cfd::Actor`. I trust that @thomaseizinger can find a way, but I had to give up :disappointed:

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -1,3 +1,4 @@
+use crate::maker_cfd::{NewTakerOnline, TakerStreamMessage};
 use crate::model::cfd::{Order, OrderId};
 use crate::model::{BitMexPriceEventId, TakerId};
 use crate::{log_error, maker_cfd, send_to_socket, wire};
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
> We would have liked to attach the stream of `TakerStreamMessage`s directly to the `MessageChannel`, but we failed and instead decided to let the `maker_inc_connections::Actor` send these messages to itself and then forward them to the `maker_cfd::Actor`.

I don't think you can do anything about this. If I understand [this](https://users.rust-lang.org/t/is-there-a-way-to-move-a-trait-object/707) correctly, then `xtra` would have to change the type-signature of `attach_stream` to `fn attach_stream(self: Box<Self>)` for this function to work. It appears that this was just never tested and never worked as intended :)

Pushed two follow-up commits.

luckysori:
> I don't think you can do anything about this. If I understand [this](https://users.rust-lang.org/t/is-there-a-way-to-move-a-trait-object/707) correctly, then `xtra` would have to change the type-signature of `attach_stream` to `fn attach_stream(self: Box<Self>)` for this function to work. It appears that this was just never tested and never worked as intended :)
 
I reached similar conclusions and tried to add 

```rust
impl<M: Message> MessageChannel<M> for Box<dyn MessageChannel<M>>
```
 
 but that required `Self: Sealed` which caused problems when trying to use it in `daemon`.
 
> Pushed two follow-up commits.

Thanks!

daemon/src/maker_inc_connections.rs
```diff
@@ -69,11 +69,14 @@ pub struct Actor {
 }
 
 impl Actor {
-    pub fn new(cfd_actor: &Address<maker_cfd::Actor>) -> Self {
+    pub fn new(
+        new_taker_channel: &impl MessageChannel<NewTakerOnline>,
+        taker_msg_channel: &impl MessageChannel<TakerStreamMessage>,
+    ) -> Self {
```
luckysori, line 75: :+1:

thomaseizinger:
> I reached similar conclusions and tried to add
> 
> ```rust
> impl<M: Message> MessageChannel<M> for Box<dyn MessageChannel<M>>
> ```
> 
> but that required `Self: Sealed` which caused problems when trying to use it in `daemon`.

Would that be necessary? `Box` implements `Deref<Target = T>`.

luckysori:
> Would that be necessary? `Box` implements `Deref<Target = T>`.

Maybe give it ago yourself, I'm sure you can figure something out :stuck_out_tongue:

PR #322, opened 2021-10-14
Cleanup in Cfd actors

klochowicz:
I started out trying to fix a common gotcha (forgetting to refresh Cfd feed),
and then realised that I was not far off of addressing a bunch of TODOs in the code

### DIFF
--- a/daemon/src/actors.rs
+++ b/daemon/src/actors.rs
@@ -1,10 +1,3 @@
-use crate::{
-    db,
-    model::cfd::{Cfd, CfdState, OrderId},
-};
-use sqlx::{pool::PoolConnection, Sqlite};
-use tokio::sync::watch;
-
 /// Wrapper for handlers to log errors
 #[macro_export]
 macro_rules! log_error {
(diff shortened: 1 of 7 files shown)


thomaseizinger (changes requested):
The last commit (3e021f3364a03e48db2409a1df86871ece8c9d7b) is not very Rust idiomatic. Traits are most useful if you have code that abstracts over them, not to share code between two types.

It would be better to use composition for those scenarios.

In general, I think the better approach would be if we find a way to split the CfdActors up into multiple such that both binaries can reuse individual actors.

Thanks for the effort and attempt but I think it would be better to drop the last commit for now.

klochowicz:
> The last commit ([3e021f3](https://github.com/comit-network/hermes/commit/3e021f3364a03e48db2409a1df86871ece8c9d7b)) is not very Rust idiomatic. Traits are most useful if you have code that abstracts over them, not to share code between two types.
> 
> It would be better to use composition for those scenarios.
> 
> In general, I think the better approach would be if we find a way to split the CfdActors up into multiple such that both binaries can reuse individual actors.
> 
> Thanks for the effort and attempt but I think it would be better to drop the last commit for now.

Happy to remove it! I wasn't sure whether this was a good approach or not to use a trait, but I wanted to give it a shot for learning purposes :) composition still achieves the purpose of removing the duplication, I'll update the PR.

da-kami (approved):
boom!

PR #327, opened 2021-10-14
Failing test for db

da-kami:
- [x] Failing test
- [ ] Fix queries :)


da-kami:
Note: we might even want to check that only one entry is returned, would make the test even more sound. The fact that we fail on first is because the order is by id I think, and the older entry is thus first.

thomaseizinger:
Superseded by #337.

PR #331, opened 2021-10-14
Fix default quantity in UI

### LINKED ISSUE #330: If taker does not change quantity, it's 0
If taker does not change quantity, it's 0 but the UI shows default value. This makes the maker reject the order immediately as the amount is out of bounds.

bonomat:
@thomaseizinger : I'm not sure if I understand it correctly. Can I do https://github.com/comit-network/hermes/commit/ed2ec8eb4926c9cf27d51a4659a8a872de4ad54e or is this not allowed because I change the variable independent of what is in the state? 

Alternative solution is the second commit https://github.com/comit-network/hermes/commit/ac8d6e191cc313f46ead18a5950cb0a097966f3a

resolves #330

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -527,6 +527,12 @@ impl Actor {
 
         // 2. check if order has acceptable amounts
         if quantity < current_order.min_quantity || quantity > current_order.max_quantity {
+            tracing::warn!(
+                "Order rejected because quantity {} was out of bounds. It was either <{} or >{}",
+                quantity,
+                current_order.min_quantity,
+                current_order.max_quantity
+            );
             self.takers
                 .do_send_async(maker_inc_connections::TakerMessage {
                     taker_id,
--- a/frontend/src/TakerApp.tsx
+++ b/frontend/src/TakerApp.tsx
@@ -73,7 +73,7 @@ export default function App() {
     let [margin, setMargin] = useState("0");
     let [userHasEdited, setUserHasEdited] = useState(false);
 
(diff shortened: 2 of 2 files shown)


thomaseizinger:
> @thomaseizinger : I'm not sure if I understand it correctly. Can I do [ed2ec8e](https://github.com/comit-network/hermes/commit/ed2ec8eb4926c9cf27d51a4659a8a872de4ad54e) or is this not allowed because I change the variable independent of what is in the state?
> 
> Alternative solution is the second commit [ac8d6e1](https://github.com/comit-network/hermes/commit/ac8d6e191cc313f46ead18a5950cb0a097966f3a)
> 
> resolves #330

It is not harmful but also not ideal. I would suggest renaming `quantityToShow` to something like `effectiveQuantity` and use that in the payload instead of `quantity`. I think that should also fix your problem.

bonomat:
> > @thomaseizinger : I'm not sure if I understand it correctly. Can I do [ed2ec8e](https://github.com/comit-network/hermes/commit/ed2ec8eb4926c9cf27d51a4659a8a872de4ad54e) or is this not allowed because I change the variable independent of what is in the state?
> > Alternative solution is the second commit [ac8d6e1](https://github.com/comit-network/hermes/commit/ac8d6e191cc313f46ead18a5950cb0a097966f3a)
> > resolves #330
> 
> It is not harmful but also not ideal. I would suggest renaming `quantityToShow` to something like `effectiveQuantity` and use that in the payload instead of `quantity`. I think that should also fix your problem.

please have a look :)

thomaseizinger (approved):
LGTM!

bonomat:
bors +

PR #333, opened 2021-10-14
Use `MessageChannel` in `maker_routes`

luckysori:
More work towards #231.

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -17,6 +17,7 @@ use std::borrow::Cow;
 use std::path::PathBuf;
 use tokio::select;
 use tokio::sync::watch;
+use xtra::prelude::MessageChannel;
 use xtra::Address;
 
 #[rocket::get("/feed")]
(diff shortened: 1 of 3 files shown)


daemon/src/routes_maker.rs
```diff
@@ -144,55 +143,48 @@ pub struct PromptAuthentication {
 }
 
 #[rocket::post("/cfd/<id>/<action>")]
-pub async fn post_cfd_action(
+pub fn post_cfd_action(
     id: OrderId,
     action: CfdAction,
-    cfd_actor_address: &State<Address<maker_cfd::Actor>>,
+    user_choice_channel: &State<Box<dyn MessageChannel<maker_cfd::UserChoice>>>,
```
thomaseizinger: Should we follow the other naming here and call this `maker_cfd::CfdAction`?

thomaseizinger (approved):

daemon/src/routes_maker.rs
luckysori: Are you referring to renaming `maker_cfd::UserChoice` to `maker_cfd::CfdAction`? I think `UserChoice` is clearer, but if we have a convention in place I'm happy to change it.

thomaseizinger: The HTTP endpoint uses `CfdAction` for the exact same enum, minus the `OrderId` field. I'd suggest following the naming here for now and perhaps rename both at a later point.

PR #336, opened 2021-10-15
Rejection always remove order

### LINKED ISSUE #324: Taker can take offer with same id twice
Not sure how I managed to do this, but it seems the maker's offer is not always removed correctly. 

Saw this in the taker logs, which means that we tried to take the same offer twice:

```
2021-10-14 20:24:47 ERROR Message handler failed: error returned from database: UNIQUE constraint failed: cfds.order_uuid: UNIQUE constraint failed: cfds.order_uuid
2021-10-14 20:24:58  INFO POST /api/calculate/margin text/plain:  
```

The maker should not allow this.

da-kami:
Fixes https://github.com/comit-network/hermes/issues/324 https://github.com/comit-network/hermes/issues/323

Since https://github.com/comit-network/hermes/issues/323 could not be reproduce I assume it is (already) fixed. Might be that https://github.com/comit-network/hermes/issues/323 was caused by a combination of problems.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -16,6 +16,8 @@ use bdk::bitcoin::secp256k1::schnorrsig;
 use cfd_protocol::secp256k1_zkp::Signature;
 use futures::channel::mpsc;
 use futures::{future, SinkExt};
+use rocket_db_pools::sqlx::Sqlite;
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Nice. LGTM.

thomaseizinger (approved):

Issue #200, opened 2021-10-05
Cleanup cfd state transitions


da-kami:
All cfd state transitions should be done in the `handle` function. 

Proposed changes:

* Move **all** state transitions onto the cfd's `handle` function. At the moment there are still several early stage states that are just transitioned in the cfd actor without the cfd `handle` function being involved. 
* The cfd `handle` function should return a new `Cfd` (consume self) instead of the new state.
* `insert_state_by_order_id` should be removed in favour of always updating based on the cfd. This is less efficient but also less error prone, because we can be sure what we saved is what we have.

da-kami:
Closing because it becomes irrelevant with #290 

Below part is basically covered by https://github.com/comit-network/hermes/pull/337

> The cfd handle function should return a new Cfd (consume self) instead of the new state.
> insert_state_by_order_id should be removed in favour of always updating based on the cfd. This is less efficient but also less error prone, because we can be sure what we saved is what we have.

PR #337, opened 2021-10-15
Better DB tests and various cleanups

thomaseizinger:
- Extend DB test suite
- Directly return `PoolConnection`
- Don't use Default for test instance
- Take `Cfd` by reference if we don't need to consume it
- Use pretty assertions
- Make DB tests more readable by reducing boilerplate
- Refactor `insert_new_cfd_state_by_order_id` to `append_cfd_state`
- Reduce boilerplate in constructing `CfdStateCommon`
- Make it less verbose to construct simple states
- Add test for a variety of state updates
- Run tests against in-memory DB

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -495,6 +495,7 @@ mod tests {
     use std::fs::File;
 
     use pretty_assertions::assert_eq;
+    use rand::Rng;
     use rust_decimal_macros::dec;
     use sqlx::SqlitePool;
     use tempfile::tempdir;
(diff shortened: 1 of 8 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -421,19 +415,16 @@ impl Actor {
...
-        insert_new_cfd_state_by_order_id(
-            order_id,
-            &CfdState::PendingOpen {
-                common: CfdStateCommon {
-                    transition_timestamp: SystemTime::now(),
-                },
-                dlc: dlc.clone(),
-                attestation: None,
+        let mut cfd = load_cfd_by_order_id(order_id, &mut conn).await?;
+        cfd.state = CfdState::PendingOpen {
```
da-kami, line 403: Note: Could be moved inside `handle` towards https://github.com/comit-network/hermes/issues/200 - not sure if it's worth it though, given that we might change the state model to https://github.com/comit-network/hermes/issues/290

On that note: Shall we close https://github.com/comit-network/hermes/issues/200 because https://github.com/comit-network/hermes/issues/290 will likely make it obsolete?

daemon/src/db.rs
```diff
@@ -647,14 +644,7 @@ mod tests {
...
     async fn setup_test_db() -> PoolConnection<Sqlite> {
-        let temp_db = tempdir().unwrap().into_path().join("tempdb");
-
-        // file has to exist in order to connect with sqlite
-        File::create(temp_db.clone()).unwrap();
-
-        let pool = SqlitePool::connect(format!("sqlite:{}", temp_db.display()).as_str())
-            .await
-            .unwrap();
+        let pool = SqlitePool::connect(":memory:").await.unwrap();
```
da-kami, line 647: Can we remove `tempfile` dependency now?

da-kami (approved):

daemon/src/db.rs
thomaseizinger: We might be able to!

daemon/src/maker_cfd.rs
thomaseizinger: Closing #200 sounds good to me.

da-kami:
I like the in-memory database for testing, but we should be aware, that we won't be able to "have a look" what is in the test database then:

> If the filename is ":memory:", then a private, temporary in-memory database is created for the connection. This in-memory database will vanish when the database connection is closed. 

http://www.sqlite.org/c3ref/open.html

When I tested initially I sometimes did have a look what was going on in the temporary one, but I guess we might not need that anymore. Just raising awareness :)

thomaseizinger:
> I like the in-memory database for testing, but we should be aware, that we won't be able to "have a look" what is in the database then:
> 
> > If the filename is ":memory:", then a private, temporary in-memory database is created for the connection. This in-memory database will vanish when the database connection is closed.
> 
> http://www.sqlite.org/c3ref/open.html
> 
> When I tested initially I sometimes did have a look what was going on in the temporary one, but I guess we might not need that anymore. Just raising awareness :)

Yep, thought of that but came to the same conclusion. Plus can always replace it with a path for temporary testing.

daemon/src/maker_cfd.rs
```diff
@@ -513,20 +497,7 @@ impl Actor {
...
-            tracing::warn!(
-                "Order rejected because quantity {} was out of bounds. It was either <{} or >{}",
-                quantity,
-                current_order.min_quantity,
-                current_order.max_quantity
-            );
-
-            self.reject_order(taker_id, order_id, conn).await?;
-            return Ok(());
-        }
```
thomaseizinger, line 527: @da-kami @bonomat I had to move this down because one of my refactorings changes the function signature of `append_cfd_state` to take the entire `Cfd`. Thus we need to have one available here to pass it down.

I think that actually uncovered a potential problem: I think calling `reject_order` in there never worked because we would not have had a `Cfd` in the database that we could append the state to.

PR #338, opened 2021-10-15
Refactor database queries, take #2

thomaseizinger:
Fix up SQL queries

This PR does a few things:
* cleans up the SQL to make the queries clearer in terms of intent, as well as eliminating the use of an extra transaction in some write queries.
* adds some additional testing
* (mostly) eliminates the use of `serde_json::to_string()`, making the data columns behave in a more sane manner

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1,11 +1,14 @@
-use crate::model::cfd::{Cfd, CfdState, Order, OrderId, Origin};
-use crate::model::{BitMexPriceEventId, Leverage, Position};
+use crate::model::cfd::{Cfd, CfdState, Order, OrderId};
+use crate::model::{BitMexPriceEventId, Usd};
 use anyhow::{Context, Result};
 use rocket_db_pools::sqlx;
+use rust_decimal::Decimal;
 use sqlx::pool::PoolConnection;
-use sqlx::{Acquire, Sqlite, SqlitePool};
-use std::convert::TryInto;
+use sqlx::{Sqlite, SqlitePool};
 use std::mem;
+use std::str::FromStr;
+use std::time::SystemTime;
+use time::{Duration, OffsetDateTime};
 
 pub async fn run_migrations(pool: &SqlitePool) -> anyhow::Result<()> {
     sqlx::migrate!("./migrations").run(pool).await?;
@@ -13,49 +16,48 @@ pub async fn run_migrations(pool: &SqlitePool) -> anyhow::Result<()> {
 }
 
 pub async fn insert_order(order: &Order, conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<()> {
-    let uuid = serde_json::to_string(&order.id).unwrap();
-    let trading_pair = serde_json::to_string(&order.trading_pair).unwrap();
-    let position = serde_json::to_string(&order.position).unwrap();
-    let initial_price = serde_json::to_string(&order.price).unwrap();
-    let min_quantity = serde_json::to_string(&order.min_quantity).unwrap();
-    let max_quantity = serde_json::to_string(&order.max_quantity).unwrap();
-    let leverage = order.leverage.0;
-    let liquidation_price = serde_json::to_string(&order.liquidation_price).unwrap();
-    let creation_timestamp = serde_json::to_string(&order.creation_timestamp).unwrap();
-    let term = serde_json::to_string(&order.term).unwrap();
-    let origin = serde_json::to_string(&order.origin).unwrap();
-    let oracle_event_id = order.oracle_event_id.to_string();
-
-    sqlx::query!(
-        r#"
-            insert into orders (
-                uuid,
-                trading_pair,
-                position,
-                initial_price,
-                min_quantity,
-                max_quantity,
-                leverage,
-                liquidation_price,
-                creation_timestamp,
-                term,
-                origin,
-                oracle_event_id
-            ) values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?);
-            "#,
-        uuid,
-        trading_pair,
-        position,
-        initial_price,
-        min_quantity,
-        max_quantity,
-        leverage,
-        liquidation_price,
-        creation_timestamp,
-        term,
-        origin,
-        oracle_event_id
+    sqlx::query(
+        r#"insert into orders (
+            uuid,
+            trading_pair,
+            position,
+            initial_price,
+            min_quantity,
+            max_quantity,
+            leverage,
+            liquidation_price,
+            creation_timestamp_seconds,
+            creation_timestamp_nanoseconds,
+            term_seconds,
+            term_nanoseconds,
+            origin,
+            oracle_event_id
+        ) values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14)"#,
     )
+    .bind(&order.id)
+    .bind(&order.trading_pair)
+    .bind(&order.position)
+    .bind(&order.price.to_string())
+    .bind(&order.min_quantity.to_string())
+    .bind(&order.max_quantity.to_string())
+    .bind(order.leverage.0)
+    .bind(&order.liquidation_price.to_string())
+    .bind(
+        order
+            .creation_timestamp
+            .duration_since(SystemTime::UNIX_EPOCH)?
+            .as_secs() as i64,
+    )
+    .bind(
+        order
+            .creation_timestamp
+            .duration_since(SystemTime::UNIX_EPOCH)?
+            .subsec_nanos() as i32,
+    )
+    .bind(&order.term.whole_seconds())
+    .bind(&order.term.subsec_nanoseconds())
+    .bind(&order.origin)
+    .bind(&order.oracle_event_id.to_string())
     .execute(conn)
     .await?;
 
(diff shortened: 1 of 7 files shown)


daemon/src/db.rs
```diff
@@ -1,61 +1,63 @@
...
+        order
+            .creation_timestamp
+            .duration_since(SystemTime::UNIX_EPOCH)?
+            .as_secs() as i64,
+    )
+    .bind(
+        order
+            .creation_timestamp
+            .duration_since(SystemTime::UNIX_EPOCH)?
+            .subsec_nanos() as i32,
```
da-kami, line 55: ❓ Just out of curiosity: Are nanos actually represented as `i32` or why are we not using `i64` here like for `secs`?

daemon/src/db.rs
```diff
@@ -66,97 +68,83 @@ pub async fn load_order_by_id(
...
-    let cfd_state = serde_json::to_string(&cfd.state)?;
-
-    // save cfd + state in a transaction to make sure the state is only inserted if the cfd was
-    // inserted
+        insert into cfds (
+            order_id,
+            order_uuid,
+            quantity_usd
+        )
+        select
```
da-kami, line 122: ❓ Why are the selects after the insert needed? Is this here to check that the insert worked - i.e. testing it on the fly?

Interesting pattern - I find it adds some complexity when looking at the coee but don't mind much keeping that.

daemon/src/db.rs
```diff
@@ -183,7 +171,7 @@ pub async fn append_cfd_state(cfd: &Cfd, conn: &mut PoolConnection<Sqlite>) -> a
         insert into cfd_states (
             cfd_id,
             state
-        ) values (?, ?);
+        ) values ($1, $2);
```
da-kami, line 172: ❓ if we have validation `select`s above, why not here as well?

daemon/src/db.rs
```diff
@@ -246,159 +232,205 @@ pub async fn load_cfd_by_order_id(
...
+        cfd as (
+            select
+                ord.order_id,
+                id as cfd_id,
+                quantity_usd
+            from cfds
+                inner join ord on ord.order_id = id
+        ),
+
+        state as (
```
da-kami, line 264: nit: I think this should be called `cfds_with_latest_state` - because in my understanding that is what this represents.

daemon/src/db.rs
```diff
@@ -246,159 +232,205 @@ pub async fn load_cfd_by_order_id(
...
+            select
+                id as state_id,
+                cfd.order_id,
+                cfd.quantity_usd,
+                state
+            from cfd_states
+                inner join cfd on cfd.cfd_id = cfd_states.cfd_id
+            where id in (
```
da-kami, line 272: ❓ Why is this `where id in` and not `where id = `? 
`max` should always return exactly one, can sqlite not handle that?

daemon/src/db.rs
```diff
@@ -246,159 +232,205 @@ pub async fn load_cfd_by_order_id(
...
-            orders.creation_timestamp as creation_timestamp,
-            orders.term as term,
-            orders.oracle_event_id,
-            cfds.quantity_usd as quantity_usd,
-            cfd_states.state as state
-        from cfds as cfds
-        inner join orders as orders on cfds.order_id = orders.id
-        inner join cfd_states as cfd_states on cfd_states.cfd_id = cfds.id
-        where cfd_states.state in (
+        with ord as (
```
da-kami, line 235: nit: could we name these placeholders like the tables that they represent? 
I find it hard to map from table to placeholder to next statement (...).

daemon/src/db.rs
```diff
@@ -246,159 +232,205 @@ pub async fn load_cfd_by_order_id(
...
+            ord.term_nanos as "term_nanos: i32",
+            ord.origin as "origin: crate::model::cfd::Origin",
+            ord.oracle_event_id,
+            state.quantity_usd,
+            state.state
+
+        from ord
+            inner join state on state.order_id = ord.order_id
+
+        where ord.uuid = $1
```
da-kami, line 301: Not part of this PR, but thought I mention it:

I was unable to figure out a nice way to reuse statement code. These huge statements basically do the same thing besides the final `where` clause`. Can you think of a way to re-use the code - do you think it would be a good idea?

da-kami (approved):
Looking good, just a few questions and minor suggestions.

daemon/src/db.rs
```diff
@@ -408,88 +440,119 @@ pub async fn load_cfds_by_oracle_event_id(
     oracle_event_id: BitMexPriceEventId,
     conn: &mut PoolConnection<Sqlite>,
 ) -> anyhow::Result<Vec<Cfd>> {
-    let oracle_event_id = oracle_event_id.to_string();
-
+    let event_id = oracle_event_id.to_string();
```
da-kami, line 441: I extended the tests for `load_all_cfds` - but given that these other statements duplicate the complex join SQL code we might also want to add more tests for loading by `oracle_event_id` and `order_id` to make sure we cover someone only changing one of these statements. (unless we can achieve partly re-using the SQL statement code).

daemon/src/db.rs
thomaseizinger: The only thing I can think of is adding SQL views to the DB.

daemon/src/db.rs
DeliciousHair: `i32`, yes. And not for completely arbitrary reasons either!

```rust
/// A span of time with nanosecond precision.
///
/// Each `Duration` is composed of a whole number of seconds and a fractional part represented in
/// nanoseconds.
///
/// This implementation allows for negative durations, unlike [`core::time::Duration`].
#[derive(Clone, Copy, Default, PartialEq, Eq, Hash, PartialOrd, Ord)]
pub struct Duration {
    /// Number of whole seconds.
    seconds: i64,
    /// Number of nanoseconds within the second. The sign always matches the `seconds` field.
    nanoseconds: i32, // always -10^9 < nanoseconds < 10^9
    #[allow(clippy::missing_docs_in_private_items)]
    padding: Padding,
```

daemon/src/db.rs
DeliciousHair: It is in fact [very standard SQL](https://www.w3schools.com/sql/sql_insert_into_select.asp). Basically, you're pointing the output data of the `SELECT` query to be the input data for the `INSERT` query.

daemon/src/db.rs
DeliciousHair: FWIW I've tried this already with no luck. It may be possible, but it's not simple nor obvious.

daemon/src/db.rs
DeliciousHair: Because we're not returning anything back to `rust` here.

daemon/src/db.rs
DeliciousHair: In this specific case, no. This is on account of the fact that `ORDER` is a protected keyword in SQL.

daemon/src/db.rs
DeliciousHair: Because the `GROUP BY` clause does the grouping, and the `MAX()` operation is applied to the group. Equality might work, didn't try. `MAX()` is explicitly what we want on a per-group basis.

daemon/src/db.rs
DeliciousHair: Er, scratch that, misread the question. The reason is because the function previously had `#[allow(dead_code)]`, so it was simply ignored. You're correct, should fix that.

daemon/src/db.rs
da-kami: cool, I did not know that!

daemon/src/db.rs
da-kami: I was pretty sure you had tried already :) - guess it's not worth to invest more time. It might be a good idea to extend the tests of `load_cfds_by_oracle_event_id` and `load_cfds_by_order_id` to be sure we don't into trouble if we ever change the statements. Either way, approved.

daemon/src/db.rs
thomaseizinger: I like this idea.

daemon/src/db.rs
```diff
@@ -408,88 +438,119 @@ pub async fn load_cfds_by_oracle_event_id(
...
     Ok(cfds)
 }
 
+fn convert_to_system_time(row_secs: i64, row_nanos: i32) -> Result<SystemTime> {
+    let secs = row_secs as i128;
+    let nanos = row_nanos as i128;
+    let offset_dt = OffsetDateTime::from_unix_timestamp_nanos(secs * 1_000_000_000 + nanos)?;
+
+    Ok(SystemTime::from(offset_dt))
+}
```
thomaseizinger, line 552: Can you create an issue to get rid of the nanoseconds for all these timestamps? Not worth the complexity IMO.

daemon/src/model.rs
```diff
@@ -109,7 +110,17 @@ pub struct WalletInfo {
...
+    Clone,
+    Copy,
+    SerializeDisplay,
+    DeserializeFromStr,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    sqlx::Type,
```
thomaseizinger: Adding this without `transparent` or another annotation does not do anything on structs as far as I know.

daemon/src/db.rs
thomaseizinger: Agreed!

thomaseizinger (commented):
Thank you @DeliciousHair! I can't request changes on my own PR, so I'll just post a comment.

Couple of things:

- The commit message needs some more work to follow https://chris.beams.io/posts/git-commit/.
- Daniel made a few good comments, please address them.
- There is an unnecessary derive on `BitMexPriceEventId`
- I would like a follow-up issue on the nanosecond topic.

daemon/src/db.rs
DeliciousHair: The reason it was left in is I am unclear as to how the timestamps are used initially, so that one that is created via `Systemtime::now()` and one retrieved from the DB will be unable to be matched without them, **if** that's a thing. In other words, the extra complexity here would be (potentially) replaced with the opposite complexity somewhere else.

Can do this of course, just need to be sure that it won't create issues somewhere else.

thomaseizinger: Yep that makes a lot of sense. It needs to be thought through, hence it would be good to have a follow-up issue!

daemon/src/db.rs
DeliciousHair: Added to `test_multiple_cfd_updates_per_cfd()`

daemon/src/db.rs
da-kami: was this not actioned on purpose? would be nice to say so. Resolved for me means that it was actioned.

da-kami (approved):

thomaseizinger (commented):
Looks good!

Did you do a manual test as well?

DeliciousHair:
> Did you do a manual test as well?

Yes, in addition to more EDA-style ad-hoc testing prior to the switch to an in-memory DB. All seems well-fine to me now, fingers crossed!

daemon/src/db.rs
DeliciousHair: While there is no general style guide that I am aware of for SQL in general (could be well wrong about this, I'm no expert!) there certainly isn't one for us internally. That being said, in practice most SQL aliases end up with really useful aliases like `a`, `b`, etc., which obviously suck but in my experience shorter phrases > longer ones as they tend to compound and things get progressively less readable with larger, more complex queries (which this most certainly is not, of course).

Thus, my action was to ignore the suggestion as I simply disagree with it is all; I think `state` is good alias as it captures the intent of the sub-query, that being that it is getting the "state" part of the overall query.

Apologies if you disagree, I read the above as a suggestion only is all, and since it didn't affect anything beyond stylistic concerns I chose to keep my original wording.

da-kami: all good, would just be nice to add a quick comment next time and then resolve the comment - then I know that it was on purpose and not just accidentally resolved.

PR #339, opened 2021-10-15
Make `maker_cfd::Actor` generic over other actor addresses

luckysori:
The diff is gonna be hard to review because I had to reorganise all the methods in separate `impl` blocks so that we only require the necessary trait bounds across the board.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -16,8 +16,8 @@ use bdk::bitcoin::secp256k1::schnorrsig;
 use cfd_protocol::secp256k1_zkp::Signature;
 use futures::channel::mpsc;
 use futures::{future, SinkExt};
-use rocket_db_pools::sqlx::Sqlite;
 use sqlx::pool::PoolConnection;
+use sqlx::Sqlite;
 use std::collections::HashMap;
 use std::time::SystemTime;
 use tokio::sync::watch;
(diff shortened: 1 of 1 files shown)


thomaseizinger:
What is the reason for not resorting to `MessageChannel`s here as well? Would it be too verbose due to the number of messages?

luckysori:
> What is the reason for not resorting to `MessageChannel`s here as well? Would it be too verbose due to the number of messages?

I think it would be a lot of channels, yep.

daemon/src/maker_cfd.rs
```diff
@@ -58,19 +58,19 @@ pub struct FromTaker {
     pub msg: wire::TakerToMaker,
 }
 
-pub struct Actor {
+pub struct Actor<O, M, T> {
```
da-kami, line 61:
```suggestion
pub struct Actor<O, M, G> {
```

🤪

da-kami (commented):
I actually find it somewhat nice to have the handling of specific messages grouped by the actors they relate to. But all that boilerplate code is just not very nice. @thomaseizinger any chance this can be optimized upstream, so we don't have to add all those `where` clauses?

daemon/src/maker_cfd.rs
luckysori: image

klochowicz:
> I actually find it somewhat nice to have the handling of specific messages grouped by the actors they relate to. But all that boilerplate code is just not very nice. @thomaseizinger any chance this can be optimized upstream, so we don't have to add all those `where` clauses?

we might give `spaad` crate another go (it's main goal is reducing this boilerplate...)

da-kami:
> > I actually find it somewhat nice to have the handling of specific messages grouped by the actors they relate to. But all that boilerplate code is just not very nice. @thomaseizinger any chance this can be optimized upstream, so we don't have to add all those `where` clauses?
> 
> we might give `spaad` crate another go (it's main goal is reducing this boilerplate...)

that does look noice! https://crates.io/crates/spaad

luckysori (commented):
LGTM. It's tricky to review, but I didn't notice anything wrong.

daemon/src/maker_cfd.rs
```diff
@@ -492,7 +495,7 @@ impl Actor {
             }
         };
 
-        // 2. Create a new CFD
+        // 2. Insert CFD in DB
```
luckysori, line 498: :question: Is this comment supposed to change?

luckysori (commented):
Did another pass with the two files side by side (the diff is impossible to follow in this case) and it still looks good. I think we can merge this.

daemon/src/maker_cfd.rs
thomaseizinger: Uhm I just made the same semantic change I think and just rewrote the comment.

thomaseizinger:
> Did another pass with the two files side by side (the diff is impossible to follow in this case) and it still looks good. I think we can merge this.

Go for it!

PR #342, opened 2021-10-15
Display available quantity in the taker

klochowicz:
It becomes important when someone else controls the offers from the maker.

This should be enough for now, but it would be nice to provide nicer validation field at some point.


da-kami (approved):
Sufficient for now I'd say.

In the long run we can invest time into relating the `balance` to the `margin` that is used when entering the `quantity` AND additionally guard against `min` and `max` quantity.

PR #343, opened 2021-10-15
All tooltips to action icons in CfdTable

klochowicz:
It makes it a bit easier to use, especially for newcomers


klochowicz:
we could capitalise the actions, but that is probably best done with some CSS rather than in JS/TS anyway.

bonomat (approved):
_A drop of water on a hot stone_

Definitely better, but I think we need to invest more into this :D

klochowicz:
> A drop of water on a hot stone

as they say it in Poland, "constant dropping wears away a stone " 🥌

PR #347, opened 2021-10-15
Maker actor system

luckysori:
How does this attempt at an actor system look? 

- The constructor just hooks everything up in the same way that `main` currently does. I think we want this to be consistent between tests and production. 
- The closures allow us to build whatever actors we want so long as they respect the contract expressed in the trait bounds. Maybe this is clunky. This could be leveraged to, for example, construct a `monitor::Actor` stand-in which immediately finds any transactions we subscribe to.
- I assume both the `wallet` and `db` arguments can be previously initialised for the purpose of testing, like we've done in the past.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -4,23 +4,28 @@ use bdk::bitcoin::secp256k1::schnorrsig;
 use clap::Clap;
 use daemon::auth::{self, MAKER_USERNAME};
 use daemon::db::{self, load_all_cfds};
-use daemon::model::cfd::{Order, UpdateCfdProposals};
+use daemon::maker_cfd::{FromTaker, NewTakerOnline};
+use daemon::model::cfd::{Cfd, Order, UpdateCfdProposals};
 use daemon::model::WalletInfo;
+use daemon::oracle::Attestation;
 use daemon::seed::Seed;
 use daemon::wallet::Wallet;
 use daemon::{
     bitmex_price_feed, fan_out, housekeeping, logger, maker_cfd, maker_inc_connections, monitor,
     oracle, wallet_sync,
 };
+use futures::Future;
 use rocket::fairing::AdHoc;
 use rocket_db_pools::Database;
+use sqlx::SqlitePool;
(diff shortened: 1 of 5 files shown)


daemon/src/maker.rs
```diff
@@ -318,3 +256,120 @@ async fn main() -> Result<()> {
...
+        F: Future<Output = Result<M>>,
+    {
+        let mut conn = db.acquire().await.unwrap();
+
+        housekeeping::transition_non_continue_cfds_to_setup_failed(&mut conn)
+            .await
+            .unwrap();
+        housekeeping::rebroadcast_transactions(&mut conn, &wallet)
+            .await
+            .unwrap();
```
thomaseizinger: This should be outside of the actor system as discussed.

thomaseizinger (approved):
Nice! Let's see how easy it will be to write tests against this :)

thomaseizinger:
Why is CI not running on this one?

daemon/src/maker.rs
```diff
@@ -4,23 +4,28 @@ use bdk::bitcoin::secp256k1::schnorrsig;
...
+use sqlx::SqlitePool;
 use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
 use std::task::Poll;
 use std::time::Duration;
-use tokio::sync::watch;
+use tokio::net::TcpListener;
+use tokio::sync::watch::{self, Receiver};
```
klochowicz, line 28: IMHO it would be nicer if we didn't import `Receiver`, but rather qualify it, otherwise it's pretty hard to notice what kind of channel are we talking about.

klochowicz (approved):
LGTM!

daemon/src/maker.rs
luckysori: Yep, I blame auto-import.

luckysori:
> Why is CI not running on this one?

It looks like it tried and failed before even starting: https://github.com/comit-network/hermes/actions/runs/1352840503/attempts/1.

PR #351, opened 2021-10-17
Sync with Olivia every 5 seconds

### LINKED ISSUE #349: Problems with getting oracle announcements
Keep seeing errors like this one in the logs (Maker)

```
2021-10-16 09:40:51 DEBUG Maker accepts a roll_over proposal order_id=4d28213a-63fe-40bc-a24d-321f14bb0be7
2021-10-16 09:40:51 ERROR Message handler failed: Announcement /x/BitMEX/BXBT/2021-10-16T23:00:00.price?n=20 not found
```

The announcement exists: https://h00.ooo/x/BitMEX/BXBT/2021-10-16T23:00:00.price?n=20
I feel this is a problem of the oracle actor not being in sync correctly. 

This does not only happen after startup, but also after letting the application run for a while. 
I saw this on various occasions when we request an announcement from the oracle actor, notably contract setup and rollover.

thomaseizinger:
We are only making requests to Olivia that are absolutely necessary,
i.e. only fetch attestations when they are likely ready and only
fetch attestations that we definitely need.

As a result, we can trigger the sync much more frequent.

Fixes #349.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -257,7 +257,7 @@ async fn main() -> Result<()> {
 
                 tokio::spawn(
                     oracle_actor_context
-                        .notify_interval(Duration::from_secs(60), || oracle::Sync)
+                        .notify_interval(Duration::from_secs(5), || oracle::Sync)
                         .unwrap(),
                 );
                 let actor = fan_out::Actor::new(&[&cfd_maker_actor_inbox, &monitor_actor_address])
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -224,13 +224,20 @@ impl xtra::Handler<GetAnnouncement> for Actor {
         msg: GetAnnouncement,
         _ctx: &mut xtra::Context<Self>,
     ) -> Option<Announcement> {
-        self.announcements
-            .get_key_value(&msg.0)
-            .map(|(id, (time, nonce_pks))| Announcement {
-                id: *id,
-                expected_outcome_time: *time,
-                nonce_pks: nonce_pks.clone(),
-            })
+        let announcement =
+            self.announcements
+                .get_key_value(&msg.0)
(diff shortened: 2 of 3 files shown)


da-kami:
Happy to get this in, but I am not sure it actually solves #349 - the remaining problem might just be related to rollover though.

thomaseizinger:
Checkout 6d03326 @da-kami.

Dirty AF but should work.

daemon/src/maker.rs
```diff
@@ -257,7 +257,7 @@ async fn main() -> Result<()> {
 
                 tokio::spawn(
                     oracle_actor_context
-                        .notify_interval(Duration::from_secs(60), || oracle::Sync)
+                        .notify_interval(Duration::from_secs(5), || oracle::Sync)
```
da-kami, line 260: So, the reasoning behind the commit message is, that pending attestations and pending announcements will only contain data once "necessary"? 

Is that actually true for the attestations? I thought we  have the pending attestations around once the contract was opened? 

I feel the logic should be changed to only fetch from the oracle if we don't have something. 

At the moment we always trigger both `update_pending_announcements` and `update_pending_attestations`:

```rust 
#[async_trait]
impl xtra::Handler<Sync> for Actor {
    async fn handle(&mut self, _: Sync, ctx: &mut xtra::Context<Self>) {
        self.update_pending_announcements(ctx);
        self.update_pending_attestations(ctx);
    }
}
```

but as far as I could see there is no guard that checks if we don't already have the data in `update_pending_announcements` and `update_pending_attestations`. Am I misunderstanding the logic?

thomaseizinger: Checkout this piece here:

https://github.com/comit-network/hermes/blob/dd5fb29b7845744158456d4c6caf9393ce6e710b/daemon/src/oracle.rs#L147-L155

We check locally if the attestation has likely already occurred so most of the time, this is not going to make any requests.

PR #353, opened 2021-10-18
Make `taker_cfd::Actor` generic over other actor addresses

luckysori:
Equivalent to what we did to `maker_cfd::Actor` in #339.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -67,7 +67,7 @@ enum RollOverState {
     None,
 }
 
-pub struct Actor {
+pub struct Actor<O, M> {
     db: sqlx::SqlitePool,
     wallet: Wallet,
     oracle_pk: schnorrsig::PublicKey,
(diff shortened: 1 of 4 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -67,22 +67,22 @@ enum RollOverState {
...
+pub struct Actor<O, M> {
     db: sqlx::SqlitePool,
     wallet: Wallet,
     oracle_pk: schnorrsig::PublicKey,
     cfd_feed_actor_inbox: watch::Sender<Vec<Cfd>>,
     order_feed_actor_inbox: watch::Sender<Option<Order>>,
     update_cfd_feed_sender: watch::Sender<UpdateCfdProposals>,
-    send_to_maker: Address<send_to_socket::Actor<wire::TakerToMaker>>,
-    monitor_actor: Address<monitor::Actor>,
+    send_to_maker: Box<dyn MessageChannel<wire::TakerToMaker> + Send>,
```
thomaseizinger: I thought `MessageChannel` already enforces `Send`?

daemon/src/taker_cfd.rs
```diff
@@ -249,7 +374,62 @@ impl Actor {
         }
         Ok(())
     }
+}
+
+impl<O, M> Actor<O, M>
+where
+    O: xtra::Handler<oracle::MonitorAttestation>,
+    M: xtra::Handler<monitor::StartMonitoring>,
```
thomaseizinger, line 351: I think that is the only message we are sending. Should we use a `MessageChannel` here?

thomaseizinger (approved):

daemon/src/taker_cfd.rs
luckysori: Good catch. This was a failed attempt at solving the `MessageSink` problem and I forgot to revert it.

daemon/src/taker_cfd.rs
luckysori: Another good catch. Yes, let's replace `Address<M>` with a message channel for `monitor::StartMonitoring`.

luckysori: Actually, I think this is no longer true with https://github.com/comit-network/hermes/pull/355/files#diff-bc49b56cb195451c001f467d581710fa452878ecb0f42787703e4cf1dda3259eR357-R363.

PR #355, opened 2021-10-18
Notify monitor actor to watch for collaborative close tx

da-kami:
So far we only did this on startup, but not during the actual execution of the collaborative close.


klochowicz (approved):
thanks!

daemon/src/monitor.rs
```diff
@@ -315,6 +320,17 @@ where
...
 
+    async fn handle_collaborative_settlement(
+        &mut self,
+        collaborative_settlement: CollaborativeSettlement,
+    ) -> Result<()> {
+        self.monitor_close_finality(
+            collaborative_settlement.tx,
+            collaborative_settlement.order_id,
+        );
+        Ok(())
```
thomaseizinger: This doesn't fail, don't need to return a `Result`.

thomaseizinger (approved):

luckysori (approved):

PR #362, opened 2021-10-18
Prefetch announcement upon proposing rollover

### LINKED ISSUE #361: Accepted rollover proposal fails because announcement is not yet fetched
As per title.

Fix is likely to prefetch announcement already when sending rollover proposal and hope (:crossed_fingers:) that it is there by the time the maker accepts.

thomaseizinger:
Fixes #361.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -17,6 +17,7 @@ use futures::channel::mpsc;
 use futures::{future, SinkExt};
 use std::collections::HashMap;
 use std::time::SystemTime;
+use time::OffsetDateTime;
 use tokio::sync::watch;
 use xtra::prelude::*;
 use xtra::KeepRunning;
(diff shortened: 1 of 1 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -343,6 +317,40 @@ where
...
+            proposal.order_id,
+            UpdateCfdProposal::RollOverProposal {
+                proposal: proposal.clone(),
+                direction: SettlementKind::Outgoing,
+            },
+        );
+        self.send_pending_update_proposals()?;
+
+        // we are likely going to need this one
+        self.oracle_actor
```
da-kami, line 341: Shall we do this one earlier, so we don't insert pending things in case it fails? (it is `send`, so I suppose we know if it failed)

thomaseizinger: I can only fail if the actor is down which it really should never be.

da-kami: I've heard that before :D

da-kami (approved):

PR #366, opened 2021-10-18
Introduce taker ActorSystem

klochowicz:
The connection actor is a real actor yet, however it should be enough to
abstract away from the production I/O.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -1,0 +1,44 @@
+use daemon::{send_to_socket, taker_cfd, wire};
+use futures::{Stream, StreamExt};
+use std::net::SocketAddr;
+use std::time::Duration;
+use tokio_util::codec::FramedRead;
+use xtra::prelude::MessageChannel;
+use xtra::spawn::TokioGlobalSpawnExt;
+use xtra::Actor as _;
+
+const CONNECTION_RETRY_INTERVAL: Duration = Duration::from_secs(5);
+
+pub struct Actor {
+    pub send_to_maker: Box<dyn MessageChannel<wire::TakerToMaker>>,
+    pub read_from_maker: Box<dyn Stream<Item = taker_cfd::MakerStreamMessage> + Unpin + Send>,
+}
+
+impl Actor {
+    pub async fn new(maker: SocketAddr) -> Self {
+        let (read, write) = loop {
+            let socket = tokio::net::TcpSocket::new_v4().expect("Be able ta create a socket");
+            if let Ok(connection) = socket.connect(maker).await {
+                break connection.into_split();
+            } else {
+                tracing::warn!(
+                    "Could not connect to the maker, retrying in {}s ...",
+                    CONNECTION_RETRY_INTERVAL.as_secs()
+                );
+                tokio::time::sleep(CONNECTION_RETRY_INTERVAL).await;
+            }
+        };
+
+        let send_to_maker = send_to_socket::Actor::new(write)
+            .create(None)
+            .spawn_global();
+
+        let read = FramedRead::new(read, wire::JsonCodec::default())
+            .map(move |item| taker_cfd::MakerStreamMessage { item });
+
+        Self {
+            send_to_maker: Box::new(send_to_maker),
+            read_from_maker: Box::new(read),
+        }
+    }
+}
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -25,7 +25,7 @@ use std::str::FromStr;
 use std::task::Poll;
 use std::time::Duration;
 use tokio::net::TcpListener;
-use tokio::sync::watch::{self, Receiver};
+use tokio::sync::watch;
 use tracing_subscriber::filter::LevelFilter;
 use xtra::prelude::*;
 use xtra::spawn::TokioGlobalSpawnExt;
@@ -270,9 +270,9 @@ async fn main() -> Result<()> {
 
 pub struct ActorSystem<O, M, T> {
     cfd_actor_addr: Address<maker_cfd::Actor<O, M, T>>,
-    cfd_feed_receiver: Receiver<Vec<Cfd>>,
-    order_feed_receiver: Receiver<Option<Order>>,
-    update_cfd_feed_receiver: Receiver<UpdateCfdProposals>,
+    cfd_feed_receiver: watch::Receiver<Vec<Cfd>>,
+    order_feed_receiver: watch::Receiver<Option<Order>>,
+    update_cfd_feed_receiver: watch::Receiver<UpdateCfdProposals>,
 }
 
 impl<O, M, T> ActorSystem<O, M, T>
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -283,7 +283,8 @@ where
         + xtra::Handler<oracle::Sync>,
     M: xtra::Handler<monitor::StartMonitoring>
         + xtra::Handler<monitor::Sync>
-        + xtra::Handler<monitor::CollaborativeSettlement>,
+        + xtra::Handler<monitor::CollaborativeSettlement>
+        + xtra::Handler<oracle::Attestation>,
     T: xtra::Handler<maker_inc_connections::TakerMessage>
         + xtra::Handler<maker_inc_connections::BroadcastOrder>
         + xtra::Handler<maker_inc_connections::ListenerMessage>,
@@ -324,7 +325,7 @@ where
             order_feed_sender,
             update_cfd_feed_sender,
             inc_conn_addr.clone(),
-            monitor_addr,
+            monitor_addr.clone(),
             oracle_addr.clone(),
         )
         .create(None)
@@ -353,7 +354,7 @@ where
                 .notify_interval(Duration::from_secs(5), || oracle::Sync)
                 .unwrap(),
         );
-        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &cfd_actor_addr])
+        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &monitor_addr])
             .create(None)
             .spawn_global();
 
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -205,30 +205,25 @@ async fn main() -> Result<()> {
                     .unwrap();
 
                 let ActorSystem {
-                        cfd_actor_addr,
-                        cfd_feed_receiver,
-                        order_feed_receiver,
-                        update_cfd_feed_receiver,
-                    } =
-                        ActorSystem::new(
-                            db,
-                            wallet.clone(),
-                            oracle,
-                            |cfds, channel| oracle::Actor::new(cfds, channel),
-                            {
-                                |channel, cfds| {
-                                    let electrum = opts.network.electrum().to_string();
-                                    async move {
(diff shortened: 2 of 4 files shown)


klochowicz:
note: the best way forward for a connection actor seems to be allowing multiple makers (or at least operation without the maker, which makes sense - there's no reason for not allowing to monitor our ongoing cfd's after maker goes offline).

daemon/src/taker_cfd.rs
```diff
@@ -91,7 +91,7 @@ impl<O, M> Actor<O, M> {
         cfd_feed_actor_inbox: watch::Sender<Vec<Cfd>>,
         order_feed_actor_inbox: watch::Sender<Option<Order>>,
         update_cfd_feed_sender: watch::Sender<UpdateCfdProposals>,
-        send_to_maker: Box<dyn MessageChannel<wire::TakerToMaker> + Send>,
+        send_to_maker: Box<dyn MessageChannel<wire::TakerToMaker>>,
```
thomaseizinger, line 95: I am happy to merge minor fixes like this so if you want to do this, please make a separate commit / PR :)

thomaseizinger (changes requested):
Minor fixes like imports and stuff can still go in. I would like to hold off with refactorings like the `ActorSystem` until we did the release :)

Meaning we can either split out the minor changes as commits or we just pause the entire one until we did the release.

daemon/src/taker_cfd.rs
luckysori: This is already in master. This branched wasn't rebased yet.

klochowicz: @luckysori for some reason, this one didn't land in the master - I think only maker fixes did. I was fairly sure I rebased against master before posting the PR, and doublechecked now. 
However, at this stage I might as well hold off till after the release 😉

klochowicz:
> Minor fixes like imports and stuff can still go in. I would like to hold off with refactorings like the `ActorSystem` until we did the release :)
> 
> Meaning we can either split out the minor changes as commits or we just pause the entire one until we did the release.

Happy to pause it, as the release is not far off :)

daemon/src/taker_cfd.rs
luckysori: I'm silly, I removed the `+ Send` from the field, but not from the argument :sweat_smile:

klochowicz:
> Minor fixes like imports and stuff can still go in. I would like to hold off with refactorings like the `ActorSystem` until we did the release :)
> 
> Meaning we can either split out the minor changes as commits or we just pause the entire one until we did the release.

@thomaseizinger made a release earlier today - we can think about merging this one now 💭

daemon/src/taker.rs
```diff
@@ -200,72 +179,43 @@ async fn main() -> Result<()> {
...
-                    oracle_actor_address.clone(),
+                    send_to_maker,
+                    read_from_maker,
+                    |cfds, channel| oracle::Actor::new(cfds, channel),
+                    {
+                        |channel, cfds| {
+                            let electrum = opts.network.electrum().to_string();
+                            async move {
+                                monitor::Actor::new(electrum, channel, cfds.clone()).await
+                            }
```
thomaseizinger: `async move` for a single `.await` expression should be equivalent to just passing the initial future in!

```suggestion
monitor::Actor::new(electrum, channel, cfds.clone())
```

daemon/src/connection.rs
```diff
@@ -0,0 +1,45 @@
...
+        let (read, write) = loop {
+            let socket = tokio::net::TcpSocket::new_v4().expect("Be able ta create a socket");
+            if let Ok(connection) = socket.connect(maker).await {
+                break connection.into_split();
+            } else {
+                tracing::warn!(
+                    "Could not connect to the maker, retrying in {}s ...",
+                    CONNECTION_RETRY_INTERVAL.as_secs()
+                );
+                sleep(CONNECTION_RETRY_INTERVAL);
```
thomaseizinger: This will block the thread, better use `tokio`'s `async` sleep!

daemon/src/connection.rs
```diff
@@ -0,0 +1,45 @@
...
+use std::thread::sleep;
+use std::time::Duration;
+use tokio_util::codec::FramedRead;
+use xtra::prelude::MessageChannel;
+use xtra::spawn::TokioGlobalSpawnExt;
+use xtra::Actor as _;
+
+const CONNECTION_RETRY_INTERVAL: Duration = Duration::from_secs(5);
+
+pub struct Actor {
```
thomaseizinger, line 12: Any chance we can have this split out in its own commit easily?

If it is too hard, don't bother.

daemon/src/taker.rs
```diff
@@ -200,72 +179,43 @@ async fn main() -> Result<()> {
                 housekeeping::rebroadcast_transactions(&mut conn, &wallet)
                     .await
                     .unwrap();
```
thomaseizinger, line 181: I think with https://github.com/comit-network/hermes/pull/381, there is actually no more reason why we cannot boot up the actor system outside of rocket.

Doing that would allow us for easier error handling because we can just use `?` instead of `unwrap`ing in here.

Best to be done in a follow-up though!

daemon/src/taker.rs
```diff
@@ -292,3 +242,93 @@ async fn main() -> Result<()> {
...
+                    .unwrap(),
+            ),
+        );
+
+        tokio::spawn(
+            oracle_ctx
+                .notify_interval(Duration::from_secs(5), || oracle::Sync)
+                .unwrap(),
+        );
+        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &cfd_actor_addr])
```
thomaseizinger: :stop_sign: Why are we passing the same actor twice here? That can't be right!

daemon/src/taker.rs
```diff
@@ -292,3 +242,93 @@ async fn main() -> Result<()> {
...
+        monitor_constructor: impl Fn(Box<dyn StrongMessageChannel<monitor::Event>>, Vec<Cfd>) -> F,
+    ) -> Self
+    where
+        F: Future<Output = Result<M>>,
+    {
+        let mut conn = db.acquire().await.unwrap();
+
+        let cfds = load_all_cfds(&mut conn).await.unwrap();
+
+        let (cfd_feed_sender, cfd_feed_receiver) = watch::channel(cfds.clone());
```
thomaseizinger, line 278: Is it worth it to do this inside here if we are passing the `cfd_feed_receiver` out again? We could just do this outside and only pass the `sender` in?

daemon/src/taker.rs
```diff
@@ -292,3 +242,93 @@ async fn main() -> Result<()> {
...
+        let (order_feed_sender, order_feed_receiver) = watch::channel::<Option<Order>>(None);
```
thomaseizinger, line 279: I think the same applies here. I'd rather pass the `order_feed_sender` in than passing the `order_feed_receiver` out.

daemon/src/taker.rs
```diff
@@ -292,3 +242,93 @@ async fn main() -> Result<()> {
...
+        let (update_cfd_feed_sender, update_cfd_feed_receiver) =
```
thomaseizinger, line 280: If we do the same thing with all of these, we can reduce the return values to just the `cfd_actor::Address`!

daemon/src/taker.rs
```diff
@@ -292,3 +242,93 @@ async fn main() -> Result<()> {
...
+        + xtra::Handler<monitor::CollaborativeSettlement>,
+{
+    pub async fn new<F>(
+        db: SqlitePool,
+        wallet: Wallet,
+        oracle_pk: schnorrsig::PublicKey,
+        send_to_maker: Box<dyn MessageChannel<wire::TakerToMaker>>,
+        read_from_maker: Box<dyn Stream<Item = taker_cfd::MakerStreamMessage> + Unpin + Send>,
+        oracle_constructor: impl Fn(Vec<Cfd>, Box<dyn StrongMessageChannel<Attestation>>) -> O,
+        monitor_constructor: impl Fn(Box<dyn StrongMessageChannel<monitor::Event>>, Vec<Cfd>) -> F,
```
thomaseizinger, line 269: For the tests, we might need control over the oracle and monitor actors so you will likely need to return their addresses (`O` and `F`) from this function!

thomaseizinger (changes requested):
Found a bug and noted some potential improvements :)

Thanks!

daemon/src/taker.rs
klochowicz: it was a bit weird for me too, but I copied it from already existing pattern:
https://github.com/comit-network/hermes/blob/taker-actor-system/daemon/src/maker.rs#L356

(shall I fix it both places? :) )

daemon/src/taker.rs
klochowicz: > Found a bug and noted some potential improvements :)
> 
> Thanks!

This ActorSystem closely matches the one on the maker side - perhaps it would be best for me to create a follow-up that rolls out the suggestions that apply to both sides? (I'll address the ones that solely matter for the taker here)

daemon/src/taker.rs
thomaseizinger: Yes!

daemon/src/taker.rs
klochowicz: well, it saves some of the boilerplate.... and I'm just repeating the pattern from the maker. 
I don't have any strong feelings for or against, but I see it as a bit more convenient.

daemon/src/taker.rs
klochowicz: done! I'll attach the maker commit to this PR as well.

daemon/src/taker.rs
klochowicz: As stated above, happy to do it, but it will result in more boilerplate in the tests, so we'll need to add one more wrapper that adds these watch channels (perhaps that's better, as it splits the responsibilities?)

daemon/src/taker.rs
klochowicz: happy to do it as a follow-up :)

daemon/src/connection.rs
klochowicz: not too easily, I'll keep it mind to be more atomic with my commits!

klochowicz:
oops, I accidentally autosquashed, after `git absorb` resulted in 4 fixup commits :( sorry for making it more difficult to look at it again!

thomaseizinger (approved):

daemon/src/maker.rs
```diff
@@ -353,7 +354,7 @@ where
                 .notify_interval(Duration::from_secs(5), || oracle::Sync)
                 .unwrap(),
         );
-        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &cfd_actor_addr])
+        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &monitor_addr])
```
thomaseizinger, line 352: Pretty sure this bug has consequences in monitoring for CETs in the current release as the monitoring actor will not learn about attestations and hence not monitor for CETs.

cc @da-kami @bonomat

bonomat: good catch. I think @da-kami noticed this bug already.

thomaseizinger: We should push out a 0.1.1 that is branched off the 0.1.0 tag.

PR #372, opened 2021-10-18
Print git tag when starting maker/taker

bonomat:
Prints this on latest master: 

```bash
2021-10-19 09:08:22  INFO Running version: preview-133-ga998367
```


luckysori (approved):
Nice improvement!

PR #381, opened 2021-10-19
Properly shutdown DB connection before exit

thomaseizinger:


bonomat (approved):
tested locally, tmp files are gone and I was able to reopen the db file :).

PR #383, opened 2021-10-19
Make `OrderId` always represent as `Hyphenated` in DB

thomaseizinger:


klochowicz (approved):

DeliciousHair (approved):
Approved as well, but TBH I prefer leaving the data as binary. This is only a personal preference though and I don't know of any actual difference it makes in terms of performance or extra code.

thomaseizinger:
> Approved as well, but TBH I prefer leaving the data as binary. This is only a personal preference though and I don't know of any actual difference it makes in terms of performance or extra code.

I would like to upstream an implementation of `Serialize` and `Deserialize` to `uuid` for `Hyphenated`. Pretty annoying that we can no longer derive that.

Otherwise I am preferring the data within the tables to be human-readable. Performance implications would have to be measured first before we can make a call here.

bors r+

da-kami:
Come on bors...

PR #384, opened 2021-10-19
Refer to correct ID when joining orders

thomaseizinger:


thomaseizinger:
@da-kami I've trimmed down the test a bit, let me know if it still conveys the right idea from your perspective.

DeliciousHair:
Test looks good to me!

thomaseizinger:
I verified that the test fails without the fix.

da-kami:
> @da-kami I've trimmed down the test a bit, let me know if it still conveys the right idea from your perspective.

Like the trim down, should convey the same 👍

da-kami (approved):

Issue #365, opened 2021-10-18
Make a dedicated `ExchangeRate` type


DeliciousHair:
In the same vein as #357, the `Price` type should have it's own type and (IMHO) be renamed to something more expressive such as `ExchangeRate`. The reason for this falls out of attempting to do basic algebra with the current `Price` type:

```rust
pub struct Order {
    ...
    pub price: Usd,
```

If you want to simply compute a value such as `amount_btc = amount_usd / price` this gets strange quickly, as `amount_usd` is (correctly) represented as `Usd` while `Price` is magically known to actually be `Usd/Btc` and an implementation of `Div` needs to reflect this in order to get the units correct. But then new problems arise if we want to compute, say, half the current exchange rate as division now gives us a value in BTC...

Proposal is to thus create a new type `ExchangeRate` that properly encapsulates that `Price` is not, in fact `Usd`.

DeliciousHair:
Extending this a bit to cover more type definitions, taking a bite out of #357 at the same time.

DeliciousHair:
For some unimportant reasons, went with `Price` instead of `ExchangeRate` and resolved this via #405.

PR #390, opened 2021-10-20
Fix precision of Usd and Percent to 2 decimal upon serialization

thomaseizinger:


daemon/src/model.rs
```diff
@@ -27,6 +32,7 @@ impl Usd {
         Ok(Usd(result))
     }
 
+    // TODO: Usd * Usd = Usd^2 not Usd !!!
```
bonomat, line 35: What do you mean with this todo?

bonomat (approved):
LGTM 👍

daemon/src/model.rs
thomaseizinger: The function signature claims that multiplying `Usd` with `Usd` gives you `Usd` but that is mathematically not correct. The correct result would be `Usd`^2.

You cannot just ignore units when doing arithmetic.

The argument of this function should likely be changed from `Usd` to a scalar like `Decimal` or `u32` or something.

thomaseizinger: @DeliciousHair is going to fix this with #365.

da-kami (approved):

PR #395, opened 2021-10-20
Bail only if transaction does not make it on-chain

luckysori:
We no longer wrongly report an error when failing to broadcast a transaction because it's already on-chain.

Additionally, I tried to simplify this function with https://github.com/comit-network/hermes/pull/395/commits/3e8e6bc06eb99e504504c0a14c5f01ed1422a48d, which seems a lot better to me, but I ~could have~ **missed something**.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -96,17 +96,9 @@ impl Wallet {
 
     pub async fn try_broadcast_transaction(&self, tx: Transaction) -> Result<Txid> {
         let wallet = self.wallet.lock().await;
-        // TODO: Optimize this match to be a map_err / more readable in general
         let txid = tx.txid();
 
         let result = wallet.broadcast(tx.clone());
-
-        if result.is_err() {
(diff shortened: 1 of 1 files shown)


daemon/src/wallet.rs
```diff
@@ -96,31 +96,33 @@ impl Wallet {
...
-                return Ok(txid);
+        let result = match result {
+            Ok(txid) => Ok(txid),
+            Err(bdk::Error::Electrum(electrum_client::Error::Protocol(value))) => {
+                let error_code = parse_rpc_protocol_error_code(&value).with_context(|| {
+                    format!("Failed to parse electrum error response '{:?}'", value)
+                })?;
+
+                if error_code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain) {
+                    Ok(tx.txid())
```
klochowicz: perhaps add a log message stating that it's already on chain?

klochowicz (approved):
nice one 💯

PR #398, opened 2021-10-20
Don't fail for loops

da-kami:
We should not fail for loops where we loop over cfds to process state updates based on information we learned / restart. 
We use a macro to wrap all fallible code in the loop to just print an error and continue. 

Note: I went over the codebase in search of other for loops, maybe someone else can do another grep :)

### DIFF
--- a/daemon/src/housekeeping.rs
+++ b/daemon/src/housekeeping.rs
@@ -1,10 +1,10 @@
 use crate::db::{append_cfd_state, load_all_cfds};
 use crate::model::cfd::{Cfd, CfdState};
+use crate::try_continue;
 use crate::wallet::Wallet;
 use anyhow::Result;
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
-
 pub async fn transition_non_continue_cfds_to_setup_failed(
     conn: &mut PoolConnection<Sqlite>,
 ) -> Result<()> {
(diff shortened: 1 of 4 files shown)


daemon/src/housekeeping.rs
```diff
@@ -29,29 +29,28 @@ pub async fn rebroadcast_transactions(
...
-        let txid = wallet.try_broadcast_transaction(dlc.lock.0.clone()).await?;
-
+        let txid = try_continue!(wallet.try_broadcast_transaction(dlc.lock.0.clone()).await);
         tracing::info!("Lock transaction published with txid {}", txid);
     }
 
     for cfd in cfds.iter().filter(|cfd| Cfd::is_must_refund(cfd)) {
-        let signed_refund_tx = cfd.refund_tx()?;
-        let txid = wallet.try_broadcast_transaction(signed_refund_tx).await?;
+        let signed_refund_tx = try_continue!(cfd.refund_tx());
```
thomaseizinger: These we should already handle via `filter_map`.

thomaseizinger (approved):
Nice, thanks!

daemon/src/housekeeping.rs
da-kami: Yes, good point - even though it is technically not 100% ensured that we cannot go there because those are independent code paths, it is fair to assume that this can really *never* fail :)
I amended the commit

thomaseizinger: Ha, so what I meant was, we should have an accessor om `Cfd` for `refund_tx` that returns an `Option`.

da-kami: ah, yeah, we should change those to be options like for `cfd()` and `attestation()` - I'll quickly do it

da-kami: hm... for `commit_tx` and `refund_tx` it is trivial, but for `cet()` a bit tricky because of the double result... 
I think it is better if we do these changes after the freeze. Really wanna avoid side effects with changes atm :)

PR #400, opened 2021-10-20
Cleanup `unwrap`s throughout the codebase

thomaseizinger:
- Make function that never fails non-fallible
- Disallow `.unwrap` in prod code and use expect for remaining ones
- Rewrite matrix_solve to not contain `unwrap`

### DIFF
--- a/daemon/src/payout_curve/csr_tools.rs
+++ b/daemon/src/payout_curve/csr_tools.rs
@@ -1,5 +1,6 @@
 use crate::payout_curve::compat::{To1DArray, ToNAlgebraMatrix};
 use crate::payout_curve::Error;
+use itertools::Itertools;
 use ndarray::prelude::*;
 use std::ops::Mul;
 
@@ -49,29 +50,19 @@ impl CSR {
     // is horrible.
     pub fn matrix_solve(&self, b_arr: &Array2<f64>) -> Result<Array2<f64>, Error> {
         let a_arr = self.todense();
-        let ncols = b_arr.shape()[1];
-        let mut temp = (0..ncols)
-            .rev()
-            .map(|e| {
-                let b = b_arr.slice(s![.., e]).to_owned();
+        let zeros = Array2::zeros((b_arr.nrows(), 0));
 
-                let sol = lu_solve(&a_arr, &b)?;
-                Ok(sol.to_vec())
-            })
-            .collect::<Result<Vec<_>, Error>>()?;
+        let result = b_arr
+            .columns()
+            .into_iter()
+            .map(|b| lu_solve(&a_arr, &b.to_owned()))
+            .fold_ok(zeros, |mut result, column| {
+                result
+                    .push_column(column.view())
+                    .expect("shape was initialized correctly");
 
-        let nrows = temp[0].len();
-        let mut raveled = Vec::with_capacity(nrows * temp.len());
-
-        for _ in 0..nrows {
-            for vec in &mut temp {
-                raveled.push(vec.pop().unwrap());
-            }
-        }
-
-        raveled.reverse();
-
-        let result = Array2::<f64>::from_shape_vec((nrows, ncols), raveled)?.to_owned();
+                result
+            })?;
 
         Ok(result)
     }
--- a/daemon/src/to_sse_event.rs
+++ b/daemon/src/to_sse_event.rs
@@ -262,8 +262,8 @@ impl ToSseEvent for CfdsWithAuxData {
 
                     // TODO: Depending on the state the margin might be set (i.e. in Open we save it
                     // in the DB internally) and does not have to be calculated
-                    margin: cfd.margin().unwrap(),
-                    margin_counterparty: cfd.counterparty_margin().unwrap(),
+                    margin: cfd.margin().expect("margin to be available"),
+                    margin_counterparty: cfd.counterparty_margin().expect("margin to be available"),
                     details,
                     expiry_timestamp: cfd.expiry_timestamp(),
                 }
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -74,7 +74,7 @@ impl Quote {
 
     pub fn for_taker(&self) -> Usd {
         // TODO: verify whether this is correct
-        self.mid_range().unwrap()
+        self.mid_range().expect("decimal arithmetic to not fail")
     }
 
     fn mid_range(&self) -> Result<Usd> {
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -1,3 +1,5 @@
+#![cfg_attr(not(test), warn(clippy::unwrap_used))]
+
 pub mod actors;
 pub mod auth;
 pub mod bitmex_price_feed;
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -92,7 +92,7 @@ impl Actor<bdk::electrum_client::Client> {
                     if let Some(model::cfd::CollaborativeSettlement { tx, ..}
                     ) = cfd.state.get_collaborative_close()  {
                         let close_params = (tx.txid(),
-                            tx.output.first().expect("have output").script_pubkey.clone());
+                            tx.output.first().context("transaction has zero outputs")?.script_pubkey.clone());
                         actor.monitor_close_finality(close_params,cfd.order.id);
                     }
                 }
--- a/daemon/src/payout_curve.rs
+++ b/daemon/src/payout_curve.rs
@@ -351,7 +351,7 @@ impl PayoutCurve {
             vec.push(arr[[i, 1]]);
         }
 
-        *arr = Array2::<f64>::from_shape_vec((capacity, 2), vec).unwrap();
+        *arr = Array2::<f64>::from_shape_vec((capacity, 2), vec).expect("vec is a 2D array");
     }
 
     fn modify_samples_unbounded(&self, arr: &mut Array2<f64>) {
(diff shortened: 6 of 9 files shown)


daemon/src/payout_curve/csr_tools.rs
```diff
@@ -49,29 +50,19 @@ impl CSR {
...
-        for _ in 0..nrows {
-            for vec in &mut temp {
-                raveled.push(vec.pop().unwrap());
-            }
-        }
-
-        raveled.reverse();
-
-        let result = Array2::<f64>::from_shape_vec((nrows, ncols), raveled)?.to_owned();
+        let zeros = Array2::zeros((b_arr.nrows(), 0));
```
da-kami, line 53: Can you maybe walk me through the change tomorrow? I have a hard time validating that the changed code does the same as the original. Are there tests that ensure the same behavior?

daemon/src/to_sse_event.rs
```diff
@@ -258,8 +258,8 @@ impl ToSseEvent for CfdsWithAuxData {
 
                     // TODO: Depending on the state the margin might be set (i.e. in Open we save it
                     // in the DB internally) and does not have to be calculated
-                    margin: cfd.margin().unwrap(),
-                    margin_counterparty: cfd.counterparty_margin().unwrap(),
+                    margin: cfd.margin().expect("margin to be available"),
```
da-kami, line 265: nit: The margins are actually calculated, I think a `context(...)?` is more appropriate here. 

Note: We please also remove the TODO above, I don't think we will ever action this.

daemon/src/payout_curve/csr_tools.rs
DeliciousHair: @da-kami This is simply a convenience method to solve a matrix equation of the form `AX=B`.

There are no solvers included with `ndarray` (without using BLAS or similar, and thus requiring a fortran compiler) so the pure rust LU solver from `nalgebra` was ninja-ed in that allows us to solve `Ax=b` and this simply iterates over the columns of `X` to determine the columns of `B`.

thomaseizinger: > Can you maybe walk me through the change tomorrow?

Not really 😅
I had to `dbg` things myself here as I couldn't visualize it otherwise. But I noticed that we had good test coverage of this code so I just trial-and-errorred until I had something that I think is shorter and easier to read but achieves the same thing.

Ideally, I would have liked a function on `ndarray` to build a matrix from columns. Seems trivial and I don't understand why it doesn't exist ...

There is a lot of unnecessary copying going on here.

daemon/src/to_sse_event.rs
thomaseizinger: The problem is that we are in loop here where we can't return a result.

Needs some more thinking in how we want to handle errors during projection for the UI. 

Should we not display the CFD at all? Just not show margin? In what cases can we actually fail to calculate the margin?

So many questions ...

daemon/src/payout_curve/csr_tools.rs
da-kami: > @da-kami This is simply a convenience method to solve a matrix equation of the form `AX=B`.
> 
> There are no solvers included with `ndarray` (without using BLAS or similar, and thus requiring a fortran compiler) so the pure rust LU solver from `nalgebra` was ninja-ed in that allows us to solve `Ax=b` and this simply iterates over the columns of `X` to determine the columns of `B`.

@DeliciousHair could you verify that the code change does not have any unwanted side effects?

da-kami: I'll have a look at the test coverage. I'm afraid of uncovered edge cases that were not effected by the old code but are by the new. Don't wanna be annoying here, just saying it's hard for me to reason about the change :)

DeliciousHair: I thought this was well tested already, but it seems to not be the case--maybe I didn't commit the tests or something... Regardless, some minor glitches have been found and tests added to address them. Woot!

thomaseizinger:
@da-kami Going to rebase this onto master once #406 is merged for more confidence!

daemon/src/payout_curve/csr_tools.rs
DeliciousHair: > Ideally, I would have liked a function on `ndarray` to build a matrix from columns. Seems trivial and I don't understand why it doesn't exist ...

I thought [you could do that](https://docs.rs/ndarray/0.15.3/ndarray/struct.ArrayBase.html#method.axis_iter_mut) already? In this case, you'd need to build your output array `X` first as a mutable array of zeros with the correct shape, and then fill the columns in by iterating over `B` with the solver.

I think.

thomaseizinger: How do I use this to _build_ a matrix from an iterator of columns?

DeliciousHair: Something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2b2fb891dd535dce3f1f907bdfe752d5) perhaps?

DeliciousHair: not tested, but the pattern should be:
```rust
    pub fn matrix_solve(&self, b_arr: &Array2<f64>) -> Result<Array2<f64>, Error> {
        let a_arr = self.todense();
        let mut result = Array2::zeros((b_arr.nrows(), b_arr.ncols()));

        for (i, col) in b_arr.axis_iter(Axis(1)).enumerate() {
            let mut slice = result.slice_mut(s![.., i]);
            let update = lu_solve(&a_arr, &col.to_owned());
            slice.assign(&update);
        }

        result
    }
```

thomaseizinger:
@da-kami @DeliciousHair Rebased and ready to merge at your convenience!

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -49,29 +50,19 @@ impl CSR {
...
+            .columns()
+            .into_iter()
+            .map(|b| lu_solve(&a_arr, &b.to_owned()))
+            .fold_ok(zeros, |mut result, column| {
+                result
+                    .push_column(column.view())
+                    .expect("shape was initialized correctly");
+
+                result
+            })?;
```
DeliciousHair, line 65: Can I suggest this less-verbose version of the above (tested locally against master just now):

```rust
    pub fn matrix_solve(&self, b_arr: &Array2<f64>) -> Result<Array2<f64>, Error> {
        let a_arr = self.todense();
        let mut result = Array2::zeros((b_arr.nrows(), b_arr.ncols()));

        for (i, col) in b_arr.axis_iter(Axis(1)).enumerate() {
            let mut slice = result.slice_mut(s![.., i]);
            let update = lu_solve(&a_arr, &col.to_owned())?;
            slice.assign(&update);
        }

        Ok(result)
    }
```

DeliciousHair (changes requested):

daemon/src/payout_curve/csr_tools.rs
thomaseizinger: Hmm, not really a fan of the loop, I'd rather have the mutation contained in a `fold` so it is clearer what happens, ie we are building a new matrix from a series of computations.

DeliciousHair: Up to you I guess but I'd argue this is actually less clear, and also runs awry of the conventional pattern suggested.

thomaseizinger: How is it less clear?

I see the following "noise" that I need to think about before I can understand what is happening:

- How does `axis_iter` work?
- What is `Axis(1)`?
- Why do we need `enumerate`?

Having a running index variable to access a part of a data-structure is a code-smell IMO. We are not writing C code here, we have iterators for this :D 

Compared to that, the currently suggestion version makes it clear that we are iterating over all columns, computing something with each column and then building up a matrix again. Ideally, I would have liked to use `collect` for the last part but `ndarray` doesn't support that.

If anything, the 2nd mutable variable in your code can be avoided:

```rust
let update = lu_solve(&a_arr, &col.to_owned())?;
result.slice_mut(s![.., i]).assign(&update);
```

DeliciousHair: Interesting point of view that I had honestly not considered!

By my thinking, the reason I'm claiming my suggestion is "easier" is precisely because of the use of `axis_iter()` and `Axis()`, not as code noise. This suggests a philosophical difference: it is better to use the tools provided by `ndarray` to operate on an `ndarray` object, or is it better to use more bare-bones rust (for lack of a better phrase) in all cases unless absolutely necessary?

Personally, I think the former is generally the better approach (with the condensed step you suggested) as it keeps things clear what sort of "thing" we are working with, rather than defining a "thing" and then just treating it as a generic object. The difference is thus quite subjective and I don't think there is any objective difference* at all.

\* To be sure, I tried writing some benchmarks for the two approaches and the performance difference is consistently within the noise bound of the benchmarking experiments.

da-kami (approved):
I'm happy with merging this as is and accept changes as suggested by @DeliciousHair in a follow-up.

da-kami:
@DeliciousHair if the change request is important it would be great if you could take over the PR and add them.

DeliciousHair:
> @DeliciousHair if the change request is important it would be great if you could take over the PR and add them.

As I mentioned, it makes (essentially) zero performance difference, it's a style issue only--I think I'm right and have thus used the suggested pattern in a number of places, @thomaseizinger thinks he's right and has (presumably) used his pattern elsewhere as well.

In the end, doesn't seem to matter so do feel free to accept.

DeliciousHair (approved):

PR #401, opened 2021-10-20
Taker cleans up order feed

### LINKED ISSUE #399: Taker sets own order feed to `None` if the taker knows that the order is "gone"
This is more a resilience feature. 
Sometimes the taker looses the connection to the maker and is not informed about the order being removed / replaced anymore. In such a scenario it can happen that a taker tries to take the same order twice, which results in a `unique key constraint` violation when trying to create the `Cfd` on the taker side. (The maker is actually not involved in this scenario.)

The taker can know that the order was already taken and remove it form its own feed - in addition to not saving twice.

da-kami:
fixes https://github.com/comit-network/hermes/issues/399

The first commit was mostly added for ensuring sanity - this should also help us for scenarios where we restart the taker and the maker sends over the same current order as before.  Instead of bailing in `insert` we could handle this in a less intrusive way, because in some scenarios (as the one above) it is actually expected at the moment, so we might not want to treat it as error. For now I kept it simple and just bail (otherwise the insert would fail...) - that should not cause side effects.

### DIFF
--- a/daemon/src/cfd_actors.rs
+++ b/daemon/src/cfd_actors.rs
@@ -1,3 +1,4 @@
+use crate::db::load_cfd_by_order_id;
 use crate::model::cfd::{Attestation, Cfd, CfdState, CfdStateChangeEvent, OrderId};
 use crate::wallet::Wallet;
 use crate::{db, monitor, oracle};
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):
Thanks!

Issue #357, opened 2021-10-18
Epic: Replace current domain model with something sensible


thomaseizinger:
The current domain model has been created very ad-hoc and proves to be difficult to test and maintain.

Related issues are: https://github.com/comit-network/hermes/issues/290 and https://github.com/comit-network/hermes/issues/264.

We want a new domain model that:

- Is (same as the current one) pure: no dependency on IO, blockchain, wallet, async, etc
- Is rich and hard to misuse: testing a domain-model function should give confidence that the production code also works.
- Is easy to understand: f.e. the current `handle` function is way too big

It is likely worth to draw some ideas from:

- Domain-driven design (value types, entities & aggregates)
- Event-sourcing (domain events)
- CQRS (separating read model and write model, more explicit modelling of projections)

On an abstract level, interactions could look like this:

- Actor gets a message and loads current CFD from database
- Actor _sends a command_ (i.e. calls a function) on CFD to execute the action, f.e. `cfd.start_collaborative_close()`
- Model function emits a _domain-event_ that describes what happened
- Based on the _domain-event_, appropriate actions can be taken

Following this should allow us to get rid of a lot of boilerplate:

1. Explicit modelling of domain-events allows for a generic function that handles them (i.e. refreshing feed only happens in one place)
1. Explicit modelling of commands encapsulates all necessary logic
1. Explicit modelling of commands that produce domain-events allows for easy testing

Our current `CfdState` should likely be completely encapsulated within a `Cfd`, more discussions on that in https://github.com/comit-network/hermes/issues/290.

thomaseizinger:
Concrete tasks that can be made out of this:

- Ban the use of new types with public inner fields (following the "value types" idea of DDD), f.e. https://github.com/comit-network/hermes/issues/365
- Create entities and aggregates (effectively modelling commands and domain events)

thomaseizinger:
Outdated.

PR #405, opened 2021-10-20
Fix price type

DeliciousHair:
Replacement of structs with public data to ones with private data

Addresses #357 and #365. Although not a very large change, this PR ends up touching rather a lot of code.

* Converted types `Usd`, `Leverage` and `Percent` to something that is appropriate to this application
* Created new types `Price` and `InversePrice` to use for BTC/USD exchange rate with appropriate algebraic ops implemented as well.
* Added new positive tests
* The function `daemon::model::calculate_profit()` has been changed substantially as the updated types make the existing workflow needlessly complex
* Some tests (mostly in `cfd.rs` required updating) in order to make use of the new types.
* Minor edit to `.gitignore` to avoid accidental pushing of DB to repository--should have been it's own item, added here to fix a problem that arose during this work.

NOTE:
* There may be an excess of algebraic ops implemented, some pruning may be appropriate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@
 .pnp.*
 
 # Artifacts from running the daemons
-/*.sqlite*
+*.sqlite*
 /maker_seed
 /taker_seed
 /mainnet
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1,5 +1,5 @@
 use crate::model::cfd::{Cfd, CfdState, Order, OrderId};
-use crate::model::{BitMexPriceEventId, Usd};
+use crate::model::{BitMexPriceEventId, Price, Usd};
 use anyhow::{Context, Result};
 use rust_decimal::Decimal;
 use sqlx::pool::PoolConnection;
@@ -39,7 +39,7 @@ pub async fn insert_order(order: &Order, conn: &mut PoolConnection<Sqlite>) -> a
     .bind(&order.price.to_string())
     .bind(&order.min_quantity.to_string())
     .bind(&order.max_quantity.to_string())
-    .bind(order.leverage.0)
+    .bind(order.leverage.get())
     .bind(&order.liquidation_price.to_string())
     .bind(
         order
@@ -101,11 +101,11 @@ pub async fn load_order_by_id(
         id: row.uuid,
         trading_pair: row.trading_pair,
         position: row.position,
-        price: Usd::new(Decimal::from_str(&row.initial_price)?),
-        min_quantity: Usd::new(Decimal::from_str(&row.min_quantity)?),
-        max_quantity: Usd::new(Decimal::from_str(&row.max_quantity)?),
+        price: row.initial_price.parse::<Price>()?,
+        min_quantity: row.min_quantity.parse::<Usd>()?,
+        max_quantity: row.max_quantity.parse::<Usd>()?,
         leverage: row.leverage,
-        liquidation_price: Usd::new(Decimal::from_str(&row.liquidation_price)?),
+        liquidation_price: row.liquidation_price.parse::<Price>()?,
         creation_timestamp: convert_to_system_time(row.ts_secs, row.ts_nanos)?,
         term: Duration::new(row.term_secs, row.term_nanos),
         origin: row.origin,
@@ -315,11 +315,11 @@ pub async fn load_cfd_by_order_id(
         id: row.uuid,
         trading_pair: row.trading_pair,
         position: row.position,
-        price: Usd::new(Decimal::from_str(&row.initial_price)?),
-        min_quantity: Usd::new(Decimal::from_str(&row.min_quantity)?),
-        max_quantity: Usd::new(Decimal::from_str(&row.max_quantity)?),
+        price: row.initial_price.parse::<Price>()?,
+        min_quantity: row.min_quantity.parse::<Usd>()?,
+        max_quantity: row.max_quantity.parse::<Usd>()?,
         leverage: row.leverage,
-        liquidation_price: Usd::new(Decimal::from_str(&row.liquidation_price)?),
+        liquidation_price: row.liquidation_price.parse::<Price>()?,
         creation_timestamp: convert_to_system_time(row.ts_secs, row.ts_nanos)?,
         term: Duration::new(row.term_secs, row.term_nanos),
         origin: row.origin,
@@ -417,11 +417,11 @@ pub async fn load_all_cfds(conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<
                 id: row.uuid,
                 trading_pair: row.trading_pair,
                 position: row.position,
-                price: Usd::new(Decimal::from_str(&row.initial_price)?),
-                min_quantity: Usd::new(Decimal::from_str(&row.min_quantity)?),
-                max_quantity: Usd::new(Decimal::from_str(&row.max_quantity)?),
+                price: row.initial_price.parse::<Price>()?,
+                min_quantity: row.min_quantity.parse::<Usd>()?,
+                max_quantity: row.max_quantity.parse::<Usd>()?,
                 leverage: row.leverage,
-                liquidation_price: Usd::new(Decimal::from_str(&row.liquidation_price)?),
+                liquidation_price: row.liquidation_price.parse::<Price>()?,
                 creation_timestamp: convert_to_system_time(row.ts_secs, row.ts_nanos)?,
                 term: Duration::new(row.term_secs, row.term_nanos),
                 origin: row.origin,
@@ -527,11 +527,11 @@ pub async fn load_cfds_by_oracle_event_id(
                 id: row.uuid,
                 trading_pair: row.trading_pair,
                 position: row.position,
-                price: Usd::new(Decimal::from_str(&row.initial_price)?),
-                min_quantity: Usd::new(Decimal::from_str(&row.min_quantity)?),
-                max_quantity: Usd::new(Decimal::from_str(&row.max_quantity)?),
+                price: row.initial_price.parse::<Price>()?,
+                min_quantity: row.min_quantity.parse::<Usd>()?,
+                max_quantity: row.max_quantity.parse::<Usd>()?,
                 leverage: row.leverage,
-                liquidation_price: Usd::new(Decimal::from_str(&row.liquidation_price)?),
+                liquidation_price: row.liquidation_price.parse::<Price>()?,
                 creation_timestamp: convert_to_system_time(row.ts_secs, row.ts_nanos)?,
                 term: Duration::new(row.term_secs, row.term_nanos),
                 origin: row.origin,
@@ -540,7 +540,7 @@ pub async fn load_cfds_by_oracle_event_id(
 
             Ok(Cfd {
                 order,
-                quantity_usd: Usd::new(Decimal::from_str(&row.quantity_usd)?),
+                quantity_usd: row.quantity_usd.parse::<Usd>()?,
                 state: serde_json::from_str(row.state.as_str())?,
             })
         })
@@ -807,7 +807,7 @@ mod tests {
     impl Order {
         fn dummy() -> Self {
             Order::new(
-                Usd::new(dec!(1000)),
+                Price::new(dec!(1000)).unwrap(),
                 Usd::new(dec!(100)),
                 Usd::new(dec!(1000)),
                 Origin::Theirs,
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -1,12 +1,14 @@
 use crate::olivia;
+
 use anyhow::{Context, Result};
-use bdk::bitcoin::{Address, Amount};
+use bdk::bitcoin::{Address, Amount, Denomination};
 use reqwest::Url;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
-use rust_decimal_macros::dec;
 use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
+use std::num::NonZeroU8;
+use std::ops::{Add, Div, Mul, Sub};
 use std::time::SystemTime;
 use std::{fmt, str};
 use time::{OffsetDateTime, PrimitiveDateTime, Time};
@@ -14,49 +16,34 @@ use uuid::Uuid;
 
 pub mod cfd;
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Price of zero is not allowed.")]
+    ZeroPrice,
+    #[error("Negative Price is unimplemented.")]
+    NegativePrice,
+}
+
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
 pub struct Usd(Decimal);
 
 impl Usd {
-    pub fn new(dec: Decimal) -> Self {
-        Self(dec)
-    }
-
-    pub fn checked_add(&self, other: Usd) -> Result<Usd> {
-        let result = self.0.checked_add(other.0).context("addition error")?;
-        Ok(Usd(result))
-    }
-
-    pub fn checked_sub(&self, other: Usd) -> Result<Usd> {
-        let result = self.0.checked_sub(other.0).context("subtraction error")?;
-        Ok(Usd(result))
-    }
-
-    // TODO: Usd * Usd = Usd^2 not Usd !!!
-    pub fn checked_mul(&self, other: Usd) -> Result<Usd> {
-        let result = self
-            .0
-            .checked_mul(other.0)
-            .context("multiplication error")?;
-        Ok(Usd(result))
-    }
-
-    pub fn checked_div(&self, other: Usd) -> Result<Usd> {
-        let result = self.0.checked_div(other.0).context("division error")?;
-        Ok(Usd(result))
+    pub fn new(value: Decimal) -> Self {
+        Self(value)
     }
 
     pub fn try_into_u64(&self) -> Result<u64> {
         self.0.to_u64().context("could not fit decimal into u64")
     }
 
-    pub fn half(&self) -> Usd {
-        let half = self
-            .0
-            .checked_div(dec!(2))
-            .expect("can always divide by two");
+    pub fn try_into_f64(&self) -> Result<f64> {
+        self.0.to_f64().context("Could not fit decimal into f64")
+    }
+}
 
-        Usd(half)
+impl fmt::Display for Usd {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.round_dp(2).fmt(f)
     }
 }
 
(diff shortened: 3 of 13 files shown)


daemon/src/model.rs
```diff
@@ -80,24 +99,226 @@ impl<'de> Deserialize<'de> for Usd {
...
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, sqlx::Type)]
+pub struct Leverage {
+    value: u8,
+}
+
+impl Leverage {
+    pub fn new(value: u8) -> Result<Self, Error> {
+        if value == 0 {
+            return Result::Err(Error::NonPositiveLeverage);
+        }
```
thomaseizinger: Use https://doc.rust-lang.org/std/num/struct.NonZeroU8.html here to avoid this error case.

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
-    pub fn try_into_u64(&self) -> Result<u64> {
-        self.0.to_u64().context("could not fit decimal into u64")
-    }
+impl ExchangeRate {
+    pub fn new(value: Decimal) -> Result<Self, Error> {
+        if value.to_f64().unwrap().abs() <= f64::EPSILON {
+            return Result::Err(Error::ZeroExchangeRate);
+        } else if value.to_f64().unwrap() < 0f64 {
+            return Result::Err(Error::NegativeExchangeRate);
+        };
```
thomaseizinger: Just a style thing but I don't like `else if` blocks. I would change this to:

```suggestion
        if value.to_f64().unwrap().abs() <= f64::EPSILON {
            return Result::Err(Error::ZeroExchangeRate);
        }
        
        if value.to_f64().unwrap() < 0f64 {
            return Result::Err(Error::NegativeExchangeRate);
        };
```

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
+pub struct ExchangeRate {
+    value: Decimal,
+}
 
-    pub fn try_into_u64(&self) -> Result<u64> {
-        self.0.to_u64().context("could not fit decimal into u64")
-    }
+impl ExchangeRate {
+    pub fn new(value: Decimal) -> Result<Self, Error> {
+        if value.to_f64().unwrap().abs() <= f64::EPSILON {
```
thomaseizinger: Use https://docs.rs/rust_decimal/1.16.0/rust_decimal/prelude/struct.Decimal.html#method.is_zero instead.

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
+            return Result::Err(Error::ZeroExchangeRate);
+        } else if value.to_f64().unwrap() < 0f64 {
```
thomaseizinger: Use https://docs.rs/rust_decimal/1.16.0/rust_decimal/prelude/struct.Decimal.html#method.is_sign_negative instead.

daemon/src/model.rs
```diff
@@ -80,24 +99,226 @@ impl<'de> Deserialize<'de> for Usd {
...
+        Ok(Leverage { value })
+    }
+}
+
+// add impl's to do algebra with Usd, Leverage, and ExhangeRate as required
+impl Mul<Leverage> for Usd {
+    type Output = Usd;
+
+    fn mul(self, rhs: Leverage) -> Self::Output {
+        let out = self.value * Decimal::from_u8(rhs.value).unwrap();
```
thomaseizinger: For `u8`s, I think you should be able to use `From`: https://docs.rs/rust_decimal/1.16.0/rust_decimal/prelude/struct.Decimal.html#impl-From%3Cu8%3E

And that doesn't panic.

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
-    // TODO: Usd * Usd = Usd^2 not Usd !!!
-    pub fn checked_mul(&self, other: Usd) -> Result<Usd> {
-        let result = self
-            .0
-            .checked_mul(other.0)
-            .context("multiplication error")?;
-        Ok(Usd(result))
+impl fmt::Display for Usd {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.value.fmt(f)
```
thomaseizinger: Formatting should perhaps trim to two decimal places?

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
     }
+}
 
-    pub fn checked_div(&self, other: Usd) -> Result<Usd> {
-        let result = self.0.checked_div(other.0).context("division error")?;
-        Ok(Usd(result))
-    }
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd)]
```
thomaseizinger: Do we really want to derive `Serialize` and `Deserialize` here or should we customize the implementation as we did for `Usd`?

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
+            .to_u64()
+            .context("could not fit decimal into u64")
     }
+}
 
-    pub fn checked_sub(&self, other: Usd) -> Result<Usd> {
-        let result = self.0.checked_sub(other.0).context("subtraction error")?;
-        Ok(Usd(result))
-    }
+// impl From<f64> for Result<Usd> {
```
thomaseizinger: There is `TryFrom` is you need a fallible conversion trait.

daemon/src/model.rs
```diff
@@ -1,62 +1,81 @@
...
+pub enum Error {
+    #[error("Exchange rate of zero is not allowed.")]
+    ZeroExchangeRate,
+    #[error("Negative exchange rate is not allowed.")]
+    NegativeExchangeRate,
+    #[error("Leverage must be >= 1.")]
+    NonPositiveLeverage,
+}
+
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, PartialOrd)]
```
thomaseizinger: These derives will conflict with the manually declared ones!

daemon/src/model.rs
```diff
@@ -298,4 +522,55 @@ mod tests {
...
+        assert_eq!(usd_sum.value, dec!(11.110));
+        assert_eq!(usd_diff.value, dec!(-8.642));
+        assert_eq!(half.value, dec!(0.617));
+        assert_eq!(double.value, dec!(19.752));
+    }
+
+    #[test]
+    fn algebra_with_exchange_rate() {
+        let usd = Usd::new(dec!(123.456));
+        let rate = ExchangeRate::new(usd.value).unwrap();
```
thomaseizinger: I don't think that is a good practice for the test, better initialize this also with `dec!(123.456)`.

daemon/src/model.rs
```diff
@@ -298,4 +522,55 @@ mod tests {
...
+        let res = usd / rate;
+
+        assert_eq!(res, Amount::from_sat(100_000_000));
```
thomaseizinger: There is a constant for 1 BTC.

daemon/src/model.rs
```diff
@@ -298,4 +522,55 @@ mod tests {
...
+    }
+
+    #[test]
+    fn algebra_with_leverage() {
+        let usd = Usd::new(dec!(123.456));
+        let leverage = Leverage::new(3).unwrap();
+        let res = usd * leverage / leverage;
```
thomaseizinger, line 642: I would prefer hardcoded expected values over than computed ones.

daemon/src/model.rs
```diff
@@ -298,4 +522,55 @@ mod tests {
...
+        let usd = Usd::new(dec!(123.456));
+        let rate = ExchangeRate::new(usd.value).unwrap();
+        let res = usd / rate;
+
+        assert_eq!(res, Amount::from_sat(100_000_000));
+    }
+
+    #[test]
+    fn algebra_with_leverage() {
+        let usd = Usd::new(dec!(123.456));
```
thomaseizinger: Would be great to use test values that are closer to our real world data.

daemon/src/model.rs
```diff
@@ -298,4 +522,55 @@ mod tests {
...
+        let double = usd_1 * 2;
+
+        assert_eq!(usd_sum.value, dec!(11.110));
+        assert_eq!(usd_diff.value, dec!(-8.642));
+        assert_eq!(half.value, dec!(0.617));
+        assert_eq!(double.value, dec!(19.752));
+    }
+
+    #[test]
+    fn algebra_with_exchange_rate() {
```
thomaseizinger: This might be more expressive of what is happening here.

```suggestion
    fn usd_for_1_btc_buys_1_btc() {
```

thomaseizinger (commented):
Couple of comments to mull over :)

daemon/src/model.rs
DeliciousHair: As an FYI, I actually started with this type but the problem was that except for this error case, all the other features that it offers in terms of our use-case are still experimental and thus in nightly only. So I went back to `u8` for overall convenience.

Can change though, no bid deal either way.

thomaseizinger: You can use `NonZeroU8` in the fn signature but use a `u8` internally for convenience!

Best of both worlds :)

daemon/src/db.rs
```diff
@@ -101,11 +101,11 @@ pub async fn load_order_by_id(
         id: row.uuid,
         trading_pair: row.trading_pair,
         position: row.position,
-        price: Usd::new(Decimal::from_str(&row.initial_price)?),
-        min_quantity: Usd::new(Decimal::from_str(&row.min_quantity)?),
-        max_quantity: Usd::new(Decimal::from_str(&row.max_quantity)?),
+        price: row.initial_price.parse::<Price>()?,
```
da-kami, line 104: ❓I'm surprised these type hints are actually needed - can rust not infer the type here based on what is returned?

daemon/src/model.rs
```diff
@@ -1,66 +1,107 @@
...
 use std::time::SystemTime;
 use std::{fmt, str};
 use time::{OffsetDateTime, PrimitiveDateTime, Time};
 use uuid::Uuid;
 
 pub mod cfd;
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Exchange rate of zero is not allowed.")]
```
da-kami:
```suggestion
    #[error("Price of zero is not allowed.")]
```

daemon/src/model.rs
```diff
@@ -1,66 +1,107 @@
...
+    ZeroPrice,
+    #[error("Negative exchange rate is unimplemented.")]
```
da-kami:
```suggestion
    #[error("Negative price is unimplemented.")]
```

daemon/src/model.rs
```diff
@@ -69,29 +110,287 @@ impl Serialize for Usd {
...
+    }
+}
+
+impl Mul<InversePrice> for Usd {
+    type Output = Amount;
+
+    fn mul(self, rhs: InversePrice) -> Self::Output {
+        let mut btc = self.0 * rhs.0;
+        btc.rescale(8);
+        Amount::from_str_in(&btc.to_string(), Denomination::Bitcoin).unwrap()
```
da-kami: Just to check my understanding of the motivation for `InversePrice`: 
Was `InversePrice` introduced, so we can keep these calculation impls separate from the `Price`? Is there any other motivation for using `InversePrice` over `Price`?

daemon/src/model.rs
```diff
@@ -69,29 +110,287 @@ impl Serialize for Usd {
...
+
+impl Div<Leverage> for Leverage {
+    type Output = Decimal;
+
+    fn div(self, rhs: Leverage) -> Self::Output {
+        Decimal::from(self.0) / Decimal::from(rhs.0)
+    }
+}
+
+// TODO: leaving Percent as-is for this PR as the plan is to
```
da-kami: Please remove this Todo and create an issue for it instead. We should definitely not talk about specific PRs in Todos in the code - once the PR is merged there is no context for this Todo...

daemon/src/model.rs
```diff
@@ -302,4 +598,60 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_liquidation() {
+        let usd = Usd::new(dec!(61234.5678));
+        let leverage = Leverage::new(5).unwrap();
+        let price = Price::new(dec!(61234.5678)).unwrap();
+        let expected_buyin = Amount::from_str_in("0.2", Denomination::Bitcoin).unwrap();
+
+        let liquidation_price = price * leverage / (leverage + 1);
```
da-kami, line 654: 🛠 This test does not test any code at the moment - besides constructing types. If that is intended please rename the test to reflect that.

I think the test should test the actual code where we calculate the liquidation price, not calculate the liquidation price in the test!

daemon/src/model/cfd.rs
```diff
@@ -1303,88 +1290,102 @@ pub enum CfdStateChangeEvent {
...
-    let initial_price_inverse = dec!(1)
-        .checked_div(initial_price.0)
-        .context("Calculating inverse of initial_price resulted in overflow")?;
-    let current_price_inverse = dec!(1)
-        .checked_div(closing_price.0)
-        .context("Calculating inverse of current_price resulted in overflow")?;
+    let inv_initial_price = InversePrice::new(initial_price);
+    let inv_closing_price = InversePrice::new(closing_price);
+    let long_liquidation_price = initial_price * leverage / (leverage + 1);
+    let test = closing_price <= long_liquidation_price;
```
da-kami:
```suggestion
    let is_not_liquidation = closing_price <= long_liquidation_price;
```

daemon/src/model/cfd.rs
```diff
@@ -1303,88 +1290,102 @@ pub enum CfdStateChangeEvent {
...
+            // for short:
+            //   let:
+            //     P = payout
+            //     Q = quantity
+            //     Ll = long_leverage
+            //     Ls = short_leverage
+            //     xi = initial_price
+            //     xc = closing_price
+            //
+            //     a = xi * Ll / (Ll + 1)
```
da-kami: 🛠 Hard to understand what `a` and `b` represent - can we somehow come up with better names?

daemon/src/model/cfd.rs
```diff
@@ -1303,88 +1290,102 @@ pub enum CfdStateChangeEvent {
...
+            // for long:
+            //   let:
+            //     P = payout
+            //     Q = quantity
+            //     Ll = long_leverage
+            //     Ls = short_leverage
+            //     xi = initial_price
+            //     xc = closing_price
+            //
+            //     a = xi * Ll / (Ll + 1)
```
da-kami: 🛠 Hard to understand what `a` and `b` represent - can we somehow come up with better names?

daemon/src/model/cfd.rs
```diff
@@ -1303,88 +1290,102 @@ pub enum CfdStateChangeEvent {
...
+            //     P = {
+            //          0 if <= a,
+            //          Q / (xi * Ll) + Q * (1 / xo - 1 / xc) if a < xc < b,
+            //          Q / xi * (1/Ll + 1/Ls) if xc if xc >= b
+            //     }
+            let margin = quantity / (initial_price * leverage);
+            let payout = match test {
+                true => Amount::ZERO,
+                false => {
+                    quantity / (initial_price * leverage)
```
da-kami: I could validate that the payout is calculated with the same formula as before.

daemon/src/model/cfd.rs
```diff
@@ -1303,88 +1290,102 @@ pub enum CfdStateChangeEvent {
...
+            //     a = xi * Ll / (Ll + 1)
+            //     b = xi * Ls / (Ls - 1)
+            //
+            //     P = {
+            //          Q / xi * (1/Ll + 1/Ls) if xc <= a,
+            //          Q / (xi * Ls) - Q * (1 / xo - 1 / xc) if a < xc < b,
+            //          0 if xc >= b
+            //     }
+            let margin = quantity / initial_price;
+            let payout = match test {
```
da-kami: ❓❗️ Here we are using the long liquidation price (the `test` variable) when defining the bounds for the `Sell` (= `short`) position. Is that correct? 
The liquidation price for the `short` position is substantially higher than the one for `long`. I think the liquidation price calculation has to distinguish between `long` and `short`! Also, the liquidation price for `short` must be guarded by `>= closing_price`.

Or am I misunderstanding something?

da-kami: I don't think we should calculate the liquidation price separately in here, but use the liquidation price calculation that already exists (and is hopefully tested for sanity)!

daemon/src/model/cfd.rs
```diff
@@ -1440,64 +1427,63 @@ mod tests {
 
     #[test]
     fn given_default_values_then_expected_liquidation_price() {
-        let leverage = Leverage(5);
-        let price = Usd(dec!(49000));
-        let maintenance_margin_rate = dec!(0.005);
+        let price = Price::new(dec!(46125)).unwrap();
```
da-kami, line 1413: I validated that these values correspond to our GSheet based model. 
For future reference, it would have been great to have a commit, that changes the maintenance margin to `0` prior to removing it - that way you can showcase that the calculation is the same (i.e. tests pass with maintenance margin `0` and without maintenance margin based on the same input and expect).

daemon/src/model/cfd.rs
```diff
@@ -1522,65 +1508,65 @@ mod tests {
...
-            Usd::from(dec!(10_000)),
-            Amount::from_btc(0.01984).unwrap(),
+            Price::new(dec!(50_400)).unwrap(),
+            Price::new(dec!(60_000)).unwrap(),
+            Usd::new(dec!(10_000)),
+            Leverage::new(1).unwrap(),
             Position::Buy,
             SignedAmount::from_sat(3_174_603),
-            dec!(160.01024065540194572452620968).into(),
+            dec!(15.99999899200001).into(),
```
da-kami: I think it is sane, but I have a hard time validating that change 😅

da-kami (changes requested):
Great work! This improves understandability a lot!

I have one blocking comment concerning the guard for the `short` `payout` in the `profit` calculation that is based on the `long`'s liquidation price. It might just be a misunderstanding on my side though 😅

Comment on commits: 

Please make sure to clean up the commit history! I think it's good enough to squash the code into one commit. For future work it would be nice to have some clean commits with distinct work, so we can review by commit. 
e.g. for the `.gitignore` file change it would have been great to have a separate commit and a commit message that explains why this change was done in one sentence.

daemon/src/model.rs
DeliciousHair: Nope, none. It's just a safe + convenient way of doing `1/price` is all, since the guardrails are on `Price::new()`

daemon/src/model.rs
DeliciousHair: oops!

daemon/src/db.rs
DeliciousHair: Hm, you may be correct... Lemme check that I didn't make extra work here.

EDIT:
Nope, this is "correct" in this context as we are parsing the data to `Decimal` type which is not something we can handle out-of-the-box just yet (see [this issue](https://github.com/launchbadge/sqlx/issues/1489) in particular). This is not to say it cannot be improved, only that this is where we are for the moment.

daemon/src/model/cfd.rs
DeliciousHair: > Or am I misunderstanding something?

I think you may be; can we please discuss this at some point?

daemon/src/model/cfd.rs
DeliciousHair: I'm going to say not really, as in this case I'm trying to show what the algebra does, as opposed to how the "concept" works. In this context, using variables `a` and `b` I think make good sense as it removes the context completely, which is the objective in the comment.

daemon/src/model/cfd.rs
DeliciousHair: Sorry, I'm not sure what you're saying here?

daemon/src/model/cfd.rs
DeliciousHair: Yeah, I couldn't figure out where the first number was coming from either; suspect there was actually a logic error that was being allowed in via the test, but I really don't know TBH.

daemon/src/model/cfd.rs
DeliciousHair: Fair enough, but this further ties back to detail that the `maintenance_margin` has now been stripped out completely until it has been properly defined.

daemon/src/model/cfd.rs
DeliciousHair: > I don't think we should calculate the liquidation price separately in here, but use the liquidation price calculation that already exists (and is hopefully tested for sanity)!

The existing function is:
```rust
fn calculate_liquidation_price(leverage: Leverage, price: Price) -> Price {
}
```
which I wanted to remove TBH as having such trivial calculations relegated out to distinct functions makes the code harder to read IMHO as I the maintainer have to navigate around to discover that a function that does something akin to `a + b` has been used at some location rather than just writing `a + b`.

If the math was more complicated, sure! But in this case, I really think it is just making things less clear.

Happy enough to go either way, but will make consistent.

Also, re: testing please note from `model.rs` that:
```rust
#[test]
fn test_liquidation() {
    let usd = Usd::new(dec!(61234.5678));
    let leverage = Leverage::new(5).unwrap();
    let price = Price::new(dec!(61234.5678)).unwrap();
    let expected_buyin = Amount::from_str_in("0.2", Denomination::Bitcoin).unwrap();

    let liquidation_price = price * leverage / (leverage + 1);
    let inv_price = InversePrice::new(price);
    let inv_liquidation_price = InversePrice::new(liquidation_price);

    let long_buyin = usd / (price * leverage);
    let long_payout =
        (usd / leverage) * ((leverage + 1) * inv_price - leverage * inv_liquidation_price);

    assert_eq!(long_buyin, expected_buyin);
    assert_eq!(long_payout, Amount::ZERO);
}
```

.gitignore
```diff
@@ -17,6 +17,7 @@
 .pnp.*
 
 # Artifacts from running the daemons
+*.sqlite*
 /*.sqlite*
```
klochowicz:
```suggestion
```

klochowicz: nit: I believe you don't need this one, if we're ignoring all the sqlite databases in the repo with your new change

daemon/src/model.rs
```diff
@@ -69,29 +110,290 @@ impl Serialize for Usd {
...
+        Ok(Price(dec))
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+pub struct InversePrice(Decimal);
+
+impl InversePrice {
+    pub fn new(value: Price) -> InversePrice {
+        Self(Decimal::ONE / value.0)
```
klochowicz: could we please add a check that `value.0` is not `ZERO`? (I know that currently price can't be created with zero, but it could be a footgun if we relax that requirement or perform calculations resulting in zero.

daemon/src/model.rs
```diff
@@ -69,29 +110,290 @@ impl Serialize for Usd {
...
+    type Output = Decimal;
+
+    fn div(self, rhs: Leverage) -> Self::Output {
+        Decimal::from(self.0) / Decimal::from(rhs.0)
+    }
+}
+
+// TODO
+// Create a new PnL struct at some point and incorporate this value
+// rahther than having a distinct struct for it.
```
klochowicz: Perhaps we could create a ticket instead of TODO? (I think we agreed on that during last retro)

daemon/src/model/cfd.rs
```diff
@@ -8,19 +8,22 @@ use bdk::miniscript::DescriptorTrait;
 use cfd_protocol::secp256k1_zkp::{self, EcdsaAdaptorSignature, SECP256K1};
 use cfd_protocol::{finalize_spend_transaction, spending_tx_sighash, TransactionExt};
 use rocket::request::FromParam;
-use rust_decimal::prelude::{FromPrimitive, ToPrimitive};
+use rust_decimal::prelude::FromPrimitive;
 use rust_decimal::Decimal;
-use rust_decimal_macros::dec;
+
```
klochowicz: Please remove the empty lines here as it interferes with `rustfmt` (@thomaseizinger always points it out in my reviews 😄 )

(here and below)

daemon/src/model/cfd.rs
```diff
@@ -155,23 +156,8 @@ impl Order {
     }
 }
 
-fn calculate_liquidation_price(
-    leverage: &Leverage,
-    price: &Usd,
-    maintenance_margin_rate: &Decimal,
```
klochowicz, line 161: what happened to `maintenance_margin_rate`? It seems odd that it was removed as part of this PR. If you'd like to remove it, I'd prefer to open up a separate PR so we can focus and discuss it, instead of blocking the refactor

klochowicz (commented):
it's getting close, I guess the commits should be squashed with some nice description before it can get merged.

daemon/src/model.rs
DeliciousHair: Sorry, but I disagree. The types are in fact being tested in the sense that we are creating some values and doing some algebra. The fact that the algebra is testing that the liquidation price yields a payout of zero is merely a convenient test value.

And again, computing the liquidation price is trivial--the function in another module is the exact same one-line operation.

As for renaming the test, alright.

daemon/src/model/cfd.rs
da-kami: I am always comparing what we have in code with the model that we have in Google sheets. That way I can somehow verify that we have the expected outcome (besides the tests). I just added those comments so that other reviewers might see that it was verified against the GSheet model. that what @bonomat and me did with previous PRs that changed calculations.

daemon/src/model.rs
DeliciousHair: Since `Price.0 = ZERO` is blocked, I don't  see what you're getting at here as the check would serve no purpose.

klochowicz: I mentioned above -e - if anyone ever decides to support price being `0` or be able to calculate it (as a result of some computation, not via `new()`), then this will be an error waiting to happen.
It's much easier to be explicit here, instead of forcing a reader to put trust that these types will be set in lockstep. 

Is there any harm in being a bit more explicit here stating the contract?

daemon/src/model/cfd.rs
DeliciousHair: Given that this is a comment only and is written up as formulae, I think the use of the single-letter values is appropriate within the context.

daemon/src/model/cfd.rs
DeliciousHair: Ah, gotcha!

daemon/src/model/cfd.rs
DeliciousHair: Fixed with current update, should all be well now!

daemon/src/model/cfd.rs
DeliciousHair: It was simply dropped as it was ad-hoc'd into things as the pedagogy of said item relied on some cumbersome algebra. With the current structure it is easy to compute on the fly in both the payout curve and the profit calculation(s), or can be stuck onto, say, the `Cfd` struct as a static value--doesn't really matter TBH; note that the resulting functions in `cfd.rs` are simply:

```rust
pub fn calculate_buy_margin(price: Price, quantity: Usd, leverage: Leverage) -> Amount {
    quantity / (price * leverage)
}

fn calculate_sell_margin(price: Price, quantity: Usd) -> Amount {
    quantity / price
}

fn calculate_liquidation_price(leverage: Leverage, price: Price) -> Price {
}
```

daemon/src/model.rs
DeliciousHair: I suppose not, so added it is!

daemon/src/model.rs
da-kami: It's still a Todo, can we move to a ticket?

daemon/src/model.rs
```diff
@@ -302,4 +609,60 @@ mod tests {
...
+    fn leverage_does_not_alter_type() {
+        let usd = Usd::new(dec!(61234.5678));
+        let leverage = Leverage::new(3).unwrap();
+        let res = usd * leverage / leverage;
+
+        assert_eq!(res.0, usd.0);
+    }
+
+    #[test]
+    fn test_liquidation_as_expected() {
```
da-kami:
```suggestion
    fn test_algebra() {
```

Reasoning: This test is not about testing the liquidation price but just algebra and constructing types. The liquidation price is just used as an example, but no code calculating the liquidation price is tested.

da-kami (approved):
nice, thanks!

daemon/src/model.rs
DeliciousHair: That is actually (more or less) what I had and was told to change it to (more or less) the name that is there.  I'll change it again as I prefer your suggestion, but perhaps we should create a style-guide for this sort of thing as there is clearly some inconsistency happening?

daemon/src/model/cfd.rs
klochowicz: This explanation makes sense, thank you! Would you mind mentioning it also in the commit message? otherwise people might be wondering why it went away during a refactor.

klochowicz (approved):

PR #406, opened 2021-10-21
Added tests for matrix-solver

DeliciousHair:
Added missing tests for `CSR::matrix_solve()` and fixed some
error handling in the process.

### DIFF
--- a/daemon/src/payout_curve.rs
+++ b/daemon/src/payout_curve.rs
@@ -152,6 +152,8 @@ pub enum Error {
     CannotInitCSR,
     #[error("matrix must be square")]
     MatrixMustBeSquare,
+    #[error("cannot invert singular matrix")]
+    SingularMatrix,
     #[error("evaluation outside parametric domain")]
     InvalidDomain,
     #[error("einsum error--array size mismatch?")]
--- a/daemon/src/payout_curve/csr_tools.rs
+++ b/daemon/src/payout_curve/csr_tools.rs
@@ -55,7 +55,7 @@ impl CSR {
             .map(|e| {
                 let b = b_arr.slice(s![.., e]).to_owned();
 
-                let sol = lu_solve(&a_arr, &b).unwrap();
+                let sol = lu_solve(&a_arr, &b)?;
                 Ok(sol.to_vec())
             })
             .collect::<Result<Vec<_>, Error>>()?;
@@ -89,15 +89,19 @@ impl CSR {
 }
 
 fn lu_solve(a: &Array2<f64>, b: &Array1<f64>) -> Result<Array1<f64>, Error> {
+    if !is_square(a) {
+        return Result::Err(Error::MatrixMustBeSquare);
+    }
     let a = a.to_nalgebra_matrix().lu();
     let b = b.to_nalgebra_matrix();
 
-    let x = a
-        .solve(&b)
-        .ok_or(Error::MatrixMustBeSquare)?
-        .to_1d_array()?;
+    let x = a.solve(&b).ok_or(Error::SingularMatrix)?.to_1d_array()?;
 
     Ok(x)
+}
+
+fn is_square<T>(arr: &Array2<T>) -> bool {
+    arr.shape()[0] == arr.shape()[1]
 }
 
 impl Mul<&Array1<f64>> for CSR {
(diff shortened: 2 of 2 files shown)


daemon/src/payout_curve/csr_tools.rs
```diff
@@ -89,6 +92,9 @@ impl CSR {
 }
 
 fn lu_solve(a: &Array2<f64>, b: &Array1<f64>) -> Result<Array1<f64>, Error> {
+    if a.shape()[0] != a.shape()[1] {
+        return Result::Err(Error::MatrixMustBeSquare);
+    }
```
thomaseizinger: Are we calling `lu_solve` somewhere outside of `matrix_solve`? I think we should only have one check.

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -48,6 +48,9 @@ impl CSR {
     // matrix version of `solve()`; useful for solving AX = B. Implementation
     // is horrible.
     pub fn matrix_solve(&self, b_arr: &Array2<f64>) -> Result<Array2<f64>, Error> {
+        if self.is_square() == false {
```
thomaseizinger:
```suggestion
        if !self.is_square() {
```

A bit more idiomatic I'd say.

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -127,14 +133,7 @@ mod tests {
...
 
-        for (x, expected) in x.into_iter().zip(x_expected) {
-            assert!(
-                (x - expected).abs() < f64::EPSILON * 10.,
-                "{} {}",
-                x,
-                expected
-            )
-        }
+        assert!(x.abs_diff_eq(&x_expected, 1e-10));
```
thomaseizinger, line 134: Neat!

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -149,4 +148,61 @@ mod tests {
...
+            Array1::<usize>::from_vec(vec![0, 1, 1, 2, 0, 2]),
+            Array1::<usize>::from_vec(vec![0, 2, 4, 6]),
+            (3, 3),
+        )
+        .unwrap();
+        let b = Array2::<f64>::from(vec![
+            [36., 59., 82.],
+            [204., 249., 294.],
+            [198., 262., 326.],
+        ]);
```
thomaseizinger: This could be slightly more concise using the macro: https://docs.rs/ndarray/0.15.3/ndarray/macro.array.html

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -149,4 +148,61 @@ mod tests {
...
+            Array1::<usize>::from_vec(vec![0, 1, 1, 1]),
+            (3, 3),
+        )
+        .unwrap();
+        let b = Array2::<f64>::from(vec![
+            [36., 59., 82.],
+            [204., 249., 294.],
+            [198., 262., 326.],
+        ]);
+        let e = a.matrix_solve(&b).unwrap_err();
```
thomaseizinger, line 174: I'd prefer a newline above this statement so that the test is visually grouped into:

ARRANGE

ACT

ASSERT

daemon/src/payout_curve/csr_tools.rs
```diff
@@ -149,4 +148,61 @@ mod tests {
 
         assert!(matches!(a, Error::CannotInitCSR));
     }
+
+    // test that all is good
+    #[test]
+    fn test_lu_matrix_solve_00() {
+        let a = CSR::new(
+            Array1::<f64>::from_vec(vec![11., 12., 22., 23., 31., 33.]),
```
thomaseizinger: The parameter type in `new` enforces the type of the array (`f64`), hence you wouldn't need to specify it here. That would reduce some noise in these tests.

thomaseizinger (changes requested):
A few nits, otherwise good improvement!

daemon/src/payout_curve/csr_tools.rs
DeliciousHair: Ah, you're right. We were, but that was in the mass of code that got axed. Will fix.

thomaseizinger (approved):
Thanks, LGTM!

daemon/src/payout_curve.rs
```diff
@@ -153,6 +153,8 @@ pub enum Error {
     #[error("matrix must be square")]
     MatrixMustBeSquare,
     #[error("evaluation outside parametric domain")]
+    SingularMatrix,
+    #[error("cannot invert singular matrix")]
```
da-kami, line 155: I think some lines got flipped here, looks like the `#[error("cannot invert singular matrix")]` should be above `SingularMatrix` and `#[error("evaluation outside parametric domain")]` on top of `InvalidDomain` ?

da-kami (approved):
great, thanks!

daemon/src/payout_curve.rs
DeliciousHair: Not sure what happened there, but good catch! Fixed.

PR #407, opened 2021-10-21
Xtra productivity hell yeah

da-kami:
- [x] Optimize message block generation
- [x] Use the actual variable names that were used in the function sig in the generated code
- [x] Get the xtra::Context into the macro (third argument that can be specified on function sig)


thomaseizinger:
The beauty of macros is, the code can be ugly / duplicated. As long as there is a test and it passes, happy days 😁

daemon/src/model/cfd.rs
```diff
@@ -999,11 +999,11 @@ impl Cfd {
         Ok(Some(new_state))
     }
 
-    pub fn refund_tx(&self) -> Result<Transaction> {
+    pub fn refund_tx(&self) -> Option<Transaction> {
```
da-kami: Will clean that up later, accidentally committed some changes that were not supposed to be where, `cfd.rs`

daemon/src/maker_inc_connections.rs
```diff
@@ -161,78 +208,10 @@ impl Actor {
...
-    }
-}
-
-#[async_trait]
-impl Handler<ListenerMessage> for Actor {
     async fn handle(&mut self, msg: ListenerMessage, ctx: &mut Context<Self>) -> KeepRunning {
         match msg {
             ListenerMessage::NewConnection { stream, address } => {
-                self.handle_new_connection(stream, address, ctx).await;
+                self.handle_new_connection_impl(stream, address, ctx).await;
```
thomaseizinger, line 214: I think inlining this function would be better now.

xtra_productivity/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    let actor = block.self_ty;
+
+    let code = block
+        .items
+        .into_iter()
+        .filter_map(|block_item| match block_item {
+            ImplItem::Method(method) => Some(method),
+            _ => None,
+        }).map(|method| {
+        let message_arg = method.sig.inputs[1].clone();
```
thomaseizinger: Is dprint not running on these for some reason? Why is line 18 not indented?

daemon/src/maker_cfd.rs
```diff
@@ -579,7 +579,7 @@ where
                 taker_id,
                 command: TakerCommand::NotifyOrderAccepted { id: order_id },
             })
-            .await?;
+            .await??;
```
thomaseizinger: This is a behavioural change, can we easily use `spawn_fallible` here?

daemon/src/maker_cfd.rs
```diff
@@ -759,7 +759,7 @@ where
                     oracle_event_id,
                 },
             })
-            .await?;
+            .await??;
```
thomaseizinger: Same here.

daemon/src/maker_cfd.rs
thomaseizinger: Or, in case you want to keep it, can we have it as a separate commit just in case we need to bisect a bug?

thomaseizinger (changes requested):
The behaviour change is blocking, otherwise LGTM.

daemon/src/maker_cfd.rs
klochowicz: Is it not as a separate commit already? Can't really split it further, without introducing and removing the spawn fallible in a commit it-between. 

IMHO we really should stop the contract setup if it fails instead of just logging it.

daemon/src/maker_inc_connections.rs
klochowicz: I honestly prefer the function, as it is 36 lines long, and it would be harder to follow if we inline it :( perhaps a better name, or a comment there that it's not an `xtra::Handler`, but an internal forwarding?

daemon/src/maker_cfd.rs
da-kami: yeah, the behaviour should result in the same thing, in the end we always log an error (just on the inside now instead of the outside) if I understand it correctly.

thomaseizinger: I see two commits:

1. Removing log_error and use spawn_fallible on the calling side

2. Use xtra_productivity to remove boilerplate.

I would like to retain functionality for now, feel free to propose a follow-up where we discuss changing the behaviour.

xtra_productivity/src/lib.rs
da-kami: It does not pick it up, but I really don't get why

daemon/src/maker_inc_connections.rs
thomaseizinger: Just drop the impl postfix then :)

daemon/src/maker_cfd.rs
klochowicz: I don't think it would actually be a behavioral change, as it will still result in a `log_error()` in the maker_cfd actor handler, it will just stop stop going further after this failure, as it is currently. I can be even more explicit about this behavioral change in the commit perhaps?

xtra_productivity/src/lib.rs
klochowicz: `rustfmt` doesn't seem to work there either.

da-kami: So, apparently this behaves just weird - when I changed the `ui.rs` file in the `tests` it suddenly picked it up:

```
[VERBOSE]: Reading file: /Users/dakami/CoBloX/repos/hermes/xtra_productivity/src/lib.rs
[VERBOSE]: Formatted file: /Users/dakami/CoBloX/repos/hermes/xtra_productivity/src/lib.rs in 3ms
[VERBOSE]: Writing file: /Users/dakami/CoBloX/repos/hermes/xtra_productivity/src/lib.rs
Formatted 2 files.
[VERBOSE]: Writing file: /Users/dakami/Library/Caches/dprint/cache/hermes.incremental

```

Really not sure why this happens, but I would guess it was a cache issue? Would not explain why it was not picked up in CI though.

klochowicz: for me the `rustfmt` can easily handle a simple `fn hello_world()` in that file, but it does not pick up stuff with that derive macro :(

daemon/src/maker_cfd.rs
klochowicz: ok, I'll create a follow-up with that simplification separately.

thomaseizinger (approved):

daemon/src/maker_cfd.rs
klochowicz: Started #418

daemon/src/maker_inc_connections.rs
klochowicz: oops, forgot about it, will do this along with other cleanups tho!

PR #409, opened 2021-10-21
Backport: Fix bug in fanning out messages to multiple actors in the maker

thomaseizinger:


daemon/src/maker.rs
```diff
@@ -353,7 +354,7 @@ where
                 .notify_interval(Duration::from_secs(5), || oracle::Sync)
                 .unwrap(),
         );
-        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &cfd_actor_addr])
+        let fan_out_actor = fan_out::Actor::new(&[&cfd_actor_addr, &monitor_addr])
```
da-kami, line 357: ❓ I have not noticed this in the code base, what does this line do?

da-kami (approved):

bonomat (approved):

daemon/src/maker.rs
thomaseizinger: From the docs:

> /// A fan-out actor takes every incoming message and forwards it to a set of other actors.

We are using this pattern with the oracle actor to forward the attestation to both, the cfd actor and the monitoring actor without the oracle actor having to loop over a list of addresses.

da-kami: I see, I yesterday experienced the "problem" that we did not transition from `PendingCET` to `Closed` automatically. given the change you made, I assume that could be related to the monitoring actor not gettgin the Attestation reported (and thus not looking for `CetFinality` - is that correct?

PR #411, opened 2021-10-21
Show cfd expiry in UI

bonomat:
Looks so pretty :) 

image

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -207,6 +207,10 @@ impl BitMexPriceEventId {
             .join(&self.to_string())
             .expect("Event id can be joined")
     }
+
+    pub fn timestamp(&self) -> OffsetDateTime {
+        self.timestamp
+    }
 }
 
 impl fmt::Display for BitMexPriceEventId {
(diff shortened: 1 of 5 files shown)


daemon/src/model/cfd.rs
```diff
@@ -689,6 +689,10 @@ impl Cfd {
         (self.order.term * Cfd::REFUND_THRESHOLD).as_blocks().ceil() as u32
     }
 
+    pub fn expiry_timestamp(&self) -> i64 {
+        self.order.oracle_event_id.as_unix_timestamp()
```
bonomat: Ideally this is stored as a field on the CFD and we don't have to parse it from the oracle event id but I didn't want to change the data model.

daemon/src/model.rs
```diff
@@ -207,6 +207,10 @@ impl BitMexPriceEventId {
             .join(&self.to_string())
             .expect("Event id can be joined")
     }
+
+    pub fn as_unix_timestamp(&self) -> i64 {
+        self.timestamp.unix_timestamp()
+    }
```
thomaseizinger: I think I'd prefer if we expose the strongly-typed timestamp (`OffsetDatetime`) instead of the unix timestamp.

thomaseizinger (approved):

daemon/src/model.rs
bonomat: Force pushed, please check again :)

thomaseizinger (approved):
Nice use of the `serde` module from `time`!

PR #412, opened 2021-10-21
All the term to be configurable on the maker's commandline

thomaseizinger:
- Use the original order's term for rolling over
- Allow term to be configurable on the maker's commandline

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -812,6 +812,7 @@ mod tests {
                 Usd::new(dec!(1000)),
                 Origin::Theirs,
                 BitMexPriceEventId::with_20_digits(OffsetDateTime::now_utc()),
+                time::Duration::hours(24),
             )
             .unwrap()
         }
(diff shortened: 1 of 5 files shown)


da-kami (approved):
boy that was easy :D

klochowicz (approved):

bonomat (approved):
👍

bonomat:
I rebased for you @thomaseizinger

bonomat:
I think it is important to keep in mind that the `term` should not be seen as a feature of a CFD. A CFD does not have an expiry (unlike options or a futures). A CFD never ends and hence does not have a term. This `term` here is an implementation necessity because we cannot have never ending CFDs with the features Bitcoin provides today.

thomaseizinger:
> I think it is important to keep in mind that the `term` should not be seen as a feature of a CFD. A CFD does not have an expiry (unlike options or a futures). A CFD never ends and hence does not have a term. This `term` here is an implementation necessity because we cannot have never ending CFDs with the features Bitcoin provides today. 
> 

That is a good reminder, thank you! We should definitely word this accordingly. @da-kami @klochowicz something to keep in mind for the model make-over.

PR #413, opened 2021-10-21
Remove all usages of `#[allow(dead_code)]`

thomaseizinger:
This is a legacy from not having a library crate.


klochowicz (approved):
nice! I grepped through the code and could not see any remaining.

PR #414, opened 2021-10-21
Initialize Cfd actors outside of Rocket

klochowicz:
Allows us to use `?` instead of unwrapping.

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -15,7 +15,6 @@ use daemon::{
     oracle, wallet_sync,
 };
 use futures::Future;
-use rocket::fairing::AdHoc;
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::SqlitePool;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


daemon/src/maker.rs
```diff
@@ -176,70 +175,52 @@ async fn main() -> Result<()> {
...
+
+    housekeeping::transition_non_continue_cfds_to_setup_failed(&mut conn).await?;
+    housekeeping::rebroadcast_transactions(&mut conn, &wallet).await?;
+
+    let ActorSystem {
+        cfd_actor_addr,
+        cfd_feed_receiver,
+        order_feed_receiver,
+        update_cfd_feed_receiver,
+    } = ActorSystem::new(
```
thomaseizinger, line 193: I think we `unwrap` in one of these constructors because we coulnd't return a `Result` before, can we clean that up as well?

thomaseizinger (approved):
Nice!

daemon/src/maker.rs
klochowicz: we unwrapped in both actually ;) fixed all the occurrences of unwrap, should be cleaner now!

thomaseizinger (approved):

PR #417, opened 2021-10-22
Add custom derive for `Actor`

thomaseizinger:


daemon/src/maker_cfd.rs
```diff
@@ -942,60 +945,74 @@ where
...
 }
 
 #[async_trait]
-impl<O: 'static, M: 'static, T: 'static> Handler<NewTakerOnline> for Actor<O, M, T>
+impl<O, M, T> Handler<NewTakerOnline> for Actor<O, M, T>
 where
-    T: xtra::Handler<maker_inc_connections::TakerMessage>,
+    T: xtra::Handler<maker_inc_connections::TakerMessage> + 'static + Send,
+    O: 'static + Send,
+    M: 'static + Send,
```
thomaseizinger, line 964: Putting this as a draft because of these changed requirements. Once #296 is done and `log_error` is gone we can use `xtra_productivity` for this and then applying the custom derive should be easier!

thomaseizinger:
Might pick up later.

PR #418, opened 2021-10-22
Stop contract and rollover from the maker if cannot notify the taker

klochowicz:
As for now, we were only logging the communication problem and moving on.


da-kami (approved):
It makes sense to fail in case we are unable to communicate with the taker at these points.

PR #419, opened 2021-10-22
Ensure we always know about Order::Term * announcements

bonomat:
When looking into https://github.com/comit-network/hermes/issues/360 I realized we forgot to trigger `FetchAnnouncement` when receiving a roll-over request. 
This is fixed with this PR.

I found this a bit cumbersome to have to do this manually. Instead, we should sync regularly and ensure we have attestations for the next 24h. 

This is also related to the discussion @da-kami and @thomaseizinger had in https://github.com/comit-network/hermes/issues/349

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -1,6 +1,6 @@
 use crate::model::cfd::{Cfd, CfdState};
 use crate::model::BitMexPriceEventId;
-use crate::{log_error, tokio_ext};
+use crate::{log_error, tokio_ext, try_continue};
 use anyhow::{Context, Result};
 use async_trait::async_trait;
 use cfd_protocol::secp256k1_zkp::{schnorrsig, SecretKey};
@@ -9,25 +9,18 @@ use serde::Deserialize;
 use std::collections::{HashMap, HashSet};
 use std::ops::Add;
 use time::ext::NumericalDuration;
+use time::Duration;
 use xtra::prelude::StrongMessageChannel;
 use xtra_productivity::xtra_productivity;
 
 pub struct Actor {
     announcements: HashMap<BitMexPriceEventId, (OffsetDateTime, Vec<schnorrsig::PublicKey>)>,
-    pending_announcements: HashSet<BitMexPriceEventId>,
     pending_attestations: HashSet<BitMexPriceEventId>,
     attestation_channel: Box<dyn StrongMessageChannel<Attestation>>,
+    term: Duration,
 }
 
 pub struct Sync;
-
-/// Message used to tell the `oracle::Actor` to fetch an
-/// `Announcement` from `olivia`.
-///
-/// The `Announcement` corresponds to the `OracleEventId` included in
-/// the message.
-#[derive(Debug, Clone)]
-pub struct FetchAnnouncement(pub BitMexPriceEventId);
 
 pub struct MonitorAttestation {
     pub event_id: BitMexPriceEventId,
@@ -69,6 +62,7 @@ impl Actor {
     pub fn new(
         cfds: Vec<Cfd>,
         attestation_channel: Box<dyn StrongMessageChannel<Attestation>>,
+        term: Duration,
     ) -> Self {
         let mut pending_attestations = HashSet::new();
 
@@ -100,14 +94,23 @@ impl Actor {
 
         Self {
             announcements: HashMap::new(),
-            pending_announcements: HashSet::new(),
             pending_attestations,
             attestation_channel,
+            term,
         }
     }
 
-    fn update_pending_announcements(&mut self, ctx: &mut xtra::Context<Self>) {
-        for event_id in self.pending_announcements.iter().cloned() {
+    fn ensure_having_announcements(&mut self, term: Duration, ctx: &mut xtra::Context<Self>) {
+        // we want inclusive the term length hence +1
+        for hour in 1..term.whole_hours() + 1 {
+            let event_id = try_continue!(next_announcement_after(
+                time::OffsetDateTime::now_utc() + Duration::hours(hour)
+            ));
+
+            if self.announcements.get(&event_id).is_some() {
+                tracing::trace!("Announcement already known: {}", event_id,);
+                continue;
+            }
             let this = ctx.address().expect("self to be alive");
 
             tokio_ext::spawn_fallible(async move {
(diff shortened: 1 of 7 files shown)


daemon/src/oracle.rs
```diff
@@ -19,14 +20,9 @@ pub struct Actor {
...
 
 pub struct Sync;
-
-/// Message used to tell the `oracle::Actor` to fetch an
-/// `Announcement` from `olivia`.
-///
-/// The `Announcement` corresponds to the `OracleEventId` included in
-/// the message.
-#[derive(Debug, Clone)]
-pub struct FetchAnnouncement(pub BitMexPriceEventId);
```
thomaseizinger, line 30: With the removal of this, we don't actually need to sync every 5 seconds but can increase that interval.

This was only 5 seconds because we wanted to "quickly" react to incoming messages of `FetchAnnouncement`. That was sort of a hack anyway from the beginning so I am happy that it is going away!

daemon/src/oracle.rs
```diff
@@ -199,20 +195,33 @@ impl Actor {
     }
 }
 
-#[async_trait]
-impl xtra::Handler<MonitorAttestation> for Actor {
-    async fn handle(&mut self, msg: MonitorAttestation, _ctx: &mut xtra::Context<Self>) {
-        if !self.pending_attestations.insert(msg.event_id) {
-            tracing::trace!("Attestation {} already being monitored", msg.event_id);
+impl Actor {
+    fn ensure_having_announcements(&mut self, term: Duration) {
```
thomaseizinger: Once we increase the interval, I think we can merge this with sync to simplify the interface of this actor.

daemon/src/oracle.rs
```diff
@@ -19,14 +20,9 @@ pub struct Actor {
...
-/// Message used to tell the `oracle::Actor` to fetch an
-/// `Announcement` from `olivia`.
-///
-/// The `Announcement` corresponds to the `OracleEventId` included in
-/// the message.
-#[derive(Debug, Clone)]
-pub struct FetchAnnouncement(pub BitMexPriceEventId);
+/// Message used to tell the `oracle::Actor` to fetch
+/// `Announcements` from `olivia` for the provided duration.
+pub struct FetchAnnouncements(pub Duration);
```
thomaseizinger: With https://github.com/comit-network/hermes/pull/412, the term will be configurable. Should we just make the term an argument to the `oracle::Actor`? Fits the constructor better than the message IMO, esp. given that this message is going away if we merge it with `Sync`.

daemon/src/oracle.rs
```diff
@@ -199,20 +195,33 @@ impl Actor {
...
+            if self.announcements.get(&oracle_event_id).is_some() {
+                tracing::trace!("Announcement already known: {}", oracle_event_id,);
+                continue;
+            }
+            if self.pending_announcements.insert(oracle_event_id) {
+                tracing::debug!(
+                    "Will fetch on next sync announcement for event: {}",
+                    oracle_event_id,
+                );
+            }
```
thomaseizinger: If this gets merged with `sync`, I think the entire state of `pending_announcements` can go away because you can just directly use `spawn_fallible` in here to spawn away the fetching.

daemon/src/taker.rs
```diff
@@ -296,6 +296,16 @@ where
...
         );
+
+        oracle_ctx.notify(oracle::FetchAnnouncements(Order::TERM));
+        tokio::spawn(
+            oracle_ctx
+                .notify_interval(Duration::from_secs(60 * 20), || {
+                    oracle::FetchAnnouncements(Order::TERM)
+                })
+                .map_err(|e| anyhow::anyhow!(e))?,
+        );
```
thomaseizinger: For simplicity, we can have the taker sync announcements as well, otherwise we will need two different APIs for talking to the oracle actor.

It is not strictly necessary though so we might want to keep that in mind if it ever causes problems.

thomaseizinger (changes requested):
Thanks!

I think there is a simpler design where we merge the two sync messages.

daemon/src/taker.rs
```diff
@@ -28,6 +28,8 @@ use xtra::{Actor, Address};
 mod connection;
 mod routes_taker;
 
+pub const TERM: time::Duration = time::Duration::hours(24);
```
bonomat, line 30: The changes in here are not ideal because of the what was introduced in #412. 

As mentioned in https://github.com/comit-network/hermes/pull/412#issuecomment-949986797 
a CFD does not have a term. This is an implementation detail and it should actually not be tempered with besides for test purposes. Hence, I reintroduce a constant for this.

daemon/src/oracle.rs
```diff
@@ -298,19 +285,16 @@ impl xtra::Actor for Actor {}
 #[async_trait]
 impl xtra::Handler<Sync> for Actor {
     async fn handle(&mut self, _: Sync, ctx: &mut xtra::Context<Self>) {
-        self.update_pending_announcements(ctx);
+        self.ensure_having_announcements(self.term, ctx);
```
thomaseizinger: Passing `self.term` is odd for a fn that is called on `self`.

daemon/src/taker.rs
thomaseizinger: Perhaps we should just call this `ANNOUNCEMENT_LOOKAHEAD` or something. At least on the taker side, this variable has nothing to do with the "term".

daemon/src/oracle.rs
```diff
@@ -99,16 +93,26 @@ impl Actor {
 
         Self {
             announcements: HashMap::new(),
```
thomaseizinger, line 96: I think we should actually save those to the DB. Seems unnecessary to fetch them on every startup of the app.

thomaseizinger (approved):
Nice, thank you!

Much in favor of this implementation. More resilient and smaller API surface too!

daemon/src/taker.rs
da-kami: > a CFD does not have a term

I think this is a terminology problem. Even in a traditional cfd you have to define the start and end price that the cfd is bound to. So, there is no term in the sense of closing the cfd, but there can be recurring fees that are bound to time periods, such as daily interest. 
What we call "term" just tries to put the time period between start price and end price into code. I am happy to rename it - could not come up with a better word than "term". I am not sure `ANNOUNCEMENT_LOOKAHEAD` makes this clearer. What do you think @DeliciousHair ?

DeliciousHair: I think `ANNOUNCEMENT_LOOKAHEAD` or perhaps `ANNOUNCEMENT_TIME_INTERVAL` are improvements, yes. I'm adding the second suggestion as it makes it very clear that this is a temporal value.

daemon/src/oracle.rs
da-kami: I think saving in db should be done in a follow up.

klochowicz: shall I create a ticket to track it, so it won't get lost?

da-kami (approved):
I don't see any blockers, currently rebasing this work and will see to get it merged once done!

I created a follow-up PR for renaming `term`:

https://github.com/comit-network/hermes/pull/424

daemon/src/oracle.rs
da-kami: yeah, please create one!

da-kami:
I think it is fair to assume that #426 is unrelated to the code changes in this PR. Keeping #426 for further investigation and merging this PR now.

daemon/src/oracle.rs
klochowicz: #430

PR #424, opened 2021-10-25
Rename `term` to `settlement_time_interval` and `announcement_lookahead`

da-kami:
Follow up for #419

Depending on the context, this makes it more explicit what we mean by `term` in the context of the `Cfd` and the `Oracle`.

### DIFF
--- a/daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
+++ b/daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
@@ -1,0 +1,5 @@
+alter table orders
+rename column term_seconds to settlement_time_interval_seconds;
+
+alter table orders
+rename column term_nanoseconds to settlement_time_interval_nanoseconds;
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -51,9 +51,9 @@ struct Opts {
     #[clap(short, long)]
     json: bool,
 
-    /// The term length of each CFD in hours.
+    /// The time interval until potential settlement of each CFD in hours
     #[clap(long, default_value = "24")]
-    term: u8,
+    settlement_time_interval_hours: u8,
 
     #[clap(subcommand)]
     network: Network,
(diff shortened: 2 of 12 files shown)


daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
```diff
@@ -0,0 +1,5 @@
+alter table orders
+rename column term_seconds to settlement_time_interval_seconds;
+
+alter table orders
+rename column term_nanoseconds to settlement_time_interval_nanoseconds;
```
DeliciousHair, line 5: Does this really need to be an addition SQL file, or can you simply change the existing table definitions?

da-kami: given that we had a release, I though it might be worth starting to use the migrations as intended, so we can also ensure backwards compatibility and don't always have to delete the database.

DeliciousHair: I see what you're saying, but as this is still at POC stage (release or not) do we really want / need to be backwards compatible at this point?

daemon/src/maker.rs
```diff
@@ -51,9 +51,9 @@ struct Opts {
     #[clap(short, long)]
     json: bool,
 
-    /// The term length of each CFD in hours.
+    /// The time interval until potential settlement of each CFD in hours.
     #[clap(long, default_value = "24")]
-    term: u8,
+    settlement_time_interval: u8,
```
klochowicz: perhaps `settlement_time_interval_hours`? (this would make the comment redundant!)

daemon/src/maker_cfd.rs
```diff
@@ -751,8 +752,9 @@ where
 
         let dlc = cfd.open_dlc().context("CFD was in wrong state")?;
 
-        let oracle_event_id =
-            oracle::next_announcement_after(time::OffsetDateTime::now_utc() + cfd.order.term)?;
+        let oracle_event_id = oracle::next_announcement_after(
+            time::OffsetDateTime::now_utc() + cfd.order.settlement_time_interval,
```
klochowicz: as per comment earlier: places like this would read much nices if we added `_hours` suffix

klochowicz (approved):
LGTM! 

A naming suggestion, but it's not blocking if you disagree

PR #429, opened 2021-10-26
Encrypt network communication using the noise protocol.

### LINKED ISSUE #229: Introduce authentication and encryption for the network layer
Eventually we will transition the network layer to libp2p so we have a more structured way of designing network communication.

Up until then, we still want proper identification of peers (our current `TakerId` is transient with every connection), authentication of each other's messages and encryption of the entire communication.

To make the transition to libp2p later on easier, we first will:

- Use [`snow`](https://docs.rs/snow/0.8.0/snow/index.html) for the heavy lifting of encrypting and authenticating the connection
- Use the libp2p multi-addresses specification to configure the connection details to the other party

Using snow will require that we know the public key of the party ahead of time. Using libp2p's multiaddress provides several benefits:

- Specify connection details + public key in a single argument
- Allows usage of domains as well as IP addresses (https://github.com/comit-network/hermes/issues/292)

rishflab:
- [x]   Use snow for encrypting the connection
- [ ]   Use snow for authentication the connection
- [ ]   Use the libp2p multi-addresses specification to configure the connection details to the other party

authentication can be done by using this function on the handshake state https://docs.rs/snow/0.8.0/snow/struct.TransportState.html#method.get_remote_static to check if the static_key matches the peers static key

I am trying to figure out how to associate the libp2p PeerId with the this static key

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -31,6 +31,7 @@ serde_json = "1"
 serde_plain = "1"
 serde_with = { version = "1", features = ["macros"] }
 sha2 = "0.9"
+snow = "0.8"
 sqlx = { version = "0.5", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
 thiserror = "1"
 time = { version = "0.3", features = ["serde"] }
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -1,6 +1,8 @@
-use crate::{send_to_socket, taker_cfd, wire};
+use crate::{noise, send_to_socket, taker_cfd, wire};
+use anyhow::Result;
 use futures::{Stream, StreamExt};
 use std::net::SocketAddr;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tokio_util::codec::FramedRead;
 use xtra::prelude::MessageChannel;
@@ -15,11 +17,14 @@ pub struct Actor {
 }
 
 impl Actor {
-    pub async fn new(maker: SocketAddr) -> Self {
-        let (read, write) = loop {
+    pub async fn new(maker: SocketAddr) -> Result<Self> {
+        let (read, write, noise) = loop {
             let socket = tokio::net::TcpSocket::new_v4().expect("Be able ta create a socket");
-            if let Ok(connection) = socket.connect(maker).await {
-                break connection.into_split();
+            if let Ok(mut connection) = socket.connect(maker).await {
+                let noise = noise::initiator_handshake(&mut connection).await?;
+
+                let (read, write) = connection.into_split();
+                break (read, write, Arc::new(Mutex::new(noise)));
             } else {
                 tracing::warn!(
                     "Could not connect to the maker, retrying in {}s ...",
@@ -29,16 +34,16 @@ impl Actor {
             }
         };
 
-        let send_to_maker = send_to_socket::Actor::new(write)
+        let send_to_maker = send_to_socket::Actor::new(write, noise.clone())
             .create(None)
             .spawn_global();
 
-        let read = FramedRead::new(read, wire::JsonCodec::default())
+        let read = FramedRead::new(read, wire::JsonCodec::new(noise))
             .map(move |item| taker_cfd::MakerStreamMessage { item });
 
-        Self {
+        Ok(Self {
             send_to_maker: Box::new(send_to_maker),
             read_from_maker: Box::new(read),
-        }
+        })
     }
 }
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -1,12 +1,13 @@
 use crate::maker_cfd::{FromTaker, NewTakerOnline};
 use crate::model::cfd::{Order, OrderId};
 use crate::model::{BitMexPriceEventId, TakerId};
-use crate::{forward_only_ok, maker_cfd, send_to_socket, wire};
+use crate::{forward_only_ok, maker_cfd, noise, send_to_socket, wire};
 use anyhow::{Context as AnyhowContext, Result};
 use futures::{StreamExt, TryStreamExt};
 use std::collections::HashMap;
 use std::io;
 use std::net::SocketAddr;
+use std::sync::{Arc, Mutex};
 use tokio::net::TcpStream;
 use tokio_util::codec::FramedRead;
 use xtra::prelude::*;
@@ -94,16 +95,18 @@ impl Actor {
 
     async fn handle_new_connection_impl(
         &mut self,
-        stream: TcpStream,
+        mut stream: TcpStream,
         address: SocketAddr,
         _: &mut Context<Self>,
-    ) {
+    ) -> Result<()> {
         let taker_id = TakerId::default();
 
         tracing::info!("New taker {} connected on {}", taker_id, address);
 
+        let noise = Arc::new(Mutex::new(noise::responder_handshake(&mut stream).await?));
+
         let (read, write) = stream.into_split();
-        let read = FramedRead::new(read, wire::JsonCodec::default())
+        let read = FramedRead::new(read, wire::JsonCodec::new(noise.clone()))
             .map_ok(move |msg| FromTaker { taker_id, msg })
             .map(forward_only_ok::Message);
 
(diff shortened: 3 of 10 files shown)


daemon/src/wire.rs
```diff
@@ -97,20 +99,22 @@ impl fmt::Display for MakerToTaker {
...
             inner: LengthDelimitedCodec::new(),
+            transport_state,
         }
     }
 }
 
 impl<T> Decoder for JsonCodec<T>
 where
-    T: DeserializeOwned,
+    T: DeserializeOwned + std::marker::Sync,
```
da-kami: nit: can we just import `Sync` instead of fully qualifying here?

da-kami (commented):
Looks good so far, I did not look into the specifics of noise yet though, maybe we can talk about it at some point. 

For testing: You can already rebase on master and run the test(s) in `happy-path.rs`. We spin up the actor system there, I suppose if that test passes the transport setup should work.

da-kami:
> Use the libp2p multi-addresses specification to configure the connection details to the other party

I suppose that was discussed with Thomas? What is the advantage of doing that now?

rishflab:
> > Use the libp2p multi-addresses specification to configure the connection details to the other party
> 
> I suppose that was discussed with Thomas? What is the advantage of doing that now?

yeh we dont have to. I already started on it so I have an identity to authenticate against for the authentication part.

da-kami:
> > > Use the libp2p multi-addresses specification to configure the connection details to the other party
> > 
> > 
> > I suppose that was discussed with Thomas? What is the advantage of doing that now?
> 
> yeh we dont have to. I already started on it so I have an identity to authenticate against for the authentication part.

So, the advantage is, that we don't have to implement identity setup ourselves on top of the authentication layer? What would be the alternative to useing libp2p?

klochowicz:
>     * [x]    Use snow for encrypting the connection
> 
>     * [ ]    Use snow for authentication the connection
> 
>     * [ ]    Use the libp2p multi-addresses specification to configure the connection details to the other party
> 

is it possible to create PRs on each of these stages? or is it all-or-nothing?

daemon/src/noise.rs
```diff
@@ -0,0 +1,26 @@
+use lazy_static::lazy_static;
+use snow::params::NoiseParams;
+use std::io;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpStream;
+
+lazy_static! {
+    pub static ref PARAMS: NoiseParams = "Noise_XXpsk3_25519_ChaChaPoly_BLAKE2s".parse().unwrap();
```
klochowicz: it seems a bit excessive to pull `lazy_static` when it can be easily avoided.

something like this should do the trick:

```rust
use snow::params::*`
pub const PARAMS: &str = "Noise_XXpsk3_25519_ChaChaPoly_BLAKE2s";

            let builder: Builder<'_> = Builder::new(PARAMS.parse().unwrap());
```

side note:  using `NOISE_PARAMS` name would be nicer, especially if you decide to import it in other places (if you kept `noise::PARAMS` then the name would be ok). Otherwise reader has no idea what `PARAMS` means at a glance). No need to change it though, as I suspect that it can stay local to `noise.rs`.

daemon/src/noise.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+pub static SECRET: &[u8] = b"i don't care for fidget spinners";
+
```
klochowicz, line 55: Perhaps add a factory function that encapsulates the builder, so you don't need to leak the PARAMS and SECRET at all from this module?

```rust
        // Initialize our initiator using a builder.
fn initialize() -> Result<HandshakeState> {
        let builder: Builder<'_> = Builder::new(PARAMS.parse().unwrap());
        let static_key = builder.generate_keypair().unwrap().private;
        let mut noise = builder
            .local_private_key(&static_key)
            .psk(3, SECRET)
            .build_initiator()
     }
```

daemon/src/noise.rs
klochowicz: I'm pretty green in regards to encryption - is this a good choice? (I've noticed it's the one used in the example, but I don't know about its merits). In case you're unsure, perhaps add a TODO.

daemon/Cargo.toml
```diff
@@ -15,6 +15,7 @@ futures = { version = "0.3", default-features = false }
 hex = "0.4"
 hkdf = "0.11"
 itertools = "0.10"
+lazy_static = "1.4"
```
klochowicz: I suspect this dependency can be removed if other comments are actioned

klochowicz (commented):
good start! 👍 have you tried rebasing and testing with our new test yet?

daemon/src/noise.rs
rishflab: I am also green. I don't think we need the XXpsk3 part. I can remove that along so we dont need a shared SECRET.

daemon/src/noise.rs
rishflab: yep good idea

rishflab:
> > ```
> > * [x]    Use snow for encrypting the connection
> > 
> > * [ ]    Use snow for authentication the connection
> > 
> > * [ ]    Use the libp2p multi-addresses specification to configure the connection details to the other party
> > ```
> 
> is it possible to create PRs on each of these stages? or is it all-or-nothing?

best to do it seperately, i need to do some reading on authentication with the noise framework

daemon/src/connection.rs
```diff
@@ -15,11 +17,27 @@ pub struct Actor {
...
-        let (read, write) = loop {
+    pub async fn new(maker: SocketAddr) -> Result<Self> {
+        let (read, write, noise) = loop {
             let socket = tokio::net::TcpSocket::new_v4().expect("Be able ta create a socket");
-            if let Ok(connection) = socket.connect(maker).await {
-                break connection.into_split();
+            if let Ok(mut connection) = socket.connect(maker).await {
+                let mut noise = noise::init_initiator().await?;
+
+                let mut buf = vec![0u8; 65535];
```
da-kami: ❓ `65535` is `u16::MAX` - what's the reason for that buffer size?

da-kami: If it is literally meant to be that constant, then maybe just use `u16::MAX`, makes it more descriptive.

daemon/src/connection.rs
```diff
@@ -15,11 +17,27 @@ pub struct Actor {
...
+                let len = noise.write_message(&[], &mut buf)?;
+                noise::send(&mut connection, &buf[..len]).await?;
+
+                noise.read_message(&noise::recv(&mut connection).await?, &mut buf)?;
+
+                let len = noise.write_message(&[], &mut buf)?;
+                noise::send(&mut connection, &buf[..len]).await?;
+
+                let noise = noise.into_transport_mode()?;
+                tracing::debug!("Transition the state machine into transport mode now that the handshake is complete");
```
da-kami:
```suggestion
                tracing::debug!("Handshake complete, p2p network communication ready");
```

daemon/src/maker_inc_connections.rs
```diff
@@ -94,16 +95,35 @@ impl Actor {
...
+        let len = noise.write_message(&[0u8; 0], &mut buf)?;
+        noise::send(&mut stream, &buf[..len]).await?;
+
+        noise.read_message(&noise::recv(&mut stream).await?, &mut buf)?;
+
+        let noise = noise.into_transport_mode()?;
+
+        let transport_state = Arc::new(Mutex::new(noise));
+        tracing::debug!(
+            "Transition the state machine into transport mode now that the handshake is complete"
```
da-kami: same suggestion as for taker

daemon/src/maker_inc_connections.rs
```diff
@@ -94,16 +95,35 @@ impl Actor {
...
         _: &mut Context<Self>,
-    ) {
+    ) -> Result<()> {
         let taker_id = TakerId::default();
 
         tracing::info!("New taker {} connected on {}", taker_id, address);
 
+        let mut noise = noise::init_responder().await?;
+
+        let mut buf = vec![0u8; 65535];
```
da-kami: same suggestion as for taker

daemon/src/wire.rs
```diff
@@ -136,7 +148,16 @@ where
     fn encode(&mut self, item: T, dst: &mut BytesMut) -> Result<(), Self::Error> {
         let bytes = serde_json::to_vec(&item)?;
 
-        self.inner.encode(bytes.into(), dst)?;
+        let mut buf = vec![0u8; 65535];
+
+        let mut transport = self
+            .transport_state
+            .lock()
+            .expect("acquired mutex lock on Noise object to encrypt message");
```
da-kami, line 156:
```suggestion
            .expect("to be able to acquire mutex lock on Noise object to encrypt message");
```

da-kami (approved):
thanks, few minor suggestions!

daemon/src/connection.rs
rishflab: in the spec for noise it says the max message size 65535. I think that is correct/explicit

Issue #373, opened 2021-10-18
Replace `Buy` and `Sell` enum with something more appropriate


DeliciousHair:
Extends #357 and ties to #365.

This enum doesn't actually belong in the logic flow, so it's current use is somewhat ad-hoc and a bit confusing.

da-kami:
Would it be appropriate to use `long` and `short` instead of `buy` and `sell` here or would this also be confusing?

DeliciousHair:
That was my thinking, yes.

DeliciousHair:
resolved via #433

PR #433, opened 2021-10-27
Rename Buy and Sell to Long and Short

DeliciousHair:
Addresses #373. Change made for consistency of language and removes some ambiguity in the business logic as a side-effect.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -136,7 +136,7 @@ impl Order {
         settlement_time_interval_hours: Duration,
     ) -> Result<Self> {
         let leverage = Leverage::new(2)?;
-        let liquidation_price = calculate_liquidation_price(leverage, price);
+        let liquidation_price = calculate_long_liquidation_price(leverage, price);
 
         Ok(Order {
             id: OrderId::default(),
@@ -146,7 +146,7 @@ impl Order {
             leverage,
             trading_pair: TradingPair::BtcUsd,
             liquidation_price,
-            position: Position::Sell,
+            position: Position::Short,
             creation_timestamp: SystemTime::now(),
             settlement_time_interval_hours,
             origin,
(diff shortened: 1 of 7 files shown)


daemon/src/model/cfd.rs
```diff
@@ -1301,28 +1301,33 @@ impl AsBlocks for Duration {
...
+fn calculate_short_margin(price: Price, quantity: Usd) -> Amount {
     quantity / price
 }
 
-fn calculate_liquidation_price(leverage: Leverage, price: Price) -> Price {
+fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Price {
     price * leverage / (leverage + 1)
 }
 
+// PLACEHOLDER
```
klochowicz, line 1326: nit: prefer `TODO`

frontend/src/TakerApp.tsx
```diff
@@ -206,7 +206,7 @@ export default function App() {
                                     makeNewOrderRequest(payload);
                                 }}
                             >
-                                BUY
+                                LONG
```
klochowicz: IMHO to be meaningful to the user, this button should name an action its doing, ie. use a verb. "Go long", "Take", or even "Buy" (as it used to be)? Or anything else that would be less confusing than pure "Long".

klochowicz (approved):

frontend/src/TakerApp.tsx
DeliciousHair: Is this the text that appears on the button?

daemon/src/model/cfd.rs
DeliciousHair: It's not really a TODO though, but literally is a placeholder that has meaning + functionality once maker-goes-long is implemented.

frontend/src/TakerApp.tsx
klochowicz: indeed :) this is the string that's visible in the UI. the simplest thing would be to keep it as "Buy", but I'm happy with anything that makes sense, we can adjust it later as well.

DeliciousHair: NP, I didn't realize that was the visual text!

da-kami (approved):

Issue #434, opened 2021-10-27
Use timestamp provided by Bitmex


DeliciousHair:
At present, the timestamp data is ignored and `SystemTime::now()` is used. This can lead to errors of the order of several minutes.

DeliciousHair:
Closed via #442

PR #435, opened 2021-10-27
Use timestamp data from Bitmex

DeliciousHair:
Addresses #434. Quote data includes a `timestamp` column and this PR allows us to consume it.


DeliciousHair:
#442 makes this PR pointless.

PR #436, opened 2021-10-27
Consume both feeds in test

da-kami:
Minor changes on the test we already have. When creating another test I noticed that it reads nicer to just join on both channels. We might even pack the join into the `next` function at some point and return from both - I did not go that far yet. 

Note that this is also closer to what we do in the production code, because we now send the message with `do_send_async` which is what we do there.

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -25,8 +25,10 @@ async fn taker_receives_order_from_maker_on_publication() {
 
     assert!(is_next_none(&mut taker.order_feed).await);
 
+    maker.publish_order(new_dummy_order());
+
     let (published, received) = tokio::join!(
-        maker.publish_order(new_dummy_order()),
+        next_some(&mut maker.order_feed),
         next_some(&mut taker.order_feed)
     );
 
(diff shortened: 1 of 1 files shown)


DeliciousHair (approved):
LGTM!

daemon/tests/happy_path.rs
```diff
@@ -153,11 +153,11 @@ impl Wallet {
 
 /// Maker Test Setup
 struct Maker {
-    cfd_actor_addr:
```
klochowicz, line 154: if possible, let's use the version with `_addr` suffix in both Maker and Taker - then it's the same as in production code. and I guess it's good to reiterate it's just the address , not the actual actor being stored here.

klochowicz (approved):
indeed, it reads nicer! I like this pattern.

daemon/tests/happy_path.rs
da-kami: Actually, I do mind I think. Maybe we should change it in production code.

Reasoning:
The fact that this is an address and not the actor itself is an implementation detail for me.
When we talk about it we would also say that we "send a message to the actor" and not to the address of the actor.

da-kami:
note: pushed a fixup that does not change much in the context of this test, but might change things with more complex tests.

daemon/tests/happy_path.rs
da-kami: If you insist I can change it, it's a minor thing for me :D

klochowicz: I prefer, as it's nicer to read - or you can just drop this commit, so it stays as it was :p

klochowicz: but it's not a blocker, so if you want to keep going, that's fine too.

da-kami: 🤷‍♂️

PR #437, opened 2021-10-27
Test reject order

da-kami:
There is several improvements that can be done on top of this, but we can merge this work and go on from here. 

I opted for writing specific helper functions for `next_cfd` and `next_order` that handle both maker and taker channel. I think that reads nicer and it makes it easier to use (internally it uses the generic impls so we avoid code duplication).

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -8,7 +8,7 @@ use daemon::model::cfd::{Cfd, CfdState, Order};
 use daemon::model::{Price, Usd, WalletInfo};
 use daemon::tokio_ext::FutureExt;
 use daemon::{
-    connection, db, logger, maker_cfd, maker_inc_connections, monitor, oracle, taker_cfd, wallet,
+    connection, db, maker_cfd, maker_inc_connections, monitor, oracle, taker_cfd, wallet,
 };
 use rand::thread_rng;
 use rust_decimal_macros::dec;
@@ -18,13 +18,17 @@ use std::str::FromStr;
 use std::task::Poll;
 use std::time::{Duration, SystemTime};
 use tokio::sync::watch;
+use tracing::subscriber::DefaultGuard;
 use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
 use xtra::spawn::TokioGlobalSpawnExt;
 use xtra::Actor;
 use xtra_productivity::xtra_productivity;
 
 #[tokio::test]
 async fn taker_receives_order_from_maker_on_publication() {
+    let _guard = init_tracing();
     let (mut maker, mut taker) = start_both().await;
 
     assert!(is_next_none(&mut taker.order_feed).await);
@@ -42,6 +46,7 @@ async fn taker_receives_order_from_maker_on_publication() {
 
 #[tokio::test]
 async fn taker_takes_order_and_maker_rejects() {
+    let _guard = init_tracing();
     let (mut maker, mut taker) = start_both().await;
 
     // TODO: Why is this needed? For the cfd stream it is not needed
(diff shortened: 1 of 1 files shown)


da-kami:
Fun error:

```
 failures:

---- taker_receives_order_from_maker_on_publication stdout ----
thread 'taker_receives_order_from_maker_on_publication' panicked at 'Unable to install global subscriber: SetGlobalDefaultError { _no_construct: () }', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-subscriber-0.2.25/src/fmt/mod.rs:571:14
```

Most likely that's tracing being annoying in tests... Will see if I can fix that by using a test subscriber

da-kami:
I'll convert this to a Draft until the tracing problem is fixed.

daemon/tests/happy_path.rs
```diff
@@ -36,6 +44,42 @@ async fn taker_receives_order_from_maker_on_publication() {
     assert_eq!(published.id, received.id);
 }
 
+#[tokio::test]
+async fn taker_takes_order_and_maker_rejects() {
+    let _guard = init_tracing();
+    let (mut maker, mut taker) = start_both().await;
+
+    // TODO: Why is this needed? For the cfd stream it is not needed
```
klochowicz, line 52: it's written it the function comment. what's more perplexing is why is it *not* needed in the `cfd_feed`!

daemon/tests/happy_path.rs
```diff
@@ -36,6 +44,42 @@ async fn taker_receives_order_from_maker_on_publication() {
...
+
+    maker.reject_take_request(received.clone());
+
+    let (taker_cfd, maker_cfd) = next_cfd(&mut taker.cfd_feed, &mut maker.cfd_feed).await;
+    // TODO: More elaborate Cfd assertions
+    assert_eq!(taker_cfd.order.id, received.id);
+    assert_eq!(maker_cfd.order.id, received.id);
+    assert!(matches!(taker_cfd.state, CfdState::Rejected { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::Rejected { .. }));
+}
```
klochowicz, line 81: Perhaps you could add an assertion about the maker's order feed now? This would be a nice conclusion to this test.

daemon/tests/happy_path.rs
```diff
@@ -44,6 +88,30 @@ fn new_dummy_order() -> maker_cfd::NewOrder {
     }
 }
 
+/// Returns the first `Cfd` from both channels
+///
+/// Ensures that there is only one `Cfd` present in both channels.
+async fn next_cfd(
```
klochowicz, line 94: side note: I'm still warming up to the  "next" naming, but couldn't figure out a better name myself, so it might as well stay like that. and the helpers themselves are useful!

klochowicz (approved):
I like the new helpers and the test!!! great job 🧑‍🏭

daemon/tests/happy_path.rs
da-kami: I also don't like the `next` that much - I thought about going with `updated` or something like that instead, but did not like that better so far

da-kami: I chose the `next` because that's what used in rust when processing streams

daemon/tests/happy_path.rs
da-kami: it might be related to the `Option` - not sure. But I still find it weird that we need it for the `Order` - in my understanding the `watch` channel should always return the latest element irregardless of what was there before, so, if we first send `None` and then `Some(order)` and then we borrow from it I would expect `Some(order)` not `None`.

daemon/tests/happy_path.rs
da-kami: I think we should go for more fine-granular tests at some point, so we only have one assert at the end of each test. That will make the tests easier to understand and it becomes more clear what is tested. 
The tests run really fast, so we can have really a lot of them (especially for the early ones before contract setup which takes some times...).

So far I am still pushing for getting more complex tests in to see how the test framework evolves. That said, it might be a good idea to add more assertions now and later just remove them by adding more tests. But I also don't wanna plaster the code with asserts :D

da-kami:
merging this as is for now and will see to evolve it when adding additional tests

PR #439, opened 2021-10-27
Send higher precision Usd and Price over the wire and trim it for the UI purposes

### LINKED ISSUE #438: Inconsistency with Price/Usd precision between taker and maker
Spotted by trying to compare the Order stored on maker and taker side - `liquidation_price` had much higher precision when stored in the maker. 

We could either round the calculation results or store the higher precision consistently in both backends.

klochowicz:
Fixes #438

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -25,7 +25,7 @@ pub enum Error {
     NegativePrice,
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Usd(Decimal);
 
 impl Usd {
@@ -40,31 +40,16 @@ impl Usd {
     pub fn try_into_f64(&self) -> Result<f64> {
         self.0.to_f64().context("Could not fit decimal into f64")
     }
+
+    #[must_use]
+    pub fn into_decimal(self) -> Decimal {
+        self.0
+    }
 }
 
 impl fmt::Display for Usd {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         self.0.round_dp(2).fmt(f)
-    }
-}
-
-impl Serialize for Usd {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        <Decimal as Serialize>::serialize(&self.0.round_dp(2), serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for Usd {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let dec = <Decimal as Deserialize>::deserialize(deserializer)?.round_dp(2);
-
-        Ok(Usd(dec))
     }
 }
 
@@ -77,7 +62,7 @@ impl str::FromStr for Usd {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Price(Decimal);
 
 impl Price {
@@ -100,25 +85,10 @@ impl Price {
     pub fn try_into_f64(&self) -> Result<f64> {
         self.0.to_f64().context("Could not fit decimal into f64")
     }
-}
 
-impl Serialize for Price {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        <Decimal as Serialize>::serialize(&self.0.round_dp(2), serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for Price {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let dec = <Decimal as Deserialize>::deserialize(deserializer)?.round_dp(2);
-
-        Ok(Price(dec))
+    #[must_use]
+    pub fn into_decimal(self) -> Decimal {
+        self.0
     }
 }
 
@@ -137,7 +107,7 @@ impl str::FromStr for Price {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, PartialOrd)]
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct InversePrice(Decimal);
 
 impl InversePrice {
@@ -159,26 +129,6 @@ impl InversePrice {
 
     pub fn try_into_f64(&self) -> Result<f64> {
         self.0.to_f64().context("Could not fit decimal into f64")
-    }
-}
-
-impl Serialize for InversePrice {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        <Decimal as Serialize>::serialize(&self.0.round_dp(2), serializer)
-    }
-}
-
-impl<'de> Deserialize<'de> for InversePrice {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let dec = <Decimal as Deserialize>::deserialize(deserializer)?.round_dp(2);
-
-        Ok(InversePrice(dec))
     }
 }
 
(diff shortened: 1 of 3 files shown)


klochowicz:
I'm having second thoughts - do we really need such high precision for anything? Perhaps we could round it to cents in the results of our calculations and keep the current rounding everywhere....

klochowicz:
> I'm having second thoughts - do we really need such high precision for anything? Perhaps we could round it to cents in the results of our calculations and keep the current rounding everywhere....

we could constrain the `model::Price` and `model::Usd` to double digit precision (ie. cents) in their constructors and have it like this across the board, as long this does not trip any calculations of margin and the payout curve (which should only work on dollar precision)

DeliciousHair:
> we could constrain the `model::Price` and `model::Usd` to double digit precision (ie. cents) in their constructors and have it like this across the board, as long this does not trip any calculations of margin and the payout curve (which should only work on dollar precision)

We definitely do **not** want to do this (at least with `Price`) as it will throw a spanner in the works for all BTC --> sats computations.

daemon/src/to_sse_event.rs
```diff
@@ -13,6 +13,62 @@ use std::time::{SystemTime, UNIX_EPOCH};
...
+
+impl Serialize for Price {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        <model::Price as Serialize>::serialize(&self.inner.round_dp(2), serializer)
+    }
+}
+
```
DeliciousHair, line 72: I don't think that truncating to 2 dp's like this in both values is the correct approach, as it may lead to a situation where `model::Usd / model::Price != Usd / Price` owing to the rounding, keeping in mind that division was implemented via:

```rust
impl Div<Price> for Usd {
    type Output = Amount;

    fn div(self, rhs: Price) -> Self::Output {
        let mut btc = self.0 / rhs.0;
        btc.rescale(8);
        Amount::from_str_in(&btc.to_string(), Denomination::Bitcoin)
            .expect("Error computing BTC amount")
    }
}
```

I could be wrong too, of course, but I think it's worth looking into as we do not want any discrepancy in the payout values.

da-kami:
> I'm having second thoughts - do we really need such high precision for anything? Perhaps we could round it to cents in the results of our calculations and keep the current rounding everywhere....

I think from a financial system point of view you would always calculate at high precision until you have to go for a type that cannot handle it, e.g. cutting Bitcoin to Satoshi precision. So only once converting to Bitcoin we should cut the precision I think.

daemon/src/model.rs
```diff
@@ -52,7 +60,7 @@ impl Serialize for Usd {
     where
         S: serde::Serializer,
     {
-        <Decimal as Serialize>::serialize(&self.0.round_dp(2), serializer)
+        <Decimal as Serialize>::serialize(&self.0.round_dp(PRECISION), serializer)
```
da-kami: Why do we still need this when `Serializing/De-serializing`? 

I would have gone with the approach:

Internally every Decimal number is capped at a precision of `x` (where `x=24` here).
So, not only for Serilalization/Deserialization, but for creating the any `Decimal` based number as well. 
Otherwise, we can run into the same problem again - if we treat Serialization different than the internal handling it can diverge. Or am I missing something?

daemon/src/to_sse_event.rs
```diff
@@ -13,6 +13,62 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use time::OffsetDateTime;
 use tokio::sync::watch;
 
+#[derive(Debug, Clone)]
+pub struct Usd {
+    inner: model::Usd,
+}
+
+impl Usd {
```
da-kami, line 20: I think if we never use the `to_sse_event` types for calculations it should be fine. But unfortunately it is not that simple, because the margin is calculated (in the daemon) based on values sent from the UI, see `routes_taker::margin_calc`. Note that this value is also purely for display purposes and is **not** actually used (i.e. it is not sent from the UI to the daemon for usage, but we always calculate the margin separately in the daemon with high precision). So, it could happen that what the user sees is slightly less precise than the value that is actually locked up. 

To avoid this, I think the most correct thing would be to represent the number properly on the UI side as well (using a decimal type) and then decide on top of that how to display it. I don't think we have to go that far yet though.

daemon/src/to_sse_event.rs
da-kami: So, for division we already reacale to Bitcoin precision. We might want to consider just using Bitcoin precision overall then (i.e. precision of `8` instead of `24`).

daemon/src/to_sse_event.rs
DeliciousHair: > To avoid this, I think the most correct thing would be to represent the number properly on the UI side as well (using a decimal type) and then decide on top of that how to display it. I don't think we have to go that far yet though.

I agree with this and don't think it qualifies as "that far" as it would actually be the path of least resistance in a very real sense--both sides would be using the same values and types, and the UI does some truncation for display purposes. Would eliminate the (somewhat redundant) need for the additional types introduced in this PR in fact.

daemon/src/to_sse_event.rs
klochowicz: `to_sse_event::{Price,Usd`} are types that are *only* used for the purposes of the UI (displaying, no calculations). it can only round and serialize, nothing else ;)

daemon/src/model.rs
klochowicz: Removed the custom implementations of Serialize/Deserialize from the model. there was no need for them. thanks for the prompt!

daemon/src/to_sse_event.rs
klochowicz: ideally margin should be calculated on the daemon side as well.

I don't if mind we move the rounding to the UI afterwards - as such it wasn't a focus of this PR, here I'm just keeping the current behaviour and fixing an unnecessary rounding in the backend which used to happen :)

I see good parts in both ways, on one side: 
- convenience, if the data is just for the purpose of our UI, we might as well massage it to be exactly as we want it (thus "simplifying" the UI further), 
vs
- cleaner division of responsibilities, the backend should not care about how the UI wants to represent (and not change alongside the UI, theoretically allowing for interchangeable different UIs) .

daemon/src/model.rs
```diff
@@ -100,25 +85,10 @@ impl Price {
...
-impl<'de> Deserialize<'de> for Price {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        let dec = <Decimal as Deserialize>::deserialize(deserializer)?.round_dp(2);
 
-        Ok(Price(dec))
+    #[must_use]
+    pub fn into_decimal(self) -> Decimal {
```
da-kami, line 90: 🛠 More ergonomic rust would be a `From` implementation, i.e. `From<Price> for Decimal`

daemon/src/model.rs
```diff
@@ -40,6 +40,11 @@ impl Usd {
     pub fn try_into_f64(&self) -> Result<f64> {
         self.0.to_f64().context("Could not fit decimal into f64")
     }
+
+    #[must_use]
+    pub fn into_decimal(self) -> Decimal {
```
da-kami, line 45: 🛠 More ergonomic rust would be a `From` implementation, i.e. `From<Price> for Decimal`

da-kami (approved):

daemon/src/to_sse_event.rs
```diff
@@ -12,6 +12,64 @@ use std::convert::TryInto;
...
+
+impl Serialize for Price {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        <Decimal as Serialize>::serialize(&self.inner.into_decimal(), serializer)
+    }
+}
+
```
DeliciousHair, line 72: Still a bit confused as to why this is being done, but mystery aside the implementation looks well good to me!

DeliciousHair (approved):

daemon/src/to_sse_event.rs
klochowicz: if I try to serialize e.g. from `model::Usd`, I get an error:
```
thread 'to_sse_event::tests::price_serializes_with_only_cents' panicked at 'expected Token::Str("1000.12") but serialized as NewtypeStruct { name: "Price", }', /Users/mariusz/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_test-1.0.130/src/ser.rs:198:9
```

The idea is to strip the newtypes for the UI, Typescript does not care about our structs 😄

daemon/src/model.rs
klochowicz: thank you! it would indeed be more idiomatic. I deliberately tried to make it less comfy here, as it's a bit of a hack :(

PR #440, opened 2021-10-27
Only transition maker to `ContractSetup` if we succeed to contract taker

da-kami:
Stumbled on this when writing actor tests.
We cannot create a failing test for this code at the moment, because our current test framework does not allow us to test at this granulairty level.
Nevertheless, we should fix this.

Rational:

If the maker is unable to contact the taker about the contract setup we fail now (previously we tried to continue).
This means, that we should only transition the maker's cfd to the new state if we are able to contact the maker.


klochowicz (approved):
good find!

Issue #364, opened 2021-10-18
Contract setup can fail for one but succeed for the other party


da-kami:
Following scenario:

1. Maker and Taker go into contract setup
2. Either party sends `Msg2` but does not receive `Msg2` (before the timeout)
3. The party that received both messages will continue to publish lock, the party other party will fail in contract setup and transition to `SetupFailed` upon restart (and then does not get any updates on the `Cfd` anymore)

This is critical, because the party for which the contract setup failed won't get any information on locked up funds in case this happens. Eventually, when the party for which it worked published commit and cet the party for which the setup failed will receive back the funds, but that just randomly happens without the party knowing. 

First, we could argue, that we should remove `Msg2` from the contract setup and do it afterwards. But that does not really solve the problem as a whole, because we would then still have the problem for `Msg1`.

What we need, is to persist all the information we have before we send `Msg1`, because after doing so the other party is able to publish the lock transaction. So, upon restart we have to monitor for lock, in case we have sent `Msg1` (irregardless of the contract setup failing). 

I think it is OK to keep `Msg2` in the mix, although we could also move that part of the code outside the contract setup. 

We could generalize this problem into:
`Once you signed something and sent it to the other party we have to make sure that we have that state persisted so we can reason about it`. We could solve part of this by extending the broadcast in the wallet.

thomaseizinger:
I rephrased the title because there is nothing _unsafe_ per se about the implementation.

da-kami:
This ticket concerns adding more state to the `Cfd` and is thus related to #290

da-kami:
Tape-workaround: Add `Ack` messages to both actors at the end and only publish if ack was received

da-kami:
@itchymax @bonomat @thomaseizinger 

I'll see to add the ACK messages and discard the complete setup (on both maker and taker) side in case we did not get the ACK from the other party.

This is a temporary solution that could be exploited by a counterparty that removes sending the final `ACK`. No money can be "stolen", but the rogue counterparty could "hold the funds hostage". I think this is good enough for now. In our event-model refactor we already introduce error scenarios where we pass out the incomplete DLC for monitoring, so it should be easier to add a more complete fix.

da-kami:
Removed from Umbrel milestone becuase #724 was implemented as workaround. More proper fix is expected to be implemented as part of the refactoring towards event model. Keeping the ticket open until that goes in.

da-kami:
[...]

PR #441, opened 2021-10-27
Only change Cfd state when we succeed to contact Taker

klochowicz:
The situation outlined and fix #440 was repeated a few times in the code, there
were a few other places that preemptively changed the Cfd state before
taker was successfully contacted.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -684,6 +684,13 @@ where
 
         let mut conn = self.db.acquire().await?;
 
+        let txid = self
+            .wallet
+            .send(wallet::TryBroadcastTransaction {
+                tx: dlc.lock.0.clone(),
+            })
+            .await??;
+
         let mut cfd = load_cfd_by_order_id(order_id, &mut conn).await?;
         cfd.state = CfdState::PendingOpen {
             common: CfdStateCommon::default(),
@@ -692,13 +699,6 @@ where
         };
 
         append_cfd_state(&cfd, &mut conn, &self.cfd_feed_actor_inbox).await?;
-
-        let txid = self
-            .wallet
-            .send(wallet::TryBroadcastTransaction {
-                tx: dlc.lock.0.clone(),
-            })
-            .await??;
 
         tracing::info!("Lock transaction published with txid {}", txid);
 
@@ -749,8 +749,6 @@ where
             }
         };
 
-        let dlc = cfd.open_dlc().context("CFD was in wrong state")?;
-
         let oracle_event_id = oracle::next_announcement_after(
             time::OffsetDateTime::now_utc() + cfd.order.settlement_time_interval_hours,
         )?;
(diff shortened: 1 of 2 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -684,6 +684,13 @@ where
 
         let mut conn = self.db.acquire().await?;
 
+        let txid = self
```
da-kami, line 687: ❗️ We only retry to send the transaction if we transition to `PendingOpen`, otherwise the whole `Cfd` will be transitioned to `SetupFailed` upon restart at the moment!

So, if sending the transaction fails, we won't transition to `PendingOpen` and will never retry to send upon restart.
This could lead to a similar behaviour as described here: https://github.com/comit-network/hermes/issues/364

daemon/src/maker_cfd.rs
```diff
@@ -770,6 +768,8 @@ where
             })
             .await??;
 
+        let dlc = cfd.open_dlc().context("CFD was in wrong state")?;
```
da-kami, line 771: This change does not make a whole lot of difference; both statements can fail but don't change any state.
What was the motivation for changing this?

daemon/src/taker_cfd.rs
```diff
@@ -401,6 +401,13 @@ where
         tracing::info!("Setup complete, publishing on chain now");
 
         let mut conn = self.db.acquire().await?;
+        let txid = self
```
da-kami, line 404: ❗️same problem as described above.

da-kami (changes requested):

daemon/src/maker_cfd.rs
da-kami: Note: #364 describes how we can tackle this during failure in contract setup itself; the overall idea is to be mindful about what the other party knows, because we cannot ensure the other party won't publish if we ran into an error. 
For #440 this is irrelevant because it is prior to contract setup, i.e. no signatures (...) exchanged yet :)

da-kami:
Would be cool if we could cover these changes with failing tests. This goes towards writing tests that include the restart behavior - I think we can easily achieve that, but will need some more time until we have them. 
Simulating that sending the transaction failed should be possible. This, however, is already very high test coverage level. Let's see to reach that :)

klochowicz:
> Would be cool if we could cover these changes with failing tests. This goes towards writing tests that include the restart behavior - I think we can easily achieve that, but will need some more time until we have them. Simulating that sending the transaction failed should be possible. This, however, is already very high test coverage level. Let's see to reach that :)

great idea 🏆

daemon/src/maker_cfd.rs
klochowicz: Thanks for the explanation - I have indeed missed the whole point that this particular cfd state is there for the purpose of handling this problem!

Armed with the new knowledge, I've closed the PR as it made no sense anymore :)
🐣

Issue #352, opened 2021-10-18
Do not track timestamps with nanosecond precision


DeliciousHair:
Trivial change to DB structure to remove ns from timestamp + delay data but requires checking that this does not cause conflicts with `SystemTime::now()` instances elsewhere in the codebase.

thomaseizinger:
The solution for this is likely to have our own `Timestamp` type that has only a precision of seconds.

DeliciousHair:
resolved

PR #442, opened 2021-10-27
Nanoseconds begone!

DeliciousHair:
Created new Timestamp struct that only uses seconds (as i64 in order to play nice with both `sqlx` and `chrono`) and removed use of `SytemTime::now()` throughout in the process.

This PR addresses #352 but also had the effect of doing a better job of addressing #434, making #435 pointless.

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -1,15 +1,16 @@
 use crate::olivia;
-
 use anyhow::{Context, Result};
 use bdk::bitcoin::{Address, Amount, Denomination};
+use chrono::DateTime;
 use reqwest::Url;
 use rust_decimal::prelude::ToPrimitive;
 use rust_decimal::Decimal;
 use serde::{Deserialize, Serialize};
 use serde_with::{DeserializeFromStr, SerializeDisplay};
+use std::convert::TryInto;
 use std::num::NonZeroU8;
 use std::ops::{Add, Div, Mul, Sub};
-use std::time::SystemTime;
+use std::time::{SystemTime, UNIX_EPOCH};
 use std::{fmt, str};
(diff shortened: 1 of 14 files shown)


daemon/src/model.rs
```diff
@@ -1,15 +1,16 @@
 use crate::olivia;
-
 use anyhow::{Context, Result};
 use bdk::bitcoin::{Address, Amount, Denomination};
+use chrono::prelude::*;
```
da-kami: 🛠 Please use specific imports and not `*` - or is this necessary?

daemon/src/model.rs
```diff
@@ -546,6 +547,43 @@ impl str::FromStr for BitMexPriceEventId {
     }
 }
 
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, sqlx::Type)]
+pub struct Timestamp(i64);
```
da-kami, line 551: ❓❗️ Wouldn't it be safer to keep the internals as `u64` and do the conversion from `i64` to `u64` in `parse_from_rfc3339`?
I don't really see the advantage of storing it as `i64`. I find `u64` more logical, because it should never be signed unless time goes backwards.

da-kami (approved):
thanks! would be great if you could remove the `*` import before the merge.
the other comment can also be tackled in a follow up.

daemon/src/model.rs
DeliciousHair: As mentioned in the PR note, this was done intentionally in order to avoid additional coercion in reading from the DB (`sqlx` does not support `u64` via `sqlx::Type` for `sqlite`) as well as wrapping the `parse_from_rfc3339()` method from `chrono`. I see your argument, but it actually leads to more fragility and / or faffing about hence going with `i64`.

PR #445, opened 2021-10-28
Encrypt network communication using the noise protocol

rishflab:
https://github.com/comit-network/hermes/pull/429 renamed the branch to accurately reflect what the PR was doing so I had to reopen the PR

moved the handshake implementation details into the noise module

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -1,12 +1,13 @@
 use crate::maker_cfd::{FromTaker, NewTakerOnline};
 use crate::model::cfd::{Order, OrderId};
 use crate::model::{BitMexPriceEventId, TakerId};
-use crate::{forward_only_ok, maker_cfd, send_to_socket, wire};
+use crate::{forward_only_ok, maker_cfd, noise, send_to_socket, wire};
 use anyhow::{Context as AnyhowContext, Result};
 use futures::{StreamExt, TryStreamExt};
 use std::collections::HashMap;
 use std::io;
 use std::net::SocketAddr;
+use std::sync::{Arc, Mutex};
 use tokio::net::TcpStream;
 use tokio_util::codec::FramedRead;
 use xtra::prelude::*;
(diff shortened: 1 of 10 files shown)


daemon/src/noise.rs
```diff
@@ -0,0 +1,72 @@
+use anyhow::Result;
+use snow::{Builder, TransportState};
+use std::io;
+use tokio::io::{AsyncReadExt, AsyncWriteExt};
+use tokio::net::TcpStream;
+
+// todo: investigate what these params do and whether we can use get authentication for free using
```
klochowicz: nit: prefer the uppercase version (`TODO`), as it's being parsed by many IDEs and thus easier to find later :)

daemon/src/noise.rs
```diff
@@ -0,0 +1,72 @@
...
+// them
+static PARAMS: &str = "Noise_XX_25519_ChaChaPoly_BLAKE2s";
+
+pub async fn initiator_handshake(connection: &mut TcpStream) -> Result<TransportState> {
+    let builder: Builder<'_> = Builder::new(PARAMS.parse()?);
+    let static_key = builder.generate_keypair()?.private;
```
klochowicz, line 13: do we lose anything by removing the PSK part with the secret? if it was useful, it might as well stay!

daemon/src/maker_inc_connections.rs
```diff
@@ -210,7 +215,11 @@ impl Actor {
     async fn handle(&mut self, msg: ListenerMessage, ctx: &mut Context<Self>) -> KeepRunning {
         match msg {
             ListenerMessage::NewConnection { stream, address } => {
-                self.handle_new_connection_impl(stream, address, ctx).await;
+                self.handle_new_connection_impl(stream, address, ctx)
+                    .await
+                    .expect(
```
klochowicz: IMHO this should be just a tracing::warn. We don't want a non-compliant taker to take down the whole maker - it's enough to not proceed with connecting it

daemon/src/noise.rs
klochowicz: another nit: call it `NOISE_PARAMS` for consistency with your commit message and for readability.

klochowicz (approved):
thanks, looking good! one blocker (`unwrap()`)

daemon/src/noise.rs
rishflab: Since it isnt really secret between the two parties, I dont think so

PR #451, opened 2021-10-29
Mock all the actors

klochowicz:
A POC of how things could be done:
- one more layer of indirection (a trait per actor type: Wallet, Oracle, Monitor)
- Mocks implementing the traits (with default stubbed implementations if no extra
  behaviour needed)
- references to the mocks are being passed into the tests (via Arc<Mutex>>), allowing
  for dynamically changing the behaviour and additional assertion. This also
    aids readability, as the mock setup can be collocated with a particular
    test, if the test needs something extra

Note: work-in-progress, early feedback & discussions welcome :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -640,6 +640,8 @@ dependencies = [
  "hex",
  "hkdf",
  "itertools",
+ "mockall",
+ "mockall_derive",
  "nalgebra",
  "ndarray",
  "ndarray_einsum_beta",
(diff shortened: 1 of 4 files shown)


klochowicz:
so far only MockWallet is pulled through to the tests, new ones will come soon too.

klochowicz:
superceded by #458 (which rebases on top of the more complicated test & the new structure of the test harness)

PR #458, opened 2021-11-01
Roll-out usage of a mocking framework in actor tests

klochowicz:
Mockall is a mocking framework that removes the need for writing more actors,
making tests easier to write.

Summary:
- add one more layer of indirection (a trait per actor type: Wallet, Oracle, Monitor)
- Mocks implementing the actor traits (with default stubbed implementations if no extra
  behaviour needed)
- references to the mocks are being passed into the tests (via Arc<Mutex>>), allowing
  for dynamically changing the behaviour and adding assertions. This also
    aids readability, as the mock setup can be collocated with a particular
    test, if the test needs something extra

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -1,6 +1,6 @@
-use crate::harness::mocks::monitor::Monitor;
-use crate::harness::mocks::oracle::Oracle;
-use crate::harness::mocks::wallet::Wallet;
+use crate::harness::bdk::dummy_partially_signed_transaction;
+use crate::harness::mocks::oracle::dummy_announcement;
+use crate::harness::mocks::wallet::build_party_params;
 use crate::harness::start_both;
 use anyhow::Context;
 use cfd_protocol::secp256k1_zkp::schnorrsig;
@@ -8,9 +8,10 @@ use daemon::maker_cfd;
 use daemon::model::cfd::{Cfd, CfdState, Order, Origin};
 use daemon::model::{Price, Usd};
 use daemon::tokio_ext::FutureExt;
+use harness::bdk::dummy_tx_id;
 use rust_decimal_macros::dec;
 use std::time::Duration;
-use tokio::sync::watch;
+use tokio::sync::{watch, MutexGuard};
 use tracing::subscriber::DefaultGuard;
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::util::SubscriberInitExt;
@@ -20,15 +21,7 @@ mod harness;
 #[tokio::test]
 async fn taker_receives_order_from_maker_on_publication() {
     let _guard = init_tracing();
-    let (mut maker, mut taker) = start_both(
-        Oracle::default(),
-        Monitor,
-        Wallet::default(),
-        Oracle::default(),
-        Monitor,
-        Wallet::default(),
-    )
-    .await;
+    let (mut maker, mut taker) = start_both().await;
 
     assert!(is_next_none(&mut taker.order_feed).await);
 
@@ -45,15 +38,7 @@ async fn taker_receives_order_from_maker_on_publication() {
 #[tokio::test]
 async fn taker_takes_order_and_maker_rejects() {
     let _guard = init_tracing();
-    let (mut maker, mut taker) = start_both(
-        Oracle::default(),
-        Monitor,
-        Wallet::default().with_wallet_info(),
-        Oracle::default(),
-        Monitor,
-        Wallet::default().with_wallet_info(),
-    )
-    .await;
+    let (mut maker, mut taker) = start_both().await;
 
     // TODO: Why is this needed? For the cfd stream it is not needed
     is_next_none(&mut taker.order_feed).await;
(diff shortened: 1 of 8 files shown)


daemon/tests/happy_path.rs
```diff
@@ -86,28 +71,34 @@ async fn taker_takes_order_and_maker_rejects() {
...
-            .with_psbt(harness::bdk::dummy_partially_signed_transaction())
-            .with_txid(harness::bdk::dummy_tx_id()),
-    )
-    .await;
+    let (mut maker, mut taker) = start_both().await;
+
+    mock_wallet_transaction(&mut maker.mocks.wallet().await);
+    mock_wallet_transaction(&mut taker.mocks.wallet().await);
+
+    maker.mocks.monitor().await.expect_sync().return_const(());
```
da-kami: I'm afraid this might cause brittle tests, because the sync depends on time intervals. If a test runs longer on e.g. CI we cannot guarantee we don't run into another sync interval; additionally, if we change the sync interval tests might break even though they should not be affected (we don't test the sync, we test other behavior...).

Would it be possible to just let all `sync` messages be ignored in the tests? i.e. is it possible to instruct the mocks to just not assert on `sync` messages? 
We can have specific tests that test sync sanity, but for other tests I don't think we should care about it. 
This also accounts for the oracle's sync.

daemon/tests/happy_path.rs
```diff
@@ -86,28 +71,34 @@ async fn taker_takes_order_and_maker_rejects() {
...
-        Wallet::default()
-            .with_wallet_info()
-            .with_party_params()
-            .with_psbt(harness::bdk::dummy_partially_signed_transaction())
-            .with_txid(harness::bdk::dummy_tx_id()),
-    )
-    .await;
+    let (mut maker, mut taker) = start_both().await;
+
+    mock_wallet_transaction(&mut maker.mocks.wallet().await);
```
da-kami: I am not sure this should be done in the beginning;  do you follow some kind of guide here? 
I would have always instructed what is expected directly before we need it, so I would move this code together with the expected oracle event right before the maker accepts.

daemon/tests/happy_path.rs
```diff
@@ -86,28 +71,34 @@ async fn taker_takes_order_and_maker_rejects() {
     assert!(matches!(maker_cfd.state, CfdState::Rejected { .. }));
 }
 
+// Helper function setting up a "happy path" wallet mock
+fn mock_wallet_transaction(wallet: &mut MutexGuard<'_, harness::mocks::wallet::MockWallet>) {
+    #[allow(clippy::redundant_closure)] // clippy is in the wrong here
+    wallet
```
da-kami: I am not sure this helper function is a good thing. We can do this, but I don't think `build_party_params` should be part of it - build party param is specific to the contract setup. I think combining `sign` and `txid` makes sense, because we sign and send a transaction for many cases (i.e. commit, ...).

Maybe we can keep this code explicitly in the test and extract helper functions a bit later, once we have a better idea of what makes sense to combine.

daemon/tests/harness/mocks/monitor.rs
```diff
@@ -1,29 +1,53 @@
...
-        todo!("stub this if needed")
+        self.mock.lock().await.oracle_attestation()
     }
 }
 
-impl Default for Monitor {
-    fn default() -> Self {
-        Monitor
+#[automock]
+pub trait Monitor {
```
da-kami, line 37: I think it would be nice if we can route all the messages through by default. I think the mock should always reflect the signature of the `handle` function.

daemon/tests/harness/mocks/wallet.rs
```diff
@@ -1,93 +1,81 @@
...
-        Wallet {
-            party_params: false,
-            wallet_info: None,
-            psbt: None,
-            txid: None,
-        }
+#[automock]
+pub trait Wallet {
+    fn build_party_params(&mut self, _msg: wallet::BuildPartyParams) -> Result<PartyParams> {
+        todo!()
```
da-kami:
```suggestion
        unreachable!("Generated code that will be filled in the test")
```


I find `todo!` misleading here - does `mockall` enforce that? I think`unreachable!` would be easier to understand. 
Would be nice if we can add a message so it becomes more clear what this is.

daemon/tests/harness/mod.rs
```diff
@@ -18,54 +18,42 @@ pub mod bdk;
...
-    ) -> Self {
+    pub async fn start(oracle_pk: schnorrsig::PublicKey) -> Self {
         let db = in_memory_db().await;
 
+        let mocks = mocks::Mocks::default();
+
+        let (oracle, monitor, wallet) = mocks::create_actors(&mocks);
+
+        // Sync method need to be mocked before the actors start
+        mocks.oracle.lock().await.expect_sync().return_const(());
```
da-kami, line 55: Oh my, that is a bit weird, but I see why it is needed (because we trigger the initial sync upon startup).

da-kami (approved):
Thanks! I like how the mocks tie into the tests without too much intrusion. 
Nothing blocking per se, but several comments that might be interesting to action before merging or in follow-ups. Looking forward to your replies :)

daemon/tests/happy_path.rs
klochowicz: happy to inline this helper, I just didn't want to make this test unnecessarily verbose in my opinion. It doesn't make any difference in the behaviour.

daemon/tests/harness/mocks/monitor.rs
klochowicz: I'll follow YAGNI here - let's add it when (or perhaps *if*) we need it. In this case, it would be the time someone wants to write a mock using it.

daemon/tests/harness/mod.rs
klochowicz: yeah, at least we now are explicit about it! previously we did not know when that handler got invoked :p

daemon/tests/harness/mocks/wallet.rs
klochowicz: good idea, thanks!

daemon/tests/happy_path.rs
klochowicz: from the point of the mock, it does not matter. it can be instructed arbitrarily early (it's "when this method gets called, return the following value ").

Happy to move it if it makes it more readable for others (arguably we're move specific then that we never get these calls before we expect them).

daemon/tests/happy_path.rs
klochowicz: I'm having trouble understanding the brittleness - the way it stands, I just say to return `(())` whenever the `Sync` is called (note that there's no `times()` condition). It preserves the behaviour that we had before introduction of the mock 1:1.

the mocks don't assert anything in the tests so far, and will never count how many times `sync` has been called as we don't use the `times()` here.

daemon/tests/happy_path.rs
klochowicz: ended up removing build_party_params() from this function.

I also added a sequence, that means that these handlers have to be invoked in that order.

daemon/tests/happy_path.rs
klochowicz: moved these around: build params is just before the contract setup, and sign & broadcast mock after contract setup.

daemon/tests/harness/mocks/monitor.rs
klochowicz: ah, it was so simple, I just implemented it after all :-)

daemon/tests/happy_path.rs
da-kami: Oh, then I probably just misunderstood that - that is what we want then. Could still opt for configuring the monitor sync in the setup as well if we want.

klochowicz: sure, I'll move it into the setup, so it doesn't get in the way.

da-kami (approved):

PR #459, opened 2021-11-01
Authentication

### LINKED ISSUE #229: Introduce authentication and encryption for the network layer

rishflab:
#229

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,10 +76,13 @@ jobs:
       - name: Smoke test ${{ matrix.os }} binary
         shell: bash
         run: |
-          target/debug/maker --data-dir=/tmp/maker --generate-seed testnet &
-          sleep 10s # Wait for maker to start
+          mkdir -p /tmp/maker/testnet
+          cp -r daemon/util/testnet_seeds/maker_seed /tmp/maker/testnet/maker_seed
+          target/debug/maker --data-dir=/tmp/maker testnet &
+          sleep 10s # Wait for maker to start\
 
-          target/debug/taker --data-dir=/tmp/taker --generate-seed testnet &
+          # The maker-id is generated from the makers seed found in daemon/util/testnet_seeds/maker_seed
+          target/debug/taker --data-dir=/tmp/taker --generate-seed --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e testnet &
           sleep 10s # Wait for taker to start
 
           curl --fail http://localhost:8000/api/alive
--- a/daemon/src/noise.rs
+++ b/daemon/src/noise.rs
@@ -6,15 +6,19 @@ use tokio::net::TcpStream;
 
 pub static NOISE_MAX_MSG_LEN: u32 = 65535;
 pub static NOISE_TAG_LEN: u32 = 16;
-// TODO: investigate what these params do and whether we can use get authentication for free using
-// them
-static NOISE_PARAMS: &str = "Noise_XX_25519_ChaChaPoly_BLAKE2s";
+static NOISE_PARAMS: &str = "Noise_IK_25519_ChaChaPoly_BLAKE2s";
 
-pub async fn initiator_handshake(connection: &mut TcpStream) -> Result<TransportState> {
+pub async fn initiator_handshake(
+    connection: &mut TcpStream,
+    local_priv_key: &x25519_dalek::StaticSecret,
+    remote_pub_key: &x25519_dalek::PublicKey,
+) -> Result<TransportState> {
     let builder: Builder<'_> = Builder::new(NOISE_PARAMS.parse()?);
-    let static_key = builder.generate_keypair()?.private;
 
-    let mut noise = builder.local_private_key(&static_key).build_initiator()?;
+    let mut noise = builder
+        .local_private_key(&local_priv_key.to_bytes())
+        .remote_public_key(&remote_pub_key.to_bytes())
+        .build_initiator()?;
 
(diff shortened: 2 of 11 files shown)


start_all.sh
```diff
@@ -4,4 +4,4 @@ export RUST_BACKTRACE=1
 
 # A simple command to spin up the complete package, ie. both daemons and frontends.
 # A single 'ctrl+c' stops all processes.
-(trap 'kill 0' SIGINT; cargo run --bin maker -- testnet & cargo run --bin taker -- testnet & APP=maker yarn --cwd=./frontend dev  & APP=taker yarn --cwd=./frontend dev)
+(trap 'kill 0' SIGINT; cargo run --bin maker -- testnet & cargo run --bin taker -- --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e testnet -- & APP=maker yarn --cwd=./frontend dev  & APP=taker yarn --cwd=./frontend dev)
```
klochowicz: will this magic number (`maker_id`) ever change? Perhaps add a comment about it (either a reassuring one that it's set in stone, or how/when it needs to be updated)

daemon/src/noise.rs
```diff
@@ -6,15 +6,19 @@ use tokio::net::TcpStream;
...
-// TODO: investigate what these params do and whether we can use get authentication for free using
-// them
-static NOISE_PARAMS: &str = "Noise_XX_25519_ChaChaPoly_BLAKE2s";
+static NOISE_PARAMS: &str = "Noise_IK_25519_ChaChaPoly_BLAKE2s";
 
-pub async fn initiator_handshake(connection: &mut TcpStream) -> Result<TransportState> {
+pub async fn initiator_handshake(
+    connection: &mut TcpStream,
+    local_priv_key: x25519_dalek::StaticSecret,
+    remote_pub_key: x25519_dalek::PublicKey,
```
klochowicz: nit: do you need to own these? ref would probably be enough.

daemon/src/noise.rs
```diff
@@ -23,21 +27,22 @@ pub async fn initiator_handshake(connection: &mut TcpStream) -> Result<Transport
...
 
     tracing::debug!("Noise protocol initiator handshake is complete");
 
     Ok(noise)
 }
 
-pub async fn responder_handshake(connection: &mut TcpStream) -> Result<TransportState> {
+pub async fn responder_handshake(
+    connection: &mut TcpStream,
+    local_priv_key: x25519_dalek::StaticSecret,
```
klochowicz: nit: do you need to own this? (you could probably avoid a `clone()`)

.github/workflows/ci.yml
```diff
@@ -76,10 +76,12 @@ jobs:
...
         shell: bash
         run: |
-          target/debug/maker --data-dir=/tmp/maker --generate-seed testnet &
+          cp daemon/util/testnet_seeds/maker_seed /tmp/maker/maker_seed
+          target/debug/maker --data-dir=/tmp/maker testnet &
           sleep 10s # Wait for maker to start
 
-          target/debug/taker --data-dir=/tmp/taker --generate-seed testnet &
+          cp daemon/util/testnet_seeds/taker_seed /tmp/taker/taker_seed:
+          target/debug/taker --data-dir=/tmp/taker --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e testnet &
```
klochowicz: will we need to manually upgrade this key, increasing the maintenance burden?

rishflab: if we change the seeds in the repo, yes

daemon/src/noise.rs
rishflab: yep, ref is enough

daemon/src/taker.rs
```diff
@@ -51,6 +55,12 @@ struct Opts {
     network: Network,
 }
 
+fn parse_x25519_pubkey(s: &str) -> Result<x25519_dalek::PublicKey> {
+    let mut bytes = [0u8; 32];
+    hex::decode_to_slice(s, &mut bytes)?;
+    Ok(x25519_dalek::PublicKey::from(bytes))
```
klochowicz, line 61: nit: `bytes.into()` ?

daemon/src/seed.rs
```diff
@@ -72,6 +72,16 @@ impl Seed {
...
     }
+
+    pub fn derive_noise_static_secret(&self) -> x25519_dalek::StaticSecret {
+        let mut secret = [0u8; 32];
+
+        Hkdf::<Sha256>::new(None, &self.0)
+            .expand(b"NOISE_STATIC_SECRET", &mut secret)
+            .expect("okm array is of correct length");
+
+        x25519_dalek::StaticSecret::from(secret)
```
klochowicz, line 83: nit: `secret.into()`?

klochowicz (approved):
looks good!

PR #462, opened 2021-11-01
Add `withdraw` subcommand to taker and maker

da-kami:
Withdraw is an optional subcommand of the network subcommand, so the command reads like this, e.g. taker:
`./taker mainnet withdraw --address ...`

Internally, we use the wallet actor for withdrawing for now (simplifies the implementation, otherwise we would have to extract the wallet construction outside the actor).

ToDo inside the wallet actor:

- [x] wallet drain if no amount is given
- [x] construct transaction based on amount and address
- [x] broadcast transaction and return `Txid`

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -1,7 +1,7 @@
 use anyhow::{Context, Result};
-use bdk::bitcoin;
 use bdk::bitcoin::secp256k1::schnorrsig;
-use clap::Parser;
+use bdk::{bitcoin, FeeRate};
+use clap::{Parser, Subcommand};
 use daemon::auth::{self, MAKER_USERNAME};
 use daemon::db::{self};
 
@@ -20,6 +20,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
 
+use bdk::bitcoin::Amount;
 use std::task::Poll;
 use tokio::sync::watch;
 use tracing_subscriber::filter::LevelFilter;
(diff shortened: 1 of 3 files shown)


da-kami:
@bonomat in xmr-btc we estimated the fee from a block-window using the Electrum client. Will we need a similar strategy here? If so this might become a bit more complicated, because our wallet current does not include the Electrum client itself. It can be added, but I wonder if we can do fee estimation directly from the wallet - any ideas?

https://github.com/comit-network/xmr-btc-swap/blob/bf9e2d0f777e17d75a6e5ceae3ad39f6edbb1a3c/swap/src/bitcoin/wallet.rs#L402-L409

da-kami:
> @bonomat in xmr-btc we estimated the fee from a block-window using the Electrum client. Will we need a similar strategy here? If so this might become a bit more complicated, because our wallet current does not include the Electrum client itself. It can be added, but I wonder if we can do fee estimation directly from the wallet - any ideas?
> 
> https://github.com/comit-network/xmr-btc-swap/blob/bf9e2d0f777e17d75a6e5ceae3ad39f6edbb1a3c/swap/src/bitcoin/wallet.rs#L402-L409

Had some more thoughts, probably the best choice is keeping the withdraw wallet separate for now. That was the original plan. That way the wallet used for withdraw can have additional features that the wallet in the actor does not have. I don't think we can work around the fee estimation without the electrum client.

bonomat:
> @bonomat in xmr-btc we estimated the fee from a block-window using the Electrum client. Will we need a similar strategy here? If so this might become a bit more complicated, because our wallet current does not include the Electrum client itself. It can be added, but I wonder if we can do fee estimation directly from the wallet - any ideas?
> 
> [comit-network/xmr-btc-swap@`bf9e2d0`/swap/src/bitcoin/wallet.rs#L402-L409](https://github.com/comit-network/xmr-btc-swap/blob/bf9e2d0f777e17d75a6e5ceae3ad39f6edbb1a3c/swap/src/bitcoin/wallet.rs#L402-L409)

I'll respond in the code for this.

bonomat:
> @bonomat in xmr-btc we estimated the fee from a block-window using the Electrum client. Will we need a similar strategy here? If so this might become a bit more complicated, because our wallet current does not include the Electrum client itself. It can be added, but I wonder if we can do fee estimation directly from the wallet - any ideas?
> 
> [comit-network/xmr-btc-swap@`bf9e2d0`/swap/src/bitcoin/wallet.rs#L402-L409](https://github.com/comit-network/xmr-btc-swap/blob/bf9e2d0f777e17d75a6e5ceae3ad39f6edbb1a3c/swap/src/bitcoin/wallet.rs#L402-L409)

Withdrawing is a simple strategy and I think we can get away without access to electrum client. 
There are two strategies we can follow: 
1. CPFP (Child pays for parent): this strategy is always available and means that a follow-up transaction can spend a higher transaction fee to bump the withdraw transaction. 
2. RBF (Replace by fee): by setting the seqencenumber to a particular value you signal that you would want to replace this transaction with one with a higher transaction fee. The miners have an incentive to pick the transaction with the highest fee to maximize their return. 

Luckily it is not an or but an and and we can use both strategies: 
When having access to the `wallet` you can use the transaction builder like this: 

```rust
	
 let sats = <optional_parameter>.unwrap_or(1.0)
 tx_builder
     .add_recipient(to_address.script_pubkey(), 50_000)
     .fee_rate(FeeRate::from_sat_per_vb(sats))
     // Turn on RBF signaling
     .enable_rbf();
```

Now the transaction is very cheap (1 sat/vbyte) by default. If the user wants, he could provide a second argument (`optional_parameter`) to bump that transaction later on. 


I hope that helps.

da-kami:
Note: I tested it by withdrawing a specific amount on testnet from maker to taker. I did not test the fee-rate, but expect that to work as long as one enters a sane value. I did not test draining the wallet yet but suppose that it should work, given that most of the code is borrowed from the xmr-btc project. 

Note: I thought about adding an actor test, but given that we would use a dummy wallet such a test would not really cover any logic, so I did not add it.

klochowicz (approved):
LGTM!

daemon/src/maker.rs
```diff
@@ -91,13 +91,17 @@ enum Network {
...
+#[derive(Subcommand)]
 enum Withdraw {
     Withdraw {
         /// Optionally specify the amount of Bitcoin to be withdrawn. If not specified the wallet
-        /// will be drained.
+        /// will be drained. Amount is to be specified with denomination, e.g. "0.1 BTC"
         #[clap(long)]
         amount: Option<Amount>,
+        /// Optionally specify the fee-rate for the transaction. The fee-rate is specified as btc
+        /// per kvb, e.g. 0.00001
```
bonomat: what's `kvb` ? :D 

Can we do `sats_per_vbyte` instead?
Maybe add an explanation to what it is in BTC if you don't want to allow the user to provide it in sats?

bonomat (approved):
Looks good. Just one thing on the fee.

da-kami:
I just gave it a try with draining the wallet and it just keeps being stuck. It does not print an error though, just keeps being stuck. Can draining take significant time? (i.e. more than 10 mins?)

da-kami:
Note: Came around and tested draining as well. Ran into an issue where the wallet is not properly synced before we withdraw. Re-arranged code to make sure we sync the wallet actor before withdraw.

PR #476, opened 2021-11-04
Display connection errors to the daemon in the UI

### LINKED ISSUE #460: Provide feedback if the frontend does not have connection to the daemon
If the daemon is not running, we should  indicate that the connection got lost.

klochowicz:
Periodically send a keepalive message; if the daemon does not come back with a
response, display a persistent toast notification (at the top of the screen, so
it's harder to miss or confuse).

Notification goes away when the connection to the daemon is re-established.

Fixes #460

### DIFF
--- a/frontend/src/MakerApp.tsx
+++ b/frontend/src/MakerApp.tsx
@@ -19,6 +19,7 @@ import {
 import React, { useEffect, useState } from "react";
 import { useAsync } from "react-async";
 import { useEventSource } from "react-sse-hooks";
+import { useBackendMonitor } from "./components/BackendMonitor";
 import { CfdTable } from "./components/cfdtables/CfdTable";
 import CurrencyInputField from "./components/CurrencyInputField";
 import CurrentPrice from "./components/CurrentPrice";
@@ -43,6 +44,7 @@ export default function App() {
     const priceInfo = useLatestEvent<PriceInfo>(source, "quote");
 
     const toast = useToast();
+    useBackendMonitor(toast, 5000); // 5s timeout
 
     let [minQuantity, setMinQuantity] = useState<string>("10");
     let [maxQuantity, setMaxQuantity] = useState<string>("100");
--- a/frontend/src/TakerApp.tsx
+++ b/frontend/src/TakerApp.tsx
@@ -18,6 +18,7 @@ import {
 import React, { useEffect, useState } from "react";
 import { useAsync } from "react-async";
 import { useEventSource } from "react-sse-hooks";
+import { useBackendMonitor } from "./components/BackendMonitor";
 import { CfdTable } from "./components/cfdtables/CfdTable";
 import CurrencyInputField from "./components/CurrencyInputField";
 import createErrorToast from "./components/ErrorToast";
@@ -63,6 +64,7 @@ async function getMargin(payload: MarginRequestPayload): Promise<MarginResponse>
 
 export default function App() {
     const toast = useToast();
+    useBackendMonitor(toast, 5000); // 5s timeout
 
     document.title = "Hermes Taker";
 
--- a/frontend/src/components/BackendMonitor.tsx
+++ b/frontend/src/components/BackendMonitor.tsx
@@ -1,0 +1,50 @@
+import { useEffect, useRef } from "react";
+
+async function fetchWithTimeout(resource: any, timeout: number) {
+    const controller = new AbortController();
+    const id = setTimeout(() => controller.abort(), timeout);
+    const response = await fetch(resource, {
+        signal: controller.signal,
+    });
+    clearTimeout(id);
+    return response;
+}
+
+// Check for backend's presence by sending a request to '/alive' endpoint.
+// When the backend's not there, the request is likely to timeout, triggering a
+// persistent toast notification that goes away when the daemon is back online.
+export function useBackendMonitor(toast: any, timeout_ms: number): void {
+    const toastIdRef = useRef();
+
+    const checkForBackend: () => void = async () => {
+        try {
+            const res = await fetchWithTimeout("/alive", timeout_ms);
+            // In case we get a response, but it's not what we expected
+            if (res.status.toString().startsWith("2")) {
+                if (toastIdRef.current) {
+                    toast.close(toastIdRef.current);
+                }
+            } else {
+                throw new Error(res.statusText);
+            }
+        } catch (e) {
+            if (!toastIdRef.current) {
+                toastIdRef.current = toast({
+                    title: "Connection Error",
+                    description: "Daemon process is not running",
+                    status: "error",
+                    position: "top",
+                    duration: timeout_ms * 100000, // we don't want this to be closed
+                    isClosable: false,
+                });
+            }
+        }
+    };
+
+    useEffect(() => {
+        const interval = setInterval(() => checkForBackend(), timeout_ms);
+        return () => {
+            clearInterval(interval);
+        };
+    });
+}
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -33,6 +33,7 @@ export default defineConfig({
     server: {
         proxy: {
             "/api": `http://localhost:${backendPorts[app]}`,
+            "/alive": `http://localhost:${backendPorts[app]}`,
         },
     },
 });


da-kami (commented):
This is a nice workaround, but I think the more proper solution would be a `keep-alive` on the event stream.
Technically the http backend's presence is independent of the stream connection.
f the user triggers an action (i.e. request) against the client the user will see an error if the request fails. The problem is more if the frontend is no connected to the stream anymore and does not receive updates.

Here is the solution that I had in my mind:

1. Add `keep-alive` event to the event stream that includes the current timestamp
2. Store the `keep-alive` in the frontend (always overwrite with latest)
3. Add `useInterval` to periodically check if the stored `keep-alive` is up to date.

If it is *not* up to date: 

1. try to reconnect to the stream
2. If that fails the backend must be gone, error.

I think this way is more resilient, and does not require requests from client to server, but just depends on the stream.  
Most public event streams have such kind of keep alive events for exactly that purpose. 
Let me know what you think

klochowicz:
> This is a nice workaround, but I think the more proper solution would be a `keep-alive` on the event stream. Technically the http backend's presence is independent of the stream connection...
> 
> Here is the solution that I had in my mind:
> 
>     1. Add `keep-alive` event to the event stream that includes the current timestamp
> 
>     2. Store the `keep-alive` in the frontend (always overwrite with latest)
> 
>     3. Add `useInterval` to periodically check if the stored `keep-alive` is up to date.
> 
> 
> If it is _not_ up to date:
> 
>     1. try to reconnect to the stream
> 
>     2. If that fails the backend must be gone, error.
> 
> 
> I think this way is more resilient, and does not require requests from client to server, but just depends on the stream. Most public event streams have such kind of keep alive events for exactly that purpose. Let me know what you think

would you mind explaining why it would be more resilient this way? We already expose the '`/alive` call anyway. IMHO it's simpler and works just as good (not to mention that we can use it anywhere, not just in this UI).

da-kami:
> > This is a nice workaround, but I think the more proper solution would be a `keep-alive` on the event stream. Technically the http backend's presence is independent of the stream connection...
> > Here is the solution that I had in my mind:
> > ```
> > 1. Add `keep-alive` event to the event stream that includes the current timestamp
> > 
> > 2. Store the `keep-alive` in the frontend (always overwrite with latest)
> > 
> > 3. Add `useInterval` to periodically check if the stored `keep-alive` is up to date.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If it is _not_ up to date:
> > ```
> > 1. try to reconnect to the stream
> > 
> > 2. If that fails the backend must be gone, error.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this way is more resilient, and does not require requests from client to server, but just depends on the stream. Most public event streams have such kind of keep alive events for exactly that purpose. Let me know what you think
> 
> would you mind explaining why it would be more resilient this way? We already expose the '`/alive` call anyway. IMHO it's simpler and works just as good (not to mention that we can use it anywhere, not just in this UI).

Not sure how much I can argue for resilience. It is mostly a concern issue - the stream should make sure that the consumer of the stream can be sure the stream still works. The keep alive endpoints are used for a different purpose. 
I think ensuring the keep-alive on the stream is less invasive - not in the sense of "what already exists", but who's concern it is to deal with this. The stream becomes a more complete, independent component then.

We already have the stream, with your solution we send unnecessary requests from frontend to backend.

To come back to the resilience: I think it *can* theoretically happen that you still reach the keep alive endpoint, but are disconnected from the event stream. So, the keep alive endpoint might just not always solve the problem you are trying to solve. 
Under the hood this is different technology, one being http endpoints, the other server side events which I think use websockets under the hood. I would have to dig deeper to have a proper technology discussion on that though ;)

klochowicz:
I agree that the solution would be nicer as it removes the need for the requests. I'm happy to come back and optimize it as a separate PR when I finish all my tasks scheduled for the current iteration ;) 

For now, it's an improvement, as we don't detect lack of connectivity at all (not to mention that the component would be readily useful in any project that does not use SSE events).

I think with public endpoints/streams the biggest problem is scalability - it's not an issue here, as we have exactly 1 backend and 1 frontend.

da-kami:
> I agree that the solution would be nicer as it removes the need for the requests. I'm happy to come back and optimize it as a separate PR when I finish all my tasks scheduled for the current iteration ;)
> 
> For now, it's an improvement, as we don't detect lack of connectivity at all (not to mention that the component would be readily useful in any project that does not use SSE events).
> 
> I think with public endpoints/streams the biggest problem is scalability - it's not an issue here, as we have exactly 1 backend and 1 frontend.

I see the argument of this already being an improvement and the possibility to put it on the stream later.

Could we go one step further in this PR and add re-trying to connect to the stream in case we detect that the keep-alive endpoint is not available? Can this easily be triggered with the infrastructure we have currently in place?

klochowicz:
> > I agree that the solution would be nicer as it removes the need for the requests. I'm happy to come back and optimize it as a separate PR when I finish all my tasks scheduled for the current iteration ;)
> > For now, it's an improvement, as we don't detect lack of connectivity at all (not to mention that the component would be readily useful in any project that does not use SSE events).
> > I think with public endpoints/streams the biggest problem is scalability - it's not an issue here, as we have exactly 1 backend and 1 frontend.
> 
> I see the argument of this already being an improvement and the possibility to put it on the stream later.
> 
> Could we go one step further in this PR and add re-trying to connect to the stream in case we detect that the keep-alive endpoint is not available? Can this easily be triggered with the infrastructure we have currently in place?

That's not easy :( it will require more invasive change into the frontend, whilst I tried to keep this `BackendCheck` as a standalone component that can be used in any prototype we create. 

Moving forward, I believe we don't even need to define any custom keepalive message, SSE events should be able to handle sending a heartbeat message and detecting the loss. I will raise a ticket to cover this improvement.

da-kami:
> > > I agree that the solution would be nicer as it removes the need for the requests. I'm happy to come back and optimize it as a separate PR when I finish all my tasks scheduled for the current iteration ;)
> > > For now, it's an improvement, as we don't detect lack of connectivity at all (not to mention that the component would be readily useful in any project that does not use SSE events).
> > > I think with public endpoints/streams the biggest problem is scalability - it's not an issue here, as we have exactly 1 backend and 1 frontend.
> > 
> > 
> > I see the argument of this already being an improvement and the possibility to put it on the stream later.
> > Could we go one step further in this PR and add re-trying to connect to the stream in case we detect that the keep-alive endpoint is not available? Can this easily be triggered with the infrastructure we have currently in place?
> 
> That's not easy :( it will require more invasive change into the frontend, whilst I tried to keep this `BackendCheck` as a standalone component that can be used in any prototype we create.
> 
> Moving forward, I believe we don't even need to define any custom keepalive message, SSE events should be able to handle sending a heartbeat message and detecting the loss. I will raise a ticket to cover this improvement.

Probably my wording was not correct, a heartbeat is what I meant all along. But maybe that does not even need a specific message then and can be handled in an even simpler way. I have not looked into heartbeat implementations for sse in detail.

da-kami (approved):
Go for it then, I don't see blockers, just future improvements :)

klochowicz:
> > > > I agree that the solution would be nicer as it removes the need for the requests. I'm happy to come back and optimize it as a separate PR when I finish all my tasks scheduled for the current iteration ;)
> > > > For now, it's an improvement, as we don't detect lack of connectivity at all (not to mention that the component would be readily useful in any project that does not use SSE events).
> > > > I think with public endpoints/streams the biggest problem is scalability - it's not an issue here, as we have exactly 1 backend and 1 frontend.
> > > 
> > > 
> > > I see the argument of this already being an improvement and the possibility to put it on the stream later.
> > > Could we go one step further in this PR and add re-trying to connect to the stream in case we detect that the keep-alive endpoint is not available? Can this easily be triggered with the infrastructure we have currently in place?
> > 
> > 
> > That's not easy :( it will require more invasive change into the frontend, whilst I tried to keep this `BackendCheck` as a standalone component that can be used in any prototype we create.
> > Moving forward, I believe we don't even need to define any custom keepalive message, SSE events should be able to handle sending a heartbeat message and detecting the loss. I will raise a ticket to cover this improvement.
> 
> Probably my wording was not correct, a heartbeat is what I meant all along. But maybe that does not even need a specific message then and can be handled in an even simpler way. I have not looked into heartbeat implementations for sse in detail.

It's not about wording, I tried to say that it seems like we do not need to add any custom events, they might even already be there (depends on Rocket defaults, would need to check, and enable the heartbeat if it isn't). The trick will be to detect it appropriately on the client side, I'll look into that next week.

klochowicz:
> Go for it then, I don't see blockers, just future improvements :)

cool, I'll capture the thoughts from here in a separate ticket to address soon!

#482

PR #479, opened 2021-11-04
Auto rollover

### LINKED ISSUE #456: Automate roll-over
Right now the user has to trigger roll-over of a contract. This should be automated, e.g. on an regular interval the contracts should be rolled-over for another term (e.g. 24h). 

This will simplify our UI because it removes the user to manually have to roll-over. 

This only applies for the Taker, the Maker should still have to manually accept/decline the roll-over request.

da-kami:
fixes #456 

Planning to write an actor test against it, but it won't be totally trivial so I am planning to do it in a follow up PR.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -3,6 +3,7 @@ use crate::db::load_all_cfds;
 use crate::maker_cfd::{FromTaker, NewTakerOnline};
 use crate::model::cfd::{Cfd, Order, UpdateCfdProposals};
 use crate::oracle::Attestation;
+use crate::taker_cfd::RollOverParams;
 use anyhow::Result;
 use connection::ConnectionStatus;
 use maia::secp256k1_zkp::schnorrsig;
@@ -85,7 +86,7 @@ where
             Box<dyn MessageChannel<NewTakerOnline>>,
             Box<dyn MessageChannel<FromTaker>>,
         ) -> T,
-        settlement_time_interval_hours: time::Duration,
+        settlement_interval: time::Duration,
         n_payouts: usize,
     ) -> Result<Self>
     where
@@ -107,7 +108,7 @@ where
         let cfd_actor_addr = maker_cfd::Actor::new(
             db,
             wallet_addr,
-            settlement_time_interval_hours,
+            settlement_interval,
             oracle_pk,
             cfd_feed_sender,
             order_feed_sender,
@@ -189,6 +190,7 @@ where
         identity_sk: x25519_dalek::StaticSecret,
         oracle_constructor: impl FnOnce(Vec<Cfd>, Box<dyn StrongMessageChannel<Attestation>>) -> O,
         monitor_constructor: impl FnOnce(Box<dyn StrongMessageChannel<monitor::Event>>, Vec<Cfd>) -> F,
+        roll_over_params: RollOverParams,
         n_payouts: usize,
     ) -> Result<Self>
     where
@@ -208,9 +210,10 @@ where
 
         let (monitor_addr, mut monitor_ctx) = xtra::Context::new(None);
         let (oracle_addr, mut oracle_ctx) = xtra::Context::new(None);
-
+        let (cfd_actor_addr, cfd_context) = xtra::Context::new(None);
         let (connection_actor_addr, connection_actor_ctx) = xtra::Context::new(None);
-        let cfd_actor_addr = taker_cfd::Actor::new(
+
+        tokio::spawn(cfd_context.run(taker_cfd::Actor::new(
             db,
             wallet_addr,
             oracle_pk,
@@ -221,9 +224,8 @@ where
             monitor_addr.clone(),
             oracle_addr,
             n_payouts,
-        )
-        .create(None)
-        .spawn_global();
+            roll_over_params,
+        )));
 
         tokio::spawn(connection_actor_ctx.run(connection::Actor::new(
             maker_online_status_feed_sender,
@@ -253,6 +255,12 @@ where
             .spawn_global();
 
         tokio::spawn(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
+
+        // TODO: Optimize initial auto-rollover to take time into account so that we don't want to
+        // trigger auto-rollover after an immediate restart
+        cfd_actor_addr
+            .do_send_async(taker_cfd::AutoRollover)
+            .await?;
 
         Ok(Self {
             cfd_actor_addr,
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -115,7 +115,7 @@ pub struct Order {
     pub creation_timestamp: Timestamp,
 
     /// The duration that will be used for calculating the settlement timestamp
-    pub settlement_time_interval_hours: Duration,
+    pub settlement_interval: Duration,
 
     pub origin: Origin,
 
@@ -132,7 +132,7 @@ impl Order {
         max_quantity: Usd,
         origin: Origin,
         oracle_event_id: BitMexPriceEventId,
-        settlement_time_interval_hours: Duration,
+        settlement_interval: Duration,
     ) -> Result<Self> {
         let leverage = Leverage::new(2)?;
         let liquidation_price = calculate_long_liquidation_price(leverage, price);
@@ -147,7 +147,7 @@ impl Order {
             liquidation_price,
             position: Position::Short,
             creation_timestamp: Timestamp::now()?,
-            settlement_time_interval_hours,
+            settlement_interval,
             origin,
             oracle_event_id,
         })
@@ -671,7 +671,7 @@ impl Cfd {
     }
 
     pub fn refund_timelock_in_blocks(&self) -> u32 {
-        (self.order.settlement_time_interval_hours * Cfd::REFUND_THRESHOLD)
+        (self.order.settlement_interval * Cfd::REFUND_THRESHOLD)
             .as_blocks()
             .ceil() as u32
     }
@@ -680,20 +680,26 @@ impl Cfd {
         self.order.oracle_event_id.timestamp()
     }
 
-    /// A factor to be added to the CFD order settlement_time_interval_hours for calculating the
+    pub fn creation_timestamp(&self) -> Result<OffsetDateTime> {
+        let timestamp =
+            OffsetDateTime::from_unix_timestamp(self.order.creation_timestamp.seconds())?;
+        Ok(timestamp)
+    }
+
+    /// A factor to be added to the CFD order settlement_interval for calculating the
     /// refund timelock.
     ///
     /// The refund timelock is important in case the oracle disappears or never publishes a
     /// signature. Ideally, both users collaboratively settle in the refund scenario. This
     /// factor is important if the users do not settle collaboratively.
-    /// `1.5` times the settlement_time_interval_hours as defined in CFD order should be safe in the
+    /// `1.5` times the settlement_interval as defined in CFD order should be safe in the
     /// extreme case where a user publishes the commit transaction right after the contract was
     /// initialized. In this case, the oracle still has `1.0 *
-    /// cfdorder.settlement_time_interval_hours` time to attest and no one can publish the refund
+    /// cfdorder.settlement_interval` time to attest and no one can publish the refund
     /// transaction.
     /// The downside is that if the oracle disappears: the users would only notice at the end
-    /// of the cfd settlement_time_interval_hours. In this case the users has to wait for another
-    /// `1.5` times of the settlement_time_interval_hours to get his funds back.
+    /// of the cfd settlement_interval. In this case the users has to wait for another
+    /// `1.5` times of the settlement_interval to get his funds back.
     const REFUND_THRESHOLD: f32 = 1.5;
 
     pub const CET_TIMELOCK: u32 = 12;
--- a/daemon/src/taker.rs
+++ b/daemon/src/taker.rs
@@ -6,6 +6,7 @@ use clap::{Parser, Subcommand};
 use daemon::connection::ConnectionStatus;
 use daemon::model::WalletInfo;
 use daemon::seed::Seed;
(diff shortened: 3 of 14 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -196,6 +198,26 @@ impl<O, M, W> Actor<O, M, W> {
             })?;
         Ok(())
     }
+
+    async fn handle_auto_rollover(&mut self) -> Result<()> {
+        tracing::debug!("Auto rollover");
```
bonomat: Imho we should do a _smart_ auto rollover: 
check if the CFD's expiry is within 23h, if so, we rollover for another 24h. That way we can run it more regular for a low cost, e.g. every X minutes instead of hourly.

bonomat (commented):
> fixes #456
> 
> Todo:
> 
>     * [ ]  Change the rollover endpoint to toggle auto-rollover rather than allowing rollover itself.

I don't think we need the feature of toggling auto rollover.

daemon/src/taker_cfd.rs
da-kami: Note to future Daniel: `within 23h` will cover the CFD already being past attestation as well.

da-kami:
> > fixes #456
> > Todo:
> > ```
> > * [ ]  Change the rollover endpoint to toggle auto-rollover rather than allowing rollover itself.
> > ```
> 
> I don't think we need the feature of toggling auto rollover.

Without having the ability to put your position on the market (i.e. sell your position) I think we should add this feature. Otherwise users can only stop the cfd by going offline which is weird.

bonomat:
> Without having the ability to put your position on the market (i.e. sell your position) I think we should add this feature. Otherwise users can only stop the cfd by going offline which is weird.

The taker can always propose a settlement. If the maker is gone, then they will close the position and go on-chain and they will then receive the money at attestation point. 

This should happen automatically with: #483

da-kami:
> > Without having the ability to put your position on the market (i.e. sell your position) I think we should add this feature. Otherwise users can only stop the cfd by going offline which is weird.
> 
> The taker can always propose a settlement. If the maker is gone, then they will close the position and go on-chain and they will then receive the money at attestation point.
> 
> This should happen automatically with: #483

Fair enough

scratchscratchscratchy:
bors ping

scratchscratchscratchy:
bors ping

scratchscratchscratchy:
bors ping

daemon/src/lib.rs
```diff
@@ -254,6 +256,12 @@ where
 
         tokio::spawn(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
 
+        // TODO: Optimize initial auto-rollover to take time into account so that we don't want to
```
da-kami, line 259: Outdated, should be removed

daemon/src/taker.rs
```diff
@@ -249,6 +250,7 @@ async fn main() -> Result<()> {
                 monitor::Actor::new(electrum, channel, cfds)
             }
         },
+        RollOverParams::new(1, time::Duration::hours(1))?,
```
da-kami, line 253: This is the production code - why is the settlement interval set to `1 hour`?

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +189,71 @@ impl<O, M, W> Actor<O, M, W> {
...
+                self.roll_over_params.settlement_interval
+            );
+
+            let now = OffsetDateTime::now_utc();
+            let elapsed_since_creation = now - cfd.creation_timestamp()?;
+
+            // trigger rollover if rollover window has elapsed
+            if cfd.expiry_timestamp() > now
+                && elapsed_since_creation
+                    > self.roll_over_params.rollover_offset
```
da-kami, line 218: I cannot follow this logic, please explain what the values represent after your changes and how they relate to each other.

da-kami (commented):
@rishflab it would be great if you could open a new PR, so I can properly give this a review. Since this PR was created by me I can only comment - but the code does not reflect the initial solution anymore. 

I am confused by how the solution was changed.

Please explain in more detail:

* Why is the settlement interval set to `1 hour` in production code? It was set to `24 hours` before with an offset of `1 hour`. The settlement interval represents when the CFD will be settled (it defines how far in the future the oracle event should be). With the current changes that is not reflected anymore.
* The commit message does not reflect the code at all, it is my original commit message! As far as I can see e.g. the changes for removing the configurable settlement interval on the maker were dropped - why?

daemon/tests/happy_path.rs
```diff
@@ -59,6 +66,167 @@ async fn taker_takes_order_and_maker_rejects() {
...
+
+    let (taker_cfd, maker_cfd) = next_cfd(&mut taker.cfd_feed, &mut maker.cfd_feed).await;
+
+    // TODO: More elaborate Cfd assertions
+    assert_eq!(taker_cfd.order.id, received.id);
+    assert_eq!(maker_cfd.order.id, received.id);
+    assert!(matches!(taker_cfd.state, CfdState::PendingOpen { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::PendingOpen { .. }));
+
+    taker.trigger_auto_rollover().await;
```
da-kami, line 165: why is this called here already and then again later in the test?

daemon/src/model/cfd.rs
```diff
@@ -680,20 +680,26 @@ impl Cfd {
...
     ///
     /// The refund timelock is important in case the oracle disappears or never publishes a
     /// signature. Ideally, both users collaboratively settle in the refund scenario. This
     /// factor is important if the users do not settle collaboratively.
-    /// `1.5` times the settlement_time_interval_hours as defined in CFD order should be safe in the
+    /// `1.5` times the settlement_interval as defined in CFD order should be safe in the
     /// extreme case where a user publishes the commit transaction right after the contract was
     /// initialized. In this case, the oracle still has `1.0 *
-    /// cfdorder.settlement_time_interval_hours` time to attest and no one can publish the refund
+    /// cfdorder.settlement_interval` time to attest and no one can publish the refund
```
thomaseizinger, line 698: Can you put this rename into a different commit?

daemon/src/taker_cfd.rs
```diff
@@ -82,6 +101,7 @@ where
         + xtra::Handler<wallet::Sign>
         + xtra::Handler<wallet::BuildPartyParams>,
 {
+    #[allow(clippy::too_many_arguments)]
```
thomaseizinger, line 104: That seems pretty useless.

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,24 @@ enum RollOverState {
...
+    settlement_interval: Duration,
+    /// The number of settlement intervals in which an auto-rollover is triggered
+    rollover_offset: u32,
+}
+
+impl RollOverParams {
+    pub fn new(rollover_offset: u32, settlement_interval: Duration) -> Result<Self> {
+        if rollover_offset < 1 {
+            bail!("It does not make sense to rollover a cfd until at least 1 settlement interval has passed")
+        }
```
thomaseizinger, line 73: Use `NonZeroU32` instead.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +189,71 @@ impl<O, M, W> Actor<O, M, W> {
...
+            let now = OffsetDateTime::now_utc();
+            let elapsed_since_creation = now - cfd.creation_timestamp()?;
+
+            // trigger rollover if rollover window has elapsed
+            if cfd.expiry_timestamp() > now
+                && elapsed_since_creation
+                    > self.roll_over_params.rollover_offset
+                        * self.roll_over_params.settlement_interval
+            {
+                tracing::debug!("proposed roll over");
```
thomaseizinger, line 221: Make this log part of the `propose_roll_over` function and move it to the bottom so we don't erroneously log this in case the function fails.

daemon/tests/happy_path.rs
```diff
@@ -59,6 +66,167 @@ async fn taker_takes_order_and_maker_rejects() {
...
+        .expect_start_monitoring()
+        .return_const(());
+
+    #[allow(clippy::redundant_closure)] // clippy is in the wrong here
+    taker
+        .mocks
+        .monitor()
+        .await
+        .expect_start_monitoring()
+        .return_const(());
```
thomaseizinger, line 134: Please reduce this noise similar to what @klochowicz did in https://github.com/itchysats/itchysats/pull/511.

daemon/tests/happy_path.rs
```diff
@@ -59,6 +66,167 @@ async fn taker_takes_order_and_maker_rejects() {
...
+
+    #[allow(clippy::redundant_closure)] // clippy is in the wrong here
+    maker
+        .mocks
+        .monitor()
+        .await
+        .expect_start_monitoring()
+        .return_const(());
+
+    #[allow(clippy::redundant_closure)] // clippy is in the wrong here
```
thomaseizinger, line 128: There is no closure in this statement?

daemon/tests/happy_path.rs
```diff
@@ -59,6 +66,167 @@ async fn taker_takes_order_and_maker_rejects() {
...
+        .times(1)
+        .returning(|sign| Ok(sign.psbt))
+        .in_sequence(&mut seq);
+    wallet
+        .expect_broadcast()
+        .times(1)
+        .returning(|_| Ok(dummy_tx_id()))
+        .in_sequence(&mut seq);
+}
+
```
thomaseizinger, line 229: We have moved on to a different pattern, see https://github.com/itchysats/itchysats/pull/511.

daemon/tests/harness/mod.rs
```diff
@@ -151,6 +172,8 @@ impl Taker {
         oracle_pk: schnorrsig::PublicKey,
         maker_address: SocketAddr,
         maker_noise_pub_key: x25519_dalek::PublicKey,
+        settlement_interval: Duration,
+        rollover_offset: u32,
```
thomaseizinger, line 176: Do we ever configure this with something else other than the hardcoded values further up? If not, I'd prefer to not (yet) expose this as parameters.

daemon/tests/happy_path.rs
```diff
@@ -59,6 +66,167 @@ async fn taker_takes_order_and_maker_rejects() {
...
+        let update_proposals = next(&mut maker.update_proposals).await;
+        if let Some(update_proposal) = update_proposals.get(&maker_cfd.order.id) {
+            assert!(matches!(
+                update_proposal,
+                UpdateCfdProposal::RollOverProposal { .. }
+            ));
+            break;
+        }
+    }
+
```
thomaseizinger, line 200: Extract this into a helper please so we don't clutter up the test.

thomaseizinger: The helper should return the next update proposal by ID. The assertion should stay in the test.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +189,71 @@ impl<O, M, W> Actor<O, M, W> {
...
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn propose_roll_over(&mut self, order_id: OrderId) -> Result<()> {
+        if self.current_pending_proposals.contains_key(&order_id) {
+            anyhow::bail!("An update for order id {} is already in progress", order_id)
+        }
```
thomaseizinger, line 232: Can this ever be hit? We cleared the list on line 200?

daemon/src/model/cfd.rs
```diff
@@ -680,20 +680,26 @@ impl Cfd {
         self.order.oracle_event_id.timestamp()
     }
 
-    /// A factor to be added to the CFD order settlement_time_interval_hours for calculating the
+    pub fn creation_timestamp(&self) -> Result<OffsetDateTime> {
+        let timestamp =
+            OffsetDateTime::from_unix_timestamp(self.order.creation_timestamp.seconds())?;
+        Ok(timestamp)
+    }
```
thomaseizinger, line 687: I think we have some unnecessary conversions here. Can we get rid of these failure paths or at least change them to panics because they are unreachable?

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,24 @@ enum RollOverState {
     None,
 }
 
+pub struct RollOverParams {
+    settlement_interval: Duration,
+    /// The number of settlement intervals in which an auto-rollover is triggered
+    rollover_offset: u32,
```
thomaseizinger, line 66: Naming and docs need improvement. It is hard to workout what this value does without reading the code.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +189,71 @@ impl<O, M, W> Actor<O, M, W> {
...
+        // cleanup all pending proposals because auto-rollover will create a new one
+        self.current_pending_proposals.clear();
+
+        // Only cfds in state `Open` are allowed for auto-rollover proposal
+        for cfd in cfds
+            .iter()
+            .filter(|cfd| matches!(cfd.state, CfdState::Open { .. }))
+        {
+            tracing::debug!(
+                "settlement_interval: {:?}",
```
thomaseizinger, line 208: Debug formatting a duration is going to be noisy. Please either do seconds or something like this: https://docs.rs/humantime/2.1.0/humantime/fn.format_duration.html

daemon/src/taker_cfd.rs
rishflab: not currently, but if someones makes `propose_rollover` pub in the future, it would be nice to have this check. I dont think it does any harm to keep it?

daemon/src/taker_cfd.rs
rishflab: removing, i don't think we need to print this anymore

daemon/src/taker_cfd.rs
thomaseizinger: It does harm. It introduces complexity because its presence makes you think that it can be hit. Unnecessary code is one of the worst forms of complexity because you end up wasting hours of looking at it to only realize that is unrelated to the problem you are debugging.

PR #487, opened 2021-11-08
Transition to `SetupFailed` if the conract setup failed

da-kami:
Otherwise, the user does not get feedback until restarting the application.

Note: Given what is described in #364 this will be solved in a more complete way. This PR more of a workaround to get user feedback that the contract setup failed. At the moment we don't know until restarting the application which sucks.


bonomat (approved):
LGTM

PR #488, opened 2021-11-08
Shutdown taker when maker goes offline

klochowicz:
- periodically send heartbeat messages from maker to all connected takers,
- monitor connection to the maker using a connection_monitor actor
- publish the maker status on a watch::channel feed outside TakerActorSystem
- observe changes on the watch channel and shut down Rocket if maker goes offline

Note: this is a stepping stone resulting in temporary behaviour. Eventually,
     we'd want to support reconnecting to the maker as well as operating the
     taker without the maker being online (with a limited set of actions).


daemon/src/connection_monitor.rs
```diff
@@ -0,0 +1,144 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::connection_monitor;
+    use crate::tokio_ext::FutureExt;
+    use anyhow::Context;
+    use xtra::spawn::TokioGlobalSpawnExt;
+    use xtra::Actor as _;
+
+    // TODO: remove this copy-paste from happy_path.rs and extract to reusable place
```
da-kami: I think that comment is outdated

daemon/src/connection_monitor.rs
```diff
@@ -0,0 +1,143 @@
...
+            .spawn_global();
+
+        // Send the first heartbeat
+        maker_monitor_addr
+            .send(connection_monitor::Heartbeat)
+            .await
+            .unwrap()
+            .unwrap();
+
+        println!("Monitored entity should be online after heartbeat");
```
bonomat: Please no printlns. If you want to fail with a message, you could do this: 

```
assert_eq!(true, true, "true is always true!");
```

daemon/src/taker.rs
```diff
@@ -279,9 +281,20 @@ async fn main() -> Result<()> {
...
-        .await?;
+        );
+
+    let rocket = rocket.ignite().await?;
+    let shutdown_handle = rocket.shutdown();
+    rocket::tokio::spawn(rocket.launch());
+
+    // shutdown the rocket server maker if goes offline
+    tokio::spawn(async move {
+        maker_online_status_feed_receiver.changed().await.unwrap();
```
bonomat: Under which circumstances can this `unwrap` fail ?

bonomat (commented):
Not sure about this approach. I'll let smarter people review this. 
cc @da-kami / @thomaseizinger

daemon/src/connection_monitor.rs
klochowicz: nice, I'll do that!

daemon/src/taker.rs
klochowicz: "This method returns an error if and only if the Sender is dropped." - the sender will outlive the request, as it's managed. shall I use `.expect()` with some explanation instead?

klochowicz:
@thomaseizinger, I looked at the review and the test in the revised version does not handle coming back online. It's a substantial feature that `ConnectionMonitor` actor provided - otherwise we'll need one more actor to keep restarting the previous one?

Also: the feed to the UI will be needed too - the UI should indicate that there's no maker.

klochowicz:
also: in the recent commit we have a dummy actor that doesn't indicate that there's no connection in our actor tests, as it didn't need to be mocked :( I'd prefer to have the real actor inside, so that we can cover it in the tests properly, without the need to mock it too. (IMHO mocking only makes sense if it's needed to control application flow in a way to provoke certain responses or events, and that's not the case here)

thomaseizinger:
> @thomaseizinger, I looked at the review and the test in the revised version does not handle coming back online. It's a substantial feature that `ConnectionMonitor` actor provided - otherwise we'll need one more actor to keep restarting the previous one?
> 
> Also: the feed to the UI will be needed too - the UI should indicate that there's no maker.

What do you mean coming back online? We are shutting down the entire application, there is no coming back online?

I think the dead-man switch is a useful primitive but given that we don't know yet, how the actual connection management will look like, it may be best to just hold off on this entirely until we simply have dynamic connections?

klochowicz:
> > @thomaseizinger, I looked at the review and the test in the revised version does not handle coming back online. It's a substantial feature that `ConnectionMonitor` actor provided - otherwise we'll need one more actor to keep restarting the previous one?
> > Also: the feed to the UI will be needed too - the UI should indicate that there's no maker.
> 
> What do you mean coming back online? We are shutting down the entire application, there is no coming back online?
> 
> I think the dead-man switch is a useful primitive but given that we don't know yet, how the actual connection management will look like, it may be best to just hold off on this entirely until we simply have dynamic connections?

I have clearly stated in the PR that I do not want to shut down the application, and this is only a temporary state until the subsequent PR that I'm about to post on top of it. We absolutely want to have a running taker without the maker that can re-instantiate the connection, and IMHO we want it rather sooner than later (in fact, this is a ticket I was working on, and this is the first part of it!)

I do think that it's an useful primitive indeed (and love the name and usage of channels), but I don't think it makes my work here any easier - perhaps we could have both things, and you can resubmit your part as a separate PR? 
Anyway, I believe we should sync up.

da-kami:
My opinion: I think for the taker to know that we are not connected to the maker anymore is of vital importance - especially if we push for a demo that is to be used without our supervision against a public testnet maker that may be restarted at any point. 

That said, I think shutting down the application is a harsh step and this in between step should only be temporary. Nevertheless, I think *any* feedback that we are not connected anymore is an improvement. I think even printing just an Error in the logs would be an improvement.

da-kami (approved):
I don't think @klochowicz work should be blocked. I think we should merge this as it is an improvement (i.e. the taker knows that the daemon has to be restarted).
Of course this is just a temporary state and not optimal, but it is better than what we have now, where we just don't know anything about a lost connection. 

Further improvements can go in follow ups. Would that be OK with you @thomaseizinger ?

thomaseizinger:
> I don't think @klochowicz work should be blocked. I think we should merge this as it is an improvement (i.e. the taker knows that the daemon has to be restarted). Of course this is just a temporary state and not optimal, but it is better than what we have now, where we just don't know anything about a lost connection.
> 
> Further improvements can go in follow ups. Would that be OK with you @thomaseizinger ?

@klochowicz and myself will pair on a new iteration of this.

PR #492, opened 2021-11-09
SQL query fix

DeliciousHair:
Not completely necessary, but the previous ALTER commands only worked
because of being operated in an edge-case. The replacement here accomplishes
exactly the same thing but in a more stable and sqlite-compliant manner,
which makes the downstream workflow clearer at the same time.

### DIFF
--- a/daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
+++ b/daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
@@ -1,8 +1,59 @@
-alter table orders
-rename column term_seconds to settlement_time_interval_seconds;
+PRAGMA foreign_keys=off;
 
-alter table orders
-drop column creation_timestamp_nanoseconds;
+create temporary table orders_backup (
+  id,
+  uuid,
+  trading_pair,
+  position,
+  initial_price,
+  min_quantity,
+  max_quantity,
+  leverage,
+  liquidation_price,
+  creation_timestamp_seconds,
+  settlement_time_interval_seconds,
+  origin,
+  oracle_event_id 
+);
 
-alter table orders
-drop column term_nanoseconds;
+insert into orders_backup
+  select
+    id,
+    uuid,
+    trading_pair,
+    position,
+    initial_price,
+    min_quantity,
+    max_quantity,
+    leverage,
+    liquidation_price,
+    creation_timestamp_seconds,
+    term_seconds,
+    origin,
+    oracle_event_id 
+  from orders;
+
+drop table orders;
+
+create table orders (
+  id                                integer primary key autoincrement,
+  uuid                              text unique not null,
+  trading_pair                      text        not null,
+  position                          text        not null,
+  initial_price                     text        not null,
+  min_quantity                      text        not null,
+  max_quantity                      text        not null,
+  leverage                          integer     not null,
+  liquidation_price                 text        not null,
+  creation_timestamp_seconds        integer     not null,
+  settlement_time_interval_seconds  integer     not null,
+  origin                            text        not null,
+  oracle_event_id                   text        not null
+);
+
+insert into orders
+  select *
+  from orders_backup;
+drop table orders_backup;
+
+PRAGMA foreign_keys=on;


da-kami:
Is this really necessary? I don't think it adds much value to create a backup table here given that we just rename a column. I don't think this is more understandable than what was there before, but that's probably just me. 

One thing we could have fixed is the rename from `Buy`/`Sell` to `Long`/`Short` because that was actually not backwards compatible. But I would not care at this stage - I think all of us already deleted that db version.

DeliciousHair:
> Is this really necessary?

Technically, no. However, I put it in as the existing use of `ALTER` only works because of it being applied to one of the (very limited) cases where `ALTER` can be used by sqlite. The backup-drop-recrecate approach is the usual approach, so this was stuck in for consistency downstream is all.

daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
```diff
@@ -1,8 +1,59 @@
-alter table orders
-rename column term_seconds to settlement_time_interval_seconds;
+PRAGMA foreign_keys=off;
```
bonomat, line 1: You should never change old migration sql files. 
These are meant to be incremental. 

Here a short how to use sqlx-cli: 
https://github.com/launchbadge/sqlx/blob/master/sqlx-cli/README.md#sqlx-cli

If you look at the filename it says : 

`20211025050345_rename_term_to_settlement_time_interval.sql`

So on `2021 - 11 - 02` we renamed the term `settlement time` to `interval`. 

You changed the content to something completely different and the filename does not fit anymore

Differently said, any change to the db should have its own migration file. 

This way you can track how the database has evolved over time.

bonomat (dismissed):
You should not change db migrations file but create new ones if the schema changes.

daemon/migrations/20211025050345_rename_term_to_settlement_time_interval.sql
DeliciousHair: Note that the `sqlx.json` file is unaffected by this, meaning that I am not altering the incremental revision approach at all but merely changing the syntax used to be something that is more "correct" in the sense of it not requiring us to be in a limited-case scenario, which is what was done previously.

The outcomes are identical in both cases.

bonomat: Thanks for pointing this out. 

Regarding the changes: it's not clear why this is necessary. In your commit message you mention: 

> Not completely necessary, but the previous ALTER commands only worked
because of being operated in an edge-case.

What are the edge-cases and when could the old approach not work? 

The old file was much smaller and imho easier to understand.

DeliciousHair: Please refer to the SQLite docs for a complete answer; short version is that the `ALTER` command only worked because of there being no fkeys involved, and **possibly** (not sure about this) being so early in the game. In other words, it is implicitly working on an edge-case where it manages to work by accident, not design. This edit makes things clearer going forward regarding the use of the recommended process is all.

As I have stated already though, not completely necessary at all.

bonomat: I don't understand what you are saying :( 

> short version is that the ALTER command only worked because of there being no fkeys involved, 

Differently said, it works if no foreign keys are involved. Which there aren't, so that's fine then? 
Since schema migration files are run sequentially, this fact will never change, i.e. i.e. the columns being altered here will never become foreign keys. Differently said, if it works now, it will always work. Or am I misunderstanding something? 

> and possibly (not sure about this) being so early in the game

What do you mean with this?

Maybe you can point me to that part of the SQLite docs you are referring to please?

DeliciousHair: Since this PR doesn't actually "do" anything (it's for syntax purposes only) I propose this simply gets closed as it is clearly causing too much hassle.

da-kami:
closing for now, please re-open if necessary

PR #511, opened 2021-11-10
Test cleanup

klochowicz:
A small refactor, hopefully reducing entry barrier in creating new tests.

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -1,17 +1,16 @@
-use crate::harness::bdk::dummy_partially_signed_transaction;
-use crate::harness::mocks::oracle::dummy_announcement;
-use crate::harness::mocks::wallet::build_party_params;
+use crate::harness::mocks::{
+    mock_oracle_annoucement, mock_party_params, mock_wallet_sign_and_broadcast,
+};
 use crate::harness::start_both;
 use anyhow::Context;
 use daemon::maker_cfd;
 use daemon::model::cfd::{Cfd, CfdState, Order, Origin};
 use daemon::model::{Price, Usd};
 use daemon::tokio_ext::FutureExt;
-use harness::bdk::dummy_tx_id;
 use maia::secp256k1_zkp::schnorrsig;
 use rust_decimal_macros::dec;
 use std::time::Duration;
-use tokio::sync::{watch, MutexGuard};
+use tokio::sync::watch;
 use tracing::subscriber::DefaultGuard;
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::util::SubscriberInitExt;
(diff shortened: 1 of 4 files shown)


daemon/tests/happy_path.rs
```diff
@@ -101,39 +84,13 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
...
-        .expect_get_announcement()
-        .returning(|_| Some(dummy_announcement()));
-
-    taker
-        .mocks
-        .oracle()
-        .await
-        .expect_get_announcement()
-        .returning(|_| Some(dummy_announcement()));
+    mock_oracle_annoucement(&mut maker.mocks.oracle().await);
```
thomaseizinger: What about:

```suggestion
    maker.mocks.mock_oracle_announcement();
```

thomaseizinger (approved):

daemon/tests/happy_path.rs
klochowicz: cool! did it for all of them. accessing mutexguard is `async`, so I had to `.await` all of these calls - but they still look nicer than before :) and it should be easier to discover the calls with the IDE.

daemon/tests/happy_path.rs
```diff
@@ -76,13 +73,13 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
     taker.take_order(received.clone(), Usd::new(dec!(5)));
     let (_, _) = next_cfd(&mut taker.cfd_feed, &mut maker.cfd_feed).await;
 
-    mock_oracle_annoucement(&mut maker.mocks.oracle().await);
-    mock_oracle_annoucement(&mut taker.mocks.oracle().await);
+    maker.mocks.mock_oracle_annoucement().await;
```
thomaseizinger, line 76: In an ideal world, we can even get rid of the `.mocks` indirection without too much duplication!

daemon/tests/harness/mocks/mod.rs
```diff
@@ -32,6 +32,38 @@ impl Mocks {
...
+            .returning(|_| Ok(dummy_tx_id()))
+            .in_sequence(&mut seq);
+    }
+
+    pub async fn mock_oracle_annoucement(&mut self) {
+        self.oracle()
+            .await
+            .expect_get_announcement()
+            .return_const(Some(oracle::dummy_announcement()));
+    }
```
thomaseizinger, line 58: One reason I am liking this is because we can easily make `dummy_announcement` configurable if we need at some point!

thomaseizinger (approved):

daemon/tests/happy_path.rs
klochowicz: I don't mind this small indirection as it enforces separation of concerns, as it gets rid of any duplication (and there will be more methods...)

klochowicz:
@thomaseizinger @da-kami - could anyone bors this? I don't have permissions 😢

PR #512, opened 2021-11-10
Use xtra-productivity from comit-network

klochowicz:


Cargo.toml
```diff
@@ -1,9 +1,10 @@
...
-members = ["daemon", "xtra_productivity"]
+members = ["daemon"]
 resolver = "2"
 
 [patch.crates-io]
 rocket = { git = "https://github.com/SergioBenitez/Rocket" } # Need to patch rocket dependency of `rocket_basicauth` until there is an official release.
 xtra = { git = "https://github.com/Restioson/xtra" } # Latest master has crucial patches.
 secp256k1-zkp = { git = "https://github.com/ElementsProject/rust-secp256k1-zkp" } # Latest master has crucial patches.
 maia = { git = "https://github.com/comit-network/maia", rev = "70fc548da0fe4f34478fb34ec437fa9a434c7ee3" }
+xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity", rev = "eda11a682dbbf16a9004cfe5e30baa1c60062a67" }
```
thomaseizinger, line 10: Strictly speaking, we wouldn't need to use `patch` here because this is a top-level dependency only and thus does not need replacement across the dependency graph. A git dependency within `daemon` would be sufficient. One could still argue that it is better for consistency reasons. I am not sure.

thomaseizinger (approved):

Cargo.toml
klochowicz: Yes, I went with consistency - so it's easier to find it and remove it when we publish it to `crates.io` - it will disappear then anyway :)

klochowicz:
bors ping

PR #514, opened 2021-11-10
Publish docker image improvements

1010Tom:
Note: the only thing we need from the checkout is the dockerfile. The rest does not matter. 
Nevertheless, when manually releasing docker images by release tag, we should check out that particular release tag to ensure we are using the correct dockerfile

### DIFF
--- a/.github/workflows/build-release-binary.yml
+++ b/.github/workflows/build-release-binary.yml
@@ -155,7 +155,7 @@ jobs:
       contents: read
       packages: write
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_TOKEN: ${{ secrets.BOTTY_GITHUB_TOKEN }}
     steps:
       - name: Checkout tagged commit
         uses: actions/checkout@v2.4.0
--- a/.github/workflows/build-release-binary.yml
+++ b/.github/workflows/build-release-binary.yml
@@ -154,6 +154,11 @@ jobs:
     permissions:
       contents: read
       packages: write
+    strategy:
+      matrix:
+        include:
+          - bin: maker
+          - bin: taker
     env:
       GITHUB_TOKEN: ${{ secrets.ITCHY_GITHUB_TOKEN }}
     steps:
@@ -176,7 +181,7 @@ jobs:
 
       - name: Download release from github
         run: |
-          gh release download ${{ github.event.release.tag_name }} -p '*Linux_x86_64*' -p '*aarch64*' -p '*armv7*'
+          gh release download ${{ github.event.release.tag_name }} -p '${{matrix.bin}}*Linux_x86_64*' -p '${{matrix.bin}}*aarch64*' -p '${{matrix.bin}}*armv7*'
 
       - name: Unpack archives
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release-docker.yml
```diff
@@ -20,7 +20,8 @@ jobs:
     steps:
       - name: Checkout tagged commit
         uses: actions/checkout@v2.4.0
-      # TODO: checkout the tag and not master
+        with:
+          ref: ${{ github.event.inputs.realease_tag }}
```
thomaseizinger: Typo!

```suggestion
ref: ${{ github.event.inputs.release_tag }}
```

.github/workflows/release-docker.yml
```diff
@@ -58,22 +58,34 @@ jobs:
 
       - name: Build maker docker image
         run: |
+          VERSION=${{ github.event.inputs.realease_tag }}
```
thomaseizinger:
```suggestion
VERSION=${{ github.event.inputs.release_tag }}
```

.github/workflows/build-release-binary.yml
```diff
@@ -164,7 +164,7 @@ jobs:
       contents: read
       packages: write
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      GITHUB_TOKEN: ${{ secrets.BOTTY_GITHUB_TOKEN }}
```
thomaseizinger: If the user is not botty, we should name this something else!

thomaseizinger (changes requested):

.github/workflows/release-docker.yml
1010Tom: Same typo is also in the action input parameters 🙈

thomaseizinger: Let's fix it everywhere then!

1010Tom: done and force pushed

.github/workflows/build-release-binary.yml
1010Tom: Done, renamed all to `ITCHY_`

.github/workflows/build-release-binary.yml
```diff
@@ -206,22 +206,34 @@ jobs:
...
               --push \
               --platform linux/amd64,linux/arm64,linux/arm/v7 \
               --label "org.opencontainers.image.source https://github.com/${{ github.repository }}" \
-              --tag ghcr.io/${{ github.repository }}/maker:${{ github.event.release.tag_name }} \
+              --tag ghcr.io/${{ github.repository }}/maker:${VERSION} \
               --build-arg BINARY_PATH=maker \
               .
 
 
       - name: Build taker docker image
```
thomaseizinger: Could we handle this with a `matrix` to avoid the duplication?

.github/workflows/build-release-binary.yml
```diff
@@ -206,22 +206,34 @@ jobs:
 
       - name: Build maker docker image
         run: |
+          VERSION=${{ github.event.release.tag_name }}
+
+          if [ "$VERSION" = "preview" ]; then
+              VERSION=latest
+          fi
```
thomaseizinger: Something like this might be cleaner: https://stackoverflow.com/a/3953666/2489334

thomaseizinger (approved):
Few more comments.

I only just noticed that we have two workflows now. Might be worth looking into reducing this duplication: https://docs.github.com/en/actions/learn-github-actions/reusing-workflows

1010Tom:
> I only just noticed that we have two workflows now. Might be worth looking into reducing this duplication: docs.github.com/en/actions/learn-github-actions/reusing-workflows

@thomaseizinger : I hope this second workflow can go away once the automated one works flawlessly again.

1010Tom:
@thomaseizinger : changed my mind :D reusing the workflow might actually be good and allows us to manually trigger releasing images or have it automated.

.github/workflows/release-docker.yml
```diff
@@ -15,12 +21,18 @@ jobs:
     permissions:
       contents: read
       packages: write
+    strategy:
+      matrix:
+        include:
+          - bin: maker
+          - bin: taker
```
thomaseizinger:
```suggestion
        bin: [maker, taker]
```

`include` is designed for specifying combinations manually. If all you need is permutations, there is a simpler way.

thomaseizinger (approved):

.github/workflows/release-docker.yml
1010Tom: thanks :)

1010Tom:
Still testing on my fork btw: https://github.com/itchymax/itchysats/actions/runs/1450463643

1010Tom:
> @thomaseizinger : changed my mind :D reusing the workflow might actually be good and allows us to manually trigger releasing images or have it automated.

Unfortunately I can't get it working. 
I tried with an extra workflow which calls the `docker-release` workflow using an `input` variable. This works but when triggered from the `build-release` workflow the variable `${{ github.event.release.tag_name }}` does not seem to be propagated down to the called workflow. 
Its always undefined....

I undid the last commit and will go for a merge with this solution.

1010Tom:
worked on my fork: https://github.com/itchymax/itchysats/actions/runs/1450960567

bors r+

Issue #510, opened 2021-11-10
Fix "mocked" behavior on frontend


1010Tom:
A few things show up on the frontend but are not actually implemented: 

- Go long dialog checkbox: `Always show this dialog` has no effect
- Close position: `Always show this dialog` has no effect

bonomat:
resolved in #515

PR #515, opened 2021-11-10
Remove checkboxes

bonomat:
I thought this is a better solution for #510 for now


klochowicz (approved):
it's an improvement :) can always revert it when we get to that piece of functionality.

PR #527, opened 2021-11-11
Generate seed automatically if no seed is present

### LINKED ISSUE #505: `-generate-seed` is a bit cumbersome
Having to add the argument `--generate-seed` is a bit cumbersome, especially that you have to add it once and then not again. 

Ideally it would not be required and the seed would be automatically generated. Or, `generate-seed` is ignored if a seed was already generated before.

klochowicz:
Smooth out user experience by automatically generating a new seed if there
is none present (and by writing that fact in the log file).

Fixes #505

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,7 @@ jobs:
           sleep 10s # Wait for maker to start\
 
           # The maker-id is generated from the makers seed found in daemon/util/testnet_seeds/maker_seed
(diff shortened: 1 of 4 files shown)


bonomat (approved):
Afaik you need to update the ci.yml file as well. look for the smoke test or `--generate-seed `

klochowicz:
> Afaik you need to update the ci.yml file as well. look for the smoke test or `--generate-seed `

thanks! `ripgrep` seems to by default exclude `.github` from search, probably need to tweak its settings, as it's not the first time I got tripped by that :(

Issue #524, opened 2021-11-11
Taker UI displays `0.0` when there's no order from the maker


klochowicz:
This is confusing, as it can be assumed that the app broke or the connection to the maker got lost.

also: it's not clear in the first place that the price is the offer price. I actually thought it would be current BTC price.

1010Tom:
> also: it's not clear in the first place that the price is the offer price. I actually thought it would be current BTC price.

I propose to do exactly that, i.e. show the current price from BTC/USD here (Bitmex offers that as an event source afaik it should be `wss://www.bitmex.com/realtime?subscribe=instrument:.BXBT` ).

klochowicz:
but then we still need to show the offer price somewhere else?

1010Tom:
I'd do that in the dialog then. I.e. this is a _market order_. You don't 100% know what rate you get before you hit send.

1010Tom:
@da-kami : will look into why the taker does not pick up the latest price and I will update the price to show the BXBT rate.

da-kami:
One cause of this error is  https://github.com/itchysats/itchysats/issues/526 which I can reproduce locally. 
So far, without going over the actual wire I am unable to reproduce the behavior where we don't receive the order after taking a cfd. 

Some more notes on what was observed so far to get a more complete picture:

When testing over the wire it was observed, that even though there was a connection with the maker (taker and maker were in the middle of exchanging contract setup parameters), the taker did not properly receive the new order that was re-published. 
Re-publishing another order manually did not help. Refreshing the taker did not help. Only restarting the daemon fixed the problem. 
This behavior is weird, because the connection was definitely not dropped (contract setup messages were exchanged). But the order messages somehow did not reach the taker UI.

da-kami:
was able to reproduce locally, checking :)

PR #529, opened 2021-11-11
Use BXBT index for main price

1010Tom:
Improves #524


taker-frontend/src/App.tsx
```diff
@@ -42,6 +44,22 @@ async function postCfdOrderRequest(payload: CfdOrderRequestPayload) {
 export const App = () => {
     const toast = useToast();
 
+    const {
+        lastMessage,
+        readyState,
+    } = useWebSocket("wss://www.bitmex.com/realtime?subscribe=instrument:.BXBT", {
```
da-kami, line 50: We should do this in the daemon. After discussion with @itchymax we realized that the daemon price feed might not be representing all prices we need. 
The way we internally use the price is not correct in all cases. We should properly clean this up, but this might take more effort.

If we decide to do this in a follow up, please create a ticket that describes the motivation of the different prices and which one to use when.

da-kami (approved):
Since getting the reference price into the daemon might require more thinking I'm ok with getting this into the UI for now so what we display to the user makes sense.

taker-frontend/src/App.tsx
1010Tom: Issue created #533

PR #530, opened 2021-11-11
Remove order before transition to `IncomingOrderRequest`

### LINKED ISSUE #524: Taker UI displays `0.0` when there's no order from the maker

da-kami:
fixes https://github.com/itchysats/itchysats/issues/524


bonomat (approved):
👍

PR #531, opened 2021-11-11
Cleanup frontend code

klochowicz:
Remove the `APP` variable from `maker-frontend` (from now on, it only builds maker).
Update README, CI and start_all.sh script.

note: the old taker component will be remove in the next PR, as I'm still mining it for ideas.


.github/workflows/build-release-binary.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         run: |
           cd maker-frontend;
           yarn
-          APP=maker yarn build
+          yarn build
```
thomaseizinger: We could write this as:

```suggestion
          cd ${{ matrix.bin }}-frontend;
          yarn
          yarn build
```

And get rid of the `if`.

maker-frontend/vite.config.ts
```diff
@@ -1,18 +1,9 @@
...
-if (!app || (app !== "maker" && app !== "taker")) {
-    throw new Error("APP environment variable needs to be set to `maker` `taker`");
-}
-
-const backendPorts = {
-    "taker": 8000,
-    "maker": 8001,
-};
+const app = "maker";
+const backendPort = 8001;
```
thomaseizinger: I guess this indirection can also go away now!

thomaseizinger (approved):

.github/workflows/build-release-binary.yml
klochowicz: wouldn't this also make it a bit harder to find out which job failed? (right now they're named separately, so you see at the first glance which frontend failed on the CI)

edit: nevermind, I see the point in avoiding needless duplication too ;) Added as a separate commit.

PR #532, opened 2021-11-11
Taker handles order received from maker already known

### LINKED ISSUE #526: Taker cannot handle "same order" on startup
When the taker receives the same order from the maker we see this error:

```
2021-11-11 02:54:02 ERROR Message handler failed: error returned from database: UNIQUE constraint failed: orders.uuid: UNIQUE constraint failed: orders.uuid
```

I have only observed this on startup. It happens when we shut down the taker, start it again and the order sent to the taker by the maker upon establishing a connection is the same as before we shutdown the taker. 

The taker has to properly guard against already knowing an order. As long as no cfd was created for that order yet we can just proceed with an already known order.
If we have a cfd in the database for an order yet, but the maker is still broadcasting said order, we should error in the taker daemon (but still not try to save the known order again...).

da-kami:
fixes https://github.com/itchysats/itchysats/issues/526

In case we already have a cfd for the received order we bail, because that should not happen.
If we already know the order, but don't have a related cfd we don't insert but still send it out on the order feed.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -9,7 +9,7 @@ use crate::model::{BitMexPriceEventId, Price, Timestamp, Usd};
 use crate::monitor::{self, MonitorParams};
 use crate::wire::{MakerToTaker, RollOverMsg, SetupMsg};
 use crate::{log_error, oracle, setup_contract, wallet, wire};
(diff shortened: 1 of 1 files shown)


1010Tom (approved):
👍

PR #541, opened 2021-11-11
Disable leverage slider

thomaseizinger:
Before:

image

After:

image

### DIFF
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -33,6 +33,7 @@ import {
     Tbody,
     Td,
     Text,
+    Tooltip,
     Tr,
     useColorModeValue,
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
👍

da-kami (approved):
How about adding a tooltip that explains that this is fixed to `2` at the moment?
I think it is confusing to the user if we don't state that it cannot be changed at the moment, users might expect that they "have to do something" so it can be changed...

thomaseizinger:
> How about adding a tooltip that explains that this is fixed to `2` at the moment? I think it is confusing to the user if we don't state that it cannot be changed at the moment, users might expect that they "have to do something" so it can be changed...

Done!

image

da-kami (approved):

PR #544, opened 2021-11-12
Use fewer payouts in the tests

thomaseizinger:
This makes our test run fast by default without the need for a feature flag.


da-kami (approved):
Cool, that should make the tests fast! Thanks!

You'll have to remove running with the removed feature flag on CI

PR #548, opened 2021-11-12
Release version 0.1.2

1010Tom:
Hi @scratchscratchscratchy!
This PR was created in response to a manual trigger of the release workflow here: https://github.com/itchysats/itchysats/actions/runs/1451669850.
I've bumped the versions in the manifest files in this commit: 68cad44f709194d6b3fd1687f733c76b23005ee3.
Merging this PR will create a GitHub release and publish the library to crates.io!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -601,7 +601,7 @@ dependencies = [
 
 [[package]]
 name = "daemon"
-version = "0.1.1"
+version = "0.1.2"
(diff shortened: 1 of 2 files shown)


scratchscratchscratchy:
decided that we don't need this release because we use the latest release for now in the posts. has more advantages than dis-, because we keep moving so fast with fixing things.

Issue #550, opened 2021-11-12
Maker randomly closes connection to taker


bonomat:
I don't have much information unfortunately, but the maker looks like it's randomly closing the connection to the taker: 

```
2021-11-12 16:25:16  INFO POST /api/order/sell application/json:
2021-11-12 16:25:25 DEBUG Taker wants to take an order taker_id=5a0b1a14-b7b5-4992-af26-d93aadf7032c quantity=10 order_id=b46ad6be-cdcd-4d0b-823b-7bf6f2db2ae3
2021-11-12 16:25:25  WARN Taker tried to take order with outdated id b46ad6be-cdcd-4d0b-823b-7bf6f2db2ae3
2021-11-12 16:25:36 ERROR Closing connection to taker 5a0b1a14-b7b5-4992-af26-d93aadf7032c
```

da-kami:
This happens here:

https://github.com/itchysats/itchysats/blob/b7a8370b0b4638b4359abd48577d41985cac32cf/daemon/src/maker_inc_connections.rs#L124-L135

This results in having to restart the taker (the taker does not receive new orders ..., because the connection is gone). - The taker does not see any errors. I don't understand why the maker closes  the connection at this point.

da-kami:
Did some more tests over the wire. The maker sees an incoming connections and sends an initial order:

```
2021-11-12 16:50:26 DEBUG new order Some(Order { id: OrderId(Hyphenated(b6f24ca9-35b9-4543-873d-17d00fa4c512)), trading_pair: BtcUsd, position: Short, price: Price(65047.619999999995), min_quantity: Usd(10), max_quantity: Usd(1000), leverage: Leverage(2), liquidation_price: Price(43365.079999999996666666666667), creation_timestamp: Timestamp(1636696216), settlement_time_interval_hours: Duration { seconds: 86400, nanoseconds: 0 }, origin: Ours, oracle_event_id: BitMexPriceEventId { timestamp: OffsetDateTime { utc_datetime: PrimitiveDateTime { date: Date { year: 2021, ordinal: 317 }, time: Time { hour: 6, minute: 0, second: 0, nanosecond: 0 } }, offset: UtcOffset { hours: 0, minutes: 0, seconds: 0 } }, digits: 20 } })
```

That is received on the taker side (timestamps match). 

The connection is dead, any other order created does not arrive.
There is no close connection recorded on either taker or maker.
The maker does not see an error.
The taker does not see an error. 

[...]

PR #551, opened 2021-11-12
Configurable log levels on both taker and maker

da-kami:
Add configurable log levels on taker and maker and start tracing connection related information to track down a problem where the taker does not receive orders but there are not connection errors on both sides.

https://github.com/itchysats/itchysats/issues/550


bonomat (approved):
image

PR #552, opened 2021-11-12
Use a connection actor for the taker

klochowicz:
Connection actor takes care of establishing a connection to the maker and is
monitoring its status.

The connection actor:
- listens for periodically sent heartbeat messages from the maker,
- publish the maker status on a watch::channel feed outside TakerActorSystem,
- observe changes on the watch channel and shut down Rocket if maker goes offline

Note: this is a stepping stone resulting in temporary behaviour. Eventually,
     we'd want to support reconnecting to the maker as well as operating the
     taker without the maker being online (with a limited set of actions).

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -1,57 +1,174 @@
-use crate::{noise, send_to_socket, taker_cfd, wire};
+use crate::{log_error, noise, send_to_socket, wire};
 use anyhow::Result;
-use futures::{Stream, StreamExt};
+use futures::future::RemoteHandle;
+use futures::{FutureExt, StreamExt};
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
-use std::time::Duration;
+use std::time::{Duration, SystemTime};
+use tokio::sync::watch;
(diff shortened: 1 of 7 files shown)


daemon/src/connection.rs
```diff
@@ -1,57 +1,174 @@
...
 use tokio_util::codec::FramedRead;
 use xtra::prelude::MessageChannel;
-use xtra::spawn::TokioGlobalSpawnExt;
-use xtra::Actor as _;
+use xtra::KeepRunning;
+use xtra_productivity::xtra_productivity;
 
-const CONNECTION_RETRY_INTERVAL: Duration = Duration::from_secs(5);
+struct ConnectedState {
+    pub last_heartbeat: SystemTime,
```
thomaseizinger: I think this can be private.

thomaseizinger (approved):

klochowicz:
For those interested:

The main differences between this PR and the previous unmerged one ( #488 ):
- send the heartbeat down from maker_inc_connections actor (so maker_cfd can be unaware of heartbeats)
-  encapsulate whole connection management inside a single actor

PR #563, opened 2021-11-15
Cleanup bad connections for good

thomaseizinger:


daemon/src/maker_inc_connections.rs
```diff
@@ -84,14 +84,16 @@ impl Actor {
...
         let conn = self
             .write_connections
-            .get(&taker_id)
+            .get(taker_id)
             .context("no connection to taker_id")?;
 
-        // use `.send` here to ensure we only continue once the message has been sent
-        conn.send(msg).await?;
+        if conn.send(msg).await.is_err() {
+            tracing::info!(%taker_id, "Bad connection to taker, removing ...");
```
klochowicz: nit: `s/Bad/Lost` ? (a bad connection is an ambiguous term, it can often refer to a connection that is still active, but unreliable/noisy)

klochowicz (approved):
great! a good step towards stability 👍

klochowicz (approved):

daemon/src/maker_inc_connections.rs
```diff
@@ -156,57 +160,50 @@ impl Actor {
...
-                .await
-                .is_err()
-            {
-                tracing::trace!(%taker_id, "removing outdated connection to taker because unable to send order: {:?}", order);
-                self.write_connections.remove(&taker_id);
-            } else {
-                tracing::trace!(%taker_id, "sent new order: {:?}", order);
-            }
+        for taker_id in self.write_connections.clone().keys() {
+            self.send_to_taker(taker_id, wire::MakerToTaker::CurrentOrder(order.clone())).await.expect("send_to_taker only fails on missing hashmap entry and we are iterating over those entries");
```
da-kami, line 164: ❓ Can it happen that we access the hashmap from a different handle block and an element gets removed while we are looping in here? 

It would be great if we can remove the failure scenario from the for loop. Regardless of "it cannot happen". I'm afraid that changing code might re-introduce problems that we already solved. I think we should not fail in the loop no matter what. Maybe just wrap in a `try_continue` for now?

da-kami (dismissed):

daemon/src/maker_inc_connections.rs
thomaseizinger: We are holding a mutable reference to `Actor` here. These kind of race conditions are impossible in Rust.

thomaseizinger: If this panics, someone modified the `send_to_taker` function without thinking through all call-sites which is a bug and should be fixed.

da-kami: I still think it is bad to have a failure scenario in a loop and I don't think we should expect this, because if `send_to_taker` changes we might re-introduce an error that was hard to track down.

da-kami:
Can you at least please confirm that we will see the `expect` message if it ever happens? Previously we had a `?` here and never saw an error.

thomaseizinger:
> Can you at least please confirm that we will see the `expect` message if it ever happens? Previously we had a `?` here and never saw an error.

Yes, panics result in a stacktrace in the logs.

da-kami (approved):
Thanks for cleaning this up!

da-kami:
> > Can you at least please confirm that we will see the `expect` message if it ever happens? Previously we had a `?` here and never saw an error.
> 
> Yes, panics result in a stacktrace in the logs.

Any idea why we did not see the `?`ed error that was previously there? I was wondering if we have more code parts that swallow errors and if that's expected...?

thomaseizinger:
> > > Can you at least please confirm that we will see the `expect` message if it ever happens? Previously we had a `?` here and never saw an error.
> > 
> > 
> > Yes, panics result in a stacktrace in the logs.
> 
> Any idea why we did not see the `?`ed error that was previously there? I was wondering if we have more code parts that swallow errors and if that's expected...?

Yes, both call-sites where we send `BroadcastOrder` use `do_send_async` which is async and hence discard the return value.

PR #566, opened 2021-11-15
Migrate to Rust 2021

thomaseizinger:
Nothing to do because our code is already idiomatic!

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.55"
+channel = "1.56"
 components = ["clippy"]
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.55"
+channel = "1.56"
```
klochowicz, line 2:
```suggestion
channel = "1.56.1"
```

I believe that that's the current stable

klochowicz (approved):
sweet! 👯‍♂️ 👯‍♀️

rust-toolchain.toml
thomaseizinger: It will auto-pick the highest patch release AFAIK.

PR #568, opened 2021-11-15
Minor fixes improving UX of taker shutdown

klochowicz:
- reduce log level of heartbeat message
- improve the toast message to clearly state possible action

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -138,7 +138,7 @@ impl Actor {
 
         match msg {
             wire::MakerToTaker::Heartbeat => {
-                tracing::info!("got the heartbeat");
+                tracing::trace!("received a heartbeat message from maker");
                 self.connected_state
                     .as_mut()
                     .expect("wire messages only to arrive in connected state")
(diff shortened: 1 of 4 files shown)


daemon/src/connection.rs
```diff
@@ -138,7 +138,7 @@ impl Actor {
 
         match msg {
             wire::MakerToTaker::Heartbeat => {
-                tracing::info!("got the heartbeat");
+                tracing::debug!("received a heartbeat message from maker");
```
da-kami:
```suggestion
                tracing::trace!("received a heartbeat message from maker");
```

Let's do trace please. This is not important information and we run on debug by default. It heavily clutters the logs :)

da-kami (changes requested):
Please also change the daemon message `Maker is offline. Shutting down the taker` to something like `Lost connection to maker, shutting down. Please restart the daemon to reconnect`

For now the user will likely see both messages (in UI and daemon) and it would be great if they roughly say the same things.

thomaseizinger (approved):
ACK modulo Daniel's suggestion

da-kami (approved):

bonomat (approved):
👍

PR #570, opened 2021-11-15
Ban use of xtra functions that lead to hard-to-spot bugs

thomaseizinger:
Read commit messages for details.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -622,8 +622,9 @@ where
         tokio::spawn(async move {
             let dlc = contract_future.await;
 
-            this.do_send_async(CfdSetupCompleted { order_id, dlc })
+            this.send(CfdSetupCompleted { order_id, dlc })
                 .await
+                .expect("always connected to ourselves");
         });
 
         // 6. Record that we are in an active contract setup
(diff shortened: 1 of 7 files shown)


da-kami (approved):

thomaseizinger:
Hmm, the integration tests are failing with this change. Weird.

da-kami:
> Hmm, the integration tests are failing with this change. Weird.

It seems that the next order is not received correctly. I think it is the way we wait for the next order initially in the tests.
I was already wondering why we have to await `None` first there - I think we should not have to. (For the cfd list we don't have to wait for an empty vec initially...)

Maybe this change also has something to do with it (but I think the tests are actually failing earlier, on the initial order after sending it out):

```
self.takers
            .send(maker_inc_connections::BroadcastOrder(None))
            .await?;
```

I cannot explain why it behaves like this though 😅

daemon/src/maker_cfd.rs
```diff
@@ -346,17 +349,27 @@ where
...
                 command: TakerCommand::NotifySettlementAccepted { id: order_id },
             })
-            .await?;
+            .await?
+        {
+            Ok(_) => {
+                self.current_agreed_proposals
+                    .insert(order_id, self.get_settlement_proposal(order_id)?);
+                self.remove_pending_proposal(&order_id)
+                    .context("accepted settlement")?;
```
klochowicz, line 364: nit: this probably could be pulled out of this clause, at it is the same for both cases?

klochowicz (approved):
nice! looking forward to testing this more :)

daemon/src/maker_cfd.rs
thomaseizinger: True!

thomaseizinger:
bors retry

PR #573, opened 2021-11-16
Should not be able to trigger close in `PendingOpen`

### LINKED ISSUE #517: Getting stuck in `SettlementProposed`
When should `Close` be available in the UI?
I think it only makes sense once the CFD is actually on chain, because otherwise closing would not have any effect besides wasting tx fees.

At the moment the `Close` button is clickable already in `PendingOpen` - which can lead to getting stuck in `SettlementProposed`, because of an internal error in the taker daemon before actually proposing the settlement.

The `Close` button should only be clickable if we are in Open state. Anything else can lead to very weird side effects.

We should also consider guarding that when posting the action to the daemon, and return an error if the action is not allowed. (This would be a change on the daemon rather than the UI)

scratchscratchscratchy:
fixes https://github.com/itchysats/itchysats/issues/517

Triggering a settlement proposal prior to being `Open` does not make sense.

### DIFF
--- a/taker-frontend/src/components/History.tsx
+++ b/taker-frontend/src/components/History.tsx
@@ -98,9 +98,7 @@ const CfdDetails = ({ cfd }: CfdDetailsProps) => {
     });
 
     const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED
-        || [StateKey.OPEN_COMMITTED, StateKey.OUTGOING_SETTLEMENT_PROPOSAL, StateKey.PENDING_CLOSE].includes(
-            cfd.state.key,
-        );
+        || !(cfd.state.key === StateKey.OPEN);
 
     return (
         <HStack bg={useColorModeValue("gray.100", "gray.700")} rounded={5}>
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1221,6 +1221,10 @@ impl Cfd {
         )
     }
 
+    pub fn is_collaborative_settle_possible(&self) -> bool {
+        matches!(self.state.clone(), CfdState::Open { .. })
+    }
+
     pub fn role(&self) -> Role {
         self.order.origin.into()
     }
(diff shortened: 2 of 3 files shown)


klochowicz (commented):
would you mind also emitting an error from the daemon API if this is undesired behaviour?

scratchscratchscratchy:
Note: I'm actually not 100% that Close should not be allowed in `PendingOpen` - theoretically we should be able to publish the settlement tx together with the lock tx in the same block. But it does not really make sense to do that, so I would not allow it for now.

Unfortunately I did not record what error actually got me stuck in `SettlementProposed` when writing up #517

taker-frontend/src/components/History.tsx
```diff
@@ -97,7 +97,12 @@ const CfdDetails = ({ cfd }: CfdDetailsProps) => {
...
     const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED
-        || [StateKey.OPEN_COMMITTED, StateKey.OUTGOING_SETTLEMENT_PROPOSAL, StateKey.PENDING_CLOSE].includes(
+        || [
+            StateKey.PENDING_OPEN,
+            StateKey.OPEN_COMMITTED,
+            StateKey.OUTGOING_SETTLEMENT_PROPOSAL,
+            StateKey.PENDING_CLOSE,
+        ].includes(
             cfd.state.key,
         );
```
bonomat: Would it be easier to just say: 

```
const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED || !StateKey.OPEN === cfd.state.key
```

bonomat (approved):
👍

bonomat:
> Note: I'm actually not 100% that Close should not be allowed in `PendingOpen` - theoretically we should be able to publish the settlement tx together with the lock tx in the same block. But it does not really make sense to do that, so I would not allow it for now.
> 
> Unfortunately I did not record what error actually got me stuck in `SettlementProposed` when writing up #517

It could happen that the two parties have underpaid the `tx_lock` and then the money is pending. 
Closing in `PendingOpen` is not wrong and could be useful in certain circumstances. I would assume though that traders prefer to have a position open for a longer period than a very short one. 

That being said, I'm fine with not having the `Close` available in `PendingOpen`.

PR #575, opened 2021-11-16
Rollover CFD's automatically when a configurable rollover offset has …

rishflab:
…elapsed

The rollover offset is the number of settlement intervals after which rollover is triggered.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -3,6 +3,7 @@ use crate::db::load_all_cfds;
 use crate::maker_cfd::{FromTaker, NewTakerOnline};
 use crate::model::cfd::{Cfd, Order, UpdateCfdProposals};
 use crate::oracle::Attestation;
+use crate::taker_cfd::AutoRolloverParams;
 use anyhow::Result;
 use connection::ConnectionStatus;
 use maia::secp256k1_zkp::schnorrsig;
@@ -47,6 +48,7 @@ pub mod wallet_sync;
 pub mod wire;
 
 const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
+pub const SETTLEMENT_INTERVAL: time::Duration = time::Duration::hours(24);
 
 pub const N_PAYOUTS: usize = 200;
 
@@ -189,6 +191,7 @@ where
         identity_sk: x25519_dalek::StaticSecret,
         oracle_constructor: impl FnOnce(Vec<Cfd>, Box<dyn StrongMessageChannel<Attestation>>) -> O,
         monitor_constructor: impl FnOnce(Box<dyn StrongMessageChannel<monitor::Event>>, Vec<Cfd>) -> F,
+        roll_over_params: AutoRolloverParams,
         n_payouts: usize,
     ) -> Result<Self>
     where
@@ -208,9 +211,10 @@ where
 
         let (monitor_addr, mut monitor_ctx) = xtra::Context::new(None);
         let (oracle_addr, mut oracle_ctx) = xtra::Context::new(None);
-
+        let (cfd_actor_addr, cfd_context) = xtra::Context::new(None);
         let (connection_actor_addr, connection_actor_ctx) = xtra::Context::new(None);
-        let cfd_actor_addr = taker_cfd::Actor::new(
+
+        tokio::spawn(cfd_context.run(taker_cfd::Actor::new(
             db,
             wallet_addr,
             oracle_pk,
@@ -221,9 +225,8 @@ where
             monitor_addr.clone(),
             oracle_addr,
             n_payouts,
-        )
-        .create(None)
-        .spawn_global();
+            roll_over_params,
+        )));
 
         tokio::spawn(connection_actor_ctx.run(connection::Actor::new(
             maker_online_status_feed_sender,
@@ -253,6 +256,10 @@ where
             .spawn_global();
 
         tokio::spawn(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
+
+        cfd_actor_addr
+            .do_send_async(taker_cfd::AutoRollover)
+            .await?;
 
         Ok(Self {
             cfd_actor_addr,
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -2,7 +2,7 @@ use crate::model::{
     BitMexPriceEventId, InversePrice, Leverage, Percent, Position, Price, TakerId, Timestamp,
     TradingPair, Usd,
 };
-use crate::{monitor, oracle, payout_curve};
+use crate::{monitor, oracle, payout_curve, SETTLEMENT_INTERVAL};
 use anyhow::{bail, Context, Result};
 use bdk::bitcoin::secp256k1::{SecretKey, Signature};
 use bdk::bitcoin::{Address, Amount, PublicKey, Script, SignedAmount, Transaction, Txid};
@@ -132,7 +132,7 @@ impl Order {
         max_quantity: Usd,
         origin: Origin,
         oracle_event_id: BitMexPriceEventId,
-        settlement_time_interval_hours: Duration,
+        settlement_time_interval_hours: Duration
     ) -> Result<Self> {
         let leverage = Leverage::new(2)?;
         let liquidation_price = calculate_long_liquidation_price(leverage, price);
@@ -680,20 +680,20 @@ impl Cfd {
         self.order.oracle_event_id.timestamp()
     }
 
-    /// A factor to be added to the CFD order settlement_time_interval_hours for calculating the
+    /// A factor to be added to the CFD order settlement_interval for calculating the
     /// refund timelock.
     ///
     /// The refund timelock is important in case the oracle disappears or never publishes a
     /// signature. Ideally, both users collaboratively settle in the refund scenario. This
     /// factor is important if the users do not settle collaboratively.
-    /// `1.5` times the settlement_time_interval_hours as defined in CFD order should be safe in the
+    /// `1.5` times the settlement_interval as defined in CFD order should be safe in the
     /// extreme case where a user publishes the commit transaction right after the contract was
     /// initialized. In this case, the oracle still has `1.0 *
-    /// cfdorder.settlement_time_interval_hours` time to attest and no one can publish the refund
+    /// cfdorder.settlement_interval` time to attest and no one can publish the refund
     /// transaction.
     /// The downside is that if the oracle disappears: the users would only notice at the end
-    /// of the cfd settlement_time_interval_hours. In this case the users has to wait for another
-    /// `1.5` times of the settlement_time_interval_hours to get his funds back.
+    /// of the cfd settlement_interval. In this case the users has to wait for another
+    /// `1.5` times of the settlement_interval to get his funds back.
     const REFUND_THRESHOLD: f32 = 1.5;
 
     pub const CET_TIMELOCK: u32 = 12;
@@ -1159,6 +1159,11 @@ impl Cfd {
                 | CfdState::Accepted { .. }
                 | CfdState::ContractSetup { .. }
         )
+    }
+
+    pub fn is_elligble_for_rollover(&self, open_until_rollover: Duration) -> bool {
+        let cfd_time_since_open =  OffsetDateTime::now_utc() -(self.expiry_timestamp() - SETTLEMENT_INTERVAL);
+        matches!(self.state, CfdState::Open {attestation: None, ..}) &&  (open_until_rollover >= cfd_time_since_open)
     }
 
     pub fn role(&self) -> Role {
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -1,5 +1,5 @@
 use crate::cfd_actors::{self, append_cfd_state, insert_cfd_and_send_to_feed};
-use crate::db::{insert_order, load_cfd_by_order_id, load_order_by_id};
+use crate::db::{insert_order, load_all_cfds, load_cfd_by_order_id, load_order_by_id};
 use crate::model::cfd::{
     Cfd, CfdState, CfdStateChangeEvent, CfdStateCommon, CollaborativeSettlement, Dlc, Order,
     OrderId, Origin, Role, RollOverProposal, SettlementKind, SettlementProposal, UpdateCfdProposal,
@@ -8,13 +8,14 @@ use crate::model::cfd::{
 use crate::model::{BitMexPriceEventId, Price, Timestamp, Usd};
 use crate::monitor::{self, MonitorParams};
 use crate::wire::{MakerToTaker, RollOverMsg, SetupMsg};
-use crate::{log_error, oracle, setup_contract, wallet, wire};
+use crate::{log_error, oracle, setup_contract, try_continue, wallet, wire};
 use anyhow::{bail, Context as _, Result};
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
 use futures::channel::mpsc;
 use futures::{future, SinkExt};
 use std::collections::HashMap;
+use time::{Duration, OffsetDateTime};
 use tokio::sync::watch;
 use xtra::prelude::*;
 
@@ -27,9 +28,6 @@ pub enum CfdAction {
     ProposeSettlement {
         order_id: OrderId,
         current_price: Price,
-    },
-    ProposeRollOver {
-        order_id: OrderId,
     },
     Commit {
         order_id: OrderId,
@@ -46,6 +44,8 @@ pub struct CfdRollOverCompleted {
     pub dlc: Result<Dlc>,
 }
 
+pub struct AutoRollover;
+
 enum SetupState {
     Active {
         sender: mpsc::UnboundedSender<SetupMsg>,
@@ -58,6 +58,32 @@ enum RollOverState {
         sender: mpsc::UnboundedSender<RollOverMsg>,
     },
     None,
+}
+
+pub struct AutoRolloverParams {
+    /// The maximum time can remain open before it is settled
+    settlement_interval: Duration,
+    /// Defines the duration a CFD needs to be open for until a roll-over is triggered.
+    open_until_rollover: Duration,
+}
+
+impl AutoRolloverParams {
+    pub fn new(open_until_rollover: Duration, settlement_interval: Duration) -> Result<Self> {
+        Ok(Self {
+            settlement_interval,
+            open_until_rollover,
+        })
+    }
+
+    // Two conditions need to be met for a rollover to be triggered.
+    // The cfd should not be open longer than the settlement interval.
+    // The cfd should be open longer than open_until_rollover duration.
+    pub fn is_cfd_due_for_rollover(&self, expiry_timestamp: OffsetDateTime) -> bool {
+        let now = OffsetDateTime::now_utc();
+        let time_until_cfd_settlement = now - (expiry_timestamp - self.roll_over_params.settlement_interval);
+        self.settlement_interval > time_until_cfd_settlement
+            && (self.settlement_interval - time_until_cfd_settlement) >= self.open_until_rollover
+    }
 }
 
 pub struct Actor<O, M, W> {
@@ -73,6 +99,7 @@ pub struct Actor<O, M, W> {
     roll_over_state: RollOverState,
     oracle_actor: Address<O>,
     current_pending_proposals: UpdateCfdProposals,
+    roll_over_params: AutoRolloverParams,
     n_payouts: usize,
 }
 
@@ -94,6 +121,7 @@ where
         monitor_actor: Address<M>,
         oracle_actor: Address<O>,
         n_payouts: usize,
+        roll_over_params: AutoRolloverParams,
     ) -> Self {
         Self {
             db,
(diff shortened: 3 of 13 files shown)


daemon/src/model/cfd.rs
```diff
@@ -680,20 +680,26 @@ impl Cfd {
         self.order.oracle_event_id.timestamp()
     }
 
-    /// A factor to be added to the CFD order settlement_time_interval_hours for calculating the
+    pub fn creation_timestamp(&self) -> Result<OffsetDateTime> {
+        let timestamp =
+            OffsetDateTime::from_unix_timestamp(self.order.creation_timestamp.seconds())?;
+        Ok(timestamp)
+    }
```
thomaseizinger: I think this might be dead code now?

daemon/src/lib.rs
```diff
@@ -254,6 +256,12 @@ where
 
         tokio::spawn(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
 
+        // TODO: Optimize initial auto-rollover to take time into account so that we don't want to
+        // trigger auto-rollover after an immediate restart
+        cfd_actor_addr
```
bonomat: Why should we not trigger it immediately after start?

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,21 @@ enum RollOverState {
     None,
 }
 
+pub struct RollOverParams {
+    settlement_interval: Duration,
+    /// The time since cfd creation where rollover is triggered
```
bonomat: I think this doc is outdated.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+        for cfd in cfds
+            .iter()
+            .filter(|cfd| matches!(cfd.state, CfdState::Open { .. }))
+        {
+            let now = OffsetDateTime::now_utc();
+
+            let elapsed_since_creation =
+                now - (cfd.expiry_timestamp() - self.roll_over_params.settlement_interval);
+
+            // Two conditions need to be bet for a roll over to be triggered.
```
bonomat:
```suggestion
            // Two conditions need to be met for a roll over to be triggered.
```

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+            // The cfd should not of run longer that the settlement interval.
```
bonomat:
```suggestion
            // The cfd should not run longer that the settlement interval.
```

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,21 @@ enum RollOverState {
     None,
 }
 
+pub struct RollOverParams {
```
thomaseizinger: I think these concern themselves with our "auto" roll-over feature, not roll-over per se.

cc @da-kami

```suggestion
pub struct AutoRollOverParams {
```

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,21 @@ enum RollOverState {
...
+    settlement_interval: Duration,
+    /// The time since cfd creation where rollover is triggered
+    rollover_interval: Duration,
```
thomaseizinger: To me, the docs and variable name here still don't make sense to me.

As far as I understand, this interval defines the minimum time that has to pass for a CFD to be automatically rolled over. How about this:

```suggestion
    /// Defines the minimum duration a CFD needs to be open for until a roll-over is triggered.
    min_open_until_rollover: Duration,
```

cc @da-kami @bonomat

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,21 @@ enum RollOverState {
     None,
 }
 
+pub struct RollOverParams {
+    settlement_interval: Duration,
```
thomaseizinger, line 65:
```suggestion
    /// Defines the maximum duration a CFD may be open for to still qualify for automatic roll-over.
    max_open: Duration,
```

cc @da-kami @bonomat

daemon/src/lib.rs
```diff
@@ -85,7 +86,7 @@ where
             Box<dyn MessageChannel<NewTakerOnline>>,
             Box<dyn MessageChannel<FromTaker>>,
         ) -> T,
-        settlement_time_interval_hours: time::Duration,
+        settlement_interval: time::Duration,
```
thomaseizinger: I know the name is ugly but it would be great to minimize the diff of this PR to just the auto-roll over feature. Can you please undo this rename and squash all commits afterwards?

Feel free to send a follow-up PR that cleans up these names.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+        self.current_pending_proposals.clear();
+
+        // Only cfds in state `Open` are allowed for auto-rollover proposal
+        for cfd in cfds
+            .iter()
+            .filter(|cfd| matches!(cfd.state, CfdState::Open { .. }))
+        {
+            let now = OffsetDateTime::now_utc();
+
+            let elapsed_since_creation =
```
thomaseizinger: I think this variable name is now stale.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+        self.send_pending_update_proposals()?;
+
+        self.send_to_maker
+            .send(wire::TakerToMaker::ProposeRollOver {
+                order_id: proposal.order_id,
+                timestamp: proposal.timestamp,
+            })
+            .await?;
+
+        tracing::debug!("proposed roll over");
```
thomaseizinger, line 242: I think this is important enough to warrant an `info` log, plus it would be good to include some context.

```suggestion
        tracing::info!(%order_id, "Proposed roll over");
```

daemon/tests/harness/mocks/mod.rs
```diff
@@ -73,12 +73,26 @@ impl Mocks {
...
+            .expect_monitor_attestation()
+            .return_const(());
+    }
+
+    pub async fn mock_start_monitoring(&mut self) {
+        self.monitor()
+            .await
+            .expect_start_monitoring()
+            .return_const(());
+    }
```
thomaseizinger: I think we should include these in `mock_common_empty_handlers` otherwise the test is super noisy.

daemon/tests/happy_path.rs
```diff
@@ -59,6 +64,150 @@ async fn taker_takes_order_and_maker_rejects() {
...
+        .expect_sign()
+        .times(1)
+        .returning(|sign| Ok(sign.psbt))
+        .in_sequence(&mut seq);
+    wallet
+        .expect_broadcast()
+        .times(1)
+        .returning(|_| Ok(dummy_tx_id()))
+        .in_sequence(&mut seq);
+}
```
thomaseizinger: This is stale, we have a function on `Mocks` that does the same thing.

daemon/tests/happy_path.rs
```diff
@@ -59,6 +64,150 @@ async fn taker_takes_order_and_maker_rejects() {
...
+        .cfd_actor_addr
+        .send(monitor::Event::LockFinality(taker_cfd.order.id))
+        .await
+        .unwrap();
+
+    let (taker_cfd_open_state, maker_cfd_open_state) =
+        next_cfd(&mut taker.cfd_feed, &mut maker.cfd_feed).await;
+
+    assert!(matches!(taker_cfd_open_state.state, CfdState::Open { .. }));
+    assert!(matches!(maker_cfd_open_state.state, CfdState::Open { .. }));
```
thomaseizinger: All of this is essentially setup code, at least from the perspective of this test. Please create a helper like `start_both_in_open` that does all of this minus the assertions (setup code should not contain assertions).

daemon/tests/harness/mod.rs
```diff
@@ -134,6 +147,29 @@ impl Maker {
...
+            .send(CfdAction::AcceptRollOver { order_id: order.id })
+            .await
+            .unwrap()
+            .unwrap();
+    }
+
+    pub async fn get_update_cfd_proposal(
+        &mut self,
+        order_id: &OrderId,
+    ) -> anyhow::Result<UpdateCfdProposal> {
```
thomaseizinger: No need to qualify `Result`.

```suggestion
    ) -> Result<UpdateCfdProposal> {
```

daemon/src/lib.rs
thomaseizinger: For future PRs, can you please quickly scan the entire diff of the PR yourself just before you request a review? :)
These things are easy to spot!

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+            let now = OffsetDateTime::now_utc();
+
+            let elapsed_since_creation =
+                now - (cfd.expiry_timestamp() - self.roll_over_params.settlement_interval);
+
+            // Two conditions need to be bet for a roll over to be triggered.
+            // The cfd should not of run longer that the settlement interval.
+            // The cfd should of lived longer than the rollover interval.
+            if self.roll_over_params.settlement_interval > elapsed_since_creation
+                && elapsed_since_creation < self.roll_over_params.rollover_interval
```
thomaseizinger: You could simplify this by moving this check _onto_ the `RollOverParams` struct. Note how you are accessing self.roll_over_params` twice in here.

This is an instance of _Tell, don't ask_.

thomaseizinger (changes requested):
Thanks, some more comments!

daemon/src/taker_cfd.rs
bonomat: I would prefer `settlement_interval` or something with `*interval` in it.

daemon/src/taker_cfd.rs
bonomat: I find it confusing if I have to think about the expiry from the beginning. 
I would rather think about it from the end, i.e. the CFD should always have a remaining expiry of 24h, if it doesn't, roll over.

daemon/src/taker_cfd.rs
```diff
@@ -167,6 +185,65 @@ impl<O, M, W> Actor<O, M, W> {
...
+        {
+            let now = OffsetDateTime::now_utc();
+
+            let elapsed_since_creation =
+                now - (cfd.expiry_timestamp() - self.roll_over_params.settlement_interval);
+
+            // Two conditions need to be bet for a roll over to be triggered.
+            // The cfd should not of run longer that the settlement interval.
+            // The cfd should of lived longer than the rollover interval.
+            if self.roll_over_params.settlement_interval > elapsed_since_creation
```
bonomat: I find it confusing if I have to think about the time elapsed since the beginning. 
.
I would rather think about it from the end, i.e. the CFD should always have a remaining expiry of 24h, if it doesn't, roll over.
e.g. 
```
if cfd.expiry_timestamp() - now() <= self.roll_over_params.settlement_interval
  rollover()
```

daemon/src/taker_cfd.rs
thomaseizinger: Why interval though? An interval is a duration that happens regularly. This value is not an interval. If you look at the logic how this is used, it only represents a cut-off time after which we no longer trigger roll over.

daemon/src/taker_cfd.rs
thomaseizinger: >  I would rather think about it from the end, i.e. the CFD should always have a remaining expiry of 24h, if it doesn't, roll over.

That is not how it is currently implemented though. I based my suggestion on how things are implemented. If we implement it differently, I am happy to go with a name such as `min_time_to_expiry`.

thomaseizinger: Do we even need this? Shouldn't this always be the same value as `Order#settlement_time_interval_hours`?

daemon/src/lib.rs
rishflab: sorry thats my bad

daemon/src/taker_cfd.rs
rishflab: do you mean that it should be renamed? how about `elapsed_since_open`

thomaseizinger: You are using `cfd.expiry_timestamp()` yeah? So this computation is no longer based on the _creation_ timestamp. Or am I missing something?

daemon/src/lib.rs
```diff
@@ -254,6 +256,12 @@ where
 
         tokio::spawn(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
 
+        // TODO: Optimize initial auto-rollover to take time into account so that we don't want to
```
da-kami: I think this is an old ToDo from when I worked on it that is outdated now that we handle the roll-over based on the time left until the CFD settles. (A previous version of the code did not implement that, that's when the the ToDo was written).

daemon/src/taker_cfd.rs
da-kami: I think in the context of rollover  `max_open` does make more sense than `settlement_interval`. 

---

A few words about different name on the "same" thing in a different context. 

Similar to the oracle's `announcement_lookahead` this variable has to be aligned with the `settlement_interval`. 
Initially we had a `term` that was pulled through the codebase and it was clear what it was, because it was named the same everywhere. That caused confusion in the context of a specific actor, but was easy to understand in "the big picture". Now I feel we have the opposite - in the context of each actor we try to find a name, but the big picture is hard to grasp.

I am not sure we can call it "an implementation detail" that the cfd's `settlement_interval` has to be used for the oracle's `announcement_lookahead` and the rollover's `max_open` - if those values are not aligned weird side effects can happen because we use different time constraints that are supposed to be aligned.

That said, maybe we should take a step back and think about creating a component that encapsulates this knowledge. I don't want to block this PR on this, but this is not the first time we run into this issue.

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
...
+
+impl AutoRolloverParams {
+    pub fn new(rollover_interval: Duration, settlement_interval: Duration) -> Result<Self> {
+        Ok(Self {
+            settlement_interval,
+            max_open_until_rollover: rollover_interval,
+        })
+    }
+
+    // Two conditions need to be bet for a roll over to be triggered.
```
da-kami:
```suggestion
    // Two conditions need to be met for a roll over to be triggered.
```

or can one use `bet` in this context in English?

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
...
+    // The cfd should not of run longer that the settlement interval.
+    // The cfd should of lived longer than the rollover interval.
+    pub fn is_cfd_due_for_rollover(&self, elapsed_since_cfd_opened: Duration) -> bool {
```
da-kami: I would find it more helpful if the complete time handling would be encapsulated in this function. 
i.e. I would pass in `cfd.expiry_timestamp()` as parameter into the function and all the rest should happen inside this function.

I find the current split up of logic is very hard to follow. Maybe that's a personal thing

daemon/tests/happy_path.rs
rishflab: I am not sure how to do this. in order to have setup and assertions seperate I would have to return multiple states from the setup function and assert on them inside the test. I think this might be a bit confusing.

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
...
+        self.settlement_interval > elapsed_since_cfd_opened
```
da-kami: Just to make sure I understand correct.

The `settlement_interval` is a constant set to `24` hours. This allows rollover if that constant is greater than the time elapsed since the CFD was opened. 
So, if `elapsed_since_cfd_opened` is `25` hours we won't roll over.

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
...
+impl AutoRolloverParams {
+    pub fn new(rollover_interval: Duration, settlement_interval: Duration) -> Result<Self> {
+        Ok(Self {
+            settlement_interval,
+            max_open_until_rollover: rollover_interval,
+        })
+    }
+
+    // Two conditions need to be bet for a roll over to be triggered.
+    // The cfd should not of run longer that the settlement interval.
```
da-kami:
```suggestion
    // The cfd should not run longer that the settlement interval.
```

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
...
+    // The cfd should of lived longer than the rollover interval.
```
da-kami: ??? - this sentence is somehow incomplete and does not reflect the (re-)names of variables anymore I feel.

daemon/src/taker_cfd.rs
```diff
@@ -60,6 +60,29 @@ enum RollOverState {
     None,
 }
 
+pub struct AutoRolloverParams {
+    settlement_interval: Duration,
+    /// Defines the minimum duration a CFD needs to be open for until a roll-over is triggered.
+    max_open_until_rollover: Duration,
```
da-kami: I find it confusing that the comment states `minimum time ...` but then the variable is called `max_...` 

This represents the `1 hour` a CFD has to be at least open until we trigger a rollover. Maybe just remove the `max` ? 

Sry, I think you had a discussion with @thomaseizinger on that one before. If you decide to keep the variable name as is it would still be nice to align the comment.

da-kami (changes requested):
My initial work included removing the configurable settlement interval from the maker.

See: https://github.com/itchysats/itchysats/commit/9e9ed8632db2674206804ff6cfb34baa387efb83#L55

Was it a conscious decision to not include this into this PR, or was it just accidentally dropped?

We don't actively use this feature and with introducing logic into the taker, that is unaware of what the maker set, this can cause weird side effects.
I think we should remove the configurable settlement interval as part of this PR and use **one** constant that represents what is called `settlement_interval` for both the maker and taker in the production code.

daemon/tests/happy_path.rs
thomaseizinger: I think it is fine to have setup functions that don't have any assertions. We have a separate test that has detailed assertions things that are meant to happen on the way to an open CFD.

We can have a setup function that assumes that this works and just creates an open CFD. If anything breaks within that part of the code base, all these tests will fail (and probably with different errors). It is then up to the dev to check which test is the one that points out the problem with the most detail and which tests are just a symptom. That should be easy. If something panics in the setup, no need to investigate that test.

daemon/src/taker_cfd.rs
thomaseizinger: I don't think it has to be aligned, has it? `max_open` should be **equal or less** than `settlement_interval`. 

> That caused confusion in the context of a specific actor,

I don't think I was part of this conversation. Can you provide more context?

If a value MUST be the same across multiple parts of the system for it to work then we should absolutely have it as a single value.

daemon/src/taker_cfd.rs
thomaseizinger: My original suggestion called it `min_open_until_rollover` :D

daemon/src/taker_cfd.rs
da-kami: For me this is also easier to follow.

daemon/src/lib.rs
rishflab: I think this is stale now as we check if the `time_until_rollover` has elapsed before triggering rollover. Can you confirm @da-kami

daemon/tests/happy_path.rs
rishflab: out of curiosity what is the problem in leaving the assertions in the setup functions. Wont it make it easier to debug in the case something goes wrong in the future?

daemon/src/taker_cfd.rs
rishflab: ok i will flip it around

daemon/src/taker_cfd.rs
da-kami: > I don't think it has to be aligned, has it? max_open should be equal or less than settlement_interval.

It has to be equal to make sense. It *can* be less for the system to still work, but that is out of the context of what we want to achieve. 

> I don't think I was part of this conversation. Can you provide more context?

Yes, this happened when you were on vacation 😬
Here is the PR:
https://github.com/itchysats/itchysats/pull/424

> If a value MUST be the same across multiple parts of the system for it to work then we should absolutely have it as a single value.

I am not sure the rename was a good idea. The rename is correct in the context of each actor, but the fact that the same value has to be used gets lost. Theoretically it does not have to be *exactly the same* as you pointed out above, but in the context of this system it *should* be. 

@rishflab and me are adding a commit that uses exactly *one* constant to represent the value (i.e. the 24 hours).
I hope that helps. My initial implementation of auto-rollover included that, but somehow it was lost on the way. I should have structured the commits more explicitly.

daemon/src/taker_cfd.rs
rishflab: How would you know when to  trigger the rollover?

daemon/src/taker_cfd.rs
rishflab: yep

PR #589, opened 2021-11-16
Test maker actor system shutdown

klochowicz:
- maker actor system now stops all its async tasks when it gets destroyed
- taker notices absence of maker in a newly written test

Note: I did not include the roll-out on the taker side (it would be done the
same way), as I wanted to first get some feedback from the reviewers.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -3,8 +3,10 @@ use crate::db::load_all_cfds;
 use crate::maker_cfd::{FromTaker, NewTakerOnline};
 use crate::model::cfd::{Cfd, Order, UpdateCfdProposals};
 use crate::oracle::Attestation;
+use crate::tokio_ext::FutureExt;
 use anyhow::Result;
 use connection::ConnectionStatus;
+use futures::future::RemoteHandle;
 use maia::secp256k1_zkp::schnorrsig;
 use sqlx::SqlitePool;
 use std::collections::HashMap;
@@ -12,7 +14,6 @@ use std::future::Future;
 use std::time::Duration;
 use tokio::sync::watch;
 use xtra::message_channel::{MessageChannel, StrongMessageChannel};
-use xtra::spawn::TokioGlobalSpawnExt;
 use xtra::{Actor, Address};
 
 pub mod actors;
@@ -50,12 +51,18 @@ const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
 
 pub const N_PAYOUTS: usize = 200;
 
+/// Struct controlling the lifetime of the async tasks,
+/// such as running actors and periodic notifications.
+/// If it gets dropped, all tasks are cancelled.
+pub struct Tasks(Vec<RemoteHandle<()>>);
+
 pub struct MakerActorSystem<O, M, T, W> {
     pub cfd_actor_addr: Address<maker_cfd::Actor<O, M, T, W>>,
     pub cfd_feed_receiver: watch::Receiver<Vec<Cfd>>,
     pub order_feed_receiver: watch::Receiver<Option<Order>>,
     pub update_cfd_feed_receiver: watch::Receiver<UpdateCfdProposals>,
     pub inc_conn_addr: Address<T>,
+    pub tasks: Tasks,
 }
 
 impl<O, M, T, W> MakerActorSystem<O, M, T, W>
@@ -104,7 +111,9 @@ where
         let (oracle_addr, mut oracle_ctx) = xtra::Context::new(None);
         let (inc_conn_addr, inc_conn_ctx) = xtra::Context::new(None);
 
-        let cfd_actor_addr = maker_cfd::Actor::new(
+        let mut tasks = vec![];
+
+        let (cfd_actor_addr, cfd_actor_fut) = maker_cfd::Actor::new(
             db,
             wallet_addr,
             settlement_time_interval_hours,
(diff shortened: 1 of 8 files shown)


daemon/src/tokio_ext.rs
```diff
@@ -27,3 +28,20 @@ where
         timeout(duration, self)
     }
 }
+
+pub trait SpawnWithHandleExt: Future {
+    /// Spawn the future on a task in the runtime and return a RemoteHandle to it.
+    /// The task will be stopped when the handle gets dropped.
+    fn spawn_with_handle(self) -> RemoteHandle<Self::Output>;
```
thomaseizinger: You can put this on the `FutureExt` trait on line 19.

Creating a dedicated extension trait for each function will only bloat our imports.

daemon/src/lib.rs
```diff
@@ -223,13 +247,15 @@ where
             n_payouts,
         )
         .create(None)
-        .spawn_global();
+        .run();
```
thomaseizinger, line 256: This change seems unrelated?

daemon/src/lib.rs
```diff
@@ -190,6 +213,7 @@ where
         oracle_constructor: impl FnOnce(Vec<Cfd>, Box<dyn StrongMessageChannel<Attestation>>) -> O,
         monitor_constructor: impl FnOnce(Box<dyn StrongMessageChannel<monitor::Event>>, Vec<Cfd>) -> F,
         n_payouts: usize,
+        maker_heartbeat_interval: Duration,
```
thomaseizinger, line 216: Why do we need to make this configurable?

daemon/src/taker.rs
```diff
@@ -168,14 +168,15 @@ async fn main() -> Result<()> {
...
+    let (wallet, wallet_fut) = wallet::Actor::new(
         opts.network.electrum(),
         &data_dir.join("taker_wallet.sqlite"),
         ext_priv_key,
     )
     .await?
     .create(None)
-    .spawn_global();
+    .run();
+    let _wallet_handle = wallet_fut.spawn_with_handle();
```
thomaseizinger, line 179: This is unrelated if you really only wanted to do the maker in this PR :)

daemon/tests/harness/mod.rs
```diff
@@ -168,7 +181,8 @@ impl Taker {
         let (oracle, monitor, wallet) = mocks::create_actors(&mocks);
         mocks.mock_common_empty_handlers().await;
 
-        let wallet_addr = wallet.create(None).spawn_global();
+        let (wallet_addr, wallet_fut) = wallet.create(None).run();
+        tokio::spawn(wallet_fut);
```
thomaseizinger, line 175: So is this.

daemon/tests/harness/mod.rs
```diff
@@ -153,6 +162,10 @@ impl Taker {
         &mut self.system.order_feed_receiver
     }
 
+    pub fn maker_status_feed(&mut self) -> &mut watch::Receiver<ConnectionStatus> {
+        &mut self.system.maker_online_status_feed_receiver
+    }
```
thomaseizinger, line 157: Is this used?

daemon/src/lib.rs
```diff
@@ -46,16 +47,20 @@ pub mod wallet;
 pub mod wallet_sync;
 pub mod wire;
 
-const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
+pub const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
 
 pub const N_PAYOUTS: usize = 200;
 
+/// Struct controlling the lifetime of the actors
```
thomaseizinger: I think this needs more docs. Can you explain a bit here, how this works? :)

thomaseizinger: Also, I think `Tasks` might be a better name because it is not just actors but also notifier tasks.

thomaseizinger (changes requested):
Nice work!

A couple of changes seem to be unrelated, would you mind getting rid of those again?
Also, if you really want to make the heartbeat interval configurable, can we do that as a separate patch?

daemon/tests/harness/mod.rs
thomaseizinger: Seems to belong to the patch after this.

daemon/src/lib.rs
klochowicz: sure! `Tasks` sounds like good name. I'll describe more works.

> Nice work!
> 
> A couple of changes seem to be unrelated, would you mind getting rid of those again? Also, if you really want to make the heartbeat interval configurable, can we do that as a separate patch?



> Nice work!
> 
> A couple of changes seem to be unrelated, would you mind getting rid of those again? Also, if you really want to make the heartbeat interval configurable, can we do that as a separate patch?

Sure, I'll remove heartbeat interval changes from this one (it's not that related, it was only a test optimisation, and can go in later / if at all)

daemon/tests/harness/mod.rs
klochowicz: moved into the correct patch.

klochowicz:
> Nice work!
> 
> A couple of changes seem to be unrelated, would you mind getting rid of those again? Also, if you really want to make the heartbeat interval configurable, can we do that as a separate patch?

Removed the heartbeat, rolled out the same changes in the taker side. My apologies for no `fixup`s, but it was a huge change in total ;D

daemon/tests/harness/mod.rs
```diff
@@ -64,7 +63,8 @@ impl Maker {
         let (oracle, monitor, wallet) = mocks::create_actors(&mocks);
         mocks.mock_common_empty_handlers().await;
 
-        let wallet_addr = wallet.create(None).spawn_global();
+        let (wallet_addr, wallet_fut) = wallet.create(None).run();
+        tokio::spawn(wallet_fut);
```
thomaseizinger, line 67: Not a blocker but it would be good to always strive for a minimal diff. This change doesn't add anything as far as I can tell?

daemon/tests/happy_path.rs
```diff
@@ -1,9 +1,13 @@
...
+use std::time::Duration;
+
+use crate::harness::flow::{is_next_none, next, next_cfd, next_order, next_some};
 use crate::harness::{assert_is_same_order, dummy_new_order, init_tracing, start_both};
+use daemon::connection::ConnectionStatus;
 use daemon::model::cfd::CfdState;
 use daemon::model::Usd;
 use maia::secp256k1_zkp::schnorrsig;
 use rust_decimal_macros::dec;
+use tokio::time::sleep;
```
thomaseizinger: I'd prefer to keep all `use` statements in a single block.

thomaseizinger (approved):
Nice work!

daemon/src/lib.rs
```diff
@@ -50,12 +51,18 @@ const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
...
+/// If it gets dropped, all tasks are cancelled.
+pub struct Tasks(Vec<RemoteHandle<()>>);
+
 pub struct MakerActorSystem<O, M, T, W> {
     pub cfd_actor_addr: Address<maker_cfd::Actor<O, M, T, W>>,
     pub cfd_feed_receiver: watch::Receiver<Vec<Cfd>>,
     pub order_feed_receiver: watch::Receiver<Option<Order>>,
     pub update_cfd_feed_receiver: watch::Receiver<UpdateCfdProposals>,
     pub inc_conn_addr: Address<T>,
+    pub tasks_handle: Tasks,
```
thomaseizinger: Just `tasks` perhaps?

thomaseizinger: Plus make it private? Should not be modified I guess.

thomaseizinger:
Should we ban `tokio::spawn` to avoid creating zombie tasks that are not stopped when the actor system gets dropped?

daemon/tests/happy_path.rs
klochowicz: I'll try to keep an eye on it (or maybe write a lint against it), this is unintentional outcome of running a code action in my emacs ;/
fixed!

daemon/tests/harness/mod.rs
klochowicz: it's a preparation to spot the stray `tokio::spawn` calls - I will need to ban `spawn_global()` as well. 
I'll keep this change as it's minimal if you're ok with it

daemon/src/lib.rs
klochowicz: renamed. 

 tried to make it private in the very beginning, but it was getting dropped if I didn't destructure it properly and keep it in scope.

An alternative would be manage the whole actor system sin Rocket, but we're not doing right now (should we?). Then it could be private, as it should be!

That's the reason I made the newtype, which at least makes the contents of `Tasks` private, so you can't add/remove/access the internals.

daemon/tests/harness/mod.rs
thomaseizinger: We can disable the tokio feature for `xtra` then we don't have `spawn_global` available :)

klochowicz:
> Should we ban `tokio::spawn` to avoid creating zombie tasks that are not stopped when the actor system gets dropped?

I contemplated that as well! I'll create a separate PR for that.

daemon/src/lib.rs
thomaseizinger: > tried to make it private in the very beginning, but it was getting dropped if I didn't destructure it properly and keep it in scope.

Ah fair enough.

> An alternative would be manage the whole actor system sin Rocket, but we're not doing right now (should we?). Then it could be private, as it should be!

Yeah that might be worth it at some point. Could even consider making all fields private and only expose what we need.

thomaseizinger (approved):

daemon/src/lib.rs
klochowicz: I'll put it on my list to deal with at some point.

PR #601, opened 2021-11-16
Use the taker's public key for `TakerId`

thomaseizinger:


daemon/src/model.rs
```diff
@@ -383,18 +383,41 @@ pub enum Position {
...
+        D: serde::Deserializer<'de>,
+    {
+        let hex = String::deserialize(deserializer)?;
+
+        let mut bytes = [0u8; 32];
+        hex::decode_to_slice(&hex, &mut bytes).map_err(D::Error::custom)?;
+
+        Ok(Self(x25519_dalek::PublicKey::from(bytes)))
     }
 }
```
klochowicz, line 416: perhaps add a unit test to cover our custom serialisation/deserialisation?

klochowicz (approved):
great work!

daemon/src/model.rs
```diff
@@ -655,4 +655,16 @@ mod tests {
 
         assert_eq!(ts_b.seconds() - ts_a.seconds(), -36000);
     }
+
+    #[test]
```
klochowicz, line 659: nice! do we have a rule of thumb when we should move things into a test module? I'm happy for it to stay as it is, as it seems to be a trivial test that won't bloat lib size or compile time, asking just out of curiosity.

klochowicz (approved):
thank you for adding the test 👍

daemon/src/model.rs
thomaseizinger: > that won't bloat lib size or compile time

That is why this is behind `cfg(test)` (line 548)! `cfd` is conditional compilation in Rust so non of this gets compiled unless you want to run the tests.

bonomat (approved):
Nice, can we store the key in the CFD in the db so that we know who we had this CFD with?

thomaseizinger:
> Nice, can we store the key in the CFD in the db so that we know who we had this CFD with?

Of course we can. But that should come with actually using the key to authorize changes to a CFD which is a much more invasive change that is best left to a follow-up PR.

bonomat:
> > Nice, can we store the key in the CFD in the db so that we know who we had this CFD with?
> 
> Of course we can. But that should come with actually using the key to authorize changes to a CFD which is a much more invasive change that is best left to a follow-up PR.

sounds good, I'll create a ticket for that.

daemon/src/model.rs
klochowicz: oops, missed that one! 😅

PR #602, opened 2021-11-16
Bug hunting cleanups

thomaseizinger:
Some changes I made whilst bug hunting.

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -11,14 +11,14 @@ async fn taker_receives_order_from_maker_on_publication() {
     let _guard = init_tracing();
     let (mut maker, mut taker) = start_both().await;
 
-    assert!(is_next_none(taker.order_feed()).await);
+    assert!(is_next_none(taker.order_feed()).await.unwrap());
 
     maker.publish_order(dummy_new_order()).await;
 
     let (published, received) =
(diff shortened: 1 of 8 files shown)


daemon/tests/happy_path.rs
```diff
@@ -11,14 +11,14 @@ async fn taker_receives_order_from_maker_on_publication() {
     let _guard = init_tracing();
     let (mut maker, mut taker) = start_both().await;
 
-    assert!(is_next_none(taker.order_feed()).await);
+    assert!(is_next_none(taker.order_feed()).await.unwrap());
```
klochowicz, line 14: IMHO these `unwrap()`s bubbling out in the test just make it read worse - is it really an improvement? 

Do they aid in debugging, giving a better line number in case of an error? (I ran into issues with that too, if that's the case, then it trumps a small decrease in readability :) )

klochowicz (approved):
Nice cleanup

daemon/tests/happy_path.rs
thomaseizinger: > Do they aid in debugging, giving a better line number in case of an error? (I ran into issues with that too, if that's the case, then it trumps a small decrease in readability :) )

Yes, see the commit message of that commit. `unwrap` shows the line number where it got called (because it has the `#[track_caller]` attribute: https://doc.rust-lang.org/src/core/result.rs.html#1294).

PR #607, opened 2021-11-17
Don't use tokio::spawn

klochowicz:
see commit descriptions for more info

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -45,7 +45,7 @@ tracing = { version = "0.1" }
 tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "ansi", "env-filter", "chrono", "tracing-log", "json"] }
 uuid = { version = "0.8", features = ["serde", "v4"] }
 x25519-dalek = { version = "1.1" }
-xtra = { version = "0.6", features = ["with-tokio-1"] }
+xtra = { version = "0.6" }
 xtra_productivity = { version = "0.1.0" }
 
 [[bin]]
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -15,7 +15,7 @@ use xtra_productivity::xtra_productivity;
 
 struct ConnectedState {
     last_heartbeat: SystemTime,
-    _pulse_handle: RemoteHandle<()>,
+    _tasks: Vec<RemoteHandle<()>>,
 }
 
 pub struct Actor {
@@ -97,22 +97,26 @@ impl Actor {
 
         let send_to_socket = send_to_socket::Actor::new(write, noise.clone());
 
-        tokio::spawn(self.send_to_maker_ctx.attach(send_to_socket));
+        let mut tasks = vec![self
+            .send_to_maker_ctx
+            .attach(send_to_socket)
+            .spawn_with_handle()];
 
         let read = FramedRead::new(read, wire::EncryptedJsonCodec::new(noise))
             .map(move |item| MakerStreamMessage { item });
 
         let this = ctx.address().expect("self to be alive");
-        tokio::spawn(this.attach_stream(read));
+        tasks.push(this.attach_stream(read).spawn_with_handle());
 
-        let pulse_remote_handle = ctx
-            .notify_interval(self.timeout, || MeasurePulse)
-            .expect("we just started")
-            .spawn_with_handle();
+        tasks.push(
+            ctx.notify_interval(self.timeout, || MeasurePulse)
+                .expect("we just started")
+                .spawn_with_handle(),
+        );
 
         self.connected_state = Some(ConnectedState {
             last_heartbeat: SystemTime::now(),
-            _pulse_handle: pulse_remote_handle,
+            _tasks: tasks,
         });
         self.status_sender
             .send(ConnectionStatus::Online)
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -1,7 +1,5 @@
-use crate::tokio_ext::FutureExt;
-use crate::{log_error, noise, send_to_socket, wire};
+use crate::{log_error, noise, send_to_socket, wire, Tasks};
 use anyhow::Result;
-use futures::future::RemoteHandle;
 use futures::StreamExt;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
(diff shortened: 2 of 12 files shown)


daemon/src/connection.rs
```diff
@@ -15,7 +15,7 @@ use xtra_productivity::xtra_productivity;
 
 struct ConnectedState {
     last_heartbeat: SystemTime,
-    _pulse_handle: RemoteHandle<()>,
+    tasks: Vec<RemoteHandle<()>>,
```
thomaseizinger: What do you think of using the `Tasks` struct in more places and extend it with an API like:

```rust
impl Tasks {
	pub fn add(&mut self, f: impl Future) {
		let handle = f.spawn_with_handle();
		self.0.push(handle);
	}
}
```

Then we could do things like:

```rust
tasks.add(self.send_to_maker_ctx.attach(send_to_socket));
```

Which would be a bit shorter.

daemon/src/maker_cfd.rs
```diff
@@ -765,18 +770,19 @@ where
             .address()
             .expect("actor to be able to give address to itself");
 
+        let task = async move {
+            let dlc = contract_future.await;
+
+            log_error!(this.do_send_async(CfdRollOverCompleted { order_id, dlc }));
```
thomaseizinger: I don't think we should `log_error` here, this only ever fails with `Disconnected` which is unlikely (impossible?) because we are sending the message to ourselves.

daemon/src/maker_cfd.rs
```diff
@@ -603,17 +607,18 @@ where
...
-        tokio::spawn(async move {
+        let task = async move {
             let dlc = contract_future.await;
 
-            this.do_send_async(CfdSetupCompleted { order_id, dlc })
-                .await
-        });
+            log_error!(this.do_send_async(CfdSetupCompleted { order_id, dlc }))
+        }
+        .spawn_with_handle();
```
thomaseizinger: I think this is unnecessary, this task will end by itself sooner or later. I think the idea of using `spawn_with_handle` should only be applied to tasks that contain endless loops.

thomaseizinger (commented):
Nice, thank you!

Some ideas to mull over.

daemon/src/connection.rs
klochowicz: great idea!

daemon/src/maker_cfd.rs
klochowicz: I agree that main use-case is dealing with tasks containing endless loops, however in this case cleanup during shutdown could deal with it in a nice way IMHO (especially that we already have a state for in-flight setup, we're making the code clearer this way for a reader)

daemon/src/maker_cfd.rs
klochowicz: it was possible to get this one if the actor got shut down whilst doing contract setup (which takes ~1 min). With the guard that I've added here, it shouldn't ever happen, so I'm happy to do an `.expect("actor to be connected")` here. 

I'm not fussed, I just didn't want to swallow the error silently, and I also didn't want to open the doors for anything else than `RemoteHandle<()>`

klochowicz:
I've rebased on top of latest master, no functional changes (except catching one unneeded `spawn_with_handle` before `tasks.add`)

daemon/Cargo.toml
```diff
@@ -45,7 +45,7 @@ tracing = { version = "0.1" }
 tracing-subscriber = { version = "0.2", default-features = false, features = ["fmt", "ansi", "env-filter", "chrono", "tracing-log", "json"] }
 uuid = { version = "0.8", features = ["serde", "v4"] }
 x25519-dalek = { version = "1.1" }
-xtra = { version = "0.6", features = ["with-tokio-1"] }
+xtra = { version = "0.6" }
```
thomaseizinger, line 48: I believe this patch by itself will fail to compile yes? It should probably come after you've migrated all usages of `spawn_global`.

thomaseizinger (changes requested):
Commit message of 7723cb2174db1dd3fe87cf30df92b932cb0bc203 is a bit odd, is that on purpose?

One nit regard the order of the commit messages!

daemon/src/maker_cfd.rs
klochowicz: used `expect()` (as the master branch does now too)

daemon/src/connection.rs
klochowicz: Added the API, guarded by a `debug_assert!` to prevent calling `spawn_with_handle()` twice (as it makes no sense). Guard is covered with a test, and it was helpful when refactoring the code (I missed 2!)

daemon/Cargo.toml
klochowicz: It didn't fail to compile. We were not using `spawn_global` before this changeset. I moved it after as it didn't matter.

klochowicz:
> Commit message of [7723cb2](https://github.com/itchysats/itchysats/commit/7723cb2174db1dd3fe87cf30df92b932cb0bc203) is a bit odd, is that on purpose?
> 
> One nit regard the order of the commit messages!

it is, indeed! I forgot to squash it inside the other commit, squashed now ;)

daemon/src/tokio_ext.rs
```diff
@@ -39,12 +40,46 @@ where
...
+    async fn panics_when_called_spawn_with_handle_on_remote_handle() {
+        let result = panic::catch_unwind(|| {
+            let handle = sleep(Duration::from_secs(2)).spawn_with_handle();
+            let _handle_to_a_handle = handle.spawn_with_handle();
+        });
+        assert!(
+            result.is_err(),
+            "Spawning a remote handle into a separate task should panic"
+        );
+    }
```
thomaseizinger: This test will fail if we build the tests with `--release`. I think that is fine for now but perhaps we should just replace the `debug_assert` with an actual assert?

thomaseizinger (approved):

daemon/src/tokio_ext.rs
klochowicz: thanks for pointing it out! I've changed the test to only panic in debug:

```rust
+        if cfg!(debug_assertions) {
+            assert!(
+                result.is_err(),
+                "Spawning a remote handle into a separate task should panic_in_debug_mode"
+            );
+        } else {
+            assert!(result.is_ok(), "Do not panic in release mode");
+        }
```

Issue #572, opened 2021-11-16
Encorporate UI projections into actor system


thomaseizinger:
Currently, the projections necessary for the UI are happening outside of the actor system in the "rocket layer" via multiple `tokio::sync::watch` channels and the `ToSseEvent` trait. This has multiple downsides:

1. It is not tested as part of our integration test because our tests start at a different layer. We don't want to change the layer we are testing at.
2. It introduces complexity in the actors because they need to send updates on various "watch" channels.

To fix these problems, we want to introduce a dedicated _projection_ actor which takes care about applying the projection for the UI. The goal is manifold:

1. Reduce the logic happening in the rocket layer. `ToSseEvent` can likely go away as part of this.
2. Reduce the complexity for other actors in what needs to be updated when. All they should care about is sending updates to the projection actor.
3. Improve test coverage. With a dedicated actor that does the projection, we should be able to write assertions in our integration tests that are closer to the UI.

bonomat:
Reopening: #645 set the ground work but did not solve this ticket yet.

klochowicz:
This has been addressed in the PRs:

#697  #709 #708  #719

PR #624, opened 2021-11-18
New http api for maker to retrieve CFDs

bonomat:
Adding a new HTTP GET endpoint is useful for a service built on top of the maker. 
Using SSE we need to continuously listen and wait for things to change while an HTTP GET can simply be triggered if needed. 

The second commit moves all sse structs into an own file to remove duplication.   
Only change in there is making some fields public so that they can be accessed from `to_sse_events`


thomaseizinger (commented):
I am happy to add such an endpoint.

The current vision for the rocket layer is to include the projections in the actor system. See https://github.com/itchysats/itchysats/issues/572.

If we would have this projection actor already, retrieving the current list of CFDs as shown for the UI would be a simple message to that projection actor.

Would you be willing to work towards said projection actor and implement this endpoint on top of it?

bonomat:
closing this for now as it will end up in terrible merge conflicts after #645 was merged

PR #625, opened 2021-11-18
Disable go-long button if

### LINKED ISSUE #522: Being able to trigger `Buy` should be more restricted
Currently the user can click `Buy` unconditionally as soon as an order is present.

We should at least validate:

- The balance is enough to fulfill the margin

I propose to just add this condition to the `order != null` condition for now. Additionally a tooltip that shows what is the problem. More creative: Display link to wallet under button if the balance is not enough (...).

Going down the route of adding proper validation by using a form and e.g. `formik` is a bit of a rabbit hole, so I would postpone that to later.

### LINKED ISSUE #525: No indication and validation in the taker about valid order range
the new UI lost the ability to communicate about the available offers from the maker.

I tried ordering $103 from the maker, and it failed during contract setup with a really confusing error:
`Failed to setup contract with maker: CET signatures don't verify: no enc sig from other party for price range 55360..=55363`

bonomat:
- balance < required margin
- entered quantity > max_quantity or < min_quantity
- entered quantity <= 0

resolves #525
resolves  #522

Only a cosmetic change. Should we duplicate this logic into the controller? Let me know and I'll add it and convert this PR to `ready for review`

### DIFF
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -1,5 +1,9 @@
 import { BoxProps } from "@chakra-ui/layout";
 import {
+    Alert,
+    AlertDescription,
+    AlertIcon,
+    AlertTitle,
     Box,
     Button,
     ButtonGroup,
@@ -56,10 +60,23 @@ interface TradeProps {
     leverage?: number;
     quantity: string;
     liquidationPrice?: number;
-    canSubmit: boolean;
     isSubmitting: boolean;
     onQuantityChange: any;
+    walletBalance?: number;
     onLongSubmit: (payload: CfdOrderRequestPayload) => void;
+}
+
+interface AlertBoxProps {
+    title: string;
+    description: string;
+}
+
+function AlertBox({ title, description }: AlertBoxProps) {
+    return (<Alert status="error">
+        <AlertIcon />
+        <AlertTitle mr={2}>{title}</AlertTitle>
+        <AlertDescription>{description}</AlertDescription>
+    </Alert>);
 }
 
 const Trade = (
@@ -73,9 +90,9 @@ const Trade = (
         margin: marginAsNumber,
         leverage,
         liquidationPrice: liquidationPriceAsNumber,
-        canSubmit,
         onLongSubmit,
         orderId,
+        walletBalance,
     }: TradeProps,
 ) => {
     let outerCircleBg = useColorModeValue("gray.100", "gray.700");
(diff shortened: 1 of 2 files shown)


klochowicz:
nice improvement! 
Would you mind adding some way of communicating the user possible way of going back to the happy path? (ie. why they can't move on with the order).
As @da-kami pointed out somewhere, a tooltip would be probably the simplest way of telling the user why this button is disabled (it's not perfect as it's not easily discoverable and would not work on mobile, but it's better than nothing).

IMHO for fully solving the linked issue we should also clearly state the available quantity in the order, and also communicate when there's no order (old UI did these 2 things) - I might add it if I get a rush of blood :) Second-guessing is tiresome to the user, if they could just pick a quantity that it's valid in the first place.

bonomat:
image

image

image

taker-frontend/src/components/Trade.tsx
```diff
@@ -93,14 +97,55 @@ const Trade = (
...
+        && !quantityTooHigh && !quantityTooLow && quantityGreaterZero;
+
+    let alertBox;
+
+    if (balanceTooLow) {
+        alertBox = <Alert status="error">
+            <AlertIcon />
+            <AlertTitle mr={2}>Your balance is too low!</AlertTitle>
+            <AlertDescription>Pleas deposit more into you wallet.</AlertDescription>
+        </Alert>;
```
klochowicz: nit: it might be worthwhile to extract this as a separate component (to ensure consistent styling and ease readability) - then `title` and `description` could be passed in as props.
(this can be always done later)

klochowicz (approved):
nice! this UI feedback should improve the UX quite a bit :)

taker-frontend/src/components/Trade.tsx
```diff
@@ -101,6 +105,47 @@ const Trade = (
...
+            <AlertIcon />
+            <AlertTitle mr={2}>Quantity too low!</AlertTitle>
+            <AlertDescription>Min quantity is ${minQuantity}.</AlertDescription>
+        </Alert>;
+    }
+    if (!orderId) {
+        alertBox = <Alert status="error">
+            <AlertIcon />
+            <AlertTitle mr={2}>No liquidity!</AlertTitle>
+            <AlertDescription>The maker you are connected to did not create any offers.</AlertDescription>
```
klochowicz: nit: "has not created"? Present perfect tense might be slightly better

klochowicz (approved):

taker-frontend/src/components/Trade.tsx
bonomat: done :)

taker-frontend/src/components/Trade.tsx
bonomat: done :)

PR #627, opened 2021-11-18
Another auto rollover

da-kami:
After discussion with @thomaseizinger we decided to put the rollover logic on the cfd and unit-test it. 

Auto-rollover actually only depends on the cfd expiry. We hardcode the roll-over offset as `1 hour` for the moment. The unit-tests should cover all bounds and should help understand the logic. (If not please scream :)
Auto-rollover itself is not tested as part of the actor tests. The unit tests should cover the logic. The rest is just wiring things together and should not require testing.

We are adding an actor test for `ProposeRollover` (to test the logic related to rollover itself - accept, reject...) in a follow up PR. This test will be triggered through sending the `ProposeRollover` message to the actor in the tests. 
In the production code we don't send this message to the cfd actor from the outside (it is triggered internally upon auto-rollover).

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -39,7 +39,7 @@ pub async fn insert_order(order: &Order, conn: &mut PoolConnection<Sqlite>) -> a
     .bind(order.leverage.get())
     .bind(&order.liquidation_price.to_string())
     .bind(&order.creation_timestamp.seconds())
-    .bind(&order.settlement_time_interval_hours.whole_seconds())
+    .bind(&order.settlement_interval.whole_seconds())
     .bind(&order.origin)
     .bind(&order.oracle_event_id.to_string())
     .execute(conn)
@@ -90,7 +90,7 @@ pub async fn load_order_by_id(
         leverage: row.leverage,
         liquidation_price: row.liquidation_price.parse()?,
         creation_timestamp: row.ts_secs,
-        settlement_time_interval_hours: Duration::new(row.settlement_time_interval_secs, 0),
+        settlement_interval: Duration::new(row.settlement_time_interval_secs, 0),
         origin: row.origin,
         oracle_event_id: row.oracle_event_id.parse()?,
     })
@@ -300,7 +300,7 @@ pub async fn load_cfd_by_order_id(
         leverage: row.leverage,
         liquidation_price: row.liquidation_price.parse()?,
         creation_timestamp: row.ts_secs,
-        settlement_time_interval_hours: Duration::new(row.settlement_time_interval_secs, 0),
+        settlement_interval: Duration::new(row.settlement_time_interval_secs, 0),
         origin: row.origin,
         oracle_event_id: row.oracle_event_id.parse()?,
     };
@@ -398,7 +398,7 @@ pub async fn load_all_cfds(conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<
                 leverage: row.leverage,
                 liquidation_price: row.liquidation_price.parse()?,
                 creation_timestamp: row.ts_secs,
-                settlement_time_interval_hours: Duration::new(row.settlement_time_interval_secs, 0),
+                settlement_interval: Duration::new(row.settlement_time_interval_secs, 0),
                 origin: row.origin,
                 oracle_event_id: row.oracle_event_id.parse()?,
             };
@@ -504,7 +504,7 @@ pub async fn load_cfds_by_oracle_event_id(
                 leverage: row.leverage,
                 liquidation_price: row.liquidation_price.parse()?,
                 creation_timestamp: row.ts_secs,
-                settlement_time_interval_hours: Duration::new(row.settlement_time_interval_secs, 0),
+                settlement_interval: Duration::new(row.settlement_time_interval_secs, 0),
                 origin: row.origin,
                 oracle_event_id: row.oracle_event_id.parse()?,
             };
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -541,7 +541,7 @@ mod tests {
     async fn test_insert_and_load_order() {
         let mut conn = setup_test_db().await;
 
-        let order = Order::dummy().insert(&mut conn).await;
+        let order = Order::dummy_db().insert(&mut conn).await;
         let loaded = load_order_by_id(order.id, &mut conn).await.unwrap();
 
         assert_eq!(order, loaded);
@@ -551,7 +551,7 @@ mod tests {
     async fn test_insert_and_load_cfd() {
         let mut conn = setup_test_db().await;
 
-        let cfd = Cfd::dummy().insert(&mut conn).await;
+        let cfd = Cfd::dummy_db().insert(&mut conn).await;
         let loaded = load_all_cfds(&mut conn).await.unwrap();
 
         assert_eq!(vec![cfd], loaded);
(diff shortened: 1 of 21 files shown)


da-kami:
After discussion with @thomaseizinger we decided to put the rollover logic on the cfd and unit-test it. 

Auto-rollover itself is not tested as part of the actor tests. The unit tests should cover the logic. The rest is just wiring things together and should not require testing.

We are adding a test for `ProposeRollover` (to test the logic related to rollover itself - accept, reject...) in a follow up PR though. This test will be triggered through sending the `ProposeRollover` message to the actor in the tests. 
In the production code we don't send this message to the cfd actor from the outside (it is triggered internally upon auto-rollover).

daemon/src/taker_cfd.rs
```diff
@@ -326,6 +329,51 @@ impl<O, M, W> Actor<O, M, W> {
...
+
+impl<O: 'static, M: 'static, W: 'static> Actor<O, M, W> {
+    async fn handle_auto_rollover(&mut self, ctx: &mut Context<Self>) -> Result<()> {
+        tracing::debug!("Auto rollover");
+
+        let mut conn = self.db.acquire().await?;
+        let cfds = load_all_cfds(&mut conn).await?;
+
+        // cleanup all pending proposals because auto-rollover will create a new one
+        self.current_pending_proposals.clear();
```
da-kami, line 366: Note for tomorrow Daniel: Is this actually needed given that we filter?

daemon/src/model/cfd.rs
```diff
@@ -1274,6 +1274,31 @@ impl Cfd {
...
     }
+
+    /// Only cfds in state `Open` that have not received an attestation and are within 23 hours
+    /// until expiry are eligible for rollover
+    pub fn rollover_proposal(&self) -> Option<RollOverProposal> {
+        let now = OffsetDateTime::now_utc();
+        let remaining_till_expiry = self.expiry_timestamp() - now;
+        let rollover_window = Duration::hours(1);
+
+        if matches!(
```
bonomat: My personal preference is to do an early return instead of if/else, e.g. 

```
if !matches!(self.state.clone(),
            CfdState::Open {
                attestation: None,
                ..
            })
{
return None;
}
```

bonomat (approved):
Good work 👍

daemon/src/lib.rs
```diff
@@ -52,6 +52,8 @@ pub const HEARTBEAT_INTERVAL: std::time::Duration = Duration::from_secs(5);
 
 pub const N_PAYOUTS: usize = 200;
 
+pub const SETTLEMENT_INTERVAL: time::Duration = time::Duration::hours(24);
```
thomaseizinger, line 65: Some docs would be nice on this!

daemon/src/db.rs
```diff
@@ -760,15 +760,10 @@ mod tests {
...
-
-        fn with_event_id(mut self, id: BitMexPriceEventId) -> Self {
-            self.order.oracle_event_id = id;
-            self
-        }
     }
 
     impl Order {
-        fn dummy() -> Self {
+        fn dummy_db() -> Self {
```
thomaseizinger, line 766: I don't understand the value of this rename?

daemon/src/model/cfd.rs
thomaseizinger: Second this!

daemon/src/lib.rs
```diff
@@ -298,6 +303,10 @@ where
 
         tasks.add(oracle_ctx.run(oracle_constructor(cfds, Box::new(fan_out_actor))));
 
+        cfd_actor_addr
+            .do_send_async(taker_cfd::AutoRollover)
```
thomaseizinger: Why is this usage not flagged by clippy :cry: 

Are you rebased onto latest master?

thomaseizinger: Needs to use `send`.

daemon/src/taker_cfd.rs
```diff
@@ -351,7 +353,7 @@ impl<O, M, W> Actor<O, M, W> {
 
 impl<O: 'static, M: 'static, W: 'static> Actor<O, M, W> {
     async fn handle_auto_rollover(&mut self, ctx: &mut Context<Self>) -> Result<()> {
-        tracing::debug!("Auto rollover");
+        tracing::trace!("Auto rollover message received");
```
thomaseizinger, line 360: I _think_ we have a general `trace` log already for message that we receive.

daemon/src/db.rs
da-kami: The problem is, that the impl blocks in the tests are accessible over all tests. So, if we have a `dummy` function on `Order` / `Cfd` then that dummy will be accessible from any test.
To make it clearer which dummy is used where I started being more explicit about the dummies. Keep the `dummy_...` function private (because that is specific to a test block and should not be re-used across tests because it could cause weird side effects if we change the dummy...) and be more explicit about the name.

We don't have to do this rename, but I would make it a convention to add where the dummy is used.

daemon/src/lib.rs
da-kami: When I wrote this code we still had `do_send_async` `Sync` messages in `lib.rs` - I thought this was a concious decision to not block the startup. 
We did rebase on `master` yesterday, but the removal of `do_send_async` + clippy rule landed later. I'll fix it :)

daemon/src/taker_cfd.rs
da-kami: I'll rebase on latest master and see. I just ran it and did not see a log for these messages besides the logs I added.

daemon/src/model/cfd.rs
```diff
@@ -1666,179 +1869,196 @@ mod tests {
...
+        let cfd = Cfd::dummy_open().with_event_id(BitMexPriceEventId::with_20_digits(
+            now + Duration::hours(23) + Duration::seconds(59),
+        ));
 
-    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
-    pub maker_lock_amount: Amount,
-    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
-    pub taker_lock_amount: Amount,
+        assert!(cfd.rollover_proposal().is_some());
+    }
```
thomaseizinger: I think this test should be `none`.

daemon/src/lib.rs
da-kami: done! let me know if that's what you had in mind

daemon/src/model/cfd.rs
da-kami: this changed completely, I think you will like the change :)

daemon/src/taker_cfd.rs
da-kami: Keeping this statement here means that we clean up rollovers every 5 minutes. I think this is fine. We could also just do it in case we have a proposal. I don't think it should do any harm to do this here - what do you think @thomaseizinger ?

da-kami:
somehow the logic is still messed up. sry, needs another round

da-kami:
> somehow the logic is still messed up. sry, needs another round

correction, the logic seems to be correct (the unit tests are sane - we still improved them...), but the behavior is still not what is expected. Have to dig deeper and see if we pick the oracle event correctly when rolling over.

da-kami:
Unfortunately we found a severe problem: Roll over actually is incomplete and  a non-collaborative close would have never worked on the current implementation of rollover (master).

The problem is, that we never update the oracle event id that we use for rollover in the cfd's data. the only oracle event id that is actually stored in the database is the initial one that is stored as part of the order. 
That's probably why on some instances we see problems like this #644 

At this point I think we have to take a step back and start tackling the cfd model restructuring, otherwise this will be a never ending story.
What we have to do is don't use the `Order` as source of the `event_id` on the cfd, but put the `event_id` on the `Cfd` itself (in the database as well). This will require updating the `Cfd` and not updating the state. 
Introducing the `event_id` into the state is not a good idea, it won't help and will make our current state handling even more messy. 

@thomaseizinger described more details in #646 

@scratchscratchscratchy @itchymax

thomaseizinger:
Outdated, will tackle feature once new model is introduced.

PR #629, opened 2021-11-18
Rollover test

rishflab:


thomaseizinger:
Outdated, will tackle feature once new model is introduced.

PR #643, opened 2021-11-19
Don't allow rollover that is too close to expiry

da-kami:
When writing tests we noticed, that we will have unpredictable behaviour allowing rollover that is too close to the expiry time.

@thomaseizinger we discussed hat earlier and I implemented it, but then came to the conclusion that we should not use it. 
I still add it as a draft up for discussion. 

Here's my reasoning:

The taker should be able to propose a rollover until the very last second. If the maker already committed, the maker already committed. The rollover should be rejected then (rollovers should only be accepted if we are in an open state, once committed we can't roll over anymore). 
I think that behaviour is resilient enough and we don't have to introduce a `too_late_to_rollover` - I don't think it would add value. From a taker-UX point of view: We need feedback from maker->taker upon rejection. If there is no feedback then the UX will suffer. But I would solve that by sending proper errors over rather than introducing this interval. 

If I am missing something please let me know.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3826,7 +3826,6 @@ dependencies = [
  "futures-timer",
  "futures-util",
  "pollster",
- "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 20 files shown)


da-kami:
closing because I don't think we need it

PR #645, opened 2021-11-19
Introduce projection actor

### LINKED ISSUE #572: Encorporate UI projections into actor system

bonomat:
Ground work for #572

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1,6 +1,6 @@
 use crate::model::cfd::{Cfd, CfdState, Order, OrderId};
 use crate::model::{BitMexPriceEventId, Usd};
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use rust_decimal::Decimal;
 use sqlx::pool::PoolConnection;
 use sqlx::{Sqlite, SqlitePool};
@@ -97,6 +97,13 @@ pub async fn load_order_by_id(
 }
 
 pub async fn insert_cfd(cfd: &Cfd, conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<()> {
+    if load_cfd_by_order_id(cfd.order.id, conn).await.is_ok() {
+        bail!(
+            "Cannot insert cfd because there is already a cfd for order id {}",
+            cfd.order.id
+        )
+    }
+
     let state = serde_json::to_string(&cfd.state)?;
     let query_result = sqlx::query(
         r#"
@@ -714,6 +721,22 @@ mod tests {
         let data = load_all_cfds(&mut conn).await.unwrap();
 
         assert_eq!(data.len(), 100);
+    }
+
+    #[tokio::test]
+    async fn inserting_two_cfds_with_same_order_id_should_fail() {
+        let mut conn = setup_test_db().await;
+
+        let cfd = Cfd::dummy().insert(&mut conn).await;
+
+        let error = insert_cfd(&cfd, &mut conn).await.err().unwrap();
+        assert_eq!(
+            error.to_string(),
+            format!(
+                "Cannot insert cfd because there is already a cfd for order id {}",
+                cfd.order.id
+            )
+        );
     }
 
     fn random_simple_state() -> CfdState {
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -529,16 +529,14 @@ pub async fn load_cfds_by_oracle_event_id(
 
 #[cfg(test)]
 mod tests {
-    use crate::cfd_actors;
     use pretty_assertions::assert_eq;
     use rand::Rng;
     use rust_decimal_macros::dec;
     use sqlx::SqlitePool;
     use time::macros::datetime;
     use time::OffsetDateTime;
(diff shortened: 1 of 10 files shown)


daemon/src/db.rs
```diff
@@ -716,6 +695,22 @@ mod tests {
         assert_eq!(data.len(), 100);
     }
 
+    #[tokio::test]
+    async fn inserting_two_cfds_with_same_order_id_should_fail() {
+        let mut conn = setup_test_db().await;
+
+        let cfd = Cfd::dummy().insert(&mut conn).await;
```
thomaseizinger, line 702: I think you could literally just call `insert` twice here.

daemon/src/projection.rs
```diff
@@ -0,0 +1,49 @@
...
+            tx_order,
+            tx_quote,
+            tx_settlements,
+        }
+    }
+}
+
+pub struct Update<T>(pub T);
+
+impl Update<Cfd> {}
```
thomaseizinger: Can be deleted.

thomaseizinger (approved):
Looks good!

daemon/src/db.rs
```diff
@@ -97,6 +97,13 @@ pub async fn load_order_by_id(
 }
 
 pub async fn insert_cfd(cfd: &Cfd, conn: &mut PoolConnection<Sqlite>) -> anyhow::Result<()> {
+    if load_cfd_by_order_id(cfd.order.id, conn).await.is_ok() {
```
da-kami, line 100: I thought I had introduced this check a long time ago...?
Weird.

da-kami: Would have been nice to have a separate commit for the db changes, I don't see how they are related to adding the projection actor and it would have been nice to have commit messages with explanations.

daemon/src/projection.rs
```diff
@@ -0,0 +1,49 @@
+use crate::bitmex_price_feed::Quote;
+use crate::{Cfd, Order, UpdateCfdProposals};
+use tokio::sync::watch;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor {
+    tx_cfds: watch::Sender<Vec<Cfd>>,
```
da-kami, line 7: what does the `tx` here stand for?

daemon/src/maker.rs
```diff
@@ -270,9 +272,27 @@ async fn main() -> Result<()> {
...
         time::Duration::hours(opts.settlement_time_interval_hours as i64),
         N_PAYOUTS,
+        projection_actor.clone(),
     )
     .await?;
 
+    let (task, init_quote) = bitmex_price_feed::new(projection_actor).await?;
+    tasks.add(task);
+
+    let cfds = load_all_cfds(&mut conn).await?;
```
da-kami, line 282: It's not expensive to do this, but now we load all cfds twice upon startup - here and in the actor system in `lib.rs`.

da-kami (commented):
Is it still a draft or is it ready to go in?

daemon/src/db.rs
bonomat: This would panic because `insert` uses `unwrap` inside.

daemon/src/projection.rs
bonomat: > Is it still a draft or is it ready to go in?

Sorry, it's ready for review and ready to ship from my perspective

bonomat:
> Is it still a draft or is it ready to go in?

Sorry, it's ready for review and ready to ship from my perspective

daemon/src/projection.rs
bonomat: `tx` is the counterpart to `rx`.
- `tx` - used for the _transmitting_ side
- `rx` - used for the _receiving_ side

afaik it's the best practice as in tokio, e.g. look [here](https://docs.rs/tokio/1.14.0/tokio/sync/watch/fn.channel.html)

daemon/src/db.rs
bonomat: mhm... that's in a separate commit with an explanation?

daemon/src/maker.rs
bonomat: Right... I don't think this is an issue here or do you have any concerns?

bonomat: Let's fix this in a follow up if you think this is important.

PR #663, opened 2021-11-22
Add connected takers feed

klochowicz:
- add a new watch channel with information about connected takers from the maker's actor system
- test the watch channel behaviour with a unit test
- add a list of connected takers to maker's UI (for testing/monitoring)

This PR will enable us to write more elaborate assertions (e.g. reconnection of the same taker & resuming its operation) and also enables us to gauge the length & amount of inteactions with the maker

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -10,7 +10,6 @@ use connection::ConnectionStatus;
 use futures::future::RemoteHandle;
 use maia::secp256k1_zkp::schnorrsig;
 use maker_cfd::TakerDisconnected;
-use model::TakerId;
 use sqlx::SqlitePool;
 use std::future::Future;
 use std::time::Duration;
@@ -91,7 +90,6 @@ impl Default for Tasks {
 
 pub struct MakerActorSystem<O, M, T, W> {
     pub cfd_actor_addr: Address<maker_cfd::Actor<O, M, T, W>>,
-    pub connected_takers_feed_receiver: watch::Receiver<Vec<TakerId>>,
     pub inc_conn_addr: Address<T>,
     pub tasks: Tasks,
 }
@@ -135,9 +133,6 @@ where
 
         let cfds = load_all_cfds(&mut conn).await?;
 
-        let (connected_takers_feed_sender, connected_takers_feed_receiver) =
-            watch::channel::<Vec<TakerId>>(Vec::new());
-
         let (monitor_addr, mut monitor_ctx) = xtra::Context::new(None);
         let (oracle_addr, mut oracle_ctx) = xtra::Context::new(None);
         let (inc_conn_addr, inc_conn_ctx) = xtra::Context::new(None);
@@ -150,7 +145,6 @@ where
             settlement_interval,
             oracle_pk,
             projection_actor,
-            connected_takers_feed_sender,
             inc_conn_addr.clone(),
             monitor_addr.clone(),
             oracle_addr.clone(),
@@ -196,7 +190,6 @@ where
 
         Ok(Self {
             cfd_actor_addr,
-            connected_takers_feed_receiver,
             inc_conn_addr,
             tasks,
         })
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -1,7 +1,7 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 #![warn(clippy::disallowed_method)]
 use crate::db::load_all_cfds;
-use crate::maker_cfd::{FromTaker, NewTakerOnline};
+use crate::maker_cfd::{FromTaker, TakerConnected};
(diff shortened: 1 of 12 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -40,7 +40,11 @@ pub struct NewOrder {
...
 }
 
-pub struct NewTakerOnline {
+pub struct TakerAdded {
+    pub id: TakerId,
+}
+
+pub struct TakerRemoved {
     pub id: TakerId,
 }
```
thomaseizinger: Maybe `TakerConnected` and `TakerDisconnected` would be better names?

daemon/src/maker_cfd.rs
```diff
@@ -345,6 +355,21 @@ where
             })
             .await?;
 
+        if !self.connected_takers.insert(taker_id) {
+            anyhow::bail!("Taker already connected: {:?}", &taker_id);
```
thomaseizinger: Is this really a failure as such?

daemon/src/maker_cfd.rs
```diff
@@ -345,6 +355,21 @@ where
...
+        }
+        self.connected_takers_feed_sender
+            .send(self.connected_takers.clone().into_iter().collect())?;
+
+        Ok(())
+    }
+
+    async fn handle_taker_removed(&mut self, taker_id: TakerId) -> Result<()> {
+        if !self.connected_takers.remove(&taker_id) {
+            anyhow::bail!("Removed unknown taker: {:?}", &taker_id);
```
thomaseizinger: Same here, perhaps logging is fine here?
Would make the function idempotent which is good I think.

daemon/tests/happy_path.rs
```diff
@@ -133,3 +133,22 @@ async fn taker_notices_lack_of_maker() {
...
+        next(maker.connected_takers_feed()).await.unwrap()
+    );
+
+    std::mem::drop(taker);
+
+    assert_eq!(
+        Vec::<TakerId>::new(),
+        next(maker.connected_takers_feed()).await.unwrap()
+    );
+}
```
thomaseizinger: Pretty nice!

daemon/src/maker_cfd.rs
klochowicz: thanks for the suggestion, they are better. Will change the names & handlers.

daemon/src/maker_cfd.rs
klochowicz: It's not a failure, but an indication that something is not right with our code....

I've downgraded both to `tracing::warn!`, ideally I'd probably add a `debug_assert!` there to catch such inconsistencies early if they'd happen.

daemon/src/maker_cfd.rs
klochowicz: converted to log.

daemon/tests/happy_path.rs
klochowicz: thanks :)

klochowicz:
@thomaseizinger I've rebased on top of latest master and have added one commit on top (moved connected_takers_feed inside projection actor). I have added a TODO to just use the same channel in the taker as well, but I would prefer to add it on top of this PR, as it already suffered from severe merge conflicts :)

daemon/src/lib.rs
```diff
@@ -1,14 +1,16 @@
...
+use crate::maker_cfd::{FromTaker, TakerConnected};
 use crate::model::cfd::{Cfd, Order, UpdateCfdProposals};
 use crate::oracle::Attestation;
 use crate::tokio_ext::FutureExt;
 use anyhow::Result;
 use connection::ConnectionStatus;
 use futures::future::RemoteHandle;
 use maia::secp256k1_zkp::schnorrsig;
+use maker_cfd::TakerDisconnected;
+use model::TakerId;
```
thomaseizinger: These imports are a bit odd, should be part of the others.

maker-frontend/src/MakerApp.tsx
```diff
@@ -41,6 +42,8 @@ export default function App() {
     const order = useLatestEvent<Order>(source, "order", intoOrder);
     const walletInfo = useLatestEvent<WalletInfo>(source, "wallet");
     const priceInfo = useLatestEvent<PriceInfo>(source, "quote");
+    const takersOrUndefined = useLatestEvent<TakerId[]>(source, "takers");
+    let takers = takersOrUndefined ? takersOrUndefined! : [];
```
thomaseizinger: There is a shortcut for this:

```suggestion
    let takers = takersOrUndefined || [];
```

thomaseizinger (approved):

daemon/src/lib.rs
klochowicz: thanks, tidied up the commits - `TakerId` should have been added earlier to ensure the tests run (as well as missing import for `TimeStamp`, as tests would not have compiled.

PR #664, opened 2021-11-22
Withdraw through UI

### LINKED ISSUE #608: Withdraw in the UI

bonomat:
Enable withdraw through UI.

If the withdraw fails the error messages are being swallowed somewhere and I don't know how to get it. Maybe @da-kami or @thomaseizinger has an idea?
The HTTP post request fails and I get an `Internal Server Error` but the actual error message is not shown. 

I solved the _withdraw all_ feature by saying: if input == 0, withdraw all. 

resolves #608   

image

URL: https://mempool.space/testnet/tx/c33de791867148c3fa7e67858b48521a72bd83edd4efe21d0666af26e5939d0d

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -5,7 +5,7 @@ use daemon::model::cfd::{Cfd, Order, OrderId, Role, UpdateCfdProposals};
 use daemon::model::{Price, Usd, WalletInfo};
 use daemon::routes::EmbeddedFileExt;
 use daemon::to_sse_event::{CfdAction, CfdsWithAuxData, ToSseEvent};
-use daemon::{bitmex_price_feed, maker_cfd};
+use daemon::{bitmex_price_feed, maker_cfd, wallet};
 use http_api_problem::{HttpApiProblem, StatusCode};
 use rocket::http::{ContentType, Header, Status};
 use rocket::response::stream::EventStream;
@@ -206,6 +206,52 @@ pub fn dist<'r>(file: PathBuf, _auth: Authenticated) -> impl Responder<'r, 'stat
 pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 'static> {
     let asset = Asset::get("index.html").ok_or(Status::NotFound)?;
     Ok::<(ContentType, Cow<[u8]>), Status>((ContentType::HTML, asset.data))
+}
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct WithdrawRequest {
+    address: bdk::bitcoin::Address,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    amount: bdk::bitcoin::Amount,
+    fee: f32,
+}
+
+#[rocket::post("/withdraw", data = "<withdraw_request>")]
+pub async fn post_withdraw_request(
+    withdraw_request: Json<WithdrawRequest>,
+    wallet: &State<Address<wallet::Actor>>,
+    network: &State<Network>,
+    _auth: Authenticated,
+) -> Result<String, HttpApiProblem> {
+    let amount =
+        (withdraw_request.amount != bdk::bitcoin::Amount::ZERO).then(|| withdraw_request.amount);
+
+    let txid = wallet
+        .send(wallet::Withdraw {
+            amount,
+            address: withdraw_request.address.clone(),
+            fee: Some(bdk::FeeRate::from_sat_per_vb(withdraw_request.fee)),
+        })
+        .await
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Could not proceed with withdraw request")
+                .detail(e.to_string())
+        })?
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::BAD_REQUEST)
+                .title("Could not withdraw funds")
+                .detail(e.to_string())
+        })?;
+
+    let url = match network.inner() {
+        Network::Bitcoin => format!("https://mempool.space/tx/{}", txid),
+        Network::Testnet => format!("https://mempool.space/testnet/tx/{}", txid),
+        Network::Signet => format!("https://mempool.space/signet/tx/{}", txid),
+        Network::Regtest => txid.to_string(),
+    };
+
+    Ok(url)
 }
 
 #[cfg(test)]
--- a/daemon/src/routes_taker.rs
+++ b/daemon/src/routes_taker.rs
@@ -3,7 +3,7 @@ use daemon::model::cfd::{calculate_long_margin, Cfd, Order, OrderId, Role, Updat
 use daemon::model::{Leverage, Price, Usd, WalletInfo};
 use daemon::routes::EmbeddedFileExt;
 use daemon::to_sse_event::{CfdAction, CfdsWithAuxData, ToSseEvent};
-use daemon::{bitmex_price_feed, taker_cfd};
+use daemon::{bitmex_price_feed, taker_cfd, wallet};
 use http_api_problem::{HttpApiProblem, StatusCode};
 use rocket::http::{ContentType, Status};
 use rocket::response::stream::EventStream;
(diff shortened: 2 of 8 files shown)


daemon/src/routes_maker.rs
```diff
@@ -208,6 +208,63 @@ pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 's
...
+
+    tracing::info!(%txid, "Withdraw successful");
+    let url = match network.inner() {
+        Network::Bitcoin => format!("https://mempool.space/tx/{}", txid),
+        Network::Testnet => format!("https://mempool.space/testnet/tx/{}", txid),
+        Network::Signet => format!("https://mempool.space/signet/tx/{}", txid),
+        Network::Regtest => txid.to_string(),
+    };
+
+    Ok(status::Accepted(Some(url)))
```
thomaseizinger: Is the use of `Accepted` here intentional?

With `Accepted` you could include a `Location` header in the response which points the user to a "progress" page about the status of the processing.

I guess one could spin it such that the mempool page is about the progress of getting the TX included? I think I wouldn't bother with these semantics though and just return 200 OK.

daemon/src/routes_maker.rs
```diff
@@ -208,6 +208,63 @@ pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 's
...
+        withdraw_request.amount,
+        withdraw_request.address,
+        withdraw_request.fee
+    );
+
+    let amount = if withdraw_request.amount == bdk::bitcoin::Amount::ZERO {
+        None
+    } else {
+        Some(withdraw_request.amount)
+    };
```
thomaseizinger: There is a `.then`  combinator on boolean's that does exactly this.

daemon/src/routes_maker.rs
```diff
@@ -208,6 +208,63 @@ pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 's
...
+    wallet: &State<Address<wallet::Actor>>,
+    network: &State<Network>,
+    _auth: Authenticated,
+) -> Result<status::Accepted<String>, HttpApiProblem> {
+    tracing::info!(
+        "Withdrawing {:?} to {} with fee {}",
+        withdraw_request.amount,
+        withdraw_request.address,
+        withdraw_request.fee
+    );
```
thomaseizinger: I think this log should move into the wallet so you can omit it from both endpoints.

daemon/src/routes_maker.rs
```diff
@@ -208,6 +208,63 @@ pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 's
...
+                .title("Withdraw failed")
+                .detail(e.to_string())
+        })?
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Withdraw failed")
+                .detail(e.to_string())
+        })?;
+
+    tracing::info!(%txid, "Withdraw successful");
```
thomaseizinger: Same here, I think this should go into the wallet actor.

daemon/src/routes_maker.rs
```diff
@@ -208,6 +208,63 @@ pub fn index<'r>(_paths: PathBuf, _auth: Authenticated) -> impl Responder<'r, 's
     Ok::<(ContentType, Cow<[u8]>), Status>((ContentType::HTML, asset.data))
 }
 
+#[derive(Debug, Clone, Deserialize)]
+pub struct WithdrawRequest {
+    pub address: bdk::bitcoin::Address,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    pub amount: bdk::bitcoin::Amount,
+    pub fee: f32,
```
thomaseizinger: Strictly speaking, these wouldn't need to be `pub`.

daemon/src/routes_taker.rs
```diff
@@ -211,3 +211,59 @@ pub fn index<'r>(_paths: PathBuf) -> impl Responder<'r, 'static> {
...
+            amount,
+            address: withdraw_request.address.clone(),
+            fee: Some(bdk::FeeRate::from_sat_per_vb(withdraw_request.fee)),
+        })
+        .await
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Withdraw failed")
+                .detail(e.to_string())
+        })?
```
thomaseizinger: The first `Result` only happens when the actor gets disconnected, I think we can write a better error message here (500 is still a good code).

daemon/src/routes_taker.rs
```diff
@@ -211,3 +211,59 @@ pub fn index<'r>(_paths: PathBuf) -> impl Responder<'r, 'static> {
...
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Withdraw failed")
+                .detail(e.to_string())
```
thomaseizinger: Is 500 appropriate for errors within the wallet actor? Read: Is it the user's fault that the error happened (in which case 4xx would be better) or can they expect that it works at a later stage by retrying the same request (then 5xx is okay)?

daemon/src/routes_taker.rs
```diff
@@ -211,3 +211,59 @@ pub fn index<'r>(_paths: PathBuf) -> impl Responder<'r, 'static> {
...
+    let amount = if withdraw_request.amount == Amount::ZERO {
+        None
+    } else {
+        Some(withdraw_request.amount)
+    };
+
+    let txid = wallet
+        .send(wallet::Withdraw {
+            amount,
+            address: withdraw_request.address.clone(),
```
thomaseizinger, line 235: Does the wallet verify the address' network against the connected one?

taker-frontend/src/App.tsx
```diff
@@ -54,6 +55,15 @@ async function postCfdOrderRequest(payload: CfdOrderRequestPayload) {
     }
 }
 
+export async function postWithdraw(payload: WithdrawRequest) {
+    let res = await fetch(`/api/withdraw`, { method: "POST", body: JSON.stringify(payload) });
+    if (!res.status.toString().startsWith("2")) {
+        console.log(`Error${JSON.stringify(res)}`);
+        throw new Error("failed to withdraw: " + res.status + ", " + res.statusText);
```
thomaseizinger: Your error is being swallowed because you are completely ignoring the body here. Check how `createErrorToast` is used in other places. I believe it has capabilities to parse an RFC7807 response which should give you the error.

taker-frontend/src/components/Wallet.tsx
```diff
@@ -13,14 +40,54 @@ export default function Wallet(
...
+        setWithdrawAmount(Number.parseFloat(amount));
+    };
+    const [fee, setFee] = useState(1);
+    const onFeeChange = (amount: string) => {
+        setFee(Number.parseFloat(amount));
+    };
+    const [withdrawAddress, setWithdrawAddress] = useState("");
+    const onWithdrawAddressChange = (address: string) => {
+        setWithdrawAddress(address);
+    };
```
thomaseizinger: This seems unnecessary?

taker-frontend/src/components/Wallet.tsx
```diff
@@ -13,14 +40,54 @@ export default function Wallet(
...
+                });
+                toast({
+                    title: "Withdraw successful",
+                    description: <Link href={url} isExternal>
+                        {url}
+                    </Link>,
+                    status: "info",
+                    duration: 10000,
+                    isClosable: true,
+                });
```
thomaseizinger, line 70: I think I'd prefer if we open a new tab in case the response is a URL. The toast may go away and then the user has no way of retrieving the transaction ID.

thomaseizinger: I think ideally, we would persist these transaction IDs to the DB and display them as a "withdrawal TX list" underneath the wallet.

taker-frontend/src/components/Wallet.tsx
```diff
@@ -44,6 +111,65 @@ export default function Wallet(
...
+                            </Input>
+                        </FormControl>
+                        <HStack>
+                            <FormControl id="amount">
+                                <FormLabel>Amount</FormLabel>
+                                <NumberInput
+                                    min={0}
+                                    max={balance}
+                                    default={0}
+                                    onChange={(event) => onWithdrawAmountChange(event)}
```
thomaseizinger: I think `event` is not true here, it is directly the amount? I don't see an code anywhere that extracts the number from this "event".

taker-frontend/src/components/Wallet.tsx
```diff
@@ -44,6 +111,65 @@ export default function Wallet(
...
+                                </NumberInput>
+                                <FormHelperText>How much do you want to withdraw? (0 to withdraw all)</FormHelperText>
+                            </FormControl>
+                            <FormControl id="fee" w={"30%"}>
+                                <FormLabel>Fee</FormLabel>
+                                <NumberInput
+                                    min={1}
+                                    max={100}
+                                    default={0}
+                                    onChange={(event) => onFeeChange(event)}
```
thomaseizinger: Same here.

thomaseizinger (commented):
Some comments!

daemon/src/routes_maker.rs
thomaseizinger: See https://httpstatuses.com/202.

bonomat: It wasn't on purpose but I learned something new :) 
Changing to 200 OK instead

daemon/src/routes_maker.rs
bonomat: Such a log already exists 🙈

daemon/src/routes_taker.rs
bonomat: Good point. I guess in this case it is very likely that it failed because of invalid user input (amount too high)

daemon/src/routes_taker.rs
bonomat: Of course not :D

taker-frontend/src/App.tsx
bonomat: I use it the same as in other places (throw an error here and the caller catches and uses `createErrorToast`)
 but maybe that has never worked?

bonomat: 🤦‍♂️
... `import createErrorToast from "../../../maker-frontend/src/components/ErrorToast";`

taker-frontend/src/components/Wallet.tsx
bonomat: > I think I'd prefer if we open a new tab in case the response is a URL. The toast may go away and then the user has no way of retrieving the transaction ID.

I'll do both. Useful in case a user has blocked the site from opening new tabs.

bonomat: > I think ideally, we would persist these transaction IDs to the DB and display them as a "withdrawal TX list" underneath the wallet.

I think this is a topic story and goes towards showing a complete history of the wallet actions/transactions.

taker-frontend/src/components/Wallet.tsx
bonomat: On `NumberInput` this event is the value itself as string or number: 

```
    /**
     * The callback fired when the value changes
     */
    onChange?(valueAsString: string, valueAsNumber: number): void;
```

thomaseizinger: It is not an `event` then is it? `value` would be a more appropriate name I think.

bonomat: yes, changed it already :)

daemon/src/routes_maker.rs
thomaseizinger: Let's remove it from here then?

bonomat: Already gone :)

da-kami (approved):

daemon/src/routes_maker.rs
thomaseizinger: Huh? I still see a "Withdrawing" log in the diff? What am I missing? :)

bonomat: Nothing, you are completely right, I forgot that one :(

taker-frontend/src/App.tsx
klochowicz: @bonomat there's a copy in the `taker-frontend/src/components`! (I copied it there)

It works, as the other toasts for failed actions are displayed correctly.

bonomat: I know, I've imported the wrong one which did not work 😅

PR #665, opened 2021-11-22
Refactor model to use event sourcing

### LINKED ISSUE #375: Simplify `append_cfd_state` to a single SQL query
Currently, we are only loading the latest state to make a `mem::discriminant` check if we are appending the same state again.

This is a legacy from when we thought that we should never be appending the current state. This assumption turned out to be wrong which is why the log was changed to `trace`.

We don't need this check at all actually, meaning we can simplify these SQL statements to a single one the same way we insert the CFD based on the order.

### LINKED ISSUE #644: No CET for oracle event
```
2021-11-19 15:45:07 DEBUG Fetching attestation for /x/BitMEX/BXBT/2021-11-10T07:00:00.price?n=20
2021-11-19 15:45:07  INFO Fetched new attestation for /x/BitMEX/BXBT/2021-11-10T07:00:00.price?n=20
2021-11-19 15:45:07 DEBUG Learnt latest oracle attestation for event: /x/BitMEX/BXBT/2021-11-10T07:00:00.price?n=20
2021-11-19 15:45:07  INFO Fetched new attestation for /x/BitMEX/BXBT/2021-11-05T08:00:00.price?n=20
2021-11-19 15:45:07 ERROR No CET for oracle event /x/BitMEX/BXBT/2021-11-05T08:00:00.price?n=20 and cfd with order id e0d42219-f50b-4e08-adde-6bfd1a82d3be
```

I have not seen this in a while, I thought it was fixed with https://github.com/itchysats/itchysats/pull/475 - it seems there is another problem related to this.

### LINKED ISSUE #646: Enforce privacy on domain models
We need to enforce privacy on our domain models:

- No direct assignment of `state`, all state transitions must go through a "command" function (This will enable an easier transition of the internal state model to a graph or other things)
- No direct reaching into fields (like `cfd.order`). The order MUST become an implementation detail of the CFD. These fields MUST NOT be `pub`.

FYI @bonomat @klochowicz @rishflab @da-kami @luckysori. Tagging everyone here so you guys are aware. When you come across this issue in other development tasks, please fix it on the way.

### LINKED ISSUE #725: sqlx query warning in the test
when running with `--nocapture`, I've noticed this:

```
[2mNov 26 15:27:09.136[0m [33m WARN[0m sqlx::query: with ord as ( …; rows: 1, elapsed: 9.645s

with ord as (
  select
    id as order_id,
    uuid,
    trading_pair,
    position,
    initial_price,
    min_quantity,
    max_quantity,
    leverage,
    liquidation_price,
    creation_timestamp_seconds as ts_secs,
    settlement_time_interval_seconds as settlement_time_interval_secs,
    origin,
    oracle_event_id
  from
    orders
),
cfd as (
  select
    ord.order_id,
    id as cfd_id,
    quantity_usd
  from
    cfds
    inner join ord on ord.order_id = cfds.order_id
),
state as (
  select
    id as state_id,
    cfd.order_id,
    cfd.quantity_usd,
    state
  from
    cfd_states
    inner join cfd on cfd.cfd_id = cfd_states.cfd_id
  where
    id in (
      select
        max(id) as id
      from
        cfd_states
      group by
        (cfd_id)
    )
)
select
  ord.uuid as "uuid: crate::model::cfd::OrderId",
  ord.trading_pair as "trading_pair: crate::model::TradingPair",
  ord.position as "position: crate::model::Position",
  ord.initial_price as "initial_price: crate::model::Price",
  ord.min_quantity as "min_quantity: crate::model::Usd",
  ord.max_quantity as "max_quantity: crate::model::Usd",
  ord.leverage as "leverage: crate::model::Leverage",
  ord.liquidation_price as "liquidation_price: crate::model::Price",
  ord.ts_secs as "ts_secs: crate::model::Timestamp",
  ord.settlement_time_interval_secs as "settlement_time_interval_secs: i64",
  ord.origin as "origin: crate::model::cfd::Origin",
  ord.oracle_event_id as "oracle_event_id: crate::model::BitMexPriceEventId",
  state.quantity_usd as "quantity_usd: crate::model::Usd",
  state.state
from
  ord
  inner join state on state.order_id = ord.order_id
where
  ord.uuid = $1
  
```

should this take 10s with one cfd inside the in-memory db?

### LINKED ISSUE #780: CFD forever remains in `PendingOpen` after permanent transaction publication error
I opened two CFD positions right after each other - but for the second one I did not have enough balance. 
Since we don't check for the balance during contract setup this was not detected. Only once publishing the lock transaction we failed with:

```
2021-12-01 23:15:46  INFO daemon::taker_cfd: Setup complete, publishing on chain now
2021-12-01 23:15:47 ERROR daemon::taker_cfd: Message handler failed: Broadcasting transaction failed. Txid: 0a541980849c67bff93f02517478fa28e3c8c945ef3f16f5b192d05d274d9163. Raw transaction: 0200000000010541e861f99814684960c7fbd20ed36416d46fd3ef2029ee27ba311f2e2d4631790000000000ffffffff4331eb97425bf125208331075ea2657faf6097c08df3489f2266337b1c9137f90000000000ffffffffbfcef016f49e37e3fe51c2a27264fca5a0876cdf6ab57a654dd590e39ce784350100000000ffffffff6f62ded11af428a1738950bf31f4e47a20a8f469e8a0f700e919eb7fb8142dcf0200000000ffffffffb0c3c38c699b33758aa21f6b6e37398beae7108b05b26d7fff3b17260c56fba50100000000ffffffff0343fb030000000000220020160b230ac2f2b7a4c7a6a8eda7cfbbeda62bc05253ead86dd02caaf7d0a25929476f0200000000001600144ac706ce3cb3f325b63895798a9f53db170ae876380a000000000000160014eebffd19957659b6464c3a49cb52cff26de89587024830450221008ceb534e6967011babbeb0010772a1bb5b0940a335fdf5af77266a8f88266f3502205bb63c9a901456df87c515ab361b21db31da5a723cfb5861e611e279824da7900121034dcbbcfd6a820d6cb1b5ba411cddcaa5165d708e07ddfa26381889285dc9d76e02483045022100ef4f3c1651e25bf8dd995084a6335f1ba4d1b24835470e701cbfedacb458e84302207d9ace5c6c0ad17c8f503320c73851d08596f919da20032667f3b6bff75a42a60121034dcbbcfd6a820d6cb1b5ba411cddcaa5165d708e07ddfa26381889285dc9d76e0247304402206af9bcbf58b6961909a1ac1c7d133d42d846e361fe7bc62dc5093b487892e9a802202ec4f672508097ab9aa5bf87bd3dea035cbaf67f49814f206cf0f44e7c99cfb10121029f2c252444dd9cd11a5e3b2ba0b0b96ed859157f481519dae72f7b67d7630e1302483045022100a5e4e2e9fb796a5e74fe69d68603b2621dc1d5f06d1759f1ffcd4d87c2738ebe02207ecd244739a824b670e1b01516c5201ed5f554ec0f781bdf9081812b9bc99b590121032a8fc2901a0022aba66c1c45c1b1880ee6fd5e4d01bb0be6cf0b49413940fc150248304502210093d4d10834e42a2d7b6f2fcf90945ddbe484a3d6843cf42dce693584fce212c602206756cd0b66f458940ebc65aeaef05e6cb57dd79903b72a4980f4917a3d7eea1e012103afbdb3b7636cb6a512eda57016eb9f7e283d57d05c1c762b994835f2362f2a5100000000: Electrum(Protocol(String("sendrawtransaction RPC error: {\"code\":-26,\"message\":\"txn-mempool-conflict\"}")))
```

transaction:
first: https://mempool.space/testnet/tx/4a6d8da59131666c9f3c73f38c40643e6174fafc167065b8ba6c07e84487c30e (made it into mempool)
second: https://mempool.space/testnet/tx/0a541980849c67bff93f02517478fa28e3c8c945ef3f16f5b192d05d274d9163 (never made it into mempool)

However, in the UI the CFD is displayed as `PendingOpen`. 

image

The problem is, that we don't handle "failing" to broadcast the transaction - i.e. there is no state changed recorded here. But that means that when restarting the daemon we will always try to re-broadcast (which will always fail), so this CFD will remain stuck in this "retry-loop" forever.

One way to deal with this is avoiding such kind of scenarios through bookkeeping - i.e. deduct the margin of cfds that are in contract setup from the balance, so the user cannot open another position while still setting up.

For more resilience we should deal with different failure scenarios when publishing transactions. If the publication runs into a permament error the CFD should be considered failed (and treated as final).

da-kami:
Make the change easy, then make the easy change.

## Work that makes / will make this change easier:

- [x] https://github.com/itchysats/itchysats/pull/796
- [x] https://github.com/itchysats/itchysats/pull/795
- [x] https://github.com/itchysats/itchysats/pull/791
- [x] https://github.com/itchysats/itchysats/pull/765
- [x] https://github.com/itchysats/itchysats/pull/743
- [x] https://github.com/itchysats/itchysats/pull/769
- [x] https://github.com/itchysats/itchysats/pull/746
- [x] https://github.com/itchysats/itchysats/pull/744
- [x] https://github.com/itchysats/itchysats/pull/740
- [x] https://github.com/itchysats/itchysats/pull/715
- [x] https://github.com/itchysats/itchysats/pull/707
- [x] https://github.com/itchysats/itchysats/pull/718
- [x] https://github.com/itchysats/itchysats/pull/708
- [x] https://github.com/itchysats/itchysats/pull/684
- [x] https://github.com/itchysats/itchysats/pull/704
- [x] https://github.com/itchysats/itchysats/pull/695
- [x] https://github.com/itchysats/itchysats/pull/690
- [x] https://github.com/itchysats/itchysats/pull/645
- [x] #743
- [x] https://github.com/itchysats/itchysats/pull/819
- [x] https://github.com/itchysats/itchysats/pull/775
- [x] https://github.com/itchysats/itchysats/pull/806
- [x] https://github.com/itchysats/itchysats/pull/863 (this reduces the number of merge conflicts during rebases)
- [x] https://github.com/itchysats/itchysats/pull/876 (this will make rebasing easier because `CfdAggregate` has getters)
- [ ] Move all failure cases for incoming new order into protocol actor (fetching oracle announcement, etc)
- [ ] Remove this check? https://github.com/itchysats/itchysats/blob/0a9240abe28a1d37fd8e7a98f521d9147dd93491/daemon/src/taker_cfd.rs#L211-L213
- [x] https://github.com/itchysats/itchysats/pull/899
- [x] https://github.com/itchysats/itchysats/pull/900
- [x] https://github.com/itchysats/itchysats/pull/917

## Missing bits in this PR

- [x] Manually switch to `imort-granularity: item` before the next rebase
- [x] Delete old `Cfd` and rename `CfdAggregate` to `Cfd` (this will make rebasing a lot easier because there are no naming conflicts)
- [x] Create dedicated projection from events for projection actor
- [x] Create dedicated projection from events for monitoring actor (for startup)
- [ ] Grep for TODOs and fix them
- [ ] ~Consider introducing a central `DomainEventActor` right away instead of doing another iteration~
- [ ] ~Consider introducing a housekeeping task that moves closed CFDs into an archive table and prunes all events~

## Features to re-introduce

- [ ] fee-rate inside CFD
- [ ] unit tests covering rollover

Fixes https://github.com/itchysats/itchysats/issues/646.
Fixes https://github.com/itchysats/itchysats/issues/644.
Fixes https://github.com/itchysats/itchysats/issues/375 (will be obsolete).
Fixes https://github.com/itchysats/itchysats/issues/725 (will be obsolete due to change in DB model).
Fixes https://github.com/itchysats/itchysats/issues/780.


da-kami:
@thomaseizinger I added a million ToDos where I ran into "friction", if you have time please have a look, otherwise let's go over it tomorrow. Once we sorted the ToDos that require decisions out it should be fairly straight forwards to pull the model through.

What do you think about mapping to the old `CfdState` in the first iteration, so we don't have to touch removing that yet? 
Not sure it's a good idea, but that would push some of the work to later.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
     }
 }
 
+trait Aggregate {
```
thomaseizinger: Unless we want to abstract over aggregates, I would ditch this for now.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+
+    pub enum Command {
+        CompleteContractSetup(Result<Dlc, ContractSetupError>),
+
+        // TODO: Should this `Complete` represent Accepted/Rejected/Failed, or should we model
+        // accept/reject separated from setup exchange?  Error could represent rejected as
+        // well
+        CompleteRollover(Result<(Dlc, RevokedCommit)>),
+
+        // TODO: Should this `Complete` represent Accepted/Rejected/Failed, or should we model
```
thomaseizinger: Yes needs to model all of them.

thomaseizinger: Can be done using a Result if you want.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+
+        ConfirmCetTimelock,
+        ConfirmRefundTimelock,
+
+        // oracle
+        HandleAttestation(Attestation),
+    }
+
+    pub enum Event {
+        ContractSetupCompleted(Dlc),
```
thomaseizinger: Each variant needs to have the CFDs ID + a timestamp.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+            self.version
+        }
+
+        fn apply(mut self, evt: &Self::Item) -> CfdAggregate {
+            self.version += 1;
+
+            match evt {
+                Event::ContractSetupCompleted(dlc) => self.dlc = Some(dlc.clone()),
+                Event::OracleAttested(attestation) => {
+                    self.attestation = Some(attestation.clone());
```
thomaseizinger: Taking ownership would allow us to avoid the clone.

daemon/src/model/cfd.rs
```diff
@@ -1685,6 +1955,9 @@ pub struct Cet {
 /// To use any of these transactions, we need to re-sign them with the correct secret key.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Dlc {
+    // TODO: Add the oracle event id to the dlc?
```
thomaseizinger: Might be a good idea actually.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+            todo!(
+                "should this advance state? if it does not then there is not point in having this"
+            )
+        }
+
+        // TODO: Opted for one single function that handles all commands rather than a lot of
+        // different functions.  This makes it easier to see all commands and have the logic
+        // in one place. But it is one big function again - pretty cleaned up though.
+        //  Somewhat less typing when implementing it like this, not sure if better than having
+        // separate functions...  Separate functions might be more readable...?
```
thomaseizinger: IMO separate functions have multiple advantages:

1. Shorter
2. Less indentation because no `match`
3. Less typing on the call site
4. Less complexity: An enum suggests that there is at least one place in the code where we need to deal with `Command`s dynamically, ie store them without knowing which concrete instance it is but that is not true. We always know which command we want to execute. Modelling this using an enum is using the wrong tool of the programming language for this job.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+        Self: Sized;
+}
+
+pub mod taker {
+
+    use super::*;
+
+    /// Models the cfd state
+    ///
+    /// Upon `Command`s, that are reaction to something happening in the system, we decide to
```
thomaseizinger: I think this is misleading, a command is not a reaction as such. Yes we are "reacting" to timelocks etc but these commands are the _essence_ of the system. They are active controls.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+                collaborative_settlement_finality: false,
+                cet_timelock_expired: false,
+                refund_timelock_expired: false,
+                cfd_state: CfdState::outgoing_order_request(),
+            }
+        }
+
+        pub fn start_contract_setup(&self) {
+            todo!(
+                "should this advance state? if it does not then there is not point in having this"
```
thomaseizinger: It shouldn't advance state but the point of this is to push this logic into the domain model.

Additionally, starting a contract setup is not valid in all states hence there is value in having this on here, esp. once the state machine for performing the contract setup.

Lastly, having this on here allows us to access private data nicely that will be needed for contract setup instead of exposing dumb getters and composing things outside of the domain model.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+
+        // TODO: Opted for one single function that handles all commands rather than a lot of
+        // different functions.  This makes it easier to see all commands and have the logic
+        // in one place. But it is one big function again - pretty cleaned up though.
+        //  Somewhat less typing when implementing it like this, not sure if better than having
+        // separate functions...  Separate functions might be more readable...?
+        pub fn handle_command(&self, command: Command) -> Vec<Event> {
+            match command {
+                Command::CompleteContractSetup(result) => match result {
+                    Ok(dlc) => vec![Event::ContractSetupCompleted(dlc)],
```
thomaseizinger: These commands will need to validate that executing them on the current state makes sense.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+        lock_finality: bool,
+        commit_finality: bool,
+        refund_finality: bool,
+        cet_finality: bool,
+        collaborative_settlement_finality: bool,
+
+        cet_timelock_expired: bool,
+        refund_timelock_expired: bool,
+
+        // TODO: Can we map to cfd state for now to have contained changes in the legacy code?
```
thomaseizinger: I don't think I would bother, it would be nice to chuck it all out in one go.

daemon/src/model/cfd.rs
```diff
@@ -1276,6 +1279,273 @@ impl Cfd {
...
+                    vec![Event::CollaborativeSettlementConfirmed]
+                }
+                Command::HandleAttestation(attestation) => {
+                    // TODO: The way I imagine this:
+                    //  Rather than having function `cet(&self) -> Option<Cet>` on the aggregate we
+                    // explicitly transition into having a ready cet
+                    //  I find "tell don't ask" cleaner here, i.e. event that tells the cet is ready
+                    // rather than check from the outside if "a cet is available".
+                    //  Note: When handling CetReadyForPublication on the outside we would
+                    // apply+save the event and then publish the tx.
```
thomaseizinger: I think that is good.

thomaseizinger (commented):
Thanks for ramping this up! Some thoughts :)

daemon/src/model/cfd.rs
da-kami: Should the `handle_command` return a result? Otherwise, in a case where I don't get an event (but expect one...?) - what should the behavior be?

daemon/src/model/cfd.rs
da-kami: Would you mind adding a suggestion?

daemon/src/model/cfd.rs
da-kami: I can see to model it like that. One thing to keep in mind is that we use the event-id (in particular its timestamp) for calculations outside the DLC. I feel we might introduce a dependency here that we don't want.

daemon/src/model/cfd.rs
da-kami: I can model it to be more driven by the `Event`.

My thinking so far was:
When we handle a command we already know for which cfd it was emitted. 
The `Event` is emitted by the `Aggregate`. The `Aggregate` knows it's own id, so why would the `Event` need it?
But that will be annoying when saving things (...). Probably better to model it on the `Event` and make the `Event` the citizen that is concerned with this.

thomaseizinger: We need to store the ID with the event so the aggregate that emits it needs to embed it otherwise we lose the association.

daemon/src/model/cfd.rs
```diff
@@ -1336,6 +1339,303 @@ impl Cfd {
...
+        cet: Transaction,
+    },
+
+    RefundTimelockConfirmed {
+        refund_tx: Transaction,
+    },
+
+    // TODO: Once we use multiple aggregates in different actors we could change this to something like
+    // CetReadyForPublication that is emitted by the CfdActor. The Oracle actor would take care of saving and broadcasting an attestation event
+    // that can be picked up by the cfd actor which can then decide to publish the CetReadyForPublication event.
```
thomaseizinger: That is an interesting idea.

scratchscratchscratchy: could even go further and let `CetReadyForPublication` be picked up by the wallet actor that just broadcasts. We could do that with all transactions. The wallet actor could even record another event that records the broadcast. Not sure where we would draw the line :D

For now the line is: Only change the current messaging between the actors were we have to (avoid as much as possible). The cfd actor stays in charge of the database and the wallet. once we have the aggregate in we can go wilder.

da-kami: Thanks @scratchscratchscratchy - we just had another round of discussion about this after we agreed on this earlier @thomaseizinger

daemon/src/projection.rs
```diff
@@ -627,6 +630,21 @@ impl From<Order> for CfdOrder {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, derive_more::Display)]
+pub struct Identity(String);
```
thomaseizinger: I am not sure I follow your reasoning on the commit message. The `model::Identity` serializes exactly to what you need. What is the issue?

daemon/src/projection.rs
```diff
@@ -426,7 +426,7 @@ impl Cfd {
                 self.state = CfdState::IncomingSettlementProposal;
 
                 if role == Role::Maker {
-                    self.actions = vec![CfdAction::AcceptSettlement, CfdAction::AcceptSettlement];
+                    self.actions = vec![CfdAction::AcceptSettlement, CfdAction::RejectSettlement];
```
da-kami, line 436: This maker was too good to be true!

daemon/src/collab_settlement_maker.rs
```diff
@@ -174,6 +174,7 @@ impl Actor {
     }
 
     async fn reject(&mut self, ctx: &mut xtra::Context<Self>) {
+        // TODO: For the taker, we force-close on rejection, should we do the same thing here?
```
da-kami: I would say no. Because the maker does not have an incentive to close the position. 
If you see it from the taker's point of view: The taker wants to close because the taker clicks the close button. So, if collaborative is not an option it falls back to non-collaborative. 

The maker never clicked the close button, and should leave it up to the taker how to proceed.

daemon/src/projection.rs
```diff
@@ -62,41 +72,412 @@ pub struct Feeds {
...
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    pub margin: Amount,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    pub margin_counterparty: Amount,
+
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc::opt")]
+    pub profit_btc: Option<SignedAmount>,
+    pub profit_percent: Option<String>,
+
+    // TODO: Either don't store it as on Option or expose a getter that unwraps the option
```
da-kami: oudated TODO

thomaseizinger:
Merry Christmas!

bors r+

PR #678, opened 2021-11-23
We can safely assume that creating a timestamp should not fail

da-kami:
Expect instead of returning a `Result`.


thomaseizinger (approved):

da-kami:
bors retry

PR #679, opened 2021-11-23
Refactor `useAsync` usages into `usePostRequest` hook

thomaseizinger:


taker-frontend/src/components/History.tsx
```diff
@@ -86,16 +76,7 @@ const CfdDetails = ({ cfd }: CfdDetailsProps) => {
...
-        deferFn: async ([orderId, action]: any[]) => {
-            try {
-                console.log(`Closing: ${orderId} ${action}`);
-                await doPostAction(orderId, action);
-            } catch (e) {
-                createErrorToast(toast, e);
-            }
-        },
-    });
+    let [settle, isSettling] = usePostRequest(`/api/cfd/${cfd.order_id}/settle`);
```
klochowicz, line 79: this will need to become general really soon as @rishflab is working on using "commit" action if there's no maker online). Perhaps it's worth keeping it general with `postAction, isActioning` with the action name argument?

klochowicz (approved):
cool, nice cleanup! :)

taker-frontend/src/components/History.tsx
thomaseizinger: You can just make the URL dynamic? That is easily supported :)

thomaseizinger:
Rebased and enhanced the implementation to support the new usecase of returning response as text.

PR #684, opened 2021-11-23
Introduce short-lived `setup_taker::Actor`

luckysori:


daemon/src/setup_taker.rs
```diff
@@ -0,0 +1,213 @@
...
+        //   `update_sender.send(db::load_all_cfds(conn).await?)?;`
+        // I see two options:
+        // - make `update_sender` a mpmc (via `crossbeam` for example).
+        // - send a message to the `taker_cfd` to update the feed.
+
+        let offer_announcement = self
+            .get_announcement
+            .send(oracle::GetAnnouncement(cfd.order.oracle_event_id))
+            .await?
+            .with_context(|| format!("Announcement {} not found", cfd.order.oracle_event_id))?;
```
thomaseizinger: Could we pass the announcement in on constructing the actor? That would remove `get_announcement` channel.

daemon/src/setup_taker.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        self.monitor_attestation
+            .send(oracle::MonitorAttestation {
+                event_id: offer_announcement.id,
+            })
+            .await?;
```
thomaseizinger: Can we do as within `taker_cfd::Actor` once it setup finishes?

thomaseizinger (commented):
Yep that looks really good! Can't spot anything obvious why the tests would be failing ...

daemon/src/connection.rs
```diff
@@ -25,6 +27,8 @@ pub struct Actor {
     /// Max duration since the last heartbeat until we die.
     timeout: Duration,
     connected_state: Option<ConnectedState>,
+    // QUESTION: Does this need to be generic?
```
luckysori: :question: Is it okay to set `wallet::Actor` here?

daemon/src/setup_taker.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        let mut conn = self.db.acquire().await?;
+        let cfd = Cfd::new(order.clone(), *quantity, CfdState::accepted());
+        db::append_cfd_state(&cfd, &mut conn).await?;
+        // TODO:
+        // Updating the cfd feed is missing i.e.
+        //   `update_sender.send(db::load_all_cfds(conn).await?)?;`
+        // I see two options:
+        // - make `update_sender` a mpmc (via `crossbeam` for example).
+        // - send a message to the `taker_cfd` to update the feed.
```
luckysori: :question: What do you think about this problem, @thomaseizinger?

daemon/src/connection.rs
thomaseizinger: No I think that will become an issue once you try and insert something into this hashmap.

Best to remove the type parameter from `setup_contract` and pass two `MessageChannel`s in. If that works, try and land that as a separate patch :)

daemon/src/setup_taker.rs
thomaseizinger: Can we move the update back into `taker_cfd::Actor` and you just send it a message back? Will require another `MessageChannel` but for now, that is probably the least invasive thing.

luckysori:
@thomaseizinger: I tackled the improvements suggested and figured out what the bug was: I forgot to actually put anything in these two `setup_actors` hashmaps, so we could never get any messages from the maker to the corresponding `setup_taker::Actor`.

Looking into this bug made me think about a couple of things:

- I'm not too fond of using the `wire` message types for anything other than sending over the wire. I couldn't simply add the address of the `setup_taker::Actor` to the message that was already going to the `connection::Actor` because the same message needed to be serializable and relevant to the maker.
- Isn't it a suspicious sign that we keep the same `setup_actors` state in two different actors? I may have followed your original branch a little blindly in that regard. Perhaps it would be better if only the `taker_cfd::Actor` kept this state and it was responsible for both ensuring that there's only one contract setup per order ID at any given moment and that messages from the maker go to the correct `setup_taker::Actor`.

daemon/src/taker_cfd.rs
```diff
@@ -329,6 +266,85 @@ impl<O, M, W> Actor<O, M, W> {
...
+    /// forward it to the `connection::Actor`.
+    async fn handle_register_setup_address(
+        &mut self,
+        msg: setup_taker::RegisterAddress,
+    ) -> Result<()> {
+        self.setup_actors.insert(msg.order_id, msg.address.clone());
+        self.register_setup_actor.send(msg).await?;
+
+        Ok(())
+    }
```
thomaseizinger: This should not be necessary? You know the address of the actor when you create it.

daemon/src/setup_taker.rs
```diff
@@ -0,0 +1,252 @@
...
+impl xtra::Actor for Actor {
+    async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
+        let order_id = self.order.id;
+        self.maker
+            .send(wire::TakerToMaker::TakeOrder {
+                order_id,
+                quantity: self.quantity,
+            })
+            .await
+            .expect("connection actor to be up");
```
thomaseizinger: I would have used a different message and actor here.

If you take an address instead of a message channel to the `connection::Actor`, you can make a new message `StartContractSetup` or something that contains:

- Your own address
- The order ID
- The quantity

The connection actor can then send the first message on your behalf.

thomaseizinger (commented):
> @thomaseizinger: I tackled the improvements suggested and figured out what the bug was: I forgot to actually put anything in these two `setup_actors` hashmaps, so we could never get any messages from the maker to the corresponding `setup_taker::Actor`.
> 
> Looking into this bug made me think about a couple of things:
> 
>  I'm not too fond of using the `wire` message types for anything other than sending over the wire. I couldn't simply add the address of the `setup_taker::Actor` to the message that was already going to the `connection::Actor` because the same message needed to be serializable and relevant to the maker.

Agreed, I would introduce separate messages between the actors.

>  Isn't it a suspicious sign that we keep the same `setup_actors` state in two different actors? I may have followed your original branch a little blindly in that regard. Perhaps it would be better if only the `taker_cfd::Actor` kept this state and it was responsible for both ensuring that there's only one contract setup per order ID at any given moment and that messages from the maker go to the correct `setup_taker::Actor`.

While that might look better at the moment, it doesn't actually take responsibility away from the `taker_cfd::Actor`. We want to introduce _boundaries_ between the actors which means the `taker_cfd::Actor` cannot act as a proxy for messages.

The fact that the `connection` actor has to keep a HashMap of addresses is an artifact of us _not_ having a transport that supports multiplexing. If we would be able to open an isolated "channel" over the connection (for example https://docs.rs/quinn/0.8.0/quinn/struct.Connection.html#method.open_bi), then the connection actor would not need to act as a proxy but would only need to hand the channel to the setup actor and could otherwise stop caring about it.

That is what I want to get to long-term and for that, removing the proxying from the `taker_cfd::Actor` is a step forward :)

daemon/src/taker_cfd.rs
luckysori: Yes, you're right. In fact, I was already inserting it into the hashmap, just not calling `insert`. Don't let me work late at night.

daemon/src/setup_taker.rs
luckysori: I considered something like what you describe at first, but decided against it because I thought that taking an `Address` to the `connection::Actor` would force us to change the `taker_cfd::Actor` in ways that would break the actor tests. Admittedly, I didn't give it a go, so maybe I was wrong about that. I'll give it a go now.

luckysori:
@thomaseizinger: Does https://github.com/itchysats/itchysats/pull/684/commits/365ce8e33fbc59089b1319f0aca414e4c735fd41 do the trick for you?

thomaseizinger:
> @thomaseizinger: Does [365ce8e](https://github.com/itchysats/itchysats/commit/365ce8e33fbc59089b1319f0aca414e4c735fd41) do the trick for you?

I commented on the commit.

thomaseizinger (approved):
Nice! Thank you! :)

daemon/src/taker_cfd.rs
```diff
@@ -766,22 +720,39 @@ where
...
+        use setup_taker::Completed::*;
+        match msg {
+            NewContract { order_id, dlc } => {
+                log_error!(self.handle_new_contract(order_id, dlc))
+            }
+            Rejected { order_id } => log_error!(self.handle_order_rejected(order_id)),
+            Failed { order_id, error } => {
+                log_error!(self.handle_contract_setup_failed(order_id, error))
+            }
+        }
```
thomaseizinger: Actually, there is one more thing.

Can we _not_ split `Completed` into individual handlers here? With our new model, we will pass the entire result to the aggregate and no longer make this decision in the actor.

thomaseizinger: Sorry for only noticing this now!

da-kami (approved):
Great work!
As Thomas said it'd help us integrating the new model if we don't split up the contract setup completed in the handler.

daemon/src/taker_cfd.rs
luckysori: @thomaseizinger @da-kami: I think this is what you guys meant, but the difference is pretty small, so maybe you were referring to something else: https://github.com/itchysats/itchysats/pull/684/commits/d7f7d8f91292e11edea561832aafb3e170816d59

thomaseizinger: Nah that is perfect, thank you!

thomaseizinger (approved):

luckysori:
> bors r+

Oh, no, @da-kami. I hadn't tested locally yet and there was a `fixup` commit in there :(

thomaseizinger:
> > bors r+
> 
> Oh, no, @da-kami. I hadn't tested locally yet and there was a `fixup` commit in there :(

Not the end of the world, at least both build cleanly.

PR #687, opened 2021-11-23
Implement sqlx::Type for more datatypes

thomaseizinger:
This maker our DB code slightly simpler because we can directly use
our newtypes in the query and they are also constructed right away
by sqlx.

@da-kami @klochowicz Could one of you run `./prepare-db.sh` locally? It doesn't work on my machine for some reason ...

### DIFF
--- a/daemon/sqlx-data.json
+++ b/daemon/sqlx-data.json
@@ -1,101 +1,5 @@
 {
   "db": "SQLite",
-  "04399897350d026ee0830ccaba3638f8aa8f4ef9694d59286f32b9e2449a99fa": {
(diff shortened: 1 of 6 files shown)


thomaseizinger:
Thanks! Just squash it together please while you are at it! 😊

da-kami:
Done and pushed :)

Did you check that the temporary database was cleaned up completely? I had trouble with that in the past, where the `shm` and `wal` file remained and then the script would fail. (Note: Had other issues as well where it did not work, but was unable to reproduce them at the time...)

thomaseizinger:
Can I get a review as well @da-kami ? :D

da-kami (approved):

PR #690, opened 2021-11-24
Move settlement oracle event id into the Dlc

da-kami:
Oracle event-id and Dlc have to exist together, one without the other is incomplete, hence we move the event id inside.
Note that the ID is actually already know  to the Dlc as part of the `cet` hashmap, but this was not cleaned up in this patch.

This patch goes towards getting the model boundaries right to transition the architecture towards an event model.

Note: This would theoretically allow us to land auto-rollover, because we don't now pass the correct event-id through. However, we decided that it's better to pull the architecture refactoring through before landing the rollover feature 😬

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -434,17 +434,13 @@ where
         self.monitor_actor
             .send(monitor::StartMonitoring {
                 id: order_id,
-                params: MonitorParams::new(
-                    dlc,
-                    cfd.refund_timelock_in_blocks(),
-                    cfd.order.oracle_event_id,
-                ),
+                params: MonitorParams::new(dlc.clone(), cfd.refund_timelock_in_blocks()),
             })
             .await?;
 
         self.oracle_actor
             .send(oracle::MonitorAttestation {
-                event_id: cfd.order.oracle_event_id,
+                event_id: dlc.settlement_event_id,
             })
             .await?;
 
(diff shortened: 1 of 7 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -483,12 +479,6 @@ where
             .await?
             .with_context(|| format!("Announcement {} not found", cfd.order.oracle_event_id))?;
 
-        self.oracle_actor
```
da-kami, line 486: This was removed because it is just a weird duplicate that should not have been here. we trigger monitoring for attestation upon contract-setup/rollover completion, not when an order is accepted...

daemon/src/to_sse_event.rs
```diff
@@ -318,7 +318,10 @@ impl ToSseEvent for CfdsWithAuxData {
                     margin: cfd.margin().expect("margin to be available"),
                     margin_counterparty: cfd.counterparty_margin().expect("margin to be available"),
                     details,
-                    expiry_timestamp: cfd.expiry_timestamp(),
+                    expiry_timestamp: match cfd.expiry_timestamp() {
```
da-kami, line 321: This is to avoid "not sending a timestamp" to the UI - in case we don't have a Dlc yet it do use the one from the order here for display in the UI.

thomaseizinger (approved):

PR #693, opened 2021-11-24
Instead of persisting the order we only keep it in memory for the taker

da-kami:


da-kami:
cannot really do this in separate patch

PR #695, opened 2021-11-24
Remove `maker_cfd::CfdAction` enum

thomaseizinger:
Originally we thought we need this enum because we would need to use
`MessageChannel`s from the rocket layer. But we can actually fully
qualify the address without issues.

We also introduce `xtra_productivity` to remove some of the indirection
that is happening in this file.


luckysori (approved):
I remember we talked about this, nice!

PR #703, opened 2021-11-24
Explicitly drop database connection acquired during application setup

### LINKED ISSUE #477: Dirty daemon shutdown
We have reported this problem in the past, we still see errors like this when shutting down:

```
2021-11-04 18:20:53  INFO Received shutdown request. Waiting for pending I/O...
2021-11-04 18:20:55  WARN Wallet feed receiver not available, stopping wallet sync
2021-11-04 18:20:58 ERROR Closing connection to taker 907107d8-6ccb-4cce-9366-a477c9fd98da
2021-11-04 18:20:59 ERROR Server failed to shutdown cooperatively. Terminating.
```

On that particular shutdown the database was not properly merged, i.e. the `maker.sqlite-shm` and  `maker.sqlite-wal` file were present after terminating the application. 

Can we somehow ensure that the database always gets closed properly as part of the rocket shutdown? Or manage database outside of rocket, but ensure proper closing upon shutdown?

klochowicz:
The dirty shutdown we observed were caused by the sqlite pool patiently waiting
until we close the connection we've opened in the application setup.
This is turn was eventually triggering Rocket timeout (which is by default
configured to wait 5s for cleanup).

When loading cfds gets moved into the projection actor, the connection can be
moved to a tighter scope (only for the housekeeping functions).

Resolves #477

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -235,10 +235,7 @@ async fn main() -> Result<()> {
         .context("Db migrations failed")?;
 
     // Create actors
-    let mut conn = db.acquire().await?;
-
-    housekeeping::transition_non_continue_cfds_to_setup_failed(&mut conn).await?;
-    housekeeping::rebroadcast_transactions(&mut conn, &wallet).await?;
+    housekeeping::new(&db, &wallet).await?;
 
     let (projection_actor, projection_context) = xtra::Context::new(None);
 
@@ -275,7 +272,12 @@ async fn main() -> Result<()> {
     let (task, init_quote) = bitmex_price_feed::new(projection_actor).await?;
     tasks.add(task);
 
-    let cfds = load_all_cfds(&mut conn).await?;
+    // TODO: Move to projection actor
+    let cfds = {
+        let mut conn = db.acquire().await?;
+
+        load_all_cfds(&mut conn).await?
+    };
     let (cfd_feed_sender, cfd_feed_receiver) = watch::channel(cfds.clone());
     let (order_feed_sender, order_feed_receiver) = watch::channel::<Option<Order>>(None);
     let (update_cfd_feed_sender, update_cfd_feed_receiver) =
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
Fantastic, thanks @klochowicz !

daemon/src/maker.rs
```diff
@@ -336,6 +337,8 @@ async fn main() -> Result<()> {
         .launch()
         .await?;
 
+    std::mem::drop(conn);
```
thomaseizinger: Perhaps an explicit `conn.close()` might be even better?

Another way of solving this would be to limit the scope of `conn`:

```rust
let cfds = {

	load_all_cfds(&mut conn).await?
};
```

klochowicz: Limiting scope sounds best for me, I didn't opt for it initially as the connection was used for 2 distinct purposes - housekeeping and cfds. then it dawned at me that I could have just acquired 2 separate conns and make the scope much tighter!

daemon/src/lib.rs
```diff
@@ -67,6 +67,15 @@ pub const N_PAYOUTS: usize = 200;
...
 pub const SETTLEMENT_INTERVAL: time::Duration = time::Duration::hours(24);
 
+/// Perform necessary housekeeping before actor system startup
+pub async fn perform_housekeeping(db: &SqlitePool, wallet: &Address<wallet::Actor>) -> Result<()> {
+    let mut conn = db.acquire().await?;
+
+    housekeeping::transition_non_continue_cfds_to_setup_failed(&mut conn).await?;
+    housekeeping::rebroadcast_transactions(&mut conn, wallet).await?;
+    Ok(())
+}
```
thomaseizinger: Can we move this function to `housekeeping::new` and make the other two private?

klochowicz: great idea

thomaseizinger (approved):

PR #704, opened 2021-11-25
Move transaction signing onto `Dlc`

thomaseizinger:


da-kami (approved):
Yes! Thanks for doing that, I also had it in mind 😁

PR #707, opened 2021-11-25
Contract setup `Completed` is the same for taker and maker

da-kami:
Either of the roles has to record the outcome of the setup, which should be the same for them.
In order to handle this in our event model we need `Completed` in the model so it is not role specific.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1661,6 +1661,22 @@ impl CollaborativeSettlement {
     }
 }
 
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
ACK

Wait for @luckysori's approval before merge pls.

luckysori (approved):
LGTM!

thomaseizinger:
@klochowicz Seems there is still something not quite right with the timeouts :(

klochowicz:
bors retry

Issue #697, opened 2021-11-24
Move `to_sse_event` logic into projection actor


thomaseizinger:
Goal(s):

1. Tests assert on projected data
2. `to_sse_event::Cfd` becomes `projection::Cfd`
3. Messages sent to `projection::Actor` do not contain data (where possible) but just notify of change (i.e. `CfdsChanged`), `projection::Actor` can then load from DB by itself
4. Types in `watch::Receiver`s are `projection::*` types (sort of implicit through (1))

PR #708, opened 2021-11-25
Refactor projection actor before adding new functionality

klochowicz:
Create watch channels within the actor, returning a single struct called `Feeds`.
Manage the struct with Rocket and use it in SSE events.

Changing the watch channel types etc. will be much easier after this cleanup.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use crate::bitmex_price_feed::Quote;
 use crate::model::TakerId;
 use crate::{Cfd, Order, UpdateCfdProposals};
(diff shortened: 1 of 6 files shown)


klochowicz:
First part of #697 (actual changes will be built on top of this PR)

daemon/src/projection.rs
```diff
@@ -1,55 +1,82 @@
...
-            tx_order,
-            tx_quote,
-            tx_settlements,
-            tx_connected_takers,
-        }
+    pub fn new(init_cfds: Vec<Cfd>, init_quote: Quote) -> (Self, Feeds) {
+        let (tx_cfds, rx_cfds) = watch::channel(init_cfds);
+        let (tx_order, rx_order) = watch::channel::<Option<Order>>(None);
+        let (tx_update_cfd_feed, rx_update_cfd_feed) =
+            watch::channel::<UpdateCfdProposals>(HashMap::new());
```
thomaseizinger: Don't think you should need turbo fish here if you are directly assigning to fields.

thomaseizinger (approved):

daemon/src/projection.rs
klochowicz: indeed! changed :)

PR #709, opened 2021-11-25
Move Quote and CfdOrder projections inside projection actor

klochowicz:
This task is a major one, so I'm submitting in stages to make it easier for both reviewers and myself (hopefully less rebase conflicts)

Still need to be done:
- connected takers (will be replaced by "counterparties")
- Cfds (will incorporate all logic of CfdWithAuxData)

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -2,7 +2,7 @@ use std::time::Duration;
 
 use crate::harness::flow::{is_next_none, next, next_cfd, next_order, next_some};
 use crate::harness::{
-    assert_is_same_order, dummy_new_order, init_tracing, start_both, Maker, MakerConfig, Taker,
+    dummy_new_order, init_tracing, order_from_cfd, start_both, Maker, MakerConfig, Taker,
     TakerConfig,
 };
 use daemon::connection::ConnectionStatus;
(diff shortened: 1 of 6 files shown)


daemon/tests/happy_path.rs
```diff
@@ -46,8 +46,8 @@ async fn taker_takes_order_and_maker_rejects() {
     taker.take_order(received.clone(), Usd::new(dec!(10))).await;
 
     let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
-    assert_is_same_order(&taker_cfd.order, &received);
-    assert_is_same_order(&maker_cfd.order, &received);
+    assert_eq!(order_from_cfd(&taker_cfd), received);
+    assert_eq!(order_from_cfd(&maker_cfd), received);
```
thomaseizinger, line 50: Maybe just asserting the same ID is enough? Unit tests should / could cover the actual projection.

thomaseizinger (approved):

daemon/tests/happy_path.rs
klochowicz: perhaps, but I tried to just keep the tests the way they were. :) ideally, this should all disappear, we should really address the TODO and compare the Cfd on both sides.

klochowicz:
bors retry

PR #714, opened 2021-11-25
Make it clear what the payout is

### LINKED ISSUE #557: "At the current rate you would receive" does not match "Unrealized P/L"
This might be a user error on my side but I don't understand why these two metrics disagree:

image

P/L indicates that I am making a loss but right next to it, it says that I will be making a profit.
I also don't understand what "At the current rate" actually means. Does this refer to the exchange rate? Interest rate?

thomaseizinger:
image


bonomat (approved):
Step by step :)

da-kami (approved):
I thought it's always `margin +/- unrealized P/L = payout` - but it seems due to rounding this is sometimes not exact 😅

da-kami:
Note: one thing that "gets lost" with this change is, that this view reflects the payout at the current price. Not sure if that needs explicit stating. But, the fact that the payout "locks in" once we go for either settlement is not immediately clear. (Out of the scope of this PR though)

klochowicz:
bors retry

PR #716, opened 2021-11-26
Improve margin calc and quantity input

### LINKED ISSUE #688: Quantity input feels sluggish
The quantity input feels sluggish, that's because every input is being sent to the server and then updates the UI. 

Imho best is to do an initial calculation of the margin in the frontend and then, after clicking Long/Short, send to the backend and show the actual calculated numbers in the confirmation dialog.

thomaseizinger:
Quantity input change now looks like this:

image

Fixes #688.

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -11,7 +11,7 @@ import {
     VStack,
 } from "@chakra-ui/react";
 import * as React from "react";
-import { useEffect, useState } from "react";
+import { useState } from "react";
 import { Route, Routes } from "react-router-dom";
 import { useEventSource } from "react-sse-hooks";
 import useWebSocket from "react-use-websocket";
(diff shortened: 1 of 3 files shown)


taker-frontend/src/App.tsx
```diff
@@ -81,22 +82,23 @@ export const App = () => {
...
-            price: order.price,
-            quantity,
-        };
-        calculateMargin(payload);
-    }, // Eslint demands us to include `calculateMargin` in the list of dependencies.
-     // We don't want that as we will end up in an endless loop. It is safe to ignore `calculateMargin` because
-    // nothing in `calculateMargin` depends on outside values, i.e. is guaranteed to be stable.
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [margin, effectiveQuantity, order]);
+    useDebouncedEffect(
```
da-kami, line 85: 👍 nice one! We should use that more often

da-kami (approved):

PR #718, opened 2021-11-26
Rename `TakerId` to `Identity`

thomaseizinger:
With #665, we will use this struct on both maker and taker side.


klochowicz:
I was about to do the same locally ;D 

this will also be beneficial in the projection actor work (I'll be able remove the `maker_online_status feed`, and just use the same feed to send maker id instead)

klochowicz (approved):
👍

da-kami (approved):

PR #720, opened 2021-11-26
Increase the timeout in debug mode to fix CI/bors

klochowicz:
Contract setup test keeps failing when bors runs it.

Note: contract setup got recently refactored into a dedicated actor, this might
     have increased the time it takes to finish the process.

PR #722, opened 2021-11-26
Increase heartbeat timeout interval in debug mode to 45s

klochowicz:
As we do not process messages concurrently, heartbeat messages are not processed
during lengthy operations, such as contract setup.

The spam in the amount of commit messages is a bit unfortunate, but only failing
tests on various machines can indicate that the timing might cause issues.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -57,7 +57,7 @@ pub mod wire;
 // Increasing the value for debug mode makes sure that we don't cause problems
 // when testing / CI, whilst the release can still detect status faster
 pub const HEARTBEAT_INTERVAL: std::time::Duration =
-    Duration::from_secs(if cfg!(debug_assertions) { 30 } else { 5 });
+    Duration::from_secs(if cfg!(debug_assertions) { 45 } else { 5 });
 
 pub const N_PAYOUTS: usize = 200;
 


da-kami:
Bit confused the PR title says `60`s but you actually set it to `45`s - what is correct? 

Also, why this if we already got https://github.com/itchysats/itchysats/pull/720 in?

klochowicz:
> Bit confused the PR title says `60`s but you actually set it to `45`s - what is correct?
Thanks, I forgot to update it code to 60 :) I just went with something that would work for sure this time....

> 
> Also, why this if we already got #720 in?

#720 only increased a timeout it the waiting on the watch channel, nothing to do with the heartbeat processing. This one actually changes the behaviour of our daemons running in debug mode (not just tests), ie. the connection will be closed only after 120s of not seeing any heartbeats, which should be more than enough to deal with contract setup block :/

I have created an issue explaining what's the problem with the timeouts: #723

da-kami:
> > Bit confused the PR title says `60`s but you actually set it to `45`s - what is correct?
> > Thanks, I forgot to update it code to 60 :) I just went with something that would work for sure this time....
> 
> > Also, why this if we already got #720 in?
> 
> #720 only increased a timeout it the waiting on the watch channel, nothing to do with the heartbeat processing. This one actually changes the behaviour of our daemons running in debug mode (not just tests), ie. the connection will be closed only after 120s of not seeing any heartbeats, which should be more than enough to deal with contract setup block :/
> 
> I have created an issue explaining what's the problem with the timeouts: #723

Do I understand correctly, that this PR is a workaround for #723 until we add better support for concurrent messaging?

klochowicz:
> > > Bit confused the PR title says `60`s but you actually set it to `45`s - what is correct?
> > > Thanks, I forgot to update it code to 60 :) I just went with something that would work for sure this time....
> > 
> > 
> > > Also, why this if we already got #720 in?
> > 
> > 
> > #720 only increased a timeout it the waiting on the watch channel, nothing to do with the heartbeat processing. This one actually changes the behaviour of our daemons running in debug mode (not just tests), ie. the connection will be closed only after 120s of not seeing any heartbeats, which should be more than enough to deal with contract setup block :/
> > I have created an issue explaining what's the problem with the timeouts: #723
> 
> Do I understand correctly, that this PR is a workaround for #723 until we add better support for concurrent messaging?

yes. in fact, this is just a fine-tuning of the workaround that's already in place - payout curve calculations in debug mode take a while, the previous adjustments were good for local runs & CI, but not enough for bors :(

da-kami (approved):

thomaseizinger:
> As we do not process messages concurrently, heartbeat messages are not processed
during lengthy operations, such as contract setup.

I don't think this is necessarily true? Contract setup is its own task so I don't understand why it would be holding a lock on the actor.

We should be able to process messages concurrently. There is something fishy here!

klochowicz:
> > As we do not process messages concurrently, heartbeat messages are not processed
> > during lengthy operations, such as contract setup.
> 
> I don't think this is necessarily true? Contract setup is its own task so I don't understand why it would be holding a lock on the actor.
> 
> We should be able to process messages concurrently. There is something fishy here!

You're right - the problem was first observed when contract setup was not in its task yet, that's probably why I was lead the wrong trail here. This log last night might shed more light into this issue:

https://github.com/itchysats/itchysats/issues/723#issuecomment-979715200

thomaseizinger:
But it has been its own task from the beginning? We always spawned the future into the runtime, never blocking the handler.

PR #733, opened 2021-11-28
Create CloseButton component that non-collaboratively closes when the…

### LINKED ISSUE #623: Allow running taker with limited functionality without the maker online
Some functions should be still possible when the user cannot connect to the maker.

For example:
- withdrawing funds,
- checking the status of the open/closed CFDs,
- closing a CFD

The UI should clearly indicate that only a subset of normal actions is possible and that there is no connection to the maker.

rishflab:


taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+    let [settle, isSettling] = usePostRequest(`/api/cfd/${cfd.order_id}/settle`);
+    let [commit, isCommiting] = usePostRequest(`/api/cfd/${cfd.order_id}/commit`);
+
+    const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED
+        || !(cfd.state.key === StateKey.OPEN);
+
+    let closeButton;
+
+    if (connectedToMaker) {
+        closeButton = <Box w={"45%"}>
```
klochowicz: it's really hard to reason about code and maintain it when there's a lot of duplication inside it, with small changes (ie. which action gets triggered, the button name, and the text). IMHO it would be much nicer if you extracted the duplicated elements into a component, instead of a component that hides the duplication.

consider something like this:

```ts
if (connectedToMaker) {
   <CloseButton action={settle} [...] />
} 
   else {
  <CloseButton action={commit} [...] />
}

and then you wouldn't need to move the post requests inside the new component, as they're just passed in as props.

File: taker-frontend/src/components/CloseButton.tsx [hunk 2]
```diff
@@ -0,0 +1,129 @@
...
+                                size="sm"
+                                colorScheme="red"
+                                onClick={() => {
+                                    console.log(`Commiting CFD ${cfd.order_id}`);
+                                    commit({});
+                                    onClose();
+                                }}
+                                isLoading={isCommiting}
+                            >
+                                Non-Collaborative Close
```
klochowicz: as discussed, let's try to keep the size of the button relatively short - perhaps "Commit" would be enough?

klochowicz (approved):
thanks, it's a step forward towards the umbrel milestone.

taker-frontend/src/components/CloseButton.tsx
thomaseizinger: Agreed! Bonus-points for immutability:

```tsx
const closeButton = connectedToMaker ? <CloseButton action={settle} [...] /> : <CloseButton action={commit} [...] />;
```

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+                placement="bottom"
+                closeOnBlur={true}
+            >
+                {({ onClose }) => (<>
+                    <PopoverTrigger>
+                        <Button colorScheme={"blue"} disabled={disableCloseButton}>Commit</Button>
+                    </PopoverTrigger>
+                    <PopoverContent color="white" bg="blue.800" borderColor="blue.800">
+                        <PopoverHeader pt={4} fontWeight="bold" border="0">
+                            Non-Collaboratively Close your position
```
thomaseizinger: @bonomat sometimes calls this "force-close".

Thoughts? Don't need to discuss this at length now, the text can easily be changed later :)

bonomat: I prefer force-close but non-collaboratively close works as well.. no hard feelings here :)

taker-frontend/src/components/CloseButton.tsx
bonomat: I'd just call it `close`. I don't think it is clear to users what `commit` means.

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+        </Box>;
+    } else {
+        closeButton = <Box w={"45%"}>
+            <Popover
+                placement="bottom"
+                closeOnBlur={true}
+            >
+                {({ onClose }) => (<>
+                    <PopoverTrigger>
+                        <Button colorScheme={"blue"} disabled={disableCloseButton}>Commit</Button>
```
bonomat: I'd call it `Close` as I don't think users will understand the difference here. 
```suggestion
                        <Button colorScheme={"blue"} disabled={disableCloseButton}>Close</Button>
```

klochowicz: won't they be puzzled if they don't get they money until the next day? IMHO having the same button with the same name will eventually lead to users being unable to accurately create a mental map of software's functionality. Perhaps at least we could color-code it differently?

taker-frontend/src/components/CloseButton.tsx
rishflab: I think I will still have to pass in the `connectedToMaker` bool to decide which text is shown.  don't think it is a good idea to pass in `connectedToMaker` bool and the post request as it implies that you can settle when `connectedToMaker` is false.

I guess I could also pass this through but it might be a bit too much.

taker-frontend/src/components/CloseButton.tsx
rishflab: i went with force-close

taker-frontend/src/components/CloseButton.tsx
rishflab: i went with close and force-close

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+    cfd: Cfd;
+    action: (req: any) => void;
+    status: boolean;
+    buttonName: String;
+}
+
+/// Button that can be used to trigger non-collaborative close if the maker is offline and collaborative close if the
+/// maker is online.
+export const CloseButton = ({ cfd, action, status, buttonName }: CloseButtonProps) => {
+    let closeButton;
```
thomaseizinger: This variable is now pointless.

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+interface CloseButtonProps {
+    cfd: Cfd;
+    action: (req: any) => void;
+    status: boolean;
+    buttonName: String;
+}
+
+/// Button that can be used to trigger non-collaborative close if the maker is offline and collaborative close if the
+/// maker is online.
+export const CloseButton = ({ cfd, action, status, buttonName }: CloseButtonProps) => {
```
thomaseizinger: I think we should prefer the format of:

```tsx
export default function CloseButton({ cfd, action, status, buttonName }: CloseButtonProps) {

}
```

1. If there is only one component in the file, the default export should be that component
2. `function`s are declarations and can thus be `default` exported, `const`s are declarations and cannot
```

File: taker-frontend/src/components/CloseButton.tsx [hunk 7]
```diff
@@ -0,0 +1,79 @@
...
+    PopoverContent,
+    PopoverFooter,
+    PopoverHeader,
+    PopoverTrigger,
+    Text,
+} from "@chakra-ui/react";
+import * as React from "react";
+import { Cfd, StateGroupKey, StateKey } from "../types";
+
+interface CloseButtonProps {
```
thomaseizinger: Within the specific component, `Props` is unique and would be shorter.

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+            placement="bottom"
+            closeOnBlur={true}
+        >
+            {({ onClose }) => (<>
+                <PopoverTrigger>
+                    <Button colorScheme={"blue"} disabled={disableCloseButton}>{buttonName}</Button>
+                </PopoverTrigger>
+                <PopoverContent color="white" bg="blue.800" borderColor="blue.800">
+                    <PopoverHeader pt={4} fontWeight="bold" border="0">
+                        Close your position
```
thomaseizinger: This should also use `buttonName` (and perhaps be "action" then?).

thomaseizinger (approved):
Thanks! Couple of stylistic things but otherwise good.

thomaseizinger (approved):
Nice! Squash it all and merge :)

PR #738, opened 2021-11-29
Introduce `AddressMap` primitive

thomaseizinger:
This should remove some duplication as we will repeat this pattern
a couple of times.


luckysori (approved):
I like it!

PR #740, opened 2021-11-29
Move more types into projection actor

klochowicz:
A stepping stone to the PR with incorporating Cfds into the projection actor.


thomaseizinger (approved):
Ace!

PR #743, opened 2021-11-29
Move collaborative settlement protocol execution into individual actors

thomaseizinger:


thomaseizinger:
Tested locally, settling works. Rejecting settlement is still in testing.

thomaseizinger:
Tested rejecting of settlement successfully locally!

daemon/src/taker_cfd.rs
```diff
@@ -198,61 +187,78 @@ where
...
+        let (order_id, settlement) = match msg {
+            collab_settlement_taker::Completed::Confirmed {
+                order_id,
+                settlement,
+            } => (order_id, settlement),
+            collab_settlement_taker::Completed::Rejected { .. } => {
+                return Ok(());
+            }
+            collab_settlement_taker::Completed::Failed { order_id, error } => {
+                tracing::warn!(%order_id, "Collaborative settlement failed: {:#}", error);
```
luckysori, line 242: :question: I ask this for myself mainly: why did you choose `warn` over `error` here?

daemon/src/address_map.rs
```diff
@@ -49,3 +90,24 @@ impl<'a, K, A> Disconnected<'a, K, A> {
...
+mod tests {
+    use super::*;
+    use xtra::Context;
+
+    #[test]
+    fn gc_removes_address() {
+        let (addr, _) = Context::<Dummy>::new(None);
+        let mut map = AddressMap::default();
+        map.insert("foo", addr.clone());
+
```
klochowicz, line 106: it would be beneficial to have a `"bar" address there too, to show that it doesn't stop all addresses

daemon/src/collab_settlement_taker.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    async fn stopping(&mut self, ctx: &mut xtra::Context<Self>) -> xtra::KeepRunning {
+        // inform the connection actor that we stopping so it can GC the address from the hashmap
+        let me = ctx.address().expect("we are still alive");
+        let _ = self.connection.send(Stopping { me }).await;
+
+        xtra::KeepRunning::StopAll
+    }
```
luckysori, line 149: :+1: Didn't consider using this, nice!

daemon/src/collab_settlement_taker.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::model::Price;
+use crate::{connection, projection, wire};
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use xtra::prelude::MessageChannel;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor {
+    cfd: Cfd,
+    projection: xtra::Address<projection::Actor>,
```
luckysori, line 14: :question: Should we talk directly to the `projection::Actor` in the other setup actors then?

daemon/src/collab_settlement_taker.rs
```diff
@@ -0,0 +1,194 @@
...
+            wire::maker_to_taker::Settlement::Confirm => match self.handle_confirmed().await {
+                Ok(settlement) => Completed::Confirmed {
+                    settlement,
+                    order_id,
+                },
+                Err(e) => Completed::Failed { error: e, order_id },
+            },
+            wire::maker_to_taker::Settlement::Reject => {
+                if let Err(e) = self.handle_rejected().await {
+                    Completed::Failed { error: e, order_id }
```
luckysori, line 185: :thinking: A little annoying that handling a rejection can fail.

daemon/src/collab_settlement_taker.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        Ok(())
+    }
+
+    async fn handle_confirmed(&mut self) -> Result<CollaborativeSettlement> {
+        let order_id = self.cfd.order.id;
+
+        tracing::info!(%order_id, "Settlement proposal got accepted");
+
+        self.update_proposal(None).await?;
```
luckysori, line 71: :question: Is this supposed to take `None` as an argument here. You do the same for handling a rejection.

luckysori (approved):
LGTM!

daemon/src/collab_settlement_taker.rs
thomaseizinger: In this case I would say yes because the `taker_cfd::Actor` doesn't actually care about the ongoing settlement state other than to forward it to the `projection::Actor`.

daemon/src/collab_settlement_taker.rs
thomaseizinger: Yes, because in any case, the proposal is now "complete".

daemon/src/collab_settlement_taker.rs
thomaseizinger: We could actually make it non-fallible. If the `projection::Actor` goes away (which is the only error case of `send`), we don't need to fail the entire settlement.

daemon/src/taker_cfd.rs
thomaseizinger: Should probably be an `error`? :D

daemon/src/taker_cfd.rs
```diff
@@ -198,61 +187,78 @@ where
...
+        &mut self,
+        msg: ProposeSettlement,
+        ctx: &mut xtra::Context<Self>,
+    ) -> Result<()> {
         let ProposeSettlement {
             order_id,
             current_price,
         } = msg;
 
+        let disconnected = self
```
klochowicz, line 201: what does "disconnected" mean in this context?

daemon/src/projection.rs
```diff
@@ -133,21 +133,21 @@ impl State {
...
-            }
-            tracing::trace!(%order, "Removed cfd proposal");
-        };
-        Ok(())
+
+            return;
+        }
+
+        if self.proposals.remove(&order).is_none() {
+            tracing::trace!(%order, "Cannot remove cfd proposal: unknown");
```
klochowicz, line 145: Good change - the caller doesn't really care if the other part of the system misbehaved as it can't fix it anyway.

One thing though:
perhaps this should be at least `debug`, or even `error`? I'm a bit worried that if we bury this log message too deep we will not notice if our logic is flawed. Ideally, this should never happen.

thomaseizinger: Can make it debug yeah.

`error` I would only use if the answer to "Would I want an automated monitoring tool to wake me up at 3am if this happens?" is yes 😁

daemon/src/taker_cfd.rs
thomaseizinger: It is an instance of `address_map::Disconnected`. Do you have an idea for a better name?

daemon/src/taker_cfd.rs
thomaseizinger: Actually: https://github.com/itchysats/itchysats/pull/743#discussion_r759831579.

daemon/src/taker_cfd.rs
klochowicz: hard to think of a better one inded. If I find one, I'll propose something!

daemon/src/projection.rs
klochowicz: debug sounds good 👍

klochowicz (approved):
nice! 👍

thomaseizinger:
bors retry

PR #745, opened 2021-11-29
Refactor `maker_inc_connections::Actor`

luckysori:
I'm clearly doing something wrong because the contract setup actor test doesn't pass. For some reason I'm not seeing the solution clearly even though we talked about it, @thomaseizinger.


daemon/src/maker_inc_connections.rs
```diff
@@ -106,39 +129,32 @@ impl Actor {
             .map_ok(move |msg| FromTaker { taker_id, msg })
             .map(forward_only_ok::Message);
 
-        let (out_msg_actor_address, mut out_msg_actor_context) = xtra::Context::new(None);
-
-        let (forward_to_cfd, forward_to_cfd_fut) =
+        let (forward_only_ok, forward_only_ok_fut) =
             forward_only_ok::Actor::new(self.taker_msg_channel.clone_channel())
```
luckysori: I can already see that I meant for the `forward_only_ok::Actor` to forward to the `maker_inc_connections::Actor` (i.e. this one), so this is wrong.

daemon/src/maker_inc_connections.rs
```diff
@@ -106,39 +129,32 @@ impl Actor {
...
-
-        self.write_connections
-            .insert(taker_id, out_msg_actor_address);
+        let heartbeat_interval = self.heartbeat_interval;
+        let heartbeat_fut = out_msg_actor_context
+            .notify_interval(heartbeat_interval, || wire::MakerToTaker::Heartbeat)
+            .expect("actor not to shutdown");
+        taker_tasks.add(heartbeat_fut);
+
+        out_msg_actor_context.run(send_to_socket_actor).await;
```
thomaseizinger: This is blocking but must not block the current future. Needs to be a separate task.

daemon/src/maker_inc_connections.rs
thomaseizinger: I am not sure if we still need `forward_only_ok` now.

daemon/src/maker_inc_connections.rs
```diff
@@ -106,39 +129,32 @@ impl Actor {
...
-        let taker_disconnected_channel = self.taker_disconnected_channel.clone_channel();
-        self.tasks.add(async move {
-            let mut actor = send_to_socket::Actor::new(write, transport_state.clone());
-
-            let _heartbeat_handle = out_msg_actor_context
-                .notify_interval(heartbeat_interval, || wire::MakerToTaker::Heartbeat)
-                .expect("actor not to shutdown")
-                .spawn_with_handle();
+        taker_tasks.add(forward_only_ok_fut);
+        taker_tasks.add(forward_only_ok.clone().attach_stream(read));
```
thomaseizinger: I would probably not use `attach_stream` and implement a loop myself. You'll need it once you want to dispatch to different addresses based on the message type.

daemon/src/maker_inc_connections.rs
luckysori: I too wasn't sure, but it is useful to track when the reading end of the stream fails: the actor will go down and we can react accordingly by dropping all the tasks related to that taker ID.

daemon/src/maker_inc_connections.rs
luckysori: This was the bug, thank you!

daemon/src/maker_inc_connections.rs
luckysori: > You'll need it once you want to dispatch to different addresses based on the message type.

Is that so? I think we're sending the filtered messages back to this actor, who can now handle them and dispatch to different addresses.

But I agree that there must be a more direct way of doing this that doesn't involve this `attach_stream` and the `forward_only_ok::Actor` in general.

thomaseizinger: Maybe I am not seeing the big picture, I'll checkout the code and have a look!

daemon/src/maker_cfd.rs
```diff
@@ -439,9 +439,12 @@ where
         // state change. Once we know that we go for either an accept/reject scenario we
         // have to remove the current order.
         self.current_order_id = None;
-        self.takers
-            .send(maker_inc_connections::BroadcastOrder(None))
-            .await?;
+
+        // TODO: If I don't comment this out we hang forever!
```
thomaseizinger: Probably another dead-lock! Either this one or the `send` within `maker_inc_connections` where you dispatch messages needs to be `do_send_async`.

daemon/src/maker_inc_connections.rs
thomaseizinger: Yeah I think there is quite some unnecessary indirection here (probably introduced by me :joy: ).

I reckon take a step back and see if you can implement all of this functionality in a simpler way. In a nutshell what we need is:

- Forward messages from the `read` side to the `taker_msg_channel`
- Drop the connections on error on either the read or writing side
- Send a heart-beat every X seconds
- Make sure to stop the heartbeat on errors

daemon/src/maker_cfd.rs
luckysori: How could I have known that a deadlock could occur here? What should I be looking out for?

daemon/src/maker_cfd.rs
```diff
@@ -439,9 +439,11 @@ where
         // state change. Once we know that we go for either an accept/reject scenario we
         // have to remove the current order.
         self.current_order_id = None;
+
         self.takers
-            .send(maker_inc_connections::BroadcastOrder(None))
+            .do_send_async(maker_inc_connections::BroadcastOrder(None))
```
thomaseizinger, line 448: This should need a `allow(clippy::disallowed_method)`?

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
...
+        let this = ctx.address().expect("self to be alive");
+        let taker_msg_channel = self.taker_msg_channel.clone_channel();
+        let read_fut = async move {
+            while let Ok(res) = read.try_next().await {
+                match res {
+                    Some(msg) => taker_msg_channel
+                        .send(msg)
+                        .await
+                        .expect("connection to taker to exist"),
+                    None => continue,
```
thomaseizinger: `None` means the stream is gone, I think we should `break` here.

thomaseizinger: Which means I think you can collapse the `while let` into `while let Ok(Some(msg))`.

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
         let transport_state = Arc::new(Mutex::new(transport_state));
 
         let (read, write) = stream.into_split();
-        let read = FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()))
-            .map_ok(move |msg| FromTaker { taker_id, msg })
-            .map(forward_only_ok::Message);
+        let mut read =
+            FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()))
+                .map_ok(move |msg| FromTaker { taker_id, msg });
```
thomaseizinger: I think now that we are not using the `Stream` with `attach_stream`, it would be more idiomatic to construct the `FromTaker` message just upon `taker_msg_channel.send()`.

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
...
+
+        let this = ctx.address().expect("self to be alive");
+        let taker_msg_channel = self.taker_msg_channel.clone_channel();
+        let read_fut = async move {
+            while let Ok(res) = read.try_next().await {
+                match res {
+                    Some(msg) => taker_msg_channel
+                        .send(msg)
+                        .await
+                        .expect("connection to taker to exist"),
```
thomaseizinger: What do you think of using `break` here as well? We have a sane error handling strategy here, could use it?

daemon/src/maker_inc_connections.rs
```diff
@@ -186,6 +192,15 @@ impl Actor {
             }
         }
     }
+
+    async fn handle_read_fail(&mut self, msg: ReadFail, _: &mut Context<Self>) {
```
thomaseizinger: The macro can handle fn definitions without `Context`.

```suggestion
    async fn handle_read_fail(&mut self, msg: ReadFail) {
```

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
...
-        let read = FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()))
-            .map_ok(move |msg| FromTaker { taker_id, msg })
-            .map(forward_only_ok::Message);
+        let mut read =
+            FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()))
+                .map_ok(move |msg| FromTaker { taker_id, msg });
+
+        let this = ctx.address().expect("self to be alive");
+        let taker_msg_channel = self.taker_msg_channel.clone_channel();
+        let read_fut = async move {
```
thomaseizinger, line 117: Not sure if much cleaner but we could inline this variable?

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
...
-        let (forward_to_cfd, forward_to_cfd_fut) =
-            forward_only_ok::Actor::new(self.taker_msg_channel.clone_channel())
-                .create(None)
-                .run();
-        self.tasks.add(forward_to_cfd_fut);
+        let (out_msg, mut out_msg_actor_context) = xtra::Context::new(None);
+        let send_to_socket_actor = send_to_socket::Actor::new(write, transport_state.clone());
 
-        // only allow outgoing messages while we are successfully reading incoming ones
         let heartbeat_interval = self.heartbeat_interval;
```
thomaseizinger: Can we inlined, no longer needs to be separate because the `async move` is gone.

daemon/src/maker_inc_connections.rs
```diff
@@ -92,8 +99,10 @@ impl Actor {
         &mut self,
         mut stream: TcpStream,
         taker_address: SocketAddr,
-        _: &mut Context<Self>,
+        ctx: &mut Context<Self>,
     ) -> Result<()> {
+        let mut taker_tasks = Tasks::default();
```
thomaseizinger: Perhaps?

```suggestion
        let mut connection_tasks = Tasks::default();
```

daemon/src/maker_inc_connections.rs
```diff
@@ -102,43 +111,40 @@ impl Actor {
...
+        let heartbeat_fut = out_msg_actor_context
+            .notify_interval(heartbeat_interval, || wire::MakerToTaker::Heartbeat)
+            .expect("actor not to shutdown");
+        taker_tasks.add(heartbeat_fut);
 
-            let _heartbeat_handle = out_msg_actor_context
-                .notify_interval(heartbeat_interval, || wire::MakerToTaker::Heartbeat)
-                .expect("actor not to shutdown")
-                .spawn_with_handle();
+        taker_tasks.add(out_msg_actor_context.run(send_to_socket_actor));
```
thomaseizinger: Could we re-arrange this so that all `tasks.add` are in the same place? I think that would be more readable.

thomaseizinger: Do we have a constructor on `Tasks` that takes a `Vec`? Could remove the mutability of `tasks` then and construct and store it in the `HashMap` in one go.

thomaseizinger (approved):
Some ideas, take as you wish :)

daemon/src/maker_inc_connections.rs
luckysori: Good to know, I had a suspicion!

daemon/src/maker_cfd.rs
luckysori: Clippy doesn't complain :O

daemon/src/maker_inc_connections.rs
luckysori: Good point!

daemon/src/maker_inc_connections.rs
luckysori: Nice!

daemon/src/maker_inc_connections.rs
luckysori: Yep, you're right!

daemon/src/maker_cfd.rs
thomaseizinger: 1. Actors need a `&mut self` to invoke a handler
2. Sending a message invokes a handler

This is the important code-part in xtra: https://github.com/Restioson/xtra/blob/224004bc95dac12b8b8d2a14fea5ddf37eef6ae0/src/context.rs#L215-L249

If `.tick` never returns, it cannot read any more messages. But if a handler can only complete by returning from a `.send` call, it can never return. It would be unsound if two handlers were running concurrently because they both receive `&mut self`.

daemon/src/maker_inc_connections.rs
luckysori: Something like this would be nice and compiles

```rust
pub fn new(fs: Vec<impl Future<Output = ()> + Send + 'static>) -> Self {
    let handles = fs.into_iter().map(|f| f.spawn_with_handle()).collect();
    Self(handles)
}
```

but in practice

```
distinct uses of `impl Trait` result in different opaque types
```

Any way we can get around that?

thomaseizinger: You'd have to `Box` which is quite inconvenient.

Could do an `trait IntoHandles` that converts to a `Vec<RemoteHandle>` and is implement for `(A: Future)`, `(A: Future, B: Future)`, etc. Then you could pass in tuples of various sizes ...

luckysori: I'm going to pass on this one, but it's an interesting idea.

daemon/src/maker_cfd.rs
thomaseizinger: Can you add it anyway? I might be due to some cached artifacts and I don't want to blow up later down the line!

With a comment about the dead lock please.

thomaseizinger (approved):

PR #746, opened 2021-11-29
Use UI types in the projection actor

klochowicz:


daemon/src/projection.rs
```diff
@@ -87,7 +83,7 @@ struct State {
 }
 
 impl State {
-    pub fn to_cfd(&self) -> Vec<Cfd> {
+    pub fn as_cfds(&self) -> Vec<Cfd> {
```
thomaseizinger: Nit: `as_` usually means "cheap conversion of references". `to_cfds` was correct (reference to owned conversion).

See https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv.

thomaseizinger (approved):
Nice, thank you!

daemon/src/projection.rs
klochowicz: thank you, I'll keep it in mind!

thomaseizinger (approved):

PR #761, opened 2021-11-30
Introduce a 10s connect timeout

thomaseizinger:
- Cleanup logs for establishing connection
- Introduce a 10s timeout for connecting to the maker


da-kami (approved):

bonomat (approved):
10s should be plenty. Thanks. 👍

PR #763, opened 2021-11-30
Refactor bitmex price feed to Actor that reconnects on error

thomaseizinger:


daemon/src/bitmex_price_feed.rs
```diff
@@ -1,46 +1,106 @@
...
+            tasks: Tasks::default(),
+            receiver: Box::new(receiver),
+        }
+    }
+
+    async fn connect_until_successful(&self, ctx: &mut xtra::Context<Self>) {
+        let this = ctx.address().expect("we are alive");
+
+        while let Err(e) = this
+            .send(Connect)
```
da-kami: This can result in a deadlock if we send the message to ourselves.

da-kami (approved):
thanks!
works :)

PR #765, opened 2021-11-30
Setup maker actor

luckysori:
Tests are still failing, but I think it's close, so take a look if you can.

### DIFF
--- a/daemon/src/address_map.rs
+++ b/daemon/src/address_map.rs
@@ -29,6 +29,13 @@ where
         Ok(Disconnected { entry })
     }
 
+    pub fn get_connected(&self, key: &K) -> Option<&xtra::Address<A>> {
+        match self.inner.get(key) {
+            Some(addr) if addr.is_connected() => Some(addr),
+            _ => None,
+        }
+    }
+
     /// Garbage-collect an address that is no longer active.
     pub fn gc(&mut self, stopping: Stopping<A>) {
         self.inner.retain(|_, candidate| stopping.me != *candidate);
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -1,3 +1,4 @@
+use crate::address_map::{AddressMap, Stopping};
 use crate::cfd_actors::{self, append_cfd_state, insert_cfd_and_send_to_feed};
 use crate::db::{insert_order, load_cfd_by_order_id, load_order_by_id};
 use crate::model::cfd::{
@@ -10,10 +11,11 @@ use crate::projection::{
     try_into_update_rollover_proposal, try_into_update_settlement_proposal, Update,
     UpdateRollOverProposal, UpdateSettlementProposal,
 };
-use crate::setup_contract::{RolloverParams, SetupParams};
+use crate::setup_contract::RolloverParams;
 use crate::tokio_ext::FutureExt;
 use crate::{
-    log_error, maker_inc_connections, monitor, oracle, projection, setup_contract, wallet, wire,
+    log_error, maker_inc_connections, monitor, oracle, projection, setup_contract, setup_maker,
+    wallet, wire, Tasks,
 };
 use anyhow::{Context as _, Result};
 use async_trait::async_trait;
@@ -27,6 +29,7 @@ use sqlx::Sqlite;
 use std::collections::{HashMap, HashSet};
 use time::Duration;
 use xtra::prelude::*;
+use xtra::Actor as _;
 use xtra_productivity::xtra_productivity;
 
 pub struct AcceptOrder {
@@ -64,11 +67,6 @@ pub struct TakerDisconnected {
     pub id: Identity,
 }
 
-pub struct CfdSetupCompleted {
-    pub order_id: OrderId,
-    pub dlc: Result<Dlc>,
-}
-
 pub struct CfdRollOverCompleted {
     pub order_id: OrderId,
     pub dlc: Result<Dlc>,
@@ -93,7 +91,7 @@ pub struct Actor<
     takers: Address<T>,
     current_order_id: Option<OrderId>,
     monitor_actor: Address<M>,
-    setup_state: SetupState,
+    setup_actors: AddressMap<OrderId, setup_maker::Actor>,
     roll_over_state: RollOverState,
     oracle_actor: Address<O>,
     // Maker needs to also store Identity to be able to send a reply back
@@ -101,15 +99,7 @@ pub struct Actor<
     current_agreed_proposals: HashMap<OrderId, (SettlementProposal, Identity)>,
     connected_takers: HashSet<Identity>,
     n_payouts: usize,
-}
-
-enum SetupState {
-    Active {
-        taker: Identity,
-        sender: mpsc::UnboundedSender<wire::SetupMsg>,
-        _task: RemoteHandle<()>,
-    },
-    None,
+    tasks: Tasks,
 }
 
 enum RollOverState {
@@ -143,13 +133,14 @@ impl<O, M, T, W> Actor<O, M, T, W> {
             takers,
             current_order_id: None,
             monitor_actor,
-            setup_state: SetupState::None,
+            setup_actors: AddressMap::default(),
             roll_over_state: RollOverState::None,
             oracle_actor,
             current_pending_proposals: HashMap::new(),
             current_agreed_proposals: HashMap::new(),
             n_payouts,
             connected_takers: HashSet::new(),
+            tasks: Tasks::default(),
         }
     }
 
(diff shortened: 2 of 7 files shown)


daemon/src/address_map.rs
```diff
@@ -27,6 +27,13 @@ where
 
         Ok(Disconnected { entry })
     }
+
+    pub fn get(&self, key: &K) -> Option<&xtra::Address<A>> {
```
thomaseizinger: `get_connected` perhaps?

daemon/src/maker_cfd.rs
```diff
@@ -467,122 +469,80 @@ where
         // Since rejection is tied to the cfd state at the moment we can only do this after creating
         // a cfd.
         if quantity < current_order.min_quantity || quantity > current_order.max_quantity {
```
thomaseizinger: What do you think of moving this check also into the `setup_maker` actor?

daemon/src/maker_inc_connections.rs
```diff
@@ -41,6 +55,7 @@ pub struct Actor {
     taker_msg_channel: Box<dyn MessageChannel<FromTaker>>,
     noise_priv_key: x25519_dalek::StaticSecret,
     heartbeat_interval: Duration,
+    setup_actors: HashMap<OrderId, xtra::Address<setup_maker::Actor>>,
```
thomaseizinger: I think we may also need a message back to this actor to clean this up, otherwise we have unbounded growth of this hashmap.

daemon/src/setup_maker.rs
```diff
@@ -0,0 +1,238 @@
...
+
+            self.on_completed
+                .send(Completed::Failed { order_id, error })
+                .await
+                .expect("maker_cfd actor to be available")
+                .expect("maker_cfd to handle failed");
+
+            ctx.stop();
+            return;
+        }
```
thomaseizinger: Perhaps worth extracting this into a function that returns `Result` and only doing the `if let Err` dance once.

daemon/src/maker_inc_connections.rs
```diff
@@ -113,10 +129,9 @@ impl Actor {
             FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()));
 
         let this = ctx.address().expect("self to be alive");
-        let taker_msg_channel = self.taker_msg_channel.clone_channel();
         let read_fut = async move {
             while let Ok(Some(msg)) = read.try_next().await {
-                let res = taker_msg_channel.send(FromTaker { taker_id, msg }).await;
+                let res = this.send(FromTaker { taker_id, msg }).await;
```
thomaseizinger, line 135: Not sure if worth this indirection? Could `match` on `msg` here as well.

daemon/src/maker_cfd.rs
```diff
@@ -467,122 +469,80 @@ where
...
-        // Use `.send` here to ensure we only continue once the message has been sent
-        // Nothing done after this call should be able to fail, otherwise we notified the taker, but
-        // might not transition to `Active` ourselves!
-        self.takers
-            .send(maker_inc_connections::TakerMessage {
-                taker_id,
-                msg: wire::MakerToTaker::ConfirmOrder(cfd.order.id),
-            })
-            .await??;
+        tracing::info!(%order_id, "Maker accepts an order" );
```
thomaseizinger: Can we move this to the `setup_maker::Actor`?

thomaseizinger (commented):
Nice work so far!

Some comments!

daemon/src/maker_cfd.rs
luckysori: It felt weird to spawn an actor to let it "fail" immediately, but I guess the `maker_cfd` actor shouldn't even be involved in all this anyway.

thomaseizinger: https://github.com/Restioson/xtra/blob/224004bc95dac12b8b8d2a14fea5ddf37eef6ae0/src/context.rs#L194-L200 :man_shrugging:

luckysori: Hahaha, I feel seen.

daemon/src/setup_maker.rs
luckysori: I agree. I wish we had try blocks: https://github.com/rust-lang/rust/issues/31436.

thomaseizinger: In the mean-time, you could abuse an `async` for it :)

daemon/src/maker_inc_connections.rs
luckysori: That was my original intention, but at the time of handling a new connection the `setup_maker` actor doesn't exist. I think we would need to pass in a `Arc<Mutex<AddressMap>>` to this future, which would be clunky anyway.

daemon/src/maker_inc_connections.rs
luckysori: I copied your solution :D Should we do the same for the CFD actors?

thomaseizinger: I think so yes.

thomaseizinger: I think there might even be an abstraction where the entire `AddressMap` is an actor so this pattern of "cleaning up after the actor is dead" is encapsulated. Something to mull over ...

daemon/src/maker_inc_connections.rs
thomaseizinger: Ah yes, good point.

daemon/src/maker_cfd.rs
```diff
@@ -471,130 +484,81 @@ where
...
             Cfd {
-                state: CfdState::IncomingOrderRequest { taker_id, .. },
+                state: CfdState::IncomingOrderRequest { .. },
                 ..
-            } => taker_id,
+            } => {}
             _ => {
                 anyhow::bail!("Order is in invalid state. Ignoring trying to accept it.")
             }
         };
```
thomaseizinger: I think we can drop this now, the state within the actor is not going to change!

daemon/src/maker_cfd.rs
```diff
@@ -471,130 +484,81 @@ where
...
-        // Nothing done after this call should be able to fail, otherwise we notified the taker, but
-        // might not transition to `Active` ourselves!
-        self.takers
-            .send(maker_inc_connections::TakerMessage {
-                taker_id,
-                msg: wire::MakerToTaker::ConfirmOrder(cfd.order.id),
-            })
-            .await??;
+        // 2. Forward accept message to setup_maker::Actor
+        setup_actor.send(setup_maker::Accepted).await?;
```
thomaseizinger: You can use `AddressMap::send` for this!

daemon/src/maker_cfd.rs
```diff
@@ -1060,6 +960,64 @@ where
...
 }
 
+#[xtra_productivity]
+impl<O, M, T, W> Actor<O, M, T, W>
+where
+    O: xtra::Handler<oracle::MonitorAttestation>,
+    M: xtra::Handler<monitor::StartMonitoring>,
+    W: xtra::Handler<wallet::TryBroadcastTransaction>,
+{
+    async fn handle_setup_completed(&mut self, msg: setup_maker::Completed) -> Result<()> {
```
thomaseizinger: There is a lot that can go wrong here, please use `log_error` for the time being until we have something better.

Note that you are also dropping this error within `setup_maker::Actor` so it won't show up anywhere.

daemon/src/setup_maker.rs
```diff
@@ -0,0 +1,278 @@
...
+        (taker, confirm_order, taker_id): (
+            &(impl MessageChannel<maker_inc_connections::TakerMessage> + 'static),
+            &(impl MessageChannel<maker_inc_connections::ConfirmOrder> + 'static),
+            Identity,
+        ),
+        on_completed: &(impl MessageChannel<Completed> + 'static),
+        (on_stopping0, on_stopping1): (
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+        ),
```
thomaseizinger, line 48: Could use a `fan_out::Actor` here.

daemon/src/setup_maker.rs
```diff
@@ -0,0 +1,278 @@
...
+            };
+
+            Ok(())
+        });
+
+        Ok(())
+    }
+
+    async fn complete(&mut self, completed: Completed, ctx: &mut xtra::Context<Self>) {
+        let _ = self.on_completed.send(completed).await;
```
thomaseizinger, line 117: We are ignoring the error here. I think we should either use `log_error` on the other side or at least not drop the `Result` here and log it :)

thomaseizinger (approved):
Thank you!

Couple of comments, nothing blocking though so we can do these in follow-ups if you want!

daemon/src/setup_maker.rs
luckysori: I think you are right, but you may need to do the same in `collab_settlement_taker.rs`, since I copied this snippet from there.

thomaseizinger (approved):
Nice! Let's get it in.

PR #766, opened 2021-11-30
Rewrite docker container and binary release process 

### LINKED ISSUE #537: Version displayed on startup is not correct
```
2021-11-11 22:26:32  INFO Running version: 0.1.1
```

I am fairly certain I am running a version that is _newer_ than the 0.1.1 tag (I pulled the `preview` docker tag) but this is not reflected in the version number.

### LINKED ISSUE #735: Re-design release process
With the introduction of publishing docker images, our release process might need some refinements.

Originally, the idea was to publish pre-built binaries via releases and for that reason a `preview` release was introduced. However, this was more of a duct-tape solution than anything else as can be seen by the number of workarounds that we need to put in place for it to work correctly:

- https://github.com/itchysats/itchysats/pull/734
- https://github.com/itchysats/itchysats/blob/697771e71550b06148169e0c53fb2db20960670e/.github/workflows/preview-release.yml#L23

Both of these stem from the fact that tags are _meant_ to be immutable but our `preview` tag is not.

Unfortunately, the current process of releasing a `latest` docker image depends on the release binaries being built upon merges to master and attaching them to the `preview` release. The same workflow is also used to attach binaries to actual releases.

A re-designed process should take into account:

- No tag hacks like the above
- Ideally no duplication between workflows

Feature-wise, I think the bare minimum we need is:

- Updating the `latest` docker image every time we push to master
- Releasing a versioned docker tag upon releases (like 0.1.2)

With Umbrel becoming our main target user base, I think we can ditch attaching release binaries to releases.

Another thing that is currently not accounted for in our release process is doing releases from hotfix branches. Our automation for making a release there doesn't currently work because it always targets the master branch, regardless of which branch it was run from.

What would also be nice is if we would automatically tag an image with the minor version. i.e. releasing an image of version `0.3.6` should update the `0.3` tag so users can stay up to date with the latest compatible minor version without having to specify patch!

A major pain-point that I'd like to solve with this is that our current release process waits for all binaries to be built although for the docker images, we only need the linux binaries. In cases such as [this one](https://github.com/itchysats/itchysats/runs/5015035657?check_suite_focus=true), this resulted in a delay of 30 minutes until the change is deployed.

Thoughts welcome!

thomaseizinger:
Fixes #735. (Most of it at least)

The commit message summarizes most of the details but there is something else that we need to discuss:

With this patch, the container released from `master` is no longer tagged as `latest` but instead it is tagged as `master`. That is a conscious decision. After doing some research, I came to the conclusion that the `latest` tag in docker should likely point to the latest _stable_ release. Our `master` tag is absolutely bleeding edge so I don't think we should be using `latest` for that.

This patch actually does not tag any docker container as `latest`. It only tags releases with their according version number and latest master as `master`. I think that covers all the requirements we currently have and I would like to defer the tagging of a container as `latest` to a follow-up ticket.

You can see successful testruns here: https://github.com/thomaseizinger/itchysats/actions


thomaseizinger:
Here is an example on how to build multi-arch images step by step: https://github.com/bwt-dev/bwt/blob/master/scripts/docker-release.sh

This should be useful in constructing a new release workflow where we use github's build matrix to build all the containers in parallel and then simply finish with one manifest creation.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,79 @@
...
+        uses: docker/login-action@v1.11.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build docker image
+        run: |
+          docker build --platform linux/${{ matrix.docker_arch }} --pull --build-arg BINARY_PATH=./target/${{ matrix.target }}/release/${{ matrix.bin }} --tag ghcr.io/${{ github.repository }}/${{matrix.bin}}:${{ inputs.tag }}-${{ matrix.docker_arch }} .
+          docker push ghcr.io/${{ github.repository }}/${{matrix.bin}}:${{ inputs.tag }}-${{ matrix.docker_arch }}
```
bonomat, line 56: I'm not 100% convinced about the approach of building a separate docker image for each binary and then combine it using manifest. 

What's the value we get from this? I don't think it is speed. Because you still have to wait for all images to build before you can combine them using the `create_manifest_image` step. 
Building the docker images using buildx is quite fast and I never thought of this being a bottleneck. The only thing we do is copy the binary into the container, and this does not take much time. 

The problem I see is that we might have half-backed images out there. E.g. building `amd64` works but for `arm64` fails. Then we have one container image `maker-amd64` but not the one for `arm64` and not the combined one.
Are we happy with this?

bonomat (commented):
Nice improvement using composite actions. I'm not sure about the split of building images and combining them in two steps is worth it.

.github/workflows/docker.yml
thomaseizinger: > What's the value we get from this? I don't think it is speed. Because you still have to wait for all images to build before you can combine them using the `create_manifest_image` step.
> Building the docker images using buildx is quite fast and I never thought of this being a bottleneck. The only thing we do is copy the binary into the container, and this does not take much time.

Correct! I did not optimize for speed but for simplicity!

- The previous approach was tricky to understand because we had to copy the binaries into a particular directory structure to allow `buildx`'s internall parallelism to work.
- This also leaked into the Dockerfile definition where it was unclear, what these build parameters do.
- The buildx approach also depended on being able to download the binaries from the _preview_ release which was a legacy concept that we were no longer actively utilizing and had its own tech-debt built up. This could be mitigated by using the `actions/{upload,download}-artifact` actions but that alone also adds complexity to the workflow again.

I think the buildx approach is really powerful and simple if the entire build process is within the Dockerfile. Then building a multi-arch image becomes a one-liner. We don't do that for performance reasons so it IMO loses a lot of its appeal.

We already have a build-matrix for building all the different binaries so it felt natural to me to just tuck the building of the specific container on top of that. The convenience of a multi-arch image is then just something we add on top, once all the pipelines are done!

> The problem I see is that we might have half-backed images out there. E.g. building `amd64` works but for `arm64` fails. Then we have one container image `maker-amd64` but not the one for `arm64` and not the combined one.
> Are we happy with this?

I am obviously biased towards the proposed solution but also objectively considered, I don't think it is an actual issue. We can delete containers if we want or we trigger the pipeline again if it was a ephemeral error.

klochowicz: I don't have any  objections - it is a bit "cleaner" approach, and our `Dockerfile` is extremely simple, so I wouldn't expect many errors that leave us in a wrong state (and re-triggering the pipeline should do the trick in these cases, as observed above).

.github/workflows/build-release-binary.yml
```diff
@@ -1,12 +1,12 @@
-name: "Build release binary"
+name: "Attach binaries to GitHub release"
```
klochowicz, line 1: nit: "Build & attach"? (if it was named "Attach", I'd have some troubles finding it today)

klochowicz (commented):
as I haven't had much to do with multi-arch images - will this have negative impact on local storage? Is Docker smart enough to only download the parts that matter to the target platform?

.github/workflows/docker.yml
bonomat: Imho we exchanged one complexity for a different one, but that's fine with me.
I'm happy to go with your proposal and if we ever have partially built docker images just have to clean them up manually 👍

bonomat (approved):

thomaseizinger:
> as I haven't had much to do with multi-arch images - will this have negative impact on local storage? Is Docker smart enough to only download the parts that matter to the target platform?

To clarify, we are already using multi-arch images but it is "hidden" behind the `buildx` command so nothing changes in regards to how much storage docker consumes etc.

.github/workflows/docker.yml
thomaseizinger: > Imho we exchanged one complexity for a different one, but that's fine with me.

That is probably true.

> I'm happy to go with your proposal

Cool, thanks! What is being proposed here makes releases a bit easier because we don't need to manually trigger the building of docker images for a release after they have been attached so I'd say it is a net-positive overall. We can always iterate on the implementation details of how the containers are built.

.github/workflows/build-release-binary.yml
thomaseizinger: Feel free to send a follow-up patch :)

klochowicz:
> > as I haven't had much to do with multi-arch images - will this have negative impact on local storage? Is Docker smart enough to only download the parts that matter to the target platform?
> 
> To clarify, we are already using multi-arch images but it is "hidden" behind the `buildx` command so nothing changes in regards to how much storage docker consumes etc.

thank you for the clarification, much appreciated!

PR #767, opened 2021-11-30
Update patched `xtra` dependency

luckysori:
The new revision includes https://github.com/Restioson/xtra/commit/413dc807bd70d70cfe4a02cb3fd7d49d10d827e9
which could prevent some bugs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,6 @@ resolver = "2"
 [patch.crates-io]
 rocket = { git = "https://github.com/SergioBenitez/Rocket" } # Need to patch rocket dependency of `rocket_basicauth` until there is an official release.
 xtra = { git = "https://github.com/Restioson/xtra" } # We need to use unreleased patches.
-secp256k1-zkp = { git = "https://github.com/ElementsProject/rust-secp256k1-zkp" } # Latest master has crucial patches.
-maia = { git = "https://github.com/comit-network/maia", rev = "70fc548da0fe4f34478fb34ec437fa9a434c7ee3" }
-xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity", rev = "eda11a682dbbf16a9004cfe5e30baa1c60062a67" }
+secp256k1-zkp = { git = "https://github.com/ElementsProject/rust-secp256k1-zkp" } # We need to use unreleased patches.
+maia = { git = "https://github.com/comit-network/maia" } # Unreleased
+xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity" } # Unreleased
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -4,7 +4,7 @@ resolver = "2"
 
 [patch.crates-io]
 rocket = { git = "https://github.com/SergioBenitez/Rocket" } # Need to patch rocket dependency of `rocket_basicauth` until there is an official release.
-xtra = { git = "https://github.com/Restioson/xtra" } # Latest master has crucial patches.
+xtra = { git = "https://github.com/Restioson/xtra", rev = "413dc807bd70d70cfe4a02cb3fd7d49d10d827e9" } # Latest master has crucial patches.
```
thomaseizinger: Did a `cargo update -p xtra` not suffice?

IMO the lockfile should be enough here, we don't need to duplicate this information.

luckysori: I find it misleading to imply "latest" master here and then discover in the `lockfile` that we actually point to a specific revision that is not necessarily the latest one.

But if you feel strongly I can remove it.

thomaseizinger: Even with your patch, we are not guaranteed that we point to latest master :)

Feel free to change the comment to something like: "We want to utilize unreleased patches." if you find that that is clearer!

luckysori: What I mean is that in this case (probably in most cases) we don't want to point to the moving target that is latest master. When this was added we wanted to include some specific commits which happened to include latest master at the time. And that information was recorded in the lockfile.

My preference would be to always specify the revision here, so that one doesn't have to go to the lockfile (which one never edits manually) to find out what we are actually depending on. AFAIK this duplicated information cannot get out of sync. If it can then I would accept your suggestion and remove this.

thomaseizinger: A matter of taste I guess :)

In case you didn't know, `cargo` has a `precise` flag to update a dependency to a particular version, so you could run:

```
cargo update -p xtra --precise 413dc807bd70d70cfe4a02cb3fd7d49d10d827e9
```

I agree that the lockfile should not be edited manually.

thomaseizinger (approved):

Cargo.toml
luckysori: Actually, something really weird happens in the lockfile using `rev`. Currently on this branch:
```
source = "git+https://github.com/Restioson/xtra?rev=413dc807bd70d70cfe4a02cb3fd7d49d10d827e9#413dc807bd70d70cfe4a02cb3fd7d49d10d827e9"
```
After running `cargo update -p xtra --precise 224004bc95dac12b8b8d2a14fea5ddf37eef6ae0`, actually pointing to latest master:

```
source = "git+https://github.com/Restioson/xtra?rev=413dc807bd70d70cfe4a02cb3fd7d49d10d827e9#224004bc95dac12b8b8d2a14fea5ddf37eef6ae0"
```

The patch information is duplicated and the one after the hash seems to take precedence. I think I'm gonna do as you originally suggested.

thomaseizinger: That is indeed really weird!

thomaseizinger (approved):
Thanks for cleaning up the others as well!

PR #769, opened 2021-11-30
Pass changes in cfd proposals into projection actor

klochowicz:
Instead of passing the whole HashMap to the projection actor, only send what has
actually changed.

Provide separate messages to update settlement proposal and roll over proposals,
as they will be sent from different actors.

Some type conversions added in this commit are meant to be temporary and will disappear when cfd actors will not be storing a joint hashmap of proposals (both settlement and rollover) anymore.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -8,6 +8,10 @@ use crate::model::cfd::{
 };
 use crate::model::{BitMexPriceEventId, Price, Timestamp, Usd};
 use crate::monitor::{self, MonitorParams};
+use crate::projection::{
+    try_into_update_rollover_proposal, try_into_update_settlement_proposal, UpdateRollOverProposal,
+    UpdateSettlementProposal,
+};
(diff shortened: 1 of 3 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -162,14 +179,13 @@ where
...
-                proposal: proposal.clone(),
-                direction: SettlementKind::Outgoing,
-            },
-        );
-        self.send_pending_update_proposals().await?;
+        let new_proposal = UpdateCfdProposal::RollOverProposal {
+            proposal: proposal.clone(),
+            direction: SettlementKind::Outgoing,
+        };
+        self.projection_actor
```
thomaseizinger, line 189: Are we on purpose not inserting here?

thomaseizinger (approved):
Thanks!

daemon/src/taker_cfd.rs
klochowicz: oops, I missed that one! thanks for spotting! added the missing `insert`

PR #775, opened 2021-12-01
Move collab. settlement protocol into dedicated actors for maker

thomaseizinger:


daemon/src/collab_settlement_maker.rs
```diff
@@ -0,0 +1,212 @@
...
+        tracing::info!(
+            order_id = %self.proposal.order_id,
+            price = %self.proposal.price,
+            "Received settlement proposal"
+        );
+
+        self.update_proposal(Some((self.proposal.clone(), SettlementKind::Incoming)))
+            .await;
+
+        todo!("Update maker_inc_connections::Actor with our address");
```
thomaseizinger: @luckysori We need to deal with this one :D

daemon/src/maker_cfd.rs
```diff
@@ -575,61 +536,23 @@ where
...
-                tracing::warn!("Failed to notify taker of accepted settlement: {}", e);
-                self.remove_pending_proposal(&order_id)
-                    .await
-                    .context("accepted settlement")?;
-            }
-        }
+        self.settlement_actors
+            .send(&order_id, collab_settlement_maker::Accepted)
+            .await
+            .then(|| ())
```
thomaseizinger: What is this for?

thomaseizinger (commented):
Two comments.

daemon/src/collab_settlement_maker.rs
luckysori: Ah, forgot to remove it. We should be registered as soon as the user accepts.

daemon/src/maker_cfd.rs
luckysori: `bool -> Option -> Result` so that we can keep the `?`.

thomaseizinger: Should be change the API of `send` then?

thomaseizinger: If you introduce an `ActorName` trait, you could even return a really descriptive error message that actor XYZ is not connected.

luckysori: I tried your suggestions in https://github.com/itchysats/itchysats/pull/775/commits/09ebb963db46cd621e83f78bca32ac08f7543e5a and https://github.com/itchysats/itchysats/pull/775/commits/d2dbb44e7abd4926eac4c0ba0bc7a51cce6b5439.

daemon/src/actor_name.rs
```diff
@@ -0,0 +1,3 @@
+pub trait ActorName {
+    fn actor_name() -> String;
+}
```
thomaseizinger: I'd move this into `address_map` because that is where the code lives that motivates this piece. If we ever delete `address_map`, this should probably go away too!

If this assumption is no longer true at some point, we can move it.

thomaseizinger (commented):
ACK

luckysori (approved):
Great work me!

PR #776, opened 2021-12-01
Rounding

### LINKED ISSUE #729: The quantity in the UI has to be rounded to 100 precision
You cannot go more precise than 100-steps on BitMEX. To allow a proper hedging calculation we should apply the same on the taker UI.

rishflab:
closes #729

### DIFF
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -112,9 +112,11 @@ export default function Trade({
     const quantityTooHigh = maxQuantity < parse(quantity);
     const quantityTooLow = minQuantity > parse(quantity);
     const quantityGreaterZero = parse(quantity) > 0;
+    const quantityIncrement = 100;
+    const quantityIsEvenlyDivisibleByIncrement = isEvenlyDivisible(parse(quantity), quantityIncrement);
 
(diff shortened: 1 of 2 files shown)


taker-frontend/src/components/Trade.tsx
```diff
@@ -289,6 +297,7 @@ function Quantity({ min, max, onChange, quantity }: QuantityProps) {
                     min={min}
                     max={max}
                     default={min}
+                    step={100}
```
bonomat: You could use the same variable here as well: 
```suggestion
                    step={quantityIncrement}
```

taker-frontend/src/components/Trade.tsx
```diff
@@ -345,3 +354,11 @@ function Margin({ margin }: MarginProps) {
         </VStack>
     );
 }
+
+export function isEvenlyDivisible(numerator: number, divisor: number): boolean {
+    if (numerator % divisor == 0.0) {
```
bonomat: Nit: 

```suggestion
    return (numerator % divisor === 0.0);
```

bonomat (approved):
👍

thomaseizinger (approved):
Nice work, thanks for adding tests!

PR #779, opened 2021-12-01
Hello message

### LINKED ISSUE #577: Version check during first connection between maker and taker
We should have a version compatibility check between maker and taker. 

For example, the taker could say to the maker `running version X.Y.Z` and the maker returns his version. If they not match, either party should cut the connection.

da-kami:
fixes https://github.com/itchysats/itchysats/issues/577

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -2,11 +2,11 @@ use crate::address_map::{AddressMap, Stopping};
 use crate::model::cfd::OrderId;
 use crate::model::{Identity, Price, Timestamp, Usd};
 use crate::tokio_ext::FutureExt;
-use crate::wire::EncryptedJsonCodec;
+use crate::wire::{EncryptedJsonCodec, TakerToMaker, Version};
 use crate::{collab_settlement_taker, log_error, noise, send_to_socket, setup_taker, wire, Tasks};
-use anyhow::{Context, Result};
+use anyhow::{bail, Context, Result};
 use bdk::bitcoin::Amount;
-use futures::StreamExt;
+use futures::{SinkExt, StreamExt, TryStreamExt};
 use std::collections::HashMap;
 use std::net::SocketAddr;
 use std::sync::{Arc, Mutex};
@@ -55,7 +55,17 @@ struct MeasurePulse;
 #[derive(Clone, Debug, PartialEq)]
 pub enum ConnectionStatus {
     Online,
-    Offline,
+    Offline {
+        reason: Option<ConnectionCloseReason>,
+    },
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum ConnectionCloseReason {
+    VersionMismatch {
+        taker_version: Version,
+        maker_version: Version,
+    },
 }
 
 /// Message sent from the `setup_taker::Actor` to the
@@ -197,8 +207,51 @@ impl Actor {
             (read, write, Arc::new(Mutex::new(noise)))
         };
 
-        let read = FramedRead::new(read, wire::EncryptedJsonCodec::new(noise.clone()));
-        let write = FramedWrite::new(write, EncryptedJsonCodec::new(noise));
+        let mut read = FramedRead::new(read, wire::EncryptedJsonCodec::new(noise.clone()));
+        let mut write = FramedWrite::new(write, EncryptedJsonCodec::new(noise));
+
+        let our_version = Version::current();
+        write.send(TakerToMaker::Hello(our_version.clone())).await?;
+
+        match read
+            .try_next()
+            .timeout(Duration::from_secs(10))
+            .await
+            .with_context(|| {
+                format!(
+                    "Maker {} did not send Hello within 10 seconds, dropping connection",
+                    maker_identity
+                )
+            })? {
+            Ok(Some(wire::MakerToTaker::Hello(maker_version))) => {
+                if our_version != maker_version {
+                    self.status_sender
+                        .send(ConnectionStatus::Offline {
+                            reason: Some(ConnectionCloseReason::VersionMismatch {
+                                taker_version: our_version.clone(),
+                                maker_version: maker_version.clone(),
+                            }),
+                        })
+                        .expect("receiver to outlive the actor");
+
+                    bail!(
+                        "Network version mismatch, we are on version {} but taker is on version {}",
+                        our_version,
+                        maker_version,
+                    )
+                }
+            }
+            unexpected_message => {
+                bail!(
+                    "Unexpected message {:?} from maker {}",
+                    unexpected_message,
+                    maker_identity
+                )
+            }
+        }
+
+        tracing::info!(address = %maker_addr, "Established connection to maker");
+
         let this = ctx.address().expect("self to be alive");
 
         let send_to_socket = send_to_socket::Actor::new(write);
@@ -218,8 +271,6 @@ impl Actor {
         self.status_sender
             .send(ConnectionStatus::Online)
             .expect("receiver to outlive the actor");
-
-        tracing::info!(address = %maker_addr, "Established connection to maker");
 
         Ok(())
     }
@@ -287,6 +338,9 @@ impl Actor {
                     tracing::warn!(%order_id, "No active collaborative settlement");
                 }
             }
+            wire::MakerToTaker::Hello(_) => {
+                tracing::warn!("Ignoring unexpected Hello message from maker. Hello is only expected when opening a new connection.")
+            }
             other => {
                 // this one should go to the taker cfd actor
                 log_error!(self.maker_to_taker.send(other));
@@ -307,7 +361,7 @@ impl Actor {
 
         if time_since_last_heartbeat > self.heartbeat_timeout {
             self.status_sender
-                .send(ConnectionStatus::Offline)
+                .send(ConnectionStatus::Offline { reason: None })
                 .expect("watch receiver to outlive the actor");
             self.connected_state = None;
         }
@@ -325,7 +379,8 @@ pub async fn connect(
     maker_addresses: Vec<SocketAddr>,
 ) {
     loop {
-        if maker_online_status_feed_receiver.borrow().clone() == ConnectionStatus::Offline {
+        let connection_status = maker_online_status_feed_receiver.borrow().clone();
+        if matches!(connection_status, ConnectionStatus::Offline { .. }) {
             tracing::debug!("No connection to the maker");
             'connect: loop {
                 for address in &maker_addresses {
@@ -339,7 +394,7 @@ pub async fn connect(
                         .await
                         .expect("Taker actor to be present")
                     {
-                        tracing::trace!(%address, "Failed to establish connection: {:#}", e);
+                        tracing::warn!(%address, "Failed to establish connection: {:#}", e);
                         continue;
                     }
                     break 'connect;
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -1,6 +1,6 @@
 use crate::address_map::{AddressMap, Stopping};
 use crate::model::cfd::OrderId;
-use crate::model::{Price, Timestamp, Usd};
+use crate::model::{Identity, Price, Timestamp, Usd};
 use crate::tokio_ext::FutureExt;
 use crate::wire::EncryptedJsonCodec;
 use crate::{collab_settlement_taker, log_error, noise, send_to_socket, setup_taker, wire, Tasks};
@@ -41,7 +41,7 @@ pub struct Actor {
 }
 
 pub struct Connect {
-    pub maker_identity_pk: x25519_dalek::PublicKey,
+    pub maker_identity: Identity,
     pub maker_addr: SocketAddr,
 }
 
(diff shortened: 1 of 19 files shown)


daemon/src/connection.rs
```diff
@@ -285,6 +291,20 @@ impl Actor {
                     tracing::warn!(%order_id, "No active collaborative settlement");
                 }
             }
+            wire::MakerToTaker::Goodbye(goodbye) => {
```
thomaseizinger: I think we should actively close the connection here.

daemon/src/maker_inc_connections.rs
```diff
@@ -99,19 +101,50 @@ impl Actor {
...
 
         let (read, write) = stream.into_split();
         let mut read =
             FramedRead::new(read, wire::EncryptedJsonCodec::new(transport_state.clone()));
 
+        if let Ok(Some(TakerToMaker::Hello(version))) = read
+            .try_next()
+            .timeout(Duration::from_secs(10))
+            .await
+            .context("Taker did not sent Hello within 10 seconds, dropping connection")?
```
thomaseizinger: Would be useful to include the identity here.

daemon/src/wire.rs
```diff
@@ -109,6 +130,30 @@ pub mod maker_to_taker {
...
+            self.message.fmt(f)
+        }
+    }
+
+    /// A fixed set of goodbye messages from maker to taker
+    ///
+    /// This helps the taker to understand the reason why a connection was closed.
+    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+    pub enum Goodbye {
+        VersionMismatch { maker_version: Version },
```
thomaseizinger: I think this expresses the semantics a bit better.

```suggestion
        UnsupportedVersion { minimum_version: Version },
```

daemon/src/wire.rs
```diff
@@ -18,6 +18,23 @@ use std::ops::RangeInclusive;
 use std::sync::{Arc, Mutex};
 use tokio_util::codec::{Decoder, Encoder, LengthDelimitedCodec};
 
+const NETWORK_VERSION: u8 = 1;
```
thomaseizinger: Eventually, we could make this a bit more sophisticated and use semantic versioning here.

Then the taker could automatically detect that there is a newer version (1.0.0 < 1.5.0 f.e.) but still be compatible.

daemon/src/wire.rs
```diff
@@ -109,6 +130,30 @@ pub mod maker_to_taker {
...
+
+    /// A generic string message that can be sent from maker to taker
+    ///
+    /// This can be used to e.g. warn the taker user that a version will be deprecated and an
+    /// upgrade is coming. This message is not strongly typed on purpose to keep it configurable
+    /// for the maker. This message should only be used very mindfully.
+    #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+    pub struct Alarm {
+        pub message: String,
+    }
```
thomaseizinger: Sending free-form messages to clients can be a security problem: https://cointelegraph.com/news/phishing-attack-on-electrum-wallet-nets-hacker-almost-1-million-in-hours-report

thomaseizinger (changes requested):
I am actually not particularly comfortable to include this free-form `Alarm` message here. Can we solve this problem instead by:

1. Introducing a `MakerToTaker::Hello` message
2. That includes a version <=> deprecation date map

Then the taker can figure out by itself if it is on an outdated version and can display a warning to the user.

daemon/src/wire.rs
da-kami: I think it could happen that a maker did not uppgrade, and the taker is one version ahead. That's why I opted for the `maker_version` instead of minimum. Minimum also makes it harder to decide what that means - I think if we want that we would have to use semantic versioning already, so we can decide what "minimum supported version" means.

daemon/src/wire.rs
da-kami: I agree, but wanted to push that away a bit. Not sure it is a good idea to push it away too long though.

da-kami:
What is to be done after discussion: 

* Remove `Alarm`
* Keep `Hello` Taker2Maker
* Keep `Goodbye`  Maker2Taker with reason
* Introduce `Hello` Maker2Taker (which just sends the version for now)

More sophisticated "deprecation warning" messages can be added later, because we can always force the user to upgrade by publishing a new version that introduces the new messages.

daemon/src/connection.rs
da-kami:
```
 self.send_to_maker_ctx.stop();
```

will this be sufficient to actively close the connection?

daemon/src/connection.rs
```diff
@@ -197,8 +198,33 @@ impl Actor {
...
+                    "Maker {} did not send Hello within 10 seconds, dropping connection",
+                    maker_identity
+                )
+            })? {
+            Ok(Some(wire::MakerToTaker::Hello(_))) => {
+                tracing::info!(address = %maker_addr, "Established connection to maker");
+            }
+            _ => {
+                tracing::error!(
+                    "Maker {} did not send Hello within 10 seconds, dropping connection",
```
thomaseizinger: Wrong error message :)

thomaseizinger: Should probably also we a `bail!`.

daemon/src/connection.rs
thomaseizinger: I think so but better test it :)

da-kami:
Connection status when on "wrong" version is not especially beautiful:

image

I'd see to improve that in a follow-up PR

da-kami:
This includes semantic versioning, but we don't apply any logic yet. i.e. if the versions don't match the maker sends `Goodbye`. We could extend this to allow minor version upgrades, but I refrained from adding more complex logic until we know better what we want (also in terms of display).

daemon/src/connection.rs
da-kami: I did not explicitly test this code path because at the moment it is actually not used. 
Shall I remove it for now?
The version check always happens upon setting up the connection - we did not introduce any other `Goodbye` message yet, so we never hit this.

thomaseizinger: I didn't expect that we would "greet" with a `Goodbye` message but rather do:

```
taker -> maker: hello
maker -> taker: hello
maker -> taker: goodbye
```

Then we would only have one codepath for "goodbye" and "hello" would be a completely separate "handshake protocol". Hello would basically be our version of libp2p's identify :)

Happy to merge this as well though.

thomaseizinger (approved):

daemon/src/connection.rs
da-kami: Hm... if I implement it like this I run into the problem that the taker just immediately tries to re-connect forever:

```
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
2021-12-03 14:04:11 DEBUG daemon::connection: Connecting to maker address=127.0.0.1:9999
2021-12-03 14:04:11 DEBUG daemon::noise: Noise protocol initiator handshake is complete
2021-12-03 14:04:11  INFO daemon::connection: Established connection to maker address=127.0.0.1:9999
2021-12-03 14:04:11 TRACE daemon::connection: Received 'Goodbye'
2021-12-03 14:04:11 DEBUG daemon::connection: No connection to the maker
...
```

I think that is problematic behavior and it's better to just fail to connect to that maker initially of now.

thomaseizinger (commented):
The reconnecting is indeed problematic.

How about this design:

1. Both parties send a `Hello` message with their version
2. Once they receive `Hello` from the other party, both do a version check.
3. If the version check fails, they both actively terminate the connection (no goodbye) version.

da-kami:
> The reconnecting is indeed problematic.
> 
> How about this design:
> 
>     1. Both parties send a `Hello` message with their version
> 
>     2. Once they receive `Hello` from the other party, both do a version check.
> 
>     3. If the version check fails, they both actively terminate the connection (no goodbye) version.

done.
Note that the taker treats not establishing a connection as `warn`, the maker as `debug`.
The logs on both sides get spammed quite a bit, so I actually thought about downgrading the message on the maker to `trace`.

daemon/src/maker_inc_connections.rs
```diff
@@ -117,19 +118,54 @@ impl Actor {
...
+                write.send(MakerToTaker::Hello(our_version.clone())).await?;
+
+                if our_version != taker_version {
+                    tracing::debug!(
+                        "Network version mismatch, we are on version {} but taker is on version {}",
+                        our_version,
+                        taker_version
+                    );
+
+                    // A taker running an old version is not treated as error for the maker
```
thomaseizinger: Note: You are comparing `!=` but the comment says "older" which would imply a comparison like `<`?

thomaseizinger (approved):
Thanks!

I think we could also actively disconnect from the maker but it is not that important, can add that later.

PR #781, opened 2021-12-02
Disable trade button while any cfd is still setting up

da-kami:
image

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -31,6 +31,7 @@ import {
     MarginResponse,
     Order,
     StateGroupKey,
+    StateKey,
     WalletInfo,
 } from "./types";
 import useDebouncedEffect from "./useDebouncedEffect";
@@ -63,6 +64,10 @@ export const App = () => {
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
     cfds.sort((a, b) => a.order_id.localeCompare(b.order_id));
+    const istSettingUpCfd = cfds.some((cfd) =>
+        [StateKey.OUTGOING_ORDER_REQUEST, StateKey.CONTRACT_SETUP].includes(cfd.state.key)
+    );
+
     const connectedToMakerOrUndefined = useLatestEvent<boolean>(source, "maker_status");
     const connectedToMaker = connectedToMakerOrUndefined ? connectedToMakerOrUndefined! : false;
 
@@ -131,6 +136,7 @@ export const App = () => {
                                     }}
                                     onLongSubmit={makeNewOrderRequest}
                                     isLongSubmitting={isCreatingNewOrderRequest}
+                                    isSettingUpCfd={istSettingUpCfd}
                                 />
                                 <History
                                     connectedToMaker={connectedToMaker}
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -3,6 +3,7 @@ import {
     Alert,
     AlertDescription,
     AlertIcon,
+    AlertStatus,
     AlertTitle,
     Box,
     Button,
@@ -66,15 +67,17 @@ interface TradeProps {
     walletBalance: number;
     onLongSubmit: (payload: CfdOrderRequestPayload) => void;
     isLongSubmitting: boolean;
+    isSettingUpCfd: boolean;
 }
 
 interface AlertBoxProps {
     title: string;
     description: string;
+    status?: AlertStatus;
 }
 
-function AlertBox({ title, description }: AlertBoxProps) {
-    return (<Alert status="error">
+function AlertBox({ title, description, status = "error" }: AlertBoxProps) {
+    return (<Alert status={status}>
         <AlertIcon />
         <AlertTitle mr={2}>{title}</AlertTitle>
         <AlertDescription>{description}</AlertDescription>
@@ -96,6 +99,7 @@ export default function Trade({
     isLongSubmitting,
     orderId,
     walletBalance,
+    isSettingUpCfd,
 }: TradeProps) {
     let outerCircleBg = useColorModeValue("gray.100", "gray.700");
     let innerCircleBg = useColorModeValue("gray.200", "gray.600");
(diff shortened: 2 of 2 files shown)


da-kami:
Note: Since this includes `PendingOpen` the user might see this message for 10 minutes

thomaseizinger (approved):
Fine with me.

bonomat:
Just to summarize: we want to disable the button because our wallet could run into a "double spend" situation where it tries to use the same UTXOs for different CFDs. 

I'm not sure if this solution improves things but it definitely comes at the cost of UX.   

I just wrote a little PoC where I send two transactions in a row. If I do a `wallet.sync()` in between, the wallet does not use the same UTXOs. Without, it does. 

That being said, we should solve this issue differently, e.g. have a state indicating that the `tx_lock` was seen in the mempool and/or, do a `wallet.sync` before going into `contract_setup`. That way we can reduce the possibilities that this error happens. 

Solving this problem in the backend is also preferable, because this error could also happen for the maker. 

Back to your PR: At the very least I would change the `AlertBox` to use `status="warn"` instead of `status="error"` for this message. Otherwise it feels like something is wrong.

thomaseizinger:
> I just wrote a little PoC where I send two transactions in a row. If I do a wallet.sync() in between, the wallet does not use the same UTXOs. Without, it does.

Thanks for this! Triggering a sync before contract setup sounds okay to me.

bonomat:
> > I just wrote a little PoC where I send two transactions in a row. If I do a wallet.sync() in between, the wallet does not use the same UTXOs. Without, it does.
> 
> Thanks for this! Triggering a sync before contract setup sounds okay to me.

Thinking about this more: can we use a mutex of the wallet and lock it while we are in the contract setup and release it when done , i.e. published the transaction and called `wallet.sync`? That way we should be able to guarantee that the wallet is not reusing the same utxos.

thomaseizinger:
> > > I just wrote a little PoC where I send two transactions in a row. If I do a wallet.sync() in between, the wallet does not use the same UTXOs. Without, it does.
> > 
> > Thanks for this! Triggering a sync before contract setup sounds okay to me.
> 
> Thinking about this more: can we use a mutex of the wallet and lock it while we are in the contract setup and release it when done , i.e. published the transaction and called `wallet.sync`? That way we should be able to guarantee that the wallet is not reusing the same utxos. 

That is tricky because we can only broadcast once we merged the two PSBTs, meaning we would hold a lock that will only be released by network communication (signature of other party coming in).

I wouldn't want to hold a Mutex for this long, that is just asking for trouble.

An optimistic approach and retrying is preferable IMO.

da-kami:
To summarize what is to be changed:

* disable the button in `OutgoingOrderRequest` and `ContractSetup` but not in `PendingOpen`.
* add a `wallet.sync()` before going into contract setup

I think this will reduce the probability of running into this issue, but it will not cover all cases, because contract setup just takes some time and we cannot ensure that a lock tx from one contract setup is seen in mempool at the starting point of another contract setup. 
We could lock the wallet while in contract setup, but that seems to be tricky.

I think to solve this problem completely we will also have to add the handling of failing to publish the (lock) transaction.
That way we cover handling the problem when it happens (and avoid keeping a `PendingOpen` forever cfd around). 

I will see to invest the time to solve this completely later.

thomaseizinger:
Adding the wallet sync might be not as useful as we think.

1. We would have to do it _before_ contract setup as `build_party_params` is among the first things that happen.
2. Contract setup can take up to a minute to complete.
3. We very regularly call wallet sync already.

I think disabling the button should improve things a bit but it is not going to help with concurrency on the maker side.

I suggest we open a feature request with bdk to mark selected UTXOs as "spent" when funding a PSBT. That should actually solve our problem.

As a workaround, the maker should probably hold off with publish a new order until the contract setup with one party either completes or fails. That should help with this error on the maker side if triggered by concurrent takers.

bonomat:
> I suggest we open a feature request with bdk to mark selected UTXOs as "spent" when funding a PSBT. That should actually solve our problem.

There is a function on the tx builder which we could make use of: 

https://github.com/bitcoindevkit/bdk/blob/aa13186fb08b5bcf630b439ae77053d58ba41d52/src/wallet/tx_builder.rs#L394-L401

This would allow us to tell the tx builder which UTXOs not to select (i.e. the ones we've selected for another tx lock). 

I found this comment here: https://github.com/bitcoindevkit/bdk/issues/98#issuecomment-697228850: Even though this discussion is about coin blacklisting, I think that they prefer that locking and remembering locked transactions is something they would see on our responsibility and not in the bdk. 

So my thinking is making use of this flow: 

1. in contract_setup we load _locked_ utxos (from our db) and create a tx_lock
2. as soon as we created tx_lock we update the _locked_ utxos in the db
3. if publishing of tx_lock failed we remove the locked utxos from our db
  a. on a regular interval we could clean up this table so that it does not outgrow too much

thomaseizinger:
> I found this comment here: [bitcoindevkit/bdk#98 (comment)](https://github.com/bitcoindevkit/bdk/issues/98#issuecomment-697228850): Even though this discussion is about coin blacklisting, I think that they prefer that locking and remembering locked transactions is something they would see on our responsibility and not in the bdk.
> 
> So my thinking is making use of this flow:
> 
>     1. in contract_setup we load _locked_ utxos (from our db) and create a tx_lock
> 
>     2. as soon as we created tx_lock we update the _locked_ utxos in the db
> 
>     3. if publishing of tx_lock failed we remove the locked utxos from our db
>        a. on a regular interval we could clean up this table so that it does not outgrow too much

Good find! I think you are describing a good solution and also the reasoning in the BDK ticket for why it is not a feature that is present in BDK makes sense to me.

One suggestion: Do you think it would be enough to store the locked UTXOs in-memory? I think this list could easily be stored within `wallet::Actor`. Then it would "auto-clean" itself on restarts.

bonomat:
> One suggestion: Do you think it would be enough to store the locked UTXOs in-memory? I think this list could easily be stored within `wallet::Actor`. Then it would "auto-clean" itself on restarts.

True, that should be sufficient. I'll create a ticket for that.

da-kami:
I think disabling the button during contract setup is still a useful feature, updated and changed alert box to `warning`.

da-kami:
Closing because fix for https://github.com/itchysats/itchysats/issues/790 should make this obsolete

PR #785, opened 2021-12-03
Replace automated reconnecting with button in UI

thomaseizinger:
Not only does this simplify the implementation a bit, it also gives
the user more control and a way of actioning an error state. It is
also more efficient because we don't try to re-connect in an endless
loop.

When disconnected:

image

When connected:

image

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -18,9 +18,6 @@ use xtra::prelude::MessageChannel;
 use xtra::KeepRunning;
 use xtra_productivity::xtra_productivity;
 
-/// Time between reconnection attempts
-const CONNECT_TO_MAKER_INTERVAL: Duration = Duration::from_secs(5);
-
 struct ConnectedState {
     last_heartbeat: SystemTime,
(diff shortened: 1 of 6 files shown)


klochowicz (commented):
I'm a bit worried about the UX implications of this changeset. If maker goes down for a few minutes, the user will get fed up and annoyed after a few unsuccessful attempts at reconnecting.

On the other hand, if we kept reconnecting, we could notify the user with a push notification that it's back online!

thomaseizinger:
> If maker goes down for a few minutes, the user will get fed up and annoyed after a few unsuccessful attempts at reconnecting.

Same argument could be made for the current solution. The website just says "Maker status: Offline" without giving the user any explanation why and no way of doing anything about it. Sounds pretty annoying to me :)

> On the other hand, if we kept reconnecting, we could notify the user with a push notification that it's back online!

But with the current solution, the user doesn't know that retries are happening!

I am not against automated retries but at least from an engineering perspective, and endless loop with a constant sleep value is a pretty bad solution.

PR #788, opened 2021-12-03
Upgrade to Rust 1.57

thomaseizinger:
The clippy version included with 1.57 flagged several problems:

1. Multiple fields were never read, removed those.
2. The `Default` impl on `Tasks` could be derived.
3. A redundant closure
4. A large enum variant


klochowicz (approved):
cool! nice to know that we can derive `Default` in such cases now!

thomaseizinger:
> cool! nice to know that we can derive `Default` in such cases now!

We could do that before already! `Vec` implements `Default` :)

bonomat (approved):
nice

PR #791, opened 2021-12-03
Test collaborative settlement

klochowicz:


daemon/tests/happy_path.rs
```diff
@@ -186,3 +222,68 @@ async fn maker_notices_lack_of_taker() {
...
+    taker.mocks.mock_oracle_monitor_attestation().await;
+
+    maker.mocks.mock_monitor_start_monitoring().await;
+    taker.mocks.mock_monitor_start_monitoring().await;
+
+    maker.accept_take_request(received.clone()).await;
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert!(matches!(taker_cfd.state, CfdState::ContractSetup { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::ContractSetup { .. }));
```
thomaseizinger, line 274: Is this necessary in the setup function?

daemon/tests/happy_path.rs
```diff
@@ -186,3 +222,68 @@ async fn maker_notices_lack_of_taker() {
...
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert!(matches!(taker_cfd.state, CfdState::PendingOpen { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::PendingOpen { .. }));
+
+    deliver_lock_finality_event(&maker, &taker, received.id).await;
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert!(matches!(taker_cfd.state, CfdState::Open { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::Open { .. }));
```
thomaseizinger, line 287: Is this necessary in the setup function?

daemon/tests/happy_path.rs
```diff
@@ -125,6 +126,41 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
...
+
+    maker.mocks.mock_monitor_collaborative_settlement().await;
+    taker.mocks.mock_monitor_collaborative_settlement().await;
+
+    maker.accept_settlement_proposal(order_id).await;
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert!(matches!(taker_cfd.state, CfdState::PendingClose { .. }));
+    assert!(matches!(maker_cfd.state, CfdState::PendingClose { .. }));
```
thomaseizinger, line 154: Can we someone assert that this is a collab. close?

thomaseizinger (approved):

daemon/tests/happy_path.rs
klochowicz: `PendingClose` is actually a collab-close only close! a force-close would assert on `PendingCommit`

daemon/tests/happy_path.rs
klochowicz: we need `next_cfd` everywhere - this assertions give the users confidence that we're always in that state when we use the function. no room for error.

daemon/tests/happy_path.rs
klochowicz: the `next_cfd` call is crucial, otherwise we try to trigger events too early :(
I like describing with the assertions what state we are in - it gives more context to the reader. As this function is moved into a helper, I think it shouldn't decrease readability of the tests.

klochowicz: Next week I'll try to come up with a function that just waits until you're in a particular state on the cfd, a one-liner that would be much more readable than this code littered everywhere. I've got some ideas, I'll ping you on Monday :)

PR #795, opened 2021-12-03
Allow projection actor to retrieve latest Cfd state itself from DB

klochowicz:
Instead of retrieving Cfds inside Cfd actors, send a msg that they changed and
retrieve them by the projection actor itself.

### DIFF
--- a/daemon/src/cfd_actors.rs
+++ b/daemon/src/cfd_actors.rs
@@ -4,15 +4,13 @@ use anyhow::{bail, Context, Result};
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
 
-pub async fn insert_cfd_and_send_to_feed(
+pub async fn insert_cfd_and_update_feed(
     cfd: &Cfd,
     conn: &mut PoolConnection<Sqlite>,
     projection_address: &xtra::Address<projection::Actor>,
 ) -> Result<()> {
     db::insert_cfd(cfd, conn).await?;
-    projection_address
-        .send(projection::Update(db::load_all_cfds(conn).await?))
-        .await?;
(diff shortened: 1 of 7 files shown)


thomaseizinger (approved):
Nice, thank you!

I think the two functions `cfd_actors` are pretty useless now so wouldn't mind inlining them :)

klochowicz:
> Nice, thank you!
> 
> I think the two functions `cfd_actors` are pretty useless now so wouldn't mind inlining them :)

they're pretty small, but IMHO not useless! we had a problem previously that we forgot to update the feed in some cases... when we will move to looking at events in projection actor, then it will be useless - I agree :)

thomaseizinger:
> Build failed:
> 
>     * [test_daemons (macos-latest)](https://github.com/itchysats/itchysats/runs/4405024373?check_suite_focus=true)

Well that is concerning:

> process didn't exit successfully: `/Users/runner/work/itchysats/itchysats/target/debug/deps/happy_path-795f3ae0e3f41528` (signal: 11, SIGSEGV: invalid memory reference)

How can our tests SEGFAULT??

luckysori (approved):

klochowicz:
bors retry

klochowicz:
> 
I have never seen a segfault on my macos, I suspect it might be some linked to the gh actions container.... is there any way to obtain a code dump from gh actions? 😅

thomaseizinger:
> > 
> 
> I have never seen a segfault on my macos, I suspect it might be some linked to the gh actions container.... is there any way to obtain a code dump from gh actions? sweat_smile

I don't think so but it might be the same issue we saw on the Windows tests at times!

PR #796, opened 2021-12-03
Introduce short-lived `rollover_taker::Actor`

luckysori:
Very similar to the other ones we've done.

Things to consider:
- Shall we consolidate the usage of "rollover" over "roll over" across the entire codebase? I don't think I have been able to stay consistent even in this PR :(
- I've added an `OrderId` to the `wire::MakerToTaker::ConfirmRollOver` message, because the maker could be talking about more than one rollover.
- The `taker_cfd::Actor` no longer handles the raw `wire::MakerToTaker` enumeration of messages :tada: After creating all these short-lived actors it only needs to handle a dedicated `CurrentOrder` message originating from the maker.
- Error handling is worth reviewing in detail as I've messed that up in previous related PRs.

### DIFF
--- a/daemon/src/rollover_taker.rs
+++ b/daemon/src/rollover_taker.rs
@@ -1,0 +1,318 @@
+use crate::address_map::Stopping;
+use crate::connection;
+use crate::model::cfd::{Cfd, Dlc, OrderId, Role, RollOverProposal, SettlementKind};
+use crate::model::{BitMexPriceEventId, Timestamp};
+use crate::oracle::{self, GetAnnouncement};
+use crate::projection::{self, UpdateRollOverProposal};
+use crate::setup_contract::{self, RolloverParams};
+use crate::tokio_ext::spawn_fallible;
+use crate::wire::{self, RollOverMsg};
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use futures::channel::mpsc::{self, UnboundedSender};
+use futures::{future, SinkExt};
+use maia::secp256k1_zkp::schnorrsig;
+use xtra::prelude::MessageChannel;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor {
+    cfd: Cfd,
+    n_payouts: usize,
+    oracle_pk: schnorrsig::PublicKey,
+    timestamp: Timestamp,
+    maker: xtra::Address<connection::Actor>,
+    get_announcement: Box<dyn MessageChannel<GetAnnouncement>>,
+    projection: xtra::Address<projection::Actor>,
+    on_completed: Box<dyn MessageChannel<Completed>>,
+    on_stopping: Vec<Box<dyn MessageChannel<Stopping<Self>>>>,
+    rollover_msg_sender: Option<UnboundedSender<RollOverMsg>>,
+}
+
+impl Actor {
+    pub fn new(
+        (cfd, n_payouts): (Cfd, usize),
+        oracle_pk: schnorrsig::PublicKey,
+        maker: xtra::Address<connection::Actor>,
+        get_announcement: &(impl MessageChannel<GetAnnouncement> + 'static),
+        projection: xtra::Address<projection::Actor>,
+        on_completed: &(impl MessageChannel<Completed> + 'static),
+        (on_stopping0, on_stopping1): (
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+        ),
+    ) -> Self {
+        Self {
+            cfd,
+            n_payouts,
+            oracle_pk,
+            timestamp: Timestamp::now(),
+            maker,
+            get_announcement: get_announcement.clone_channel(),
+            projection,
+            on_completed: on_completed.clone_channel(),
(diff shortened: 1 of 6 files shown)


daemon/src/rollover_taker.rs
```diff
@@ -0,0 +1,318 @@
...
+        &mut self,
+        msg: RollOverAccepted,
+        ctx: &mut xtra::Context<Self>,
+    ) -> Result<()> {
+        let RollOverAccepted { oracle_event_id } = msg;
+        let announcement = self
+            .get_announcement
+            .send(oracle::GetAnnouncement(oracle_event_id))
+            .await?
+            .with_context(|| format!("Announcement {} not found", oracle_event_id))?;
```
thomaseizinger, line 89: I guess here we can't avoid it so could potentially include this in the other protocols as well? But then the question is, where do you draw the line? Should each of these actors perhaps also load the CFD from the DB themselves? :thinking:

thomaseizinger: Perhaps not such a bad idea actually because then even more error cases are covered ...

daemon/src/taker_cfd.rs
```diff
@@ -510,92 +391,67 @@ where
...
-
-        let (sender, receiver) = mpsc::unbounded();
+        let ProposeRollOver { order_id } = msg;
 
-        if let RollOverState::Active { .. } = self.roll_over_state {
-            anyhow::bail!("Already rolling over a contract!")
-        }
+        let disconnected = self
+            .rollover_actors
+            .get_disconnected(order_id)
```
thomaseizinger, line 409: A pattern seems to be forming here where we:

1. Get a `Disconnected` entry
2. Start a new actor
3. Add the actor to the map

Might be worth exploring an API such as:

```rust
/// Inserts a new actor under the given key if the entry is either empty or the existing address is disconnected.
pub fn insert_new(&mut self, key: K, actor_ctor: impl Fn() -> A) -> Result<(), StillConnected>;
```

That would save a couple of lines and remove the awkwardly named `disconnected` variable.

daemon/src/rollover_taker.rs
```diff
@@ -0,0 +1,318 @@
...
+            .send(connection::ProposeRollOver {
+                order_id: self.cfd.order.id,
+                timestamp: self.timestamp,
+                address: this,
+            })
+            .await??;
+
+        self.update_proposal(Some((
+            RollOverProposal {
+                order_id: self.cfd.order.id,
```
thomaseizinger, line 69: This seems redundant? We already include the ID in the `UpdateRollOverProposal` struct.

thomaseizinger (approved):
Thanks!

The actual rollover protocol is currently neither tested nor in use so I don't think we need to care too much about it at this stage. Let's get this is to complete the cleanup and test it more once we start using it again!

@da-kami Thoughts on that?

daemon/src/rollover_taker.rs
luckysori: I was leaning towards even loading from the `db` originally. I'm really not sure what the CFD actors should be doing other than handling user actions. Even handling completion of these ephemeral protocols may be unnecessary.

thomaseizinger: > I'm really not sure what the CFD actors should be doing other than handling user actions.

Someone needs to track the currently on-going protocols to make sure we are not executing them in parallel for the same CFD. In theory, we could overcome this (once we have event-sourcing) by recording the _expected_ version of the CFD in the protocol and if you try to apply the result to a different one, it would fail (and it means you'd have to retry). But if you do that, how would you communicate this back to the user / the original interaction? Thinking about it more though, it is really almost impossible to trigger multiple of these through the UI.

We could take the approach of not guarding against this in the API and let it only fail eventually with a log message in case someone manages to somehow trigger this?

> Even handling completion of these ephemeral protocols may be unnecessary.

Once we do event-sourcing, I imagine these to go away yes because we will simply broadcast the resulting event of applying the completed state to the CFD.

luckysori: To proceed, shall we just leave it as it is and iterate on all this as we go?

thomaseizinger: Yes, leave as is for now. Mostly something to mull over :)

daemon/src/taker_cfd.rs
luckysori: I've attempted to implement this quickly, but in practice we run into issues with the `actor_ctor` due to the borrow checker.

daemon/src/rollover_taker.rs
luckysori: Do you mean that you think we should remove the `order` from `RollOverProposal`. I don't think I added this stuff.

thomaseizinger: It seems that we can optimize this a bit. I also think this timestamp is not actually used anywhere :rofl: 

Don't need to change it now!

daemon/src/taker_cfd.rs
thomaseizinger: Thanks for trying, probably needs some iteration yeah!

thomaseizinger (approved):
LGTM!

I think there is low risk in merging this so happy to get it in :)

PR #804, opened 2021-12-06
Handle wallet sync errors more gracefully

thomaseizinger:
See individual patches for details.


daemon/src/wallet_sync.rs
```diff
@@ -1,29 +0,0 @@
-use crate::model::WalletInfo;
-use crate::wallet;
-use std::time::Duration;
-use tokio::sync::watch;
-use tokio::time::sleep;
-use xtra::Address;
-
-pub async fn new(wallet: Address<wallet::Actor>, sender: watch::Sender<Option<WalletInfo>>) {
```
klochowicz: awesome, I've always felt this could have been a part of the wallet actor 👍

daemon/src/to_sse_event.rs
```diff
@@ -46,13 +46,13 @@ pub struct WalletInfo {
     last_updated_at: Timestamp,
 }
 
-impl ToSseEvent for model::WalletInfo {
+impl ToSseEvent for Option<model::WalletInfo> {
```
klochowicz, line 49: don't we need to adjust our frontend code to deal with not having the wallet info as well?

klochowicz (approved):
(I forgot to click publish on my comments earlier!)

LGTM, provided clippy is happy and the frontend isn't broken by this (see my comment on the review)

daemon/src/to_sse_event.rs
thomaseizinger: I think we already handle this case in the frontend because we need to deal with the fact that the feed might never emit a `WalletInfo` event.

thomaseizinger:
Still needs manual testing.

PR #806, opened 2021-12-06
Move maker rollover logic to dedicated actor

rishflab:
The rollover actor is spawned the the maker receives a rollover proposal message from the taker. After the rollover is complete, it sends itself
a completion message triggering a "cleanup" handler. Cleanup involves sending messages to maker_cfd actor and the maker_inc_connections to
trigger removal of stored references to the rollover actor.

potential todo:
- [x] Move projection actor to `maker_rollover::Actor` instead of doing all the ui updates in the "maker_cfd::Actor"

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -10,15 +10,11 @@ use crate::maker_inc_connections;
 use crate::model::cfd::Cfd;
 use crate::model::cfd::CfdState;
 use crate::model::cfd::CfdStateCommon;
-use crate::model::cfd::Dlc;
 use crate::model::cfd::Order;
 use crate::model::cfd::OrderId;
 use crate::model::cfd::Origin;
-use crate::model::cfd::Role;
 use crate::model::cfd::RollOverProposal;
-use crate::model::cfd::SettlementKind;
 use crate::model::cfd::SettlementProposal;
-use crate::model::cfd::UpdateCfdProposal;
 use crate::model::Identity;
 use crate::model::Price;
 use crate::model::Timestamp;
@@ -27,14 +23,10 @@ use crate::monitor::MonitorParams;
 use crate::monitor::{self};
 use crate::oracle;
 use crate::projection;
-use crate::projection::try_into_update_rollover_proposal;
 use crate::projection::Update;
-use crate::projection::UpdateRollOverProposal;
-use crate::projection::UpdateSettlementProposal;
-use crate::setup_contract;
-use crate::setup_contract::RolloverParams;
+use crate::rollover_maker;
+use crate::rollover_maker::Completed;
 use crate::setup_maker;
-use crate::tokio_ext::FutureExt;
 use crate::wallet;
 use crate::wire;
 use crate::wire::TakerToMaker;
@@ -43,13 +35,8 @@ use anyhow::Context as _;
 use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
-use futures::channel::mpsc;
-use futures::future;
-use futures::future::RemoteHandle;
-use futures::SinkExt;
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
-use std::collections::HashMap;
 use std::collections::HashSet;
 use time::Duration;
 use xtra::prelude::*;
@@ -74,6 +61,10 @@ pub struct AcceptRollOver {
 pub struct RejectRollOver {
     pub order_id: OrderId,
 }
+pub struct RollOverProposed {
+    pub order_id: OrderId,
+    pub address: xtra::Address<rollover_maker::Actor>,
+}
 pub struct Commit {
     pub order_id: OrderId,
 }
@@ -92,11 +83,6 @@ pub struct TakerDisconnected {
     pub id: Identity,
 }
 
-pub struct CfdRollOverCompleted {
-    pub order_id: OrderId,
-    pub dlc: Result<Dlc>,
-}
-
 pub struct FromTaker {
     pub taker_id: Identity,
     pub msg: wire::TakerToMaker,
@@ -108,27 +94,16 @@ pub struct Actor<O, M, T, W> {
     settlement_interval: Duration,
     oracle_pk: schnorrsig::PublicKey,
     projection_actor: Address<projection::Actor>,
+    rollover_actors: AddressMap<OrderId, rollover_maker::Actor>,
     takers: Address<T>,
     current_order: Option<Order>,
     monitor_actor: Address<M>,
     setup_actors: AddressMap<OrderId, setup_maker::Actor>,
     settlement_actors: AddressMap<OrderId, collab_settlement_maker::Actor>,
-    roll_over_state: RollOverState,
     oracle_actor: Address<O>,
-    // Maker needs to also store Identity to be able to send a reply back
-    current_pending_proposals: HashMap<OrderId, (UpdateCfdProposal, Identity)>,
     connected_takers: HashSet<Identity>,
     n_payouts: usize,
     tasks: Tasks,
-}
-
-enum RollOverState {
-    Active {
-        taker: Identity,
-        sender: mpsc::UnboundedSender<wire::RollOverMsg>,
-        _task: RemoteHandle<()>,
-    },
-    None,
 }
 
 impl<O, M, T, W> Actor<O, M, T, W> {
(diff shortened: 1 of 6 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -632,141 +690,33 @@ where
...
-            sender,
-            taker: taker_id,
-            _task: task,
-        };
+        self.projection_actor
+            .send(UpdateRollOverProposal {
+                order: msg.order_id,
+                proposal: None,
+            })
+            .await?;
```
thomaseizinger: Move this into the rollover actor please.

daemon/src/maker_cfd.rs
```diff
@@ -205,18 +167,15 @@ impl<O, M, T, W> Actor<O, M, T, W> {
...
-        }
+        let rollover_actor = self
+            .rollover_actors
+            .get_connected(&order_id)
+            .context("rollover actor is not connected")?;
+
+        rollover_actor.send(msg).await?;
 
         Ok(())
     }
```
thomaseizinger: This never happens because you are doing the routing inside the `maker_inc_connections::Actor`.

daemon/src/maker_rollover.rs
```diff
@@ -0,0 +1,245 @@
...
+use async_trait::async_trait;
+use futures::channel::mpsc;
+use futures::channel::mpsc::UnboundedSender;
+use futures::{future, SinkExt};
+use xtra::prelude::MessageChannel;
+use xtra::{Context, Handler, KeepRunning};
+use xtra_productivity::xtra_productivity;
+
+pub struct AcceptRollOver {
+    pub order_id: OrderId,
```
thomaseizinger: An instance of this actor is for one specific `Cfd`, no need to send the order ID here. That is just an unnecessary error case. What if they don't match?

daemon/src/maker_rollover.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+            RollOverCompleted::Failed {
+                ref order_id,
+                ref error,
+            } => {
+                tracing::error!(%order_id, "Rollover failed: {}", error);
+                self.maker_cfd_actor.send(msg).await??;
+                ctx.stop();
+            }
+        }
```
thomaseizinger: This seems unnecessarily repetitive?

```suggestion
	self.maker_cfd_actor.send(msg).await??;
        ctx.stop();
```

Note that "logging AND passing on an error" is almost always undesired and will result in duplicate log messages.

thomaseizinger (changes requested):
Thanks! Some comments, nothing big :)

daemon/src/maker_cfd.rs
rishflab: what about accept and completed? Should those also be handled in the rollover actor instead of the `maker_cfd::Actor`

thomaseizinger: Yep, ideally the entire protocol is handled in the actor and the `maker_cfd::Actor` only gets informed about the outcome so it can update the state in DB.

daemon/src/maker_cfd.rs
```diff
@@ -395,6 +334,94 @@ where
...
+            .insert(proposal.order_id, rollover_actor_addr);
+
+        let new_proposal = UpdateCfdProposal::RollOverProposal {
+            proposal: proposal.clone(),
+            direction: SettlementKind::Incoming,
+        };
+
+        self.projection_actor
+            .send(try_into_update_rollover_proposal(new_proposal)?)
+            .await?;
```
thomaseizinger: Can we move this into the `started` hook of the actor?

daemon/src/maker_cfd.rs
```diff
@@ -632,141 +659,26 @@ where
...
-            time::OffsetDateTime::now_utc() + cfd.order.settlement_interval,
-        )?;
-        let announcement = self
-            .oracle_actor
-            .send(oracle::GetAnnouncement(oracle_event_id))
-            .await?
-            .with_context(|| format!("Announcement {} not found", oracle_event_id))?;
+impl<O, M, T, W> Actor<O, M, T, W> {
+    async fn handle_reject_rollover(&mut self, msg: RejectRollOver) -> Result<()> {
+        tracing::info!(%msg.order_id, "Maker rejects a roll_over proposal" );
```
thomaseizinger: Move these log statements into the actor as well please.

daemon/src/maker_cfd.rs
```diff
@@ -792,36 +704,51 @@ where
...
+                    })
+                    .await?;
+            }
+            RollOverCompleted::Rejected { order_id } => {
+                self.projection_actor
+                    .send(projection::UpdateRollOverProposal {
+                        order: order_id,
+                        proposal: None,
+                    })
+                    .await?;
```
thomaseizinger: This already happens in the `rollover_maker::Actor`.

daemon/src/maker_cfd.rs
```diff
@@ -792,36 +704,51 @@ where
...
+                tracing::info!(%order_id, "Rollover rejected");
+            }
+            RollOverCompleted::Failed { order_id, .. } => {
+                self.projection_actor
+                    .send(projection::UpdateRollOverProposal {
+                        order: order_id,
+                        proposal: None,
+                    })
+                    .await?;
```
thomaseizinger: Same here.

daemon/src/maker_inc_connections.rs
```diff
@@ -78,6 +79,7 @@ impl Actor {
             noise_priv_key,
             heartbeat_interval,
             setup_actors: AddressMap::default(),
+            rollover_actors: AddressMap::default(),
```
thomaseizinger, line 130: We do not yet write to this map so this can't work at the moment.

I think you'll have to introduce the equivalent of `maker_inc_connections::ConfirmOrder` for rollover and send the address across.

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1,246 @@
...
+use futures::channel::mpsc;
+use futures::channel::mpsc::UnboundedSender;
+use futures::{future, SinkExt};
+use xtra::prelude::MessageChannel;
+use xtra::{Context, Handler, KeepRunning};
+
+pub struct AcceptRollOver;
+
+pub struct RejectRollOver {
+    pub order_id: OrderId,
```
thomaseizinger, line 56: Same as for `AcceptRollOver`, we shouldn't send this ID across.

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1,246 @@
...
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+            &(impl MessageChannel<Stopping<Self>> + 'static),
+        ),
+        projection_actor: xtra::Address<projection::Actor>,
+    ) -> Self {
+        Self {
+            send_to_taker_actor: send_to_taker_actor.clone_channel(),
+            cfd,
+            taker_id,
+            n_payouts: 0,
```
thomaseizinger: Uhm, probably better not :D

thomaseizinger (changes requested):
Thanks for pushing this forward!

A few more comments plus the merge conflicts need to be resolved.

thomaseizinger:
> A few more comments plus the merge conflicts need to be resolved.

Ah, just saw the force-push! I'll check again :)

daemon/src/maker_cfd.rs
rishflab: to do this we will have pass the proposal into the actor. Not sure if that is a good idea. I am seeing the proposal and the rollover as different domain objects.

thomaseizinger: The proposal is part of the rollover protocol. In an ideal world, I would like to kick off the protocol directly from the user interaction and perhaps even move the loading of the CFD into the database into that actor as well. The entire process should be encapsulated in an actor.

daemon/src/maker_inc_connections.rs
rishflab:
```
        self.rollover_actors
            .insert(proposal.order_id, rollover_actor_addr);
```

I add the rollover actor to this map in `maker_cfd::handle_propose_rollover`

daemon/src/rollover_maker.rs
rishflab: are you talking about `n_payouts?`

thomaseizinger: Yes, we should not set to zero :D

rishflab: ok using the constant (200)

luckysori: I think we're meant to use the field in `maker_cfd::Actor`: https://github.com/itchysats/itchysats/blob/de36ff69028fe1043ae39e68406693b1c5533312/daemon/src/maker_cfd.rs#L96

thomaseizinger: Yes, you are meant to pass it on from the `maker_cfd::Actor`!

daemon/src/maker_inc_connections.rs
thomaseizinger: But here we are talking about the `maker_inc_connections::Actor`! :)

This map is (unfortunately) in two places and we need to populate both of them.

daemon/src/maker_cfd.rs
```diff
@@ -67,9 +60,19 @@ pub struct TakerDisconnected {
     pub id: Identity,
 }
 
-pub struct CfdRollOverCompleted {
-    pub order_id: OrderId,
-    pub dlc: Result<Dlc>,
+#[allow(clippy::large_enum_variant)]
+pub enum RollOverCompleted {
```
thomaseizinger: I think for the other protocols, we put this struct into the same file as the actor that produces it. Please do the same here.

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1,296 @@
...
+use futures::{future, SinkExt};
+use xtra::prelude::MessageChannel;
+use xtra::{Context, Handler, KeepRunning};
+
+pub struct AcceptRollOver;
+
+pub struct RejectRollOver;
+
+#[allow(clippy::large_enum_variant)]
+pub enum Completed {
```
thomaseizinger: This is duplicated? We only need one of those!

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1,296 @@
...
+            Role::Maker,
+            dlc,
+            self.n_payouts,
+        )
+        .await;
+
+        match rollover {
+            Ok(dlc) => self.handle_updated_contract(order_id, dlc, ctx).await?,
+            Err(error) => self.handle_failed(order_id, error, ctx).await?,
+        };
```
thomaseizinger: This won't work. You are blocking this handler function here until the protocol is complete, yet you expect it to make progress by taking messages out of `receiver` which is only populated by another handler function (line 278).

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.maker_cfd_actor.send(msg).await?;
+        ctx.stop();
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl Handler<AcceptRollOver> for Actor {
+    async fn handle(&mut self, msg: AcceptRollOver, ctx: &mut Context<Self>) {
+        log_error!(self.handle_accept_rollover(msg, ctx));
```
thomaseizinger: We are trying to phase out `log_error`, please don't add any more uses unless really necessary. Have a look at what Lucas did in the other actors for alternatives.

dprint.json
```diff
@@ -11,7 +11,7 @@
   "excludes": ["**/target", "**/sqlx-data.json", "maker-frontend/dist", "taker-frontend/dist", "**/node_modules"],
   "plugins": [
     "https://plugins.dprint.dev/markdown-0.9.2.wasm",
-    "https://plugins.dprint.dev/rustfmt-0.4.0.exe-plugin@c6bb223ef6e5e87580177f6461a0ab0554ac9ea6b54f78ea7ae8bf63b14f5bc2",
+    "rustfmt.exe-plugin@7723878c1be78a571c37dcb7f92a0c5780fcfd714a23498260b575efb7e610f3",
```
thomaseizinger: Why is this change necessary and what does it accomplish?

daemon/src/maker_inc_connections.rs
thomaseizinger: Checkout how `maker_inc_connections::settlement::Response` is done.

thomaseizinger (changes requested):
We are currently not testing rollover but this wouldn't work unfortunately :/

Left some comments inline!

daemon/src/rollover_maker.rs
rishflab: I don't understand this. Sorry this might be a really dumb question.

Doesn't `setup_contract::roll_over().await` execute asynchronously, allowing other handler functions to run while it is executing?

the `sender` is set on line 157, `self.sent_from_taker = Some(sender)`,  before `setup_contract::roll_over().await`. The sender is used in `handle_rollover_msg` handler to send the `RolloverMsg` to the receiver. Shouldn't this happen concurrently?

daemon/src/maker_inc_connections.rs
rishflab: > But here we are talking about the `maker_inc_connections::Actor`! :)
> 
> This map is (unfortunately) in two places and we need to populate both of them.

Ahh i see the problem now

daemon/src/maker_cfd.rs
```diff
@@ -347,6 +238,96 @@ where
...
+                ..
+            } => (),
+            _ => {
+                anyhow::bail!("Order is in invalid state. Cannot propose roll over.")
+            }
+        };
+
+        let this = ctx.address().expect("acquired own address");
+
+        let (rollover_actor_addr, _rollover_actor_future) = rollover_maker::Actor::new(
```
thomaseizinger: You can't just ignore this future otherwise the actor will not run :upside_down_face:

Add it to `self.tasks`.

daemon/src/maker_cfd.rs
```diff
@@ -642,104 +588,17 @@ where
...
-            .expect("actor to be able to give address to itself");
-
-        let task = async move {
-            let dlc = contract_future.await;
-
-            this.send(CfdRollOverCompleted { order_id, dlc })
-                .await
-                .expect("always connected to ourselves")
+impl<O, M, T, W> Actor<O, M, T, W> {
+    async fn handle_reject_rollover(&mut self, msg: RejectRollOver) -> Result<()> {
```
thomaseizinger: Can we move that in the same impl block as `accept`?

thomaseizinger (approved):
One more critical thing, otherwise ACK.

PR #812, opened 2021-12-07
Make wallet sync more resilient

thomaseizinger:


daemon/src/wallet.rs
```diff
@@ -118,15 +118,15 @@ impl Actor {
 impl Actor {
     pub fn handle_sync(&mut self, _msg: Sync) -> Result<()> {
         let wallet_info_update = match self.sync_internal() {
-            Ok(wallet_info) => Some(wallet_info),
+            Ok(wallet_info) => wallet_info,
             Err(e) => {
                 tracing::debug!("{:#}", e);
 
-                None
+                std::process::exit(1); // Let's hope we have a supervising process.
```
klochowicz: Perhaps we could add more info in the commit description that motivated us to do this for now?

e.g. "After wallet sync fails, the user cannot add/withdraw funds. As a temporary workaround, we're shutting down the application in anticipation that a supervisor will start it again (the wallet DB gets created from scratch at startup, fixing the bug)"

Also: the commit summary could be more specific here "Kill the app if wallet sync fails" (we have sync messages from monitor and oracle too)

klochowicz (approved):
LGTM! (let's keep talking about a better way than exiting app, actor supervision is the way to go beyond umbrel milestone)

I'll try to break it with my "lucky" seeds :)

daemon/src/wallet.rs
thomaseizinger: Updated.

PR #815, opened 2021-12-07
Use async-native esplora backend based on reqwest

thomaseizinger:
This is blocked on figuring out what to do with the `monitor::Actor` which has been using the `electrum_client` re-exported from bdk.


thomaseizinger:
Closing for now, not much value in here.

PR #819, opened 2021-12-07
Don't store order in database

### LINKED ISSUE #657: Order and CFD model is entangled
Currently a `Cfd` _embeds_ an `Order` in our model.

This is causing problems (see #627) because an `Order` is immutable in our domain whereas a `Cfd` can get updated through roll-over. Ultimately, we need to untangle the two of them such that a `Cfd` _copies_ all necessary data from an `Order` upon creation so we can update it individually if we want to.

Note that "Update" here doesn't necessarily mean we use an `SQL UPDATE`. The current `Cfd` could still be a projection of multiple, immutable events (an event-sourcing-ish design) but even to support that, we need to untangle the `Order` from it to make it clear, where the information is coming from.

I would recommend tackling this by first enforcing privacy of a `Cfd`s fields (https://github.com/itchysats/itchysats/issues/646) because it will make this change much more contained and touch less areas of the codebase.

thomaseizinger:
Todo:

- [x] Needs manual testing

Tested opening a CFD locally. All fine.

### DIFF
--- a/daemon/migrations/20211210000000_remove-order-from-database.sql
+++ b/daemon/migrations/20211210000000_remove-order-from-database.sql
@@ -1,0 +1,31 @@
+drop table cfd_states;
+drop table cfds;
+drop table orders;
+
+create table if not exists cfds
+(
+    id                               integer primary key autoincrement,
+    uuid                             text unique not null,
+    trading_pair                     text        not null,
+    position                         text        not null,
+    initial_price                    text        not null,
+    leverage                         integer     not null,
+    liquidation_price                text        not null,
+    creation_timestamp_seconds       integer     not null,
+    settlement_time_interval_seconds integer     not null,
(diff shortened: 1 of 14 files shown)


da-kami:
@itchymax @scratchscratchscratchy this wipes the DB again, is that OK or do we want a migration in this case?
The migration is possible but somewhat complex. If we decide to let the next release include the event sourcing DB changes already the migration is probably not worth it - because event sourcing cannot be migrated. We will have to wipe. 
So, the question for me would be: When is event sourcing gonna go into the DB?

thomaseizinger (commented):
Thank you for finishing this!

klochowicz (approved):
nice refactor!

da-kami:
Solving merge conflicts :)

PR #826, opened 2021-12-07
A few frontend improvements

bonomat:
1. fix: improves reference price being undefined

2. fix: moves the alert box below the grid to allow for longer messages
old: 
image

3. fix: @da-kami thought this was a cashing problem 🙈

4. fix: make the input full width
5. fix: makes the button size fix width

new

image
image

### DIFF
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -156,133 +156,135 @@ export default function Trade({
     }
 
     return (
-        <Center>
-            <Grid
-                templateRows="repeat(1, 1fr)"
-                templateColumns="repeat(1, 1fr)"
-                gap={4}
-            >
-                <GridItem colSpan={1}>
-                    <Center>
-                        <MotionBox
-                            variants={{
-                                pulse: {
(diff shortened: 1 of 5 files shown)


klochowicz (approved):
thanks for getting rid of these annoyances! :) 👍

taker-frontend/src/components/Trade.tsx
```diff
@@ -214,13 +214,20 @@ export default function Trade({
                                 padding="3"
                                 spacing="6"
                             >
-                                <Button colorScheme="red" size="lg" disabled h={16}>
+                                <Button colorScheme="red" size="lg" disabled h={16} w={"40"}>
```
thomaseizinger, line 217: Would `minWidth` maybe be a better choice to also account for large quantities?

thomaseizinger (approved):
Message of 2nd commit is a bit too long and thus breaks in the rendering on GitHub. If you keep the subject line to 50 chars, the commit message looks better in most places where it is being rendered.

da-kami (approved):

PR #831, opened 2021-12-08
Keep BitMex API connection API open

### LINKED ISSUE #736: bitmex price feed - failed to read quote from websocket
Since Friday, I sporadically bump into this:

```
2021-11-28 23:08:15  WARN Error on websocket stream: WebSocket protocol error: Connection reset without closing handshake
2021-11-28 23:08:15  WARN Failed to read quote from websocket
```

has anyone experienced same issues?

thomaseizinger:
To keep the BitMex API connection open, we need to send `Ping` messages to it.
To send `Ping` messages, we need access to the `Sink` API of the websocket connection.
By get access to the `Sink` API, we must not use any of the `Stream` combinators on the connection.
Not using any of the stream combinators makes it clunky to retrieve an initial quote before we start the loop.

Relying on an initial quote being available has been annoying in the past, and also doesn't help with ideas like https://github.com/itchysats/itchysats/discussions/753.

We also already had some hacky solution in place that defaulted to 0 for errors in calculating the profit.

Fix both problems by rendering only a part of the CFD blank:

Screenshot from 2021-12-08 11-47-35

Fixes #736.

### DIFF
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -43,7 +43,7 @@ impl Actor {
         self.tasks.add(connect_until_successful(this));
     }
 
-    async fn handle(&mut self, _: Connect, ctx: &mut xtra::Context<Self>) -> Result<Quote> {
+    async fn handle(&mut self, _: Connect, ctx: &mut xtra::Context<Self>) -> Result<()> {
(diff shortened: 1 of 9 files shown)


luckysori (approved):
LGTM, but I haven't really looked at this part of the codebase until now.

da-kami (approved):
Definitely an improvement! Thank you.

PR #836, opened 2021-12-08
Introduce a supervisor actor that can restart another actor.

thomaseizinger:


daemon/src/supervisor.rs
```diff
@@ -0,0 +1,114 @@
...
+        tracing::info!("Spawning new instance of actor"); // TODO: Include name
+
+        let this = ctx.address().expect("we are alive");
+        let actor = (self.ctor)(this);
+
+        self.tasks.add(self.context.attach(actor));
+    }
+}
+
+// TODO: Should a supervisor act as proxy?
```
klochowicz: interesting idea! if you'll go with this, perhaps it would be easier to reframe our thinking and call them `Supervised<Actor>` (ie. supervised actors?).

klochowicz: it seems like it is already owning the actor that is being supervised, so it would be more convenient to prevent handing out any references (`xtra::Address` in this case) that could outlive the supervisor - it would be easiest to only hand out the ones to the supervisor to control the lifetime of both.

daemon/src/supervisor.rs
```diff
@@ -0,0 +1,114 @@
+use std::fmt;
+
+use crate::Tasks;
+use async_trait::async_trait;
+use xtra::{Address, Context, Message};
+use xtra_productivity::xtra_productivity;
+
+/// A supervising actor reacts to messages from the actor it is supervising and restarts it based on
```
klochowicz, line 7: this is amazing. I am really looking forward to using it.
Perhaps someday we could plug-in some "Hypervisor" that could monitor the health of the whole actor system (ie. which are essential to run, which failures are fatal etc.) - some rudimentary way of doing that could be possible even now passing in a custom restart policy (this shows how extendable is this design, nice!).

I guess it would be beneficial for merging the PR to add a few tests of the recovery and (potentially) forwarding into the actor that's being supervised from outside.

daemon/src/supervisor.rs
thomaseizinger: I am currently leaning towards "no" to this question and instead hand out the address as part of the constructor like it is currently done.

daemon/src/supervisor.rs
```diff
@@ -0,0 +1,90 @@
...
+            ctor: Box::new(ctor),
+            tasks: Tasks::default(),
+            restart_policy: Box::new(restart_policy),
+        };
+
+        (supervisor, address)
+    }
+
+    fn spawn_new(&mut self, ctx: &mut Context<Self>) {
+        tracing::info!("Spawning new instance of actor"); // TODO: Include name
```
luckysori, line 43: 🤔 Perhaps a good opportunity to use the `ActorName` trait that was recently added.

luckysori (approved):
I like it!

PR #837, opened 2021-12-08
Counterparty identity

da-kami:
Disclaimer: This resets the database! - i.e. it drops `cfd_states` and `cfds` and re-creates them. 

Since there are no users this does not matter, but for those how were testing on mainnet ( @klochowicz ): Remaining open will be committed + cet from the maker side.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -3,6 +3,7 @@
 use crate::db::load_all_cfds;
 use crate::maker_cfd::{FromTaker, TakerConnected};
 use crate::model::cfd::{Cfd, Order, UpdateCfdProposals};
+use crate::model::Identity;
 use crate::oracle::Attestation;
 use crate::tokio_ext::FutureExt;
 use address_map::Stopping;
@@ -235,6 +236,7 @@ where
         maker_heartbeat_interval: Duration,
         connect_timeout: Duration,
         projection_actor: Address<projection::Actor>,
+        maker_identity: Identity,
     ) -> Result<Self>
     where
         F: Future<Output = Result<M>>,
@@ -261,6 +263,7 @@ where
             monitor_addr.clone(),
             oracle_addr,
             n_payouts,
+            maker_identity,
         )
         .create(None)
         .run();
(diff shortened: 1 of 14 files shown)


daemon/src/lib.rs
```diff
@@ -261,6 +263,7 @@ where
             monitor_addr.clone(),
             oracle_addr,
             n_payouts,
+            maker_identity,
```
klochowicz, line 266: nit: `maker_id` should be sufficient? (it's called `taker_id` in the other actor already...)

daemon/src/model/cfd.rs
```diff
@@ -577,16 +577,24 @@ pub struct Cfd {
     pub order: Order,
     pub quantity_usd: Usd,
     pub state: CfdState,
+
+    pub counterparty_identity: Identity,
```
klochowicz: nit: `counterparty_id` or even `counterparty`?

daemon/src/projection.rs
```diff
@@ -500,6 +500,8 @@ pub struct Cfd {
 
     #[serde(with = "::time::serde::timestamp")]
     pub expiry_timestamp: OffsetDateTime,
+
+    pub counterparty_identity: Identity,
```
klochowicz: don't you need to also add the Identity in the both frontends to be displayed/parsed correctly? will it not break the UI as it is (have you tested both maker and taker?)

`interface Cfd` would probably require changes it 2 places, as we copied the file with creation of the new frontend :(

klochowicz (commented):
just a question about the frontend state after changing the type in the Cfd feed and some naming suggestions (feel free to ignore those if you don't agree / don't have time)

daemon/src/projection.rs
bonomat: It will just be ignored in the UI.

klochowicz: good to know! but maybe it would be worthwhile to keep them in sync? (we don't need to display them rn, we do it already for a few fields)

daemon/src/lib.rs
da-kami: That's not aligned throughout the codebase. I'd say we do that in a followup.

thomaseizinger (approved):

klochowicz (approved):
LGTM!

PR #861, opened 2021-12-10
Phase out `log_error` macro

### LINKED ISSUE #553: Remove instances of `log_error!`
Logging is a form of error handling. By logging the error in the handler, we prematurely perform a form of error handling, thus making it impossible to chose a different form of error handling at other call sites.

Goal: Remove all usages of `log_error!` and log them on the call-site of the message sending. In a later stage, we can then change the error handling to something different.

luckysori:
Fixes #553.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -26,7 +26,6 @@ use xtra::Address;
 
 pub mod sqlx_ext; // Must come first because it is a macro.
 
-pub mod actors;
 pub mod address_map;
 pub mod auth;
 pub mod auto_rollover;
@@ -67,6 +66,7 @@ pub mod try_continue;
 pub mod tx;
 pub mod wallet;
 pub mod wire;
+pub mod xtra_ext;
 
 // Certain operations (e.g. contract setup) take long time in debug mode,
 // causing us to lag behind in processing heartbeats.
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -20,6 +20,7 @@ use crate::wire::EncryptedJsonCodec;
 use crate::wire::MakerToTaker;
 use crate::wire::TakerToMaker;
 use crate::wire::Version;
+use crate::xtra_ext::LogFailure;
 use crate::Tasks;
 use anyhow::bail;
 use anyhow::Context;
@@ -138,6 +139,7 @@ impl Actor {
             let _ = self
                 .taker_disconnected_channel
                 .send(maker_cfd::TakerDisconnected { id: *taker_id })
+                .log_failure("Failed to inform about taker disconnect")
                 .await;
             let _ = self.connection_tasks.remove(taker_id);
         }
@@ -219,7 +221,7 @@ impl Actor {
         let this = ctx.address().expect("self to be alive");
         let read_fut = async move {
             while let Ok(Some(msg)) = read.try_next().await {
-                let res = this.send(FromTaker { taker_id, msg }).await;
+                let res = this.send(FromTaker { taker_id, msg }).log_failure("").await;
 
                 if res.is_err() {
                     break;
@@ -249,6 +251,7 @@ impl Actor {
         let _ = self
             .taker_connected_channel
             .send(maker_cfd::TakerConnected { id: taker_id })
+            .log_failure("Failed to report new taker connection")
             .await;
 
         Ok(())
@@ -342,7 +345,7 @@ impl Actor {
 
 #[xtra_productivity(message_impl = false)]
 impl Actor {
-    async fn handle_msg_from_taker(&mut self, msg: FromTaker) {
+    async fn handle_msg_from_taker(&mut self, msg: FromTaker) -> Result<()> {
         use wire::TakerToMaker::*;
         match msg.msg {
             Protocol { order_id, msg } => match self.setup_actors.get_connected(&order_id) {
@@ -380,6 +383,8 @@ impl Actor {
                 let _ = self.taker_msg_channel.send(msg);
             }
         }
+
+        Ok(())
     }
 
     async fn handle_setup_actor_stopping(&mut self, message: Stopping<setup_maker::Actor>) {
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -1,4 +1,3 @@
-use crate::log_error;
 use crate::model;
 use crate::model::cfd::CetStatus;
 use crate::model::cfd::Cfd;
@@ -333,7 +332,7 @@ where
         Ok(())
     }
 
-    async fn handle_oracle_attestation(&mut self, attestation: oracle::Attestation) -> Result<()> {
+    async fn handle_oracle_attestation(&mut self, attestation: oracle::Attestation) {
         for (order_id, MonitorParams { cets, .. }) in self
             .cfds
             .clone()
@@ -342,8 +341,6 @@ where
         {
             try_continue!(self.monitor_cet_finality(cets, attestation.clone(), order_id))
         }
-
-        Ok(())
     }
 
     async fn update_state(
@@ -433,7 +430,7 @@ where
 
                     for (target_status, event) in reached_monitoring_target {
                         tracing::info!(%txid, target = %target_status, current = %status, "Bitcoin transaction reached monitoring target");
-                        self.event_channel.send(event).await?;
+                        self.event_channel.send(event).await??;
                     }
                 }
             }
@@ -628,7 +625,7 @@ fn map_cets(
 }
 
 impl xtra::Message for Event {
-    type Result = ();
+    type Result = Result<()>;
 }
 
 impl xtra::Message for Sync {
@@ -669,14 +666,16 @@ where
     C: bdk::electrum_client::ElectrumApi + Send + 'static,
 {
     async fn handle(&mut self, _: Sync, _ctx: &mut xtra::Context<Self>) {
-        log_error!(self.sync());
+        if let Err(e) = self.sync().await {
+            tracing::warn!("Sync failed: {:#}", e);
+        }
     }
 }
 
 #[async_trait]
 impl xtra::Handler<oracle::Attestation> for Actor {
     async fn handle(&mut self, msg: oracle::Attestation, _ctx: &mut xtra::Context<Self>) {
-        log_error!(self.handle_oracle_attestation(msg));
+        self.handle_oracle_attestation(msg).await
     }
 }
 
@@ -852,8 +851,10 @@ mod tests {
 
     #[async_trait]
     impl xtra::Handler<Event> for MessageRecordingActor {
-        async fn handle(&mut self, message: Event, _ctx: &mut xtra::Context<Self>) {
+        async fn handle(&mut self, message: Event, _ctx: &mut xtra::Context<Self>) -> Result<()> {
             self.events.push(message);
+
+            Ok(())
         }
     }
 
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -1,9 +1,9 @@
-use crate::log_error;
 use crate::model::cfd::Cfd;
 use crate::model::cfd::CfdState;
 use crate::model::BitMexPriceEventId;
 use crate::tokio_ext;
 use crate::try_continue;
+use crate::xtra_ext::LogFailure;
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
@@ -188,6 +188,7 @@ impl Actor {
                     id: event_id,
                     attestation,
                 })
+                .log_failure("Failed to send attestation to oracle::Actor")
                 .await?;
 
                 Ok(())
@@ -257,8 +258,13 @@ pub struct NoAnnouncement(pub BitMexPriceEventId);
 
 #[async_trait]
 impl xtra::Handler<NewAttestationFetched> for Actor {
-    async fn handle(&mut self, msg: NewAttestationFetched, _ctx: &mut xtra::Context<Self>) {
-        log_error!(self.handle_new_attestation_fetched(msg.id, msg.attestation));
+    async fn handle(
+        &mut self,
+        msg: NewAttestationFetched,
+        _ctx: &mut xtra::Context<Self>,
+    ) -> Result<()> {
+        self.handle_new_attestation_fetched(msg.id, msg.attestation)
+            .await
     }
 }
 
@@ -305,7 +311,7 @@ impl xtra::Message for Attestation {
 }
 
 impl xtra::Message for NewAttestationFetched {
-    type Result = ();
+    type Result = Result<()>;
 }
 
 mod olivia_api {
--- a/daemon/src/rollover_maker.rs
+++ b/daemon/src/rollover_maker.rs
@@ -20,6 +20,7 @@ use crate::tokio_ext::spawn_fallible;
 use crate::wire;
 use crate::wire::MakerToTaker;
 use crate::wire::RollOverMsg;
+use crate::xtra_ext::LogFailure;
 use crate::Cfd;
 use crate::Stopping;
 use anyhow::Context as _;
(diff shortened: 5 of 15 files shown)


daemon/src/send_future_ext.rs
```diff
@@ -0,0 +1,35 @@
...
+    async fn warn(self);
+}
+
+#[async_trait]
+impl<E: Error + Send, A: xtra::Actor, M: xtra::Message<Result = Result<(), E>>> Warn
+    for xtra::address::SendFuture<A, M>
+{
+    async fn warn(self) {
+        let msg = match self.await {
+            Err(e) => format!("{:#}", e),
```
thomaseizinger: I think we should still return`Disconnected` here.

daemon/src/send_future_ext.rs
```diff
@@ -0,0 +1,35 @@
+use async_trait::async_trait;
+use std::error::Error;
+
+#[async_trait]
+pub trait Warn {
+    async fn warn(self);
```
thomaseizinger: Perhaps pass in a `&str` so we  can add a message to the error?

daemon/src/lib.rs
```diff
@@ -47,6 +47,7 @@ pub mod projection;
 pub mod rollover_taker;
 pub mod routes;
 pub mod seed;
+pub mod send_future_ext;
```
thomaseizinger:
```suggestion
pub mod xtra_ext;
```

Probably go after the crate.

daemon/src/send_future_ext.rs
thomaseizinger:
```suggestion
    async fn log_failure(self);
```

A bit longer but I think easier to understand.

thomaseizinger (commented):
I think that looks quite good already!

Some suggestions.

daemon/src/send_future_ext.rs
luckysori: I don't think we've been making a clear distinction in terms of handling actor disconnects and handler failures. Doing this would suggest to me that we should almost always bubble up actor disconnects. Is that what you see?

Otherwise, wouldn't this change just make this trait method less convenient?

daemon/src/send_future_ext.rs
luckysori: > Perhaps pass in a `&str` so we can add a message to the error?

What does the final log string look like in your mind. I'm not convinced we need this, but maybe it could be an `Option`.

daemon/src/send_future_ext.rs
thomaseizinger: We should be making a distinction as in most cases, an actor being disconnected is a bug I'd say? Now the question is, do we want to handle that in one place or explicitely at every call site?

The only example I can think of where it might not be a bug is if we tie the lifetime of a network connection to an actor.

daemon/src/send_future_ext.rs
thomaseizinger: I see it as an equivalent to `anyhow::Context`.

Very likely, a fallible handler has more than one `?` in it so the overall context of what bigger operation failed is lost.

What we could do is introduce a trait similar to `ActorName` for messages and include a generic message here that says `"Message handler for {} failed: {:#}"`.

Probably can even unify that to a single `Name` trait or - dare I say it - use `Display` 😁

daemon/src/oracle.rs
```diff
@@ -195,7 +196,10 @@ impl Actor {
     ) -> Result<()> {
         tracing::info!("Fetched new attestation for {}", id);
 
-        let _ = self.attestation_channel.send(attestation).await;
+        self.attestation_channel
+            .send(attestation)
+            .log_failure()
+            .await?;
```
thomaseizinger: To preserve semantics, we should do `let _ = ` here.

daemon/src/fan_out.rs
```diff
@@ -1,32 +1,92 @@
...
 }
 
-impl<M> xtra::Actor for Actor<M> where M: xtra::Message<Result = ()> {}
+impl<M> xtra::Actor for Actor<M> where M: xtra::Message<Result = Result<()>> {}
 
 #[async_trait::async_trait]
 impl<M> xtra::Handler<M> for Actor<M>
 where
-    M: xtra::Message<Result = ()> + Clone + Sync + 'static,
+    M: xtra::Message<Result = Result<()>> + Clone + Sync + 'static,
```
thomaseizinger: Is this change necessary?

daemon/src/fan_out.rs
```diff
@@ -1,32 +1,92 @@
...
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn errors_display() -> Result<()> {
```
thomaseizinger: Can we assert the actual display impl here?

daemon/src/send_future_ext.rs
luckysori: I agree that it's always unexpected when the receiving actor is down. But I think we have been primarily logging these errors, so I suspect if we return the `Disconnected` error we're just gonna log at every call site.

If you want the flexibility of being able to handle the `Disconnected` error, could we introduce another trait which warns about the disconnect. I'm just worried about ergonomics and code readability if we just match against the error everywhere.

daemon/src/fan_out.rs
luckysori: Sorry, I was just testing things for myself. Do you think it's worth it to keep this test?

daemon/src/oracle.rs
luckysori: Do we want to preserve the semantics though? We weren't taking advantage of returning a `Result` before. I can put in a separate commit if you prefer.

thomaseizinger: I think in this first transition yes. We can then re-evaluate, where we fail for what reason.

daemon/src/fan_out.rs
thomaseizinger: Well, I am not convinced we need this entire construct? Do we need to change `fan_out` to return a result? Note that I think we can also just drop the usage of `fan_out` if it is too painful.

daemon/src/fan_out.rs
luckysori: I don't think it's strictly necessary, but I think it goes with the theme of phasing out `log_error`:

1. We want to, in general, shift from logging at the handler to returning errors to the sender.
2. The `fan_out::Actor` was preventing us from passing errors through back to the sender.

I think it would be better if we could fan out messages of both types here i.e. `()` and `Result<()>`. Is there any way we could do that?

thomaseizinger: > We want to, in general, shift from logging at the handler to returning errors to the sender.

Yes but in this case, `fan_out` was explicitly designed to only work with messages that never fail to avoid exactly the decision making of "what do we do if one of the messages fails?"

So unless this is a change motivated by another change that is more related to the `log_error` macro, I'd leave that in place for now.

daemon/src/send_future_ext.rs
luckysori: I prefer `Name` here since we probably only want to log the name of the type. If not I would use `Debug`.

daemon/src/oracle.rs
luckysori: I've tried to follow your advice and preserve semantics all across the board. But I may have slipped, so do keep an eye out for mistakes like this one in your next review.

luckysori:
I'm not sure if I will want to keep all these commits, but they should help at review time, @thomaseizinger.

daemon/src/oracle.rs
```diff
@@ -308,6 +309,12 @@ impl xtra::Message for Attestation {
     type Result = Result<()>;
 }
 
+impl xtra_ext::Name for Attestation {
+    fn name() -> String {
+        "Attestation".to_string()
+    }
+}
```
thomaseizinger: This wants to be a custom derive :D

daemon/src/oracle.rs
```diff
@@ -257,8 +259,13 @@ pub struct NoAnnouncement(pub BitMexPriceEventId);
 
 #[async_trait]
 impl xtra::Handler<NewAttestationFetched> for Actor {
```
thomaseizinger, line 260: Perhaps make a note somewhere that this can use `xtra-productivity` now.

daemon/src/monitor.rs
```diff
@@ -666,8 +665,8 @@ impl<C> xtra::Handler<Sync> for Actor<C>
 where
     C: bdk::electrum_client::ElectrumApi + Send + 'static,
 {
-    async fn handle(&mut self, _: Sync, _ctx: &mut xtra::Context<Self>) {
-        log_error!(self.sync());
+    async fn handle(&mut self, _: Sync, _ctx: &mut xtra::Context<Self>) -> Result<()> {
```
thomaseizinger: Does this actually compile by itself? We are sending this `Sync` message using notify_interval I believe (which would be nice if we would start this from the `started()` lifecycle hook. If this compiles, then I think we might kill the sync if we return a `Result` because then this loop might stop.

daemon/src/maker_inc_connections.rs
```diff
@@ -246,10 +247,10 @@ impl Actor {
         tasks.add(write_fut);
         self.connection_tasks.insert(taker_id, tasks);
 
-        let _ = self
-            .taker_connected_channel
+        self.taker_connected_channel
             .send(maker_cfd::TakerConnected { id: taker_id })
-            .await;
+            .log_failure("Failed to report new taker connection")
+            .await?;
```
thomaseizinger: Not semantically equivalent.

daemon/src/maker_inc_connections.rs
```diff
@@ -136,10 +136,15 @@ impl Actor {
...
-            let _ = self
+            if self
                 .taker_disconnected_channel
                 .send(maker_cfd::TakerDisconnected { id: *taker_id })
-                .await;
+                .log_failure("Failed to inform about taker disconnect")
+                .await
+                .is_err()
+            {
+                tracing::warn!("TakerDisconnected receiver actor is disconnected");
```
thomaseizinger: I am not sure we really need those? Bloats things quite a bit.

daemon/src/rollover_maker.rs
```diff
@@ -140,7 +140,7 @@ impl Actor {
             order_id: self.cfd.id(),
             dlc,
         };
-        self.maker_cfd_actor.send(msg).await?;
+        self.maker_cfd_actor.send(msg).await??;
```
thomaseizinger: This should use `.log_error`?

daemon/tests/harness/mod.rs
```diff
@@ -367,9 +367,13 @@ impl Taker {
 #[macro_export]
 macro_rules! deliver_event {
     ($maker:expr, $taker:expr, $event:expr) => {
-        tracing::debug!("Delivering event: {:?}", $event);
-        $taker.system.cfd_actor_addr.send($event).await.unwrap();
-        $maker.system.cfd_actor_addr.send($event).await.unwrap();
+        #[allow(unused_must_use)]
```
thomaseizinger, line 370: Why? Shouldn't we `unwrap` here?

thomaseizinger (approved):
Nice work! Definitely some follow-up things that we can do here.

I think most of this is sane :)

thomaseizinger:
> I'm not sure if I will want to keep all these commits, but they should help at review time, @thomaseizinger.

Sorry for the late review, I got carried away coding half way through!
The individual commits were definitely helpful, but I think we should squash for the merge.

daemon/src/oracle.rs
luckysori: For simplicity let's just settle on using `Display`. We can always tweak the `Display` implementations if we want more out of them.

daemon/src/oracle.rs
luckysori: I've opened #905 to keep track of this in general.

daemon/src/monitor.rs
luckysori: Okay, I think I get you. The compiler doesn't complain, but I'll revert this change and log in the handler like we used to.

I think it wouldn't actually kill the notification loop since it ignores the variant of the inner result:

```rust
async move {
    Delay::new(duration).await;
    while addr.do_send(constructor()).is_ok() {
        Delay::new(duration).await
    }
};
```

But we would lose the logging of the error nonetheless.

daemon/src/maker_inc_connections.rs
luckysori: I agree about the bloat, which is why my first thought was to include handling this in the `log_failure` trait. So you would just ignore these errors?

thomaseizinger: At the moment we are ignoring them right? So I don't think we are degrading anything if we continue to ignore them!

It is hard to introduce changes like the current one if you are at the same time trying to change behaviour too much.

Reason I would ignore them in most cases is because they can never happen as by default, actors don't shut down in xtra. It is an explicit action.

daemon/tests/harness/mod.rs
luckysori: Unfortunately we cannot `unwrap` twice unconditionally, as sometimes sending only returns `Result<(), Disconnected>`.

daemon/src/maker_inc_connections.rs
luckysori: > It is hard to introduce changes like the current one if you are at the same time trying to change behaviour too much.

I agree on principal, but logging instead of silently dropping is never going to have unintended side-effects.

> Reason I would ignore them in most cases is because they can never happen as by default, actors don't shut down in xtra. It is an explicit action.

What if we explicitly shut down the receiving end and still want to send messages there? In this case I agree that we can ignore the error, but it seems wrong as a general rule.

luckysori: It the argument is that it can never happen, shouldn't we be `expect`ing?

luckysori:
@thomaseizinger: One final review (hopefully) and then I'll squash everything and complete the rebase.

daemon/src/oracle.rs
luckysori: I think `std::any::type_name` may be all we need for logging and way more concise.

daemon/src/xtra_ext.rs
```diff
@@ -33,22 +33,18 @@ where
...
-    M: xtra::Message<Result = anyhow::Result<()>> + Name,
+    M: xtra::Message<Result = anyhow::Result<()>>,
 {
     async fn log_failure(self, context: &str) -> Result<(), Disconnected> {
         if let Err(e) = self.await? {
             tracing::warn!(
                 "{}: Message handler for message {} failed: {:#}",
                 context,
-                M::name(),
+                std::any::type_name::<M>(),
```
thomaseizinger, line 43: image

thomaseizinger (approved):

daemon/src/maker_inc_connections.rs
```diff
@@ -354,32 +357,26 @@ impl Actor {
...
+                let _ = self
                     .settlement_actors
                     .send(&order_id, collab_settlement_maker::Initiated { sig_taker })
-                    .await
-                    .is_err()
-                {
-                    tracing::warn!(%order_id, "No active settlement");
-                }
+                    .await;
             }
```
thomaseizinger: Sorry to bring this up only now but this change is kind of unintended. The (ephemeral) actors being disconnected here is very much possible which is why I would like to keep these messages.

daemon/tests/harness/mod.rs
thomaseizinger: Ah that is a shame. @klochowicz are you okay with this for now? You introduced the macro but unfortunately, it is no longer as generically applicable. We might need two macros (`deliver_fallible_event` and `deliver_event`) or create some other abstraction on the actual `Maker` and `Taker` structs.

thomaseizinger (changes requested):
One comment, otherwise good to go!

daemon/src/maker_inc_connections.rs
luckysori: Ahh, my bad. I thought I didn't mess this up.

thomaseizinger (approved):

PR #862, opened 2021-12-10
Cleanup ephemeral

luckysori:
Trying to align setup actors with the rest of the ephemeral actors. Let me know if there's anything else.

### DIFF
--- a/daemon/src/address_map.rs
+++ b/daemon/src/address_map.rs
@@ -1,3 +1,4 @@
+use anyhow::Result;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::hash::Hash;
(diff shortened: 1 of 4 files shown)


daemon/src/address_map.rs
```diff
@@ -61,6 +61,32 @@ where
             _ => Err(NotConnected::new::<A>()),
         }
     }
+
+    /// Sends a message to the actor stored with the given key.
+    pub async fn send_fallible<M>(
+        &self,
+        key: &K,
+        msg: M,
+    ) -> Result<anyhow::Result<()>, NotConnected>
```
thomaseizinger: Do we need to qualify here? Don't think so :)

daemon/src/address_map.rs
```diff
@@ -61,6 +61,32 @@ where
...
+            Some(addr) if addr.is_connected() => {
+                if let Err(e) = addr
+                    .send(msg)
+                    .await
+                    .expect("we checked that we are connected")
+                {
+                    return Ok(Err(e));
+                };
+
+                Ok(Ok(()))
```
thomaseizinger: Isn't this equivalent:

```suggestion
                let result = addr.send(msg).await.expect("we checked that we are connected");
                Ok(result)
```

thomaseizinger (approved):

daemon/src/address_map.rs
luckysori: We don't if we import `anyhow::Result`!

daemon/src/address_map.rs
luckysori: Equivalent and better.

thomaseizinger (approved):

PR #864, opened 2021-12-10
Introduce `SendAsyncSafe` trait

luckysori:
This was mentioned in https://github.com/itchysats/itchysats/issues/553#issuecomment-983191673.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -27,6 +27,7 @@ use crate::projection;
 use crate::projection::Update;
 use crate::rollover_maker;
 use crate::rollover_maker::Completed;
+use crate::send_async_safe::SendAsyncSafe;
 use crate::setup_maker;
 use crate::wallet;
 use crate::wire;
@@ -205,11 +206,8 @@ where
     T: xtra::Handler<maker_inc_connections::TakerMessage>,
 {
     async fn handle_taker_connected(&mut self, taker_id: Identity) -> Result<()> {
-        // Need to use `do_send_async` here because we are being invoked from the
-        // `maker_inc_connections::Actor`. Using `send` would result in a deadlock.
-        #[allow(clippy::disallowed_method)]
         self.takers
-            .do_send_async(maker_inc_connections::TakerMessage {
+            .send_async_safe(maker_inc_connections::TakerMessage {
                 taker_id,
                 msg: wire::MakerToTaker::CurrentOrder(self.current_order.clone()),
             })
@@ -413,12 +411,8 @@ where
         // have to remove the current order.
         self.current_order = None;
 
-        // Need to use `do_send_async` here because invoking the
-        // corresponding handler can result in a deadlock with another
-        // invocation in `maker_inc_connections.rs`
-        #[allow(clippy::disallowed_method)]
         self.takers
-            .do_send_async(maker_inc_connections::BroadcastOrder(None))
+            .send_async_safe(maker_inc_connections::BroadcastOrder(None))
             .await?;
 
         self.projection_actor.send(projection::Update(None)).await?;
--- a/daemon/src/send_async_safe.rs
+++ b/daemon/src/send_async_safe.rs
@@ -1,0 +1,57 @@
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait SendAsyncSafe<M, R>
+where
+    M: xtra::Message<Result = R>,
+{
+    /// Send a message to an actor without waiting for them to handle
+    /// it.
+    ///
+    /// As soon as this method returns, we know if the receiving actor
+    /// is alive. If they are, the message we are sending will
+    /// eventually be handled by them, but we don't wait for them to
+    /// do so.
+    async fn send_async_safe(&self, msg: M) -> Result<(), xtra::Disconnected>;
+}
+
+#[async_trait]
+impl<A, M> SendAsyncSafe<M, ()> for xtra::Address<A>
+where
+    A: xtra::Handler<M>,
+    M: xtra::Message<Result = ()>,
+{
+    async fn send_async_safe(&self, msg: M) -> Result<(), xtra::Disconnected> {
+        #[allow(clippy::disallowed_method)]
+        self.do_send_async(msg).await
+    }
+}
+
+#[async_trait]
+impl<A, M, E> SendAsyncSafe<M, Result<(), E>> for xtra::Address<A>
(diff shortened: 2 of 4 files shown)


daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[async_trait]
+impl<A, M> SendAsyncSafe<M> for xtra::Address<A>
+where
+    A: xtra::Handler<M>,
+    M: xtra::Message<Result = ()>,
+{
+    async fn send_async_safe(&self, msg: M) -> Result<(), xtra::Disconnected> {
+        #[allow(clippy::disallowed_method)]
+        self.do_send_async(msg).await
```
thomaseizinger, line 26: Do a `tokio::spawn` here and use `.send` and log the error if it happens.

thomaseizinger (changes requested):

daemon/src/send_async_safe.rs
thomaseizinger: Sorry, for this one there is no error but we can do that for the TODO above!

luckysori: I don't understand these two comments, could you rephrase?

thomaseizinger: Ignore this, I'll post again!

daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,28 @@
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait SendAsyncSafe<M>
```
thomaseizinger: Can we extend this with a type parameter `R`?

daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,28 @@
...
+where
+    M: xtra::Message<Result = ()>,
```
thomaseizinger: And use `R` here?

daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,28 @@
...
+    ///
+    /// As soon as this method returns, we know if the receiving actor
+    /// is alive. If they are, the message we are sending will
+    /// eventually be handled by them, but we don't wait for them to
+    /// do so.
+    async fn send_async_safe(&self, msg: M) -> Result<(), xtra::Disconnected>;
+}
+
+#[async_trait]
+impl<A, M> SendAsyncSafe<M> for xtra::Address<A>
```
thomaseizinger: And implement this twice:

```rust
impl<A, M> SendAsyncSafe<M, ()> for xtra::Address<A> {}
impl<A, M, E: Error> SendAsyncSafe<M, Result<(), E>> for xtra::Address<A> {}
```
And in the 2nd impl we can do:

```rust
let fut = self.clone().send(msg);

tokio::spawn(async move {
	if let Err(e) = fut.await {
		tracing::warn!("Asynchronous message invocation failed: {:#}", e);
	}
})
```

luckysori: Two things:

- I cannot implement the same trait for the same generic type twice (even if the restrictions on its type parameter are different). Is there a workaround for this? I think we may have talked about this recently, but I can't recall.
- What is the difference between your second snippet and simply calling `do_send_async`?

thomaseizinger: > I cannot implement the same trait for the same generic type twice (even if the restrictions on its type parameter are different). Is there a workaround for this? I think we may have talked about this recently, but I can't recall.

Not even if the trait is parameterized with different things? Note how the 2nd type parameter in `SendAsyncSafe` is different between the two impl blocks.

> What is the difference between your second snippet and simply calling do_send_async?

The 2nd impl will log an error if the async invocation fails whereas `do_send_async` will just drop the result, regardless of what it is.

luckysori: > Not even if the trait is parameterized with different things? Note how the 2nd type parameter in SendAsyncSafe is different between the two impl blocks.

I was doing something wrong and your tip fixed it!

> The 2nd impl will log an error if the async invocation fails whereas do_send_async will just drop the result, regardless of what it is.

Of course, that makes sense. Thanks!

daemon/src/maker_cfd.rs
```diff
@@ -301,6 +302,11 @@ where
             None => None,
         };
 
+        // TODO: How do we get rid of this one? The handler on the
+        // other side is fallible, yet we know that it cannot fail
+        // because we know we just added the connection to this new
+        // taker.
```
thomaseizinger: This is resolved now :)

daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,57 @@
...
+            return Err(xtra::Disconnected);
+        }
+
+        let send_fut = self.send(msg);
+
+        #[allow(clippy::disallowed_method)]
+        tokio::spawn(async {
+            let e = match send_fut.await {
+                Ok(Err(e)) => format!("{:#}", e),
+                Err(e) => format!("{:#}", e),
```
thomaseizinger, line 48: This case is extremely unlikely I guess?

daemon/src/send_async_safe.rs
```diff
@@ -0,0 +1,57 @@
...
+                Ok(Ok(())) => return,
+            };
+
+            tracing::warn!("Async message invocation failed: {:#}", e)
```
thomaseizinger, line 52: Looking forward to the `Name` trait adding some context here. (Although I think using `Display` wouldn't be wrong either: https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug)

thomaseizinger (approved):
Nice work!

Eventually, it would be nice to migrate all these xtra utilities into an `xtras` crate :drum:

daemon/src/send_async_safe.rs
luckysori: You're right, it's almost impossible with the check at the top of this method. Worth expecting the first error then?

daemon/src/send_async_safe.rs
luckysori: I can see `Display` working for `Actor`s, since you will almost always just want the name. For some reason I'd expect `Message` impls of `Display` to also show the contents of the message though.

thomaseizinger: For messages, I think I could be convinced either way :D

thomaseizinger (approved):

PR #876, opened 2021-12-13
Make all fields of CFD private

thomaseizinger:


daemon/src/model.rs
```diff
@@ -390,7 +390,7 @@ impl From<Decimal> for Percent {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, sqlx::Type)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, sqlx::Type)]
 pub enum TradingPair {
```
klochowicz, line 394: do you need to implement `Clone` anymore here?

klochowicz (approved):
good work!
I like the new getter `state_mut()` as it makes it a bit clearer that we're going to change the state here. (and it will be easier to grep for places where we actually change the state :) ).

daemon/src/model.rs
thomaseizinger: You need `Clone` to derive `Copy`!

thomaseizinger:
> I like the new getter `state_mut()` as it makes it a bit clearer that we're going to change the state here. (and it will be easier to grep for places where we actually change the state :) ).

Well, it is going away with event-sourcing again :joy:

daemon/src/model.rs
klochowicz: ah, indeed!

PR #877, opened 2021-12-13
Remove oracle event id from cfd

da-kami:
Loading the CFD by oracle event id, based on the "initial oracle event id" that we store in the CFD is wrong.
We should actually not store this even id at all, as it only becomes relevant once the setup is completed and we have a DLC.
If we load based on the event id that was stored in the CFD we might still load based on an outdated id once we roll over!
Thus, the event id of the cfd was removed. We always use the one stored in the DLC.
Since it is very hard to load a CFD based on the event id in the DLC (because that only exists in some states) this was removed.
Instead, upon attestation, we load all CFDs and then decide if we care about the attestation.
This should be optimized to loading all **open** CFDs at some point.

### DIFF
--- a/daemon/migrations/20211213000000_remove-oracle_event_id_from_cfd.sql
+++ b/daemon/migrations/20211213000000_remove-oracle_event_id_from_cfd.sql
@@ -1,0 +1,29 @@
+drop table cfd_states;
+drop table cfds;
+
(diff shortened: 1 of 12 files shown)


thomaseizinger (approved):
Nice work Daniel :)

Issue #865, opened 2021-12-10
Seedphrase based wallet that can easily be backed up


da-kami:
This is something we should really get on Umbrel because it is hard to to achieve it there.

rishflab:
#  Proposed solution: backup/restore through the UI

## Features
1. User can put seed phrase in UI to restore a wallet.
2. User can let UI create a new wallet, displaying a seed phrase once.
3. User can ask UI to show seed phrase anytime they are "logged in" (not sure about this one)

### Tasks

note: pitfall tasks are in bold

####  Daemon
- [ ] **delay wallet init until seed bytes from rest endpoint arrives** (this might drastically change the architecture/actor initialisation)
- [ ] restore from wallet from seed bytes REST endpoint

####  UI
- [ ] Generate 24-word seed phrase
- [ ] Convert seed phrase to bytes
- [ ] Restore wallet from inputted seed phrase
- [ ] Generate wallet/seed phrase and show seed phrase
- [ ] Prompt/component to restore/generate wallet
- [ ] Reveal seed phrase button/seed phrase display in wallet component
- [ ] Store/wipe stored seed phrase securely?

## Follow up features/ Not in scope
1. User can ask UI to show the descriptor allowing user to export to third party wallets

### Daemon
- [ ] export descriptor endpoint
- [ ] export wallet from descriptor methods to wallet actor

thomaseizinger:
Thanks for the initial design!

Here are some ideas:

1. I think we should generate the new seed phrases in the daemon and hand it out to the actor and not generate it in the frontend.
2. As an initial step, I would change our wallet actor to be initialized by a seed phrase and let it derive the descriptor internally. That is an API we will certainly need, regardless of how we design things around it. For now, you can derive the seed-words deterministically from the seedfile. This can be its own PR already.
3. Delaying the wallet initialization is a bit trickier because we currently do this as part of the actor system startup.
	- To delay instantiation of the wallet actor, we need to keep its `xtra::Context` around (we can later call `Context::attach` with an instance of the actor)
	- In the REST endpoint, we only have access to the addresses of actors but not their contexts. Instead of extending this, I would go towards https://github.com/itchysats/itchysats/issues/774. With that, we can `rocket.manage` the entire actor system and thus provide an API on it along the lines of `TakerActorSystem::initialize_wallet_from_seed_phrase(seed_words: &str)` which will internally create a wallet actor and attach it to the context.
[...]

Issue #774, opened 2021-12-01
Too many details about the actor system leak into rocket


thomaseizinger:
Currently, we are managing `xtra::Address`es within rocket and send messages to dedicated actors upon calls to certain endpoints.

As we move away from a monolithic `{taker,maker}_cfd::Actor`, which actor to contact for a certain behaviour may change. For example, the short-lived collaborative settlement actor _could_ be spawned directly from the `maker_inc_connections::Actor` upon receiving the incoming message. That would have the advantage that we would only maintain the state about ongoing collaborative settlements once (as an `AddressMap` inside the `maker_inc_connections::Actor`).

However, changing which actor to send the `Accept` or `Reject` message to (from `maker_cfd::Actor` to `maker_inc_connections::Actor`) requires an unnecessary change to the rocket-managed state.

This highlights that too many details about the actor system are currently leaked to its users.

There are likely multiple ways of fixing this. We should consider doing something like:

```rs
impl MakerActorSystem {
	pub async fn accept_settlement(&self, order_id: OrderId) -> Result<()> {
		todo!()
	}
}
```

And `rocket.manage` the entire `MakerActorSystem`. This would allow us to make changes to the internals of the actor system without them leaking into upper layer.

thomaseizinger:
Another hint that this might be a good idea is that we essentially do exactly this in our tests with the `Maker` and `Taker` structs, thus duplicating the logic of how to for example take an offer. If this functionality were to be encapsulated in the actor system struct, it would be easier to use.

thomaseizinger:
This has been resolved.

PR #878, opened 2021-12-13
Test force-closing CFD

### LINKED ISSUE #564: Actor test for non-collaborative close
Happy path test for non-collaborative close.
Assert on `PendingCet` and `Close`. 

Note that events can happen in different order, i.e. Oracle Attestation can happen before/after Cet timelock expiry (depending on when we publish the commit tx).
We might want to have multiple tests that cover different combinations for this scenario.

klochowicz:
Resolves #564

### DIFF
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -5,6 +5,8 @@ use crate::harness::flow::next_cfd;
 use crate::harness::flow::next_order;
 use crate::harness::flow::next_some;
 use crate::harness::init_tracing;
+use crate::harness::maia::OliviaData;
+use crate::harness::mocks::oracle::dummy_wrong_attestation;
 use crate::harness::start_both;
 use crate::harness::Maker;
 use crate::harness::MakerConfig;
@@ -14,6 +16,7 @@ use daemon::connection::ConnectionStatus;
 use daemon::model::cfd::OrderId;
 use daemon::model::Usd;
 use daemon::monitor::Event;
+use daemon::oracle;
 use daemon::projection::CfdState;
 use daemon::projection::Identity;
 use maia::secp256k1_zkp::schnorrsig;
@@ -129,7 +132,8 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
 #[tokio::test]
 async fn collaboratively_close_an_open_cfd() {
     let _guard = init_tracing();
-    let (mut maker, mut taker, order_id) = start_from_open_cfd_state().await;
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(OliviaData::example_0().announcement()).await;
 
     taker.propose_settlement(order_id).await;
 
@@ -149,6 +153,38 @@ async fn collaboratively_close_an_open_cfd() {
 
     sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
 
+    assert_next_state!(CfdState::Closed, maker, taker, order_id);
+}
+
+#[tokio::test]
+async fn force_close_an_open_cfd() {
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement()).await;
+
+    // Taker initiates force-closing
+    taker.force_close(order_id).await;
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+    assert_next_state!(CfdState::OpenCommitted, maker, taker, order_id);
+
+    // After CetTimelockExpired, we're only waiting for attestation
+    deliver_event!(maker, taker, Event::CetTimelockExpired(order_id));
+
+    // Delivering the wrong attestation does not move state to `PendingCet`
+    deliver_event!(maker, taker, dummy_wrong_attestation());
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+    assert_next_state!(CfdState::OpenCommitted, maker, taker, order_id);
+
+    // Delivering correct attestation moves the state `PendingCet`
+    deliver_event!(maker, taker, oracle_data.attestation());
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+    assert_next_state!(CfdState::PendingCet, maker, taker, order_id);
+
+    deliver_event!(maker, taker, Event::CetFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
     assert_next_state!(CfdState::Closed, maker, taker, order_id);
 }
 
@@ -217,7 +253,8 @@ async fn maker_notices_lack_of_taker() {
 /// Hide the implementation detail of arriving at the Cfd open state.
 /// Useful when reading tests that should start at this point.
 /// For convenience, returns also OrderId of the opened Cfd.
-async fn start_from_open_cfd_state() -> (Maker, Taker, OrderId) {
+/// `announcement` is used during Cfd's creation.
+async fn start_from_open_cfd_state(announcement: oracle::Announcement) -> (Maker, Taker, OrderId) {
     let heartbeat_interval = Duration::from_secs(60);
     let maker_listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
     let mut maker = Maker::start(
@@ -240,8 +277,14 @@ async fn start_from_open_cfd_state() -> (Maker, Taker, OrderId) {
         .await
         .unwrap();
 
-    taker.mocks.mock_oracle_announcement().await;
-    maker.mocks.mock_oracle_announcement().await;
+    taker
+        .mocks
+        .mock_oracle_announcement_with(announcement.clone())
+        .await;
+    maker
+        .mocks
+        .mock_oracle_announcement_with(announcement)
+        .await;
 
     taker.take_order(received.clone(), Usd::new(dec!(5))).await;
     let (_, _) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
--- a/daemon/tests/harness/mocks/oracle.rs
+++ b/daemon/tests/harness/mocks/oracle.rs
@@ -1,9 +1,12 @@
-use crate::harness::maia::OliviaData;
+use daemon::model::BitMexPriceEventId;
 use daemon::oracle;
 use mockall::*;
 use std::sync::Arc;
+use time::OffsetDateTime;
 use tokio::sync::Mutex;
 use xtra_productivity::xtra_productivity;
+
+use crate::harness::maia::OliviaData;
 
 /// Test Stub simulating the Oracle actor.
 /// Serves as an entrypoint for injected mock handlers.
@@ -50,6 +53,18 @@ pub trait Oracle {
     }
 }
 
-pub fn dummy_announcement() -> oracle::Announcement {
-    OliviaData::example_0().announcement()
+/// We do *not* depend on the current time in our tests, the valid combination of
+/// announcement/attestation is hard-coded in OliviaData struct (along with event id's).
+/// Therefore, an attestation based on current utc time will always be wrong.
+pub fn dummy_wrong_attestation() -> oracle::Attestation {
+    let oracle::Attestation {
+        id: _,
+        price,
+        scalars,
+    } = OliviaData::example_0().attestation();
+    oracle::Attestation {
+        id: BitMexPriceEventId::with_20_digits(OffsetDateTime::now_utc()),
+        price,
+        scalars,
+    }
 }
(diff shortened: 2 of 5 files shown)


daemon/clippy.toml
```diff
@@ -1,3 +1,5 @@
 disallowed-methods = [
   "tokio::spawn", # tasks can outlive the actor system, prefer spawn_with_handle()
+  "std::time::SystemTime::now", # prefer our clock provider - clock::now()
+  "time::OffsetDateTime::now_utc", # prefer our clock provider - clock::now().into()
```
thomaseizinger: Latest clippy version allows you to specify this reason as part of the metadata here so it will show up in the error message! https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_methods

daemon/tests/happy_path.rs
```diff
@@ -152,6 +154,46 @@ async fn collaboratively_close_an_open_cfd() {
...
+
+    let start_time = OliviaData::example_0().announcement().expected_outcome_time
+        - daemon::SETTLEMENT_INTERVAL
+        - time::Duration::minutes(20);
+
+    let clock_handle = daemon::clock::set_fake_clock(start_time.into());
+
+    let (mut maker, mut taker, order_id) = start_from_open_cfd_state().await;
+
+    std::mem::drop(clock_handle); // No need to fake time anymore
```
thomaseizinger: This would do the dropping for you :)

```suggestion
    let (mut maker, mut taker, order_id) = {
    	let _clock_handle = daemon::clock::set_fake_clock(start_time.into());
    	start_from_open_cfd_state().await
    };
```

daemon/tests/happy_path.rs
```diff
@@ -152,6 +154,46 @@ async fn collaboratively_close_an_open_cfd() {
...
+
+    // Taker initiates force-closing
+    taker.force_close(order_id).await;
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+
+    assert_next_state!(CfdState::OpenCommitted, maker, taker, order_id);
```
thomaseizinger: Should we just `sleep` as part of `assert_next_state`?

daemon/tests/happy_path.rs
```diff
@@ -152,6 +154,46 @@ async fn collaboratively_close_an_open_cfd() {
...
+
+    sleep(Duration::from_secs(5)).await;
```
thomaseizinger: This seems unnecessary?

daemon/tests/harness/mocks/oracle.rs
```diff
@@ -53,3 +53,7 @@ pub trait Oracle {
 pub fn dummy_announcement() -> oracle::Announcement {
     OliviaData::example_0().announcement()
 }
+
+pub fn dummy_end_attestation() -> oracle::Attestation {
```
thomaseizinger: I think we should inline these. It would be good to see in the test that we are using `example_0` for both.

daemon/tests/harness/mod.rs
```diff
@@ -352,6 +352,15 @@ impl Taker {
...
     }
+
+    pub async fn force_close(&self, order_id: OrderId) {
+        self.system
+            .cfd_actor_addr
+            .send(taker_cfd::Commit { order_id })
+            .await
+            .unwrap()
+            .unwrap();
+    }
```
thomaseizinger, line 363: Doesn't need to happen now but I thought I'd point you towards https://github.com/itchysats/itchysats/issues/774.

If we would have these functions directly on the actor system, we could likely re-design the tests to leverage composition even more and directly operate on the `TakerActorSystem` hence not needing to write this indirection.

.github/workflows/ci.yml
```diff
@@ -92,10 +92,10 @@ jobs:
       - name: Setup rust toolchain
         run: rustup show
       - uses: Swatinem/rust-cache@v1.3.0
-      - run: cargo build --bins --tests
-      - run: cargo test --workspace
+      - run: cargo build --features fake_clock --bins --tests
```
thomaseizinger: Note on the commit message. Perhaps change to:

```
Enable fake_clock feature flag on the CI

Otherwise the force-close test does not run.
```

thomaseizinger (approved):
Nice work! Looking forward to landing this!

daemon/tests/happy_path.rs
```diff
@@ -152,6 +154,46 @@ async fn collaboratively_close_an_open_cfd() {
     assert_next_state!(CfdState::Closed, maker, taker, order_id);
 }
 
+#[tokio::test]
+#[cfg_attr(not(feature = "fake_clock"), ignore)]
+async fn force_close_an_open_cfd() {
+    let _guard = init_tracing();
+
+    let start_time = OliviaData::example_0().announcement().expected_outcome_time
```
luckysori: 🤔 I think the `fake_clock` feature flag is a well-executed and interesting solution, but I'm not totally sold.

The way I see it, we're introducing quite a bit of complexity to get past this check https://github.com/itchysats/itchysats/blob/90999363343d6321443aeb37a63aef04755a88da/daemon/src/oracle.rs#L159-L166 in this test. Couldn't we just make the timestamp in the event ID dynamic?

I'm reluctant to add something to production code that is there just to help us test things. If there is no valid alternative then we can merge this.

luckysori (commented):
Well executed, just not sure about the solution.

daemon/tests/happy_path.rs
thomaseizinger: This is not the only thing actually. With the fake clock, we are also working around how we compute what the next bitmex price event ID is when we send out a new offer for example.

daemon/tests/happy_path.rs
```diff
@@ -152,6 +154,46 @@ async fn collaboratively_close_an_open_cfd() {
...
+
+    let (mut maker, mut taker, order_id) = start_from_open_cfd_state().await;
+
+    std::mem::drop(clock_handle); // No need to fake time anymore
+
+    // Taker initiates force-closing
+    taker.force_close(order_id).await;
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
```
da-kami, line 170: Can we wait for the state transition and if it does *not* happen within x time we fail? 
That would make this more efficient, because we don't always have to wait for 5 seconds. 
Would be cool if we could optimize all the sleeps away.

da-kami (approved):
concept ack! I agree to @thomaseizinger comments and added one additional one. Not blocking though

daemon/tests/happy_path.rs
klochowicz: @luckysori  thanks to the feature flag, we're not adding anything except a forwarding call to `clock` module in production (we could say we're adding one level of abstraction, but no functional changes whatsoever). Compared to the gymnastics we had to go through to allow testing maker and taker actors, it's IMHO negligible (especially that we have clippy lint and it's a 1:1 correspondence with calls, so no mental overhead).

This is fairly lightweight solution, a different one that was contemplated was having a clock provider as an actor that can send messages with current time on a tick and also respond with current time to a message querying the time. The obvious downside would be all the cost of rolling out the actor's address into all the actors that need to ask for the time). The upside of that solution would be not needing a feature flag (we would only mock the clock provider in the test harness).

I'm afraid there's no easier approach here :(

daemon/tests/happy_path.rs
klochowicz: due to how `next_cfd` is implemented, we can't easily optimize them away. I could have a stab at refactoring that function at some point, but I would prefer not to do it as a part of this review.
In short, testing showed that  `next_cfd` currently can return when only one of the parties is in desired state, and we want them both being in there.

daemon/tests/harness/mod.rs
klochowicz: that's interesting! I'm happy to give it a shot on top of this. if not the full redesign, even a short macro `deliver_command!` would get rid of all these functions ;D

daemon/clippy.toml
klochowicz: That's great! I'll open a PR that uses it for all our cases for consistency and better error messages ;)

daemon/tests/happy_path.rs
luckysori: I definitely prefer what you've done over introducing a clock provider actor. If you guys reckon there is no simple way to handle this problem then I guess we should go with this.

daemon/tests/happy_path.rs
klochowicz: indeed, this one was unnecessary!

klochowicz:
@luckysori spotted that on recent master, the test still passes without faking the clock (it didn't beforehand). I'll investigate this closer before resubmitting an update to the review. converting to draft for now.

daemon/clippy.toml
klochowicz: unrelated for this review - will still post a PR with using new clippy functionality though.

daemon/tests/happy_path.rs
klochowicz: it's only needed sometimes - I intend to refactor it away someday, so it's better not to bury it to remind it :)

daemon/tests/harness/mocks/oracle.rs
klochowicz: inlined! and used a common struct in the test to see that they're matched from the same source (`oracle_data`)

daemon/tests/happy_path.rs
klochowicz: thanks to @luckysori trying out the changeset after @da-kami's refactor we don't need to rely on particular implementation of our lookup of the matching attestation, and we do not need to fake the clock anymore. check out the new simplified version :)

daemon/tests/happy_path.rs
```diff
@@ -152,6 +156,40 @@ async fn collaboratively_close_an_open_cfd() {
...
+
+    sleep(Duration::from_secs(5)).await;
+
+    assert_next_state!(CfdState::PendingCet, maker, taker, order_id);
+
+    deliver_event!(maker, taker, Event::CetFinality(order_id));
+
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+
+    assert_next_state!(CfdState::Closed, maker, taker, order_id);
```
thomaseizinger: A nit on test structure, I find excessive use of newlines actually decreases readability. What do you think of this:

```suggestion
    // Delivering the wrong attestation does not move state to `PendingCet`
    deliver_event!(maker, taker, dummy_wrong_attestation());
    assert_next_state!(CfdState::OpenCommitted, maker, taker, order_id);

    deliver_event!(maker, taker, oracle_data.attestation());
    deliver_event!(maker, taker, Event::CetTimelockExpired(order_id));

    sleep(Duration::from_secs(5)).await;
    assert_next_state!(CfdState::PendingCet, maker, taker, order_id);

    deliver_event!(maker, taker, Event::CetFinality(order_id));
    
    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
    assert_next_state!(CfdState::Closed, maker, taker, order_id);
```

The idea is:
- Delivering the two events is related.
- Sleeping before asserting is related.
- Delivering the wrong event is related to the assertion on `OpenCommitted`.

daemon/tests/happy_path.rs
```diff
@@ -152,6 +156,40 @@ async fn collaboratively_close_an_open_cfd() {
...
+    // Taker initiates force-closing
+    taker.force_close(order_id).await;
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+
+    // Delivering the wrong attestation does not move state to `PendingCet`
+    deliver_event!(maker, taker, dummy_wrong_attestation());
+
+    assert_next_state!(CfdState::OpenCommitted, maker, taker, order_id);
```
thomaseizinger: I think this is not 100% sane because nothing would happen unless the CET timelock also expires.

Hence if you want to include this, I think we should deliver that event first. Then do

- wrong attestation
- sleep
- assert `OpenCommitted`
---
- correct attestation
- sleep
- assert `PendingCet`

thomaseizinger (approved):

daemon/tests/happy_path.rs
klochowicz: indeed, that is more foolproof. actioned the suggestion and added more comments in the code to highlight the behaviour.

daemon/tests/happy_path.rs
klochowicz: I've changed the whitespaces that combines the delivery/sleep/assertion together in a group (the structure changed after your second comment so couldn't just apply your suggestion).

daemon/tests/harness/mod.rs
thomaseizinger: #865 for example highlights that this is useful outside of the tests as well. I think the tests here just show us a shortcoming of the abstraction in our production code.

thomaseizinger (approved):

PR #879, opened 2021-12-13
Resilient broadcast

### LINKED ISSUE #873: Be more resilient when broadcasting transactions
When broadcasting transaction we should make sure to distinguish different errors upon transaction publication.

Our code for broadcasting transactions should check for more error cases that should not be treated as error case.
At the moment, we do check for one case of "the transaction already be included", but it seems this is not inclusive enough.

rishflab:
can drop the first commit (probably should)

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -186,9 +186,14 @@ impl Actor {
     pub fn handle_try_broadcast_transaction(
         &mut self,
         msg: TryBroadcastTransaction,
+        ctx: &mut xtra::Context<Self>,
     ) -> Result<Txid> {
         let tx = msg.tx;
         let txid = tx.txid();
+
+        let this = ctx.address().expect("self to be alive");
+
+        this.send(Sync).await??;
 
         let result = self.wallet.broadcast(&tx);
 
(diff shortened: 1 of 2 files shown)


daemon/src/wallet.rs
```diff
@@ -186,10 +186,15 @@ impl Actor {
...
         &mut self,
         msg: TryBroadcastTransaction,
+        ctx: &mut xtra::Context<Self>,
     ) -> Result<Txid> {
         let tx = msg.tx;
         let txid = tx.txid();
 
+        let this = ctx.address().expect("self to be alive");
+
+        this.send(Sync).await??;
```
thomaseizinger, line 196: How is this going to make a difference? Syncing will scan the blockchain for coins sent to addresses derived from the descriptor. Any transaction related to our protocol is completely separate.

thomaseizinger (changes requested):
I think there is a misunderstanding in what the wallet sync does.

daemon/src/wallet.rs
rishflab: hmm i see, back to the drawing board.

PR #880, opened 2021-12-13
Auto-rollover

### LINKED ISSUE #456: Automate roll-over

### LINKED ISSUE #892: lots of log spam after upgrading umbrel to latest itchysats
I get literally swamped with the logs (and for quite a while I couldn't start the app because of them):
```
2021-12-14 09:27:11  WARN daemon::tokio_ext: Task failed: Failed to GET https://h00.ooo/x/BitMEX/BXBT/2021-12-20T14:00:00.price?n=20: error sending request for url (https://h00.ooo/x/BitMEX/BXBT/2021-12-20T14:00:00.price?n=20): error trying to connect: dns error: failed to lookup address information: Temporary failure in name resolution: error trying to connect: dns error: failed to lookup address information: Temporary failure in name resolution: dns error: failed to lookup address information: Temporary failure in name resolution: failed to lookup address information: Temporary failure in name resolution
2021-12-14 09:27:16 DEBUG daemon::oracle: Fetching announcement for /x/BitMEX/BXBT/2021-12-21T10:00:00.price?n=20
```

da-kami:
resolves #456 

Note: I'm looking forward to the event sourcing model. Getting this working in the current state model is quite cumbersome.

Fixes https://github.com/itchysats/itchysats/issues/892. (settlement interval is reduced)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Off-chain perpetual CFD rollover.
+  Hourly, the taker sends a rollover request to the maker for each open position.
+  The maker can accept or reject a rollover proposal.
+  Upon acceptance the taker and maker collaboratively agree on an oracle price event further in the future and generate new payout transactions accordingly.
+  The previous payout transactions are invalidated.
+  The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
+  In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
+
 ## [0.3.0] - 2021-12-09
 
 Initial release for mainnet.
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -1,0 +1,194 @@
+use crate::address_map::AddressMap;
+use crate::address_map::Stopping;
+use crate::cfd_actors::append_cfd_state;
+use crate::connection;
+use crate::db;
+use crate::db::load_cfd;
+use crate::model::cfd::CannotRollover;
+use crate::model::cfd::CfdState;
+use crate::model::cfd::CfdStateCommon;
+use crate::model::cfd::OrderId;
+use crate::monitor;
+use crate::monitor::MonitorParams;
+use crate::oracle;
+use crate::projection;
+use crate::rollover_taker;
+use crate::Tasks;
+use anyhow::Result;
+use async_trait::async_trait;
+use maia::secp256k1_zkp::schnorrsig;
+use std::time::Duration;
+use xtra::Actor as _;
+use xtra::Address;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor<O, M> {
+    db: sqlx::SqlitePool,
+    oracle_pk: schnorrsig::PublicKey,
+    projection_actor: Address<projection::Actor>,
+    conn_actor: Address<connection::Actor>,
+    monitor_actor: Address<M>,
+    oracle_actor: Address<O>,
+    n_payouts: usize,
+
+    rollover_actors: AddressMap<OrderId, rollover_taker::Actor>,
+
+    tasks: Tasks,
+}
+
+impl<O, M> Actor<O, M> {
+    pub fn new(
+        db: sqlx::SqlitePool,
+        oracle_pk: schnorrsig::PublicKey,
+        projection_actor: Address<projection::Actor>,
+        conn_actor: Address<connection::Actor>,
+        monitor_actor: Address<M>,
+        oracle_actor: Address<O>,
+        n_payouts: usize,
+    ) -> Self {
+        Self {
+            db,
+            oracle_pk,
+            projection_actor,
+            conn_actor,
+            monitor_actor,
+            oracle_actor,
+            n_payouts,
+            rollover_actors: AddressMap::default(),
+            tasks: Tasks::default(),
+        }
+    }
+}
+
+#[xtra_productivity]
+impl<O, M> Actor<O, M>
+where
+    M: xtra::Handler<monitor::StartMonitoring>,
+    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+{
+    async fn handle(&mut self, _msg: AutoRollover, ctx: &mut xtra::Context<Self>) -> Result<()> {
+        tracing::trace!("Checking all CFDs for rollover eligibility");
+
+        let mut conn = self.db.acquire().await?;
+        let cfds = db::load_all_cfds(&mut conn).await?;
+
+        let this = ctx
+            .address()
+            .expect("actor to be able to give address to itself");
+
+        for cfd in cfds {
+            let disconnected = match self.rollover_actors.get_disconnected(cfd.id()) {
+                Ok(disconnected) => disconnected,
+                Err(_) => {
+                    tracing::debug!(order_id=%cfd.id(), "Rollover already in progress");
+                    continue;
+                }
+            };
+
+            let (addr, fut) = rollover_taker::Actor::new(
+                (cfd, self.n_payouts),
+                self.oracle_pk,
+                self.conn_actor.clone(),
+                &self.oracle_actor,
+                self.projection_actor.clone(),
+                &this,
+                (&this, &self.conn_actor),
+            )
+            .create(None)
+            .run();
+
+            disconnected.insert(addr);
+            self.tasks.add(fut);
+        }
+
+        Ok(())
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
+impl<O, M> Actor<O, M>
+where
+    O: 'static,
+    M: 'static,
+    M: xtra::Handler<monitor::StartMonitoring>,
+    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+{
+    async fn handle_rollover_completed(&mut self, msg: rollover_taker::Completed) -> Result<()> {
+        use rollover_taker::Completed::*;
+        let (order_id, dlc) = match msg {
+            UpdatedContract { order_id, dlc } => (order_id, dlc),
+            Rejected { .. } => {
+                return Ok(());
+            }
+            Failed { order_id, error } => {
+                tracing::warn!(%order_id, "Rollover failed: {:#}", error);
+                return Ok(());
+            }
+            NoRollover { order_id, reason } => {
+                match reason {
+                    CannotRollover::NoDlc => {
+                        tracing::warn!(%order_id, "Not rolled over: {:#}", reason);
+                    }
+                    CannotRollover::AlreadyExpired
+                    | CannotRollover::WasJustRolledOver
+                    | CannotRollover::WrongState { .. } => {
+                        tracing::debug!(%order_id, "Not rolled over: {:#}", reason);
+                    }
+                }
+                return Ok(());
+            }
+        };
+
+        let mut conn = self.db.acquire().await?;
+        let mut cfd = load_cfd(order_id, &mut conn).await?;
+        *cfd.state_mut() = CfdState::Open {
+            common: CfdStateCommon::default(),
+            dlc: dlc.clone(),
+            attestation: None,
+            collaborative_close: None,
+        };
+
+        append_cfd_state(&cfd, &mut conn, &self.projection_actor).await?;
+
+        self.monitor_actor
+            .send(monitor::StartMonitoring {
+                id: order_id,
+                params: MonitorParams::new(dlc.clone(), cfd.refund_timelock_in_blocks()),
+            })
+            .await?;
+
+        self.oracle_actor
+            .send(oracle::MonitorAttestation {
+                event_id: dlc.settlement_event_id,
+            })
+            .await?;
+
+        Ok(())
+    }
+
+    async fn handle_rollover_actor_stopping(&mut self, msg: Stopping<rollover_taker::Actor>) {
+        self.rollover_actors.gc(msg);
+    }
+}
+
+#[async_trait]
+impl<O, M> xtra::Actor for Actor<O, M>
+where
+    O: 'static,
+    M: 'static,
+    Self: xtra::Handler<AutoRollover>,
+{
+    async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
+        let fut = ctx
+            .notify_interval(
+                rollover_taker::MAX_ROLLOVER_DURATION + Duration::from_secs(60),
+                || AutoRollover,
+            )
+            .expect("we are alive");
+
+        self.tasks.add(fut);
+    }
+}
+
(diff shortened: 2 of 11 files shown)


daemon/src/lib.rs
```diff
@@ -267,20 +268,33 @@ where
...
+            connection_actor_addr.clone(),
+            monitor_addr.clone(),
+            oracle_addr.clone(),
+            n_payouts,
+            maker_identity,
+        )
+        .create(None)
+        .run();
+
+        let (_, auto_rollover_fut) = auto_rollover::Actor::new(
```
thomaseizinger: Need to be careful here. Might need to keep that address around to keep the actor alive. At that point, the question is, should this maybe just be a normal task with an endless loop? I didn't think about that before!

daemon/src/model/cfd.rs
```diff
@@ -1306,6 +1307,56 @@ impl Cfd {
...
+
+    /// Only cfds in state `Open` that have not received an attestation and are within 23 hours
+    /// until expiry are eligible for rollover
+    pub fn can_roll_over(&self, now: OffsetDateTime) -> Result<(), CannotRollover> {
+        let expiry_timestamp = match self.expiry_timestamp() {
+            Some(expiry_timestamp) => expiry_timestamp,
+            None => {
+                // This should not really happen, so let's log a warning in case it does
+                tracing::warn!("Rollover proposal triggered for a CFD without a DLC");
+                return Err(CannotRollover::NoDlc);
```
thomaseizinger: Don't log an error AND return one. This will lead to duplicated logs. Once the log msg is gone, use `ok_or_else`.

daemon/src/model/cfd.rs
```diff
@@ -1306,6 +1307,56 @@ impl Cfd {
...
+            });
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(thiserror::Error, Debug, PartialEq)]
+pub enum CannotRollover {
+    #[error("Cfd does not have a dkc")]
```
```suggestion
    #[error("Cfd does not have a dlc")]
```

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,179 @@
...
+    async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
+        let fut = ctx
+            .notify_interval(Duration::from_secs(60 * 5), || AutoRollover)
+            .expect("we are alive");
+
+        self.tasks.add(fut);
+    }
+}
+
+/// Module private message to check for rollover eligibility on a regular interval.
```
thomaseizinger: Lie.

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,179 @@
+use crate::address_map::AddressMap;
+use crate::address_map::Stopping;
+use crate::cfd_actors::append_cfd_state;
+use crate::connection;
+use crate::db::load_cfd_by_order_id;
+use crate::db::{self};
```
thomaseizinger: Weird import.

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,179 @@
...
+use crate::model::cfd::CfdState;
+use crate::model::cfd::CfdStateCommon;
+use crate::model::cfd::OrderId;
+use crate::monitor::MonitorParams;
+use crate::monitor::{self};
```
thomaseizinger: Here too.

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,179 @@
...
+        let (order_id, dlc) = match msg {
+            UpdatedContract { order_id, dlc } => (order_id, dlc),
+            Rejected { .. } => {
+                return Ok(());
+            }
+            Failed { order_id, error } => {
+                tracing::warn!(%order_id, "Rollover failed: {:#}", error);
+                return Ok(());
+            }
+            CannotRollover { order_id, reason } => {
```
thomaseizinger: Cannot is a bit of a weird naming as it implies that it should have succeeded although there are plenty of legitimate reasons for not rolling over.


```suggestion
            NoRollover { order_id, reason } => {
```

thomaseizinger (commented):
Nice work!

A few comments!

daemon/src/lib.rs
da-kami: yes! that's what causes my current problems!

daemon/src/model/cfd.rs
da-kami: This is on purpose. (I even added a comment to explain why...)
The trace here logs on `WARN` and is just there to notice that something is fishy when looking at the logs. 
This is one of those "we should never see this" problems, hence the extra warning trace.
I could treat it different by treating this error variant different on the outside. Not sure that is better.

da-kami:
Independent of merging this PR (can be changed later): 
Should we stick to the name `rollover`? Or shall this be renamed to `NewFundingRound` or similar? 
I think the rollover is actually the funding round, whereas the "funding" would be the interest.

daemon/src/model/cfd.rs
thomaseizinger: > I could treat it different by treating this error variant different on the outside. Not sure that is better.

I think it is better. I don't see why this particular situation is any different to not warrant following the "either log OR return error" philosophy.

If the error is unexpected, then you can still log that on the outside. At the moment, this just duplicates information.

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+#[xtra_productivity]
+impl<O, M> Actor<O, M>
+where
+    M: xtra::Handler<monitor::StartMonitoring>,
+    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+{
+    async fn handle(&mut self, _msg: AutoRollover, ctx: &mut xtra::Context<Self>) -> Result<()> {
+        tracing::trace!("Auto Rollover");
```
thomaseizinger:
```suggestion
        tracing::trace!("Checking all CFDs for rollover eligibilty");
```

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,191 @@
...
+                match reason {
+                    CannotRollover::NoDlc => {
+                        tracing::warn!(%order_id, "Not rolled over: {:#}", reason);
+                    }
+                    CannotRollover::AlreadyExpired
+                    | CannotRollover::WasJustRolledOver
+                    | CannotRollover::WrongState { .. } => {
+                        tracing::debug!(%order_id, "Not rolled over: {:#}", reason);
+                    }
+                }
```
thomaseizinger, line 137: Does this work?

```suggestion
                let log_level = match reason {
                    CannotRollover::NoDlc => {
                        tracing::Level::WARN
                    }
                    CannotRollover::AlreadyExpired
                    | CannotRollover::WasJustRolledOver
                    | CannotRollover::WrongState { .. } => {
                        tracing::Level::DEBUG
                    }
                };
                
                tracing::event!(log_level, %order_id, "Not rolled over: {:#}", reason);
```

thomaseizinger (approved):
Two more comments, otherwise LGTM.

Did we test that this actually works? With merging this PR, we are essentially enabling perpetual CFDs.

Also, can you add a changelog entry please?

daemon/src/auto_rollover.rs
da-kami: no it does not:

```
error[E0435]: attempt to use a non-constant value in a constant
   --> daemon/src/auto_rollover.rs:139:33
    |
139 |                 tracing::event!(log_level, %order_id, "Not rolled over: {:#}", reason);
    |                 ----------------^^^^^^^^^---------------------------------------------
    |                 |               |
    |                 |               non-constant value
    |                 help: consider using `let` instead of `static`: `let CALLSITE`
```

don't really think it is important enough to invest time into this.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Off-chain perpetual CFD rollover. Hourly, the taker sends a rollover request to the maker for each open position. The maker can accept or reject a rollover proposal.
+  Upon acceptance the taker and maker collaboratively agree on an oracle price event further in the future and generate new payout transactions accordingly.
+  The previous payout transactions are invalidated. The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
+  In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
```
thomaseizinger:
```suggestion
- Off-chain perpetual CFDs.
  Hourly, the taker sends a rollover request to the maker for each open position.
  The maker can accept or reject a rollover proposal.
  Upon acceptance the taker and maker collaboratively agree on an oracle price event further in the future and generate new payout transactions accordingly.
  The previous payout transactions are invalidated.
  The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
  In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
```

One line per sentence please.

thomaseizinger (approved):
I think this is good to go. We will see in our testnet tests whether or not it works :D

thomaseizinger:
I'll rebase and merge this some time this week.

da-kami:
Some notes after some testing:

There was a problem of `cleanup` being triggered too fast (with 10 seconds it is almost always triggered). Since rollover is actually a long running (yet ephemeral) message exchange, we need to set the timeout higher. 
I thus introduced `MAX_ROLLOVER_DURATION` that is set to 4 minutes. The auto-rollover interval is then set to `MAX_ROLLOVER_DURATION + 1min`. 
After running some manual test again I can confirm that the logic works now. The rollover is triggered after `MAX_ROLLOVER_DURATION + 1min` (i.e. 5 mins) and only CFDs that were not rolled over within the last hour are rolled over (i.e. a rollover proposal is sent to the maker). If the rollover actor runs longer than 4 minutes it will send a `Failed` message ending the rollover.

---

Note that this solution is not *exactly* what we want to have:

Auto-rollover is currently **not** triggered upon startup, but only after the first intertval of `MAX_ROLLOVER_DURATION + 1min` (i.e. 5 mins) is over. 
While we could just trigger an initial auto-rollover by sending the message, this is actually not good enough, because we have to ensure we are connected to a maker first. If we don't ensure to send this after a connection was established I think it would always fail in case establishing a connection takes longer then triggering rollover...
Now, reporting the connection status to the auto-rollover actor is something that we already started discussing when `cleanup` was implemented internally. Probably we should actually re-design auto-rollover to be more aware of the connection status and only trigger auto-rollover upon an active connection. The two `AutoRollover` triggers would then be:

1. Upon establishing a connection to a maker
2. Upon auto-rollover interval *if* we have an active connection to a maker

This PR is not necessarily blocked on this issue, but we have to be aware that rollover is not triggerd (directly) on startup at the moment.

da-kami:
Rebased 😘

thomaseizinger:
> There was a problem of `cleanup` being triggered too fast (with 10 seconds it is almost always triggered). Since rollover is actually a long running (yet ephemeral) message exchange, we need to set the timeout higher.
> I thus introduced `MAX_ROLLOVER_DURATION` that is set to 4 minutes. The auto-rollover interval is then set to `MAX_ROLLOVER_DURATION + 1min`.
> After running some manual test again I can confirm that the logic works now. The rollover is triggered after `MAX_ROLLOVER_DURATION + 1min` (i.e. 5 mins) and only CFDs that were not rolled over within the last hour are rolled over (i.e. a rollover proposal is sent to the maker). If the rollover actor runs longer than 4 minutes it will send a `Failed` message ending the rollover.

Another solution would be to not cancel right away within `cleanup` but just send a message to ourselves that X amount of time has passed and then we can evaluate whether or not the protocol has progressed (did the maker answer?) or not and trigger the potential cleanup from in there. Then this timeout is independent of how long the entire rollover protocol takes.



> Auto-rollover is currently **not** triggered upon startup, but only after the first intertval of `MAX_ROLLOVER_DURATION + 1min` (i.e. 5 mins) is over. While we could just trigger an initial auto-rollover by sending the message, this is actually not good enough, because we have to ensure we are connected to a maker first. If we don't ensure to send this after a connection was established I think it would always fail in case establishing a connection takes longer then triggering rollover... Now, reporting the connection status to the auto-rollover actor is something that we already started discussing when `cleanup` was implemented internally. Probably we should actually re-design auto-rollover to be more aware of the connection status and only trigger auto-rollover upon an active connection. The two `AutoRollover` triggers would then be:
> 
>     1. Upon establishing a connection to a maker
> 
>     2. Upon auto-rollover interval _if_ we have an active connection to a maker
> 
> 
> This PR is not necessarily blocked on this issue, but we have to be aware that rollover is not triggerd (directly) on startup at the moment.

Is this actually a big issue? For our current userbase (people on umbrel) we can expect that this daemon is running the entire time. We trigger this every 5 minutes so if the connection to the maker is unavailable, we will try again in another 5 minutes. I am not sure adding additional triggers here is worth the complexity. It is not like this is super time-sensitive. We have almost 24h to trigger a rollover.

thomaseizinger:
> I am not sure adding additional triggers here is worth the complexity. It is not like this is super time-sensitive. We have almost 24h to trigger a rollover.

That being said, it would be pretty easy to trigger `AutoRollover` from other actors, like on startup or from the connection actor.

thomaseizinger:
> That being said, it would be pretty easy to trigger `AutoRollover` from other actors, like on startup or from the connection actor.

Given that we connect on startup, the only real additional trigger is probably establishing a connection to the maker. The time-based one is independent. I don't think I would bother to restrict that to only when we have a connection.

da-kami:
> 

Ok with me, let's see how it proves itself once in production :)
Feel free to squash and merge

daemon/src/auto_rollover.rs
```diff
@@ -0,0 +1,194 @@
...
+    }
+}
+
+#[xtra_productivity]
+impl<O, M> Actor<O, M>
+where
+    M: xtra::Handler<monitor::StartMonitoring>,
+    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+{
+    async fn handle(&mut self, _msg: AutoRollover, ctx: &mut xtra::Context<Self>) -> Result<()> {
```
klochowicz, line 69: nit: this would be much more readable if the message was a verb, e.g. `StartAutoRollover`.

klochowicz (commented):
It's certainly a much-wanted feature, but I'm not sold on this implementation. The size of the auto-rollover actor constructor indicates to me that something's not right with our module boundaries - it has the same arguments as the cfd actor pretty much.... 
Let's hope event sourcing will fix it - and I hope we'll get a test for it before refactoring :)

daemon/src/auto_rollover.rs
thomaseizinger: Agreed, feel free to send a follow-up :D

PR #887, opened 2021-12-14
Resilient broadcast 2

### LINKED ISSUE #873: Be more resilient when broadcasting transactions

rishflab:


daemon/src/wallet.rs
```diff
@@ -195,17 +196,29 @@ impl Actor {
...
+            if rpc_error.code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain) {
                 tracing::trace!(
                     %txid, "Attempted to broadcast transaction that was already on-chain",
                 );
 
                 return Ok(txid);
             }
+
+            // We do this check because electrum sometimes returns bad
+            if rpc_error.code == i64::from(RpcErrorCode::RpcVerifyAlreadyInChain)
```
thomaseizinger:
```suggestion
            if rpc_error.code == i64::from(RpcErrorCode::RpcVerifyError)
```

Seems to be a typo.

thomaseizinger (approved):

thomaseizinger (approved):
LGTM.

PR #891, opened 2021-12-14
Use dprint's SQL plugin to format SQL files

thomaseizinger:
Thoughts? dprint all the things!

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -8,13 +8,14 @@ use crate::model::Leverage;
 use crate::model::Position;
 use crate::model::Price;
 use crate::model::Usd;
+use anyhow::Context;
 use anyhow::Result;
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
 use time::Duration;
 
-pub async fn run_migrations(pool: &SqlitePool) -> anyhow::Result<()> {
+pub async fn run_migrations(pool: &SqlitePool) -> Result<()> {
     sqlx::migrate!("./migrations")
         .run(pool)
         .await
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -15,7 +15,11 @@ use sqlx::SqlitePool;
 use time::Duration;
 
 pub async fn run_migrations(pool: &SqlitePool) -> anyhow::Result<()> {
-    sqlx::migrate!("./migrations").run(pool).await?;
+    sqlx::migrate!("./migrations")
+        .run(pool)
+        .await
+        .context("Failed to run migrations")?;
+
     Ok(())
 }
 
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -11,11 +11,37 @@ use crate::model::Usd;
 use anyhow::Context;
 use anyhow::Result;
 use sqlx::pool::PoolConnection;
+use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
+use std::path::PathBuf;
 use time::Duration;
 
-pub async fn run_migrations(pool: &SqlitePool) -> Result<()> {
+pub async fn connect(path: PathBuf) -> Result<SqlitePool> {
+    let pool = SqlitePool::connect_with(
+        SqliteConnectOptions::new()
+            .create_if_missing(true)
+            .filename(path),
+    )
+    .await?;
+
+    run_migrations(&pool).await?;
+
+    Ok(pool)
+}
+
+pub async fn memory() -> Result<SqlitePool> {
+    // Note: Every :memory: database is distinct from every other. So, opening two database
+    // connections each with the filename ":memory:" will create two independent in-memory
+    // databases. see: https://www.sqlite.org/inmemorydb.html
+    let pool = SqlitePool::connect(":memory:").await?;
+
+    run_migrations(&pool).await?;
+
+    Ok(pool)
+}
+
+async fn run_migrations(pool: &SqlitePool) -> Result<()> {
     sqlx::migrate!("./migrations")
         .run(pool)
         .await
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -10,6 +10,9 @@ use crate::model::Price;
 use crate::model::Usd;
 use anyhow::Context;
 use anyhow::Result;
+use futures::future::BoxFuture;
+use futures::FutureExt;
+use sqlx::migrate::MigrateError;
 use sqlx::pool::PoolConnection;
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::Sqlite;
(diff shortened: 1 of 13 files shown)


luckysori (approved):

bonomat (approved):

thomaseizinger:
Need to test whether or not sqlx uses a hash to check migrations in existing databases. Let's hope not otherwise this will break stuff!

thomaseizinger:
Ah yeah, that is what I was thinking:

```
Error: Db migrations failed

Caused by:
    migration 20210903050345 was previously applied but has been modified
```

thomaseizinger:
I think we have 3 options here:

1. Don't use an SQL formatter, closing this PR.
2. Exclude existing migrations in dprint's config, only formatting future migrations.
3. Include a snippet to completely wipe the DB with the next release so the migrations can be reapplied.

Thoughts?

klochowicz:
> 

I'd vote for 3, because there's no users other than us yet and we could release before umbrel gets merged! 


a different question: can we somehow tell dprint to try to format sql snippets in our rust files?

luckysori:
3 is the least clunky imo.

thomaseizinger:
> a different question: can we somehow tell dprint to try to format sql snippets in our rust files?

We can use https://docs.rs/sqlx/0.5.9/sqlx/macro.query_file.html instead! Now that we are no longer building on windows, the [line-ending encoding issues](https://github.com/launchbadge/sqlx/issues/1494) might no longer affect us!

thomaseizinger:
> > a different question: can we somehow tell dprint to try to format sql snippets in our rust files?
> 
> We can use [docs.rs/sqlx/0.5.9/sqlx/macro.query_file.html](https://docs.rs/sqlx/0.5.9/sqlx/macro.query_file.html) instead! Now that we are no longer building on windows, the [line-ending encoding issues](https://github.com/launchbadge/sqlx/issues/1494) might no longer affect us!

Actually, just reading the issue: Using `.gitattributes`, we would have been able to work around that!

thomaseizinger:
@bonomat Would (3) be acceptable?

bonomat:
> @bonomat Would (3) be acceptable?

If the next release includes the event-sourcing refactoring we will need to do this anyways right?

thomaseizinger:
> > @bonomat Would (3) be acceptable?
> 
> If the next release includes the event-sourcing refactoring we will need to do this anyways right?

From a user-facing perspective yes. The technical implementation is different (dropping the tables vs nuking the DB file).

thomaseizinger:
The easiest implementation would probably be to switch to a new name for the DB file, otherwise we have to build all kinds of trickery around detecting whether or not we have already deleted it in a prior run of the same release.

bonomat:
> The easiest implementation would probably be to switch to a new name for the DB file, otherwise we have to build all kinds of trickery around detecting whether or not we have already deleted it in a prior run of the same release.

I think eitherway is fine. In general, the sooner we have a stable db the better. Right now there are not many users and it's easier to break things...

thomaseizinger:
@luckysori @bonomat @klochowicz thoughts on 375ff65?

klochowicz (commented):
this solves one problem but creates another one - with this change we'd have a new db every release, even when there's no db changes. from user's pov it's almost the same an nuking the db. the fact that it's stored somewhere on disk or inside container it's hard to notice. 

Perhaps we could use a custom env variable that only gets bumped when we have a breaking db change?

thomaseizinger:
> this solves one problem but creates another one - with this change we'd have a new db every release, even when there's no db changes. from user's pov it's almost the same an nuking the db. the fact that it's stored somewhere on disk or inside container it's hard to notice.
> 
> Perhaps we could use a custom env variable that only gets bumped when we have a breaking db change?

Well, I think by the looks of it, we will have breaking changes in the DB on every release anyway. I don't intend this to be a long term solution but it would do the trick for now?

To get around this particular problem, we could rename the DB if the migrations fail?

klochowicz:
> > this solves one problem but creates another one - with this change we'd have a new db every release, even when there's no db changes. from user's pov it's almost the same an nuking the db. the fact that it's stored somewhere on disk or inside container it's hard to notice.
> > Perhaps we could use a custom env variable that only gets bumped when we have a breaking db change?
> 
> Well, I think by the looks of it, we will have breaking changes in the DB on every release anyway. I don't intend this to be a long term solution but it would do the trick for now?
> 
> To get around this particular problem, we could rename the DB if the migrations fail?

That is probably the best solution as it only is applied when we have to. Maybe we could rename it to `maker_{$TIMESTAMP}.sqlite` in case of failed migration, and tell user that they can still use the older version to access that info?

thomaseizinger:
> > > this solves one problem but creates another one - with this change we'd have a new db every release, even when there's no db changes. from user's pov it's almost the same an nuking the db. the fact that it's stored somewhere on disk or inside container it's hard to notice.
> > > Perhaps we could use a custom env variable that only gets bumped when we have a breaking db change?
> > 
> > 
> > Well, I think by the looks of it, we will have breaking changes in the DB on every release anyway. I don't intend this to be a long term solution but it would do the trick for now?
> > To get around this particular problem, we could rename the DB if the migrations fail?
> 
> That is probably the best solution as it only is applied when we have to. Maybe we could rename it to `maker_{$TIMESTAMP}.sqlite` in case of failed migration, and tell user that they can still use the older version to access that info?

Yeah I am liking this solution as well. It bring 0 disruption for new users that don't have an existing DB and the restart-on-failure from docker compose means we will restart immediately and be in a good state then. This would also allow us to clean up our migration scripts at some point (if we want to).

thomaseizinger:
> > > > this solves one problem but creates another one - with this change we'd have a new db every release, even when there's no db changes. from user's pov it's almost the same an nuking the db. the fact that it's stored somewhere on disk or inside container it's hard to notice.
> > > > Perhaps we could use a custom env variable that only gets bumped when we have a breaking db change?
> > > 
> > > 
> > > Well, I think by the looks of it, we will have breaking changes in the DB on every release anyway. I don't intend this to be a long term solution but it would do the trick for now?
> > > To get around this particular problem, we could rename the DB if the migrations fail?
> > 
> > 
> > That is probably the best solution as it only is applied when we have to. Maybe we could rename it to `maker_{$TIMESTAMP}.sqlite` in case of failed migration, and tell user that they can still use the older version to access that info?
> 
> Yeah I am liking this solution as well. It bring 0 disruption for new users that don't have an existing DB and the restart-on-failure from docker compose means we will restart immediately and be in a good state then. This would also allow us to clean up our migration scripts at some point (if we want to).

This is the log now:

```
2021-12-20 17:08:47  INFO daemon::logger: Initialized logger
2021-12-20 17:08:47  INFO maker: Running version: 0.3.1-139-gd21173b
2021-12-20 17:08:47  INFO maker: CFDs created with this release will settle after 24 hours
2021-12-20 17:08:49  INFO maker: Listening on 0.0.0.0:9999
2021-12-20 17:08:49 ERROR daemon::db: Failed to run migrations: migration 20210903050345 was previously applied but has been modified
2021-12-20 17:08:49  INFO daemon::db: Backing up old database at /home/thomas/src/github.com/comit-network/hermes/testnet/maker.sqlite to /home/thomas/src/github.com/comit-network/hermes/testnet/maker.sqlite-1639980529-backup
2021-12-20 17:08:51 DEBUG daemon::oracle: Fetching announcement for /x/BitMEX/BXBT/2021-12-20T14:00:00.price?n=20
2021-12-20 17:08:51 DEBUG daemon::oracle: Fetching announcement for /x/BitMEX/BXBT/2021-12-20T17:00:00.price?n=20
2021-12-20 17:08:51 DEBUG daemon::oracle: Fetching announcement for /x/BitMEX/BXBT/2021-12-20T10:00:00.price?n=20
2021-12-20 17:08:51 DEBUG daemon::oracle: Fetching announcement for /x/BitMEX/BXBT/2021-12-20T18:00:00.price?n=20
```

I decided to not fail but just re-connect after the database has been moved so it also works if you are not running in a supervisor!

da-kami (approved):
Read through the decision making process. ack with creating a backup if the migration fails and then just start from scratch. We obviously have to be mindful about this in the future.

daemon/src/db.rs
```diff
@@ -11,17 +13,45 @@ use std::mem;
...
+            );
+
+            tokio::fs::rename(&path, &new_path)
+                .await
+                .context("Failed to rename database file")?;
+
+            tracing::info!("Starting with a new database!");
+
+            // recurse to reconnect (async recursion requires a `BoxFuture`)
+            return connect(path).await;
```
luckysori, line 70: 🤔 So the assumption is that `run_migrations` cannot fail if the database has just been created, right? Otherwise I can see us creating an endless loop here.

daemon/src/db.rs
```diff
@@ -49,6 +49,8 @@ pub fn connect(path: PathBuf) -> BoxFuture<'static, Result<SqlitePool>> {
             return connect(path).await;
         }
 
+        tracing::info!("Opened database at {}", path.display());
```
luckysori: 🔧 Can probably squash this with the previous commit, unless there was a reason for it.

luckysori (approved):
Thank you for taking the time to figure this out and solve it cleanly. Inspiring PR :D

daemon/tests/harness/mod.rs
```diff
@@ -365,17 +364,6 @@ macro_rules! deliver_event {
     };
 }
 
-async fn in_memory_db() -> SqlitePool {
-    // Note: Every :memory: database is distinct from every other. So, opening two database
-    // connections each with the filename ":memory:" will create two independent in-memory
-    // databases. see: https://www.sqlite.org/inmemorydb.html
```
klochowicz, line 378: nit: any reason for getting rid of this comment? I for one found it useful (and it doesn't look like it can/will rot)

klochowicz (approved):
nice to see SQL files formatted 👍 approval pending the potential endless loop issue raised by @luckysori

daemon/tests/harness/mod.rs
thomaseizinger: No reason, just got lost in the refactor!

daemon/src/db.rs
thomaseizinger: Yeah, these are the possible errors: https://docs.rs/sqlx-core/0.5.5/src/sqlx_core/migrate/error.rs.html#3

I guess there are some in here that are not resolved by using a new DB? We should probably only try again on `VersionMismatch` and `VersionMissing`. All others seem permanent!

luckysori (approved):
LGTM, thanks!

PR #899, opened 2021-12-15
Rename database function for loading CFDs

thomaseizinger:
With the removal of the order from the DB, we are loading a CFD by
its ID which just happens to be the order id but that is no longer
relevant at this stage.


luckysori (approved):

klochowicz (approved):
I already updated that on the event-sourcing branch as well, so it should rebase without a problem against this one :)

PR #906, opened 2021-12-15
Remove workaround for longer heartbeat interval in debug mode

klochowicz:
Before contract setup was split into dedicated actors on each side, heartbeats
were not being delivered correctly and the taker would close the connection
half-way through contract setup.
This was especially noticeable in debug mode, as the contract setup could take even
~60 seconds to complete. A workaround was set in place (45s for heartbeat
interval, making it drop connection only after 90s, longer than contract setup).

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -43,6 +43,9 @@ const CONNECT_TO_MAKER_INTERVAL: Duration = Duration::from_secs(5);
 enum State {
     Connected {
         last_heartbeat: SystemTime,
+        /// Last pulse measurement time. Used for checking whether measuring
+        /// task is not lagging too much.
+        last_pulse: SystemTime,
         write: wire::Write<wire::MakerToTaker, wire::TakerToMaker>,
         _tasks: Tasks,
     },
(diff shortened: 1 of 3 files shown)


luckysori (approved):
Noice.

klochowicz:
of course it passed the tests on my machine before posting, but seems to have failed on CI :( if the workaround's needed, the issue is still there :(

thomaseizinger (approved):

thomaseizinger:
Perhaps this is due to our tests only using a single-threaded tokio runtime. If one task is busy (verifying signatures), it might be that the other tasks never get to run! We could experiment with using https://docs.rs/tokio/1.15.0/tokio/task/fn.spawn_blocking.html to do the verification and signing?

klochowicz:
@thomaseizinger @luckysori @da-kami  I have come up with a different, IMHO less hacky workaround - instead of trying to have crazy long times for heartbeats in debug mode, I'd like to propose not acting on a "false readings" (ie. when our measurements fell behind due to some compute-heavy task blocking the thread). Please let me know what you think.

@thomaseizinger I also tried using `spawn_blocking()` and multi-threaded tokio runtime in tests, none of which produced promised results :( happy to look into that someday again, but I hope this one will do the trick in a least hacky way.

luckysori (approved):
Very interesting problem. Good job! I would say we can merge this and see if it solves it.

PR #908, opened 2021-12-16
Recover wallet from 256bit string through REST endpoint

### LINKED ISSUE #865: Seedphrase based wallet that can easily be backed up

rishflab:
The choice of using a 256 bit string is temporary. This will be replaced
in favour of a 24 word seed phrase which will be used to create the
seed (BIP39).

This PR was exploratory. I will merge it after #909.

will squash after approval

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -1,5 +1,6 @@
 use crate::model::Timestamp;
 use crate::model::WalletInfo;
+use crate::tokio_ext::spawn_fallible;
 use crate::Tasks;
 use anyhow::bail;
 use anyhow::Context;
@@ -40,6 +41,7 @@ pub struct Actor {
     used_utxos: HashSet<OutPoint>,
     tasks: Tasks,
     sender: watch::Sender<Option<WalletInfo>>,
+    electrum_rpc_url: String,
 }
 
 #[derive(thiserror::Error, Debug, Clone, Copy)]
@@ -70,6 +72,7 @@ impl Actor {
             tasks: Tasks::default(),
             sender,
             used_utxos: HashSet::default(),
+            electrum_rpc_url: electrum_rpc_url.to_string(),
         };
 
         Ok((actor, receiver))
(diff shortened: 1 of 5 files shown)


daemon/src/routes_maker.rs
```diff
@@ -186,6 +187,31 @@ pub async fn post_cfd_action(
     Ok(status::Accepted(None))
 }
 
+#[derive(Debug, Clone, Deserialize)]
+pub struct WalletRecoverRequest {
+    pub seed_words: String,
+}
+
+#[rocket::post("/wallet/recover", data = "<wallet_recover_request>")]
+pub async fn post_wallet_recover(
```
thomaseizinger: Don't need this for the maker :)

daemon/src/wallet.rs
```diff
@@ -131,6 +135,36 @@ impl Actor {
 
 #[xtra_productivity]
 impl Actor {
+    pub fn handle_recover(&mut self, msg: Recover, ctx: &mut xtra::Context<Self>) -> Result<()> {
+        let client = bdk::electrum_client::Client::new(&self.electrum_rpc_url)
+            .context("Failed to initialize Electrum RPC client")?;
+
+        let seed = Seed::from_seed_words(&msg.seed_words)?;
+
+        let ext_priv_key = seed.derive_extended_priv_key(self.wallet.network())?;
```
thomaseizinger: This is not ideal. I think for now, we should be creating the `ExtendedPrivKey` directly from the seed words, not roundtripping through `Seed` because that also derives the network identity. Let's keep this change isolated to the wallet for now otherwise you would have to re-initialize the connection here as well because on next startup (if you were to use the new seed) you would suddenly have a different identity.

daemon/src/wallet.rs
```diff
@@ -304,6 +338,10 @@ pub struct BuildPartyParams {
 /// Private message to trigger a sync.
 struct Sync;
 
+pub struct Recover {
```
thomaseizinger:
```suggestion
pub struct Reinitialize {
```

Recover is too usecase specific IMO. From the wallet perspective, we are just re-initializing with a different seed.

daemon/src/wallet.rs
```diff
@@ -40,6 +42,7 @@ pub struct Actor {
     used_utxos: HashSet<OutPoint>,
```
thomaseizinger, line 41: You need to clear these upon re-init for your implementation to be correct.

(Side note: If we would restart the entire actor, this would be automatically handled. Managing state is hard :upside_down_face: )

daemon/src/seed.rs
```diff
@@ -12,6 +12,11 @@ use std::path::Path;
 pub struct Seed([u8; 256]);
 
 impl Seed {
+    pub fn from_seed_words(words: &str) -> Result<Self> {
+        let bytes: [u8; 256] = words.as_bytes().try_into()?;
+        Ok(Self(bytes))
+    }
```
thomaseizinger: Don't do this for now because it will also influence the network identity.

daemon/src/routes_maker.rs
```diff
@@ -186,6 +187,31 @@ pub async fn post_cfd_action(
...
+    wallet_recover_request: Json<WalletRecoverRequest>,
+    wallet_actor: &State<Wallet>,
```
thomaseizinger: Instead of doing this, I would prefer if we go towards https://github.com/itchysats/itchysats/issues/774.

thomaseizinger (dismissed):
Thanks!

Couple of comments.

daemon/src/routes_maker.rs
rishflab: I will make a new PR to address #774 and rebase this onto it

thomaseizinger (approved):

thomaseizinger:
Actually, can you add a changelog entry please?

PR #909, opened 2021-12-16
Hide internal actors from rocket state

### LINKED ISSUE #774: Too many details about the actor system leak into rocket

rishflab:
Partially addresses #774 

will squash after review

### DIFF
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -278,23 +278,7 @@ async fn main() -> Result<()> {
         projection::Actor::new(db.clone(), Role::Maker, bitcoin_network).await?;
     tasks.add(projection_context.run(proj_actor));
 
-    let listener_stream = futures::stream::poll_fn(move |ctx| {
-        let message = match futures::ready!(listener.poll_accept(ctx)) {
-            Ok((stream, address)) => {
-                maker_inc_connections::ListenerMessage::NewConnection { stream, address }
-            }
-            Err(e) => maker_inc_connections::ListenerMessage::Error { source: e },
-        };
-
-        Poll::Ready(Some(message))
-    });
-
-    tasks.add(
-        maker_actor_system
-            .inc_conn_addr
-            .clone()
-            .attach_stream(listener_stream),
-    );
+    maker.listen_on(listener);
 
     rocket::custom(figment)
         .manage(projection_feeds)
--- a/daemon/src/maker.rs
+++ b/daemon/src/maker.rs
@@ -30,7 +30,6 @@ use sqlx::SqlitePool;
 use std::net::SocketAddr;
 use std::path::PathBuf;
 use std::str::FromStr;
-use std::task::Poll;
 use tracing_subscriber::filter::LevelFilter;
 use xtra::Actor;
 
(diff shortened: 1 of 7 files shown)


daemon/src/routes_maker.rs
```diff
@@ -152,27 +146,25 @@ pub struct PromptAuthentication {
 pub async fn post_cfd_action(
     id: OrderId,
     action: CfdAction,
-    cfd_actor: &State<Maker>,
+    maker_actor_system: &State<Maker>,
```
thomaseizinger: Nit: I think naming this `maker` would be expressive enough.

```suggestion
    maker: &State<Maker>,
```

daemon/src/maker.rs
```diff
@@ -294,15 +290,19 @@ async fn main() -> Result<()> {
         Poll::Ready(Some(message))
     });
 
-    tasks.add(incoming_connection_addr.attach_stream(listener_stream));
+    tasks.add(
+        maker_actor_system
+            .inc_conn_addr
+            .clone()
+            .attach_stream(listener_stream),
```
thomaseizinger:
```suggestion
        maker_actor_system.listen_on(listener)
```

I think we could hide a few more details here, that should allow us to make all fields of `MakerActorSystem` private.

thomaseizinger (approved):
Thank you! Some ideas but good to see this idea play out as envisioned!

thomaseizinger (approved):
Thanks! Nice work!

Feel free to squash everything into one commit.

rishflab:
I think I have broken it:

```

    pub fn listen_on(&mut self, listener: TcpListener) {
        let listener_stream = futures::stream::poll_fn(move |ctx| {
            let message = match futures::ready!(listener.poll_accept(ctx)) {
                Ok((stream, address)) => {
                    maker_inc_connections::ListenerMessage::NewConnection { stream, address }
                }
                Err(e) => maker_inc_connections::ListenerMessage::Error { source: e },
            };

            Poll::Ready(Some(message))
        });

        self.tasks
            .add(self.inc_conn_addr.clone().attach_stream(listener_stream));
            ```

Issue #244, opened 2021-10-07
Funding rate


da-kami:
The maker has to be able to define interest on the CFD, i.e. a percentage of the taker's amount that will always go to the maker regardless of the price change.

scratchscratchscratchy:
Besides the fact the this will require changes on the protocol, what would be the trigger for this in the current program flow?

How does collecting interest (i.e. deduct the interest from re-setting up the CFD) work together with auto-rollover #456 and automatic closing #483 the CFD?

Here's what's in my head:

* auto-rollover without amount changes every hour, until we reach the 23rd hour of rollover.
* once in the 23rd hour of rollover the roll-over proposal should take the interest into account
* Both taker and maker have to validate that the amounts exchanged make sense according to what kind of roll-over (without interest or with interest) is happening.

This requires that we distinguish between initial creation and roll-over in the database.

Trigger for auto-close would be: If no roll over happens until attestation then close.

Note: The 23rd hour more generally means: `settlement_interval - 1 hour`, since we use `24` h as settlement interval I used 23h here. 

@itchymax what do you think?

1010Tom:
In my head this is relatively straight forward. Think of the scenario like this: 

- Alice goes long and puts up 0.01 BTC
- Bob goes short and puts up 0.01 BTC

Alice should pay Bob in a bull market , i.e. long pays short.

when creating the set of CETs we need to include the interest rate already, let's be expansive and say it's 10%/day. 
We construct the CETs as follow: 

```
{
	alice_out: AMOUNT_ALICE,
	bob_out: AMOUNT_BOB
	bob_interest: INTEREST
}
```

So for the initial CETs with an expiry of 24h the payout would look like this: 

```
{
	alice_out: AMOUNT_ALICE_n,
	bob_out: AMOUNT_BOB
	bob_interest: 0.001
}
```

where `AMOUNT_ALICE_n = AMOUNT_ALICE-0.001`.

We can feed `AMOUNT_ALICE_n` and `AMOUNT_BOB` directly into our payout curve calculation as before. 
`AMOUNT_ALICE` and `AMOUNT_BOB` both depend on the payout curve.

After 1h we rollover and we should update the fee accordingly: 
i.e. the new CET 

```
{
	alice_out: AMOUNT_ALICE_n2,
	bob_out: AMOUNT_BOB
	bob_interest: 0.001+(AMOUNT_ALICE_n2-AMOUNT_BOB/10%/24)
}
```

[...]

PR #911, opened 2021-12-16
Introduce funding fee in payout calculations

luckysori:
Work towards #244.

### DIFF
--- a/daemon/src/payout_curve.rs
+++ b/daemon/src/payout_curve.rs
@@ -53,7 +53,7 @@ pub fn calculate(
     leverage: Leverage,
     n_payouts: usize,
 ) -> Result<Vec<Payout>> {
-    let payouts = calculate_payout_parameters(price, quantity, leverage, n_payouts)?
+    let payouts = calculate_payout_parameters(price, quantity, leverage, n_payouts, 0)?
         .into_iter()
         .map(PayoutParameter::into_payouts)
         .flatten_ok()
@@ -74,6 +74,7 @@ fn calculate_payout_parameters(
     quantity: Usd,
     long_leverage: Leverage,
     n_payouts: usize,
+    funding_fee: u64,
 ) -> Result<Vec<PayoutParameter>> {
     let initial_rate = price
         .try_into_f64()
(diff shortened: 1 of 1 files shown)


daemon/src/payout_curve.rs
```diff
@@ -511,19 +512,53 @@ mod tests {
...
+                    left_bound,
+                    right_bound,
+                    long_amount: long_amount.wrapping_sub(funding_fee),
+                    short_amount: short_amount + funding_fee,
+                },
+            )
+            .collect::<Vec<_>>();
+
+        pretty_assertions::assert_eq!(actual_payouts, adjusted_payouts);
+    }
```
thomaseizinger: Is there any way we can assert this differently? You are basically reimplementing what you did above, that is not exactly the best assertion :D 

What about checking:

- That the sum for each payout is the same as without the fee
- That the number of payouts didn't change if you use a fee
- That the short gets the fee on every payout (map both iterators to just the short amount, zip them, subtract one from the other, assert that it all items are equal to the funding fee)

Combination of 1 and 3 should give us good confidence.

thomaseizinger (approved):

daemon/src/payout_curve.rs
luckysori: I agree that this assertion is poor. I'll work on it and I'll try to incorporate your ideas.

daemon/src/payout_curve.rs
```diff
@@ -721,13 +727,62 @@ mod tests {
...
+                .iter()
+                .zip(payouts_with_fee.clone())
+                .all(|(payout, payout_with_fee)| {
+                    let total = payout_with_fee.short_amount + payout_with_fee.long_amount;
+                    let space_for_fee = total - payout_with_fee.short_amount;
+                    let expected_fee = funding_fee.min(space_for_fee);
+                    let fee = payout_with_fee.short_amount - payout.short_amount;
+
+                    fee == expected_fee
+                });
```
thomaseizinger:
```suggestion
        let actual_fees =
            payouts
                .iter()
                .zip(payouts_with_fee.clone())
                .map(|(payout, payout_with_fee)| payout_with_fee.short_amount - payout.short_amount);
                
        assert!(actual_fees.all(|fee| fee == 200));
```

Best to split the assertion out and assert against something static!

daemon/src/payout_curve.rs
```diff
@@ -721,13 +727,62 @@ mod tests {
...
+        .unwrap();
+
+        let sum_of_payouts_is_constant =
+            payouts
+                .iter()
+                .zip(payouts_with_fee.clone())
+                .all(|(payout, payout_with_fee)| {
+                    dbg!(payout.short_amount) + dbg!(payout.long_amount)
+                        == dbg!(payout_with_fee.short_amount) + dbg!(payout_with_fee.long_amount)
+                });
```
thomaseizinger: Same here, I think it would be nice to map both to the same and then do a separate step where we assert that they are equal.

thomaseizinger: Could even do that by collecting them both into `Vec` and use `assert_eq`.

thomaseizinger (approved):
Some ideas!

daemon/src/payout_curve.rs
luckysori: This does look better but it won't pass for the payout which pays the short all of the money. I had to introduce the `expected_fee` to account for ranges which already pay the short position so much that there isn't enough money left from the total to cover the entire fee.

luckysori: > This does look better but it won't pass for the payout which pays the short all of the money.

We could treat it as a special case, but the same problem could happen for any payout which would originally pay the long position less than the funding fee.

I've been thinking about it and I think it's gonna be hard to have a nice

```rust
assert!(fees.all(|fee| fee == funding_fee));
```

like you propose.

thomaseizinger: Can we assert against an array of fees then at least? Then we can add comments to the special cases?

thomaseizinger (approved):

PR #917, opened 2021-12-17
Make oracle and monitor actor load CFDs from DB

thomaseizinger:
With event sourcing, these actors will use a different model. Passing
them the DB connection directly now makes this change easier.


klochowicz (approved):
I also like that we don't need to create them with the initial values anymore!

luckysori (approved):

PR #923, opened 2021-12-19
Remove next dependency in favor of `skipLibCheck`

thomaseizinger:
See https://github.com/chakra-ui/chakra-ui/issues/5082#issuecomment
-979039787.

### DIFF
--- a/taker-frontend/package.json
+++ b/taker-frontend/package.json
@@ -26,7 +26,6 @@
     "eslint": "^8.5.0",
     "eslint-config-react-app": "^7.0.0",
     "framer-motion": "^5",
-    "next": "^12.0.4",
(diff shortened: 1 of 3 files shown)


klochowicz:
this potentially can impact our type system accuracy according to the docs:
https://www.typescriptlang.org/tsconfig#skipLibCheck

Perhaps migrating to yarn3 could mitigate it, but I'm not sure which trade-off is more worth it here? (extra dependency vs disabling a check that has a slim chance of biting us someday)

thomaseizinger:
Should we leave a comment in the config file to switch it back to `false` when we do the `yarn 3` upgrade?

da-kami (approved):

klochowicz (approved):

PR #924, opened 2021-12-19
Don't manually lock EncryptedJsonCodec

thomaseizinger:
There is a utility in `futures` for splitting anything that implements
`AsyncRead` + `AsyncWrite` into two components. It also uses a lock
interally but at least we don't have to implement it ourselves. In this
case, it also fixes using the wrong mutex (std::sync). A side-effect of
this change is that our codec now needs to have two type parameters
because we only construct it once and we have two different messages for
sending and receiving. The implications of this are minimal though.

Might help with #759.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -26,14 +26,11 @@ use futures::SinkExt;
 use futures::StreamExt;
 use futures::TryStreamExt;
 use std::net::SocketAddr;
-use std::sync::Arc;
-use std::sync::Mutex;
 use std::time::Duration;
 use std::time::SystemTime;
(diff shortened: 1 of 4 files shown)


rishflab (approved):
nice! that mutex was gross

klochowicz (approved):
great find! 👍 

as a side note, shall we completely ban usage of `std::sync::Mutex` as it can block the whole thread instead of just the future?

thomaseizinger:
> as a side note, shall we completely ban usage of `std::sync::Mutex` as it can block the whole thread instead of just the future?

Yes that is a good idea!

PR #927, opened 2021-12-20
Remove `send_to_socket::Actor`

thomaseizinger:
This removes a level of indirection for actually sending messages which
simplifies the overall handling of the connection and unifies how
heartbeats are sent to the taker. This is important because we don't
want heartbeats to be handled in a special way as we use them to detect
issues with the connection.

May help with https://github.com/itchysats/itchysats/issues/759.


daemon/src/maker_inc_connections.rs
```diff
@@ -107,7 +105,26 @@ pub struct Actor {
...
+/// A connection to a taker.
+struct Connection {
+    taker: Identity,
+    write: wire::Write<wire::TakerToMaker, wire::MakerToTaker>,
+    _tasks: Tasks,
+}
+
+impl Connection {
+    async fn send(&mut self, msg: wire::MakerToTaker) -> Result<()> {
+        let msg_str = msg.to_string();
```
bonomat, line 119: Shall we add a trace here to log what message is being send?

thomaseizinger: Addressed in https://github.com/itchysats/itchysats/pull/927/commits/21b20f6fdbff870393cd2f92a9d2caee9bfda50c.

daemon/src/maker_inc_connections.rs
```diff
@@ -118,6 +118,8 @@ impl Connection {
     async fn send(&mut self, msg: wire::MakerToTaker) -> Result<()> {
         let msg_str = msg.to_string();
 
+        tracing::trace!(target = "wire", taker_id = %self.taker, "Sending {}", msg_str);
```
bonomat, line 121: nice, that's useful 👍

bonomat (approved):
Awesome, thanks.

thomaseizinger:
Let's see how it performs on testnet.

thomaseizinger:
Just stress testing this against the testnet maker and works like charm so far.

PR #936, opened 2021-12-20
Update default blockstream electrum url

da-kami:
It appears our URLs are not what is recommended.
See: https://blog.blockstream.com/en-esplora-and-other-alternatives-to-electrumx/

According to https://github.com/comit-network/xmr-btc-swap/pull/866 connection to these servers should be more stable.


thomaseizinger (approved):
Assuming you tested this.

klochowicz (approved):

PR #944, opened 2021-12-21
Bip39 Mnemonic + backup and generate mnemonic endpoints

rishflab:
The seed path being stored in the wallet actor is clunky. I think I should move it to the `TakerActorSystem`.


daemon/tests/harness/mod.rs
```diff
@@ -292,6 +293,7 @@ impl Taker {
             Duration::from_secs(10),
             projection_actor,
             maker_identity,
+            PathBuf::from_str("/dev/null").expect("Valid path string"),
```
rishflab: Not sure what is the best way to deal with this.

daemon/src/lib.rs
```diff
@@ -458,11 +465,25 @@ where
...
+        let wallet_seed = Seed::read_from(&self.wallet_seed_path).await?;
+        let mnemonic = wallet_seed.try_into()?;
+        Ok(mnemonic)
+    }
+
+    pub fn generate_mnenomic(&self) -> Result<Mnemonic> {
+        let mut entropy = [0u8; 256];
+        rand::thread_rng().fill(&mut entropy);
+        Ok(Mnemonic::from_entropy(&entropy)?)
+    }
```
thomaseizinger: This doesn't really use `self` so I don't think it needs to be on here? Also, I think `https://docs.rs/bip39/1.0.1/bip39/struct.Mnemonic.html#method.generate_in_with` might be a more appropriate function to use and we can just call this wherever needed and don't need to encapsulate it?

daemon/src/routes_taker.rs
```diff
@@ -159,27 +161,55 @@ pub async fn post_cfd_action(
 }
 
 #[derive(Debug, Clone, Deserialize)]
-pub struct WalletReinitialiseRequest {
-    pub seed_words: String,
+pub struct WalletRestoreRequest {
+    pub mnemonic: String,
```
thomaseizinger: `Mnemonic` implements `Deserialize`: https://docs.rs/bip39/1.0.1/bip39/struct.Mnemonic.html#impl-Deserialize%3C%27de%3E

Can we use that here?

daemon/src/routes_taker.rs
```diff
@@ -159,27 +161,55 @@ pub async fn post_cfd_action(
...
+pub async fn post_mnemonic(
+    taker: &State<Taker>,
+) -> Result<status::Accepted<String>, HttpApiProblem> {
+    let mnemonic = taker.generate_mnenomic().map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Generate new mnemonic request failed")
+            .detail(e.to_string())
+    })?;
+    Ok(status::Accepted(Option::from(mnemonic.to_string())))
+}
```
thomaseizinger: This interface seems to be unnecessary chatty:

The two functionalities we need are:

1. Restore a wallet from a mnemonic
2. Re-initialize the wallet with a new one

Currently, for (1), we would need to do:

```
POST /wallet/mnemonic
```

And for (2), we would need to do:

```
POST /mnemonic
POST /wallet/mnemonic
```

I don't think we need to split (2) into two requests. What do you think of the following interface:

1. `POST /wallet/restore`

Takes seedwords in the body and re-initializes the wallet with the given ones.

2. `POST /wallet/reinitialize`

Has no request body, generates random seedwords, re-initializes the wallet and returns the seedwords used in the response body.

A design note: Try to design your API around the functionality you want to offer to the client, not so much in the individual steps it takes to get that functionality done. In this case, we have two different things the user may want to achieve. Creating **one** API endpoint per usecase allows us to better capture the users intent and execute the appropriate action. The resulting API is also likely to be more stable because it will only change with those specific requirements.

daemon/src/seed.rs
```diff
@@ -88,3 +89,20 @@ impl Default for Seed {
         Self(seed)
     }
 }
+
+impl TryFrom<Mnemonic> for Seed {
+    type Error = anyhow::Error;
+
+    fn try_from(mnemonic: Mnemonic) -> Result<Self> {
+        let entropy = <[u8; 256]>::try_from(mnemonic.to_entropy().as_slice())?;
+        Ok(Self { 0: entropy })
```
thomaseizinger: Lol, does this compile? I had no idea that is valid syntax for referring to fields in tuples!

daemon/src/wallet.rs
```diff
@@ -134,16 +138,13 @@ impl Actor {
...
         let client = bdk::electrum_client::Client::new(&self.electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
 
-        let ext_priv_key =
-            ExtendedPrivKey::new_master(self.wallet.network(), msg.seed_words.as_bytes())?;
+        let seed = Seed::try_from(msg.mnemonic)?;
+
+        let ext_priv_key = seed.derive_extended_priv_key(self.wallet.network())?;
 
         let db = bdk::database::MemoryDatabase::new();
```
thomaseizinger: We need to store the new seed words on disk here, otherwise they are completely lost!

daemon/src/taker.rs
```diff
@@ -179,10 +179,21 @@ async fn main() -> Result<()> {
...
+        )
+        .await?;
+    }
+
+    let wallet_seed = Seed::initialize(&data_dir.join("taker_wallet_seed")).await?;
 
     let bitcoin_network = opts.network.bitcoin_network();
-    let ext_priv_key = seed.derive_extended_priv_key(bitcoin_network)?;
-    let (_, identity_sk) = seed.derive_identity();
+    let ext_priv_key = wallet_seed.derive_extended_priv_key(bitcoin_network)?;
```
thomaseizinger: I think we should change the interface of `wallet::Actor` to be initialized by a `Mnemonic`. That will clean up this back and forth here.

daemon/src/taker.rs
```diff
@@ -179,10 +179,21 @@ async fn main() -> Result<()> {
...
+    // If the wallet seed does not exist, create one using the seed
+    if !&data_dir.join("taker_wallet_seed").exists() {
+        tokio::fs::copy(
+            &data_dir.join("taker_seed"),
+            &data_dir.join("taker_wallet_seed"),
+        )
+        .await?;
+    }
+
+    let wallet_seed = Seed::initialize(&data_dir.join("taker_wallet_seed")).await?;
```
thomaseizinger: Why are we round-tripping through `Seed`? I think we should store the seed words on disk here!

thomaseizinger (changes requested):
Thanks!

Some comments.

daemon/src/routes_taker.rs
rishflab: > The two functionalities we need are:
> 
>     1. Restore a wallet from a mnemonic
> 
>     2. Re-initialize the wallet with a new one

We also need the backup functionality in case the user wants to store the mnemonic somewhere for later use:
 
3. Backup a wallet by showing the user the mnemonic 

> Takes seedwords in the body and re-initializes the wallet with the given ones.
> 
>     POST /wallet/reinitialize

yep makes sense to package mnemonic generation and restore together

daemon/src/wallet.rs
rishflab: I moved saving to disk to the `TakerActorSystem` where `Restore` is sent. This means the `wallet::Actor` does not need to know about the path of the wallet seed file.

daemon/src/taker.rs
rishflab: I am not sure if there is a huge advantage in storing the seed words instead of the entropy on disk. 

We will have to change the implementation of `Seed` or make a seperate `WalletSeed` type which is almost exactly the same as `Seed`. If we change the implementation of `Seed`, we will have to write some migration code.

daemon/src/lib.rs
rishflab: I opted to do it the same as our seed generation because I did not know which random number generator i should use when calling `generate_in_with`.

thomaseizinger: Same as the one you are using currently, `thread_rng` is good enough :)

daemon/src/routes_taker.rs
thomaseizinger: > We also need the backup functionality in case the user wants to store the mnemonic somewhere for later use:
> 
>     3. Backup a wallet by showing the user the mnemonic

True! But that is just covered by `GET /wallet/mnemonic` which I would probably just change to `GET /wallet/seedwords`.

daemon/src/taker.rs
thomaseizinger: The plain seed words are easier to back up and I would say there is some advantage in being able to do that without having to go through the UI.

The migration code is trivial:

- Make `Mnemonic` from existing seed such that getting key from `Mnemonic` is the same key (btw, we need a test that given an existing `Seed` and making a `Mnemonic` from it doesn't change the generated `ExtendedPrivKey`!)
- Write words to disk

What I am getting at here is that in the end, we don't need `Seed` at all for the wallet. On startup, we should:

- Check if seed words exist, if yes, read them and generate `Mnemonic`
- Pass `Mnemonic` to `wallet::Actor`

daemon/src/wallet.rs
thomaseizinger: Ah right, that makes sense yes. That will make it also easier to solve this via restart the wallet actor if we ever want to.

daemon/src/taker.rs
rishflab: On second thought I think it is worth creating a seperate `WalletSeed`. The current `Seed` is not bip39 compliant.

daemon/src/routes_taker.rs
```diff
@@ -162,25 +161,22 @@ pub async fn post_cfd_action(
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct WalletRestoreRequest {
-    pub mnemonic: String,
+    pub mnemonic: Mnemonic,
 }
```
thomaseizinger: Can we add a test so we can see what we actually need to pass in? Is it space-separated words?

daemon/src/routes_taker.rs
```diff
@@ -195,14 +194,15 @@ pub async fn get_wallet_mnemonic(
...
-                .title("Generate new mnemonic request failed")
-                .detail(e.to_string())
-        })?;
+#[rocket::post("/wallet/reinitialize")]
+pub async fn post_wallet_reinitalize(
+    taker: &State<Taker>,
+) -> Result<status::Accepted<String>, HttpApiProblem> {
+    let mnemonic = taker.reinitialize_wallet().await.map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Generate new mnemonic request failed")
```
thomaseizinger: Stale error message.

daemon/src/routes_taker.rs
```diff
@@ -195,14 +194,15 @@ pub async fn get_wallet_mnemonic(
...
+            .detail(e.to_string())
+    })?;
     Ok(status::Accepted(Option::from(mnemonic.to_string())))
```
thomaseizinger: I know you've just continued the existing pattern here but technically 202 ACCEPTED is not really appropriate here. Same as with HTTP methods, if you don't want to communicate anything specific and it is successful, use 200 OK.

daemon/src/routes_taker.rs
```diff
@@ -195,14 +194,15 @@ pub async fn get_wallet_mnemonic(
...
```
thomaseizinger: Returning the seed words seem unnecessary. That is exactly the information the caller gave us. Just creates confusion on whether or not the client should do something with this.

daemon/src/wallet_seed.rs
```diff
@@ -0,0 +1,89 @@
...
+use bdk::bitcoin::util::bip32::ExtendedPrivKey;
+use bdk::bitcoin::Network;
+use bip39::Language;
+use bip39::Mnemonic;
+use rand::thread_rng;
+use std::path::Path;
+use std::str::FromStr;
+
+#[derive(Clone)]
+pub struct WalletSeed(Mnemonic);
```
thomaseizinger: Something to mull over:

Wrapping a type is useful if you want maintain certain invariants because you can limit the functions that are called on it and put additional checks before and after.

In this case, we are

a) allowing to pass in any arbitrary `Mnemonic`
b) exposing the inner instance

and hence not enforcing any invariants. That could be a hint that we are wrapping this type unnecessarily. In such situations an alternative, flatter design would be to use an extension trait:

```rust
pub trait MnemonicExt {
	async fn initialize(seed_file: &Path) -> Self;
	/// etc
}

impl MnemonicExt for Mnemonic {

}
```

daemon/src/wallet_seed.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        if seed_file.exists() {
+            bail!("Failed restore wallet from mnemonic. Old wallet seed file was not successfully deleted.");
+        }
+
+        let seed = Self::new(mnemonic);
+        seed.write_to(seed_file).await?;
+
+        Ok(seed)
+    }
```
thomaseizinger: Quite a bit of code duplication here.

daemon/src/wallet_seed.rs
thomaseizinger: The benefit of an extension trait is that there is no change in hierachy and a user can use other, already existing APIs without having to wrap / unwrap a variable.

thomaseizinger (commented):
A few comments, nothing big.

Would like to see a test for the API model.

daemon/src/wallet_seed.rs
rishflab: cheers, this a pretty cool pattern/idea

daemon/src/routes_taker.rs
rishflab: This is the endpoint to backup your wallet. It has to return the seed words.

rishflab: We need this because we automatically generate seed words for the user upon startup and the user needs a way to get them.

daemon/src/routes_taker.rs
```diff
@@ -260,3 +282,19 @@ pub async fn post_withdraw_request(
 
     Ok(tx::to_mempool_url(txid, *network.inner()))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn wallet_restore_request_deserializes_from_valid_mnemonic() {
```
thomaseizinger, line 294: Nice, thanks!

Always good to verify, what your API models actually look like on the wire.

daemon/src/seed.rs
```diff
@@ -88,3 +89,20 @@ impl Default for Seed {
...
+    }
+}
+
+impl TryInto<Mnemonic> for Seed {
+    type Error = anyhow::Error;
+
+    fn try_into(self) -> Result<Mnemonic> {
+        Ok(Mnemonic::from_entropy(&self.0)?)
+    }
+}
```
thomaseizinger: Do we still use this?

daemon/src/wallet.rs
```diff
@@ -134,16 +136,13 @@ impl Actor {
...
-    ) -> Result<()> {
+    pub fn handle_restore(&mut self, msg: Restore, ctx: &mut xtra::Context<Self>) -> Result<()> {
         let client = bdk::electrum_client::Client::new(&self.electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
 
-        let ext_priv_key =
-            ExtendedPrivKey::new_master(self.wallet.network(), msg.seed_words.as_bytes())?;
+        let seed = Seed::try_from(msg.mnemonic)?;
+
+        let ext_priv_key = seed.derive_extended_priv_key(self.wallet.network())?;
```
thomaseizinger: Shouldn't this use functionality from `MnemonicExt`?

daemon/tests/harness/mod.rs
thomaseizinger: Yeah that is fine, wallet is mocked. You can use `from` though to avoid the error:

```suggestion
            PathBuf::from("/dev/null"), // wallet is mocked in tests, path doesn't matter
```

daemon/src/wallet_seed.rs
```diff
@@ -0,0 +1,89 @@
...
+use bdk::bitcoin::util::bip32::ExtendedPrivKey;
+use bdk::bitcoin::Network;
+use bip39::Language;
+use bip39::Mnemonic;
+use rand::thread_rng;
+use std::path::Path;
+use std::str::FromStr;
+
+#[async_trait]
+pub trait MnemonicExt {
```
thomaseizinger: By convention, I usually put these extension traits into a module that is named after the crate that they are extending. So this could go into `bip39_ext`.

daemon/src/routes_taker.rs
```diff
@@ -159,25 +160,46 @@ pub async fn post_cfd_action(
...
+
+#[rocket::get("/wallet/mnemonic")]
+pub async fn get_wallet_mnemonic(taker: &State<Taker>) -> Result<String, HttpApiProblem> {
+    let mnemonic = taker.backup_wallet().await.map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Wallet mnemonic request failed")
+            .detail(e.to_string())
+    })?;
+
+    Ok(mnemonic.to_string())
```
thomaseizinger: It [seems](https://docs.rs/bip39/1.0.1/bip39/struct.Mnemonic.html#impl-Serialize) to implement `Serialize`, so we could just return `mnemonic` here.

thomaseizinger: Also, this API is not quite consistent with the `POST` requests.

For the POST requests, you require a JSON body with a field `mnemonic` whereas here you just write the string to the HTTP body (which will set content-type `text/plain`).

Both are fine but I think we should be consistent. Either use JSON for both or neither of them. In the frontend, we have an abstraction to very easily make JSON-encoded POST requests so I'd suggest we roll with that. It is also easier to extend if we ever need to send more data back.

Same thing applies for `/wallet/reinitialize`.

thomaseizinger (changes requested):
Thanks! Looking good!

I think there are some left-overs from earlier iterations and I think we are not deriving the key correctly yet in the wallet!

daemon/src/routes_taker.rs
rishflab: i was thinking about doing this but then I saw one of our endpoints returns `Txid` as a `String`.

daemon/src/routes_taker.rs
```diff
@@ -159,25 +160,48 @@ pub async fn post_cfd_action(
...
+
+#[rocket::get("/wallet/mnemonic")]
+pub async fn get_wallet_mnemonic(taker: &State<Taker>) -> Result<Json<Mnemonic>, HttpApiProblem> {
+    let mnemonic = taker.backup_wallet().await.map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("Wallet mnemonic request failed")
+            .detail(e.to_string())
+    })?;
+
+    Ok(Json::from(mnemonic))
```
thomaseizinger: Have you verified what this actually serializes to? I am unsure.

rishflab: `Mnemonic` serializes to a space separated word list but I have no idea what rocket + wrapping it `Json` does.

thomaseizinger: Can we verify with a test that it actually serializes to JSON that we want to handle on the frontend?

daemon/src/routes_taker.rs
```diff
@@ -299,4 +300,13 @@ mod tests {
...
         assert!(serde_json::from_str::<WalletRestoreRequest>(json).is_ok());
     }
+
+    #[test]
+    fn mnemonic_serializes_to_space_seperated_word_list() {
+        let mnemonic =
+            Mnemonic::from_str("clay logic upset stadium treat nothing lumber seek inject require tourist disagree atom call lady eagle caught equip dove game connect peanut noble quick").unwrap();
+        let json = serde_json::to_string(&mnemonic).unwrap();
+
+        assert_eq!(json, "\"clay logic upset stadium treat nothing lumber seek inject require tourist disagree atom call lady eagle caught equip dove game connect peanut noble quick\"");
```
thomaseizinger, line 310: So we are sending a quoted string in the HTTP body back to the client? That is kinda odd and sort of defeats the point of using JSON. Why not send an object instead?

rishflab: There are two endpoints which would return this JSON object. Would you reuse the same type or make two different types.

thomaseizinger: Yeah same is fine. Can even use the same one from the request. As long as there is a test that freezes the impl, that is all we need.

thomaseizinger:
Closing as no longer relevant with #1016.

PR #949, opened 2021-12-22
Include name of actor in supervisor logs

thomaseizinger:
We now print things like:

```
2021-12-22 16:36:42  INFO daemon::supervisor: Spawning new instance of bitmex_price_feed::Actor
```

### DIFF
--- a/daemon/src/supervisor.rs
+++ b/daemon/src/supervisor.rs
@@ -41,12 +41,28 @@ where
     }
 
     fn spawn_new(&mut self, ctx: &mut Context<Self>) {
-        tracing::info!("Spawning new instance of actor"); // TODO: Include name
+        tracing::info!("Spawning new instance of {}", T::name());
 
         let this = ctx.address().expect("we are alive");
         let actor = (self.ctor)(this);
 
         self.tasks.add(self.context.attach(actor));
+    }
+}
+
+trait ActorName {
+    fn name() -> String;
+}
+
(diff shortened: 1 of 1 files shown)


luckysori (approved):
I like the solution! The only thing I was wondering is if you actually benefit from using the `ActorName` trait. Wouldn't a free function suffice?

thomaseizinger:
> I like the solution! The only thing I was wondering is if you actually benefit from using the `ActorName` trait. Wouldn't a free function suffice?

It would but `T::name()` is shorter to call :)

thomaseizinger:
> The only thing I was wondering is if you actually benefit from using the `ActorName` trait.

@luckysori I forgot that we had that one! Creating the name automatically and having some tests / sanity checks in place seems better to me though?
Perhaps we should move the trait to `xtra_ext` and write a bunch of tests in `lib.rs` to verify the name of several actors?

thomaseizinger:
> > The only thing I was wondering is if you actually benefit from using the `ActorName` trait.
> 
> @luckysori I forgot that we had that one! Creating the name automatically and having some tests / sanity checks in place seems better to me though? Perhaps we should move the trait to `xtra_ext` and write a bunch of tests in `lib.rs` to verify the name of several actors?

I opened https://github.com/itchysats/itchysats/issues/952.

da-kami (approved):

PR #967, opened 2021-12-24
Introduce process manager actor

### LINKED ISSUE #956: Introduce a process manager
The end-goal of the process manager is to de-couple components from each other and perform the "if this event happens, then I want that component to do X". For example, "if contract setup completes, we need to start monitoring".

The process manager should be an actor with a handler for a _single_ event. Only an event that is successfully saved to the DB must be processed further. If saving an event fails (for example, due to [OCC](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)), we have to report that **back** to the caller to inform them that whatever command they ran operated on outdated data.

The process manager should be decoupled an not know about the specific other actors. It thus cannot use addresses but SHOULD use one`MessageChannel`s per action that it sets off. For example `MessageChannel<StartMonitoring>`.

In a nutshell, the flow within `ProcessManager` will be:

1. Handler with new event gets called
2. Save it to DB
3. `match` on the event
4. Decide based on the event, which message to send into which `MessageChannel`

Initially, we can skip implementing OCC.

da-kami:
Resolves #956

1. Event Saving: Only the process manager saves events into the dabase. The initial CFD creation (prior to events) is done in the cfd actors.
2. Event post-processing: Only the process manager handles post-processing of events
3. UI updates through projection: Only the process manager updates the UI after an event is generated. (Note: Upon initial cfd creation the UI is updated through the cfd actor when inserting the cfd; Upon startup the projection actor updates itself)

As a result of this the cfd actor now does not know about the monitor anymore; and furthermore we don't have to keep bounds for wallet/oracle on most of the function calls that are handled by the process manager. This massively reduces the trait bounds and impl blocks in the cfd actors.

### DIFF
--- a/daemon/src/cfd_actors.rs
+++ b/daemon/src/cfd_actors.rs
@@ -25,7 +25,7 @@ pub async fn insert_cfd_and_update_feed(
 pub async fn handle_monitoring_event(
     event: monitor::Event,
     db: &SqlitePool,
-    process_manager_address: &xtra::Address<process_manager::Actor>,
+    process_manager: &xtra::Address<process_manager::Actor>,
 ) -> Result<()> {
     let mut conn = db.acquire().await?;
 
@@ -50,7 +50,7 @@ pub async fn handle_monitoring_event(
         monitor::Event::RevokedTransactionFound(_) => cfd.handle_revoke_confirmed(),
     };
 
-    if let Err(e) = process_manager_address
+    if let Err(e) = process_manager
         .send(process_manager::Event::new(event.clone()))
         .await?
     {
@@ -92,12 +92,12 @@ pub async fn load_cfd(order_id: OrderId, conn: &mut PoolConnection<Sqlite>) -> R
 pub async fn handle_commit(
     order_id: OrderId,
     conn: &mut PoolConnection<Sqlite>,
-    process_manager_address: &xtra::Address<process_manager::Actor>,
+    process_manager: &xtra::Address<process_manager::Actor>,
 ) -> Result<()> {
     let cfd = load_cfd(order_id, conn).await?;
 
     let event = cfd.manual_commit_to_blockchain()?;
-    if let Err(e) = process_manager_address
+    if let Err(e) = process_manager
         .send(process_manager::Event::new(event.clone()))
         .await?
     {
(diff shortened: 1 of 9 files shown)


da-kami:
@thomaseizinger I'd stop here for now. I think further steps (as in sending events directly from the setup actors to the process manager, moving `TryBroadcastTransaction` into the monitor, ...) can be done on top. I think this is in a good state to get merged :)

daemon/src/cfd_actors.rs
```diff
@@ -25,15 +22,11 @@ pub async fn insert_cfd_and_update_feed(
...
+pub async fn handle_monitoring_event(
     event: monitor::Event,
     db: &SqlitePool,
-    wallet: &xtra::Address<W>,
-    projection_address: &xtra::Address<projection::Actor>,
-) -> Result<()>
-where
-    W: xtra::Handler<wallet::TryBroadcastTransaction>,
-{
+    process_manager_address: &xtra::Address<process_manager::Actor>,
```
thomaseizinger:
```suggestion
    process_manager: &xtra::Address<process_manager::Actor>,
```

Might make things a bit shorter.

daemon/src/process_manager.rs
```diff
@@ -0,0 +1,177 @@
...
+        // 2. Post process event
+        match event.event {
+            CfdEvent::ContractSetupCompleted { dlc } => {
+                tracing::info!("Setup complete, publishing on chain now");
+
+                let lock_tx = dlc.lock.0.clone();
+                let txid = self
+                    .try_broadcast_transaction
+                    .send(wallet::TryBroadcastTransaction { tx: lock_tx })
+                    .await??;
```
thomaseizinger, line 74: I think we should not fail the entire function here because we will already have saved the event to the DB so this failing should not signal to the caller that they should try again.

Can we extract helpers on `Self` for each command that log on failure and are thus infallible?

thomaseizinger (approved):
Thanks!

luckysori (approved):
LGTM, but I know very little about these changes. I trust in your call that this is the correct direction.

daemon/src/process_manager.rs
da-kami: It can be done, but I think it is dangerous to do this, because we might not notice problems.
I think the proper way is to return different kinds of errors, where some can lead to a retry and some can lead to letting the user know that it failed. But I did not go there yet, needs a bit more thinking. 

I think it is better to fail for now than to try to "cover this up". If we fail the function the user is more likely to notice something is wrong and can e.g. try to restart. Since the event is recorded we should then try to e.g. broadcast again. 

I think retry should be designed on top, for now I would just fail. Retry will have to deal with transient and permanent errors and then act accordingly. I think this should be designed properly once we implement it.

da-kami: If we would save more fine-granular events we could record that e.g. publishing a tx failed. I started thinking about that as well, but am not sure about it yet.

da-kami:
bors ping

PR #968, opened 2021-12-24
Use heartbeat to detect loss of connection to taker daemon

### LINKED ISSUE #482: Utilise SSE heartbeat to detect loss of connection to the backend
As @da-kami pointed out, we can use SSE event heartbeat to detect the loss of the event source.

This would get rid of the keepalive requests from the frontend.

luckysori:
Fixes #482.

A few notes:
- I haven't done the same thing for the maker yet. I'm unsure if we actually intend to keep supporting the maker frontend. If so I'll do it in a separate PR, following a similar approach.
- This does not yet support reconnecting once the daemon is up again, it simply prompts the user to refresh the page. It may be simple to add this feature, but it isn't strictly necessary at this stage, as discussed with @thomaseizinger.
- I am unsure about the way I've handled the toast. In fact we get a big warning every time it renders:

```
Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
App@http://localhost:3000/src/App.tsx:32:45
Router@http://localhost:3000/node_modules/.vite/react-router-dom.js?v=b11c8a18:438:7
BrowserRouter@http://localhost:3000/node_modules/.vite/react-router-dom.js?v=b11c8a18:980:7
EnvironmentProvider@http://localhost:3000/node_modules/.vite/chunk-IDFFKOBS.js?v=b11c8a18:3863:18
ColorModeProvider@http://localhost:3000/node_modules/.vite/chunk-IDFFKOBS.js?v=b11c8a18:4024:15
ThemeProvider2@http://localhost:3000/node_modules/.vite/chunk-IDFFKOBS.js?v=b11c8a18:2145:43
ThemeProvider4@http://localhost:3000/node_modules/.vite/chunk-IDFFKOBS.js?v=b11c8a18:5922:28
IdProvider<@http://localhost:3000/node_modules/.vite/@chakra-ui_react.js?v=b11c8a18:791:18
ChakraProvider2@http://localhost:3000/node_modules/.vite/@chakra-ui_react.js?v=b11c8a18:1478:18 react-dom.development.js:67:29

```


daemon/src/to_sse_event.rs
```diff
@@ -109,3 +111,66 @@ impl ToSseEvent for Option<Quote> {
...
+        let mut heartbeat = serializer.serialize_struct("Heartbeat", 2)?;
+        heartbeat.serialize_field("event", "heartbeat")?;
+
+        let format = format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]");
+        let timestamp = self.0.format(format).expect("should always format");
+        heartbeat.serialize_field("timestamp", &timestamp)?;
+
+        heartbeat.end()
+    }
+}
```
thomaseizinger: Why this manual impl? I don't think you need to manually set the `event` field? We don't do this for any other structs?

taker-frontend/src/useEventSource.ts
```diff
@@ -0,0 +1,77 @@
...
+            {
+                title: "Connection error",
+                description: "Please ensure taker daemon is up and refresh the page to reconnect.",
+                status: "error",
+                position: "top",
+                duration: null,
+                isClosable: false,
+            },
+        );
+    }
```
thomaseizinger: Unsure but maybe we need a useEffect here that triggers on isConnected changing?

thomaseizinger (approved):
This looks great!

An idea regarding the error in the console!

taker-frontend/src/useEventSource.ts
luckysori: I'll give that a go before merging anything!

thomaseizinger (approved):

taker-frontend/src/useEventSource.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }, [isConnected, toast]);
+
+    return source;
+}
+
+const HEARTBEAT_EVENT_NAME = "heartbeat";
+const HEARTBEAT_TIMEOUT = 10000; // milliseconds
+
+export type EventSourceEvent = Event & { data: string };
```
thomaseizinger: This declaration is duplicated in `useLatestEvent`?

taker-frontend/src/useEventSource.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    // Initial timeout which will declare the event source
+    // disconnected if we don't receive a heartbeat in time
+    useEffect(() => {
+        const timeout = setTimeout(() => {
+            setIsConnected(false);
+            setSource(null);
+        }, HEARTBEAT_TIMEOUT);
+        timeoutHandle.current = timeout;
+        return;
```
thomaseizinger: Is this `return` necessary?

thomaseizinger (approved):
Needs a rebase.

taker-frontend/src/useEventSource.ts
luckysori: Also appears to not be used.

PR #971, opened 2021-12-24
Fix volume path

lukechilds:
Using the relative path like this instead of an abolute path when trying to start a container I just get the error:

```
ERROR: for itchysats_web_1  Cannot start service web: OCI runtime create failed: invalid mount {Destination:data Type:bind Source:/var/lib/docker/volumes/3373facda23457bc501bfff215d91789cc1ee4f20c043d4b929bed56fc691305/_data Options:[rbind]}: mount destination data not absolute: unknown

ERROR: for web  Cannot start service web: OCI runtime create failed: invalid mount {Destination:data Type:bind Source:/var/lib/docker/volumes/3373facda23457bc501bfff215d91789cc1ee4f20c043d4b929bed56fc691305/_data Options:[rbind]}: mount destination data not absolute: unknown
ERROR: Encountered errors while bringing up the project.
```

It seems like only certain versions of Docker fail without this, I can get your current Docker image to run on some machines but not others without this change. However with this change it runs correctly on all of my machines.

Related: https://github.com/getumbrel/umbrel/pull/1149#pullrequestreview-839455468


1010Tom (approved):
Thanks for the fix @lukechilds

bonomat (approved):
Awesome

PR #972, opened 2021-12-27
Create LICENSE

bonomat:
License file was missing

MIT was chosen because it is very permissive. It is not only welcoming open-source developers, but also businesses who can build proprietary software. This allows it to be both business-friendly and open-source friendly, while still making it possible to be monetized.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 ItchySats
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Can you add some context to either the PR or better the commit message on why MIT was chosen?

bonomat:
> Can you add some context to either the PR or better the commit message on why MIT was chosen?

Let me know if the updated commit message help.

thomaseizinger (approved):
Thanks!

luckysori (approved):

klochowicz:
IANAL, so it's not my area of expertise, but have we considered any sort of dual licensing to ensure that no other commercial entity takes the code and uses it in their proprietary software? (recent examples of elasticsearch etc come to mind).

PR #976, opened 2021-12-27
Add taker http basic auth

### LINKED ISSUE #609: Taker UI authentication
Implement basic auth for taker.

bonomat:
Resolves #609 

This is the first step to have authentication for the taker. 
For our current use of running the taker on umbrel this is sufficient. Eventually we might want to allow the user to change the password from inside of the application.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
   In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
 - Changed username for HTTP authentication to `itchysats`
+- Add taker HTTP basic authentication:
+  When starting the taker, an additional argument (`--password`) can be passed in which will be used for HTTP authentication.
 
 ## [0.3.0] - 2021-12-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The previous payout transactions are invalidated.
   The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
   In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
+- Changed username for HTTP authentication to `itchysats`
 
 ## [0.3.0] - 2021-12-09
 
--- a/daemon/src/auth.rs
+++ b/daemon/src/auth.rs
@@ -28,17 +28,17 @@ pub enum Error {
 }
 
 #[derive(PartialEq)]
-pub struct Password([u8; 32]);
+pub struct Password(String);
 
 impl From<[u8; 32]> for Password {
     fn from(bytes: [u8; 32]) -> Self {
-        Self(bytes)
+        Self(hex::encode(bytes))
     }
 }
 
 impl fmt::Display for Password {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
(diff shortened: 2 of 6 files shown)


bonomat:
Note: we should backport this to 0.3.x because it is mostly needed for https://github.com/getumbrel/umbrel/pull/1149

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   The previous payout transactions are invalidated.
   The new payout transactions spend from the same lock transaction, so the rollover happens off-chain.
   In case a maker rejects a rollover request from a taker the old oracle price event and payout transactions stay in place.
+- Changed username for HTTP authentication to `itchysats`
```
thomaseizinger, line 19: This should go under a different section, we are in `### Added` here :)

daemon/src/auth.rs
```diff
@@ -14,7 +14,8 @@ use std::str::FromStr;
 /// A request guard that can be included in handler definitions to enforce authentication.
 pub struct Authenticated {}
 
-pub const MAKER_USERNAME: &str = "maker";
+pub const USERNAME: &str = "itchysats";
+pub const TAKER_DEFAULT_PASSWORD: &str = "onbrinkofsecondbailout";
```
thomaseizinger, line 18: I'd rather not have a default password in the binary. If we want a default password for umbrel, we can always hardcode that in the docker-compose file. If `--password` is not specified, we can always derive a password from the seed.

What do you think?

daemon/src/routes_maker.rs
```diff
@@ -304,12 +304,15 @@ mod tests {
     /// Creates an "Authorization" header that matches the password above,
     /// in particular it has been created through:
     /// ```
-    /// base64(maker:hex("Now I'm feelin' so fly like a G6"))
+    /// let password = "Now I'm feelin' so fly like a G6".to_string();
+    /// let password_vec = password.as_bytes();
+    /// let password_hex = hex::encode(vec);
+    /// let basic_auth = base64::encode(&format!("{}:{}", "itchysats", password_hex));
```
thomaseizinger, line 310: That diff seems unnecessary if we are just changing the username from `maker` to `itchysats`.

daemon/src/routes_taker.rs
```diff
@@ -259,3 +267,20 @@ pub async fn post_withdraw_request(
...
+    }
+}
+
+/// A rocket responder that prompts the user to sign in to access the API.
+#[derive(rocket::Responder)]
+#[response(status = 401)]
+pub struct PromptAuthentication {
+    inner: (),
+    www_authenticate: Header<'static>,
+}
```
thomaseizinger, line 286: This could be re-used across both binaries.

daemon/src/routes_taker.rs
```diff
@@ -259,3 +267,20 @@ pub async fn post_withdraw_request(
...
 }
+
+/// A "catcher" for all 401 responses, triggers the browser's basic auth implementation.
+#[rocket::catch(401)]
+pub fn unauthorized() -> PromptAuthentication {
+    PromptAuthentication {
+        inner: (),
+        www_authenticate: Header::new("WWW-Authenticate", r#"Basic charset="UTF-8"#),
+    }
+}
```
thomaseizinger, line 278: This one probably too!

thomaseizinger (commented):
Not really a fan of the default password IN the binary.

daemon/src/auth.rs
klochowicz: also: if we would like a default password for the CI, we can always hardcode one in `cargo dev-taker` etc.

thomaseizinger:
Closed in favor of https://github.com/itchysats/itchysats/pull/982.

PR #986, opened 2021-12-29
Replace architecture and vision documents with ADRs

thomaseizinger:


docs/adr/003-external-maker-decision-logic.md
```diff
@@ -0,0 +1,38 @@
...
+Keep the maker's business decision logic for sending out offers or accepting / declining incoming orders outside the system. 
+
+## Context
+
+There are multiple aspects contributing to this decision.
+
+### 1. Business logic is financially valuable
+
+All of itchysats is open-source to allow for auditability and independent verification.
+The business logic for when to send out offers and which orders to accept is financially viable.
```
luckysori:
```suggestion
The business logic for when to send out offers and which orders to accept is financially valuable.
```

docs/adr/003-external-maker-decision-logic.md
```diff
@@ -0,0 +1,38 @@
...
+It is part of the maker's business plan and the system should allow for this to be closed source or otherwise not easily accessible.
+
+### 2. Business logic changes at a different pace
+
+The code that comprises itchysats will change with the advent of new features and / or technical improvements.
+More often than not, the business logic on when to create a new offer and which orders to accept changes in a completely different way.
+The role of "put together what changes together" suggests us that business logic and the rest of itchysats should _not_ be together.
```
luckysori:
```suggestion
The rule "put together what changes together" suggests to us that business logic and the rest of itchysats should _not_ be together.
```

klochowicz:
nit: strictly speaking, you're replacing them with an  ADL :) 

perhaps linking some docs about what ADR is would be convenient in the commit message? 
https://github.com/joelparkerhenderson/architecture-decision-record

This would give the reader a bit of context *why* this is a cool idea to use them. I for one never heard of this before, but it looks cool!

luckysori (approved):
Thank you for the effort. Very clear.

luckysori (approved):

klochowicz (approved):
this is great! I'll push a small fixup with a couple of typos on top, have a look whether you'd like to squash or discard it.

thomaseizinger:
> this is great! I'll push a small fixup with a couple of typos on top, have a look whether you'd like to squash or discard it.

Feel free to squash, looks good!

klochowicz:
> > this is great! I'll push a small fixup with a couple of typos on top, have a look whether you'd like to squash or discard it.
> 
> Feel free to squash, looks good!

rebased & squashed :)

bonomat (approved):
Nice write up. Thanks. 

ping @da-kami / @rishflab : please also have a look

docs/adr/003-external-maker-decision-logic.md
```diff
@@ -0,0 +1,38 @@
...
+
+Keep the maker's business decision logic for sending out offers or accepting / declining incoming orders outside the system.
+
+## Context
+
+There are multiple aspects contributing to this decision.
+
+### 1. Business logic is financially valuable
+
+All of itchysats is open-source to allow for auditability and independent verification.
```
da-kami, line 19: I think it could be worth adding a sentence that distinguishes what is mentioned as "business logic" from what we call "protocol". 

The business logic only allows decision making above the protocol level, i.e. the protocol is always ensured no matter what the external business logic is. 
I think for us this is clear, but for someone else reading through these docs it help.

docs/adr/007-event-sourcing-for-cfds.md
```diff
@@ -0,0 +1,34 @@
...
+Status: Active
+```
+
+## Summary
+
+Instead of recording the current state of the CFD and appending to that table, record the events that trigger the state transition and rebuild the current state on demand.
+
+## Context
+
+Currently, we are always saving the _current_ state of a CFD in the database by appending to a `states` table.
```
da-kami, line 15:
```suggestion
Previously, we were saving the _current_ state of a CFD in the database by appending to a `states` table.
```

docs/adr/009-protocols-start-is-an-event.md
```diff
@@ -0,0 +1,33 @@
...
+[ADR 007](./007-event-sourcing-for-cfds.md) introduced event sourcing for managing a CFD's state.
+Currently, only the completion (with either failure or success) of a protocol like contract-setup is modelled as an event.
+This avoids invalid start up states as the entire protocol state is in-memory until it completes.
+
+Whilst attractive, this has numerous downsides:
+
+1. We cannot easily detect that a certain protocol is already active and thus run the risk of executing it concurrently for the same CFD.
+2. State changes to a CFD are only triggered upon domain events and the UI is only updated on state changes, thus not having such an event introduces additional complexity through manual updates via the projection actor.
+3. We cannot easily detect that a _competing_ protocol is already active.
+   For example, rollover and collaborative settlement are mutually exclusive and should not be executed at the same time.
```
da-kami, line 24:
```suggestion
   For example, non-collaborative settlement and collaborative settlement are mutually exclusive and should not be executed at the same time.
```

> rollover and collaborative settlement are mutually exclusive

I think this is incorrect. 

Rollover can be triggered while collaboratively settlement and collaboratively settling can be triggered while rolling over.
Rollover does not alter the lock transaction, so agreeing to spend from it in collaborative settlement should always work regardless of a potential ongoing rollover. The only side effect might be that rollover reports a failure when trying to produce the event upon completion, because the CFD is already final. (Rollover completing while an ongoing collaborative settlement should not have any affect, because rollover does not make the cfd final.)

That said: I think we should not block collaborative settlement if there is an ongoing rollover! The user should always be able to trigger a collaborative settlement, unless the CFD is already closed non-collaboratively. Did you mean non-collaborative settlement instead of rollover here?

da-kami (approved):
Thanks for writing all this up! 
A few comments, nothing blocking though.

rishflab (approved):
awesome. thanks for the write up

docs/adr/007-event-sourcing-for-cfds.md
thomaseizinger: This was intentionally written in the context of when we made the decision: At that time of making the decision for event sourcing, we were saving the current state!

I think this is helpful if you'd read / catch up on historical decisions otherwise it is weird to come across decisions that seem to have been made with knowledge from the future.

docs/adr/009-protocols-start-is-an-event.md
thomaseizinger: No, this was actually on purpose!

1. I think non-collab settlement should never be blocked on anything.
2. I thought that from a business perspective, rolling over a contract is the opposite of settling it so they should not happen concurrently. You can settle right after or before rollover but not while it is active. I don't mind how we implement it but restricting this made sense to me. On BitMex, rollover is atomic so not really a concern but I doubt you'd be able to close / sell your position if you were in the middle of being charged interest.

docs/adr/003-external-maker-decision-logic.md
thomaseizinger: Feel free to send a follow-up patch that makes this clearer!

thomaseizinger: We never explicitly talked about these so I just winged these :)

thomaseizinger:
Thanks to everyone having a read! I hope to make it a habit for us as a team to write these so expect to be the author of some of these in the future!

PR #987, opened 2021-12-29
Allow concurrent connection setup for the maker

thomaseizinger:


daemon/src/maker_inc_connections.rs
```diff
@@ -441,6 +377,119 @@ impl Actor {
...
+    address: SocketAddr,
+) {
+    let (mut read, write, taker_id) = match upgrade(stream, noise_priv_key).await {
+        Ok((read, write, identity)) => (read, write, identity),
+        Err(e) => {
+            tracing::warn!(address = %address, "Failed to upgrade incoming connection: {:#}", e);
+            return;
+        }
+    };
+
```
bonomat, line 396: Can we have a log here telling that a new taker connected, e.g. 
```
tracing::info!(%taker_id, address = %taker_address, "New taker connected");
```

Could be useful to see because the next log will only happen if the connection was upgraded

bonomat (approved):
Cool, let's hope this solves our bug #759

daemon/src/maker_inc_connections.rs
thomaseizinger: How is that log different from the one in `upgrade`?

```rust
tracing::info!(taker_id = %taker_id, %taker_address, "Connection upgrade successful");
```

bonomat: If I understand it correctly then the one in `upgrade` is only printed if the handshake was successful, while logging here would show us all connection attempts.

daemon/src/maker_inc_connections.rs
```diff
@@ -441,6 +377,119 @@ impl Actor {
...
+        }
+    });
+    tasks.add({
+        let this = this.clone();
+
+        async move {
+            while let Ok(()) = this.send(SendHeartbeat(taker_id)).await {
+                tokio::time::sleep(heartbeat_interval).await;
+            }
+        }
```
da-kami, line 420: How is connection cleanup handled? 
Given that we can e.g. not send the heartbeat we just end this task here. How would that effect the inserted connection?

daemon/src/maker_inc_connections.rs
```diff
@@ -441,6 +377,119 @@ impl Actor {
...
+    wire::Read<wire::TakerToMaker, wire::MakerToTaker>,
+    wire::Write<wire::TakerToMaker, wire::MakerToTaker>,
+    Identity,
+)> {
+    let taker_address = stream.peer_addr().context("Failed to get peer address")?;
+
+    tracing::info!(%taker_address, "Upgrade new connection");
+
+    let transport_state = noise::responder_handshake(&mut stream, &noise_priv_key)
+        .timeout(Duration::from_secs(20))
```
da-kami, line 448: So, essentially, this is what we discussed earlier when we came to the conclusion that the handshake might block in some scenarios.
Does 20 secs make sense? When will the taker retry? If the taker retries before this times out, can it happen that we have multiple connection setups for the same taker in parallel?

da-kami (approved):
Thanks for tackling this!
I thought it through and I think it's OK but I find the connection lifecycle hard to follow and understand. I hope we will be able to plug in Quick soon, that will hopefully make the flow easier to understand?

One question at the start:
Was it previously possible to have concurrent connection setups with different takers? I thought that already worked.
I am not sure we need concurrent connection setups with the same taker to be honest - what would be the advantage of that?
I thought that we would just add a timeout for the handshake and that would do. What are the advantages of this solution?

Here is my understanding of the current solution:

When the taker connects we add a task for setting up the connection. This tasks ends after `setup_new_connection` completes.
The `setup_new_connection` task first  upgrades the connection.
Upgrading the connection includes the noise handshake and the hello message. 
If upgrading fails we just log an error and return from `setup_new_connection`, which will result in the task completing without setting up the connection.
If upgrade worked, we add tasks for reading messages and the heartbeat and return the connection to ourselves via `ConnectionReady`. Since `setup_new_connection` is handled as task we can use `send` for sending `ConnectionReady`.
Handling `ConnectionReady` notifies the cfd actor of the taker connection and then inserts the connection. The inserted connection is used by `sendToTaker` and when broadcasting the order.

We remove a connection (by calling `drop_taker_connection`) in three different scenarios: 

1. we are unable to send the order upon broadcasting (we clean them up after finishing the loop)
2. we are unable to send a `MakerToTaker` message to the taker
3. we are unable to read from the connection

The lifecycle of the tasks that we add in `setup_new_connection` is handled by completing them upon failure to `send` using the connection. 

We don't drop the connection when sending the heartbeat fails. I find that a bit odd, but I think it is a conscious decision (?) because we just end the task, the connection cleanup is handled somewhere else. I feel we could also call drop there. 
I am not entirely sure why we have a `ReadFail` message and not just a message for dropping the connection? - I think that could align dropping the connection for all cases where we run in an error sending/reading.

daemon/src/maker_inc_connections.rs
thomaseizinger: I am not sure I follow.

- "Here" as to where the comment is made is after `upgrade` completes _successfully_ so that is already logged inside `upgrade`.
- Your log suggests to log `taker_id` but that is only available after an upgrade. Noise needs to complete the handshake to learn the other party's identity. That is what `upgrade` does. We could log between the noise and the `Hello` handshake but I am not sure that is worth it?
- There is a log inside `upgrade` that logs for every connection attempt. It is on line 445.

daemon/src/maker_inc_connections.rs
thomaseizinger: The taker only retries after failed attempts? If we are within this timeout, the handshake is still going on so the taker does not know about the outcome yet?

20 seconds should be plenty.

thomaseizinger:
> Thanks for tackling this! I thought it through and I think it's OK but I find the connection lifecycle hard to follow and understand. I hope we will be able to plug in Quick soon, that will hopefully make the flow easier to understand?

The protocol is called QUIC, not Quick. What QUIC will improve is the routing of messages to particular _instances_ of actors. Additionally, because QUIC implies TLS, we are not going to have the noise handshake. The rest of the connection lifecycle stays pretty much the same. We still need to accept connections, do the `Hello` handshake and send heartbeats (if we want them).

> One question at the start: Was it previously possible to have concurrent connection setups with different takers? I thought that already worked. I am not sure we need concurrent connection setups with the same taker to be honest - what would be the advantage of that? I thought that we would just add a timeout for the handshake and that would do. What are the advantages of this solution?

No it was not possible. We previously blocked the `handle_new_connection_impl` on the completion of the upgrade so new items were not pulled from the stream of incoming connections until the current upgrade completed.

> Here is my understanding of the current solution:
> 
> When the taker connects we add a task for setting up the connection. This tasks ends after `setup_new_connection` completes. The `setup_new_connection` task first upgrades the connection. Upgrading the connection includes the noise handshake and the hello message. If upgrading fails we just log an error and return from `setup_new_connection`, which will result in the task completing without setting up the connection. If upgrade worked, we add tasks for reading messages and the heartbeat and return the connection to ourselves via `ConnectionReady`. Since `setup_new_connection` is handled as task we can use `send` for sending `ConnectionReady`. Handling `ConnectionReady` notifies the cfd actor of the taker connection and then inserts the connection. The inserted connection is used by `sendToTaker` and when broadcasting the order.

Correct. Sending a message to ourselves enables the handling multiple connection upgrades.concurrently.

> We remove a connection (by calling `drop_taker_connection`) in three different scenarios:
> 
>     1. we are unable to send the order upon broadcasting (we clean them up after finishing the loop)
> 
>     2. we are unable to send a `MakerToTaker` message to the taker
> 
>     3. we are unable to read from the connection
> 
> 
> The lifecycle of the tasks that we add in `setup_new_connection` is handled by completing them upon failure to `send` using the connection.
> 
> We don't drop the connection when sending the heartbeat fails. I find that a bit odd, but I think it is a conscious decision (?) because we just end the task, the connection cleanup is handled somewhere else. I feel we could also call drop there.

We could clean up the message upon heartbeat failure. Currently, we rely on the _taker_ to drop the connection if they don't receive a heartbeat within a certain timeout.

> I am not entirely sure why we have a `ReadFail` message and not just a message for dropping the connection? - I think that could align dropping the connection for all cases where we run in an error sending/reading.

I'd argue it is cleaner if a task only reports its failure and does not know what the consequence of that is.

thomaseizinger:
I am merging this so we can backport it!

PR #991, opened 2021-12-29
Add ContractSetupStarted Cfd event

klochowicz:
Allow reasoning about started contract setup, to be able to constrain available
actions and present the Cfd state in the UI.

Reinstate the ContractSetup UI state, both in the frontend and in the actor tests.

I wasn't sure about accessing db in `started()`, so opted for now for returning the event along with `SetupParams`. Happy to change that if deemed to be wrong.

### DIFF
--- a/daemon/src/cfd_actors.rs
+++ b/daemon/src/cfd_actors.rs
@@ -1,5 +1,6 @@
 use crate::db;
 use crate::model::cfd::Cfd;
+use crate::model::cfd::Event;
 use crate::model::cfd::OrderId;
 use crate::monitor;
 use crate::oracle;
@@ -87,6 +88,21 @@ pub async fn load_cfd(order_id: OrderId, conn: &mut PoolConnection<Sqlite>) -> R
         events,
     );
     Ok(cfd)
+}
+
+pub async fn apply_event(
+    process_manager: &xtra::Address<process_manager::Actor>,
+    event: Event,
+) -> Result<()> {
+    // FIXME: Failing to apply an event is a serious error and should not result
+    // in Ok() return from this function
+    if let Err(e) = process_manager
+        .send(process_manager::Event::new(event.clone()))
+        .await?
+    {
+        tracing::error!("Sending event to process manager failed: {:#}", e);
+    }
+    Ok(())
 }
 
 pub async fn handle_commit(
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -242,6 +242,7 @@ impl Event {
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 #[serde(tag = "name", content = "data")]
 pub enum CfdEvent {
+    ContractSetupStarted,
     ContractSetupCompleted {
         dlc: Dlc,
     },
@@ -391,6 +392,8 @@ pub struct Cfd {
 
     cet_timelock_expired: bool,
     refund_timelock_expired: bool,
+
+    during_contract_setup: bool,
 }
 
 impl Cfd {
@@ -429,6 +432,7 @@ impl Cfd {
             collaborative_settlement_finality: false,
             cet_timelock_expired: false,
             refund_timelock_expired: false,
+            during_contract_setup: false,
         }
     }
 
@@ -545,7 +549,7 @@ impl Cfd {
         self.collaborative_settlement_finality || self.cet_finality || self.refund_finality
     }
 
-    pub fn start_contract_setup(&self) -> Result<SetupParams> {
+    pub fn start_contract_setup(&self) -> Result<(Event, SetupParams)> {
         if self.version > 0 {
             bail!("Start contract not allowed in version {}", self.version)
         }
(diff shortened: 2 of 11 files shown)


daemon/src/cfd_actors.rs
```diff
@@ -89,6 +90,19 @@ pub async fn load_cfd(order_id: OrderId, conn: &mut PoolConnection<Sqlite>) -> R
...
+    process_manager: &xtra::Address<process_manager::Actor>,
+    event: Event,
+) -> Result<()> {
+    if let Err(e) = process_manager
+        .send(process_manager::Event::new(event.clone()))
+        .await?
+    {
+        tracing::error!("Sending event to process manager failed: {:#}", e);
+    }
+    Ok(())
```
thomaseizinger, line 105: I know you have only moved the code but failing to apply the event is a serious error and we should not return `Ok` in that case. Can be fixed in a follow-up because it is only moved here!

daemon/src/model/cfd.rs
```diff
@@ -545,7 +549,7 @@ impl Cfd {
         self.collaborative_settlement_finality || self.cet_finality || self.refund_finality
     }
 
-    pub fn start_contract_setup(&self) -> Result<(SetupParams, Identity)> {
+    pub fn start_contract_setup(&self) -> Result<(Event, SetupParams, Identity)> {
```
thomaseizinger: I think we can move `SetupParams` and `Identity` into the `ContractSetupStarted` event. It might be useful for the UI for example to display the params of the setup.

thomaseizinger: It will also simplify this return type and make it uniform with the other handlers!

daemon/src/model/cfd.rs
```diff
@@ -655,9 +660,10 @@ impl Cfd {
     }
 
     pub fn setup_contract(self, completed: SetupCompleted) -> Result<Event> {
-        if self.version > 0 {
+        // Version 1 is acceptable, as it means that we started contract setup
+        if self.version > 1 {
```
thomaseizinger, line 666: We could check for the `during_contract_setup` boolean here instead? I think that would be more accurate.

thomaseizinger (approved):
Thanks for this!

As a follow-up, we can now remove sending `Completed` back to the `{taker,maker}_cfd::Actor` and instead load the CFD from the DB inside the actor and call `setup_contract` directly!

daemon/src/cfd_actors.rs
klochowicz: ok, I'll leave a TODO now and deal with it in a follow-up!

daemon/src/model/cfd.rs
klochowicz: I might do it later, if UI requires it - the price for changing it now in making it really awkward to retrieve the setup params at the callsite (helper with an unwrap :/ due to downcasting to the particular event)

thomaseizinger: Yeah I saw it on Lucas' branch. That is quite unfortunate ...

Happy to leave as is for now.

daemon/src/cfd_actors.rs
```diff
@@ -89,6 +90,21 @@ pub async fn load_cfd(order_id: OrderId, conn: &mut PoolConnection<Sqlite>) -> R
     Ok(cfd)
 }
 
+pub async fn apply_event(
```
thomaseizinger, line 93: I don't think this abstraction is worth anything to be honest. We should be able to `??` this on all call-sites. Also, we could implement `xtra::Message` directly for `Event` then we wouldn't need to wrap it in `process_manager::Event` if you are concerned about boilerplate.

The error message can move into the message handler.

daemon/src/model/cfd.rs
klochowicz: I've actually got a follow-up mostly ready, but without the Accepted event it's actually a regression in functionality as per this patch, so I'll save it up for the follow-up.

daemon/src/cfd_actors.rs
klochowicz: so far it's saving me 6 lines of code on every callsite - I will clean this up as suggested in a follow-up refactor.

daemon/src/model/cfd.rs
klochowicz: We can't really, until we have the `Accepted` event too :( 
In order to do that, we need to move the `Started` event into started() method, and that it turn makes everything jump into the contract setup state (even before the order is accepted).
Added a TODO to address in follow-up.

daemon/src/cfd_actors.rs
thomaseizinger: > so far it's saving me 6 lines of code on every callsite

The way the function currently operates yes. But I'd argue it is wrong and we should fail instead of just logging! Generally, I am not a fan of these kind of helpers. I agree that duplicating these 6 lines is bad. In my experience though, hiding this pain will not actually get it solved because we forget about it and eventually, you end up with 7 layers of indirection until you reach the code that does what you want to do. Instead, I would like us to strive for making the actual code (sending the event to the process manager) as ergonomic as possible and until we are there, I am happy (and would perhaps even advocate for) leaving some pain behind to remind us of that :D

PR #992, opened 2021-12-29
Move counterparty identity into SetupParams

klochowicz:
If it's needed by contract setup, then it should be part of the setup params. It's an implementation detail from the model's perspective that one side doesn't really use it.


thomaseizinger (approved):
Probably obsolete with https://github.com/itchysats/itchysats/pull/991#discussion_r776526538.

klochowicz:
> Probably obsolete with [#991 (comment)](https://github.com/itchysats/itchysats/pull/991#discussion_r776526538).

I think it's still a bit cleaner to return just this - in this case, it will be stored inside the event!

Issue #959, opened 2021-12-23
Introduce domain event for `CollaborativeSettlementStarted`


thomaseizinger:
Same reasoning as https://github.com/itchysats/itchysats/issues/957 and https://github.com/itchysats/itchysats/issues/956.

da-kami:
done in #994

PR #994, opened 2021-12-30
Emit `CollaborativeSettlementProposed` event in taker daemon

luckysori:
Work towards #959, maker side is missing.

### DIFF
--- a/daemon/src/collab_settlement_taker.rs
+++ b/daemon/src/collab_settlement_taker.rs
@@ -1,78 +1,106 @@
 use crate::address_map::ActorName;
 use crate::address_map::Stopping;
+use crate::cfd_actors::load_cfd;
 use crate::connection;
-use crate::model::cfd::Cfd;
+use crate::model::cfd;
+use crate::model::cfd::CfdEvent;
 use crate::model::cfd::CollaborativeSettlement;
 use crate::model::cfd::Completed;
+use crate::model::cfd::OrderId;
 use crate::model::cfd::SettlementKind;
 use crate::model::cfd::SettlementProposal;
 use crate::model::Price;
+use crate::process_manager;
 use crate::projection;
 use crate::send_async_safe::SendAsyncSafe;
 use crate::wire;
 use anyhow::Result;
 use async_trait::async_trait;
-use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 
 pub struct Actor {
-    cfd: Cfd,
-    projection: xtra::Address<projection::Actor>,
-    on_completed: Box<dyn MessageChannel<Completed<CollaborativeSettlement>>>,
+    proposal: Option<SettlementProposal>,
+    order_id: OrderId,
+    current_price: Price,
+    n_payouts: usize,
     connection: xtra::Address<connection::Actor>,
-    proposal: SettlementProposal,
+    process_manager: xtra::Address<process_manager::Actor>,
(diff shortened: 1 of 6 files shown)


daemon/src/collab_settlement_taker.rs
```diff
@@ -1,78 +1,106 @@
...
+        let event = cfd.propose_collaborative_settlement(self.current_price, self.n_payouts)?;
+        let proposal = if let cfd::Event {
+            event: CfdEvent::CollaborativeSettlementProposed { ref proposal },
+            ..
+        } = event
+        {
+            proposal
+        } else {
+            unreachable!()
+        };
```
thomaseizinger, line 67: Damn that is ugly haha! I didn't think of that ...

daemon/src/process_manager.rs
```diff
@@ -63,8 +63,9 @@ impl Actor {
         append_event(event.clone(), &mut conn).await?;
 
         // 2. Post process event
+        use CfdEvent::*;
```
thomaseizinger, line 66: This could go into the prev commit.

daemon/src/taker_cfd.rs
```diff
@@ -132,39 +128,24 @@ where
...
-        let cfd = load_cfd(order_id, &mut conn).await?;
-
-        let this = ctx
-            .address()
-            .expect("actor to be able to give address to itself");
-        let (addr, fut) = collab_settlement_taker::Actor::new(
-            cfd,
-            self.projection_actor.clone(),
-            this,
+        let (_, fut) = collab_settlement_taker::Actor::new(
```
thomaseizinger: Are you sure we can just drop the address right away? Won't that reduce the ref-count to zero? I think we may have to keep it alive at least until the end of this function so that the `started` function runs (where we clone the address and send it to the connection::Actor, god that is complicated ....).

thomaseizinger (approved):
Good direction so far, I did not consider the uglyness of matching on the event returned from the command inside the actor. That is definitely a sign that we should actually not be doing this ...

daemon/src/collab_settlement_taker.rs
klochowicz: that was the reason I pushed back on similar comment on my review :P returning a tuple would probably be nicer after all, even if we store it both in the event for future use and pass back here!

luckysori: I also tried to avoid this by sending the proposal to the maker in `process_manager::Actor`, but that doesn't quite work because we'd need an address to the `collab_settlement_taker::Actor` which cannot be serialized as part of the event (and also doesn't belong there).

daemon/src/taker_cfd.rs
luckysori: We actually need to save it, because the ref-count reaches zero before the spawned task clones the address (unless we sleep here which would be ludicrous).

luckysori: It's very unfortunate because this actor doesn't care about these ephemeral actors after spawning them.

thomaseizinger: What you could do is:

```rust
self.tasks.add(async move {
	let _addr = address;
	fut.await;
});
```

To keep the address alive for as long as the context is running. I am not sure if that is elegant or hacky :joy:

luckysori: I considered this, but it's a bit suspicious...

PR #995, opened 2021-12-30
Refuse to accept 2nd connection of already connected taker

thomaseizinger:
Noticed that we currently don't check for existing connections of the same taker.

Introduce a utility on `Tasks` to make it easier to run fallible futures in the background.
I've split this out of https://github.com/itchysats/itchysats/pull/851 which is not quite done yet in the hope that these two things are pretty non-controversial!

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -104,6 +104,29 @@ impl Tasks {
         let handle = f.spawn_with_handle();
         self.0.push(handle);
     }
+
+    /// Spawn a fallible task on the runtime and remembers the handle.
+    ///
+    /// The task will be stopped if this instance of [`Tasks`] goes out of scope.
+    /// If the task fails, the `err_handler` will be invoked.
+    pub fn add_fallible<E, EF>(
+        &mut self,
+        f: impl Future<Output = Result<(), E>> + Send + 'static,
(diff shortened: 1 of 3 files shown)


daemon/src/lib.rs
```diff
@@ -109,6 +109,29 @@ impl Tasks {
...
+    pub fn add_fallible<E, EF>(
+        &mut self,
+        f: impl Future<Output = Result<(), E>> + Send + 'static,
+        err_handler: impl FnOnce(E) -> EF + Send + 'static,
+    ) where
+        E: Send + 'static,
+        EF: Future<Output = ()> + Send + 'static,
+    {
+        let fut = async move {
+            match f.await {
```
luckysori, line 121: 🔧 `if let Err(e)` might be ever so slightly cleaner.

luckysori (approved):

daemon/src/lib.rs
thomaseizinger: I wanted to be explicit that the `Ok` part returns `()` but maybe it is not worth it.

thomaseizinger:
Feel free to merge once it has some approvals.

klochowicz (approved):
LGTM!

PR #1000, opened 2021-12-31
Increase heartbeat timeout in debug mode

klochowicz:
Address sporadic test failures due to hitting this timeout in contract setup test.


klochowicz:
When we reach feature parity, I'd like to experiment with the suggestion presented in: https://github.com/itchysats/itchysats/pull/906
and remove this workaround altogether.

luckysori (approved):

PR #1001, opened 2021-12-31
Add two more collaborative-settlement-related events

### LINKED ISSUE #959: Introduce domain event for `CollaborativeSettlementStarted`

luckysori:
Untested manually but the automatic tests pass locally 🙃 

Fixes #959.

### DIFF
--- a/daemon/src/collab_settlement_maker.rs
+++ b/daemon/src/collab_settlement_maker.rs
@@ -1,28 +1,26 @@
 use crate::address_map::ActorName;
 use crate::address_map::Stopping;
+use crate::cfd_actors::load_cfd;
 use crate::maker_inc_connections;
-use crate::model::cfd::Cfd;
 use crate::model::cfd::CollaborativeSettlement;
 use crate::model::cfd::Completed;
-use crate::model::cfd::SettlementKind;
 use crate::model::cfd::SettlementProposal;
 use crate::model::Identity;
-use crate::projection;
-use crate::xtra_ext::LogFailure;
+use crate::process_manager;
 use anyhow::Context;
+use anyhow::Result;
 use async_trait::async_trait;
 use maia::secp256k1_zkp::Signature;
 use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 
 pub struct Actor {
-    cfd: Cfd,
-    projection: xtra::Address<projection::Actor>,
-    on_completed: Box<dyn MessageChannel<Completed<CollaborativeSettlement>>>,
     proposal: SettlementProposal,
     taker_id: Identity,
     connections: Box<dyn MessageChannel<maker_inc_connections::settlement::Response>>,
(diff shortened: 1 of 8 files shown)


daemon/src/model/cfd.rs
```diff
@@ -672,6 +680,41 @@ impl Cfd {
...
+    ) -> Result<Event> {
+        anyhow::ensure!(
+            !self.is_in_collaborative_settlement()
+                && self.role == Role::Maker
+                && self.can_settle_collaboratively()
+                && proposal.order_id == self.id,
+            "Failed to start collaborative settlement"
+        );
+
+        // TODO: Should we validate if the proposal has the expected
```
da-kami: I think now :) 
The business maker business logic should check that.

daemon/src/collab_settlement_maker.rs
```diff
@@ -104,100 +105,111 @@ impl xtra::Actor for Actor {
...
+
+        let this = ctx.address().expect("self to be alive");
+        self.connections
+            .send(maker_inc_connections::settlement::Response {
+                taker_id: self.taker_id,
+                order_id,
+                decision: maker_inc_connections::settlement::Decision::Accept { address: this },
+            })
             .await
+            .context("Failed to inform taker about settlement acceptance")??;
```
da-kami, line 194: Any particular reason why you handle this sending error different than in `reject`? I think it should be handled in the same way. I think for now we could opt for just dropping the error if we run into `Disconnected`, even though I'm not a fan of dropping things :)

daemon/src/model/cfd.rs
```diff
@@ -261,6 +261,10 @@ pub enum CfdEvent {
     CollaborativeSettlementProposed {
         proposal: SettlementProposal,
     },
+    CollaborativeSettlementProposalReceived {
```
da-kami: Let's do this as discussed with using a `CollaborativeSettlementStarted` for both, similar to what we prepared for rollover.

da-kami (changes requested):

daemon/src/collab_settlement_maker.rs
luckysori: Good point, but I think I still prefer my approach here.

If we fail to send this message to the taker, we must `complete` the collaborative settlement protocol with `Failed`, much like with the rest of the errors in this method. Keep in mind that this `accept` method is called by `handle(&mut self, msg: Accepted)`, where any error originated from this results in completing the protocol with `Failed`.

I did reject differently because it is simpler to complete the protocol with `Rejected` unconditionally. If the rejection message doesn't reach the taker, I don't see a clear upside to recording the fact that it failed (whilst it was being rejected).

da-kami (approved):

luckysori:
bors retry

PR #1002, opened 2021-12-31
Emit ContractSetupProposed event as soon as setup_contract actors are started

klochowicz:
In order to represent states adequately in the UI, we need to differentiate
between the proposed contract setup (the order got requested by the taker) and
the actual contract setup protocol (the order got accepted by the maker, so the
protocol is started).

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -244,7 +244,10 @@ impl Event {
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 #[serde(tag = "name", content = "data")]
 pub enum CfdEvent {
-    ContractSetupStarted,
+    ContractSetupProposed,
+    ContractSetupStarted {
+        params: SetupParams,
+    },
     ContractSetupCompleted {
         dlc: Dlc,
     },
@@ -399,7 +402,7 @@ pub struct Cfd {
     cet_timelock_expired: bool,
     refund_timelock_expired: bool,
 
-    during_contract_setup: bool,
+    during_contract_setup: bool, // from contract setup proposal until completion
 }
 
 impl Cfd {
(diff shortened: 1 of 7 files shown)


daemon/src/model/cfd.rs
```diff
@@ -244,7 +244,10 @@ impl Event {
 #[derive(Serialize, Deserialize, PartialEq, Debug, Clone)]
 #[serde(tag = "name", content = "data")]
 pub enum CfdEvent {
-    ContractSetupStarted,
+    ContractSetupProposed,
```
thomaseizinger, line 247: I am not sure this is a modelling step in the right direction.

"Proposing contract setup" is a completely technical term, having little to do with our domain. (Check the term ubiqitous language from DDD for more)

Domain-wise, what we are doing here is placing an order with the maker (at least within our current also flawed offer/order terminology) for the current offer.

In the UI we are displaying this on the CFD but I've also always found that weird. From my experience, your placed orders show up in a different place from your currently open positions.

I think the only immediate conclusion here is that we need to sort this mess out 😅
I have questions like:

- When does a CFD's lifetime start in our model?
- What is an offer?
- What is an order?
- When do we transition from one to the other?

These events were only intended to model CFDs, not orders or offers. My assumption was always that we only ever have a CFD once our (market?)-order is filled(?) by the maker. So one could argue that the state of placing an order should be associated with the order instead, not the CFD.

Is there a hacky way of how we can solve this without adding a new event?

thomaseizinger (commented):
A comment to mull over, lets do a workshop on this soon!

daemon/src/model/cfd.rs
klochowicz: a "hacky" way (or at least, inconsistent with the other ephemeral actors), is the current one that got merged. I'm happy to leave this one hanging - the primary motivation for it was that I didn't want to leave your previous feedback unaddressed (we only need additional event if we move the first event into the `started()` method of the actor, if it stays where it is now, it's fine).

Happy to chat further when you're back.

luckysori:
Any updates on this, @klochowicz @thomaseizinger?

thomaseizinger:
> Any updates on this, @klochowicz @thomaseizinger?

I think we need to do some more modelling to get to the bottom of this.

klochowicz:
@thomaseizinger I agree. This was an attempt to bring the protocols more in line, addressing something that @da-kami mentioned again recently.

luckysori:
Should we close it? Open PRs stress me out haha.

klochowicz:
> Should we close it? Open PRs stress me out haha.

I feel you, I also prefer for that list to be as small as possible 👍 . I've opened an issue to track further discussion. It certainly is not ready to be merged with that in mind.  #1097

PR #1003, opened 2021-12-31
Run all actor tests by default

klochowicz:
Contract setup test is expensive, but it does not deserve special treatment
compared to collab and non-collab settlement tests, which take even longer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,8 +95,8 @@ jobs:
       - uses: Swatinem/rust-cache@v1.3.0
       - run: cargo build --bins --tests
       - run: cargo test --workspace
-      # Ignored tests should be run on the CI
-      - run: cargo test --workspace -- --ignored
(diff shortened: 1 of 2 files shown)


thomaseizinger (approved):

.github/workflows/ci.yml
```diff
@@ -95,8 +95,8 @@ jobs:
       - uses: Swatinem/rust-cache@v1.3.0
       - run: cargo build --bins --tests
       - run: cargo test --workspace
-      # Ignored tests should be run on the CI
-      - run: cargo test --workspace -- --ignored
+      # Ignored tests that should be run on the CI - reenable if there are any
+      # - run: cargo test --workspace -- --ignored
```
luckysori, line 99: 🔧 I would just remove this if we are not using it atm.

luckysori (approved):

klochowicz:
Bors r+

PR #966, opened 2021-12-24
Re-implement handling of `RolloverCompleted` for taker

thomaseizinger:

PR #1014, opened 2022-01-04
Auto-/rollover with event sourcing

### LINKED ISSUE #466: Actor test for roll-over
Recently introduced encryption in #452  can have an impact rollover behaviour due to large message size. 
Add a test to ensure that we didn't break the functionality.

### LINKED ISSUE #958: Introduce domain event for `RolloverStarted`
Similar as to https://github.com/itchysats/itchysats/issues/957, we want an event for `RolloverStarted`.

In addition to https://github.com/itchysats/itchysats/issues/957, this is also important because we want to:

a) disallow rolling over the same CFD concurrently (something that is much more likely to happen)
b) disallow triggering protocols like collaborative settlement whilst "rollover" is operating

### LINKED ISSUE #1013: Re-introduce the auto-rollover unit tests and auto-rollover logic

da-kami:
Finishes the rollover / auto-rollover implementation with event sourcing. This should finish our perpetual swap setup.

Fixes #1013 
Fixes https://github.com/itchysats/itchysats/issues/958
Fixes #466

### DIFF
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -4,51 +4,47 @@ use crate::cfd_actors::load_cfd;
 use crate::connection;
 use crate::db;
 use crate::model::cfd::OrderId;
-use crate::model::cfd::RolloverCompleted;
-use crate::monitor;
 use crate::oracle;
-use crate::projection;
+use crate::process_manager;
 use crate::rollover_taker;
+use crate::try_continue;
 use crate::Tasks;
+use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
 use maia::secp256k1_zkp::schnorrsig;
 use std::time::Duration;
+use time::OffsetDateTime;
 use xtra::Actor as _;
 use xtra::Address;
 use xtra_productivity::xtra_productivity;
 
-pub struct Actor<O, M> {
+pub struct Actor<O> {
     db: sqlx::SqlitePool,
     oracle_pk: schnorrsig::PublicKey,
-    projection_actor: Address<projection::Actor>,
-    conn_actor: Address<connection::Actor>,
-    _monitor_actor: Address<M>,
-    oracle_actor: Address<O>,
+    process_manager: Address<process_manager::Actor>,
+    conn: Address<connection::Actor>,
+    oracle: Address<O>,
     n_payouts: usize,
-
     rollover_actors: AddressMap<OrderId, rollover_taker::Actor>,
-
     tasks: Tasks,
 }
 
-impl<O, M> Actor<O, M> {
+impl<O> Actor<O> {
     pub fn new(
         db: sqlx::SqlitePool,
         oracle_pk: schnorrsig::PublicKey,
-        projection_actor: Address<projection::Actor>,
-        conn_actor: Address<connection::Actor>,
-        monitor_actor: Address<M>,
-        oracle_actor: Address<O>,
+        process_manager: Address<process_manager::Actor>,
+        conn: Address<connection::Actor>,
+        oracle: Address<O>,
         n_payouts: usize,
     ) -> Self {
         Self {
             db,
             oracle_pk,
-            projection_actor,
-            conn_actor,
-            _monitor_actor: monitor_actor,
-            oracle_actor,
+            process_manager,
+            conn,
+            oracle,
             n_payouts,
             rollover_actors: AddressMap::default(),
             tasks: Tasks::default(),
@@ -57,10 +53,9 @@ impl<O, M> Actor<O, M> {
 }
 
 #[xtra_productivity]
-impl<O, M> Actor<O, M>
+impl<O> Actor<O>
 where
-    M: xtra::Handler<monitor::StartMonitoring>,
-    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+    O: xtra::Handler<oracle::GetAnnouncement>,
 {
     async fn handle(&mut self, _msg: AutoRollover, ctx: &mut xtra::Context<Self>) -> Result<()> {
         tracing::trace!("Checking all CFDs for rollover eligibility");
@@ -73,57 +68,61 @@ where
             .expect("actor to be able to give address to itself");
 
         for id in cfd_ids {
-            let disconnected = match self.rollover_actors.get_disconnected(id) {
-                Ok(disconnected) => disconnected,
-                Err(_) => {
-                    tracing::debug!(order_id=%id, "Rollover already in progress");
-                    continue;
-                }
-            };
+            try_continue!(async {
+                let cfd = load_cfd(id, &mut conn).await?;
+                cfd.can_auto_rollover_taker(OffsetDateTime::now_utc())?;
 
-            // TODO: Shall this have a try_continue?
-            let cfd = load_cfd(id, &mut conn).await?;
+                this.send(Rollover(id)).await??;
 
-            let (addr, fut) = rollover_taker::Actor::new(
-                (cfd, self.n_payouts),
-                self.oracle_pk,
-                self.conn_actor.clone(),
-                &self.oracle_actor,
-                self.projection_actor.clone(),
-                &this,
-                (&this, &self.conn_actor),
-            )
-            .create(None)
-            .run();
-
-            disconnected.insert(addr);
-            self.tasks.add(fut);
+                anyhow::Ok(())
+            }
+            .await
+            .context("Cannot roll over"));
         }
 
         Ok(())
     }
-}
 
-#[xtra_productivity(message_impl = false)]
-impl<O, M> Actor<O, M>
-where
-    O: 'static,
-    M: 'static,
-    M: xtra::Handler<monitor::StartMonitoring>,
-    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
-{
-    async fn handle_rollover_completed(&mut self, _: RolloverCompleted) -> Result<()> {
-        // TODO: Implement this in terms of event sourcing
+    async fn handle(
+        &mut self,
+        Rollover(order_id): Rollover,
+        ctx: &mut xtra::Context<Self>,
+    ) -> Result<()> {
+        let disconnected = match self.rollover_actors.get_disconnected(order_id) {
+            Ok(disconnected) => disconnected,
+            Err(_) => {
+                tracing::debug!(%order_id, "Rollover already in progress");
+                return Ok(());
+            }
+        };
+
+        let this = ctx
+            .address()
+            .expect("actor to be able to give address to itself");
+        let (addr, fut) = rollover_taker::Actor::new(
+            order_id,
+            self.n_payouts,
+            self.oracle_pk,
+            self.conn.clone(),
+            &self.oracle,
+            self.process_manager.clone(),
+            (&this, &self.conn),
+            self.db.clone(),
+        )
+        .create(None)
+        .run();
+
+        disconnected.insert(addr);
+        self.tasks.add(fut);
 
         Ok(())
     }
 }
 
 #[xtra_productivity(message_impl = false)]
-impl<O, M> Actor<O, M>
+impl<O> Actor<O>
 where
-    M: xtra::Handler<monitor::StartMonitoring>,
-    O: xtra::Handler<oracle::MonitorAttestation> + xtra::Handler<oracle::GetAnnouncement>,
+    O: xtra::Handler<oracle::GetAnnouncement>,
 {
     async fn handle_rollover_actor_stopping(&mut self, msg: Stopping<rollover_taker::Actor>) {
         self.rollover_actors.gc(msg);
@@ -131,10 +130,9 @@ where
 }
 
 #[async_trait]
-impl<O, M> xtra::Actor for Actor<O, M>
+impl<O> xtra::Actor for Actor<O>
 where
     O: 'static,
-    M: 'static,
     Self: xtra::Handler<AutoRollover>,
 {
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
@@ -149,5 +147,11 @@ where
     }
 }
 
-/// Message to trigger roll-over on a regular interval
+/// Message sent to ourselves at an interval to check if rollover can
+/// be triggered for any of the CFDs in the database.
 pub struct AutoRollover;
+
+/// Message used to trigger rollover internally within the `auto_rollover::Actor`
+///
+/// This helps us trigger rollover in the tests unconditionally of time.
+pub struct Rollover(pub OrderId);
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -137,10 +137,7 @@ where
 {
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let fut = ctx
-            .notify_interval(
-                rollover_taker::MAX_ROLLOVER_DURATION + Duration::from_secs(60),
-                || AutoRollover,
-            )
+            .notify_interval(Duration::from_secs(5 * 60), || AutoRollover)
             .expect("we are alive");
 
         self.tasks.add(fut);
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -275,14 +275,14 @@ where
 
     pub async fn accept_rollover(&self, order_id: OrderId) -> Result<()> {
         self.cfd_actor_addr
-            .send(maker_cfd::AcceptRollOver { order_id })
+            .send(maker_cfd::AcceptRollover { order_id })
             .await??;
         Ok(())
     }
 
     pub async fn reject_rollover(&self, order_id: OrderId) -> Result<()> {
         self.cfd_actor_addr
-            .send(maker_cfd::RejectRollOver { order_id })
+            .send(maker_cfd::RejectRollover { order_id })
             .await??;
         Ok(())
     }
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -1,7 +1,6 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 #![warn(clippy::disallowed_method)]
 use crate::bitcoin::Txid;
-use crate::model::cfd::Cfd;
 use crate::model::cfd::Order;
 use crate::model::cfd::OrderId;
 use crate::model::cfd::Role;
@@ -315,6 +314,7 @@ pub struct TakerActorSystem<O, W> {
     pub connection_actor_addr: Address<connection::Actor>,
     pub maker_online_status_feed_receiver: watch::Receiver<ConnectionStatus>,
     wallet_actor_addr: Address<W>,
+    pub auto_rollover_addr: Address<auto_rollover::Actor<O>>,
     _tasks: Tasks,
 }
 
@@ -376,7 +376,7 @@ where
             wallet_actor_addr.clone(),
             oracle_pk,
             projection_actor.clone(),
-            process_manager_addr,
+            process_manager_addr.clone(),
             connection_actor_addr.clone(),
             oracle_addr.clone(),
             n_payouts,
@@ -385,18 +385,16 @@ where
         .create(None)
         .run();
 
-        let (auto_rollover_address, auto_rollover_fut) = auto_rollover::Actor::new(
+        let (auto_rollover_addr, auto_rollover_fut) = auto_rollover::Actor::new(
             db,
             oracle_pk,
-            projection_actor,
+            process_manager_addr,
             connection_actor_addr.clone(),
-            monitor_addr.clone(),
             oracle_addr,
             n_payouts,
         )
         .create(None)
         .run();
-        std::mem::forget(auto_rollover_address); // leak this address to avoid shutdown
 
         tasks.add(cfd_actor_fut);
         tasks.add(auto_rollover_fut);
@@ -432,6 +430,7 @@ where
             connection_actor_addr,
             maker_online_status_feed_receiver,
             wallet_actor_addr,
+            auto_rollover_addr,
             _tasks: tasks,
         })
     }
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -52,10 +52,10 @@ pub struct AcceptSettlement {
 pub struct RejectSettlement {
     pub order_id: OrderId,
 }
-pub struct AcceptRollOver {
+pub struct AcceptRollover {
     pub order_id: OrderId,
 }
-pub struct RejectRollOver {
+pub struct RejectRollover {
     pub order_id: OrderId,
 }
 pub struct Commit {
@@ -86,14 +86,14 @@ pub struct Actor<O, T, W> {
     wallet: Address<W>,
     settlement_interval: Duration,
     oracle_pk: schnorrsig::PublicKey,
-    projection_actor: Address<projection::Actor>,
-    process_manager_actor: Address<process_manager::Actor>,
+    projection: Address<projection::Actor>,
+    process_manager: Address<process_manager::Actor>,
     rollover_actors: AddressMap<OrderId, rollover_maker::Actor>,
     takers: Address<T>,
     current_order: Option<Order>,
     setup_actors: AddressMap<OrderId, setup_maker::Actor>,
     settlement_actors: AddressMap<OrderId, collab_settlement_maker::Actor>,
-    oracle_actor: Address<O>,
+    oracle: Address<O>,
     connected_takers: HashSet<Identity>,
     n_payouts: usize,
     tasks: Tasks,
@@ -106,10 +106,10 @@ impl<O, T, W> Actor<O, T, W> {
         wallet: Address<W>,
         settlement_interval: Duration,
         oracle_pk: schnorrsig::PublicKey,
-        projection_actor: Address<projection::Actor>,
-        process_manager_actor: Address<process_manager::Actor>,
+        projection: Address<projection::Actor>,
+        process_manager: Address<process_manager::Actor>,
         takers: Address<T>,
-        oracle_actor: Address<O>,
+        oracle: Address<O>,
         n_payouts: usize,
     ) -> Self {
         Self {
@@ -117,13 +117,13 @@ impl<O, T, W> Actor<O, T, W> {
             wallet,
             settlement_interval,
             oracle_pk,
-            projection_actor,
-            process_manager_actor,
+            projection,
+            process_manager,
             rollover_actors: AddressMap::default(),
             takers,
             current_order: None,
             setup_actors: AddressMap::default(),
-            oracle_actor,
+            oracle,
             n_payouts,
             connected_takers: HashSet::new(),
             settlement_actors: AddressMap::default(),
@@ -132,7 +132,7 @@ impl<O, T, W> Actor<O, T, W> {
     }
 
     async fn update_connected_takers(&mut self) -> Result<()> {
-        self.projection_actor
+        self.projection
             .send(Update(
                 self.connected_takers
                     .clone()
@@ -181,7 +181,7 @@ where
     W: 'static,
     Self: xtra::Handler<Stopping<rollover_maker::Actor>>,
 {
-    async fn handle_propose_roll_over(
+    async fn handle_propose_rollover(
         &mut self,
         RolloverProposal { order_id, .. }: RolloverProposal,
         taker_id: Identity,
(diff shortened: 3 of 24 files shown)


daemon/src/model/cfd.rs
```diff
@@ -522,26 +516,31 @@ impl Cfd {
...
         Ok(())
     }
 
-    fn can_roll_over(&self) -> bool {
-        self.lock_finality && !self.commit_finality && !self.is_final() && !self.is_attested()
+    pub fn is_rollover_possible_maker(&self) -> Result<(), CannotRollover> {
+        self.can_roll_over()
+    }
+
+    fn can_roll_over(&self) -> Result<(), CannotRollover> {
```
luckysori: 🔧 To be consistent we could call this `is_rollover_possible`. Btw, I think we have settled on `rollover` vs `roll_over`, so we should fix that anyway.

daemon/src/model/cfd.rs
```diff
@@ -591,8 +590,8 @@ impl Cfd {
     }
 
     pub fn start_rollover(&self) -> Result<(RolloverParams, Dlc, Duration)> {
-        if !self.can_roll_over() {
-            bail!("Start rollover only allowed when open")
+        if let Err(e) = self.can_roll_over() {
+            bail!(e)
```
luckysori: 🔧 I think you can just `?` here and below.

luckysori: ❓ Also, isn't it weird that internally we are using the non-actor-specific version of this API.

daemon/src/rollover_taker.rs
```diff
@@ -190,10 +193,19 @@ impl xtra::Actor for Actor {
...
+                    | CannotRollover::Final => {
+                        // TODO: This is actually somewhat incorrect. The rollover is not rejected
+                        // but not possible!  Previously we distinguished
+                        // between Rejected / NotPossible, but the Completed abstraction makes that
+                        // difficult.
+                        RolloverCompleted::Rejected {
+                            order_id: self.cfd.id(),
+                            reason: e.into(),
+                        }
+                    }
```
luckysori: I can look into this in a follow-up PR.

luckysori (approved):
LGTM

daemon/src/model/cfd.rs
da-kami: True. I was unsure, because for e.g. collab settlement we use `can`. Happy to change it to `is` for concistency here. That might need another round of cleanup at some point.

daemon/src/model/cfd.rs
da-kami: > ❓ Also, isn't it weird that internally we are using the non-actor-specific version of this API.

Both maker and taker share a set up rules, but the taker additionally checks the time since last rollover (I think the maker should not care about that). I wanted to keep the API consistent, but don't duplicate the code for the shared logic...

da-kami: Your comment made me think again. The logic is actually incorrect at the moment I think. The rollover actor should check if it is possible to roll over, similar to the collaborative settlement actor. The cfd actor should not call `is_rollover_possible` for either party. For the taker the auto-rollover actor will call the taker specific one for decision making! I'll change that.

luckysori: I think I agree!

da-kami:
@luckysori please take over :) thank you

da-kami:
@luckysori @klochowicz please add the rollover test to this PR so we see if it works :)

Once the test passes we can do a manual test (letting it run for over an hour to see rollover triggering correctly) and then merge.

daemon/src/auto_rollover.rs
```diff
@@ -73,68 +70,71 @@ where
...
-                &self.oracle_actor,
-                self.projection_actor.clone(),
-                &this,
-                (&this, &self.conn_actor),
-            )
-            .create(None)
-            .run();
+            let cfd = try_continue!(load_cfd(id, &mut conn).await);
+
+            if let Err(e) = cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
```
klochowicz: Can you check this with an event instead? I'd assume expired cfds have certain events appended to them? Otherwise, we really need to get back to mocking time to make this test work....

(additionally, both sides could check whether it got expired if we just analyse events? then at least we get rid of asymmetry, but it's a minor issue)

daemon/src/rollover_taker.rs
```diff
@@ -29,73 +27,88 @@ use futures::future;
...
         }
     }
 
-    async fn propose(&self, this: xtra::Address<Self>) -> Result<()> {
-        tracing::trace!(order_id=%self.cfd.id(), "Proposing rollover");
+    async fn propose(&mut self, this: xtra::Address<Self>) -> Result<()> {
+        let mut conn = self.db.acquire().await?;
+        let cfd = load_cfd(self.id, &mut conn).await?;
+
+        let (event, (rollover_params, dlc)) = cfd.start_rollover_taker()?;
```
klochowicz: would it be possible to only retrieve the `(rollover_params, dlc)` after acceptance? do we need it earlier?

daemon/src/auto_rollover.rs
luckysori: Because we have introduced the indirection of the `auto_rollover::Actor` sending a message to itself to spawn a `rollover_taker::Actor` we can actually directly send said message in the actor tests, circumventing this check.

daemon/src/rollover_taker.rs
luckysori: I'll take a look tomorrow!

daemon/src/auto_rollover.rs
```diff
@@ -70,39 +69,40 @@ where
...
-            if let Err(e) = cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
-                tracing::trace!(%id, "Cannot roll over: {:#}", e);
-                continue;
-            }
+                    this.send(Rollover(id)).await??;
 
-            this.send_async_safe(Rollover { id }).await?;
+                    anyhow::Ok(())
+                }
+                .await
```
luckysori: 🔧 Let's add the "Cannot roll over" context to the result of this async block, so that any of the errors which could prevent the roll over have the appropriate context.

daemon/src/auto_rollover.rs
```diff
@@ -7,7 +7,6 @@ use crate::model::cfd::OrderId;
 use crate::oracle;
```
luckysori, line 7: @da-kami Most of the changes to this file are because I used a version that @klochowicz and I worked on which is almost exactly the same as what you ended up doing, so not much to see here.

daemon/src/lib.rs
```diff
@@ -397,7 +398,6 @@ where
         )
         .create(None)
         .run();
-        std::mem::forget(auto_rollover_address); // leak this address to avoid shutdown
```
luckysori, line 399: We are no longer leaking this because we can just keep it around and use it in the tests. Any reason why leaking was chosen over storing it in the actor system, @da-kami? (You may have been part of that work).

daemon/src/maker_cfd.rs
```diff
@@ -217,6 +217,13 @@ where
 
         self.tasks.add(rollover_actor_future);
 
+        self.takers
+            .send(RollOverProposed {
+                order_id: proposal.order_id,
+                address: rollover_actor_addr.clone(),
+            })
+            .await?;
```
luckysori: @da-kami: I had to reintroduce this because we still need to tell the `maker_inc_connections::Actor` where to forward messages from the taker about the current rollover of this particular CFD. This sucks, but it may be solved down the line with QUIC's multiplexing which Thomas has been talking about.

daemon/src/model/cfd.rs
```diff
@@ -640,7 +633,7 @@ impl Cfd {
         );
 
         Ok((
-            Event::new(self.id, CfdEvent::CollaborativeSettlementProposalAccepted),
+            Event::new(self.id, CfdEvent::RolloverAccepted),
```
luckysori: Just a little copy-paste bug here.

daemon/src/model/cfd.rs
```diff
@@ -655,6 +648,14 @@ impl Cfd {
         ))
     }
 
+    pub fn handle_rollover_accepted_taker(&self) -> Result<Event> {
```
luckysori: I chose to add this for two reasons:

1. I want to be able to emit the accepted event on the taker side, so that we can update de UI.
2. I want to ensure that we only proceed with the rollover if the CFD's state still agrees with it.

daemon/src/model/cfd.rs
```diff
@@ -774,19 +775,19 @@ impl Cfd {
...
+    pub fn roll_over(self, completed: RolloverCompleted) -> Result<Event> {
         // TODO: Compare that the version that we started the rollover with is the same as the
         // version now. For that to work we should pass the version into the state machine
         // that will handle rollover and the pass it back in here for comparison.
         // Daniel says: I don't think the version check will help in addition to what we check in
         // can_rollover. What difference would it make? If we are completed rollover and are
         // still able to apply it we should, otherwise we might produce inconsistent state with the
         // other party.
-        if let Err(e) = self.can_rollover() {
+        if let (Err(e), RolloverCompleted::Succeeded { .. }) = (self.can_rollover(), &completed) {
```
luckysori: This was interesting. I noticed that we were hitting this `bail` when receiving a `RollOverCompleted::Failed` caused by failing on this very `can_rollover` check inside of a different method.

If the completion is failed or rejected, then we can safely emit the corresponding events. It's only in the case that we have succeeded in completing rollover _and_ rollover is no longer possible (e.g. because the commit transaction was found on the blockchain) that we should exit without emitting an event IMO.

klochowicz:
has the https://github.com/itchysats/itchysats/pull/966 been completely superceded by this one?

luckysori:
> has the #966 been completely superceded by this one?

Thanks for the reminder! I honestly thought that PR was already merged. I see some overlap, but I would still like to keep some of those improvements. I'll take over that one after this one merges.

daemon/src/lib.rs
da-kami: We did not store it in the actor system before because it was simply not needed (unused). But with using it in the tests it is. that is all good. I think it's fine to store it there for potential future use in production as well.

daemon/src/maker_cfd.rs
da-kami: hm... would it be possible to do this from the rollover actor instead? 
If not easily possible we can leave it here. Would it be worth adding a ticket describing that problem then?

luckysori (changes requested):
Working on applying some of the commits of #966 onto this, I've noticed that we don't do any post-processing for rollover in the `process_manager` actor yet. That will need to be fixed for roll over to actually work.

daemon/src/model/cfd.rs
da-kami: Notes: 
ad. (1): This is actually more than we did before - the taker was not informed about accept before
ad. (2): Makes sense, but the time between sending the proposal and it being accepted should (usually) be within seconds. 

I think it is nice adding this step, but I don't think it is strictly necessary. Yes, going on with rollover if the Cfd is already final might cause weird side effects. I don't think it is such a big deal though, because once we apply the rollover we check again, so it should fail then. This *is* of course dangerous, because we already exchnaged parameters with the other party then, and the other party might not have noticed finality. In the end the `Cfd` would always be final though (once lock is spent it is spent...).

That said: happy to keep it, not sure if the state "helps" the user's understanding though. But that can be fixed by not showing fine-granular states in the UI.
If we want a separate state in the UI we might have to extend the state mapping in the projection actor.

daemon/src/model/cfd.rs
da-kami: Oh my, yes! No checks needed for failed/rejected. Good catch!

da-kami:
> Working on applying some of the commits of #966 onto this, I've noticed that we don't do any post-processing for rollover in the `process_manager` actor yet. That will need to be fixed for roll over to actually work.

Oh my, yes! We have to properly trigger monitoring again.

daemon/src/rollover_taker.rs
luckysori: Good catch, I think it makes a lot more sense.

daemon/src/maker_cfd.rs
luckysori: I think moving it to the rollover actor makes more sense, let me try.

daemon/src/model/cfd.rs
luckysori: > ad. (2): Makes sense, but the time between sending the proposal and it being accepted should (usually) be within seconds.

That is true, but this snippet should allow the user to know what's going on after the other party _accepts_. It should transition to `ContractSetup` (we could eventually introduce a dedicated state). Otherwise the UI would still show `Sent` or whatever we're doing atm.

From your comments I think we're okay to keep this, but thanks for thinking it through.

da-kami (commented):
Approved!

luckysori (approved):

daemon/src/maker_cfd.rs
```diff
@@ -205,7 +204,7 @@ where
         let mut conn = self.db.acquire().await?;
         let cfd = load_cfd(proposal.order_id, &mut conn).await?;
 
-        cfd.is_rollover_possible(OffsetDateTime::now_utc())?;
+        cfd.is_rollover_possible_maker()?;
```
thomaseizinger: This very much looks like a pre-check of a command that the maker should be running against its CFD.

daemon/src/projection.rs
```diff
@@ -428,26 +414,22 @@ impl Cfd {
...
             RevokeConfirmed => todo!("Deal with revoked"),
+            RolloverStarted { .. } => match role {
+                Role::Maker => (
+                    CfdState::IncomingRollOverProposal,
+                    vec![CfdAction::AcceptRollOver, CfdAction::RejectRollOver],
+                ),
+                Role::Taker => (CfdState::OutgoingRollOverProposal, vec![]),
+            },
+            RolloverAccepted => {
+                // TODO: Might wanna introduce a separate state there...?
```
klochowicz: Indeed - especially if we decide to allow collab settlement (ie. closing the cfd) during the rollover phase after all, as @da-kami  suggested. 
It's an UI-only state, so that should be trivial to add (just aligning the frontend to handle it correctly).
can be done later though.

daemon/src/model/cfd.rs
```diff
@@ -1816,4 +1822,345 @@ mod tests {
...
+                        dlc: Dlc::dummy(Some(event_id)),
+                    },
+                },
+                Event {
+                    timestamp: Timestamp::now(),
+                    id: Default::default(),
+                    event: CfdEvent::LockConfirmed,
+                },
+            ]
+        }
```
thomaseizinger, line 2072: This is a great first step!

One advantage of an event-sourcing based system is that we can invest some effort into a test harness that constructs us the different events. I.e. if we change this to something like:

```rust
impl Event {
	fn dummy_contract_setup_started() {}
	fn dummy_contract_setup_completed() {}
	// etc
}
```

Then the code for `Cfd::dummy_open` could read like:

```rust
[
	Event::dummy_contract_setup_started(),
	Event::dummy_contract_setup_completed(),
]
.into_iter()
.fold(cfd, Cfd::apply)
```

Which reduces a layer of indirection and makes these events more reusable for other tests! Reusing the same events might help us test stuff within `monitor` for example!

daemon/src/model/cfd.rs
```diff
@@ -211,14 +211,20 @@ pub enum SettlementKind {
     Outgoing,
 }
 
+#[derive(thiserror::Error, Debug, PartialEq)]
+pub enum CannotAutoRollover {
+    #[error("Is too recent to auto-rollover")]
+    TooRecent,
+    #[error("Cfd does not have a dlc")]
+    NoDlc,
+}
```
thomaseizinger: If we don't end up sending these errors around as values, I think I wouldn't bother modelling them!

daemon/src/rollover_taker.rs
```diff
@@ -128,9 +117,9 @@ impl Actor {
             }),
             receiver,
             (self.oracle_pk, announcement),
-            rollover_params,
+            todo!("plug in rollover params from self once we have them"),
```
thomaseizinger: Not particularly clean having these in here within the commit. Not sure if you plan on squashing?

daemon/src/rollover_maker.rs
```diff
@@ -256,6 +238,54 @@ impl Actor {
...
+        };
+
+        if let Err(error) = fut.await {
+            self.complete(Completed::Failed { order_id, error }, ctx)
+                .await;
+        }
+    }
+
+    async fn stopping(&mut self, ctx: &mut Context<Self>) -> KeepRunning {
+        let address = ctx.address().expect("acquired own actor address");
```
klochowicz: nit: `"self to be alive" , just like in other case above?

daemon/src/auto_rollover.rs
```diff
@@ -21,12 +23,12 @@ use xtra::Actor as _;
 use xtra::Address;
 use xtra_productivity::xtra_productivity;
 
-pub struct Actor<O, M> {
+pub struct Actor<O> {
     db: sqlx::SqlitePool,
     oracle_pk: schnorrsig::PublicKey,
     projection_actor: Address<projection::Actor>,
+    process_manager_actor: Address<process_manager::Actor>,
```
thomaseizinger: Nit: Would prefer the naming to omit `actor` for brevity.

daemon/src/auto_rollover.rs
```diff
@@ -127,36 +128,49 @@ where
...
+
+        let event = cfd.roll_over(rollover_completed)?;
+        append_event(event.clone(), &mut conn).await?;
+
+        if let Err(e) = self
+            .process_manager_actor
+            .send(process_manager::Event::new(event.clone()))
+            .await?
+        {
+            tracing::error!("Sending event to process manager failed: {:#}", e);
```
thomaseizinger: My current view is that the `ProcessManager` failing is a hard error and should always abort the current function. If the caller wants to only log that result, that is up to them!

daemon/src/model/cfd.rs
```diff
@@ -737,22 +738,24 @@ impl Cfd {
...
-    // TODO: Pass the entire enum
-    pub fn roll_over(self, rollover_result: Result<Dlc>) -> Result<Event> {
+    pub fn roll_over(self, rollover_completed: RolloverCompleted) -> Result<Event> {
         // TODO: Compare that the version that we started the rollover with is the same as the
         // version now. For that to work we should pass the version into the state machine
         // that will handle rollover and the pass it back in here for comparison.
+        // Daniel says: I don't think the version check will help in addition to what we check in
+        // can_rollover. What difference would it make? If we are completed rollover and are
+        // still able to apply it we should, otherwise we might produce inconsistent state with the
+        // other party.
```
thomaseizinger: Looking forward to the day where we have entire conversations incode  comments :rofl:

daemon/tests/happy_path.rs
```diff
@@ -191,6 +191,43 @@ async fn force_close_an_open_cfd() {
...
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert_eq!(taker_cfd.state, CfdState::OutgoingRollOverProposal);
+    assert_eq!(maker_cfd.state, CfdState::IncomingRollOverProposal);
+
+    maker.accept_rollover_proposal(order_id).await;
+
+    assert_next_state!(CfdState::ContractSetup, maker, taker, order_id);
+    assert_next_state!(CfdState::Open, maker, taker, order_id);
+}
```
klochowicz, line 238: in a follow-up, we should really finish this test by force-closingw this cfd using a different dlc & event id, to confirm that the internals have changed (for that we'll need to extend the test framework slightly, but that should be quite easy :) )

klochowicz (approved):
great job!!! really excited to see rollover working 🎉

daemon/src/model/cfd.rs
thomaseizinger: Thomas says: For now I would make it a hard requirement to not allow any concurrency within these commands. If we find this restrictive, we can always relax this later. If the version we started rollover with is not the same as here, all kinds of state changes could have happened and reasoning about this is **not** trivial. Best to fail early here.

daemon/src/model/cfd.rs
```diff
@@ -598,13 +599,48 @@ impl Cfd {
...
+                    self.leverage,
+                    self.refund_timelock_in_blocks(),
+                    1, // TODO: Where should I get the fee rate from?
+                ),
+                self.dlc.as_ref().ok_or(CannotRollover::NoDlc)?.clone(),
+            ),
+        ))
+    }
+
+    pub fn receive_rollover_proposal(self, proposal: RolloverProposal) -> Result<Event> {
```
thomaseizinger: I am liking this!

daemon/src/rollover_maker.rs
```diff
@@ -47,102 +47,94 @@ pub struct RolloverFailed {
...
+
+        match event_fut.await {
+            Ok(event) => {
+                let _ = self
+                    .process_manager
+                    .send(process_manager::Event::new(event))
+                    .await;
+            }
+            Err(e) => {
+                tracing::warn!(%order_id, "Failed to report completion of collab settlement: {:#}", e)
```
thomaseizinger: Copy paste error in log message :)

daemon/tests/happy_path.rs
```diff
@@ -191,6 +191,43 @@ async fn force_close_an_open_cfd() {
...
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement()).await;
+
+    taker.trigger_rollover(order_id).await;
+
+    let (taker_cfd, maker_cfd) = next_cfd(taker.cfd_feed(), maker.cfd_feed()).await.unwrap();
+    assert_eq!(taker_cfd.state, CfdState::OutgoingRollOverProposal);
+    assert_eq!(maker_cfd.state, CfdState::IncomingRollOverProposal);
```
thomaseizinger: With a bit more macro trickery to accept different syntaxes, we should be able to create a version of `assert_next_state` that allows us to assert on different events for maker / taker.

See https://doc.rust-lang.org/rust-by-example/macros/overload.html for example.

daemon/src/process_manager.rs
```diff
@@ -163,14 +163,29 @@ impl Actor {
 
                 tracing::info!(%txid, "Commit transaction published");
             }
+            RolloverCompleted { dlc } => {
+                tracing::info!("Rollover complete, monitoring new txs and oracle events");
```
thomaseizinger: I think we should split this into:

1. Logging here that rollover completed
2. Logging in `monitor` that we are starting to monitor a new DLC
3. Logging in `oracle` that we are starting to monitor a new attestation

daemon/src/rollover_maker.rs
```diff
@@ -98,7 +98,7 @@ impl Actor {
         let mut conn = self.db.acquire().await?;
         let cfd = load_cfd(self.proposal.order_id, &mut conn).await?;
 
-        let event = cfd.receive_rollover_proposal(self.proposal.clone())?;
```
thomaseizinger: I think this highlights how useless the data in `RolloverProposal` currently is :joy:

thomaseizinger (commented):
Stellar work, some non-blocking comments!

daemon/src/rollover_maker.rs
luckysori: I always grep the codebase and copy paste the first hit lol.

daemon/tests/happy_path.rs
luckysori: Agreed as discussed earlier. I'll open an issue once this merges.

daemon/src/maker_cfd.rs
luckysori: Good point, I think we need to remove this.

daemon/src/model/cfd.rs
da-kami: Definitely! The original purpose of this PR was to get the unit tests for auto-rollover back in and I tried to do that in the fastest way. 
I think we should generally invest time into a test harness for unit tests for `Cfd`. What we do for auto-rollover could and should be done for other features as well! We can unit test the complete event logic in the `Cfd`. 

I'd propose to discuss this together once you are back and ramp up a simple harness that can then be steadily extended (simliar to how we handled the actor tests...).

daemon/src/rollover_taker.rs
luckysori: I think the idea is to squash most of this, yep.

daemon/src/auto_rollover.rs
luckysori: Yeah, I agree with you.

daemon/src/model/cfd.rs
luckysori: I will do a version check here.

daemon/src/model/cfd.rs
da-kami: We went back and forth with these "cases" (`TooRecent` is not even a real error case, because you just can't rollover but there is nothing erroneous about that...).

For me, returning specific error types greatly helps the understanding of the logic, because I immediately see what is the problem. Furthermore, the caller can decide to react differently. Maybe there is a better way of modelling this, that does not actually build on errors. I would not invest the time to change them know though.

daemon/src/rollover_taker.rs
da-kami: @luckysori is planning to squash all the commits together. We kept them for now to make reviewing easier. We should have probably stated that clearer 😬

daemon/src/model/cfd.rs
luckysori: Good to get some more confirmation, because I wasn't sure if these state checking methods that just emit events were what we wanted necessarily.

daemon/tests/happy_path.rs
luckysori: I briefly added something like what you suggest, but less elegant. I'll take a look at that link!

daemon/src/model/cfd.rs
da-kami: Daniel says: Not sure it's worth it, you will have to pull the version through into the rollover actor and whatnot... But feel free to do it :D

🤪

daemon/src/rollover_maker.rs
luckysori: Yep, we want to get rid of the timestamp. The order id is necessary at some level, but maybe the `RolloverProposal` itself can omit it?

da-kami: Depends on what level. When sending from taker to maker we must include the id. 
But within the context for a rollover actor or the `Cfd` we know the id anyway...

daemon/src/maker_cfd.rs
luckysori: Okay, this was removed and done the way you suggest here.

daemon/src/rollover_maker.rs
thomaseizinger: - On the wire we certainly need the ID.
- Once we loaded the CFD from the DB, a parameter-less command should be enough to move forward.

daemon/tests/happy_path.rs
luckysori: See https://github.com/itchysats/itchysats/issues/1036.

daemon/src/projection.rs
luckysori: See https://github.com/itchysats/itchysats/issues/1037.

daemon/src/model/cfd.rs
luckysori: I will contradict myself. Let's introduce this version check (or not) in a separate PR. Here's an issue where we can continue to discuss this particular case: https://github.com/itchysats/itchysats/issues/1038.

daemon/src/model/cfd.rs
luckysori: I'm keeping them around for the time being.

daemon/src/model/cfd.rs
```diff
@@ -982,11 +1089,22 @@ impl Cfd {
                 // TODO: Deal with failed contract setup
                 self.during_contract_setup = false;
             }
+            RolloverStarted => {
+                self.during_rollover = true;
+            }
+            RolloverAccepted => {} /* TODO: It's weird we have no state change for applying an */
```
luckysori: @thomaseizinger @da-kami: I'm removing this TODO because I hate all code TODOs, but the point stands. Is this a bad sign? The `RolloverAccepted` was only introduced because a different aggregate would care (in `projection`). Maybe that's fine.

daemon/src/model/cfd.rs
```diff
@@ -982,11 +1089,22 @@ impl Cfd {
...
+            // event...
             RolloverCompleted { dlc } => {
                 self.dlc = Some(dlc);
+                self.during_rollover = false;
+            }
+            RolloverFailed { .. } => {
+                // TODO: Deal with failed rollover
```
luckysori: @da-kami: Also removing this TODO. What are we not doing that we should be doing? Then I can open an issue.

luckysori:
@thomaseizinger I'm going ahead with the merge if bors allows. The main bit of clean-up that I see is taking care of this: https://github.com/itchysats/itchysats/pull/1031#discussion_r779313915

PR #1018, opened 2022-01-05
Add more logs around sending and processing heartbeats

klochowicz:
Hopefully this will give more insight into heartbeat behaviour on the CI

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -27,6 +27,7 @@ use futures::TryStreamExt;
 use std::net::SocketAddr;
 use std::time::Duration;
 use std::time::SystemTime;
+use time::OffsetDateTime;
 use tokio::net::TcpStream;
 use tokio::sync::watch;
 use tokio_util::codec::Framed;
@@ -95,6 +96,13 @@ impl State {
         *self = State::Disconnected;
 
         true
+    }
+
+    fn last_heartbeat(&self) -> Option<OffsetDateTime> {
+        match self {
+            State::Connected { last_heartbeat, .. } => Some((*last_heartbeat).into()),
+            State::Disconnected => None,
+        }
     }
 }
 
@@ -483,10 +491,17 @@ impl Actor {
     }
 
     fn handle_measure_pulse(&mut self, _: MeasurePulse) {
+        tracing::trace!(target: "wire", "measuring heartbeat pulse");
+
         if self
             .state
             .disconnect_if_last_heartbeat_older_than(self.heartbeat_timeout)
         {
+            tracing::warn!(
+                "Disconnecting due to lack of heartbeat. Last heartbeat: {:?}",
+                self.state.last_heartbeat()
+            );
+
             self.status_sender
                 .send(ConnectionStatus::Offline { reason: None })
                 .expect("watch receiver to outlive the actor");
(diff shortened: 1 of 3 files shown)


daemon/src/connection.rs
```diff
@@ -483,10 +491,17 @@ impl Actor {
     }
 
     fn handle_measure_pulse(&mut self, _: MeasurePulse) {
+        tracing::trace!("measuring pulse");
+
```
luckysori: 🔧 I guess you didn't want to commit this.

daemon/src/connection.rs
```diff
@@ -483,10 +491,17 @@ impl Actor {
...
         if self
             .state
             .disconnect_if_last_heartbeat_older_than(self.heartbeat_timeout)
         {
+            tracing::debug!(
```
luckysori: ❓ Should we promote this to `warn` or `info`? It seems like quite a relevant and hopefully not too common event to disconnect from a peer. Or would it be too noisy?

daemon/src/maker_inc_connections.rs
```diff
@@ -370,6 +370,10 @@ impl Actor {
 
             async move {
                 while let Ok(()) = this.send(SendHeartbeat(identity)).await {
+                    tracing::debug!(
```
luckysori: 🔧 I would set this to `trace` level.

luckysori (approved):
Thanks!

daemon/src/maker_inc_connections.rs
```diff
@@ -370,6 +370,10 @@ impl Actor {
...
+                        "Sent heartbeat message. Sleeping for {}",
+                        heartbeat_interval.as_secs()
+                    );
                     tokio::time::sleep(heartbeat_interval).await;
                 }
             }
```
da-kami, line 381: I'd suggest to add a trace after the loop to see if the maker ever runs into scenarios where the maker fails to send a heartbeat to the taker and stops. This might be useful to see as well.

daemon/src/connection.rs
da-kami: Let's make it a `warn` for now - if it spams we can remove / downgrade it again. Maybe this one should stay without the need for debugging.

da-kami (approved):

daemon/src/connection.rs
klochowicz: changed to `warn`

daemon/src/connection.rs
klochowicz: thanks for doublechecking, but I wanted to commit this - it's only at trace level, so no impact on production :)

daemon/src/maker_inc_connections.rs
klochowicz: great idea, added!

klochowicz: I added it as a `warn`, as this should not happen during normal operation.

daemon/src/connection.rs
klochowicz: (I have added it as a `wire` target though, so it's a bit more explicit - I need it for debugging)

luckysori: Fair enough! I assumed it was temporary because it's all lowercase :P

klochowicz: I can see that! this is probably good heuristic! I try to remind myself with temporary `FIXME` or `TODO` comments if I add something temporarily, as magit-todos tells me I have some new ones introduced on my branch!

Issue #226, opened 2021-10-06
Improve compile-times of workspace


thomaseizinger:
With more code being added, we need to make sure our workspace compiles fast to keep productivity levels up.

There are several things that can be done. [This](https://matklad.github.io/2021/09/04/fast-rust-builds.html) blogpost summarizes these nicely.

thomaseizinger:
Sketch what the workspace could look like: 

![image](https://user-images.githubusercontent.com/5486389/147739484-f46b063f-f6a8-4953-8c99-d1fe244bfdbb.png)

<details>
<summary>Diagram source</summary>

```plantuml
@startuml

title Worksapce structure

component model
component makerStandalone
component maker
component db
component takerUmbrel
component takerStandalone
component taker
component xtras
component rocketBasicAuth
component tests
component core

makerStandalone -Down-> maker
maker -Down-> core
taker -Down-> core
core -Down-> xtras
core -Down-> model
core -Down-> db
makerStandalone -Up-> rocketBasicAuth
takerStandalone -Up-> rocketBasicAuth
takerUmbrel -Up-> rocketBasicAuth
takerUmbrel -Down-> taker
takerStandalone -Down-> taker
tests -Down-> maker
tests -Down-> taker

note right of maker: Library crate of shared code across maker binaries (actor system)
note left of taker: Library crate of shared code across taker binaries (actor system)
note top of tests: Crate of integration tests between maker and taker lib
note right of core: Shared actors
note left of model: Cfd model, payout curve, etc
note right of xtras: Extensions to xtra

@enduml
```

</details>

cc @klochowicz @luckysori You might be interested in this!

PR #1019, opened 2022-01-05
Test release build on the CI

klochowicz:
Verify whether the actor tests on the CI fail the same way on the release build.

I also wanted to check whether it's feasible at all (how much time will it take to build using github).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -116,6 +116,19 @@ jobs:
             target/debug/maker
             target/debug/taker
 
+  test_daemons_release_mode:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2.4.0
+      - name: Setup rust toolchain
+        run: rustup show
+      - uses: Swatinem/rust-cache@v1.3.0
+      - run: cargo build --release --bins --tests
+      - run: cargo test --release --workspace
+
   daemons_arm_build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,6 @@ jobs:
       - name: Setup rust toolchain
         run: rustup show
       - uses: Swatinem/rust-cache@v1.3.0
-      - run: cargo build --release --bins --tests
       - run: cargo test --release --workspace
 
   daemons_arm_build:


klochowicz:
I managed to trim it to ~9mins, not sure whether that is acceptable or not. Opening the floor for discussions :)

thomaseizinger:
I think you could easily make this part of the existing matrix to avoid the duplicate declaration.

However, I am not sure we need this specifically?

klochowicz:
> I think you could easily make this part of the existing matrix to avoid the duplicate declaration.
> 
> However, I am not sure we need this specifically?

Trying to avoid one more declaration would result in longer build times - we don't need to do another smoke test etc. This trade-off seemed more worthy IMHO.

The idea behind it was being able to verify that some of our failures in CI *only* ever happen in debug mode. This is as such an useful data point when considering whether something has an impact on the released version - in general, it makes sense to also test the version that the users are actually using).

The counter-argument was that this would significantly increase build times on the CI - this PR was submitted as draft to analyse by how much we'd increase it. From looking at preliminary results, it's almost the same times.

thomaseizinger:
> Trying to avoid one more declaration would result in longer build times - we don't need to do another smoke test etc. This trade-off seemed more worthy IMHO.

How much longer would the build times be? We should be able to cut the build times down by splitting out crates. See https://github.com/itchysats/itchysats/issues/226.

> This is as such an useful data point when considering whether something has an impact on the released version - in general, it makes sense to also test the version that the users are actually using).

Shouldn't we be doing a smoke-test then? :)

thomaseizinger:
Overall, I'd appreciate if we use GitHub's build-matrix as much as possible. It would also be nice to not special case the arm build and use a self-hosted GitHub actions runner that uses an ARM cloud machine to run the tests.

luckysori (approved):

thomaseizinger (approved):

thomaseizinger:
I opened https://github.com/itchysats/itchysats/issues/1096.

thomaseizinger:
If you want these jobs to be required for bors to merge the PR, make sure you edit `bors.toml` accordingly.

klochowicz:
> > Trying to avoid one more declaration would result in longer build times - we don't need to do another smoke test etc. This trade-off seemed more worthy IMHO.
> 
> How much longer would the build times be? We should be able to cut the build times down by splitting out crates. See #226.
> 
> > This is as such an useful data point when considering whether something has an impact on the released version - in general, it makes sense to also test the version that the users are actually using).
> 
> Shouldn't we be doing a smoke-test then? :)

Smoke test is unlikely to give us any difference - the debug builds would differ mainly with optimisations used and lack of debug assertions etc. I don't mind having one for correctness.

PR #1020, opened 2022-01-05
Restore pending orders in the maker UI

klochowicz:
Seems like a tiny regression slipped in (I tried to set up a cfd locally and it was impossible)


luckysori (approved):
I believe you 🤣

da-kami (approved):
Interesting, I wonder when this slipped in 😅 
Happy to merge if it fixes things.

PR #1027, opened 2022-01-06
Align heartbeat variables in the taker

klochowicz:
Both maker and taker actor systems are now constructed with the same heartbeat
interval variable.

Taker internally configures the heartbeat timeout as 2 * heartbeat_interval.
The heartbeat pulse (rate of checking whether timeout lapsed) is configured as
heartbeat_interval / 2. Beforehand, it was set to the same value as heartbeat
timeout, which could result in reporting a timeout even if it should not do so.

Adds more comments to describe all variables in use for future readers.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -93,6 +93,11 @@ impl State {
             return false;
         }
 
+        tracing::warn!(
+            "Disconnecting due to lack of heartbeat. Last heartbeat: {:?}",
+            self.last_heartbeat()
+        );
+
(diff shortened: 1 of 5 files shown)


da-kami:
Note: The current test failure is unrelated. @klochowicz is working on another quickfix for that.

da-kami (approved):

klochowicz:
> Note: The current test failure is unrelated. @klochowicz is working on another quickfix for that.

I've rebased after the other one got merged. CI should pass with flying colours now :)

PR #1029, opened 2022-01-06
Increate timeout on the protocol messages

klochowicz:
Prevents one class of protocol error bugs

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -54,6 +54,9 @@ use std::ops::RangeInclusive;
 use std::time::Duration;
 use xtra::prelude::MessageChannel;
 
+/// How long protocol waits for the next message before giving up
+const MSG_TIMEOUT: Duration = Duration::from_secs(60);
+
 pub struct SetupParams {
     margin: Amount,
     counterparty_margin: Amount,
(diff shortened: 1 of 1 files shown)


daemon/src/setup_contract.rs
```diff
@@ -54,6 +54,10 @@ use std::ops::RangeInclusive;
 use std::time::Duration;
 use xtra::prelude::MessageChannel;
 
+/// How long protocol waits for the next message before giving up
+// TODO: Move back to 60s after issue of actor test failures has been resolved.
```
da-kami:
```suggestion
// TODO: Analyse why event sourcing refactor made us change the timeout from 60 to 70 secs. Why does it take so much longer now?
```

Let's see if the 70 secs plaster is going to make a difference. If so, we might keep them for the time being given the current code. 
Not sure we will then change them back to 60. If the 70 secs don't help we will have to analyse why it keeps failing.

da-kami (approved):

daemon/src/setup_contract.rs
klochowicz: yeah, I still think we should analyse that, but there are more pressing issues (such as 0.4 release :) ). Hopefully this will do in the meantime - I used the first sentence of your suggestion, thanks!

daemon/src/setup_contract.rs
```diff
@@ -131,9 +135,9 @@ pub async fn new(
         .context("Failed to send Msg0")?;
     let msg0 = stream
         .select_next_some()
-        .timeout(Duration::from_secs(60))
+        .timeout(MSG_TIMEOUT)
         .await
-        .context("Expected Msg0 within 60 seconds")?
+        .context(format_expect_msg_within("Msg0"))?
```
thomaseizinger, line 140: Note: This will allocate but only use the message in case of an error. To avoid the allocation in the happy path, you should use `.with_context` every time you need to allocate to construct the error message.

klochowicz: thanks, I'll submit a small fixup for it!

PR #1030, opened 2022-01-06
Remove unused handler in `maker_cfd.rs`

luckysori:
The `maker_cfd::Actor` is no longer informed about the completion of the collaborative settlement protocol, so this handler is unused.


klochowicz (approved):
nice find!

PR #1031, opened 2022-01-06
Feature/rollover event sourcing cherry picked

luckysori:
This is most of the work that @thomaseizinger did in #966, but applied on top of #1014.

I tried to throw away anything that was no longer relevant.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -212,13 +212,19 @@ pub enum SettlementKind {
 }
 
 #[derive(thiserror::Error, Debug, PartialEq)]
+pub enum CannotAutoRollover {
+    #[error("Is too recent to auto-rollover")]
+    TooRecent,
+    #[error("Cfd does not have a dlc")]
+    NoDlc,
+}
+
+#[derive(thiserror::Error, Debug, PartialEq)]
 pub enum CannotRollover {
     #[error("Cfd does not have a dlc")]
     NoDlc,
     #[error("The Cfd is already expired")]
     AlreadyExpired,
-    #[error("The Cfd was just rolled over")]
-    WasJustRolledOver,
     #[error("Cannot roll over when CFD not locked yet")]
     NotLocked,
     #[error("Cannot roll over when CFD is committed")]
@@ -258,10 +264,12 @@ pub enum CfdEvent {
     ContractSetupFailed,
     OfferRejected,
 
+    RolloverStarted,
+    RolloverAccepted,
+    RolloverRejected,
     RolloverCompleted {
         dlc: Dlc,
     },
-    RolloverRejected,
     RolloverFailed,
 
     CollaborativeSettlementStarted {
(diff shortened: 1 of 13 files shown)


daemon/src/model/cfd.rs
```diff
@@ -775,32 +824,60 @@ impl Cfd {
...
+                    | RolloverError::NotRollingOver
+                    | RolloverError::Committed
+                    | RolloverError::Attested
+                    | RolloverError::Final
+                    | RolloverError::AlreadyExpired),
+                ..
+            } => {
+                tracing::debug!(order_id = %self.id, "Rollover was not started: {:#}", error);
+                return Ok(None);
+            }
```
luckysori, line 856: I'm not sure I understand the note, given that this is applied on top of the work that hopefully fixes #958, yet we still need to handle all these errors. What was your idea?

daemon/src/model/cfd.rs
```diff
@@ -215,15 +217,19 @@ pub enum SettlementKind {
...
+    #[error("CFD does not have a DLC")]
     NoDlc,
 }
 
-#[derive(thiserror::Error, Debug, PartialEq)]
-pub enum CannotRollover {
-    #[error("Cfd does not have a dlc")]
+/// Various error cases that can happen during rollover.
+///
+/// This enum is expected to go away once we handle the entire protocol within the actor and not
```
da-kami, line 226: I think some of the enum would still remain for defining the error cases that are returned by the rollover function on the `Cfd` - or are we planning to remove this completely then?

da-kami: On that note: It might actually be worth distinguishing `CannotRollover` from `RolloverError` because it's two different concerns. Could potentially map `CannotRollover` into `RolloverError`. (NOt sure it's worth it. Can also keep as is.)

daemon/src/rollover_taker.rs
```diff
@@ -200,30 +208,37 @@ impl Actor {
...
+                tracing::warn!(%order_id, "Failed to complete rollover: {:#}", e)
             }
         }
 
         ctx.stop();
     }
+
+    /// Returns whether the maker has accepted our rollover proposal.
+    fn is_accepted(&self) -> bool {
+        self.rollover_msg_sender.is_some()
```
da-kami, line 228: ❓How is `accepted` ensured by the message *sender* being `Some`? I would have expected the message receiver being `Some` here, but that might just be a lack of understanding. Would be great to at least explain the reasoning in a comment.

da-kami (approved):
Thanks for updating this to get the changes in!

This question might need attention: https://github.com/itchysats/itchysats/pull/1031#discussion_r779279928

daemon/src/model/cfd.rs
thomaseizinger: The idea was that we would not emit a started event for cases like "was just rolled over" so really, there shouldn't be a case within the error of the completed handler.

A not completely thought through idea I had once is that with these actors operating on the CFD directly, creating one command per outcome might be better than wrapping it in an enum?

Also, the entirety of this explictely modelled error might be unnecessary because we no longer need to transport the error as a value across actors so the command can just fail and we log the error for cases where we don't produce an event here.

Hope that makes sense, sorry for the breviety.

daemon/src/rollover_taker.rs
```diff
@@ -348,16 +383,24 @@ pub struct RollOverRejected;
...
     error: anyhow::Error,
 }
 
+/// Message sent from the spawned task to `rollover_taker::Actor` to
+/// notify that the timeout has been reached.
+///
+/// It is up to the actor to reason whether or not the protocol has progressed since then.
+struct MakerResponseTimeoutReached {
+    timeout: Duration,
+}
```
thomaseizinger, line 402: This change is likely among the best to pick out of this branch. Not sure if many of the others are worth it!

daemon/src/rollover_taker.rs
```diff
@@ -166,24 +173,25 @@ impl Actor {
...
+        self.tasks.add(async move {
+            // Use an explicit type annotation to cause a compile error if someone changes the
+            // handler.
+            let _: Result<(), Disconnected> = match rollover_fut.await {
+                Ok(dlc) => this.send(RolloverSucceeded { dlc }).await,
+                Err(error) => this.send(RolloverFailed { error }).await,
             };
-
-            Ok(())
         });
```
thomaseizinger, line 183: I think this one also has value :)

luckysori:
Changes already applied on top of #1014.

PR #1032, opened 2022-01-06
Small follow-up fixes for #1029

klochowicz:
Address feedback by @thomaseizinger & fix an omission that I noticed :)


thomaseizinger (approved):
Awesome, thanks!

PR #1040, opened 2022-01-07
Add output index to mempool.space URLs when available

### LINKED ISSUE #799: Links to `mempool.space` should indicate output index
On mempool.space it is possible to point to a specific output index by appending `:{INDEX}` to the URL.

See here: https://mempool.space/testnet/tx/0de71f7ad590b930887a872ccdf58e290328c94d4ab560c3d239e1bcc819204a:0

This would make it clearer, which of the outputs our payout is, for example.

klochowicz:
If we can find the relevant tx output by the script pubkey, then use the output
index in the URL.

Fall-back gracefully to txid URL to satisfy actor tests (which don't have signed
transactions).

Fixes #799

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -23,8 +23,12 @@ use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::Amount;
 use bdk::bitcoin::Network;
+use bdk::bitcoin::Script;
 use bdk::bitcoin::SignedAmount;
+use bdk::bitcoin::Transaction;
 use bdk::bitcoin::Txid;
+use bdk::miniscript::DescriptorTrait;
+use maia::TransactionExt;
 use rust_decimal::Decimal;
 use serde::Deserialize;
 use serde::Serialize;
@@ -261,11 +265,17 @@ impl Cfd {
                 (CfdState::ContractSetup, vec![])
             }
             ContractSetupCompleted { dlc } => {
-                self.details.tx_url_list.push(TxUrl::new(
-                    dlc.lock.0.txid(),
+                // We can highlight the shared output, but cannot discern between our change outputs
+                // and the other party's.
+                let (lock_tx, lock_desc) = &dlc.lock;
+
+                let tx_url = TxUrl::from_transaction(
+                    lock_tx,
+                    &lock_desc.script_pubkey(),
                     network,
                     TxLabel::Lock,
-                ));
+                );
+                self.details.tx_url_list.push(tx_url);
                 self.latest_dlc = Some(dlc);
 
                 (CfdState::PendingOpen, vec![])
@@ -302,10 +312,13 @@ impl Cfd {
                 (CfdState::IncomingSettlementProposal, vec![])
             }
             CollaborativeSettlementCompleted {
-                spend_tx, price, ..
+                spend_tx,
+                price,
+                script,
             } => {
-                self.details.tx_url_list.push(TxUrl::new(
-                    spend_tx.txid(),
+                self.details.tx_url_list.push(TxUrl::from_transaction(
+                    &spend_tx,
+                    &script,
                     network,
                     TxLabel::Collaborative,
                 ));
@@ -349,8 +362,9 @@ impl Cfd {
             CetConfirmed => (CfdState::Closed, vec![]),
             RefundConfirmed => {
                 if let Some(dlc) = self.latest_dlc.as_ref() {
-                    self.details.tx_url_list.push(TxUrl::new(
-                        dlc.refund.0.txid(),
+                    self.details.tx_url_list.push(TxUrl::from_transaction(
+                        &dlc.refund.0,
+                        &dlc.script_pubkey_for(role),
                         network,
                         TxLabel::Refund,
                     ));
@@ -380,9 +394,17 @@ impl Cfd {
                 (CfdState::PendingCommit, vec![])
             }
             OracleAttestedPostCetTimelock { cet, price } => {
-                self.details
-                    .tx_url_list
-                    .push(TxUrl::new(cet.txid(), network, TxLabel::Cet));
+                let tx_url = if let Some(dlc) = self.latest_dlc.as_ref() {
+                    TxUrl::from_transaction(
+                        &cet,
+                        &dlc.script_pubkey_for(role),
+                        network,
+                        TxLabel::Cet,
+                    )
+                } else {
+                    TxUrl::new(cet.txid(), network, TxLabel::Cet)
+                };
+                self.details.tx_url_list.push(tx_url);
 
                 let (profit_btc, profit_percent) = self.maybe_calculate_profit(price);
                 self.profit_btc = profit_btc;
@@ -678,12 +700,6 @@ mod round_to_two_dp {
     }
 }
 
-#[derive(Debug, Clone, Serialize)]
-pub struct TxUrl {
-    pub label: TxLabel,
-    pub url: String,
-}
-
 /// Construct a mempool.space URL for a given txid
 pub fn to_mempool_url(txid: Txid, network: Network) -> String {
     match network {
@@ -694,16 +710,46 @@ pub fn to_mempool_url(txid: Txid, network: Network) -> String {
     }
 }
 
+/// Link to transaction on mempool.space for UI representation
+#[derive(Debug, Clone, Serialize)]
+struct TxUrl {
+    pub label: TxLabel,
+    pub url: String,
+}
+
 impl TxUrl {
-    pub fn new(txid: Txid, network: Network, label: TxLabel) -> Self {
+    fn new(txid: Txid, network: Network, label: TxLabel) -> Self {
         Self {
             label,
             url: to_mempool_url(txid, network),
         }
     }
+
+    /// Highlight particular transaction output in the TxUrl
+    fn with_output_index(mut self, index: u32) -> Self {
+        self.url.push_str(&("#".to_string() + &index.to_string()));
+        self
+    }
+
+    /// If the Transaction contains the script_pubkey, output will be selected
+    /// in the URL. Otherwise, fall back to the main txid URL.
+    fn from_transaction(
+        transaction: &Transaction,
+        script_pubkey: &Script,
+        network: Network,
+        label: TxLabel,
+    ) -> Self {
+        debug_assert!(label != TxLabel::Commit, "commit transaction has a single output which does not belong to either party - this won't highlight anything");
+        let tx_url = Self::new(transaction.txid(), network, label);
+        if let Ok(outpoint) = transaction.outpoint(script_pubkey) {
+            tx_url.with_output_index(outpoint.vout)
+        } else {
+            tx_url
+        }
+    }
 }
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, PartialEq, Serialize)]
 pub enum TxLabel {
     Lock,
     Commit,


klochowicz:
one question I wasn't able to answer myself was: should I try to cram the pubkey script along the transaction to places where I haven't yet added the `TxUrl::from_transaction`? Or get the script pubkey from `self.latest.dlc` ?

I marked this PR as draft as I'll proceed when I get some feedback which of the ways should I go :)

klochowicz:
Fixes #799

daemon/src/projection.rs
```diff
@@ -420,6 +429,7 @@ impl Cfd {
                 // Only allow committing once the oracle attested.
                 (CfdState::PendingCet, vec![CfdAction::Commit])
             }
+
```
luckysori: ❓ Random newline?

daemon/src/projection.rs
```diff
@@ -755,6 +765,28 @@ impl TxUrl {
             url: to_mempool_url(txid, network),
         }
     }
+
+    /// Highlight particular transaction output in the TxUrl
+    pub fn with_output_index(mut self, index: u32) -> Self {
```
luckysori: 🔧 I don't think any of these methods need to public. Including `new` above.

daemon/src/projection.rs
```diff
@@ -290,8 +293,9 @@ impl Cfd {
                 (CfdState::ContractSetup, vec![])
             }
             ContractSetupCompleted { dlc } => {
-                self.details.tx_url_list.push(TxUrl::new(
-                    dlc.lock.0.txid(),
+                self.details.tx_url_list.push(TxUrl::from_transaction(
+                    &dlc.lock.0,
+                    &dlc.script_pubkey_for(role),
```
luckysori: 📣 The lock transaction has a single output which does not belong to either party and corresponds to a shared output, so this won't highlight anything.

daemon/src/projection.rs
```diff
@@ -367,8 +374,9 @@ impl Cfd {
             CommitConfirmed => {
                 // pretty weird if this is not defined ...
                 if let Some(dlc) = self.latest_dlc.as_ref() {
-                    self.details.tx_url_list.push(TxUrl::new(
-                        dlc.commit.0.txid(),
+                    self.details.tx_url_list.push(TxUrl::from_transaction(
```
luckysori: 📣 Same as with the lock transaction, the commit transaction has a single output which does not belong to either party and corresponds to a shared output, so this won't highlight anything.

luckysori (commented):
As pointed out, this feature doesn't apply to lock or commit, because those transactions have one shared output. I think refund and collaborative settlement are handled correctly.

What's missing is handling CETs. You should do the same thing: get the `Dlc` and call `script_pubkey_for` on it to get the correct script pubkey.

daemon/src/projection.rs
klochowicz: indeed! this struct used to be in a different file, so it's probably just a historical reason. I'll remove them :)

thomaseizinger:
> As pointed out, this feature doesn't apply to lock or commit, because those transactions have one shared output. I think refund and collaborative settlement are handled correctly.

Lock will almost certainly have change outputs though so we still need to highlight the correct one!

daemon/src/projection.rs
klochowicz: `TxUrl` struct doesn't need to be public anymore either - changed that too :)

daemon/src/projection.rs
klochowicz: I've added a debug_assert to prevent anyone from trying using this with commit transaction (along with explanation)

klochowicz:
> > As pointed out, this feature doesn't apply to lock or commit, because those transactions have one shared output. I think refund and collaborative settlement are handled correctly.
> 
> Lock will almost certainly have change outputs though so we still need to highlight the correct one!

added a comment why we still try to highlight the correct one for lock transactions

daemon/src/projection.rs
```diff
@@ -264,6 +264,8 @@ impl Cfd {
                 (CfdState::ContractSetup, vec![])
             }
             ContractSetupCompleted { dlc } => {
+                // Lock transaction has shared output, but it likely has change
+                // output too that can be highlighted
```
thomaseizinger: But are you going to find the correct output based on the role? I am not sure that `dlc.script_pubkey_for` is correct here!

daemon/src/projection.rs
```diff
@@ -389,9 +390,17 @@ impl Cfd {
                 (CfdState::PendingCommit, vec![])
             }
             OracleAttestedPostCetTimelock { cet, price } => {
-                self.details
-                    .tx_url_list
-                    .push(TxUrl::new(cet.txid(), network, TxLabel::Cet));
+                let tx_url = if let Some(dlc) = self.latest_dlc.as_ref() {
+                    TxUrl::from_transaction(
+                        &cet,
+                        &dlc.script_pubkey_for(role),
```
thomaseizinger, line 400: Same here. What script is this returning? Do our CETs actually spend to this script?

luckysori: I think we use the same address for all final transactions, so this should be okay if that's the case.

daemon/src/projection.rs
luckysori: I don't think this one is correct either, because the change outputs are added by the `bdk` wallets automatically. It may be hard to highlight these without introducing weird changes to `maia`.

klochowicz: should we not bother with the lock transaction's change output in this PR then? I'm happy to create a separate issue that will track that, especially that it would need to interact with `maia`

luckysori: I would not bother. But maybe @thomaseizinger knows how to solve this quickly.

thomaseizinger: We should be able to find the lock output based on the amount right? We know how much the parties locked up so that should be possible to easily expose as a function on `Dlc` then.

luckysori: Probably a good enough heuristic given how unlikely it is that both parties generate change outputs with the same amount.

klochowicz: Created a small PR for `maia` to allow for that: https://github.com/comit-network/maia/pull/12 - I will update this one one it gets merged

klochowicz: > Created a small PR for `maia` to allow for that: [comit-network/maia#12](https://github.com/comit-network/maia/pull/12) - I will update this one one it gets merged

migrated the changes here.

daemon/src/lib.rs
```diff
@@ -64,6 +64,7 @@ pub mod supervisor;
 pub mod taker_cfd;
 pub mod to_sse_event;
 pub mod tokio_ext;
+pub mod transaction_ext;
```
thomaseizinger: Tiny nit: We usually name these `_ext` modules after the crate from where they originate. In this case `bdk` (which re-exports `bitcoin`).

daemon/src/projection.rs
```diff
@@ -261,11 +265,19 @@ impl Cfd {
...
-                    network,
-                    TxLabel::Lock,
-                ));
+                // Lock transaction has shared output, but it likely has change
+                // output too that can be highlighted
+                let tx_url = TxUrl::new(dlc.lock.0.txid(), network, TxLabel::Lock);
+                // Find the relevant vout based on the locked amount as we don't
+                // have script pubkey for it.
+                let tx_url =
+                    if let Ok(outpoint) = dlc.lock.0.outpoint_of_value(dlc.lock_amount_for(role)) {
```
luckysori: 📣 I don't think this is quite right. If the goal is to only highlight the `lock_output` you need to add up the `taker_lock_amount` and the `maker_lock_amount`. The sum of those values is what makes up the `lock_output`.

thomaseizinger: I am afraid this is still not correct :sweat_smile: 

The output we are interested in is the _sum_ of both maker and taker. I'd suggest we expose a getter for that on `Dlc`.

daemon/src/projection.rs
```diff
@@ -302,10 +314,13 @@ impl Cfd {
                 (CfdState::IncomingSettlementProposal, vec![])
             }
             CollaborativeSettlementCompleted {
-                spend_tx, price, ..
+                spend_tx,
+                price,
+                script,
```
thomaseizinger, line 317: I love how this is now also useful here and not just for monitoring. Confirms that packing this information in the event is the right thing to do!

thomaseizinger (commented):
Did you test that this actually works? I am not sure it does for lock!

daemon/src/projection.rs
luckysori: But thinking about it, we shouldn't be using amounts to identify outputs if we can help it. For some reason I thought you would use the amount to identify change outputs of the lock transaction. Those cannot be easily identified otherwise, because `bdk` chooses a change address automatically during contract setup.

To identify the `lock_output` you can use this public method from `maia`: https://github.com/comit-network/maia/blob/close-tx-fee-rate/src/protocol.rs#L251-L264. Sorry for the confusion. I didn't realise the goal was to highlight the `lock_output`.

daemon/src/lib.rs
klochowicz: Is it a recent convention? I just did the same as in `maia`...

daemon/src/projection.rs
klochowicz: Yes, this is nice. The reason I've put a draft originally was myself wondering whether we should pack the script pubkey along the transactions when we can, instead of fishing it out from DLC

daemon/src/projection.rs
klochowicz: I don't think we need to go back to `maia` here, that `lock_desc` was readily available in the `Dlc`. Now, the question arises: should we do the same for commit transaction? I can see that we have a descriptor there too?

daemon/src/lib.rs
klochowicz: removed the file - not needed after all :)

daemon/src/projection.rs
klochowicz: We can already access that, this is the other part of the tuple?

daemon/src/projection.rs
```diff
@@ -694,16 +710,46 @@ pub fn to_mempool_url(txid: Txid, network: Network) -> String {
...
+
+    /// If the Transaction contains the script_pubkey, output will be selected
+    /// in the URL. Otherwise, fall back to the main txid URL.
+    fn from_transaction(
+        transaction: &Transaction,
+        script_pubkey: &Script,
+        network: Network,
+        label: TxLabel,
+    ) -> Self {
+        debug_assert!(label != TxLabel::Commit, "commit transaction has a single output which does not belong to either party - this won't highlight anything");
```
luckysori, line 742: :+1: Haha, nice! I never use `debug_assert`s.

luckysori: To be fair, we are highlighting an output that doesn't belong to any one party in the lock transaction too, so you could just say: `commit transaction only ever has one output, nothing to highlight`.

daemon/src/projection.rs
```diff
@@ -261,11 +265,17 @@ impl Cfd {
                 (CfdState::ContractSetup, vec![])
             }
             ContractSetupCompleted { dlc } => {
-                self.details.tx_url_list.push(TxUrl::new(
-                    dlc.lock.0.txid(),
+                // Lock transaction has shared output, but it likely has change
+                // output too that can be highlighted
```
luckysori: It's more that we can highlight the shared output, but cannot discern between our change outputs and the other party's.

daemon/src/projection.rs
```diff
@@ -261,11 +265,17 @@ impl Cfd {
...
+                let (lock_tx, lock_desc) = &dlc.lock;
+
+                let tx_url = TxUrl::from_transaction(
+                    lock_tx,
+                    &lock_desc.script_pubkey(),
                     network,
-                    TxLabel::Lock,
-                ));
+                    TxLabel::Collaborative,
```
luckysori: 🔧 I think this should be `TxLabel::Lock` or similar.

klochowicz: oops, of course! Don't know how this has slipt in.

daemon/src/projection.rs
klochowicz: I got confused - why are we highlighting the one in lock transaction then, and not this one too?

klochowicz: I kept the old wording, otherwise the debug_assert makes less sense. the lock one has a comment next to the invocation.

daemon/src/projection.rs
klochowicz: changed the comment.

daemon/src/projection.rs
luckysori: > I got confused - why are we highlighting the one in lock transaction then, and not this one too?

If there's only one output do you need to highlight anything?

At least in the context of the lock transaction you can highlight the one output that matters for the protocol, even if you can't also show the change outputs that belong to that specific party.

daemon/src/model/cfd.rs
```diff
@@ -1389,6 +1389,13 @@ impl Dlc {
         }
     }
 
+    pub fn lock_amount_for(&self, role: Role) -> Amount {
+        match role {
+            Role::Maker => self.maker_lock_amount,
+            Role::Taker => self.taker_lock_amount,
+        }
+    }
+
```
thomaseizinger: Do we still need this?

thomaseizinger (approved):

luckysori (approved):

PR #1043, opened 2022-01-07
Introduce xtra_ext::SendInterval trait for periodic actor messages

### LINKED ISSUE #1042: Heartbeat pulse measurements pile up - after first one gets discarded, another one drops connection
Heartbeat measurements (and potentially other periodic messages on actors) keep piling up if actor doesn't get around to handling them. 
Recently introduced change #906 allowed for discarding "bogus" measurement that was clearly stalled. However, it was noticed that even if the out-of-place measurement got discarded, there was another measurement in short succession (e.g. 3ms afterwards) that would then kill the process. 

It would be nice if we added an extension trait on xtra that would only schedule new periodic message after previous one got handled.

klochowicz:
Stateless periodic messages (e.g. sync) should not be queued. If for some
reason, the actor blocks before handling a message, standard behaviour of
xtra (via `xtra::Context::notify_interval`) results in piling up many
messages, which then would be handled one after another, which is not the
desired behaviour.

Meanwhile, calling `send_interval()` will only schedule another message when the
previous one was handled.

Ban usage of `xtra::Context::notify_interval` via clippy lint.

Fixes #1042

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -2,4 +2,5 @@ disallowed-methods = [
   "xtra::Address::do_send_async", # discards the return value, possibly swallowing an error
   "xtra::Address::do_send", # discards the return value, possibly swallowing an error
   "xtra::message_channel::MessageChannel::do_send", # discards the return value, possibly swallowing an error
+  "xtra::Context::notify_interval", # does not wait for response from the previous handler, prefer `xtra_ext::Address::send_interval`
 ]
--- a/daemon/src/xtra_ext.rs
+++ b/daemon/src/xtra_ext.rs
@@ -1,5 +1,6 @@
 use async_trait::async_trait;
 use std::fmt;
+use std::time::Duration;
 use xtra::address;
 use xtra::message_channel;
 use xtra::Actor;
@@ -48,5 +49,44 @@ where
         }
 
         Ok(())
+    }
+}
+
+#[async_trait]
+pub trait SendInterval<A, M>
+where
+    M: Message,
+    A: xtra::Handler<M>,
+{
+    /// Similar to xtra::Context::notify_interval, however it uses `send`
+    /// instead of `do_send` under the hood.
+    /// The crucial difference is that this function waits until previous
+    /// handler returns before scheduling a new one, thus preventing them from
+    /// piling up.
+    /// As a bonus, this function is non-fallible.
+    async fn send_interval<F>(self, duration: Duration, constructor: F)
+    where
+        F: Send + Sync + Fn() -> M,
+        M: Message<Result = ()>,
+        A: xtra::Handler<M>;
+}
+
+#[async_trait]
+impl<A, M> SendInterval<A, M> for address::Address<A>
+where
+    M: Message,
+    A: xtra::Handler<M>,
+{
+    async fn send_interval<F>(self, duration: Duration, constructor: F)
+    where
+        F: Send + Sync + Fn() -> M,
+    {
+        while self.send(constructor()).await.is_ok() {
+            tokio::time::sleep(duration).await
+        }
+        tracing::warn!(
+            "Task for periodically sending message {} stopped because actor shut down",
+            std::any::type_name::<M>()
+        );
     }
 }
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -8,6 +8,7 @@ use crate::oracle;
 use crate::process_manager;
 use crate::rollover_taker;
 use crate::try_continue;
+use crate::xtra_ext::SendInterval;
 use crate::Tasks;
 use anyhow::Context;
 use anyhow::Result;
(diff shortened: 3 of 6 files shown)


daemon/src/xtra_ext.rs
```diff
@@ -1,5 +1,7 @@
 use async_trait::async_trait;
+use futures_timer::Delay;
```
thomaseizinger: It would be better to use the `tokio` primitive here, otherwise we have two timer reactors running.

We already depend on tokio, might as well use `tokio::time::sleep`!

daemon/src/xtra_ext.rs
```diff
@@ -50,3 +52,42 @@ where
...
+    /// Similar to xtra::Context::notify_interval, however it uses `send`
+    /// instead of `do_send` under the hood.
+    /// The crucial difference is that this function waits until previous
+    /// handler returns before scheduling a new one, thus preventing them from
+    /// piling up.
+    /// As a bonus, this function is non-fallible.
+    async fn send_interval<F>(self, duration: Duration, constructor: F)
+    where
+        F: Send + Sync + Fn() -> M,
+        M: Message,
```
thomaseizinger: Perhaps restrict this to handlers that return `()`?


```suggestion
        M: Message<Result = ()>,
```

daemon/src/xtra_ext.rs
```diff
@@ -50,3 +52,42 @@ where
...
+    where
+        F: Send + Sync + Fn() -> M,
+    {
+        let fut = async move {
+            Delay::new(duration).await;
+            while self.send(constructor()).await.is_ok() {
+                Delay::new(duration).await
+            }
+        };
+        fut.await
```
thomaseizinger:
```suggestion
            Delay::new(duration).await;
            while self.send(constructor()).await.is_ok() {
                Delay::new(duration).await
            }
```

This should be equivalent. Could perhaps even remove the initial sleep?

clippy.toml
```diff
@@ -2,4 +2,5 @@ disallowed-methods = [
   "xtra::Address::do_send_async", # discards the return value, possibly swallowing an error
   "xtra::Address::do_send", # discards the return value, possibly swallowing an error
   "xtra::message_channel::MessageChannel::do_send", # discards the return value, possibly swallowing an error
+  "xtra::Context::notify_interval", # does not wait for response from the previous handler, piling up messages
```
thomaseizinger: It would be good to use the syntax here that allows us to specify the reason so it helps the developer better.

Perhaps also point to the function to use instead?

thomaseizinger (approved):

daemon/src/xtra_ext.rs
```diff
@@ -50,3 +52,42 @@ where
...
+{
+    async fn send_interval<F>(self, duration: Duration, constructor: F)
+    where
+        F: Send + Sync + Fn() -> M,
+    {
+        let fut = async move {
+            Delay::new(duration).await;
+            while self.send(constructor()).await.is_ok() {
+                Delay::new(duration).await
+            }
```
da-kami: Would it be worth printing a log here, that we were unable to send a sync message? 
It's one of these scenarios that "should never happen", in this case I feel it does not harm to add warn/error log here, because we should never see it.

da-kami (approved):

daemon/src/xtra_ext.rs
klochowicz: I thought that at some point we might try to upstream our changes - in that case `tokio` dependency might be a bit harder, but I guess if that ever happens it would trivial to change then. Removed the `futures_timer` dependency and used `tokio::time::sleep`

daemon/src/xtra_ext.rs
klochowicz: great suggestion, thanks!

clippy.toml
klochowicz: I'll create a separate PR to change to the new clippy syntax for all of them. Whichever comes later, I'll rebase the other one to align with your intention.

(added mention of the preferred method)

daemon/src/xtra_ext.rs
klochowicz: sure!

daemon/src/xtra_ext.rs
```diff
@@ -82,12 +81,12 @@ where
...
-            while self.send(constructor()).await.is_ok() {
-                Delay::new(duration).await
-            }
-        };
-        fut.await
+        while self.send(constructor()).await.is_ok() {
+            tokio::time::sleep(duration).await
+        }
+        tracing::warn!(
+            "Task periodically sending message {} stopped: ",
```
thomaseizinger:
```suggestion
            "Task for periodically sending message {} stopped because actor shut down",
```

thomaseizinger (approved):

PR #1050, opened 2022-01-10
Monitor actor rebroadcast

### LINKED ISSUE #953: Re-add re-broadcasting of transactions upon startup
https://github.com/itchysats/itchysats/pull/665 temporarily removed re-broadcasting from housekeeping due to the changes in the model.

Broadcasting is currently handled by `wallet::Actor` which doesn't actually align very well with the responsibilities of a wallet (tracking coins and keys to sign transactions).

We want to move this functionality to `monitor::Actor` which already has the responsibility of connecting to the blockchain interacting with it. `monitor::Actor` also already has a startup procedure in which it re-hydrates an aggregate from all CFD events to figure out, what needs to be done. Re-broadcasting transactions can nicely be added there IF broadcasting itself is made a responsibility of `monitor::Actor`.

Solve this issue in two steps:

1. Move `TryBroadcastTransaction` from `wallet::Actor` to `monitor::Actor`
2. Implement re-broadcasting of transactions as part of `monitor::Actor`'s startup procedure

rishflab:
ill rebase and squash

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -220,9 +220,7 @@ where
         + xtra::Handler<maker_inc_connections::TakerMessage>
         + xtra::Handler<maker_inc_connections::BroadcastOrder>
         + xtra::Handler<Stopping<setup_maker::Actor>>,
-    W: xtra::Handler<wallet::Sign>
-        + xtra::Handler<wallet::BuildPartyParams>
-        + xtra::Handler<wallet::TryBroadcastTransaction>,
+    W: xtra::Handler<wallet::Sign> + xtra::Handler<wallet::BuildPartyParams>,
 {
     async fn handle_take_order(
         &mut self,
@@ -444,7 +442,7 @@ where
     O: xtra::Handler<oracle::MonitorAttestation>,
     T: xtra::Handler<maker_inc_connections::settlement::Response>
         + xtra::Handler<Stopping<collab_settlement_maker::Actor>>,
-    W: xtra::Handler<wallet::TryBroadcastTransaction>,
+    W: 'static + Send,
 {
     async fn handle_propose_settlement(
         &mut self,
@@ -557,9 +555,7 @@ where
         + xtra::Handler<maker_inc_connections::settlement::Response>
         + xtra::Handler<Stopping<collab_settlement_maker::Actor>>
         + xtra::Handler<maker_inc_connections::RegisterRollover>,
-    W: xtra::Handler<wallet::Sign>
-        + xtra::Handler<wallet::BuildPartyParams>
-        + xtra::Handler<wallet::TryBroadcastTransaction>,
+    W: xtra::Handler<wallet::Sign> + xtra::Handler<wallet::BuildPartyParams>,
 {
     async fn handle(&mut self, FromTaker { taker_id, msg }: FromTaker, ctx: &mut Context<Self>) {
         match msg {
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -1,3 +1,5 @@
+use crate::bitcoin::consensus::encode::serialize_hex;
+use crate::bitcoin::Transaction;
 use crate::db;
 use crate::model;
 use crate::model::cfd;
@@ -9,6 +11,7 @@ use crate::model::BitMexPriceEventId;
 use crate::oracle;
 use crate::oracle::Attestation;
 use crate::try_continue;
+use crate::wallet::RpcErrorCode;
 use crate::xtra_ext::SendInterval;
 use crate::Tasks;
 use anyhow::Context;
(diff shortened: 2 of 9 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -515,7 +513,7 @@ where
     O: xtra::Handler<oracle::MonitorAttestation>,
     T: xtra::Handler<maker_inc_connections::settlement::Response>
         + xtra::Handler<Stopping<collab_settlement_maker::Actor>>,
-    W: xtra::Handler<wallet::TryBroadcastTransaction>,
+    W: 'static + std::marker::Send,
```
thomaseizinger: Can we shorted this to `Send`? It is in the prelude.

daemon/src/monitor.rs
```diff
@@ -130,14 +137,17 @@ impl Cfd {
         use CfdEvent::*;
         match event.event {
```
thomaseizinger, line 154: I think there are a couple of events where we can set some of these transactions to `None`!

For example, once lock is confirmed, we no longer need to broadcast it.

thomaseizinger (approved):

daemon/src/monitor.rs
```diff
@@ -102,14 +137,17 @@ impl Cfd {
...
+                params: Some(MonitorParams::new(dlc.clone())),
                 monitor_lock_finality: true,
                 monitor_commit_finality: true,
                 monitor_cet_timelock: true,
                 monitor_refund_timelock: true,
                 monitor_refund_finality: true,
                 monitor_revoked_commit_transactions: false,
                 monitor_collaborative_settlement_finality: None,
+                lock_tx: Some(dlc.lock.0),
+                cet: self.cet,
```
da-kami: Particularly for `ContractSetupCompleted` I find this a bit confusing. Upon contract setup completed I would expect `cet` and `commit_tx` to be set to `None`. I think that is what effectively happens here, because it *will* always be `None` at this point. 

Is there any specific reason why we pass through the value here?

da-kami (approved):

daemon/src/monitor.rs
rishflab: nope, I am not super confident with these state transitions so I copied what we were doing in an other file. will fix

daemon/src/monitor.rs
```diff
@@ -708,14 +815,17 @@ impl xtra::Message for Sync {
...
+    C: Send + 'static + bdk::electrum_client::ElectrumApi,
     Self: xtra::Handler<Sync>,
 {
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let fut = ctx
             .notify_interval(Duration::from_secs(20), || Sync)
             .expect("we just started");
 
+        let addr = ctx.address().expect("Able to acquire own address");
+        let _ = self.rebroadcast_transactions(&addr).await;
```
thomaseizinger: Don't do this. The call to `send` within `rebroadcast` will deadlock this because we are not started yet.

Can we just do this as normal function calls within the constructor?

daemon/src/monitor.rs
```diff
@@ -515,6 +570,58 @@ where
...
+            let Cfd {
+                monitor_lock_finality,
+                monitor_commit_finality,
+                cet,
+                commit_tx,
+                lock_tx,
+                ..
+            } = events.into_iter().fold(Cfd::default(), Cfd::apply);
+
+            if !monitor_commit_finality {
```
thomaseizinger: This should not be necessary. The `apply` logic should set `commit_tx` to `Some` if you should broadcast and `None` if you shouldn't.

daemon/src/monitor.rs
```diff
@@ -899,14 +1060,15 @@ mod tests {
...
-        fn for_test<const N: usize>(
+        async fn for_test<const N: usize>(
             event_channel: Box<dyn StrongMessageChannel<Event>>,
             subscriptions: [((Txid, Script), Vec<(ScriptStatus, Event)>); N],
         ) -> Self {
             Actor {
                 cfds: HashMap::default(),
                 event_channel,
                 client: stub::Client::default(),
+                db: SqlitePool::connect(":memory:").await.unwrap(),
```
thomaseizinger: This shouldn't be necessary if you do it in the ctor because then you don't need to save the db as a field.

thomaseizinger (changes requested):

daemon/src/monitor.rs
rishflab: I have removed the finality checks and I rebroadcast if the txns are `Some`. I think we need to change the aggregate. Sounds like this should be an enum with three variants rather than 3 options.

daemon/src/monitor.rs
```diff
@@ -711,7 +842,7 @@ impl xtra::Message for Sync {
 #[async_trait]
 impl<C> xtra::Actor for Actor<C>
 where
-    C: Send + 'static,
+    C: Send + 'static + bdk::electrum_client::ElectrumApi,
```
thomaseizinger: This seems unnecessary now.

daemon/src/monitor.rs
```diff
@@ -900,7 +1043,7 @@ mod tests {
 
     impl Actor<stub::Client> {
         #[allow(clippy::type_complexity)]
-        fn for_test<const N: usize>(
+        async fn for_test<const N: usize>(
```
thomaseizinger: Unnecessary now?

daemon/src/monitor.rs
```diff
@@ -746,6 +877,15 @@ where
             collaborative_settlement.order_id,
         );
     }
+
+    async fn handle_try_broadcast_transaction(
+        &mut self,
+        msg: TryBroadcastTransaction,
+        _ctx: &mut xtra::Context<Self>,
```
thomaseizinger: `xtra-productivity` can handle functions that don't need `xtra::Context`! Just omit it!

thomaseizinger (approved):
Some ideas for minimizing the diff. Good to go though :)

daemon/src/monitor.rs
rishflab: sorry should of checked the diff more thoroughly

thomaseizinger (approved):

PR #1051, opened 2022-01-10
Reintroduce automatic refund

### LINKED ISSUE #954: Re-implement handling of refund transaction
https://github.com/itchysats/itchysats/pull/665 temporarily disabled handling of refund due to changes in the model. Needs to be re-implemented (and tested?).

Let's add an actor test too.

luckysori:
Fixes #954.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -340,7 +340,7 @@ pub enum CfdEvent {
         cet: Transaction,
     },
 
-    RefundTimelockConfirmed {
+    RefundTimelockExpired {
         #[serde(with = "hex_transaction")]
         refund_tx: Transaction,
     },
@@ -1109,7 +1109,7 @@ impl Cfd {
             CetConfirmed => self.cet_finality = true,
             RefundConfirmed => self.refund_finality = true,
             CollaborativeSettlementConfirmed => self.collaborative_settlement_finality = true,
-            RefundTimelockConfirmed { .. } => self.refund_timelock_expired = true,
+            RefundTimelockExpired { .. } => self.refund_timelock_expired = true,
             LockConfirmed => self.lock_finality = true,
             CommitConfirmed => self.commit_finality = true,
             CetTimelockConfirmedPriorOracleAttestation
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -268,6 +268,18 @@ pub enum RolloverError {
     },
 }
 
+/// Errors that can happen when handling the expiry of the refund
+/// timelock on the commit transaciton.
+#[derive(thiserror::Error, Debug)]
+pub enum RefundTimelockExpiryError {
+    #[error("CFD is already final")]
+    AlreadyFinal,
+    #[error("CFD does not have a DLC")]
+    NoDlc,
+    #[error("Failed to sign refund transaction")]
+    Signing(#[from] anyhow::Error),
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub struct Event {
     pub timestamp: Timestamp,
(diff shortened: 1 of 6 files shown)


daemon/src/projection.rs
```diff
@@ -362,7 +362,15 @@ impl Cfd {
             CetTimelockConfirmedPostOracleAttestation { .. } => {
                 (CfdState::PendingCet, self.actions)
             }
-            RefundTimelockConfirmed { .. } => (self.state, self.actions),
+            RefundTimelockExpired { .. } => {
+                // TODO: Not sure what to display here! There are
```
luckysori: 👀 I'm unsure about this change.

daemon/src/model/cfd.rs
```diff
@@ -268,6 +268,18 @@ pub enum RolloverError {
...
 
+/// Errors that can happen when handling the expiry of the refund
+/// timelock on the commit transaciton.
+#[derive(thiserror::Error, Debug)]
+pub enum RefundTimelockExpiryError {
+    #[error("CFD is already final")]
+    AlreadyFinal,
+    #[error("CFD does not have a DLC")]
+    NoDlc,
+    #[error("Failed to sign refund transaction: {0}")]
```
```suggestion
    #[error("Failed to sign refund transaction")]
```

Otherwise you are duplicating the error message.

daemon/src/model/cfd.rs
```diff
@@ -944,8 +956,19 @@ impl Cfd {
...
     }
 
-    pub fn handle_refund_timelock_expired(self) -> Event {
-        todo!()
+    pub fn handle_refund_timelock_expired(self) -> Result<Event, RefundTimelockExpiryError> {
+        use RefundTimelockExpiryError::*;
+
+        if self.is_final() {
+            return Err(AlreadyFinal);
+        }
```
thomaseizinger, line 964: Is this actually an error or should we just emit `None`?

daemon/tests/happy_path.rs
```diff
@@ -259,6 +259,26 @@ async fn maker_rejects_rollover_of_open_cfd() {
...
 
+#[tokio::test]
+async fn open_cfd_is_refunded() {
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement()).await;
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
```
thomaseizinger, line 270: Arent' we waiting as part of `assert_next_state`?

daemon/src/model/cfd.rs
thomaseizinger: I think the error handling in `cfd_actors` highlights that this is not actually an error.

daemon/src/projection.rs
thomaseizinger: I think it is fine to prioritize displaying the refund state here temporarily. If we end up in a race condition on which TX makes it. Whatever gets confirmed will fix it later.

thomaseizinger (approved):
Thanks!

daemon/tests/happy_path.rs
luckysori: I was following a pattern that I noticed above. I can remove all these sleeps if they don't help.

daemon/src/model/cfd.rs
luckysori: Don't I need to annotate it as `source` then? Keep in mind that we are not using `anyhow::Context` where this error can occur.

daemon/src/model/cfd.rs
luckysori: Strictly speaking it's not an error, but I found it more ergonomic both here and in the call site to treat all the "non-success" cases as part of the same `enum`.

daemon/tests/happy_path.rs
klochowicz: the sleeps are necessary in most cases, so just removing them should result in breaking at least some tests.. I'm happy for them to be inside the `assert_next_state` (much less clutter, for the price of some needless sleeps). 

Of course check first, maybe we'll get away without it! :)

daemon/src/model/cfd.rs
thomaseizinger: `#[from]` implies `#[source]`: https://docs.rs/thiserror/latest/thiserror/index.html#details

daemon/tests/happy_path.rs
thomaseizinger: Perhaps in a follow-up if anyone is motivated :D

luckysori: I'm not sure what the best way forward is in this case, so I will just leave it as is. I think the end goal should be to not need any explicit `sleep`s in these tests, but maybe that's just not feasible here.

PR #1053, opened 2022-01-10
Remove `max_giveable`

thomaseizinger:
This function was a horrendous hack introduced in the xmr-btc codebase
for figuring out how much we can maximally withdraw from a wallet.

Introducing it here is unnecessary because we can utilize BDK native
functionality to drain the entire wallet as we don't need to know
exactly, how much we are withdrawing if we are draining it.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -13,7 +13,6 @@ use bdk::bitcoin::Address;
 use bdk::bitcoin::Amount;
 use bdk::bitcoin::OutPoint;
 use bdk::bitcoin::PublicKey;
-use bdk::bitcoin::Script;
 use bdk::bitcoin::Transaction;
 use bdk::bitcoin::Txid;
 use bdk::blockchain::Blockchain;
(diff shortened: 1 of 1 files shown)


daemon/src/wallet.rs
```diff
@@ -287,27 +249,33 @@ impl Actor {
...
-            .fee_rate(fee_rate)
-            // Turn on RBF signaling
-            .enable_rbf();
+                    tx_builder.add_recipient(address.script_pubkey(), amount.as_sat());
+                }
+                None => {
+                    tracing::info!(%address, "Draining wallet");
 
-        let (mut psbt, _) = tx_builder.finish()?;
+                    tx_builder.drain_to(address.script_pubkey());
```
klochowicz: shouldn't we call `drain_wallet()` before `drain_to()`? (at least that's what the docs say: https://docs.rs/bdk/0.15.0/bdk/wallet/tx_builder/struct.TxBuilder.html#method.drain_to )

klochowicz (approved):
nice repayment of tech debt! :)

daemon/src/wallet.rs
thomaseizinger: Good find! Should probably yes. Also, I should probably be testing this :D

thomaseizinger: Tested now and works like a charm! https://mempool.space/testnet/tx/539f544898d2e2d0ae8289dc0677ad0df97eb025a9319a5c8a2c5f46723d52b8

da-kami (approved):

thomaseizinger:
bors retry

PR #1054, opened 2022-01-10
Use new clippy `disallowed_methods` syntax

klochowicz:
More clear in TOML file, hopefully will give better error messages as well.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,6 +1,6 @@
 disallowed-methods = [
-  "xtra::Address::do_send_async", # discards the return value, possibly swallowing an error
-  "xtra::Address::do_send", # discards the return value, possibly swallowing an error
-  "xtra::message_channel::MessageChannel::do_send", # discards the return value, possibly swallowing an error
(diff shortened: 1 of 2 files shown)


luckysori (approved):

thomaseizinger:
Wait, why do we have two of these files? How does that actually behave? Are they merged together?

klochowicz:
> Wait, why do we have two of these files? How does that actually behave? Are they merged together?

I think one of them concerned the workspace, the other one the `daemon` crate. I just upgraded them together. happy to merge them into one - which one would you prefer to keep?

thomaseizinger:
> I think one of them concerned the workspace, the other one the `daemon` crate. I just upgraded them together. happy to merge them into one - which one would you prefer to keep?

I think I'd keep the workspace one!

klochowicz:
> > I think one of them concerned the workspace, the other one the `daemon` crate. I just upgraded them together. happy to merge them into one - which one would you prefer to keep?
> 
> I think I'd keep the workspace one!

#1107

PR #1055, opened 2022-01-10
Scan for the next 1000 addresses

### LINKED ISSUE #965: Investigate coins not found in development mode
We sometimes run into the problem that we start with 0 balance when using our development seeds.
Perhaps #838 is related to this issue.

We should investigate, how bdk scans for coins (enable some logs?) and see how to resolve the issue. The original thinking was, because we use electrum as a backend, we should be able to re-scan from scratch on every startup and find ALL coins.

thomaseizinger:
Fixes #965.

Seems to be an usptream bug: https://github.com/bitcoindevkit/bdk/issues/521.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -115,7 +115,7 @@ impl Actor {
 
     fn sync_internal(&mut self) -> Result<WalletInfo> {
         self.wallet
-            .sync(NoopProgress, None)
+            .sync(NoopProgress, Some(1000))
             .context("Failed to sync wallet")?;
 
         let balance = self.wallet.get_balance()?;


luckysori (approved):

daemon/src/wallet.rs
```diff
@@ -115,7 +115,7 @@ impl Actor {
 
     fn sync_internal(&mut self) -> Result<WalletInfo> {
         self.wallet
-            .sync(NoopProgress, None)
+            .sync(NoopProgress, Some(1000))
```
klochowicz, line 118: would you mind adding a comment why this magic number got introduced next to it? Would be good for maintainability 👍 

(also: typo in commit desc: *upstream*)

couple of questions that the commit description could resolve:
- would it be enough for long-running umbrel nodes in production? 
- should this be reverted if upstream fixes the problem?

klochowicz (approved):
great find with the upstream bug! just some nitpicks re: maintainability

da-kami (approved):
Thanks!

❓ Will we have to increase `1000` eventually after using more than `1000` addresses? 

(I assume by then the upstream but will be fixed, still asking to see if I understand the change correctly.)

thomaseizinger:
> Thanks!
> 
> question Will we have to increase `1000` eventually after using more than `1000` addresses?
> 
> (I assume by then the upstream but will be fixed, still asking to see if I understand the change correctly.)

Yep! See commit message :)

bors r+

PR #1058, opened 2022-01-10
Introduce `command::Executor`

### LINKED ISSUE #1035: Contract setup commands ("accept", "reject") should return a failed event instead of logging error
After an error during contract setup (due to loss of connection from the taker if I remember correctly), the maker got stuck with an cfd that could be neither accepted nor rejected.
Both commands were indicating that they can't success, logs were informing me about this, but no event was produced; this meant that the cfd was stuck in this state forever.
This example shows that if we *know* what should be the next event (in this case, `ContractSetupFailed`), we should return that event from the command and apply it. 

note: this problem might be applicable to other events as well

thomaseizinger:
I initially set out to https://github.com/itchysats/itchysats/issues/1035 and made so many improvements on the way that I figured I post them up as a separate PR.

This ended up primarily introducing a concept I called `command::Executor` which wraps the DB connection and the ProcessManager and allows us to easily execute a command against a CFD.

I've used it already for some things inside taker and maker CFD actor, then wanted to move the invocation of the command on the CFD for monitoring into the monitoring actor which triggered a refactor there. Couldn't quite finish that unfortunately because our tests still rely on being able to dispatch a monitoring event to the CFD actors!

### DIFF
--- a/daemon/src/command.rs
+++ b/daemon/src/command.rs
@@ -1,0 +1,47 @@
+use crate::cfd_actors::load_cfd;
+use crate::model::cfd::Cfd;
+use crate::model::cfd::Event;
+use crate::process_manager;
+use crate::OrderId;
+use anyhow::Context;
+use anyhow::Result;
+use xtra::Address;
+
+pub struct Executor {
+    db: sqlx::SqlitePool,
+    process_manager: Address<process_manager::Actor>,
+}
+
+impl Executor {
+    pub fn new(db: sqlx::SqlitePool, process_manager: Address<process_manager::Actor>) -> Self {
+        Self {
+            db,
+            process_manager,
+        }
+    }
+
+    pub async fn execute(
+        &self,
+        id: OrderId,
+        command: impl FnOnce(Cfd) -> Result<Event>,
+    ) -> Result<()> {
+        let mut connection = self
+            .db
+            .acquire()
+            .await
+            .context("Failed to acquire DB connection")?;
+        let cfd = load_cfd(id, &mut connection)
+            .await
+            .context("Failed to load CFD")?;
+
+        let event = command(cfd).context("Failed to execute command on CFD")?;
+
+        self.process_manager
+            .send(process_manager::Event::new(event))
+            .await
+            .context("ProcessManager is disconnected")?
+            .context("Failed to process new domain event")?;
+
+        Ok(())
+    }
+}
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -68,10 +68,25 @@ pub struct Actor<C = bdk::electrum_client::Client> {
     cfds: HashMap<OrderId, MonitorParams>,
     event_channel: Box<dyn StrongMessageChannel<Event>>,
     client: C,
+    tasks: Tasks,
+    state: State,
+}
+
+/// Internal data structure encapsulating the monitoring state without performing any IO.
+struct State {
     latest_block_height: BlockHeight,
     current_status: BTreeMap<(Txid, Script), ScriptStatus>,
     awaiting_status: HashMap<(Txid, Script), Vec<(ScriptStatus, Event)>>,
-    tasks: Tasks,
+}
+
+impl State {
+    fn new(latest_block_height: BlockHeight) -> Self {
+        State {
+            latest_block_height,
+            current_status: BTreeMap::default(),
+            awaiting_status: HashMap::default(),
+        }
+    }
 }
 
 /// Read-model of the CFD for the monitoring actor.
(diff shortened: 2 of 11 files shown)


daemon/src/setup_taker.rs
```diff
@@ -141,7 +141,8 @@ impl Actor {
     }
 
     fn handle(&mut self, msg: SetupSucceeded, ctx: &mut xtra::Context<Self>) {
-        let _: Result<(), xtra::Disconnected> = self.on_completed
+        let _: Result<(), xtra::Disconnected> = self
```
luckysori, line 144: :wrench: Some of these changes seem to belong to the previous commit.

daemon/src/setup_maker.rs
```diff
@@ -133,13 +129,28 @@ impl Actor {
...
-            .await;
+        match self.execute_setup_contract_command(completed).await {
+            Ok(()) => {}
+            Err(e) => {
+                tracing::error!("Failed to execute `contract_setup` command: {:#}", e);
+            }
+        }
+
+        ctx.stop(); // outside of `execute_setup_contract_command` because we always want to shut
+                    // down
```
luckysori: ❓ Do we need a comment for this? I thought we had been doing similar stuff elsewhere.

daemon/src/monitor.rs
```diff
@@ -120,8 +370,9 @@ impl Cfd {
                     monitor_cet_timelock: true,
                     monitor_refund_timelock: true,
                     monitor_refund_finality: true,
-                    monitor_revoked_commit_transactions: true, /* After rollover, the other party
-                                                                * might publish old states. */
+                    monitor_revoked_commit_transactions: true,
+                    /* After rollover, the other party
+                     * might publish old states. */
```
luckysori: 🔧 You probably meant to put this on top of `monitor_revoked_commit_transactions` (or keep it beside it).

luckysori (approved):
Really cool changes, thank you!

I am all for what you set out to do with https://github.com/itchysats/itchysats/pull/1058/commits/449118b12f3e51dc7df88795717a6d453ef43135# but the diff is very hard to review for me. Any particular bits that you want us to check?

daemon/src/monitor.rs
thomaseizinger: That was an auto-formatting error. Will fix!

thomaseizinger:
> I am all for what you set out to do with [449118b](https://github.com/itchysats/itchysats/commit/449118b12f3e51dc7df88795717a6d453ef43135) but the diff is very hard to review for me. Any particular bits that you want us to check?

I can try and dissect it into smaller steps. It is mostly moving code around, not actually changing any logic!

luckysori:
> I can try and dissect it into smaller steps. It is mostly moving code around, not actually changing any logic!

If it's too much effort don't worry about it!

thomaseizinger:
> > I can try and dissect it into smaller steps. It is mostly moving code around, not actually changing any logic!
> 
> If it's too much effort don't worry about it!

Wasn't too hard, please check again!

luckysori (approved):
Approved again. The diff on https://github.com/itchysats/itchysats/pull/1058/commits/be110ad248e936563625240ec19acebfce6f32bf is so easy to review now! Feel free to reorganise it like you originally had it in a separate patch, if you think it helps!

daemon/src/oracle.rs
```diff
@@ -2,7 +2,6 @@ use crate::db;
 use crate::model::cfd::CfdEvent;
 use crate::model::cfd::Event;
```
klochowicz, line 3: a missing "not" in the commit description?

```
Remove spawn_fallible

Pretty much superseded by either Tasks::add or Tasks::add_fallible,
depending on the context. Tasks not tied to any lifetime are likely *not* what
we want.
```

daemon/src/command.rs
```diff
@@ -0,0 +1,47 @@
+use crate::cfd_actors::load_cfd;
+use crate::model::cfd::Cfd;
+use crate::model::cfd::Event;
+use crate::process_manager;
+use crate::OrderId;
+use anyhow::Context;
+use anyhow::Result;
+use xtra::Address;
+
+pub struct Executor {
```
klochowicz, line 10: it would actually be lovely if you inserted a comment here - even that commit description! (good for both IDEs and generated docs)

klochowicz: BTW nice abstraction!

klochowicz (approved):
really cool! got some ideas for follow-up cleanup, but I'll hold my horses until I see the things you've taken out of this commit to introduce it faster (which was a good idea too!)

daemon/src/oracle.rs
thomaseizinger: Reads as intended to me?

klochowicz: I added the *not* in my version - not in the master branch? we don't want tasks tasks that are not tied to any lifetime, right?

PR #1066, opened 2022-01-11
Funding rate taker UI

### LINKED ISSUE #244: Funding rate

da-kami:
Resolves #244 

TODO list until this can get merged:

- [x] Taker UI display costs when opening and running costs on CFD tile
- [x] Business logic (maker&taker) to pass in opening fee into contract setup
- [x] Business logic (maker&taker) to handle recurring fee on every rollover
- [x] Maker API for creating `Offer` to allow configuration of fees

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -2257,7 +2257,7 @@ mod tests {
                 Price::new(dec!(1000)).unwrap(),
                 Usd::new(dec!(100)),
                 Usd::new(dec!(1000)),
-                Origin::Theirs,
+                Origin::Ours,
                 dummy_event_id(),
                 time::Duration::hours(24),
                 1,
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -9,6 +9,7 @@ use crate::model::cfd::CfdEvent;
 use crate::model::cfd::Dlc;
 use crate::model::cfd::Event;
 use crate::model::cfd::OrderId;
+use crate::model::cfd::Origin;
 use crate::model::cfd::Role;
 use crate::model::FundingFee;
 use crate::model::FundingRate;
@@ -748,9 +749,13 @@ impl From<Order> for CfdOrder {
             min_quantity: order.min_quantity,
             max_quantity: order.max_quantity,
             parcel_size,
-            margin_per_parcel: match order.position {
-                Position::Long => calculate_long_margin(order.price, parcel_size, order.leverage),
-                Position::Short => calculate_short_margin(order.price, parcel_size),
+            margin_per_parcel: match (order.origin, order.position) {
+                (Origin::Theirs, Position::Short) | (Origin::Ours, Position::Long) => {
+                    calculate_long_margin(order.price, parcel_size, order.leverage)
+                }
+                (Origin::Ours, Position::Short) | (Origin::Theirs, Position::Long) => {
+                    calculate_short_margin(order.price, parcel_size)
+                }
             },
             leverage: order.leverage,
             liquidation_price: order.liquidation_price,
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -42,7 +42,6 @@ use serde::Serialize;
 use sqlx::pool::PoolConnection;
 use std::collections::HashSet;
 use std::time::Duration;
-use time::macros::datetime;
 use time::OffsetDateTime;
 use tokio::sync::watch;
 use xtra::prelude::MessageChannel;
@@ -697,7 +696,7 @@ pub struct CfdOrder {
     /// Note: It's the minimum possible fee, we cannot calculate the exact
     /// amount until we know the quantity taken by the taker
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc::opt")]
-    pub opening_fee: Option<Amount>,
+    pub opening_fee_per_parcel: Option<Amount>,
 
     /// The interest as annualized percentage
     ///
@@ -709,10 +708,6 @@ pub struct CfdOrder {
     /// This represents the current funding rate of the maker.
     /// The funding rate fluctuates with market movements.
     pub funding_rate_hourly_percent: String,
-
-    /// Timestamp when the next fee will be collected
-    #[serde(with = "::time::serde::timestamp")]
-    pub next_funding_event: OffsetDateTime,
 
     #[serde(with = "round_to_two_dp")]
     pub min_quantity: Usd,
@@ -767,12 +762,10 @@ impl From<Order> for CfdOrder {
                 .expect("settlement_time_interval_hours is always positive number"),
             // XXX: We cannot calculate full fee here, best we can do is to give
             // minimal fee
-            opening_fee: calculate_min_opening_fee(&order),
+            opening_fee_per_parcel: calculate_min_opening_fee(&order),
             funding_rate_annualized_percent: AnnualisedFundingRate::from(order.funding_rate)
                 .to_string(),
             funding_rate_hourly_percent: HourlyFundingRate::from(order.funding_rate).to_string(),
-            // FIXME: Replace dummy value for next funding event
-            next_funding_event: datetime!(2030-09-23 10:00:00).assume_utc(),
         }
     }
 }
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -19,6 +19,7 @@ use daemon::model::Identity;
 use daemon::model::Usd;
 use daemon::monitor::Event;
 use daemon::oracle;
+use daemon::projection::CfdOrder;
 use daemon::projection::CfdState;
 use maia::secp256k1_zkp::schnorrsig;
 use rust_decimal_macros::dec;
--- a/daemon/tests/happy_path.rs
+++ b/daemon/tests/happy_path.rs
@@ -14,6 +14,7 @@ use crate::harness::MakerConfig;
 use crate::harness::Taker;
 use crate::harness::TakerConfig;
 use daemon::connection::ConnectionStatus;
+use daemon::model::cfd::calculate_long_margin;
 use daemon::model::cfd::OrderId;
 use daemon::model::Identity;
 use daemon::model::Usd;
@@ -61,6 +62,15 @@ async fn taker_receives_order_from_maker_on_publication() {
     let (published, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
         .unwrap();
+
+    assert_eq_order(published, received);
+}
+
+fn assert_eq_order(mut published: CfdOrder, received: CfdOrder) {
+    // align margin_per_parcel to be the long margin_per_parcel
+    let long_margin_per_parcel =
+        calculate_long_margin(published.price, published.parcel_size, published.leverage);
+    published.margin_per_parcel = long_margin_per_parcel;
 
     assert_eq!(published, received);
 
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -96,7 +96,7 @@ export const App = () => {
                     <Route
                         path="/wallet"
                         element={<>
-                            <Center>
+                            <Center marginTop={20}>
                                 <VStack>
                                     {connectionStatus}
                                     <Wallet walletInfo={walletInfo} />
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -7,8 +7,11 @@ import {
     Box,
     Center,
     StackDivider,
+    useColorModeValue,
     VStack,
 } from "@chakra-ui/react";
+import dayjs from "dayjs";
+import relativeTime from "dayjs/plugin/relativeTime";
 import * as React from "react";
 import { useState } from "react";
 import { Route, Routes } from "react-router-dom";
@@ -20,7 +23,7 @@ import History from "./components/History";
 import Nav from "./components/NavBar";
 import Trade from "./components/Trade";
 import { Wallet, WalletInfoBar } from "./components/Wallet";
-import { BXBTData, Cfd, ConnectionStatus, intoCfd, intoOrder, Order, StateGroupKey, WalletInfo } from "./types";
+import { BXBTData, Cfd, ConnectionStatus, intoCfd, intoOrder, isClosed, Order, WalletInfo } from "./types";
 import { useEventSource } from "./useEventSource";
 import useLatestEvent from "./useLatestEvent";
 
@@ -41,7 +44,6 @@ export const App = () => {
     const order = useLatestEvent<Order>(source, "order", intoOrder);
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
-    cfds.sort((a, b) => a.order_id.localeCompare(b.order_id));
     const connectedToMakerOrUndefined = useLatestEvent<ConnectionStatus>(source, "maker_status");
     const connectedToMaker = connectedToMakerOrUndefined ? connectedToMakerOrUndefined : { online: false };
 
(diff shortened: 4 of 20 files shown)


daemon/src/projection.rs
```diff
@@ -132,6 +132,17 @@ pub struct Cfd {
...
     pub initial_price: Price,
 
+    /// Sum of all costs
+    ///
+    /// Includes the opening fee and all funding fees period that were already charged.
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub accumulated_funding_fees: Amount,
+
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub funding_fee_per_period: Amount,
```
da-kami: We should use `round` here instead of `period`. I think we agreed on round, introducing `period` is yet another term.

daemon/src/projection.rs
```diff
@@ -527,16 +538,30 @@ impl From<Quote> for bitmex_price_feed::Quote {
...
+    ///
+    /// The opening feel
+    pub opening_fee: Amount,
+
+    /// Funding fee that is collected for the CFD period (until the CFD expires)
+    ///
+    /// The fee that is charged upon every funding round, which is currently charged on every
+    /// roll-over. The fee represents the interest payed from the long position to the short
+    /// position. At the moment the period is 1hour (because rollover is one hour).
+    pub funding_fee_per_period: Amount,
```
da-kami: I think we might additionally need to define the `funding_round_interval` (or similar name), that expresses how often  `funding_fee_per_period` is collected. 

This should be equal to `settlement_interval`, but it is currently not, until we resolve discussion: https://github.com/itchysats/itchysats/discussions/1068

da-kami: added.

daemon/src/projection.rs
da-kami: renamed.

da-kami:
First version of the taker UI changes done:

This includes:

- The pulsating price is now the maker's ask price (because that is relevant in the context of opening a position, not the index price)
- The index price is currently not displayed anymore, because we did not think it is needed. The fact that that price is used for oracle attestations should be mentioned somewhere - but for collab close (which is the go-to path) it is actually *not* used.
- Basic display of margin and fee as table before pressing `Long`
- Open confirm modal adds a row that states that it is perpetual
- We just added the perpetual cost information to the CFD tile for now (stacke on top of the other values)

This work also includes a tile overhaul:

- Change the background and tile layout and better alignment of components
- Close button is in the right bottom corner, transactions are displayed as table.
- Color `P/L percent` and `Payout` according to profit, loss or zero.
- Distinguish between open/close display.
- Display `Lock + Payout` transactions by default
- `Quantity` was renamed to `Contracts` (to go away from the notion of "thinking in USD")

da-kami:
Current state:

image

image

still way to cluttered with information. But it's gonna be hard to hide it without breaking the layout. Not sure we should invest much more time into this version. Probably better to design it new :)

da-kami:
image

da-kami:
Note: After rebasing the displayed transaction history is a bit messed up, I'll look into that and push a fixup.
The API is stable now @klochowicz

klochowicz:
> Note: After rebasing the displayed transaction history is a bit messed up, I'll look into that and push a fixup. The API is stable now @klochowicz

the API got extracted from this PR into: #1129

klochowicz:
> Resolves #244
> 
> So far still very much work in progress :)
> 
> TODO list until this can get merged:
> 
>     * [x]  Taker UI display costs when opening and running costs on CFD tile
> 
>     * [x]  Business logic (maker&taker) to pass in opening fee into contract setup
> 
>     * [x]  Business logic (maker&taker) to handle recurring fee on every rollover
> 
>     * [x]  Maker API for creating `Offer` to allow configuration of fees
> 
> 
> (note: we could merge this with dummy values as well, but probably should not :)

I've updated the TODO items here for the ones that got already merged. There are still a few small issues that we need to resolve for the full functionality though:
- [ ] #1131 
- [ ]  #1114 
- [ ]  #1125

da-kami:
Once #1121 and #1129 merge I'll rebase again and then we can see to get this one in

daemon/src/model/cfd.rs
```diff
@@ -2257,7 +2257,7 @@ mod tests {
                 Price::new(dec!(1000)).unwrap(),
                 Usd::new(dec!(100)),
                 Usd::new(dec!(1000)),
-                Origin::Theirs,
+                Origin::Ours,
```
da-kami, line 2260: Note: This change was not strictly needed, but it is actually more correct, because we do test the `taker` with these... We should probably be more explicit once we create a more elaborate test framework...

daemon/tests/happy_path.rs
```diff
@@ -61,7 +63,16 @@ async fn taker_receives_order_from_maker_on_publication() {
         .await
         .unwrap();
 
-    assert_eq!(published, received);
+    assert_eq_order(published, received);
+}
+
+fn assert_eq_order(mut published: CfdOrder, received: CfdOrder) {
```
thomaseizinger, line 69: Oh my god ...

daemon/src/monitor.rs
```diff
@@ -334,17 +334,17 @@ impl Actor<bdk::electrum_client::Client> {
...
 
             if let Some(cet) = cet {
-                let txid = broadcast_transaction(&self.client, cet)?;
+                let txid = try_continue!(broadcast_transaction(&self.client, cet));
                 tracing::info!("CET published on chain: {}", txid);
             }
 
             if let Some(lock_tx) = lock_tx {
-                let txid = broadcast_transaction(&self.client, lock_tx)?;
+                let txid = try_continue!(broadcast_transaction(&self.client, lock_tx));
```
thomaseizinger: I am fixing this here: https://github.com/itchysats/itchysats/pull/1143

taker-frontend/src/App.tsx
```diff
@@ -107,11 +123,14 @@ export const App = () => {
...
                                     liquidationPrice={liquidationPrice}
                                     walletBalance={walletInfo ? walletInfo.balance : 0}
+                                    opening_fee={openingFee}
+                                    funding_fee_annualized_percent={fundingRateAnnualized}
                                 />
                                 <History
                                     connectedToMaker={connectedToMaker}
                                     cfds={cfds.filter((cfd) => cfd.state.getGroup() !== StateGroupKey.CLOSED)}
                                     title={"Open Positions"}
+                                    isClosed={false}
```
thomaseizinger: Does `closed` here refer to the state of the CFD?

Seems wrong to declare this at the top here. Why can't each CFD tile figure this out by themselves by looking at the state?

taker-frontend/src/components/Trade.tsx
```diff
@@ -93,19 +99,22 @@ export default function Trade({
     let outerCircleBg = useColorModeValue("gray.100", "gray.700");
     let innerCircleBg = useColorModeValue("gray.200", "gray.600");
 
-    const referencePrice = `$${referencePriceAsNumber?.toLocaleString() || "0.0"}`;
-    const askPrice = `$${askPriceAsNumber?.toLocaleString() || "0.0"}`;
-    const liquidationPrice = `$${liquidationPriceAsNumber?.toLocaleString() || "0.0"}`;
+    // Use `Number` wrapper here to ensure the "," separator is printed.
+    const askPrice = `$${Number(askPriceAsNumber)?.toLocaleString() || "0.0"}`;
```
thomaseizinger, line 104: At this point it would be good to introduce components for this stuff and add tests :)

taker-frontend/src/components/Trade.tsx
```diff
@@ -93,19 +99,22 @@ export default function Trade({
...
 
-    const balanceTooLow = walletBalance < margin;
+    let btc_to_open_position = margin + opening_fee;
+
+    const balanceTooLow = walletBalance < btc_to_open_position;
     const quantityTooHigh = maxQuantity < quantity;
     const quantityTooLow = minQuantity > quantity;
     const quantityGreaterZero = quantity > 0;
     const quantityIsEvenlyDivisibleByIncrement = isEvenlyDivisible(quantity, parcelSize);
+    const hourlyFundingRate = (Number.parseFloat(funding_fee_annualized_percent) / 365 / 24).toFixed(5);
```
thomaseizinger: Why aren't we doing this in the backend?

thomaseizinger (commented):
Not sure what is going on with this `isClosed` property, otherwise LGTM.

daemon/tests/happy_path.rs
da-kami: yeah...

daemon/src/monitor.rs
da-kami: Ok, I'll remove this then

taker-frontend/src/App.tsx
da-kami: We split the CFDs into two separate lists and I wanted to adhere to that. This just keeps the logic of "splitting it" in one place, otherwise I have to reason twice - one time outside of `Histroy` and one time inside...

taker-frontend/src/components/Trade.tsx
da-kami: oh, we actually are. That is legacy that should be changed!

da-kami:
TODO: 

- [x] Multiply opening fee by parcel size (to have proper fee)
- [x] Remove `next_funding_event` from the `CfdOrder` and just take the next full hour in the UI

taker-frontend/src/App.tsx
da-kami: I changed it into something more specific

taker-frontend/src/components/Trade.tsx
```diff
@@ -222,14 +236,12 @@ export default function Trade({
                                     onClick={onOpen}
                                     h={16}
                                     w={"40"}
+                                    fontSize={"2xl"}
```
bonomat: nit: wow... imho this now looks gigantic 😬

taker-frontend/src/components/Trade.tsx
```diff
@@ -239,12 +251,17 @@ export default function Trade({
                                         <ModalBody>
                                             <Table variant="striped" colorScheme="gray" size="sm">
                                                 <TableCaption>
-                                                    By submitting, ₿{margin} will be locked on-chain in a contract.
+                                                    By submitting, ₿{btcToOpenPosition}
+                                                    will be locked on-chain in a contract.
```
bonomat: Nit: The line-break here removes the whitespace between `₿{btcToOpenPosition}` and `will`

image

taker-frontend/src/components/Trade.tsx
```diff
@@ -335,37 +363,50 @@ function Leverage({ leverage }: LeverageProps) {
...
-                    <SliderTrack>
-                        <Box position="relative" right={10} />
-                        <SliderFilledTrack />
-                    </SliderTrack>
-                    <SliderThumb boxSize={6}>
-                        <Text color="black">{leverage}</Text>
-                    </SliderThumb>
-                </Slider>
-            </Tooltip>
+            <Slider disabled value={leverage} min={1} max={5} step={1}>
```
bonomat, line 365: That tooltop bothered me for a long time 😬

taker-frontend/src/components/History.tsx
```diff
@@ -98,90 +145,59 @@ const CfdDetails = ({ cfd, connectedToMaker }: CfdDetailsProps) => {
...
-                            <Td textAlign="right">{margin}</Td>
-                        </Tr>
-                        <Tr>
-                            <Td><Text as={"b"}>Unrealized P/L</Text></Td>
-                            <Td textAlign="right">
-                                <Skeleton isLoaded={cfd.profit_btc != null}>
-                                    <ProfitAndLoss profitBtc={cfd.profit_btc!} />
-                                </Skeleton>
-                            </Td>
+                            <Td><Text as={"b"}>Total fees</Text></Td>
```
bonomat, line 152: I'm not convinced it is a good idea to prominently tell the user how much he is losing through fees... 

Also, note that this value might be positive or negative depending on your position. Would it be clear to you what it means if `Total fees = -0.01234` ?

taker-frontend/src/components/History.tsx
```diff
@@ -53,42 +57,85 @@ export default History;
...
+                                    </Skeleton>
+                                </Tooltip>
+                            </Td>
+                        </Tr>
+                        <Tr textColor={useColorModeValue(profitColors[0], profitColors[1])}>
+                            <Td><Text as={"b"}>Payout</Text></Td>
+                            <Td textAlign="right">
+                                <Skeleton isLoaded={cfd.profit_btc != null}>
+                                    <Payout profitBtc={cfd.profit_btc!} margin={cfd.margin} />
+                                </Skeleton>
```
bonomat, line 128: I assume the `cfd.profit_btc` includes the total fees already. If so, what do you think of adding a tooltip explaining this ?

taker-frontend/src/components/History.tsx
```diff
@@ -53,42 +57,85 @@ export default History;
...
+                            </Td>
+                        </Tr>
                         <Tr>
-                            <Td><Text as={"b"}>Quantity</Text></Td>
-                            {/*TODO: Fix textAlign right hacks by using a grid instead ... */}
-                            <Td textAlign="right">{quantity}</Td>
+                            <Td><Text as={"b"}>Margin</Text></Td>
+                            <Td textAlign="right">{margin}</Td>
```
bonomat, line 133: Question: is the margin constant? 

I think, the _current_ `margin` should be _initial_ `margin` minus `fees_paid`. 
Meaning, the longer your position is running, the more fees you have paid, the lower your margin. 
That being said, we could display this to the user ala `Margin = initial funding - fees paid = 0.003 - 0.0001 = ...` (provided we have this data available.

bonomat (approved):
Well done. I like how the tiles are evolving. 

As mentioned on a call: we know already that this layout won't work once the user can go long and short because we will need to display 2 exchange rates. 
I think the approach of having only 1 exchange rate is nice, but maybe we should just remove the `Short` button for now.

taker-frontend/src/App.tsx
thomaseizinger: > We split the CFDs into two separate lists and I wanted to adhere to that. This just keeps the logic of "splitting it" in one place, otherwise I have to reason twice - one time outside of `Histroy` and one time inside...

You always reason about the same thing though which is the state of the CFD? You can easily keep this DRY by extracting that in a function.

What we currently have is pretty confusing because in order to understand what happens, you need to know that the reason this is hard-coded to `false` here is because _a different_ piece of code groups these CFDs into closed and non-closed.

So now you have semantic coupling that is not enforced by the code. If the condition on how you split these CFDs changes without updating this flag, the code inside the tiles will behave incorrectly. The CFD tile already knows its state, it should reason about that locally and not require it to be told externally something that it already knows.

taker-frontend/src/components/History.tsx
da-kami: Given that we cannot add to lock at the moment our margin is not constant but changes with every funding event (i.e. we never add "fresh" money to keep the margin constant or reduce it).

My takeaway from these comments is, that the relationship of margin and fees is currently not clear enough. I will see to display the margin as current margin (i.e. initial margin - fees) and add a tooltip that shows initial margin and costs. That way we also don't show the costs in the tile itself. (I am not sure we have the initial margin available, but if we don't we should change that...)

taker-frontend/src/components/History.tsx
da-kami: Thanks for pointing that one out - the profit currently does *not* reflect the costs at the moment. 
This is not totally trivial to adapt (will require additional tests as well and internal changes to keep track of the costs). I will align with @klochowicz on who shall tackle this, until the it is tracked here: https://github.com/itchysats/itchysats/issues/1165

taker-frontend/src/components/History.tsx
da-kami: see https://github.com/itchysats/itchysats/pull/1066#discussion_r788236094

taker-frontend/src/components/Trade.tsx
da-kami: Yeah - the previous button with the price was actually nice from the looks - it added more to the button. I will see to give this another iteration of changes once this merges.

da-kami:
> Well done. I like how the tiles are evolving.
> 
> As mentioned on a call: we know already that this layout won't work once the user can go long and short because we will need to display 2 exchange rates. I think the approach of having only 1 exchange rate is nice, but maybe we should just remove the `Short` button for now.

I opened https://github.com/itchysats/itchysats/issues/1163 and will tackle it as follow-up

taker-frontend/src/components/History.tsx
da-kami: Since this requires a bit more work I will do it as follow-up, see: https://github.com/itchysats/itchysats/issues/1164

taker-frontend/src/components/Trade.tsx
da-kami: Reduced it to `xl` - might change this with working on the prices again

klochowicz:
@da-kami you might need to rebase on recent master, as bors is now awaiting status from "block-fixup" job, that recently got introduced (which blocks merging `fixup!` commits)

taker-frontend/src/App.tsx
da-kami: extracted function `isClosed(cfd: Cfd): boolean` - tried to have the `Cfd` as `class` first so I can have a method, but could not easily achieve the mapping using the class, so I opted for a function that reasons about the `Cfd` instead.

PR #1069, opened 2022-01-11
Remove unused_must_use allow

thomaseizinger:
This is no longer needed because all event messages now return `()`!


klochowicz (approved):
nice!

luckysori (approved):

PR #1073, opened 2022-01-11
Use `command::Executor` in more places

thomaseizinger:
Bit of a quick and dirty application. Curious to hear what people think.

### DIFF
--- a/daemon/src/command.rs
+++ b/daemon/src/command.rs
@@ -89,3 +89,11 @@ impl<TOne, TTwo> ExtractEventFromTuple for (Event, TOne, TTwo) {
         (Some(self.0), (self.1, self.2))
     }
 }
+
+impl<TOne, TTwo, TThree> ExtractEventFromTuple for (Event, TOne, TTwo, TThree) {
+    type Rest = (TOne, TTwo, TThree);
+
+    fn extract_event(self) -> (Option<Event>, Self::Rest) {
+        (Some(self.0), (self.1, self.2, self.3))
+    }
+}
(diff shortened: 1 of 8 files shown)


daemon/src/setup_maker.rs
```diff
@@ -69,8 +67,7 @@ impl Actor {
         ),
     ) -> Self {
         Self {
-            db,
-            process_manager,
+            executor: command::Executor::new(db, process_manager),
```
klochowicz, line 70: perhaps we could clone the executor and pass it instead of the `db` and `process_manager` to encapsulate what can be done with them?

daemon/src/command.rs
```diff
@@ -34,14 +34,66 @@ impl Executor {
...
+                .context("Failed to process new domain event")?;
+        }
+
+        Ok(rest)
+    }
+}
+
+// TODO: Delete this weird thing once all our commands return only an `Event` and not other stuff as
+// well.
+pub trait ExtractEventFromTuple {
```
klochowicz, line 55: wow. ingenious effort. It's weird, but at least it puts all the weirdness in a single spot!

klochowicz (approved):
I like it. It is a step in the right direction. a lot of boilerplate goes away!

daemon/src/command.rs
thomaseizinger: Figured that that would impress you :)

daemon/src/setup_maker.rs
thomaseizinger: Yes. I just wanted to keep the noise down in the PR and not ripple the changes through too much!

da-kami (approved):
Noice! let's get this in

PR #1078, opened 2022-01-12
Fix taker disconnected toast

rishflab:
we also get this error in the console. I think we are triggering too many re-renders.
```
The connection to http://localhost:3000/api/feed was interrupted while the page was loading. useEventSource.ts:8:19
```

### DIFF
--- a/taker-frontend/src/useEventSource.ts
+++ b/taker-frontend/src/useEventSource.ts
@@ -10,6 +10,12 @@ export function useEventSource(url: string, withCredentials?: boolean) {
         const es = new EventSource(url, { withCredentials });
         setSource(es);
 
+        if (es.OPEN === 1) {
+            setIsConnected(true);
+        } else if (es.CLOSED || es.CONNECTING === 1) {
+            setIsConnected(false);
+        }
+
         es.addEventListener("error", () => {
             setIsConnected(false);
             setSource(null);
@@ -17,6 +23,7 @@ export function useEventSource(url: string, withCredentials?: boolean) {
 
         return () => {
             setSource(null);
+            es.removeEventListener("error", () => {});
             es.close();
         };
     }, [url, withCredentials]);


taker-frontend/src/useEventSource.ts
```diff
@@ -12,11 +12,16 @@ export function useEventSource(url: string, withCredentials?: boolean) {
 
         es.addEventListener("error", () => {
             setIsConnected(false);
+            // we call set source here which changes the state of this component causing a re-render
```
thomaseizinger: This only happens on error so I think it is fine to change the source here. It will be useless from here on!

taker-frontend/src/useEventSource.ts
```diff
@@ -12,11 +12,16 @@ export function useEventSource(url: string, withCredentials?: boolean) {
...
+            // as far as i understand this will be called everytime an event is sent from the backend
+            // have not investigated whether the backend is sending this event multiple times
```
luckysori: Calling `setSource` shouldn't trigger a re-render AFAIK, because `source` isn't part of the dependencies of this hook.

taker-frontend/src/useEventSource.ts
```diff
@@ -53,6 +58,8 @@ export function useEventSource(url: string, withCredentials?: boolean) {
     }, [source]);
 
     const toast = useToast();
+    // I dont think should be a hook.
+    // We should just conditionally render a component that displays this message
```
luckysori: If we want to keep using `useToast` we cannot use it inside a conditional.

luckysori (commented):
Just a couple of notes on the first commit because I implemented the original solution.

Nothing against the suggested direction in the second commit, but I'm no expert.

taker-frontend/src/useEventSource.ts
thomaseizinger: I think it will trigger a re-render outside of this hook for everything that depends on source!

taker-frontend/src/useEventSource.ts
```diff
@@ -1,15 +1,17 @@
-import { useToast } from "@chakra-ui/react";
 import { useEffect, useRef, useState } from "react";
 
-export function useEventSource(url: string, withCredentials?: boolean) {
+export function useEventSource(url: string, withCredentials?: boolean, setIsConnected) {
     const [source, setSource] = useState<EventSource | null>(null);
-    const [isConnected, setIsConnected] = useState<boolean>(true);
```
thomaseizinger: I think it would be better to keep this state in here and expose it additionally with the `source`, i.e. return 

```ts
{ source: EventSource, isConnected: boolean }
````

File: taker-frontend/src/App.tsx [hunk 5]
```diff
@@ -96,6 +98,7 @@ export const App = () => {
         <>
             <Disclaimer />
             <Nav walletInfo={walletInfo} connectedToMaker={connectedToMaker} />
+            <p>connected to taker: {isConnected.toString()}</p>
```
thomaseizinger: I'd like to keep the toast for now but make it persistent!

taker-frontend/src/useEventSource.ts
thomaseizinger: We can't call `useToast` conditionally but we can conditionally enable it by calling `toast()`!

luckysori: Isn't that what we are already doing?

thomaseizinger: I think every call to `toast` creates a new one. If we want to update it we would need to `useRef`: https://chakra-ui.com/docs/feedback/toast#updating-toasts

There is also some talk about setting the `id` property of a toast to prevent duplicates.

taker-frontend/src/useEventSource.ts
luckysori: That's true, but it was also intended so that we keep checking for heartbeats after we get an `EventSource`.

There's probably a better way of doing all this and I'm not surprised we've encountered a bug, but the logic still makes sense to me.

taker-frontend/src/useEventSource.ts
klochowicz: https://github.com/itchysats/itchysats/blob/master/maker-frontend/src/components/BackendMonitor.tsx#L31 - this is how we can reuse the same toast without creating duplicates

taker-frontend/src/useEventSource.ts
thomaseizinger: Yeah this logic also makes sense to me!

taker-frontend/src/useEventSource.ts
luckysori: I had something like what you guys are suggesting at some point, but I removed it because I never got duplicates when testing with the current version 🤷 I'm all for re-introducing that.

taker-frontend/src/useEventSource.ts
```diff
@@ -10,13 +10,20 @@ export function useEventSource(url: string, withCredentials?: boolean) {
         const es = new EventSource(url, { withCredentials });
         setSource(es);
 
+        if (es.OPEN === 1) {
+            setIsConnected(true);
+        } else if (es.CLOSED || es.CONNECTING === 1) {
+            setIsConnected(false);
+        }
```
thomaseizinger, line 17: This might be better represented as a `switch` perhaps?

thomaseizinger (approved):

luckysori (approved):
Wait this fixes the bug? Haha. I have such a hard time understanding how this stuff works.

rishflab:
closed in favour of #1090

PR #1081, opened 2022-01-12
Ensure we don't settle with an outdated price

### LINKED ISSUE #1057: Ensure quote is recent before collaborative settlement
If it is older than 1 minute, we consider the quote outdated and need to fail the HTTP handler.

Make sure to write a nice message because it is going to be displayed to the user.

Provides visibility into whether https://github.com/itchysats/itchysats/issues/985 is still happening.

thomaseizinger:
Fixes #1057.

### DIFF
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -9,10 +9,11 @@ use futures::TryStreamExt;
 use rust_decimal::Decimal;
 use std::convert::TryFrom;
 use std::time::Duration;
+use time::OffsetDateTime;
 use tokio_tungstenite::tungstenite;
 use xtra_productivity::xtra_productivity;
 
-const URL: &str = "wss://www.bitmex.com/realtime?subscribe=quoteBin1m:XBTUSD";
+pub const QUOTE_INTERVAL_MINUTES: i64 = 1;
 
 pub struct Actor {
     tasks: Tasks,
@@ -39,7 +40,7 @@ impl xtra::Actor for Actor {
             async move {
                 tracing::debug!("Connecting to BitMex realtime API");
 
-                let mut connection = match tokio_tungstenite::connect_async(URL).await {
+                let mut connection = match tokio_tungstenite::connect_async(format!("wss://www.bitmex.com/realtime?subscribe=quoteBin{}m:XBTUSD", QUOTE_INTERVAL_MINUTES)).await {
                     Ok((connection, _)) => connection,
                     Err(e) => {
                         let _ = this.send(StopReason::FailedToConnect { source: e }).await;
@@ -174,6 +175,12 @@ impl Quote {
     fn mid_range(&self) -> Price {
         (self.bid + self.ask) / 2
     }
+
+    pub fn is_older_than(&self, duration: time::Duration) -> bool {
+        let required_quote_timestamp = (OffsetDateTime::now_utc() - duration).unix_timestamp();
+
+        self.timestamp.seconds() < required_quote_timestamp
+    }
 }
 
 mod wire {
@@ -202,6 +209,7 @@ mod wire {
 mod tests {
     use super::*;
     use rust_decimal_macros::dec;
+    use time::ext::NumericalDuration;
 
     #[test]
     fn can_deserialize_quote_message() {
@@ -210,5 +218,31 @@ mod tests {
         assert_eq!(quote.bid, Price::new(dec!(42640.5)).unwrap());
         assert_eq!(quote.ask, Price::new(dec!(42641)).unwrap());
         assert_eq!(quote.timestamp.seconds(), 1632192000)
+    }
+
+    #[test]
+    fn quote_from_now_is_not_old() {
+        let quote = dummy_quote_at(OffsetDateTime::now_utc());
+
+        let is_older = quote.is_older_than(1.minutes());
+
+        assert!(!is_older)
+    }
+
+    #[test]
+    fn quote_from_one_hour_ago_is_old() {
+        let quote = dummy_quote_at(OffsetDateTime::now_utc() - 1.hours());
+
+        let is_older = quote.is_older_than(1.minutes());
+
+        assert!(is_older)
+    }
+
+    fn dummy_quote_at(time: OffsetDateTime) -> Quote {
+        Quote {
+            timestamp: Timestamp::new(time.unix_timestamp()),
+            bid: Price::new(dec!(10)).unwrap(),
+            ask: Price::new(dec!(10)).unwrap(),
+        }
     }
 }
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -1,6 +1,5 @@
 use crate::model::Price;
 use crate::model::Timestamp;
-use crate::projection;
 use crate::supervisor;
 use crate::Tasks;
 use anyhow::Result;
@@ -11,25 +10,21 @@ use rust_decimal::Decimal;
 use std::convert::TryFrom;
 use std::time::Duration;
 use tokio_tungstenite::tungstenite;
-use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 
 const URL: &str = "wss://www.bitmex.com/realtime?subscribe=quoteBin1m:XBTUSD";
 
 pub struct Actor {
     tasks: Tasks,
-    receiver: Box<dyn MessageChannel<projection::Update<Quote>>>,
+    latest_quote: Option<Quote>,
     supervisor: xtra::Address<supervisor::Actor<Self, StopReason>>,
 }
 
 impl Actor {
-    pub fn new(
-        receiver: impl MessageChannel<projection::Update<Quote>> + 'static,
-        supervisor: xtra::Address<supervisor::Actor<Self, StopReason>>,
-    ) -> Self {
+    pub fn new(supervisor: xtra::Address<supervisor::Actor<Self, StopReason>>) -> Self {
         Self {
             tasks: Tasks::default(),
-            receiver: Box::new(receiver),
+            latest_quote: None,
             supervisor,
         }
     }
@@ -40,7 +35,6 @@ impl xtra::Actor for Actor {
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         self.tasks.add({
             let this = ctx.address().expect("we are alive");
-            let receiver = self.receiver.clone_channel();
 
             async move {
                 tracing::debug!("Connecting to BitMex realtime API");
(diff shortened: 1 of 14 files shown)


daemon/src/lib.rs
```diff
@@ -476,6 +477,10 @@ where
             .context("Price feed not available")?
             .context("No quote available")?;
 
+        if latest_quote.is_older_than(1.minutes()) {
```
luckysori: 🤔 Perhaps a constant would make sense here. Also the error message could change depending on its value.

daemon/src/routes_taker.rs
```diff
@@ -100,9 +89,8 @@ pub async fn feed(
                     let quote = rx_quote.borrow().clone();
                     yield quote.to_sse_event();
                 }
-                Ok(()) = rx_keep_alive.changed() => {
-                    let keep_alive = *rx_keep_alive.borrow();
-                    yield keep_alive.to_sse_event();
+                _ = heartbeat.tick() => {
+                    yield Heartbeat::new().to_sse_event();
```
luckysori, line 92: 👍 Hehe this is simpler 🤦

luckysori (approved):
LGTM!

thomaseizinger:
Tests hang, convert to draft.

daemon/tests/happy_path.rs
```diff
@@ -25,45 +26,25 @@ use std::time::Duration;
...
-            taker_cfd.state, maker_cfd.state,
-            "cfd state mismatch between maker and taker"
-        );
-        assert_eq!(taker_cfd.order_id, $id, "unexpected order id in the taker");
-        assert_eq!(maker_cfd.order_id, $id, "unexpected order id in the maker");
-        assert_eq!(taker_cfd.state, $state, "unexpected cfd state in the taker");
-        assert_eq!(maker_cfd.state, $state, "unexpected cfd state in the maker");
-    };
+/// Waits until the CFDs for both maker and taker are in the given state.
+macro_rules! wait_next_state {
```
da-kami, line 30: I am not against this change, but I want to point out that 

> This can make the tests brittle if there is another update in the feed inbetween where the state is not yet the expected one.

this should not happen! What we call "brittle" here might actually be a flaw in the logic. With letting the mocked actors return very specific messages at specific points in time we can ensure that we receive the right message at the right time. I think we should always be able to ensure this. 
As @thomaseizinger pointed out we actually have e.g. quote updates that *do* send out projection updates independent from the tests. But I wonder if that *should* happen - shouldn't we be able to control this in the tests through the mocks? 

We *can* change the test setup to wait for the next state and ignore "other" state changes in between, but that makes the test less "restrictive" in testing the behavior.

daemon/tests/harness/flow.rs
```diff
@@ -8,7 +8,7 @@ use std::time::Duration;
 use tokio::sync::watch;
 
 /// Waiting time for the time on the watch channel before returning error
-const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) { 120 } else { 30 });
+const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) { 180 } else { 30 });
```
da-kami, line 11: I understand the reasoning in the commit message, but I am surprised that we only notice this *now*. 
Before we did event sourcing and concurrent actor setup I feel we did not run into these problems (neither locally nor on CI). Maybe it's a coincidence, but I still find it weird that we now set `180` when at some point `60` was enough to "not be brittle".

daemon/src/projection.rs
```diff
@@ -56,7 +61,12 @@ pub struct Feeds {
 }
 
 impl Actor {
-    pub fn new(db: sqlx::SqlitePool, _role: Role, network: Network) -> (Self, Feeds) {
+    pub fn new(
+        db: sqlx::SqlitePool,
+        _role: Role,
```
da-kami, line 71: Why do we pass the role in here if we are not using it?

daemon/tests/happy_path.rs
```diff
@@ -25,45 +26,25 @@ use std::time::Duration;
...
-            taker_cfd.state, $taker_state,
-            "unexpected cfd state in the taker"
-        );
-        assert_eq!(
-            maker_cfd.state, $maker_state,
-            "unexpected cfd state in the maker"
-        );
+    };
+    ($id:expr, $maker:expr, $taker:expr, $state:expr) => {
+        wait_next_state!($id, $maker, $taker, $state, $state)
```
luckysori, line 47: ❓ Why does this branch of the macro not use `one_cfd_with_state` like above?

daemon/tests/harness/flow.rs
```diff
@@ -62,15 +37,56 @@ pub async fn next<T>(rx: &mut watch::Receiver<T>) -> Result<T>
...
+    let wait_until_predicate = async {
+        loop {
+            rx.changed().await?;
+
+            let current = rx.borrow().clone();
+
+            match filter_map(current) {
+                Some(val) => return anyhow::Ok(val),
+                None => {
+                    rx.changed().await?;
```
luckysori: ❓ Why don't we continue onto the next iteration of the loop and wait for `changed` there?

luckysori (approved):
Looks good again, just some questions about the first few new commits.

daemon/src/bitmex_price_feed.rs
```diff
@@ -212,4 +220,30 @@ mod tests {
...
+        let is_older = quote.is_older_than(1.minutes());
+
+        assert!(!is_older)
+    }
+
+    #[test]
+    fn quote_from_one_hour_ago_is_old() {
+        let quote = dummy_quote_at(OffsetDateTime::now_utc() - 1.hours());
+
+        let is_older = quote.is_older_than(1.minutes());
```
da-kami, line 236: nit: The test would be more meaningful if it would test the boundaries. E.g.:

```
let quote = dummy_quote_at(OffsetDateTime::now_utc() - 2.minutes());

let is_older = quote.is_older_than(1.minutes());
```

might also want to cover more boundaries.

da-kami (approved):
Thanks! few comments, nothing blocking

daemon/tests/harness/flow.rs
thomaseizinger: So it turns out that the semantics of waiting for the next change are still useful for some tests where we don't use the macro.

daemon/src/projection.rs
thomaseizinger: Not sure, was already there as you can see. Do a follow-up refactor?

daemon/tests/happy_path.rs
thomaseizinger: This recurses. How should I use the filter-map function here?

daemon/tests/harness/flow.rs
thomaseizinger: This became more apparent by using spawn-blocking because I think before we were actually blocking the timer future so time did simply not progress!

daemon/src/bitmex_price_feed.rs
thomaseizinger: Well, it is really a nitpick isn't it :D

The test makes sure we are doing the comparison correctly, i.e. older and not more recent. The fact that it uses one minute is pretty obvious from the impl so risk is low, hence I didn't bother :)

daemon/tests/happy_path.rs
thomaseizinger: > this should not happen! What we call "brittle" here might actually be a flaw in the logic.

I don't think it is as clear cut to say that it shouldn't happen.

We cannot and IMO should not have control over ever single message that is sent in the system. That would require exposing way too many knobs and mocks. We have to accept that the actor system we are building are eventually consistent, meaning things will eventually be in the correct state if it works correctly.

> We _can_ change the test setup to wait for the next state and ignore "other" state changes in between, but that makes the test less "restrictive" in testing the behavior.

It would be fairly trivial to pass a list of states that it should go through before arriving at the final state. But outside of that, I don't think it is particularly likely for us to be in an unexpected state and then actually end up in the correct state in the end. And even if, is that really a bug or is that kind of self-healing functionality a feature? I don't think the user is gonna care much ...

daemon/tests/happy_path.rs
luckysori: Lol, please ignore me. I can't read code anymore.

daemon/tests/harness/flow.rs
luckysori: I'm not getting you. Why are we awaiting _twice_ in a row for another change after not finding the `expected_state`. Wouldn't this code just ignore one state update?

thomaseizinger: Ah you are right. That doesn't make any sense!

luckysori (approved):
Thanks for fixing that!

daemon/src/bitmex_price_feed.rs
da-kami: 🤷‍♂️

daemon/tests/harness/flow.rs
```diff
@@ -62,15 +37,53 @@ pub async fn next<T>(rx: &mut watch::Receiver<T>) -> Result<T>
...
+/// # Panics
+///
+/// If there is more than one CFD in the list. This is unsupported and unexpected by our test
+/// framework.
+pub fn one_cfd_with_state(expected_state: CfdState) -> impl Fn(Vec<Cfd>) -> Option<Cfd> {
+    move |cfds: Vec<Cfd>| match cfds.as_slice() {
+        [one] if one.state == expected_state => Some(one.clone()),
+        [_one_that_doesnt_match_state] => None,
+        [] => None,
+        _more_than_one => panic!("More than one CFD in feed!"),
```
klochowicz, line 87: never seen this syntax before - what does it mean? are the `[]` destructuring the array of cfds?

daemon/tests/harness/mod.rs
```diff
@@ -310,15 +325,23 @@ macro_rules! deliver_event {
...
 
 pub fn dummy_price() -> Price {
     Price::new(dec!(50_000)).expect("to not fail")
 }
 
+pub fn dummy_quote() -> Quote {
+    Quote {
+        timestamp: Timestamp::now(),
+        bid: Price::new(dec!(50_000)).expect("to not fail"),
+        ask: Price::new(dec!(50_000)).expect("to not fail"),
```
klochowicz, line 341: nit: would probably use the `dummy_price()` inside dummy quote, no need to have more magic numbers than necessary

klochowicz (approved):
nice, thanks for waiting for me to have a look ;)

daemon/tests/harness/flow.rs
thomaseizinger: This is called sub-slice pattern matching: https://github.com/rust-lang/rfcs/blob/master/text/2359-subslice-pattern-syntax.md

PR #1082, opened 2022-01-12
Make actions a hashmap of action => link

thomaseizinger:
Fixes a TODO that is in the projection code.

What is everyone's thought on this?

Draft for now because the UI and automation needs to be updated for this!
If we agree that this is a good idea, when will implement in a backwards compatible way so we don't have to update in lockstep.


daemon/src/routes_maker.rs
```diff
@@ -123,31 +122,105 @@ pub async fn post_sell_order(
...
+}
+
+#[rocket::post("/cfd/<id>/reject")]
+pub async fn reject_order(
+    id: OrderId,
+    maker: &State<Maker>,
+    _auth: Authenticated,
+) -> Result<(), HttpApiProblem> {
+    maker.reject_order(id).await.map_err(|e| {
+        HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
```
thomaseizinger, line 147: This should probably not be a 500.

daemon/src/projection.rs
```diff
@@ -163,7 +167,7 @@ pub struct Cfd {
     pub profit_percent: Option<String>,
 
     pub state: CfdState,
-    pub actions: Vec<CfdAction>, // TODO: This should be a HashMap.
```
thomaseizinger, line 166: We should make this a backwards-compatible API change.

klochowicz, line 166: Hm.... should this be a HashMap though? It strikes me that projection does not (or perhaps even worded more strongly, *should not*) know how the internals work.

I like the type safety improvements, but it seems that a HashMap of "action" -> "URI" (a dictionary :) ) transformation should happen as a separate lookup, you don't want to duplicate the URI scheme creation whenever an action is available. Nor the person stating which ones are available should care about how the URIs/routes are derived

klochowicz (commented):
lots of cool stuff here, but I have some doubts about having just a single hashmap. happy to chat about it later today!

daemon/src/projection.rs
thomaseizinger: Perhaps it is not clear what problem we are solving here!

The issue with a list of links are:

1. It is harder to build iteratively within the `apply` method because you might end up adding duplicates if you don't check upfront. A `HashMap` by design doesn't allow duplicate entries so using this data structure seems better. In other words, each event that is being applied can simply reason about "is this action available after this event" and add it, not having to consider whether or not it was added by a previous event.
2. The client (our TS code) should be programmed against links being available and not care about the structure of the link. At the moment we send out actions and the client builds the link. If we only send out links, then the client can no longer assign meaning to the individual links without hardcoding them. For example, we need to differentiate between `commit` and `settle` in the frontend to display a different text. If we only have links, then the client would need to either inspect the link again or compare against the entire link which defeats the point of treating links as opaque identifiers! With a hashmap, the client can program against the `key` in the map (does map contain action `settle`?) and conditionally enable UI elements as a result while still treating the link as opaque.

The reason we are doing the URI creation here is because we want type-safety that the links we are sending out actually map to endpoints that are available.

thomaseizinger: > It strikes me that projection does not (or perhaps even worded more strongly, _should not_) know how the internals work.

I saw the projection module as tightly coupled to our rocket layer but perhaps there is actually another layer missing and this should only be a `HashSet` so we still get the "at most once" semantics ...

thomaseizinger: The consequence of that is though that our projection data structures should **not** be serializable.

klochowicz: I didn't say that `Vec` was the solution :) I think it should be a `HashSet` actually! and then a lookup in the HashMap (the "schema" or "dictionary" of how to create a link from it)

thomaseizinger: Agreed! I'll submit a new PR!

Issue #1064, opened 2022-01-10
CI occassionally fails on tests in debug mode


klochowicz:
The actor tests running on a single thread are problematic as they sometimes result in failures due to hitting timeouts.

Some proposals to fix it
- use `tokio::spawn_blocking` for signing and verifying during contract setup
- use multi-threaded tokio scheduler in tests (this would require adjusting tracing config)

Alternatively, as a temporary workaround:
- increase the timeout yet another time (we already did from 60s to 70s, it improved things but not fully)

da-kami:
I am surprised we hit the `70s` timeout so often - are we sure there are no other (blocking / starving) issues that cause this problem? 

The assumption is, that loading from the database increases the runtime with the event sourcing changes - but I am not totally convinced that is the actual issue. We can give multi-threaded environments a try, but might also want to look into what could potentially starve message-sending during contract setup.

klochowicz:
IMHO the commits referenced above have nothing to do with the issue posted above - it got fixed last Friday, with #1084.

PR #1084, opened 2022-01-12
Run CPU expensive operations on a dedicated thread 

thomaseizinger:
Some knowledge about Rust captured in the 2nd commit message.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -166,15 +166,25 @@ pub async fn new(
         )?,
     )]);
 
-    let own_cfd_txs = create_cfd_transactions(
-        (params.maker().clone(), *params.maker_punish()),
-        (params.taker().clone(), *params.taker_punish()),
-        oracle_pk,
-        (CET_TIMELOCK, setup_params.refund_timelock),
-        payouts,
-        sk,
-        setup_params.fee_rate,
-    )
+    let own_cfd_txs = tokio::task::spawn_blocking({
+        let maker_params = params.maker().clone();
+        let taker_params = params.taker().clone();
+        let maker_punish = *params.maker_punish();
+        let taker_punish = *params.taker_punish();
+
+        move || {
+            create_cfd_transactions(
+                (maker_params, maker_punish),
+                (taker_params, taker_punish),
+                oracle_pk,
+                (CET_TIMELOCK, setup_params.refund_timelock),
+                payouts,
+                sk,
+                setup_params.fee_rate,
+            )
+        }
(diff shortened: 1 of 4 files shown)


da-kami:
Is this a complete fix for https://github.com/itchysats/itchysats/issues/1064 ?

thomaseizinger:
> Is this a complete fix for https://github.com/itchysats/itchysats/issues/1064 ?

Not yet.

thomaseizinger:
> Is this a complete fix for #1064 ?

I am not sure if it is a fix for this at all. It fixes the heartbeats not being sent in the test. If the cause for #1064 are scheduling issues, then might also fix it but I haven't looked into #1064 specifically.

daemon/tests/happy_path.rs
```diff
@@ -353,15 +353,8 @@ async fn maker_notices_lack_of_taker() {
 /// For convenience, returns also OrderId of the opened Cfd.
 /// `announcement` is used during Cfd's creation.
 async fn start_from_open_cfd_state(announcement: oracle::Announcement) -> (Maker, Taker, OrderId) {
-    let heartbeat_interval = Duration::from_secs(60);
```
klochowicz, line 356: thanks for cleaning this - this should have been cleared before, I just didn't notice it inside this function! this might have been the reason why some of the tests were failing - `heartbeat_inverval * 2` was a bug (it should have been just `heartbeat_interval`, we changed the API recently).

As such this commit is unrelated to the PR, but still very much desired.

daemon/src/setup_contract.rs
```diff
@@ -166,15 +166,25 @@ pub async fn new(
         )?,
```
klochowicz, line 166: Thank you very much for the clear write-up, it's great.

One question: will it actually spawn a separate thread even if we don't instruct `tokio::test` to be configured with multiple threads, like in the snippet below?

```rust
#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```

klochowicz (approved):
thanks for dealing with this! looks good :)

daemon/src/setup_contract.rs
thomaseizinger: Yes, the threadpool for blocking operations is completely separate from that. See the docs for more details: https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

klochowicz: cool!

luckysori (approved):
Thank you for sharing the knowledge!

PR #1089, opened 2022-01-13
Remove margin calc endpoint

thomaseizinger:
This is a breaking API change but only on the taker and we control the only client for that atm.


da-kami (approved):
I think this PR should remove `MarginRequestPayload` and `MarginResponse` from taker UI's `types.ts`

This somewhat breaks our paradigm of doing no calculations in the UI, but it simplifies handling the margin and cost calc. I will see to rebase my work on it.

thomaseizinger:
> This somewhat breaks our paradigm of doing no calculations in the UI, but it simplifies handling the margin and cost calc. I will see to rebase my work on it.

I agree but they are at least very simple calculations and for display purposes only.

thomaseizinger:
> I think this PR should remove `MarginRequestPayload` and `MarginResponse` from taker UI's `types.ts`

Done.

PR #1090, opened 2022-01-13
Prevent taker disconnected toast spam 2

### LINKED ISSUE #1047: Disconnected from feed is triggered *a lot*
What I did:

1. Start taker daemon without maker running
2. Start taker UI

Expected behaviour:

The taker does not have connection to maker (because the maker was not started).
The taker UI displays "limited functionality", i.e. the warning `No maker! You are not connected to any maker. Functionality may be limited` is shown, but the user interface is usable.

Actual behaviour:

`No maker! You are not connected to any maker. Functionality may be limited` is shown but additionally I keep getting toasts that claim I am not connected to the feed. That is incorrect. The taker UI is connected to the taker daemon. These toasts should not be triggered. The user is advised to refresh, but refresh results in the same behaviour.

Due to the problem described in #1042 the user interface is not really usable.

I am not entirely sure that the problem of the taker not having a connection to the maker causes this problem. But, in that scenario it was reproduceable. I saw the error of not being connected to the daemon generally very frequently - could it be that we are reacting to *all* error scenarios somehow and always print that error?

image

rishflab:
I think this is an improvement over #1078. I actually understand why it works now and it (at least to me) a lot simpler.

I don't think we should be using a toast to display connection status. From what I read toasts are for transient non-critical notifications.

two things are broken:
- [x] I don't know how to align the alert box in the center

Screenshot from 2022-01-13 14-49-53

- [x]  I don't what type argument to use for `setIsConnected` in the `useEventSource` component.


taker-frontend/src/useEventSource.ts
```diff
@@ -15,61 +13,25 @@ export function useEventSource(url: string, withCredentials?: boolean) {
...
+            setIsConnected(true);
+            timer = setTimeout(() => {
+                setIsConnected(false);
+            }, 5000);
+        });
+
         return () => {
             setSource(null);
+            es.removeEventListener("error", () => {});
+            es.removeEventListener("heartbeat", () => clearTimeout(timer));
```
thomaseizinger: I think unless you are providing the exact same function here (as in: use a named function for the callback), this is not going to do what you want because an arrow function has a unique identity.

taker-frontend/src/useEventSource.ts
```diff
@@ -1,9 +1,7 @@
-import { useToast } from "@chakra-ui/react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 
-export function useEventSource(url: string, withCredentials?: boolean) {
+export function useEventSource(url: string, withCredentials?: boolean, setIsConnected) {
     const [source, setSource] = useState<EventSource | null>(null);
-    const [isConnected, setIsConnected] = useState<boolean>(true);
```
thomaseizinger, line 6: Why not keep the state inside and return `isConnected` together with the event source?

taker-frontend/src/useEventSource.ts
thomaseizinger: See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener.

taker-frontend/src/useEventSource.ts
```diff
@@ -1,9 +1,7 @@
-import { useToast } from "@chakra-ui/react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
 
-export function useEventSource(url: string, withCredentials?: boolean) {
+export function useEventSource(url: string, withCredentials?: boolean, setIsConnected) {
```
thomaseizinger: I think we could easily remove the `withCredentials` argument and always set it to true inside here. We never need this _without_ that flag set.

taker-frontend/src/components/Trade.tsx
```diff
@@ -68,19 +65,6 @@ interface TradeProps {
...
-    description: string;
-}
-
-function AlertBox({ title, description }: AlertBoxProps) {
-    return (<Alert status="error">
-        <AlertIcon />
-        <AlertTitle mr={2}>{title}</AlertTitle>
-        <AlertDescription>{description}</AlertDescription>
-    </Alert>);
-}
```
thomaseizinger, line 82: Extracting something like this could easily go into a separate commit. Makes review easier if refactorings are separate from features.

taker-frontend/src/App.tsx
```diff
@@ -92,11 +93,20 @@ export const App = () => {
         500,
     );
 
+    let connectionStatus;
+    if (!isConnected) {
+        connectionStatus = <AlertBox
+            title={"Connection error!"}
+            description={"Please ensure taker daemon is running and refresh page"}
+        />;
+    }
```
thomaseizinger, line 102: Can you attach a screenshot of what this looks like when we are disconnected?

taker-frontend/src/useEventSource.ts
rishflab: In the react docs they say move state up and pass the the set functions to the children. I don't know the pros and cons of each method.

taker-frontend/src/useEventSource.ts
rishflab: we never use a different url either.

taker-frontend/src/useEventSource.ts
thomaseizinger: Generally yes. Move it up as far as where you need to modify it. You don't actually call `setIsConnected` outside of this hook so there isn't really any point tracking the state outside.

taker-frontend/src/useEventSource.ts
thomaseizinger: Fair point. Feels weird though to hardcode the URL.

thomaseizinger:
> I don't know how to align the alert box in the center

You need to move it further down the component tree inside the `Center` component.

taker-frontend/src/useEventSource.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export function useEventSource(url: string, withCredentials?: boolean) {
+export function useEventSource(url: string) {
     const [source, setSource] = useState<EventSource | null>(null);
     const [isConnected, setIsConnected] = useState<boolean>(true);
 
     // Construct a new event source if the arguments to this hook change
     useEffect(() => {
-        const es = new EventSource(url, { withCredentials });
+        const es = new EventSource(url);
```
thomaseizinger: We need to set it to `true`, otherwise our feed is not authenticated!

dprint.json
```diff
@@ -14,7 +14,7 @@
   "excludes": ["**/target", "**/sqlx-data.json", "maker-frontend/dist", "taker-frontend/dist", "**/node_modules"],
   "plugins": [
     "https://plugins.dprint.dev/markdown-0.9.2.wasm",
-    "https://plugins.dprint.dev/rustfmt-0.4.0.exe-plugin@c6bb223ef6e5e87580177f6461a0ab0554ac9ea6b54f78ea7ae8bf63b14f5bc2",
+    "rustfmt.exe-plugin@7723878c1be78a571c37dcb7f92a0c5780fcfd714a23498260b575efb7e610f3",
```
thomaseizinger: I guess you are changing this because of some NixOS stuff? :)

Perhaps look into [`git update-index --assume-unchanged`](https://www.git-scm.com/docs/git-update-index). That would allow you to make these changes locally and avoid accidentally committing them!

taker-frontend/src/App.tsx
```diff
@@ -106,14 +106,26 @@ export const App = () => {
...
-                {connectionStatus}
                 <Routes>
-                    <Route path="/wallet" element={<Wallet walletInfo={walletInfo} />} />
+                    <Route
+                        path="/wallet"
+                        element={<>
+                            <Center>
+                                <VStack>
+                                    {connectionStatus}
+                                    <Wallet walletInfo={walletInfo} />
```
thomaseizinger, line 116: I think both of these components already have a `Center` in them somewhere. Maybe it would be better to just pass the data (`isConnected`) down and render it on each page separately?

thomaseizinger (changes requested):
One thing regarding `withCredentials`, otherwise LGTM!

dprint.json
rishflab: yeh the plugin doesnt run on nixos :(

thomaseizinger (approved):

klochowicz:
thanks for pushing on this, after having a second thought indeed using an alert box for this purpose is more correct!

luckysori (approved):
Thanks for fixing this!

taker-frontend/src/App.tsx
```diff
@@ -13,6 +13,7 @@ import * as React from "react";
 import { useState } from "react";
```
klochowicz, line 13: The commit description doesn't say why you're using an alert box instead of a toast - it would be much nicer if it actually stated the reasoning you used in the PR comment - ie. that it's a persistent problem and toasts are used for transient notifications.

klochowicz (approved):
just one suggestion re commit message

Issue #1094, opened 2022-01-14
Do not hard-code `{Setup,Rollover}Params` fee rate


luckysori:
After the event-sourcing refactor, we started to hard-code the `fee_rate` used to calculate fees in `maia` for contract setup and rollover.

The three places where we currently hard-code the `fee_rate` are:

https://github.com/itchysats/itchysats/blob/849226df74d3cf43005d178222a40db49ddab140/daemon/src/model/cfd.rs#L635

https://github.com/itchysats/itchysats/blob/849226df74d3cf43005d178222a40db49ddab140/daemon/src/model/cfd.rs#L668

https://github.com/itchysats/itchysats/blob/849226df74d3cf43005d178222a40db49ddab140/daemon/src/model/cfd.rs#L695

Without this, waiting for transactions could take too long.

Some of the previously closed issues that introduced the functionality originally: #830 #677.

klochowicz:
duplicate of #1093 😀

luckysori:
> duplicate of #1093 :p

lol I just wanted to link to an issue in my PR: https://github.com/itchysats/itchysats/pull/1095 :(

da-kami:
We will need to communicate the fee from maker to taker (the maker being in control of the fees) for contract setup and rollover. 

For the contract setup we already have it in the offer. For rollover we will have to extend the accepted message from maker to taker.

PR #1095, opened 2022-01-14
Replace TODOs with GitHub issue

luckysori:
See https://github.com/itchysats/itchysats/issues/1094. This is one of the issues that were mentioned before, @rishflab.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -632,7 +632,7 @@ impl Cfd {
                 self.quantity,
                 self.leverage,
                 self.refund_timelock_in_blocks(),
-                1, // TODO: Where should I get the fee rate from?
+                1,
             ),
         ))
     }
@@ -665,7 +665,7 @@ impl Cfd {
                 self.quantity,
                 self.leverage,
                 self.refund_timelock_in_blocks(),
-                1, // TODO: Where should I get the fee rate from?
+                1,
             ),
             self.dlc.as_ref().ok_or(RolloverError::NoDlc)?.clone(),
             self.settlement_interval,
@@ -692,7 +692,7 @@ impl Cfd {
                 self.quantity,
                 self.leverage,
                 self.refund_timelock_in_blocks(),
-                1, // TODO: Where should I get the fee rate from?
+                1,
             ),
             self.dlc.as_ref().ok_or(RolloverError::NoDlc)?.clone(),
         ))


thomaseizinger (approved):

klochowicz (commented):
hm, without these TODOs  it will be harder to find all places by the person trying to fix this issue - not sure whether this is worth it, especially that there's virtually no risk of this lingering? 

Totally agree that there should have been an gh issue raised at the time this got merged to master though.

luckysori:
> hm, without these TODOs it will be harder to find all places by the person trying to fix this issue - not sure whether this is worth it, especially that there's virtually no risk of this lingering?
> 
> Totally agree that there should have been an gh issue raised at the time this got merged to master though.

I understand your point, but the TODO noise is annoying me in the codebase atm. What if I permalink to the 3 places where the fix needs to occur in the GH issue?

klochowicz:
> > hm, without these TODOs it will be harder to find all places by the person trying to fix this issue - not sure whether this is worth it, especially that there's virtually no risk of this lingering?
> > Totally agree that there should have been an gh issue raised at the time this got merged to master though.
> 
> I understand your point, but the TODO noise is annoying me in the codebase atm. What if I permalink to the 3 places where the fix needs to occur in the GH issue?

I think that there problem with the TODOs is happening because a lot of them are "someday/maybe" and these ones should be pruned without any hesitation. However, in this particular case, the task in one of our next tasks, potentially being actioned today (and definitely in the near future) - I don't get why make it harder for the assignee. 

If I was assigned, I would be grateful for these TODOs being left, and permalinks would be an annoying overhead as I would need to jump out of my IDE.

There is also the point that as per our latest agreement TODOs should not be in the code - I'm happy with you push this as it's valid from this point of view, but this PR made me think that perhaps we should relax this requirement (I'll raise a discussion at the next opportunity :) )

PS in this case it should be actually a `FIXME` 😅  (we know we're doing something wrong and it should have been fixed).

luckysori:
> There is also the point that as per our latest agreement TODOs should not be in the code - I'm happy with you push this as it's valid from this point of view, but this PR made me think that perhaps we should relax this requirement (I'll raise a discussion at the next opportunity :) )

I just assumed that we had already chosen to relax or ignore that agreement, since the number of TODOs doesn't seem to be going down.

I'll close this PR because it doesn't seem to be adding any value at the moment, but I will keep trying to remove TODOs if I get the chance.

klochowicz:
> > There is also the point that as per our latest agreement TODOs should not be in the code - I'm happy with you push this as it's valid from this point of view, but this PR made me think that perhaps we should relax this requirement (I'll raise a discussion at the next opportunity :) )
> 
> I just assumed that we had already chosen to relax or ignore that agreement, since the number of TODOs doesn't seem to be going down.
> 
> I'll close this PR because it doesn't seem to be adding any value at the moment, but I will keep trying to remove TODOs if I get the chance.

OK, I'll try to keep vigilant of meaningless TODOs too, so that the signal to noise ratio improves. Found a few now #1098 :)

PR #1100, opened 2022-01-14
Expose price of incoming settlement proposal

### LINKED ISSUE #990: Expose settlement price proposal via HTTP API
In our current model, the taker proposes a settlement price. The maker can accept or reject this request. In order to make an conscious  decision, the maker should know about the proposed settlement api which is currently only an internal implementation detail.

We should expose this information via the HTTP API. 

https://github.com/itchysats/itchysats/commit/5e4525debae6ce3298916cecde815e9e029fb149 implements this for 0.3.x in a maybe not perfect solution. There might be a better solution hence this was not done right away on master.

thomaseizinger:
Fixes #990.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -164,6 +164,9 @@ pub struct Cfd {
 
     pub counterparty: Identity,
 
+    #[serde(with = "round_to_two_dp::opt")]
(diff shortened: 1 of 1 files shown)


bonomat (approved):
👍

PR #1102, opened 2022-01-14
Pass in funding rate in daily percentage

klochowicz:
Hourly rollovers are charged a fraction of the daily amount, based on the current funding rate, e.g. `1/24 * 0.002`.

Note: Unit tests for calculating funding rates are coming in subsequent PR, as well as missing functionality. I'll open a separate issue to track the remaining backend issues.

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -1,5 +1,6 @@
 use crate::impl_sqlx_type_display_from_str;
 use crate::olivia;
+use crate::SETTLEMENT_INTERVAL;
 use anyhow::Context;
 use anyhow::Result;
 use bdk::bitcoin::Address;
@@ -39,6 +40,7 @@ pub enum Error {
     NegativePrice,
 }
 
+/// Represents "quantity" or "contract size" in Cfd terms
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
 pub struct Usd(Decimal);
 
@@ -600,6 +602,80 @@ impl Timestamp {
     }
 }
 
+/// Funding rate per SETTLEMENT_INTERVAL
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub struct FundingRate(Decimal);
+
+impl FundingRate {
+    pub fn new(rate: Decimal) -> Result<Self> {
+        anyhow::ensure!(
+            rate.is_sign_positive(),
+            "Negative funding rates (ie. paid by the maker) are not supported yet"
+        );
+        anyhow::ensure!(
+            rate <= Decimal::ONE,
+            "Funding rate can't be higher than 100%"
+        );
+
+        Ok(Self(rate))
+    }
+
+    pub fn to_decimal(&self) -> Decimal {
+        self.0
+    }
+}
+
+/// Fee paid on every contract renewal (a fraction of it if rolling over)
+#[derive(Debug, Clone)]
+pub struct FundingFee(u64);
+
+impl FundingFee {
+    pub fn as_u64(&self) -> u64 {
+        self.0
+    }
+}
+
+impl Default for FundingFee {
+    fn default() -> Self {
+        Self::new(Amount::ZERO, FundingRate(Decimal::ZERO), 1)
+            .expect("hard-coded values to be valid")
+    }
+}
+
+impl FundingFee {
+    pub fn new(margin: Amount, funding_rate: FundingRate, hours_to_charge: i64) -> Result<Self> {
+        anyhow::ensure!(hours_to_charge >= 1, "Can't charge for less than one hour");
+        anyhow::ensure!(
+            hours_to_charge <= SETTLEMENT_INTERVAL.whole_hours(),
+            "Can't change for more than a full settlement interval"
+        );
+        let fraction_of_funding_period =
+            if hours_to_charge as i64 == SETTLEMENT_INTERVAL.whole_hours() {
+                Decimal::ONE
+            } else {
+                Decimal::from(hours_to_charge)
+                    .checked_div(Decimal::from(SETTLEMENT_INTERVAL.whole_hours()))
+                    .context("can't establish a fraction")?
+            };
+        Ok(Self(calculate_funding_fee(
+            margin,
+            funding_rate,
+            fraction_of_funding_period,
+        )?))
+    }
+}
+
+fn calculate_funding_fee(
+    margin: Amount,
+    funding_rate: FundingRate,
+    fraction_of_funding_period: Decimal,
+) -> Result<u64> {
+    (Decimal::from(margin.as_sat()) * funding_rate.to_decimal() * fraction_of_funding_period)
+        .round_dp_with_strategy(0, rust_decimal::RoundingStrategy::AwayFromZero)
+        .to_u64()
+        .context("Failed to represent as u64")
+}
+
 #[cfg(test)]
 mod tests {
     use rust_decimal_macros::dec;
@@ -722,5 +798,52 @@ mod tests {
                 "2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a",
             )],
         );
+    }
+
+    #[test]
+    fn no_funding_fees() {
+        let funding_fee = FundingFee::new(
+            Amount::from_sat(100),
+            FundingRate::new(Decimal::ZERO).unwrap(),
+            SETTLEMENT_INTERVAL.whole_hours(),
+        )
+        .unwrap();
+
+        assert_eq!(funding_fee.as_u64(), 0);
+    }
+
+    #[test]
+    fn initial_funding_fee() {
+        // Initial funding fee is taken for the whole settlement period, in this
+        // case 1%
+
+        let funding_fee = FundingFee::new(
+            Amount::from_sat(100),
+            FundingRate::new(dec!(0.01)).unwrap(),
+            SETTLEMENT_INTERVAL.whole_hours(),
+        )
+        .unwrap();
+
+        assert_eq!(funding_fee.as_u64(), 1);
+    }
+
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn rollover_funding_fee_collected_incrementally_should_not_be_smaller_than_collected_once_per_settlement_interval(amount_sat in 1_000u64..1_000_000u64) {
+            let funding_fee_for_whole_interval =
+                FundingFee::new(Amount::from_sat(amount_sat), FundingRate::new(dec!(0.01)).unwrap(), SETTLEMENT_INTERVAL.whole_hours()).unwrap();
+            let funding_fee_for_one_hour =
+                FundingFee::new(Amount::from_sat(amount_sat), FundingRate::new(dec!(0.01)).unwrap(), 1).unwrap();
+
+            let total_when_collected_hourly = SETTLEMENT_INTERVAL.whole_hours() as u64 * funding_fee_for_one_hour.as_u64();
+            let total_when_collected_for_whole_interval = funding_fee_for_whole_interval.as_u64();
+
+            prop_assert!(
+                total_when_collected_hourly >= total_when_collected_for_whole_interval,
+                "when charged per hour we should not be at loss as compared to charging once per settlement interval"
+            );
+        }
     }
 }
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1,5 +1,7 @@
 use crate::maker_inc_connections::NoConnection;
 use crate::model::BitMexPriceEventId;
+use crate::model::FundingFee;
+use crate::model::FundingRate;
 use crate::model::Identity;
 use crate::model::InversePrice;
 use crate::model::Leverage;
@@ -424,6 +426,7 @@ pub struct Cfd {
     id: OrderId,
     position: Position,
     initial_price: Price,
+    funding_rate: FundingRate,
     leverage: Leverage,
     settlement_interval: Duration,
     quantity: Usd,
@@ -502,6 +505,7 @@ impl Cfd {
             during_contract_setup: false,
             during_rollover: false,
             settlement_proposal: None,
+            funding_rate: FundingRate::new(Decimal::ZERO).expect("be valid"),
         }
     }
 
@@ -640,7 +644,8 @@ impl Cfd {
                 self.leverage,
                 self.refund_timelock_in_blocks(),
                 1, // TODO: Where should I get the fee rate from?
-            ),
+                self.funding_rate,
+            )?,
         ))
     }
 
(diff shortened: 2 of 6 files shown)


daemon/src/model/cfd.rs
```diff
@@ -557,6 +557,15 @@ impl Cfd {
             .map(|dlc| dlc.settlement_event_id.timestamp)
     }
 
+    fn margin(&self, position: Position) -> Amount {
+        match self.position {
```
thomaseizinger, line 565: Bug!

You need to use `position` here, not `self.position`.

daemon/src/model/cfd.rs
```diff
@@ -608,19 +617,9 @@ impl Cfd {
...
         }
 
-        let margin = match self.position {
-            Position::Long => {
-                calculate_long_margin(self.initial_price, self.quantity, self.leverage)
-            }
-            Position::Short => calculate_short_margin(self.initial_price, self.quantity),
-        };
+        let margin = self.margin(self.position);
+        let counterparty_margin = self.margin(self.position.counter_position());
```
thomaseizinger: I think it would be cleaner to create two functions `margin` and `counterparty_margin`.

daemon/src/model/cfd.rs
```diff
@@ -50,6 +51,8 @@ use time::OffsetDateTime;
 use uuid::adapter::Hyphenated;
 use uuid::Uuid;
 
+use super::FundingRate;
```
thomaseizinger: Use absolute imports please.

daemon/src/model.rs
```diff
@@ -600,6 +602,73 @@ impl Timestamp {
...
+    }
+}
+
+impl FundingParams {
+    pub fn new(amount: Amount, funding_rate: FundingRate, hours_to_charge: i64) -> Result<Self> {
+        anyhow::ensure!(
+            funding_rate.0.is_sign_positive(),
+            "Negative funding rates (ie. paid by the maker) are not supported yet"
+        );
+        anyhow::ensure!(hours_to_charge >= 1, "Can't charge for less than one hour");
```
thomaseizinger, line 647: Use a different data type then? :)

daemon/src/model.rs
```diff
@@ -600,6 +602,73 @@ impl Timestamp {
...
+    }
+}
+
+/// Funding parameters used for calculating the fees
+#[derive(Debug, Clone)]
+pub struct FundingParams {
+    funding_rate: FundingRate,
+    /// How much of the funding round are we aiming to pay for now
+    fraction: Decimal,
+    amount_sat: Decimal,
```
thomaseizinger: Can we rename this or at least give it a doc comment. It is not clear from the implementation, what is stored in this value.

daemon/src/payout_curve.rs
```diff
@@ -92,6 +95,8 @@ fn calculate_payout_parameters(
         None,
     )?;
 
+    let funding_fee = funding_params.calculate_fee()?;
```
thomaseizinger: Why do we need to pass `FundingParams` all the way through if we end up only calling a single function here? Seems to be cleaner to do this outside and just pass the result in?

daemon/src/model/cfd.rs
```diff
@@ -657,6 +662,16 @@ impl Cfd {
...
 
+        // TODO: Adjust for total paid fees
+        let margin_minus_total_fees = self.margin(self.position);
+
+        // TODO: Calculate the time from the last rollover, don't just assume
+        // it's up-to-date
+        let hours_to_charge = 1;
+
+        let funding_params =
+            FundingParams::new(margin_minus_total_fees, self.funding_rate, hours_to_charge)?;
```
thomaseizinger: I don't really understand the need for the `FundingParams` struct.

It should be a stateless calculation to figure out the funding fee upon rollover. Why do we need to store stuff in a struct for that?

daemon/src/model.rs
```diff
@@ -600,6 +602,73 @@ impl Timestamp {
...
+        }
+    }
+}
+
+impl FundingParams {
+    pub fn new(amount: Amount, funding_rate: FundingRate, hours_to_charge: i64) -> Result<Self> {
+        anyhow::ensure!(
+            funding_rate.0.is_sign_positive(),
+            "Negative funding rates (ie. paid by the maker) are not supported yet"
+        );
```
thomaseizinger: If `FundingRate` were to make its field private, you could enforce this in the constructor and rely on it here!

daemon/src/model.rs
```diff
@@ -40,6 +41,7 @@ pub enum Error {
 }
 
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
+/// Represents "quantity" or "contract size" in Cfd terms
```
thomaseizinger, line 43: Comments above attributes please :)

thomaseizinger (commented):
Thanks!

A few ideas!

daemon/src/model/cfd.rs
klochowicz: hah, yes! fixed locally when adding test :)

daemon/src/model.rs
klochowicz: yes, I'm working on addressing the TODO for a newtype for hours, because this is really annoying. I'll use it here too.

daemon/src/payout_curve.rs
klochowicz: interesting. I wanted strong typing here, but indeed another newtype that's the result of this method would probably be better to be passed in.

daemon/src/model/cfd.rs
klochowicz: hm, perhaps it could be a free function after all! will give it a shot :)

daemon/src/model/cfd.rs
klochowicz: ah, it was already fixed in the PR, just accidentally squashed into the other commit! should have used `git absorb` :)

daemon/src/model/cfd.rs
klochowicz: good idea, it certainly improves the ux of this API!

daemon/src/model/cfd.rs
klochowicz: (it went away anyway with the other suggestion)

daemon/src/model.rs
```diff
@@ -792,4 +793,42 @@ mod tests {
...
+        assert_eq!(funding_params.calculate_fee().unwrap(), 1);
+    }
+
+    use proptest::prelude::*;
+
+    proptest! {
+        #[test]
+        fn rollover_funding_fee_throughout_a_day_should_amount_at_least_to_daily_founding_fee(amount_sat in 1_000u64..1_000_000u64) {
+            let funding_params_daily =
+                FundingParams::new(Amount::from_sat(amount_sat), FundingRate::from(dec!(0.01)), SETTLEMENT_INTERVAL.whole_hours()).unwrap();
```
thomaseizinger: Nit: The test assumes that `SETTLEMENT_INTERVAL == 24`. That might not always be true and I can guarantee you we will forget updating the name of this test once we change that constant. I'd inline that value here because the test is actually unrelated to the constant.

daemon/src/model.rs
```diff
@@ -792,4 +793,42 @@ mod tests {
...
+            let funding_params_hourly =
+                FundingParams::new(Amount::from_sat(amount_sat), FundingRate::from(dec!(0.01)), 1).unwrap();
+
+            let total_when_collected_hourly = SETTLEMENT_INTERVAL.whole_hours() as u64 * funding_params_hourly.calculate_fee().unwrap();
+            let total_when_collected_daily = funding_params_daily.calculate_fee().unwrap();
+
+            dbg!(&total_when_collected_hourly);
+            dbg!(&total_when_collected_daily);
```
thomaseizinger: Those should go away before we merge.

thomaseizinger (commented):
Noticed two things, not sure if ready for re-review yet :D

daemon/src/model.rs
klochowicz: fair enough! I certainly can see how we'd forget to change it :) renamed the test name, variables and assert message.

klochowicz:
> Noticed two things, not sure if ready for re-review yet :D

thanks, addressed them as well! I squashed the fixups, as there were just too many of them - looks quite different now. it should be ready for re-review!

daemon/src/model.rs
```diff
@@ -600,6 +602,70 @@ impl Timestamp {
...
+        );
+        anyhow::ensure!(
+            rate <= Decimal::ONE,
+            "Funding rate can't be higher than 100%"
+        );
+
+        Ok(Self(rate))
+    }
+
+    pub fn as_dec(&self) -> Decimal {
```
thomaseizinger: Nit: Rust guidelines recommend `as_` only for `&` -> `&` conversions. We are decreasing abstraction here I think the correct API as per https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv would be `pub fn to_decimal(self) -> Decimal` given that this type is `Copy` :)

daemon/src/model.rs
```diff
@@ -600,6 +602,70 @@ impl Timestamp {
...
+                .checked_div(Decimal::from(SETTLEMENT_INTERVAL.whole_hours()))
+                .context("can't establish a fraction")?
+        };
+
+        Ok(Self(
+            (Decimal::from(margin.as_sat()) * funding_rate.as_dec() * fraction)
+                .round_dp_with_strategy(0, rust_decimal::RoundingStrategy::AwayFromZero)
+                .to_u64()
+                .context("Failed to represent as u64")?,
+        ))
```
thomaseizinger: Nit: I find it a bit weird that the math of the funding fee is in here. I think I would have expected the ctor to only check invariants and have a separate free function that does the math and constructs `FundingFee` as the last step.

daemon/src/setup_contract.rs
```diff
@@ -383,13 +395,15 @@ impl RolloverParams {
         leverage: Leverage,
         refund_timelock: u32,
         fee_rate: u32,
```
thomaseizinger, line 393: Rename this to `miner_fee_rate` to distinguish it better?

thomaseizinger: Or `blockchain_fee_rate`?

daemon/src/payout_curve.rs
```diff
@@ -101,8 +103,10 @@ fn calculate_payout_parameters(
             let right_bound = row[1] as u64;
             let long_amount_btc = row[2];
 
+            // TODO: cover the case where the maker pays funding fees
+
             let long_amount = to_sats(long_amount_btc)?;
-            let long_amount_adjusted = long_amount.saturating_sub(funding_fee);
+            let long_amount_adjusted = long_amount.saturating_sub(funding_fee.as_u64());
```
thomaseizinger, line 109: How about we convert `FundingFee` into `bitcoin::Amount` instead and do the math before we call `to_sats`?

daemon/src/model/cfd.rs
thomaseizinger: Still unaddressed :)

thomaseizinger (approved):
concept ACK

We can iterate on the details in later PRs!

daemon/src/setup_contract.rs
luckysori: `tx_fee_rate`?

daemon/src/model.rs
klochowicz: renamed to 'margin' in the constructor, hopefully it's clear now.

daemon/src/model/cfd.rs
klochowicz: sorry, rust-analyzer did this :) changed manually.

daemon/src/model.rs
klochowicz: thanks, bookmarked! spent a few minutes trying to get clippy lint working for it, but I think it's broken :/

daemon/src/payout_curve.rs
klochowicz: Perhaps this could be done later if needed? this is an improvement that can be done at any stage - it was `u64` already. (note: I will need to revisit this anyway when allowing charging the maker)

daemon/src/setup_contract.rs
klochowicz: renamed to `tx_fee_rate`

daemon/src/model/cfd.rs
```diff
@@ -502,6 +506,7 @@ impl Cfd {
             during_contract_setup: false,
             during_rollover: false,
             settlement_proposal: None,
+            funding_rate: FundingRate::new(Decimal::ZERO).expect("be valid"),
```
luckysori, line 508: ❓ What happens when we calculate the funding fee with a fund rate of 0.0? Should we add a test?

daemon/src/payout_curve.rs
```diff
@@ -101,8 +103,10 @@ fn calculate_payout_parameters(
             let right_bound = row[1] as u64;
             let long_amount_btc = row[2];
 
+            // TODO: cover the case where the maker pays funding fees
```
luckysori, line 106: ❓ Are you going to do this as part of this PR? If not, I think we can safely skip this TODO.

daemon/src/payout_curve.rs
```diff
@@ -458,6 +462,9 @@ fn create_long_payout_function(
 
 #[cfg(test)]
 mod tests {
+    use crate::model::FundingRate;
+    use crate::SETTLEMENT_INTERVAL;
+
```
luckysori: 🔧 Can remove this newline.

luckysori (approved):

daemon/src/model/cfd.rs
klochowicz: it just doesn't collect fees - I will add a test to make that obvious if it helps! :)

daemon/src/payout_curve.rs
klochowicz: I'd prefer not to, as I will be working on this in the near future (like tomorrow). I will ensure that it's tracked in a separate issue.

thomaseizinger (approved):

daemon/src/model/cfd.rs
```diff
@@ -658,6 +670,16 @@ impl Cfd {
...
 
+        // TODO: Adjust for total paid fees
+        let margin_minus_total_fees = self.margin();
+
+        // TODO: Calculate the time from the last rollover, don't just assume
+        // it's up-to-date
+        let hours_to_charge = 1;
+
+        let funding_fee =
+            FundingFee::new(margin_minus_total_fees, self.funding_rate, hours_to_charge)?;
```
da-kami, line 681: Just pointing this out, we don't have to action it now, but we should keep in mind that this might not be what we want:

I am not sure collecting the fee based on our margin is correct. When discussing this with @bonomat we were talking about charging the fee on the `counterparty` margin, because that is what you borrow. 
That has the effect, that higher leverage will result in higher fees. 

Note that BitMEX is charging based on the amount of contracts (an abstraction that is still missing in our model):

> `Funding = Mark Value * Funding Rate`
> Your Mark Value is irrespective of leverage. For example, if you hold 100 XBTUSD contracts, funding is charged/received on the notional value of those contracts, and is not based on how much margin you have assigned to the position.

see: https://www.bitmex.com/app/perpetualContractsGuide#Funding

Maybe you can add a comment so we don't forget discussing this with @bonomat ?

klochowicz: My apologies, I must've missed that. This reasoning makes sense. If you don't mind, instead of a comment, I will convert it to a separate issue - this would be relatively quick and independent fix if we decide to action it before tomorrow.

bonomat: sorry for slow reply. I've created a discussion for this topic: #1167

PR #1103, opened 2022-01-14
Fix rebroadcast failure

### LINKED ISSUE #1101: Re-broadcasting transactions sometimes fails by panicking the application upon startup
That's the only way I could explain this behaviour:

```
    Finished dev [unoptimized + debuginfo] target(s) in 1.33s
     Running `/Users/dakami/CoBloX/repos/hermes/target/debug/taker --data-dir /Users/dakami/CoBloX/run/hermes/taker --maker '127.0.0.1:9999' --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e --log-level=trace testnet`
2022-01-14 16:01:16  INFO daemon::logger: Initialized logger
2022-01-14 16:01:16  INFO taker: Running version: 0.3.1-370-gdaab1cb
2022-01-14 16:01:16  INFO taker: CFDs created with this release will settle after 24 hours
2022-01-14 16:01:16  INFO taker: Authentication details: username='itchysats' password='7975e3878cab0fefc04f155563e75124ee43421ca5ec20b1d749484b0dbd0495'
2022-01-14 16:01:16  INFO daemon::db: Opened database at /Users/dakami/CoBloX/run/hermes/taker/testnet/taker.sqlite
2022-01-14 16:01:16 TRACE daemon::auto_rollover: Checking all CFDs for rollover eligibility
2022-01-14 16:01:16 ERROR daemon::auto_rollover: Cannot roll over: CFD does not have a DLC
Error: Broadcasting transaction failed. Txid: 61d31e9d416e4e0a690af131d0de1dbdd794011c80295df0bed71d0e4a648f98. Raw transaction: 02000000000101316fc32b4e0d77631888a0bcd701a671671f15d0ed23f2ef20b69e1b7068cdf000000000000c0000000268900300000000001600146788fe7e86ca60c4544e4f626928e8986411f115bd9e0100000000001600144cbcfac5a9f4d3cdded755f10ebbc660e98376ca0347304402201af578a29650f157fd3c53fe6192d7b77243abb8477643ea711823c474ed517102200d8774adc42af462e7686281db43604732a60b4d1994d392407b145c379a82fa01483045022100dcb181fb840e4b77b486d27485844731bfba38f2d5a8a8bf818f52d88efbdc3102203e9849a7db5f1470534484b5e4a8c6342bc51b779d9fa2d523dbbaf7cdd8597101e021033e98a47e239f5369973db0c83e1cf89adf18dadc63ff0921b94b7c7c3dc60e66ac646376a914e79c7cc8f3e660c6783224927772e60c0ba508d688ad76a9146b14c72b3ae1072e8ade9be24b63bceaf999d46c88ad76a914290c841a5b98d9053e824726a02309a692b87d09886776a914dad6e308dde828eb9b32a175bb2bfdbd08abc73f88ad76a9144baa34d49430ec9b5f2592cc2f04f51e517bb00f88ad76a914f5bf48257419f85581d72818c5f79d6b12ba050d886867210367bcef6a6cc421660176d1105648f91b1b77482dfa86ff75f6e4d29b34521eae68ac00000000

Caused by:
    Electrum server error: "sendrawtransaction RPC error: {\"code\":-26,\"message\":\"non-BIP68-final\"}"

Process finished with exit code 1
```

After I saw this it is always happening now. The daemon never starts up after we run into this problem.

rishflab:
closes #1101

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -352,8 +352,8 @@ pub enum CfdEvent {
     RevokeConfirmed,
     CollaborativeSettlementConfirmed,
 
-    CetTimelockConfirmedPriorOracleAttestation,
-    CetTimelockConfirmedPostOracleAttestation {
+    CetTimelockExpiredPriorOracleAttestation,
+    CetTimelockExpiredPostOracleAttestation {
         #[serde(with = "hex_transaction")]
         cet: Transaction,
     },
@@ -996,8 +996,8 @@ impl Cfd {
             .cet
             .take()
             // If we have cet, that means it has been attested
-            .map(|cet| CfdEvent::CetTimelockConfirmedPostOracleAttestation { cet })
-            .unwrap_or_else(|| CfdEvent::CetTimelockConfirmedPriorOracleAttestation);
+            .map(|cet| CfdEvent::CetTimelockExpiredPostOracleAttestation { cet })
+            .unwrap_or_else(|| CfdEvent::CetTimelockExpiredPriorOracleAttestation);
 
         Ok(self.event(cfd_event))
     }
@@ -1181,8 +1181,8 @@ impl Cfd {
             RefundTimelockExpired { .. } => self.refund_timelock_expired = true,
             LockConfirmed => self.lock_finality = true,
             CommitConfirmed => self.commit_finality = true,
-            CetTimelockConfirmedPriorOracleAttestation
-            | CetTimelockConfirmedPostOracleAttestation { .. } => {
+            CetTimelockExpiredPriorOracleAttestation
+            | CetTimelockExpiredPostOracleAttestation { .. } => {
                 self.cet_timelock_expired = true;
             }
             OfferRejected => {
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -207,11 +207,11 @@ impl Cfd {
             },
             // final states, don't monitor anything
             CetConfirmed | RefundConfirmed | CollaborativeSettlementConfirmed => Self::default(),
-            CetTimelockConfirmedPriorOracleAttestation => Self {
+            CetTimelockExpiredPriorOracleAttestation => Self {
                 monitor_cet_timelock: false,
                 ..self
             },
-            CetTimelockConfirmedPostOracleAttestation { cet, .. } => Self {
+            CetTimelockExpiredPostOracleAttestation { cet, .. } => Self {
                 cet: Some(cet),
                 monitor_cet_timelock: false,
                 ..self
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -339,8 +339,6 @@ impl Actor<bdk::electrum_client::Client> {
             }
 
             if let Some(cet) = cet {
-                // Double question mark OK because if we are in PendingCet we must have been
-                // Ready before
                 let txid = broadcast_transaction(&self.client, cet)?;
                 tracing::info!("CET published on chain: {}", txid);
             }
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -207,8 +207,12 @@ impl Cfd {
             },
             // final states, don't monitor anything
             CetConfirmed | RefundConfirmed | CollaborativeSettlementConfirmed => Self::default(),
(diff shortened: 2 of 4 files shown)


daemon/src/monitor.rs
```diff
@@ -337,11 +344,13 @@ impl Actor<bdk::electrum_client::Client> {
                 tracing::info!("Commit transaction published on chain: {}", txid);
             }
 
-            if let Some(cet) = cet {
-                // Double question mark OK because if we are in PendingCet we must have been
-                // Ready before
-                let txid = broadcast_transaction(&self.client, cet)?;
-                tracing::info!("CET published on chain: {}", txid);
+            if !monitor_cet_timelock {
```
thomaseizinger: Your bug is on line 224. The variable is even called `timelocked_cet` :upside_down_face: 

Just don't set `self.cet` there and you should be fine!

thomaseizinger (changes requested):

daemon/src/model/cfd.rs
```diff
@@ -346,8 +346,8 @@ pub enum CfdEvent {
     RevokeConfirmed,
     CollaborativeSettlementConfirmed,
 
-    CetTimelockConfirmedPriorOracleAttestation,
-    CetTimelockConfirmedPostOracleAttestation {
+    CetTimelockExpiredPriorOracleAttestation,
```
thomaseizinger, line 355: Nice!

daemon/src/monitor.rs
rishflab: why do we set it in `src/model/cfd.rs/ and not here. 

`timelocked_cet` is set below. I didnt change that.

thomaseizinger: > why do we set it in `src/model/cfd.rs/ and not here. 

It is unfortunately a bit tricky but the gist is:

- When we get the attestation, we need to decrypt the CET in any case, even if the timelock is still active. We can't wait.
- The only way to store data in event sourcing is to emit an event, hence we include it in the event.
- Re-hydration of the aggregate will pick up that event and can pull out the data.
- When we later get told that the timelock expired, we can check whether we still have a timelocked CET pending and emit it then. Otherwise we don't have to do anything.

That is why we need to record it in the model CFD's apply method.

> `timelocked_cet` is set below. I didnt change that.

 But that was introduced by your previous PR right? Prior to rebroascasting transactions, we didn't store any of these transactions in this aggregate.

daemon/src/monitor.rs
```diff
@@ -207,17 +208,17 @@ impl Cfd {
...
+                cet: Some(cet),
+                monitor_cet_timelock: false,
                 ..self
             },
-            OracleAttestedPostCetTimelock { cet, .. } => Self {
-                cet: Some(cet),
+            RefundTimelockExpired { .. } => Self {
+                monitor_refund_timelock: false,
                 ..self
             },
```
thomaseizinger: Can we put these back into the original order? Thanks :)

daemon/src/monitor.rs
thomaseizinger: You can remove this check now. `self.cet` should only be set to CETs that can be broadcasted!

daemon/src/monitor.rs
```diff
@@ -233,6 +234,7 @@ impl Cfd {
             | RolloverAccepted
             | RolloverFailed
             | ManualCommit { .. }
+            | OracleAttestedPostCetTimelock { .. }
```
thomaseizinger: This one contains a `cet` that we should re-broadcast.

daemon/src/monitor.rs
```diff
@@ -173,8 +173,9 @@ impl Cfd {
                     monitor_cet_timelock: true,
                     monitor_refund_timelock: true,
                     monitor_refund_finality: true,
-                    monitor_revoked_commit_transactions: true, /* After rollover, the other party
-                                                                * might publish old states. */
+                    monitor_revoked_commit_transactions: true,
+                    /* After rollover, the other party
+                     * might publish old states. */
```
thomaseizinger: Nit: Unnecessary diff.

thomaseizinger (changes requested):

thomaseizinger (approved):

thomaseizinger (approved):
Very clean diff, thanks for that!

PR #1104, opened 2022-01-14
Simplify CFD read model of projection actor

thomaseizinger:
Almost a pure refactor. One tiny change in when we emit commit tx.

I started out with using `HashSet`s instead of `Vec` which prevents duplicates by default and ended up refactoring the entire function to be composed of multiple pure functions. Strictly speaking, we no longer need the `HashSet`s now because we are not modifying the set but always replacing the entire collection. We could thus revert the change to a `HashSet` but I think it is overall the more suited data structure. Open to discuss though.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -364,8 +364,12 @@ pub enum CfdEvent {
     OracleAttestedPriorCetTimelock {
         #[serde(with = "hex_transaction")]
         timelocked_cet: Transaction,
-        #[serde(with = "hex_transaction")]
-        commit_tx: Transaction,
+        /// The commit transaction for the DLC of this CFD.
+        ///
+        /// If this is set to `Some`, we haven't previously attempted broadcast `commit_tx` and
+        /// need to broadcast it as a result of this event.
+        #[serde(with = "hex_transaction::opt")]
+        commit_tx: Option<Transaction>,
         price: Price,
     },
     OracleAttestedPostCetTimelock {
@@ -439,7 +443,8 @@ pub struct Cfd {
     /// However, if this is set to `Some`, there is no need to re-emit it as part of another event.
     cet: Option<Transaction>,
 
-    /// Holds the decrypted commit transaction if we have previously emitted it as part of an event.
+    /// Holds the decrypted commit transaction if we have previously emitted it as part of an
+    /// event.
     ///
     /// There is not guarantee that the transaction is confirmed if this is set to `Some`.
     /// However, if this is set to `Some`, there is no need to re-emit it as part of another event.
@@ -932,9 +937,15 @@ impl Cfd {
             ));
         }
 
+        // If we haven't yet emitted the commit tx, we need to emit it now.
+        let commit_tx_to_emit = match self.commit_tx {
+            Some(_) => None,
+            None => Some(dlc.signed_commit_tx()?),
+        };
+
         Ok(Some(self.event(CfdEvent::OracleAttestedPriorCetTimelock {
             timelocked_cet: cet,
-            commit_tx: dlc.signed_commit_tx()?,
+            commit_tx: commit_tx_to_emit,
             price,
         })))
     }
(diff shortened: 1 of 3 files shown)


daemon/src/model/cfd.rs
```diff
@@ -364,8 +364,12 @@ pub enum CfdEvent {
     OracleAttestedPriorCetTimelock {
         #[serde(with = "hex_transaction")]
         timelocked_cet: Transaction,
-        #[serde(with = "hex_transaction")]
-        commit_tx: Transaction,
+        /// The commit transaction for the DLC of this CFD.
+        ///
+        /// If this is set to `Some`, we haven't previously attempted broadcast `commit_tx` and
+        /// need to broadcast it as a result of this event.
+        #[serde(with = "hex_transaction::opt")]
```
luckysori, line 371: 🤔 I'm surprised that we need a custom implementation here, but you know all the `serde` tricks.

da-kami (approved):
let's get this in :)

daemon/src/projection.rs
```diff
@@ -481,6 +500,37 @@ impl Cfd {
...
+    fn derive_actions(&self, role: Role) -> HashSet<CfdAction> {
+        match (self.state, role) {
+            (CfdState::PendingSetup, Role::Maker) => {
+                HashSet::from([CfdAction::AcceptOrder, CfdAction::RejectOrder])
+            }
+            (CfdState::PendingSetup, Role::Taker) => HashSet::new(),
+            (CfdState::ContractSetup, _) => HashSet::new(),
+            (CfdState::Rejected, _) => HashSet::new(),
+            (CfdState::PendingOpen, _) => HashSet::new(),
+            (CfdState::Open, _) => HashSet::from([CfdAction::Commit, CfdAction::Settle]),
```
luckysori, line 452: 👍 It looks like you fixed a bug here: we had no actions after rolling over (or failing to roll over) and returning to `CfdState::Open`.

daemon/src/projection.rs
```diff
@@ -481,6 +500,37 @@ impl Cfd {
...
+            (CfdState::PendingCommit, _) => HashSet::new(),
+            (CfdState::PendingCet, _) => HashSet::new(),
+            (CfdState::PendingClose, _) => HashSet::new(),
+            (CfdState::OpenCommitted, _) => HashSet::new(),
```
luckysori, line 456: 🤔 I think the behaviour has changed here when coming from `CetTimelockConfirmedPriorOracleAttestation`: we used to keep the available actions from applying the previous event.

But this looks correct to me since there's no action to do when `commit_tx` is on-chain.

luckysori (approved):

daemon/src/projection.rs
```diff
@@ -289,56 +289,54 @@ impl Cfd {
...
+                Role::Taker => {
+                    self.state = CfdState::OutgoingSettlementProposal;
                 }
-                Role::Taker => (CfdState::OutgoingSettlementProposal, vec![]),
             },
             CollaborativeSettlementProposalAccepted => {
                 self.pending_settlement_proposal_price = None;
-
-                (CfdState::IncomingSettlementProposal, vec![])
+                self.state = CfdState::IncomingSettlementProposal;
```
klochowicz, line 332: I suspect that this is a bug on the maker side -  won't this code will show the buttons to "Accept/Reject" on the maker after accepting a proposal? (we might need to expand the pure function to look at the state + other things, or create more UI-only states....)

daemon/src/projection.rs
```diff
@@ -254,7 +255,7 @@ impl Cfd {
             state: CfdState::PendingSetup,
             actions: initial_actions,
             details: CfdDetails {
-                tx_url_list: vec![],
+                tx_url_list: HashSet::new(),
```
klochowicz, line 280: nice! this indeed makes sense :)

klochowicz:
> bors r+

oops, got borsed before I managed to finish my review 😢

thomaseizinger:
bors retry

daemon/src/projection.rs
thomaseizinger: Uhm true! We don't have an in-between state for that I think!

PR #1105, opened 2022-01-14
Disable incremental builds on the CI

klochowicz:
We don't use incremental builds on the CI - the premise is that it should
faster without it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,9 @@ on:
       - 'staging'
       - 'master'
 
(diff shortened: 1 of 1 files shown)


klochowicz:
starting as a draft, as the hypothesis found in one article needs to be verified first :)

thomaseizinger (approved):

thomaseizinger:
I think we can also drop uploading of the artifacts, we never use that :)

klochowicz:
Bors r+

PR #1107, opened 2022-01-16
Unify clippy.toml files

klochowicz:
Somehow we ended up with 2 places where we added clippy lints.
Preserve a single one at workspace level by merging contents of daemon-specific
one into it.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -2,5 +2,6 @@ disallowed-methods = [
   { path = "xtra::Address::do_send_async", reason = "discards the return value, possibly swallowing an error" },
   { path = "xtra::Address::do_send", reason = "discards the return value, possibly swallowing an error" },
   { path = "xtra::message_channel::MessageChannel::do_send", reason = "discards the return value, possibly swallowing an error" },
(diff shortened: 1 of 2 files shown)


clippy.toml
```diff
@@ -3,4 +3,5 @@ disallowed-methods = [
   { path = "xtra::Address::do_send", reason = "discards the return value, possibly swallowing an error" },
   { path = "xtra::message_channel::MessageChannel::do_send", reason = "discards the return value, possibly swallowing an error" },
   { pant = "xtra::Context::notify_interval", reason = "does not wait for response from the previous handler, prefer `xtra_ext::Address::send_interval`" },
```
thomaseizinger: There is a typo here which tells me that clippy did actually not read both of these files before!

thomaseizinger (changes requested):

clippy.toml
klochowicz: lol, indeed! and because it was a valid word my spellchecker didn't complain 😆

thomaseizinger (approved):

PR #1115, opened 2022-01-17
Set `PendingClose` after proposal is accepted

thomaseizinger:
Otherwise we will expose accept/reject buttons in the UI although
it has already been accepted.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -329,7 +329,7 @@ impl Cfd {
             CollaborativeSettlementProposalAccepted => {
                 self.pending_settlement_proposal_price = None;
 
-                self.state = CfdState::IncomingSettlementProposal;
+                self.state = CfdState::PendingClose;
             }
             CollaborativeSettlementCompleted {
                 spend_tx,


daemon/src/projection.rs
```diff
@@ -329,7 +329,7 @@ impl Cfd {
             CollaborativeSettlementProposalAccepted => {
                 self.pending_settlement_proposal_price = None;
 
-                self.state = CfdState::IncomingSettlementProposal;
+                self.state = CfdState::PendingClose;
```
klochowicz, line 332: I'm not sure whether this is strictly correct - it would be better to only put into `PendingClose` state after the protocol has run. Otherwise, if we try to accept an "old" proposal, the taker might have moved on with their cfd and e.g. force-closed it.

thomaseizinger: This is a pretty "short" state anyway because we will either get "Completed" or "Failed" afterwards. Not sure if it is worth to introduce a completely new state.

klochowicz: fair enough. It certainly is less of a concern than still displaying the action....

klochowicz (approved):
as discussed, this is a "small" shortcut for a situation where we might not care about a temporary advancement to the next state in the UI, as it will get settled to a proper one shortly afterwards.

PR #1117, opened 2022-01-17
Make heartbeat message self-descriptive

### LINKED ISSUE #1116: Report missing local taker daemon after losing connection to maker
We are briefly displaying the `Please ensure taker daemon is running` alert when we lose connection to the _maker_.

luckysori:
Closes #1116.

Actually has nothing to do with losing connection to the maker specifically. My theory is that the change in the maker status means that delivering the next heartbeat is delayed by a little bit.

### DIFF
--- a/taker-frontend/src/useEventSource.ts
+++ b/taker-frontend/src/useEventSource.ts
@@ -34,10 +34,16 @@ export function useEventSource(url: string): [EventSource | null, boolean] {
 
         es.addEventListener("heartbeat", heartbeatHandler);
 
+        const beforeUnloadHandler = () => {
+            es.close();
+        };
+        window.addEventListener("beforeunload", beforeUnloadHandler);
+
         return () => {
             setSource(null);
             es.removeEventListener("error", errorHandler);
             es.removeEventListener("heartbeat", heartbeatHandler);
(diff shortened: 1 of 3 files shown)


taker-frontend/src/useEventSource.ts
```diff
@@ -18,14 +18,14 @@ export function useEventSource(url: string): [EventSource | null, boolean] {
...
-        }, 5000);
+        }, 10000);
 
         const heartbeatHandler = () => {
             clearTimeout(timer);
             setIsConnected(true);
             timer = setTimeout(() => {
                 setIsConnected(false);
-            }, 5000);
+            }, 10000);
```
luckysori: 🤔 We could add a `next_heartbeat` field to the heartbeat message and use that value + a buffer here. That way this value can never go out of sync with the backend.

thomaseizinger: Nit: `next` would imply an absolute value? Just `interval` (or `pulse`? 😁) perhaps, otherwise concept ACK.

thomaseizinger (approved):

thomaseizinger (approved):

rishflab (approved):

luckysori:
bors retry

PR #1122, opened 2022-01-17
Populate funding rate inside offer in the maker

klochowicz:
Hard-code "0.0" until UI creates means for choosing funding rate.

Partial fix for #244

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -20,6 +20,7 @@ use bdk::FeeRate;
 use connection::ConnectionStatus;
 use futures::future::RemoteHandle;
 use maia::secp256k1_zkp::schnorrsig;
+use model::FundingRate;
 use sqlx::SqlitePool;
 use std::future::Future;
 use std::net::SocketAddr;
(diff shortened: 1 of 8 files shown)


daemon/src/model.rs
```diff
@@ -603,7 +603,7 @@ impl Timestamp {
 }
 
 /// Funding rate per SETTLEMENT_INTERVAL
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
```
thomaseizinger, line 606: Simply deriving these will allow you to side-step the invariants you are checking in the constructor. Deriving `Serialize` / `Deserialize` for a new-type is almost always incorrect :)

daemon/src/lib.rs
```diff
@@ -238,13 +239,15 @@ where
         min_quantity: Usd,
         max_quantity: Usd,
         fee_rate: Option<u32>,
+        funding_rate: FundingRate,
```
thomaseizinger: Can we make this optional here? That would make this a backwards-compatible API change.

thomaseizinger (changes requested):
Some comments!

daemon/src/lib.rs
klochowicz: sure!

daemon/src/model.rs
klochowicz: please bear in mind that the invariants are temporary, until we support charging the maker. Happy to hear how would I go about it if I was planning to have long-term invariants in a type! Custom serialise/deserialise with fallbacking to an option if the invariants are not upheld?

thomaseizinger: > fallbacking to an option if the invariants are not upheld?

No, you can just fail the deserialization with an arbitrary message! Usually, you'd just call the constructor to make sure you don't duplicate them and then just do `map_err(D::Error::custom)?;` to forward the error.

thomaseizinger (approved):

PR #1123, opened 2022-01-17
Dont log unnecasary errors

### LINKED ISSUE #1083: When handling commands we should be mindful to log errors/warnings
At the moment we log multiple errors, because we may load *all* `Cfd`s when processing something happened in the system. If e.g. we are unable to handle an attestation because a `Cfd` is already final, this is *not* an error (most likely there should not be a log at all).

Note: Eventually we should optimize this in the database, but regardless of that (not scope of this ticket!) the `Cfd`'s event logic should be more mindful with logging errors and warnings. 

Here are examples of logs where this happens:

```
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=03b5eb02-412d-4fbe-96ad-0de1660cb9a9
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=46455d75-3e8c-4aab-9066-ceaf287de88c
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=4c7aeac8-da47-44e2-8d7b-4df46e3503d9
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=4d276386-df4d-4a57-a16c-196ba982f97e
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=87d06c92-7bf5-4f27-9c00-92d8b2f0a0dd
2022-01-12 16:07:14 DEBUG daemon::model::cfd: Attestation /x/BitMEX/BXBT/2022-01-10T09:00:00.price?n=20 is irrelevant for DLC 0983ed92ec68b61de1464910674a6f50591fe726ba7058afece3dd72acbb56b9
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=a622c191-c12f-44dc-8c5e-9cec3f2f02cf
2022-01-12 16:07:14  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=ab1e0e6b-14c3-4048-bd29-c53b6300eb5c
```

and:

```
2022-01-14 11:51:56  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=4c7aeac8-da47-44e2-8d7b-4df46e3503d9
2022-01-14 11:51:56  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=4d276386-df4d-4a57-a16c-196ba982f97e
2022-01-14 11:51:56  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=87d06c92-7bf5-4f27-9c00-92d8b2f0a0dd
2022-01-14 11:51:57 ERROR daemon::cfd_actors: Failed to decrypt CET using attestation: Condition failed: `!self.is_final()`
2022-01-14 11:51:57  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=a622c191-c12f-44dc-8c5e-9cec3f2f02cf
2022-01-14 11:51:57  WARN daemon::model::cfd: Handling attestation without a DLC is a no-op order_id=ab1e0e6b-14c3-4048-bd29-c53b6300eb5c
2022-01-14 11:51:57 ERROR daemon::cfd_actors: Failed to decrypt CET using attestation: Condition failed: `!self.is_final()`
2022-01-14 11:51:58 ERROR daemon::cfd_actors: Failed to decrypt CET using attestation: Condition failed: `!self.is_final()`
```

rishflab:
Closes #1083

### DIFF
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -120,10 +120,9 @@ where
         for id in cfd_ids {
             try_continue!(async {
                 let cfd = load_cfd(id, &mut conn).await?;
-                cfd.can_auto_rollover_taker(OffsetDateTime::now_utc())?;
-
-                this.send(Rollover(id)).await??;
-
+                if let Ok(()) = cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
+                    this.send(Rollover(id)).await??;
+                }
                 anyhow::Ok(())
             }
             .await
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -68,16 +68,12 @@ where
         }
     }
 
-    async fn handle(
-        &mut self,
-        Rollover(order_id): Rollover,
-        ctx: &mut xtra::Context<Self>,
-    ) -> Result<()> {
+    async fn handle(&mut self, Rollover(order_id): Rollover, ctx: &mut xtra::Context<Self>) {
         let disconnected = match self.rollover_actors.get_disconnected(order_id) {
             Ok(disconnected) => disconnected,
             Err(_) => {
                 tracing::debug!(%order_id, "Rollover already in progress");
-                return Ok(());
+                return;
             }
         };
 
@@ -99,8 +95,6 @@ where
 
         disconnected.insert(addr);
         self.tasks.add(fut);
-
-        Ok(())
     }
 }
 
@@ -121,7 +115,7 @@ where
             try_continue!(async {
                 let cfd = load_cfd(id, &mut conn).await?;
                 if let Ok(()) = cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
-                    this.send(Rollover(id)).await??;
+                    this.send(Rollover(id)).await?;
                 }
                 anyhow::Ok(())
             }
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -950,20 +950,20 @@ impl Cfd {
 
     /// Given an attestation, find and decrypt the relevant CET.
     pub fn decrypt_cet(self, attestation: &oracle::Attestation) -> Result<Option<Event>> {
-        anyhow::ensure!(!self.is_final());
+        if self.is_final() {
+            return Ok(None);
+        }
 
         let dlc = match self.dlc.as_ref() {
             Some(dlc) => dlc,
-            None => {
-                tracing::warn!(order_id = %self.id(), "Handling attestation without a DLC is a no-op");
-                return Ok(None);
-            }
+            None => return Ok(None),
         };
 
-        let cet = match dlc.signed_cet(attestation)? {
+        let cet = dlc.signed_cet(attestation)?;
+
+        let cet = match cet {
             Ok(cet) => cet,
-            Err(e @ IrrelevantAttestation { .. }) => {
-                tracing::debug!("{}", e);
+            Err(IrrelevantAttestation { .. }) => {
                 return Ok(None);
             }
         };
--- a/daemon/tests/harness/mod.rs
+++ b/daemon/tests/harness/mod.rs
@@ -316,7 +316,6 @@ impl Taker {
             .send(auto_rollover::Rollover(id))
             .await
             .unwrap()
-            .unwrap()
     }
 }
 


daemon/src/model/cfd.rs
```diff
@@ -949,22 +963,32 @@ impl Cfd {
...
             }
         };
 
-        let cet = match dlc.signed_cet(attestation)? {
+        let cet = match cet {
             Ok(cet) => cet,
             Err(e @ IrrelevantAttestation { .. }) => {
-                tracing::debug!("{}", e);
-                return Ok(None);
+                return Err(DecryptCetError::IrrelevantAttestation(e));
```
thomaseizinger: I think we are introducing a bug here!

We will almost certainly hit this because we feed every attestation to every CFD. This needs to remain an `Ok(None)`.

daemon/src/model/cfd.rs
```diff
@@ -976,10 +1000,15 @@ impl Cfd {
             ));
         }
 
+        let commit_tx = match dlc.signed_commit_tx() {
+            Ok(commit_tx) => commit_tx,
+            Err(err) => return Err(DecryptCetError::ConstructingSignedCetFailed(err)),
```
thomaseizinger: I think this is using the wrong error variant. Also, you can replace this with a `map_err` :)

daemon/src/model/cfd.rs
```diff
@@ -949,22 +963,32 @@ impl Cfd {
...
 
     /// Given an attestation, find and decrypt the relevant CET.
-    pub fn decrypt_cet(self, attestation: &oracle::Attestation) -> Result<Option<Event>> {
-        anyhow::ensure!(!self.is_final());
+    pub fn decrypt_cet(
+        self,
+        attestation: &oracle::Attestation,
+    ) -> Result<Option<Event>, DecryptCetError> {
+        if self.is_final() {
+            return Err(DecryptCetError::CfdIsFinal);
```
thomaseizinger: WDYT of just returning `Ok(None)` here and putting a `trace` statement in that says "Cannot decrypt CET because CFD is already final."?

daemon/src/model/cfd.rs
```diff
@@ -949,22 +963,32 @@ impl Cfd {
...
+        }
 
         let dlc = match self.dlc.as_ref() {
             Some(dlc) => dlc,
             None => {
-                tracing::warn!(order_id = %self.id(), "Handling attestation without a DLC is a no-op");
-                return Ok(None);
+                return Err(DecryptCetError::NoDlc(self.id()));
```
thomaseizinger: Same here, I think it would be cleaner to just return `Ok(None)` here.

daemon/src/cfd_actors.rs
```diff
@@ -116,9 +116,17 @@ pub async fn handle_oracle_attestation(
...
-            .context("Failed to decrypt CET using attestation"));
+
+        let event = match cfd.decrypt_cet(&attestation) {
+            Ok(event) => event,
+            Err(e) => match e {
+                DecryptCetError::NoDlc(_) | DecryptCetError::CfdIsFinal => {
+                    tracing::trace!("Decrypting cet is a no-op: {:#}", e);
+                    None
+                }
+                _ => bail!(e),
```
thomaseizinger: I don't think we should bail here. Just because one CFD fails to decrypt the CET doesn't mean we shouldn't handle the others.

daemon/src/auto_rollover.rs
```diff
@@ -118,16 +117,18 @@ where
             .address()
             .expect("actor to be able to give address to itself");
         for id in cfd_ids {
-            try_continue!(async {
+            async {
```
thomaseizinger: I think I'd prefer if we extract this into a function. Makes it more readable.

thomaseizinger (changes requested):
Some ideas!

daemon/src/model/cfd.rs
rishflab: what do you think of matching on this error variant at the call site? It seems like an error from the perspective of the caller of `decrypt_cfd` to me.

daemon/src/cfd_actors.rs
rishflab: I think the bails here are proper errors. Something must of gone really wrong if this is happening?

daemon/src/model/cfd.rs
rishflab: Changing this fucntion to return `Result<Event, DecryptCetError>`

I think the question here is do we want to handle tracing inside this function or outside. I am leaning to returning concrete errors as much as possible and handling errors higher up. It will make out code easier to refactor and composable i beleive.

daemon/src/cfd_actors.rs
```diff
@@ -116,9 +114,8 @@ pub async fn handle_oracle_attestation(
 
     for id in db::load_all_cfd_ids(&mut conn).await? {
         let cfd = load_cfd(id, &mut conn).await?;
-        let event = try_continue!(cfd
-            .decrypt_cet(&attestation)
-            .context("Failed to decrypt CET using attestation"));
+
+        let event = cfd.decrypt_cet(&attestation)?;
```
thomaseizinger: I think we should keep the `try_continue` here. Just because one CFD fails doesn't mean we should abort the loop.

daemon/src/model/cfd.rs
```diff
@@ -950,22 +950,20 @@ impl Cfd {
 
     /// Given an attestation, find and decrypt the relevant CET.
     pub fn decrypt_cet(self, attestation: &oracle::Attestation) -> Result<Option<Event>> {
```
thomaseizinger, line 952: The commit message still talks about a custom error type :)

daemon/src/model/cfd.rs
```diff
@@ -950,22 +950,20 @@ impl Cfd {
...
-        let cet = match dlc.signed_cet(attestation)? {
+        let cet = dlc.signed_cet(attestation)?;
+
+        let cet = match cet {
             Ok(cet) => cet,
-            Err(e @ IrrelevantAttestation { .. }) => {
-                tracing::debug!("{}", e);
-                return Ok(None);
-            }
+            Err(e @ IrrelevantAttestation { .. }) => bail!(e),
```
thomaseizinger: I think this should remain an `Ok(None)`. This will trigger VERY often because we feed every attestation to every CFD.

daemon/src/model/cfd.rs
```diff
@@ -1537,7 +1537,7 @@ impl Dlc {
     }
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug, thiserror::Error, PartialEq)]
```
thomaseizinger: Nit: I think this diff is no longer necessary.

daemon/src/model/cfd.rs
```diff
@@ -976,10 +974,12 @@ impl Cfd {
             ));
         }
 
+        let commit_tx = dlc.signed_commit_tx()?;
+
         // If we haven't yet emitted the commit tx, we need to emit it now.
         let commit_tx_to_emit = match self.commit_tx {
             Some(_) => None,
-            None => Some(dlc.signed_commit_tx()?),
+            None => Some(commit_tx),
```
thomaseizinger: Nit: Unnecessary diff.

daemon/src/auto_rollover.rs
```diff
@@ -108,31 +108,34 @@ impl<O> Actor<O>
...
+) -> Result<()> {
+    let cfd = load_cfd(id, conn).await?;
+    match cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
+        Ok(()) => {
+            addr.send(Rollover(id)).await??;
+        }
+        Err(e) => tracing::debug!("Could not auto-rollover: {:#}", e),
+    }
+    Ok(())
+}
```
thomaseizinger: What is the purpose of extracting this function?

Matching on the result of `cfd.can_auto_rollover_taker` would also be possible within the loop above.

Also note that now, we are aborting this loop if anything fails so this is actually _less_ resilient.

daemon/src/auto_rollover.rs
```diff
@@ -108,31 +108,34 @@ impl<O> Actor<O>
...
 
+async fn load_and_trigger_rollover<O: xtra::Handler<oracle::GetAnnouncement>>(
+    id: OrderId,
+    conn: &mut PoolConnection<Sqlite>,
+    addr: &Address<Actor<O>>,
+) -> Result<()> {
+    let cfd = load_cfd(id, conn).await?;
+    match cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
+        Ok(()) => {
+            addr.send(Rollover(id)).await??;
```
thomaseizinger: Unreleated to this PR: Looking at the handler of `Rollover`, it seems that it is unnecessarily returning a `Result` so you could spare yourself one of these `?` here.

thomaseizinger (changes requested):

daemon/src/auto_rollover.rs
rishflab: You asked to extract this block to a function in first review as far as i remember?

thomaseizinger: I indeed did! I think that was because you just removed the `try_continue`, leaving behind an async block that was immediately awaited which looked odd.

Happy to keep the function but I think we should also keep the `try_continue`, otherwise we are not going through all CFDs if one fails.

thomaseizinger (approved):
Nice work!

- Minimal diff
- Relevant commit messages
- Clean-up legacy on the way

bors r+

PR #1127, opened 2022-01-18
Don't use tokio runtime in auto-rollover unit tests

klochowicz:
There are no async functions used within these tests


thomaseizinger (approved):
Well spotted.

PR #1130, opened 2022-01-18
Include transactions in more cases

thomaseizinger:
Previously, we only included links to transactions if they were part of
an event. This turned out to not be true. Some transactions, like commit
can be published by either party in which case we only learn about it as
part of getting the `CommitConfirmed` event.

Refactor projection to explicitly track the status of whether a
transaction has been published for all transactions that we can derive
from the DLC. In case they have been published, compute the transaction
URL from the DLC and include it in the response.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -28,6 +28,7 @@ use bdk::bitcoin::Script;
 use bdk::bitcoin::SignedAmount;
 use bdk::bitcoin::Transaction;
 use bdk::bitcoin::Txid;
+use bdk::miniscript::DescriptorTrait;
 use maia::TransactionExt;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
@@ -196,16 +197,13 @@ struct Aggregated {
     /// If this is present, we have an active DLC.
     latest_dlc: Option<Dlc>,
     /// If this is present, it should have been published.
-    lock_tx: Option<Transaction>,
-    /// If this is present, it should have been published.
-    commit_tx: Option<Transaction>,
-    /// If this is present, it should have been published.
     collab_settlement_tx: Option<(Transaction, Script)>,
     /// If this is present, it should have been published.
     cet: Option<Transaction>,
-    /// If this is present, it should have been published.
-    refund_tx: Option<Transaction>,
     closing_price: Option<Price>,
+
+    commit_published: bool,
+    refund_published: bool,
 }
 
 impl Cfd {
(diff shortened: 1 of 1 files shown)


daemon/src/projection.rs
```diff
@@ -341,21 +339,25 @@ impl Cfd {
 
                 self.state = CfdState::PendingClose;
             }
-            CollaborativeSettlementRejected { commit_tx } => {
-                self.aggregated.commit_tx = Some(commit_tx);
+            CollaborativeSettlementRejected { .. } => {
+                self.aggregated.commit_published = true;
                 self.pending_settlement_proposal_price = None;
```
klochowicz, line 344: not from this PR, but it caught my attention: why do we assume that commit got published after the collab settlement is rejected? AFAICT it's a higher-level business decision that the taker does, the projection should not be aware of it?

klochowicz (approved):

daemon/src/projection.rs
da-kami: We started doing this when we merged event sourcing. 
The idea is: If the taker wants to close the taker wants to close. If the maker rejects we force close. 
We can question this again, but I think it is fair to say that once the taker hits close the taker wants the position to close.

klochowicz: I don't mind living with that business decision, but it seems odd and fragile to encode it in this place? I would assume that we'd rather emit the force close event in the cfd actor after a failure/rejection? (or maybe first retry after failure?)

thomaseizinger: It is a fair question. The way I see it is that the modelling of the event encodes this business decision. The fact that it includes the commit transaction means we want it to get published. So it is not so much the projection doing this reasoning, it is the projection adhering to the contract that the event provides.

da-kami: It is part of the event logic to do that atm.
I created a discussion for it, because I'm actually not sure about it: https://github.com/itchysats/itchysats/discussions/1132

klochowicz: > It is a fair question. The way I see it is that the modelling of the event encodes this business decision. The fact that it includes the commit transaction means we want it to get published. So it is not so much the projection doing this reasoning, it is the projection adhering to the contract that the event provides.

This is an interesting point, but doesn't it also mean that our business decisions become *immutable* as a result? thanks @da-kami for creating a discussion, when funding rate is in, I aim to catch up on all the discussions ;D

klochowicz: @thomaseizinger please feel free to push this, these remarks are as such unrelated to the PR, and they might as well go in 👍

PR #1141, opened 2022-01-18
Fill in funding rate in projection

klochowicz:
Calculate hourly and annualised hourly rate in projection for UI needs.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -753,10 +753,25 @@ impl From<Order> for CfdOrder {
                 .whole_seconds()
                 .try_into()
                 .expect("settlement_time_interval_hours is always positive number"),
-            // FIXME: Replace dummy values for opening fee, funding rates and time
+            // FIXME: Replace dummy value for opening fee
             opening_fee: Amount::from_sat(123 * 24),
-            funding_rate_annualized_percent: "18.5".to_string(),
-            funding_rate_hourly_percent: "0.002345".to_string(),
+            funding_rate_annualized_percent: order
+                .funding_rate
+                .to_decimal()
+                .checked_mul(Decimal::from(
+                    (24 / order.settlement_interval.whole_hours()) * 365,
+                ))
+                .expect("not to overflow")
+                .round_dp(2)
+                .to_string(),
+            funding_rate_hourly_percent: order
+                .funding_rate
+                .to_decimal()
+                .checked_div(Decimal::from(order.settlement_interval.whole_hours()))
+                .expect("to be able to divide")
+                .round_dp(2)
+                .to_string(),
+            // FIXME: Replace dummy value for next funding event
             next_funding_event: datetime!(2030-09-23 10:00:00).assume_utc(),
         }
     }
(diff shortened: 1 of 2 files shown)


klochowicz:
Partial fix for #1142

daemon/src/projection.rs
```diff
@@ -753,10 +753,25 @@ impl From<Order> for CfdOrder {
...
             opening_fee: Amount::from_sat(123 * 24),
-            funding_rate_annualized_percent: "18.5".to_string(),
-            funding_rate_hourly_percent: "0.002345".to_string(),
+            funding_rate_annualized_percent: order
+                .funding_rate
+                .to_decimal()
+                .checked_mul(Decimal::from(
+                    (24 / order.settlement_interval.whole_hours()) * 365,
+                ))
+                .expect("not to overflow")
```
thomaseizinger: The way we handled errors here previously is to fall back to an `Option` so we can at least display as much as possible and not everything blowing up. Perhaps also worth to create a new-type `AnnualizedFundingRate` and encapsulate this.

thomaseizinger (approved):

daemon/src/projection.rs
thomaseizinger: To make sure that these things don't panic in any case, proptests would be good. If you extract this logic into a function, it should be pretty easy to proptest.

klochowicz: Happy to create a newtype, but this cannot blow up - the rate can never be above `1.0`. IMHO If this blew up, we have a serious problem that should not be swept under the rug.

klochowicz: as this is a string, I could fallback to either an empty string or `--` would that be a preferred solution for you?

klochowicz:
@thomaseizinger I improved my solution - it's strongly typed & more idiomatic, can't get out of sync, and I tried to convey that FundingRate is a sane value too. 
I could go further with using the types instead of String, but don't want to cause problems for @da-kami  at this stage, I'll do that in a follow-up!

daemon/src/projection.rs
```diff
@@ -954,6 +944,46 @@ pub enum TxLabel {
...
+                .expect("not to overflow"),
+        )
+    }
+}
+
+impl fmt::Display for AnnualisedFundingRate {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.round_dp(2).fmt(f)
+    }
+}
```
thomaseizinger, line 966: There is a module somewhere that makes it very easy to format a `Decimal` with 2 decimal places in a `serde` serializer.

If we would implement `Serialize` for this instead, you could just store `AnnualisedFundingRate` within `CfdOrder` and stay in type-safe land :)

thomaseizinger (approved):

daemon/src/projection.rs
thomaseizinger: > as this is a string, I could fallback to either an empty string or `--` would that be a preferred solution for you?

My preferred solution would be an `Option`. In other places, the frontend knows that some of these are optional and displays a skeleton if the value is not available for some reason.

daemon/src/projection.rs
klochowicz: yes, I've seen that! I'll make a follow-up with an Option of the strong type on top of @da-kami 's changes. Added to my list for the day :)

PR #1143, opened 2022-01-19
Don't do async work in ctor of oracle and monitor actor

thomaseizinger:


daemon/src/model/cfd.rs
```diff
@@ -922,7 +922,7 @@ impl Cfd {
                 price: settlement.price,
             },
             Completed::Rejected { reason, .. } => {
-                tracing::info!(order_id=%self.id(), "Collaborative close rejected: {:#}", reason);
+                tracing::info!(order_id=%self.id(), "Collaborative close rejected: {:#}, force-closing the position", reason);
```
klochowicz, line 925: again, seeing  business logic spread out in multiple places (at least cfd model and projection) makes me worry a bit from maintenance perspective. does it need to go in with a commit and PR about removing async code from oracle and monitor? (here and below).

daemon/src/monitor.rs
```diff
@@ -69,6 +67,45 @@ pub struct MonitorParams {
 
 pub struct TryBroadcastTransaction {
     pub tx: Transaction,
+    pub kind: TransactionKind,
+}
+
+pub enum TransactionKind {
```
klochowicz, line 73: this seems really similar to our `TxLabel` struct in projection - perhaps someday we could align them

daemon/src/process_manager.rs
```diff
@@ -68,13 +68,13 @@ impl Actor {
                 tracing::info!("Setup complete, publishing on chain now");
 
                 let lock_tx = dlc.lock.0.clone();
-                let txid = self
-                    .try_broadcast_transaction
-                    .send(monitor::TryBroadcastTransaction { tx: lock_tx })
+                self.try_broadcast_transaction
+                    .send(monitor::TryBroadcastTransaction {
+                        tx: lock_tx,
+                        kind: TransactionKind::Lock,
```
klochowicz, line 74: perhaps we could wrap the transactions inside Dlc with the appropriate newtype that specifies the kind? then we could use it both here and when adding `TxLabel` in projection?

klochowicz (approved):
👍

daemon/src/process_manager.rs
thomaseizinger: That sounds like a good idea yes.

daemon/src/model/cfd.rs
thomaseizinger: I don't think the business logic is "spread out". The _decision_ to include the commit transaction in the event happens here. Following through with this decision is not business logic from my PoV.

klochowicz: the fact that it is included doesn't mean that it is broadcast, it only means that it *can* be. The way I see it it is the process manager is the one that performs the decision. 

The way I would picture it (and perhaps this is due to my limited knowledge about event sourcing), this event (and other similar events) could result in a command triggered from the process manager, like `force-close`. Then the logic would be clear and easy to follow. I don't think it's as clear with "if this transaction is sent in any of these events, then it's force-closing time" - perhaps it will be easier to follow once we adorn different kinds of transactions with newtypes, but we use fields in events for all sorts of stuff, and we shouldn't change them when we decide to change our business logic, as that will cause a lot of pain for us IMHO.
Happy to schedule a call to chat about it further :-)

thomaseizinger: I get what you are saying. I guess the PM is making a decision here (and orchestrating such a process is in fact the responsibility of a PM).

We don't have another aggregate that handles force-close because there isn't really anything to do other than broadcasting commit. I guess we could split it up by doing:

1. `Cfd` emits event `CollaborativeCloseFailed`
2. PM takes event and calls `Cfd::force_close` via `command::Executor`
3. `Cfd` emits event `ForceClosed` with `commit_tx` embedded

klochowicz: yes, this was something I was thinking of! 
note that it could also allow us to perform more elaborate workflows in PM, e.g. try to settle 3 times just in case there was some trivial error like lack of comms, and then follow up with the `Cfd::force_close` etc.
The projection could also look at `ForceClosed` event to accurately read the state.

da-kami (approved):
let's get this in

thomaseizinger:
> This PR was included in a batch that successfully built, but then failed to merge into master. It will not be retried.
> 
> Additional information:
> 
> ```json
> {"message":"Required status check \"block-fixup\" is expected.","documentation_url":"https://docs.github.com/articles/about-protected-branches"}
> ```
> 

Ah this needs a rebase on latest master.

daemon/src/model/cfd.rs
thomaseizinger: The thing to consider here is that it might result in a lot of back and forth between the aggregate and the PM, something that can cause confusion in what the actual process flow is.

From what I know, PM are usually used to orchestrate processes between different aggregates but of course that is just a rule of thumb. Unless we can point out a concrete problem, there is no reason we'd have to follow this rule blindly.

PR #1152, opened 2022-01-19
Block merging of fixup commits

klochowicz:
After accidentally merging a `fixup` commit with bors, I realised that there's gotta be an easy way of safeguarding us from it.

see: https://github.com/marketplace/actions/block-fixup-commit-merge

### DIFF
--- a/.github/workflows/git.yml
+++ b/.github/workflows/git.yml
@@ -1,0 +1,12 @@
+name: Git Checks
+
+on: [pull_request]
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2.0.0
+    - name: Block Fixup Commit Merge
+      uses: 13rac1/block-fixup-merge-action@v2.0.0


thomaseizinger (approved):
The accuracy with which the description of the actions matches our problem is hilarious!

bors.toml
```diff
@@ -8,4 +8,5 @@ status = [
   "test_daemons_release_mode (ubuntu-latest)",
   "test_daemons_release_mode (macos-latest)",
   "daemons_arm_build",
+  "block_fixup_commit_merge",
```
thomaseizinger: I think it is the job name (`block-fixup`) that is being tested against, not the step name.

klochowicz: fixing up, will verify whether bors blocks it :D

thomaseizinger: It won't because you've declared it only for `pull_request` events but bors pushes on a `staging` branch and then only checks that are triggered by `push` events are executed.

We can either:

- See if the action works with `push` events
- Put a GitHub merge restriction in place which will have bors fail because it can't trigger the merge through the API

klochowicz:
> bors r-

did it get stuck? :(

bors.toml
klochowicz: the second option seems nicer - I'll have a quick look at it!

thomaseizinger:
> > bors r-
> 
> did it get stuck? :(

Yes. bors waits for every status in the config file to appear and has a timeout of a couple of hours. If a check doesn't run on `push` it will never appear so it would have eventually run into a timeout.

bors.toml
thomaseizinger: I'll activate the check.

thomaseizinger:
bors r+

Borsing this because with the activated check, no other PR can merge.

PR #1153, opened 2022-01-19
Populate minimal opening fee in the projection

klochowicz:
Unfortunately, a Cfd offer from the maker is not enough to adequately calculate
opening fees, as we need quantity to do that.

For the time being, opt for calculating the fees for the min_quantity.
The total expected fees can be easily estimated by the user from that (or the UI
could even do a simple multiplication by `quantity / min_quantity`).

Another simpler approach could be the UI displaying:
    "opening fees from xxx for min_quantity"

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -10,6 +10,7 @@ use crate::model::cfd::Dlc;
 use crate::model::cfd::Event;
 use crate::model::cfd::OrderId;
 use crate::model::cfd::Role;
+use crate::model::FundingFee;
 use crate::model::FundingRate;
 use crate::model::Identity;
 use crate::model::Leverage;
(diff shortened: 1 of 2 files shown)


daemon/src/projection.rs
```diff
@@ -691,8 +692,11 @@ pub struct CfdOrder {
     pub price: Price,
 
     /// Fee charged by the maker for opening a position
-    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
-    pub opening_fee: Amount,
+    ///
+    /// Note: It's the minimum possible fee, we cannot calculate the exact
```
thomaseizinger, line 696: Hmm, this is unfortunate. I guess we need to do the same thing as we do for the margin. Calculate it per parcel and then calculate in the frontend.

thomaseizinger (approved):

daemon/src/projection.rs
klochowicz: yes, exactly (still getting used to the new terminology!) - `quantity / `min_quantity` is parcel size after all :)

PR #1156, opened 2022-01-19
Populate next funding event in the projection

klochowicz:
Move the field from order to cfd - we don't know the next funding event in the
order, as it hasn't been taken yet.

Estimate the value based on the next full hour from the latest contract setup or
rollover completion.

In fullness of time, these events should probably communicate when the next
rollover will happen.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -19,6 +19,7 @@ use crate::model::Price;
 use crate::model::Timestamp;
 use crate::model::TradingPair;
 use crate::model::Usd;
+use crate::oracle::ceil_to_next_hour;
 use crate::send_async_safe::SendAsyncSafe;
 use crate::Order;
 use crate::Tasks;
@@ -41,7 +42,6 @@ use serde::Serialize;
 use sqlx::pool::PoolConnection;
 use std::collections::HashSet;
 use std::time::Duration;
-use time::macros::datetime;
 use time::OffsetDateTime;
 use tokio::sync::watch;
 use xtra::prelude::MessageChannel;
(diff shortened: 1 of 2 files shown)


daemon/src/projection.rs
```diff
@@ -191,6 +191,10 @@ pub struct Cfd {
     #[serde(with = "round_to_two_dp::opt")]
     pub pending_settlement_proposal_price: Option<Price>,
 
+    /// Timestamp when the next fee will be collected
+    #[serde(with = "::time::serde::timestamp::option")]
+    pub next_funding_event: Option<OffsetDateTime>,
```
da-kami, line 196: Why do we have to do this in the projection? I think internally we can do that if necessary, but for the projection it would be good to keep it in the `CfdOrder` as it was.
We are displaying it globally in the frontend, and that what it should be.

If it is not part of the `Offer` then we can also just use full hour timestamps for now - that is what will be used as we align with the oracle event ID which is every full hour. 
@thomaseizinger thoughts?

da-kami (changes requested):

daemon/src/projection.rs
klochowicz: It can't be in the `CfdOrder` as you don't have any timestamp there, so you can't reason about it - or am I missing something?

da-kami: Maybe having it in the `CfdOrder` is confusing, it was a workaround - in the end I think it should always just be the next full hour timestamp. We could also just compute that in the UI for now if that makes it easier.

klochowicz: note that it should have the same value for all open cfds. I don't mind it being globally displayed, but then it shouldn't really be inside either Order or Cfd.

maybe it's a non-cfd event?

klochowicz:
As discussed with @da-kami on jitsi, we decided it's best to just hard-code the next hour in the UI: https://github.com/itchysats/itchysats/pull/1066#issuecomment-1016097476

PR #1157, opened 2022-01-19
Make supervisor actor more useful

thomaseizinger:
1. Expose metrics
2. Catch panics

### DIFF
--- a/daemon/src/supervisor.rs
+++ b/daemon/src/supervisor.rs
@@ -1,7 +1,10 @@
 use crate::xtra_ext::ActorName;
 use crate::Tasks;
 use async_trait::async_trait;
+use futures::FutureExt;
+use std::any::Any;
 use std::fmt;
+use std::panic::AssertUnwindSafe;
 use xtra::Address;
 use xtra::Context;
 use xtra::Message;
@@ -21,6 +24,8 @@ pub struct Actor<T, R> {
 struct Metrics {
     /// How many times the supervisor spawned an instance of the actor.
     pub num_spawns: u64,
+    /// How many times the actor shut down due to a panic.
+    pub num_panics: u64,
 }
 
 impl<T, R> Actor<T, R>
@@ -55,10 +60,21 @@ where
         tracing::info!("Spawning new instance of {actor}");
 
         let this = ctx.address().expect("we are alive");
-        let actor = (self.ctor)(this);
+        let actor = (self.ctor)(this.clone());
 
         self.metrics.num_spawns += 1;
-        self.tasks.add(self.context.attach(actor));
+        self.tasks.add({
+            let task = self.context.attach(actor);
+
+            async move {
+                match AssertUnwindSafe(task).catch_unwind().await {
+                    Ok(()) => {}
+                    Err(error) => {
+                        let _ = this.send(Panicked { error }).await;
+                    }
+                }
+            }
+        });
     }
 }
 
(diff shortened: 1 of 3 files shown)


daemon/src/supervisor.rs
```diff
@@ -142,6 +192,30 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn supervisor_tracks_panic_metrics() {
+        let _guard = tracing_subscriber::fmt().with_test_writer().set_default();
+
+        std::panic::set_hook(Box::new(|_| ())); // Override hook to avoid panic printing to log.
```
thomaseizinger, line 209: I need to test whether or not we also need to do this in production.

daemon/src/supervisor.rs
```diff
@@ -98,3 +111,60 @@ where
...
+    use xtra::Actor as _;
+
+    #[tokio::test]
+    async fn supervisor_tracks_spawn_metrics() {
+        let (supervisor, address) =
+            Actor::new(|supervisor| RemoteShutdown { supervisor }, |_| true);
+        let (supervisor, task) = supervisor.create(None).run();
+        tokio::spawn(task);
+
+        let metrics = supervisor.send(GetMetrics).await.unwrap();
```
rishflab, line 190: Why does sending a message spawn a new actor?

daemon/src/supervisor.rs
```diff
@@ -13,6 +13,13 @@ pub struct Actor<T, R> {
     ctor: Box<dyn Fn(Address<Self>) -> T + Send + 'static>,
     tasks: Tasks,
     restart_policy: Box<dyn FnMut(R) -> bool + Send + 'static>,
+    metrics: Metrics,
+}
+
+#[derive(Default, Clone, Copy)]
+pub struct Metrics {
+    /// How many times the supervisor spawned an instance of the actor.
```
rishflab, line 25: looking at the test it seems like like this is the number of messages the actor has handled?

rishflab (commented):
I can't really review the second commit as I have never recorded panics. Left some questions on the first commit.

daemon/src/supervisor.rs
thomaseizinger: It doesn't! What spawns the actor initially is the `started` lifecycle hook of the supervisor:

https://github.com/itchysats/itchysats/blob/b412ac01c5e5f8387210f39fb02e995414a13626/daemon/src/supervisor.rs#L61-L63

As a consequence of that, the `num_spawns` counter is 1 as soon as the supervisor is started.

daemon/src/supervisor.rs
thomaseizinger: Nope, if you look at the implementation, it is the number of times `spawn_new` has been called! :)

daemon/src/supervisor.rs
```diff
@@ -104,6 +120,29 @@ where
...
+            Ok(reason) => {
+                tracing::error!("{actor} panicked: {reason}");
+            }
+            Err(_) => {
+                tracing::error!("{actor} panicked");
+            }
+        }
+
+        self.metrics.num_panics += 1;
+        self.spawn_new(ctx)
```
luckysori, line 138: ❓ Why is panicking treated as an automatic restart (as opposed to stopping, which depends on the restart policy)?

luckysori (approved):
Cool!

daemon/src/supervisor.rs
thomaseizinger: So I figured that because the cause of the panic is unknown at coding time, it doesn't make much sense to implement a policy.

If I would know the problem already to code the policy, I could also just fix the problem in the first place 😁

The idea of this experiment is that hopefully the panic is temporary and restarting the actor puts it in a known state where we can progress again. Perhaps only for a while if the panic was triggered by recurring data.

So we certainly need to monitor these metrics!

I am thinking of rewriting the trace statements here and putting in a dedicated subscriber. Probably in a follow-up though!

daemon/src/supervisor.rs
```diff
@@ -104,6 +120,29 @@ where
...
+where
+    T: xtra::Actor,
+    R: fmt::Display + 'static,
+{
+    async fn handle(&mut self, msg: Panicked, ctx: &mut Context<Self>) {
+        let actor = T::name();
+
+        match msg.error.downcast::<String>() {
+            Ok(reason) => {
+                tracing::error!("{actor} panicked: {reason}");
```
da-kami: Wouldn't it make sense to add the metrics to the logs so we see them?

thomaseizinger: I am thinking about that yeah. Will push an update later!

thomaseizinger: I opted for adding more structured logs, that should make it easier to compute such metrics downstream. Should I remove the metrics again? Only added them for testing really ..

thomaseizinger: I opted for making these things module private for now.

PR #1169, opened 2022-01-20
Document rollover 2.0 protocol in ADR

thomaseizinger:
Proposal for a protocol design based on the discussion in https://github.com/itchysats/itchysats/discussions/1068 and from the event storming meeting yesterday.

### DIFF
--- a/docs/adr/014-rollover-2.0.md
+++ b/docs/adr/014-rollover-2.0.md
@@ -43,6 +43,8 @@ Running rollover at exactly these points makes implementations easier because th
 
 It makes UI design easier because we can clearly state, when the next funding event happens, and it does not depend on the maker that we are connected to.
 
+Another benefit of fixed funding periods is that our oracle actor only needs to fetch attestations for specific points in time and not every hour.
+
 ### 2. Each party should start the protocol with an expectation of the funding rate
 
 In the future, the funding rate will be determined by the market.
--- a/docs/adr/014-rollover-2.0.md
+++ b/docs/adr/014-rollover-2.0.md
@@ -43,18 +43,29 @@ Running rollover at exactly these points makes implementations easier because th
 
 It makes UI design easier because we can clearly state, when the next funding event happens and it does not depend on the maker that we are connected to.
 
-### 2. Enforce agreement of the funding rate through the blockchain protocol
+### 2. Each party should start the protocol with an expectation of the funding rate.
+
+In the future, the funding rate will be determined by the market.
+As such, no party should be specifying the funding rate as part of the protocol.
+
+### 3. Maker streams current funding rate to taker
+
+To simulate a market whilst we don't yet have one, introduce a stream of funding rates from the maker to the taker.
+Once we have a market, this component will be removed.
+Up until this, this stream we 
+
+### 4. Enforce agreement of the funding rate through the blockchain protocol
 
 Maker to stream current funding rate to taker -> required for display in UI.
 Taker use this funding rate to start rollover -> if they use the same, sig-verification succeeds.
 
-### 3. Remove approval (accept / decline) from the protocol.
+### 5. Remove approval (accept / decline) from the protocol.
    
 With (1) and (2) in mind, there is really no point in an external approval step for rollover.
 As long as the CFD's is not committed on-chain, rollover will occur.
 To opt out of rollover, either party can commit the CFD on-chain which will result in a payout at the end of the funding period.
 
-### 4. Connection dialer sends initial trigger message
+### 6. Connection dialer sends initial trigger message
 
 - Need unambiguous way to trigger protocol
 - Scales to peer to peer setting
@@ -70,3 +81,4 @@ To opt out of rollover, either party can commit the CFD on-chain which will resu
 1. Should we avoid the in-efficiency in doing a rollover when both parties don't align on the funding rate?
 2. When should peers trigger rollover? 5min before the funding period ends? 10min? What is enough time?
 3. With fixed funding periods, we get a lot of load on very specific points in time. Is that an issue? 
+4. Should the stream of funding rates be incorporated into the current offer?
--- a/docs/adr/014-rollover-2.0.md
+++ b/docs/adr/014-rollover-2.0.md
@@ -41,7 +41,7 @@ The exact timestamps are not decided yet, but they will be something along the l
 
 Running rollover at exactly these points makes implementations easier because they can assume that well-behaved peer will execute the protocol at these points in time.
 
-It makes UI design easier because we can clearly state, when the next funding event happens and it does not depend on the maker that we are connected to.
+It makes UI design easier because we can clearly state, when the next funding event happens, and it does not depend on the maker that we are connected to.
 
 ### 2. Each party should start the protocol with an expectation of the funding rate
 
@@ -51,12 +51,14 @@ As such, no party should be specifying the funding rate as part of the protocol.
 ### 3. Maker streams current funding rate to taker
 
 To simulate a market whilst we don't yet have one, introduce a stream of funding rates from the maker to the taker.
+This stream can be used to display the current funding rate in the UI.
 Once we have a market, this component will be removed.
 
 ### 4. Enforce agreement of the funding rate through the blockchain protocol
 
-Maker to stream current funding rate to taker -> required for display in UI.
-Taker use this funding rate to start rollover -> if they use the same, sig-verification succeeds.
+With both parties entering the next funding period with a specific funding rate, the protocol will simply fail due to signature verification errors if the funding rates don't match.
+This should be sufficient for an MVP of this new protocol.
+More elaborate detection of what the cause of the failure was can easily be added on top.
 
 ### 5. Remove approval (accept / decline) from the protocol
    
--- a/docs/adr/014-rollover-2.0.md
+++ b/docs/adr/014-rollover-2.0.md
@@ -29,7 +29,7 @@ To harmonize, how these fees are paid and where they are coming from, we want to
 
 ## The new protocol
 
-### 1. We define fixed UTC timestamps for the funding periods.
+### 1. We define fixed UTC timestamps for the funding periods
    
 The exact timestamps are not decided yet but it will be something along the lines of:
    - Every day at 0400 UTC.
@@ -43,7 +43,7 @@ Running rollover at exactly these points makes implementations easier because th
 
 It makes UI design easier because we can clearly state, when the next funding event happens and it does not depend on the maker that we are connected to.
 
-### 2. Each party should start the protocol with an expectation of the funding rate.
+### 2. Each party should start the protocol with an expectation of the funding rate
 
 In the future, the funding rate will be determined by the market.
 As such, no party should be specifying the funding rate as part of the protocol.
@@ -59,7 +59,7 @@ Up until this, this stream we
 Maker to stream current funding rate to taker -> required for display in UI.
 Taker use this funding rate to start rollover -> if they use the same, sig-verification succeeds.
 
-### 5. Remove approval (accept / decline) from the protocol.
+### 5. Remove approval (accept / decline) from the protocol
    
 With (1) and (2) in mind, there is really no point in an external approval step for rollover.
 As long as the CFD's is not committed on-chain, rollover will occur.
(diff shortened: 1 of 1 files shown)


docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,66 @@
...
+2. Not require approval by either party.
+
+## Context
+
+Currently, rollover is a protocol that is initiated by the taker pretty much on the hour every hour.
+The exact conditions for triggering are that we want the timestamp of the price event that is used for attestation to be 24h in the future.
+Because the oracle attests on the price exactly every hour, we can rollover to a new price event on the hour.
+
+Rollover is coupled to paying fees.
+In particular, the time window until the attestation can be considered as the current funding period.
```
da-kami, line 23: So far our funding interval was one hour, i.e. we charge every hour. That is not the same as the settlement interval that was 24h so far.

Are we planning to align them, so that the settlement interval is always the same as the funding interval, i.e. if the next funding event is at `04:00 UTC` then the oracle event ID is also at `04:00 UTC`? 
We cannot rollover exactly at the time of the oracle event ID, because once attested we cannot rollover anymore but have to force close. There needs to be an agreement between the two parties to roll-over before the expiry timestamp. 

Would it make sense to hard-code that rollover is always triggered e.g. 5 minutes before the expiry event?

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,66 @@
...
+   - Every day at 0000 UTC.
+
+Running rollover at exactly these points makes implementations easier because they can assume that well-behaved peer will execute the protocol at these points in time.
+
+It makes UI design easier because we can clearly state, when the next funding event happens and it does not depend on the maker that we are connected to.
+
+### 2. Enforce agreement of the funding rate through the blockchain protocol
+
+Maker to stream current funding rate to taker -> required for display in UI.
+Taker use this funding rate to start rollover -> if they use the same, sig-verification succeeds.
```
bonomat: You would not build in an explicit check? The verification comes after the generation which is quite resource intensive. Imho it would be better to check first and don't compute the signatures if the two are not aligned.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,66 @@
...
+## Context
+
+Currently, rollover is a protocol that is initiated by the taker pretty much on the hour every hour.
+The exact conditions for triggering are that we want the timestamp of the price event that is used for attestation to be 24h in the future.
+Because the oracle attests on the price exactly every hour, we can rollover to a new price event on the hour.
+
+Rollover is coupled to paying fees.
+In particular, the time window until the attestation can be considered as the current funding period.
+For every new funding period, fees are charged depending on the current market.
+In a bull market, shorts pay longs.
```
luckysori: 📣 This contradicts this sentence from https://github.com/itchysats/itchysats/discussions/1167:

> The funding rate can be positive or negative. If the funding rate is positive then there is a bullish sentiment in the market. At these times longs pay the shorts.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,66 @@
...
+
+### 1. We define fixed UTC timestamps for the funding periods.
+   
+The exact timestamps are not decided yet but it will be something along the lines of:
+   - Every day at 0400 UTC.
+   - Every day at 0800 UTC.
+   - Every day at 1200 UTC.
+   - Every day at 1600 UTC.
+   - Every day at 2000 UTC.
+   - Every day at 0000 UTC.
```
luckysori: ❓ Is such a policy meant to be defined by makers or is it encoded in the software? If it is part of the software, could it be too restrictive for future makers?

luckysori (approved):
I like this proposal.

docs/adr/014-rollover-2.0.md
thomaseizinger: > You would not build in an explicit check?

See the explanation here: https://github.com/itchysats/itchysats/discussions/1068#discussioncomment-2001308
**tl;dr:** It is not really worth much because you can't trust it.

The linked post doesn't take into account resource consumption but given that we are talking about an edge case here, I am hesitant to design features specifically for this into the protocol. I am happy to consider alternative proposals but an explicit handshake seems superfluous.

docs/adr/014-rollover-2.0.md
thomaseizinger: > Are we planning to align them, so that the settlement interval is always the same as the funding interval, i.e. if the next funding event is at `04:00 UTC` then the oracle event ID is also at `04:00 UTC`?

Yes, I think we should do that.

> because once attested we cannot rollover anymore but have to force close.

That is the policy we currently have but I don't think we'd have to be that strict here. We could delay force-closing for a couple of minutes.

> There needs to be an agreement between the two parties to roll-over before the expiry timestamp.

Yep. I don't yet know, what the best solution is here.

docs/adr/014-rollover-2.0.md
thomaseizinger: The idea is that it is encoded in the software / protocol so that once things are p2p, all nodes simply adhere to this and it is not maker policy.

docs/adr/014-rollover-2.0.md
thomaseizinger: I've added an open question that allows us to discuss this.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+
+Running rollover at exactly these points makes implementations easier because they can assume that well-behaved peer will execute the protocol at these points in time.
+
+It makes UI design easier because we can clearly state, when the next funding event happens, and it does not depend on the maker that we are connected to.
+
+Another benefit of fixed funding periods is that our oracle actor only needs to fetch attestations for specific points in time and not every hour.
```
bonomat, line 46: For our _dynamic liquidation_ feature we will need to fetch more regularly.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+## Considerations
+
+1. The new protocol requires being online at a very specific point in time.
+   
+   This requirement is already implied by the fact that we have an off-chain channel construct and we need to be constantly online to monitor for the broadcast of old commit transactions.
+
+## Open questions
+
+1. Should we avoid the in-efficiency in doing a rollover when both parties don't align on the funding rate?
+2. When should peers trigger rollover? 5min before the funding period ends? 10min? What is enough time?
```
bonomat, line 89: A related thought for the maker: 

Should a maker commit on-chain on the second of expiry if no rollover was trigger, 1+second afterwards or maybe even before to ensure to get his funds in-time?

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+3. With fixed funding periods, we get a lot of load on very specific points in time. Is that an issue?
```
bonomat, line 90: Are you thinking of the load on maker/taker side? There is also the load on the oracle if thousands of traders rollover at the same time.

bonomat (approved):
Thanks for writing this up. Imho one important thing here is that the explicit approval/rejection of the rollover request will go away.

docs/adr/014-rollover-2.0.md
thomaseizinger: True, not a concern of this protocol though so at when viewed in isolation, this is still true!

docs/adr/014-rollover-2.0.md
thomaseizinger: I specifically called this `peer` here so this was also referring to the maker!

Yes, we will have to define a policy on when a party commits as a result of a lacking rollover.

docs/adr/014-rollover-2.0.md
thomaseizinger: The load on the oracle is negligible because it doesn't compute anything. Fetching attestations and announcements are read-only operations that can be covered with caches. AFAIK, the Olivia deployment already does that.

This was primarily referring to the maker as they are the one that is connected to many parties at the same time. In a p2p setting, this will eventually apply to all peers. It ties back in with question (2). The smaller the window for triggering rollover, the more peers will need to compute something in a short amount of time.

thomaseizinger:
> Imho one important thing here is that the explicit approval/rejection of the rollover request will go away.

Yes. Are you bringing this up because you want it mentioned more prominently?
It is my current understanding that "rollover" should evolve into a protocol that is consensus-like concept: Every peer expects every other peer to participate in it according to the network rules. If they don't, you commit on-chain. As such, it seems like we shouldn't have an accept/reject functionality in the protocol.

bonomat:
> > Imho one important thing here is that the explicit approval/rejection of the rollover request will go away.
> 
> Yes. Are you bringing this up because you want it mentioned more prominently? It is my current understanding that "rollover" should evolve into a protocol that is consensus-like concept: Every peer expects every other peer to participate in it according to the network rules. If they don't, you commit on-chain. As such, it seems like we shouldn't have an accept/reject functionality in the protocol.

Yes, I just wanted to point this out again hoping others of the team will see this as well :)

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+To opt out of rollover, either party can commit the CFD on-chain which will result in a payout at the end of the funding period.
+
+### 6. Connection listener sends initial trigger message
+
+To make the system more predictable, we should have an dedicated trigger for the protocol.
+One way of ensuring this is deterministic, is to check whether a peer is the dialer or the listener in a connection.
+
+In our current setup, the maker is the listener.
+If the have the maker send out the initial trigger message, we have _some_ control over how many rollovers we want to do in parallel.
+If the taker would send the intial trigger message and we don't build in some kind of randomness in when to send it, rollover might end up being a DoS attack against the maker.
```
luckysori, line 78: 👍 It makes sense. I suppose this is important because there is only one maker at the moment, so this design may change if/when we go fully p2p.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Considerations
+
+1. The new protocol requires being online at a very specific point in time.
+   
+   This requirement is already implied by the fact that we have an off-chain channel construct and we need to be constantly online to monitor for the broadcast of old commit transactions.
+
+## Open questions
+
+1. Should we avoid the in-efficiency in doing a rollover when both parties don't align on the funding rate?
```
luckysori, line 88: Just let rollover fail at this stage, if that's what this question is getting at.

docs/adr/014-rollover-2.0.md
```diff
@@ -0,0 +1,91 @@
...
+2. When should peers trigger rollover? 5min before the funding period ends? 10min? What is enough time?
+3. With fixed funding periods, we get a lot of load on very specific points in time. Is that an issue? 
+4. Should the stream of funding rates be incorporated into the current offer?
```
luckysori, line 91: Why would you want to do this? I would prefer the funding rate stream to be separate, particularly since we plan to remove it according to point number 2 in `The new protocol` section.

docs/adr/014-rollover-2.0.md
thomaseizinger: That would have been my initial idea as well but @bonomat rightfully pointed out that it creates an in-efficiency where we do unnecessary computations that end up not working.

I am torn here. On the one hand, we should obviously be as efficient as possible. However, there are really only two scenarios in which this could happen (I think):

- There is a bug in our software where both parties compute the wrong funding rate (assuming it somehow consensus / market-driven)
- We have a rouge peer that wants to DoS us by making us do rollovers that purposely fail

An initial check could circumvent the first problem but not the second because a rouge peer can just send on value in the initial check and then send transactions that are computed using a different funding rate.

Considering all this, I think the value of a check is pretty small. In normal operation, all peers should always compute the same funding rate and in the case of malicious peers, we would need to deal with that differently anyway (like ban / disconnect from them if they show funny behaviour).

docs/adr/014-rollover-2.0.md
thomaseizinger: The idea was that we are already sending offers to our peers so might as well include that information in there but I think you are right and it is cleaner if it is separate.

PR #1170, opened 2022-01-20
Store FundingFee in Cfd events

klochowicz:
Allows calculating the accumulated fees

Fixes #1170

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -6,6 +6,7 @@ use anyhow::Result;
 use bdk::bitcoin::Address;
 use bdk::bitcoin::Amount;
 use bdk::bitcoin::Denomination;
+use bdk::bitcoin::SignedAmount;
 use chrono::DateTime;
 use derive_more::Display;
 use reqwest::Url;
@@ -637,12 +638,20 @@ impl Default for FundingRate {
 
 /// Fee paid on every contract renewal (a fraction of it if rolling over)
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
-pub struct FundingFee(u64);
+pub struct FundingFee(i64);
 
 impl FundingFee {
     /// Value in satoshis
-    pub fn as_u64(&self) -> u64 {
+    pub fn as_i64(&self) -> i64 {
         self.0
+    }
+}
+
+impl TryFrom<FundingFee> for u64 {
+    type Error = anyhow::Error;
+
+    fn try_from(value: FundingFee) -> Result<Self, Self::Error> {
+        value.0.try_into().context("Cannot represent u64")
     }
 }
 
@@ -653,14 +662,18 @@ impl Default for FundingFee {
     }
 }
 
-impl From<FundingFee> for Amount {
+impl From<FundingFee> for SignedAmount {
     fn from(funding_fee: FundingFee) -> Self {
-        Self::from_sat(funding_fee.as_u64())
+        Self::from_sat(funding_fee.as_i64())
     }
 }
 
 impl FundingFee {
-    pub fn new(margin: Amount, funding_rate: FundingRate, hours_to_charge: i64) -> Result<Self> {
+    pub fn new(
+        margin: Amount, // Margin cannot be negative, so use `Amount`
+        funding_rate: FundingRate,
+        hours_to_charge: i64,
+    ) -> Result<Self> {
         anyhow::ensure!(hours_to_charge >= 1, "Can't charge for less than one hour");
         anyhow::ensure!(
             hours_to_charge <= SETTLEMENT_INTERVAL.whole_hours(),
@@ -686,11 +699,11 @@ fn calculate_funding_fee(
     margin: Amount,
     funding_rate: FundingRate,
     fraction_of_funding_period: Decimal,
-) -> Result<u64> {
+) -> Result<i64> {
     (Decimal::from(margin.as_sat()) * funding_rate.to_decimal() * fraction_of_funding_period)
         .round_dp_with_strategy(0, rust_decimal::RoundingStrategy::AwayFromZero)
-        .to_u64()
-        .context("Failed to represent as u64")
+        .to_i64()
+        .context("Failed to represent as i64")
 }
 
 #[cfg(test)]
@@ -826,7 +839,7 @@ mod tests {
         )
         .unwrap();
 
-        assert_eq!(funding_fee.as_u64(), 0);
+        assert_eq!(funding_fee.as_i64(), 0);
     }
 
     #[test]
@@ -841,7 +854,7 @@ mod tests {
         )
         .unwrap();
 
-        assert_eq!(funding_fee.as_u64(), 1);
+        assert_eq!(funding_fee.as_i64(), 1);
     }
 
     use proptest::prelude::*;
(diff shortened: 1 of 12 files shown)


klochowicz:
I was actually tempted to store the funding fee inside `Dlc` as that would reduce the diff quite a bit - another thing that would be useful would be to store the amount of hours the fee was collected for.

klochowicz:
converted to draft for now, as I realised that this would also show the fees the same on the maker - it needs to be a `SignedAmount` to display the positive in the maker.

klochowicz:
> converted to draft for now, as I realised that this would also show the fees the same on the maker - it needs to be a `SignedAmount` to display the positive in the maker.

Ready for review.

Note: using `funding_fee.inverse()` is a bit of a hack, but with incoming PR testing the margins and the fees, at least it will be covered by unit and actor level tests to ensure sane behaviour.

daemon/src/model/cfd.rs
```diff
@@ -1135,9 +1139,10 @@ impl Cfd {
 
         match evt.event {
             ContractSetupStarted => self.during_contract_setup = true,
-            ContractSetupCompleted { dlc } => {
+            ContractSetupCompleted { dlc, funding_fee } => {
                 self.dlc = Some(dlc);
                 self.during_contract_setup = false;
+                todo!("{funding_fee:?}");
```
thomaseizinger: Unhandled TODO? This will blow up the program no?

daemon/src/setup_contract.rs
```diff
@@ -355,25 +355,28 @@ pub async fn new(
...
+            commit: (commit_tx, msg1.commit, commit_desc),
+            cets,
+            refund: (refund_tx, msg1.refund),
+            maker_lock_amount: params.maker().lock_amount,
+            taker_lock_amount: params.taker().lock_amount,
+            revoked_commit: Vec::new(),
+            settlement_event_id,
+            refund_timelock: setup_params.refund_timelock,
+        },
+        setup_params.funding_fee,
```
thomaseizinger, line 378: We pass this into the function, why are we passing it out again? :D

daemon/src/setup_contract.rs
```diff
@@ -678,25 +681,28 @@ pub async fn roll_over(
...
+            commit: (commit_tx, msg1.commit, commit_desc),
+            cets,
+            refund: (refund_tx, msg1.refund),
+            maker_lock_amount,
+            taker_lock_amount,
+            revoked_commit,
+            settlement_event_id: announcement.id,
+            refund_timelock: rollover_params.refund_timelock,
+        },
+        rollover_params.funding_fee,
```
thomaseizinger, line 704: Same here, this seems unnecessary.

daemon/src/model/cfd.rs
```diff
@@ -310,6 +310,7 @@ pub enum CfdEvent {
     ContractSetupStarted,
```
thomaseizinger, line 310: We compute the fee together with the `Started` event. Would it make more sense to store the fee in here instead?

daemon/src/model.rs
```diff
@@ -637,30 +638,42 @@ impl Default for FundingRate {
 
 /// Fee paid on every contract renewal (a fraction of it if rolling over)
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
-pub struct FundingFee(u64);
+pub struct FundingFee(i64);
 
 impl FundingFee {
     /// Value in satoshis
-    pub fn as_u64(&self) -> u64 {
+    pub fn as_i64(&self) -> i64 {
```
thomaseizinger, line 645: Nit: I think the correct fn sig would be:

```suggestion
    pub fn to_i64(self) -> i64 {
```

because:

1. The type is copy, so consume `self`
2. The conversion is "expensive" because it involves more than pointers

thomaseizinger (approved):
Nit: Commit message of a2fe726eade9b6c3d7a0b29d7cde512784d39c55 is a bit too long :)

Overall, the changes are a bit scattered with the todo!s and stuff so I think squashing would be appropriate.

daemon/src/model/cfd.rs
```diff
@@ -1151,9 +1161,10 @@ impl Cfd {
                 self.during_rollover = true;
             }
             RolloverAccepted => {}
-            RolloverCompleted { dlc } => {
+            RolloverCompleted { dlc, funding_fee } => {
                 self.dlc = Some(dlc);
                 self.during_rollover = false;
+                todo!("{funding_fee:?}");
```
da-kami:
```suggestion
                self.total_funding_fees
                    .checked_add(funding_fee.into())
                    .expect("addition not to overflow");
```

daemon/src/rollover_maker.rs
```diff
@@ -40,6 +41,7 @@ pub struct ProtocolMsg(pub wire::RolloverMsg);
 /// notify that rollover has finished successfully.
 struct RolloverSucceeded {
     dlc: Dlc,
+    funding_fee: FundingFee,
 }
 
 /// Message sent from the spawned task to `rollover_taker::Actor` to
```
da-kami:
```suggestion
/// Message sent from the spawned task to `rollover_maker::Actor` to
```

unrelated. I noticed that this comment is confusing, I think it should be the maker here...?

da-kami (approved):
Happy to get this in, but we should remove the `todo!`!

One thing I noticed that is not directly related to this PR - but since we are adding the accumulated fees here it might be relevant:

Currently we pass the funding fee for *one* interval into rollover:

https://github.com/itchysats/itchysats/blob/28f7fa1836c12716a0595b862574204669dffbe8/daemon/src/model/cfd.rs#L724-L737

Is this correct in the context of accumulated fees - should it be `current_funding_fee + accumulated_previous_funding_fees` instead? 
How is the fee bucket managed - does the payout curve expect the accumulated fees, or just the fee for the last interval?

daemon/src/model/cfd.rs
klochowicz: not really - because you would suddenly need to reason about all the failures/rejections, which make things heaps more complicated for no good reason. I opted to put the information in when we can be sure that the user *has been* charged.

daemon/src/model/cfd.rs
klochowicz: oops! of course this needs to be dealt with, thanks for spotting the error!

klochowicz:
> Happy to get this in, but we should remove the `todo!`!
> 
> One thing I noticed that is not directly related to this PR - but since we are adding the accumulated fees here it might be relevant:
> 
> Currently we pass the funding fee for _one_ interval into rollover:
> 
> https://github.com/itchysats/itchysats/blob/28f7fa1836c12716a0595b862574204669dffbe8/daemon/src/model/cfd.rs#L724-L737
> 
> Is this correct in the context of accumulated fees - should it be `current_funding_fee + accumulated_previous_funding_fees` instead? How is the fee bucket managed - does the payout curve expect the accumulated fees, or just the fee for the last interval?

That's a good question, let's chat about it after this gets in - you might be on to something!

Note: I'm keen to write a test today to document the behaviour we want during the setup and a couple of rollovers.

thomaseizinger:
> How is the fee bucket managed - does the payout curve expect the accumulated fees, or just the fee for the last interval?

It expects accumulated fees yes.

It is a pure function so it will only take into account what you pass in.

daemon/src/model.rs
klochowicz: indeed! thanks for pointing it out!I followed bdk convention, I think they should have had `to_sats()` too!

klochowicz:
reopened after today's discussions as #1185 - it changed quite a bit since then.

PR #1172, opened 2022-01-20
Small fix for UI tx-list decision logic

da-kami:
`Force + Payout` when disconnected is only relevant in `Open`.
Otherwise collaborativly closed positions will not be shown correctly when disconnected from the maker.


bonomat (approved):
👍

PR #1181, opened 2022-01-21
Do not let maker accept settlement twice

luckysori:
Same logic as what we do in the `rollover_maker::Actor`.


daemon/src/collab_settlement_maker.rs
```diff
@@ -192,6 +199,8 @@ impl Actor {
             .await
             .context("Failed to inform taker about settlement acceptance")??;
 
+        self.has_accepted = true;
```
thomaseizinger: Somehow I would feel more comfortable if we would do this right after the check.

thomaseizinger (approved):

PR #1182, opened 2022-01-21
Profit based on actual payout if the payout is known

da-kami:
Prior to adding the fees on top of the profit and loss calculation I thought it's probably a good idea to finally fix some tech debt. 

The profit and loss should always be based on the payout, and if we know the specific payout we should use that instead of calculating from a price.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -38,8 +38,8 @@ use maia::secp256k1_zkp::SECP256K1;
 use maia::spending_tx_sighash;
 use maia::TransactionExt;
 use rocket::request::FromParam;
-use rust_decimal::prelude::FromPrimitive;
 use rust_decimal::Decimal;
+use rust_decimal_macros::dec;
 use serde::de::Error as _;
 use serde::Deserialize;
 use serde::Serialize;
@@ -1234,25 +1234,38 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
     price * leverage / (leverage + 1)
 }
 
-/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
-pub fn calculate_profit(
-    initial_price: Price,
+pub fn calculate_profit(payout: SignedAmount, margin: SignedAmount) -> (SignedAmount, Percent) {
+    let profit = payout - margin;
+
+    let profit_sats = Decimal::from(profit.as_sat());
+    let margin_sats = Decimal::from(margin.as_sat());
+    let percent = dec!(100) * profit_sats / margin_sats;
+
+    (profit, Percent(percent))
+}
+
+/// Returns the profit/loss and payout capped by the provided margin
+///
+/// All values are calculated without using the payout curve.
+/// Profit/loss is returned as signed bitcoin amount and percent.
+pub fn calculate_profit_at_price(
+    opening_price: Price,
     closing_price: Price,
     quantity: Usd,
     leverage: Leverage,
     position: Position,
-) -> Result<(SignedAmount, Percent)> {
+) -> Result<(SignedAmount, Percent, SignedAmount)> {
     let inv_initial_price =
-        InversePrice::new(initial_price).context("cannot invert invalid price")?;
+        InversePrice::new(opening_price).context("cannot invert invalid price")?;
     let inv_closing_price =
         InversePrice::new(closing_price).context("cannot invert invalid price")?;
-    let long_liquidation_price = calculate_long_liquidation_price(leverage, initial_price);
+    let long_liquidation_price = calculate_long_liquidation_price(leverage, opening_price);
     let long_is_liquidated = closing_price <= long_liquidation_price;
 
-    let long_margin = calculate_long_margin(initial_price, quantity, leverage)
+    let long_margin = calculate_long_margin(opening_price, quantity, leverage)
         .to_signed()
         .context("Unable to compute long margin")?;
-    let short_margin = calculate_short_margin(initial_price, quantity)
+    let short_margin = calculate_short_margin(opening_price, quantity)
         .to_signed()
         .context("Unable to compute short margin")?;
     let amount_changed = (quantity * inv_initial_price)
@@ -1308,11 +1321,8 @@ pub fn calculate_profit(
         }
     };
 
-    let profit = payout - margin;
-    let percent = Decimal::from_f64(100. * profit.as_sat() as f64 / margin.as_sat() as f64)
-        .context("Unable to compute percent")?;
-
-    Ok((profit, Percent(percent)))
+    let (profit_btc, profit_percent) = calculate_profit(payout, margin);
+    Ok((profit_btc, profit_percent, payout))
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
@@ -1891,7 +1901,7 @@ mod tests {
             Leverage::new(2).unwrap(),
             Position::Long,
             SignedAmount::from_sat(3_174_603),
-            dec!(31.99999798400001).into(),
+            dec!(31.999997984000016127999870976).into(),
             "long position should make a profit when price goes up",
         );
 
@@ -1902,7 +1912,7 @@ mod tests {
             Leverage::new(2).unwrap(),
             Position::Short,
             SignedAmount::from_sat(-3_174_603),
-            dec!(-15.99999899200001).into(),
+            dec!(-15.999998992000008063999935488).into(),
             "short position should make a loss when price goes up",
         );
     }
@@ -1910,7 +1920,7 @@ mod tests {
     #[allow(clippy::too_many_arguments)]
     fn assert_profit_loss_values(
         initial_price: Price,
-        current_price: Price,
+        closing_price: Price,
         quantity: Usd,
         leverage: Leverage,
         position: Position,
@@ -1918,8 +1928,11 @@ mod tests {
         should_profit_in_percent: Percent,
         msg: &str,
     ) {
-        let (profit, in_percent) =
-            calculate_profit(initial_price, current_price, quantity, leverage, position).unwrap();
+        // TODO: Assert on payout as well
+
+        let (profit, in_percent, _) =
+            calculate_profit_at_price(initial_price, closing_price, quantity, leverage, position)
+                .unwrap();
 
         assert_eq!(profit, should_profit, "{}", msg);
         assert_eq!(in_percent, should_profit_in_percent, "{}", msg);
@@ -1931,7 +1944,7 @@ mod tests {
         let closing_price = Price::new(dec!(16_000)).unwrap();
         let quantity = Usd::new(dec!(10_000));
         let leverage = Leverage::new(1).unwrap();
-        let (profit, profit_in_percent) = calculate_profit(
+        let (profit, profit_in_percent, _) = calculate_profit_at_price(
             initial_price,
             closing_price,
             quantity,
@@ -1939,7 +1952,7 @@ mod tests {
             Position::Long,
         )
         .unwrap();
-        let (loss, loss_in_percent) = calculate_profit(
+        let (loss, loss_in_percent, _) = calculate_profit_at_price(
             initial_price,
             closing_price,
             quantity,
@@ -1982,11 +1995,17 @@ mod tests {
         ];
 
         for price in closing_prices {
-            let (long_profit, _) =
-                calculate_profit(initial_price, price, quantity, leverage, Position::Long).unwrap();
-            let (short_profit, _) =
-                calculate_profit(initial_price, price, quantity, leverage, Position::Short)
+            let (long_profit, _, _) =
+                calculate_profit_at_price(initial_price, price, quantity, leverage, Position::Long)
                     .unwrap();
+            let (short_profit, _, _) = calculate_profit_at_price(
+                initial_price,
+                price,
+                quantity,
+                leverage,
+                Position::Short,
+            )
+            .unwrap();
 
             assert_eq!(
                 long_profit + long_margin + short_profit + short_margin,
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -4,6 +4,7 @@ use crate::model;
 use crate::model::cfd::calculate_long_liquidation_price;
 use crate::model::cfd::calculate_long_margin;
 use crate::model::cfd::calculate_profit;
+use crate::model::cfd::calculate_profit_at_price;
 use crate::model::cfd::calculate_short_margin;
 use crate::model::cfd::CfdEvent;
 use crate::model::cfd::Dlc;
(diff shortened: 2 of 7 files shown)


taker-frontend/src/types.ts
```diff
@@ -58,8 +58,10 @@ export interface Cfd {
 
     margin: number;
 
-    profit_btc?: number;
-    profit_percent?: number;
+    profit_btc: number;
+    profit_percent: number;
+
+    payout: number;
```
thomaseizinger: This commit adds the field here but it is not yet present in the backend? :)

daemon/src/projection.rs
```diff
@@ -208,18 +220,76 @@ pub struct Cfd {
...
+            role,
+        }
+    }
+
+    fn payout(self) -> Option<Amount> {
+        if let Some((tx, script)) = self.collab_settlement_tx {
+            return Some(extract_payout_amount(tx, script));
+        }
+
+        match (self.cet, self.timelocked_cet) {
```
thomaseizinger: You can `?` all of these and it will bubble up the `None`.

daemon/src/projection.rs
```diff
@@ -208,18 +220,76 @@ pub struct Cfd {
...
+            collab_settlement_tx: None,
+            cet: None,
+            timelocked_cet: None,
+            commit_published: false,
+            refund_published: false,
+            role,
+        }
+    }
+
+    fn payout(self) -> Option<Amount> {
```
thomaseizinger: This could take `role` as a parameter to avoid the field.

da-kami:
I noticed that there are still some todos in the code - would be great if we could split up work on fixing those (if we want them to be fixed). I am rebasing my prep-work for the funding-fee changes on top of this now to be able to estimate how much will have to be adapted :)

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
+    margin: SignedAmount,
+) -> Result<(SignedAmount, Percent)> {
+    let profit = payout - margin;
+    let percent = Decimal::from_f64(100. * profit.as_sat() as f64 / margin.as_sat() as f64)
+        .context("Unable to compute percent")?;
+
+    Ok((profit, Percent(percent)))
+}
+
+/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
```
bonomat: This description does not seem to be up2date. Do you mind to add a proper doc?

daemon/src/projection.rs
```diff
@@ -432,9 +497,28 @@ impl Cfd {
...
+                Ok((profit_btc, profit_percent)) => {
+                    (Some(profit_btc), Some(profit_percent.to_string()))
+                }
+                Err(err) => {
+                    tracing::error!(%payout, margin=%self.margin, "Profit calculation failed: {:#}", err);
+                    (None, None)
+                }
+            };
 
+            self.payout = Some(payout);
```
bonomat, line 484: I think we should be distinguishing between `realized` and `unrealized` profits and model it that way, e.g. call the fields `realized_profit` and `unrealized_profit`, `payout` and `predicted_payout` (or similar).

To decide whether to calculate `realized` or `unrealized` profits should depend on the state (a CFD is either `open` or `closed`). 
Right now the logic of how to calculate the profit (`realized` or `unrealized`) is spread into different places which I find hard to follow. 

We don't have to do this in this PR but I think the complexity here will bite us eventually.

daemon/src/projection.rs
bonomat: Can you explain this logic please: 

From the code comments I understand that : 

- If `self.cet` is present, the `cet` should have been published.
- If `self.timelocked_cet` is present, the `cet` has not been published

What if we don't know which `cet` will be published, both are `None` right? 

If this is the case, then this function returns `None` which leads to `self.profit_btc = profit_btc;` not being set (line 501+).  
Am I correct and there is a problem here?

bonomat (changes requested):
I'm not sure if there is a problem here. Please have a look.

daemon/src/projection.rs
da-kami: The current logic is, that we set the profit and payout based on the current price when calling `projection::Cfd::new` and then *overwrite* that profit and payout in case we already know the payout when calling `projection::Cfd::apply` based on an event containing a transaction spending from `lock` (transactions are collected in `projection::Aggregated`). 
I think this design could be improved - it's not easy to understand what is happening at the moment. But refactoring this was not my concern in this PR.

To answer the question:

Returning `None` means that we don't have a fixed payout yet, but keep the initially set payout & p/l based on the current price. This is expected behavior according to the current logic flow.

daemon/src/projection.rs
da-kami: I totally agree, but that would be a breaking API change. I tried to avoid that and keep what we have but make it more correct than in was before. We can go further and change the API as well, but then we will have to communicate that to users running the old version. 

If we accept breaking the API then we could even go further. We could distinguish more clearly between values that are fixed and values that are dynamic (e.g. based on current price). We could more clearly model "unrealized" (= open) and "realized" (= closed).

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
 }
 
-/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
 pub fn calculate_profit(
+    payout: SignedAmount,
+    margin: SignedAmount,
+) -> Result<(SignedAmount, Percent)> {
+    let profit = payout - margin;
+    let percent = Decimal::from_f64(100. * profit.as_sat() as f64 / margin.as_sat() as f64)
+        .context("Unable to compute percent")?;
```
thomaseizinger: We are we doing all these calculations in float and not in Decimal?

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
+
+    Ok((profit, Percent(percent)))
+}
+
+/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
+pub fn calculate_profit_from_current_price(
```
thomaseizinger: Nit:

```suggestion
pub fn calculate_profit_at_price(
```

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
     initial_price: Price,
-    closing_price: Price,
+    current_price: Price,
```
thomaseizinger: Nit: I wouldn't be upset if we find a better name here. The function doesn't necessarily know, which price it will be called with. Perhaps `target_price` is better? (From it being the calculation _target_?)

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
+    let profit = payout - margin;
+    let percent = Decimal::from_f64(100. * profit.as_sat() as f64 / margin.as_sat() as f64)
+        .context("Unable to compute percent")?;
+
+    Ok((profit, Percent(percent)))
+}
+
+/// Returns the Profit/Loss (P/L) as Bitcoin. Losses are capped by the provided margin
+pub fn calculate_profit_from_current_price(
     initial_price: Price,
```
thomaseizinger: I think we call this `opening_price` in many other places.

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,31 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
...
     quantity: Usd,
     leverage: Leverage,
     position: Position,
-) -> Result<(SignedAmount, Percent)> {
+) -> Result<(SignedAmount, Percent, SignedAmount)> {
     let inv_initial_price =
         InversePrice::new(initial_price).context("cannot invert invalid price")?;
     let inv_closing_price =
-        InversePrice::new(closing_price).context("cannot invert invalid price")?;
+        InversePrice::new(current_price).context("cannot invert invalid price")?;
```
thomaseizinger: Nit: Variable names not aligned.

daemon/src/projection.rs
thomaseizinger: > If we accept breaking the API then we could even go further. We could distinguish more clearly between values that are fixed and values that are dynamic (e.g. based on current price). We could more clearly model "unrealized" (= open) and "realized" (= closed).

If you don't remove the old field but just add two new ones, it is not a breaking change :)

daemon/src/projection.rs
```diff
@@ -432,9 +497,28 @@ impl Cfd {
 
         // If we don't have a dedicated closing price, keep the one that is set (which is
         // based on current price).
-        if let Some(closing_price) = self.aggregated.closing_price {
-            let (profit_btc, profit_percent) = self.compute_profit(closing_price);
+        // TODO: This clone can proably be optimized away
```
thomaseizinger: `Option::as_ref()` is likely your friend here.

thomaseizinger (commented):
I think the commits should be squashed or more clearly separated.

daemon/src/projection.rs
da-kami: After some more thought I am actually not sure the current UI has an advantage in a split API. We would have to introduce decision logic for "is realized already set otherwise use unrealized". 
I still think we should change the API, but I don't think it should be done as part of this PR. We could potentially also use two separate UI components for open and closed positions in the UI. I feel this needs a bit more thinking and should not be rushed as part of this work.

daemon/src/projection.rs
da-kami: OK, but how would that help me here? The condition is "if either of them is set", if I questionmark one for them I will never be able to check the other. Am I missing something?

thomaseizinger: Ah I didn't see the `|`. Nevermind then, you can't use `?` then, correct!

daemon/src/model/cfd.rs
da-kami: Good question... I think this is some more legacy code that I just blindly copy pasted :/ 
I'll change it.

daemon/src/projection.rs
thomaseizinger: What you can do is use `Option::or`: https://doc.rust-lang.org/std/option/enum.Option.html#method.or

```rust
let cet = self.cet.or(self.timelocked_cet)?;
```

daemon/src/projection.rs
```diff
@@ -208,18 +220,76 @@ pub struct Cfd {
...
+        if let Some((tx, script)) = self.collab_settlement_tx {
+            return Some(extract_payout_amount(tx, script));
+        }
+
+        match (self.cet, self.timelocked_cet) {
+            (Some(tx), _) | (_, Some(tx)) => {
+                let dlc = self
+                    .latest_dlc
+                    .as_ref()
+                    .expect("dlc to be present when we have a cet");
```
thomaseizinger: This can use a `?` to bail if the DLC is not set.

daemon/src/projection.rs
da-kami: Unfortunately it's not that simple, because in `apply` I don't know the role. We don't store the role in the `Cfd`. I could add it to the `Cfd` and avoid serialization, but then it's probably better to just keep it in the `Aggregate`.

daemon/src/model/cfd.rs
da-kami: this is not part of the changeset of this PR and it is called `initial_price` in `projection::Cfd` and `model::Cfd` so I will not change that here.

daemon/src/projection.rs
da-kami: Only if I change the return type to be a `Result`. I think it is fair to assume that the `Dlc` is always set here.

daemon/src/model/cfd.rs
da-kami: Not sure what you mean by this comment.

daemon/src/model/cfd.rs
```diff
@@ -1234,20 +1234,30 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
     price * leverage / (leverage + 1)
 }
 
+pub fn calculate_profit(payout: SignedAmount, margin: SignedAmount) -> (SignedAmount, Percent) {
+    let profit = payout - margin;
+
+    let profit_sats = Decimal::from(profit.as_sat());
+    let margin_sats = Decimal::from(margin.as_sat());
+    let percent = dec!(100) * profit_sats / margin_sats;
```
da-kami, line 1242: @thomaseizinger is it OK to use the primitive impls over `checked_mul` and `checked_div` here?
Otherwise we would need to return a `Result` again.

thomaseizinger: They will panic on overflow / underflow in debug mode and overflow / underflow in release builds. If you are okay with that, then you can use them.

daemon/src/projection.rs
```diff
@@ -211,12 +228,51 @@ struct Aggregated {
...
+    fn payout(self, role: Role) -> Option<Amount> {
+        if let Some((tx, script)) = self.collab_settlement_tx {
+            return Some(extract_payout_amount(tx, script));
+        }
+
+        let tx = self.cet.or(self.timelocked_cet)?;
+        let dlc = self
+            .latest_dlc
+            .as_ref()
+            .expect("dlc to be present when we have a cet");
```
thomaseizinger, line 249: You could `?` this instead of panicking.

daemon/src/projection.rs
```diff
@@ -211,12 +228,51 @@ struct Aggregated {
...
+
+        Some(extract_payout_amount(tx, dlc.script_pubkey_for(role)))
+    }
+}
+
+/// Returns output if it can be found or zero amount
+///
+/// If we cannot find an output for our script we assume that we were liquidated.
+fn extract_payout_amount(tx: Transaction, script: Script) -> Amount {
+    let payout = match tx
```
thomaseizinger: This `match` can be replaced with a `map` and an `unwrap_or_else`.

daemon/src/projection.rs
```diff
@@ -211,12 +228,51 @@ struct Aggregated {
...
+            return Some(extract_payout_amount(tx, script));
+        }
+
+        let tx = self.cet.or(self.timelocked_cet)?;
+        let dlc = self
+            .latest_dlc
+            .as_ref()
+            .expect("dlc to be present when we have a cet");
+
+        Some(extract_payout_amount(tx, dlc.script_pubkey_for(role)))
```
thomaseizinger: Would be more readable to extract some variables here.

thomaseizinger (approved):

daemon/src/model/cfd.rs
thomaseizinger: Line 1258 uses `inv_closing_price` and line 1259 uses `current_price`.

daemon/src/projection.rs
da-kami: Questionmarking would give a wrong impression and swallow an error - it *is* an error if we don't have a DLC but get a CET. For now I think the `expect` is the clearest solution.

daemon/src/model/cfd.rs
da-kami: I think these calculations with `SignedAmount` should always fit into the `Decimal` type. I don't think we can effectively run into an under-/overflow here. I'll keep as is.

daemon/src/projection.rs
thomaseizinger: It is an option so the None will bubble up.

thomaseizinger (approved):
Thanks!

da-kami:
I noticed that `closing_price` and `opening_price` are likely still better names for `calculate_profit_at_price`. Changing `closing` to `target` is a pain.

daemon/src/projection.rs
bonomat: > I still think we should change the API, but I don't think it should be done as part of this PR.

this 👍

daemon/src/projection.rs
bonomat: I missed that. Got it, thanks for the explanation.

bonomat (approved):
LGTM

Issue #1079, opened 2022-01-12
sats / BTC switch in the UI


da-kami:
With small amounts the BTC display can be annoying. Umbrel offers a `sats / BTC` switch, so one can switch between displaying BTC and sats. I think we should have that as well. 

In order to make that happen I propose the following changes:

- change the projection API to always return `bitcoin::Amount` `as_sat`
- Add a `Amount` component to the UI that holds the satoshi amount and allows display in BTC and sat. The conversion is trivial multiplication that should not lead to any loss; we could even opt for using a decimal type so we can just shift the dot. Use this new component for all amount displays.
- Add a toggle switch to the user interface to allow switching `sats / BTC`

Note: We don't send any bitcoin related amounts to the daemon, so no changes there. If we ever send bitcoin related amounts in request we should send them in sats.

thomaseizinger:
> * Add a `Amount` component to the UI that holds the satoshi amount and allows display in BTC and sat. The conversion is trivial multiplication that should not lead to any loss; we could even opt for using a decimal type so we can just shift the dot. Use this new component for all amount displays.
> 
> * Add a toggle switch to the user interface to allow switching `sats / BTC`

Note: If we want a global switch for this, we should utilize a `React.Context` to store the global state and pass it to all `Amount` components implicitly.

klochowicz:
we could go even further, and have Settings/Preferences that are stored in the backend (e.g. `Rocket::State` for the time being), and then just emit `String` based on the preference from Projection actor.

da-kami:
> we could go even further, and have Settings/Preferences that are stored in the backend (e.g. `Rocket::State` for the time being), and then just emit `String` based on the preference from Projection actor.

I would not enable different serialization because it will be hard to maintain. We would have to align all APIs to have both BTC and sats. This includes the feeds as well. I think the serialization should always be in `sat` - so the API should always hand out sats and the UI decides how to display. 

On top of that we can enable some settings API that allows the user to store his preferences. I see that as independent feature.

klochowicz:
> > we could go even further, and have Settings/Preferences that are stored in the backend (e.g. `Rocket::State` for the time being), and then just emit `String` based on the preference from Projection actor.
> 
> I would not enable different serialization because it will be hard to maintain. We would have to align all APIs to have both BTC and sats. This includes the feeds as well. I think the serialization should always be in `sat` - so the API should always hand out sats and the UI decides how to display.
> 
[...]

PR #1185, opened 2022-01-21
Accumulate funding fees stored in the CfdEvents

### LINKED ISSUE #1131: Store the funding fees in a Cfd event
we need to store the funding fees to be able to:
1) present aggregated fees to date to the user,
2) account for already paid fees for the rollover (the margin changes slightly due to them)

IMHO it would be most convenient to include them in  `ContractSetupCompleted` and `RolloverCompleted`, but as the events are immutable, that can't really be done? (unless we treat this as the "last" time we change them, as it's the first release including event sourcing?)
Creating a separate event to store them seems a bit excessive...

klochowicz:
- Split fees into OpeningFee and FundingFee
- Store FundingFee in CfdEvents,
- Store FundingRate inside FundingFee,
- accumulate funding fees by summing up fees stored in cfd events,
- ensure that funding fees presented in the maker are the inverse of taker's,

Fixes  #1131

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -6,6 +6,7 @@ use anyhow::Result;
 use bdk::bitcoin::Address;
 use bdk::bitcoin::Amount;
 use bdk::bitcoin::Denomination;
+use bdk::bitcoin::SignedAmount;
 use chrono::DateTime;
 use derive_more::Display;
 use reqwest::Url;
@@ -29,6 +30,8 @@ use std::time::UNIX_EPOCH;
 use time::OffsetDateTime;
 use time::PrimitiveDateTime;
 use time::Time;
+
+use self::cfd::Role;
 
 pub mod cfd;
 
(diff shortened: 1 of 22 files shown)


daemon/src/model.rs
```diff
@@ -635,32 +638,131 @@ impl Default for FundingRate {
     }
 }
 
-/// Fee paid on every contract renewal (a fraction of it if rolling over)
-#[derive(Debug, Clone)]
+/// Fee paid by the taker to the maker for opening a cfd.
+#[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
+pub struct OpeningFee {
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
```
da-kami: Is there a particular reason why you opted for serialization `as_btc`?
I think we should have a separate type for the projection. Internally I think we should always serialize `as_sat`.
We only use `as_btc` in the projection at the moment, which should also be changed to `as_sat` (see https://github.com/itchysats/itchysats/issues/1079)

Note that a lot of the structs defined in `model` should actually not live there, but should be moved into `projection`, but that is out of scope for this PR.

daemon/src/model.rs
```diff
@@ -635,32 +638,131 @@ impl Default for FundingRate {
...
+}
+
+impl From<OpeningFee> for FundingFee {
+    fn from(opening_fee: OpeningFee) -> Self {
+        // We calculate the fees for taker
+        Self(opening_fee.to_u64())
+    }
+}
+
+/// Fee paid on every contract renewal (a fraction of it if rolling over)
```
da-kami: I think this is incorrect. The `RolloverFee` is the accumulated previous funding fee (including opening fee as long as we do the bookkeeping as part of the payout calc)  + the current rollover fee.

If this is the only fee we pass in when rolling over the representation as fee + rate does not make sense.

daemon/src/setup_contract.rs
```diff
@@ -452,7 +451,7 @@ pub async fn roll_over(
             rollover_params.quantity,
             rollover_params.leverage,
             n_payouts,
-            rollover_params.funding_fee,
+            rollover_params.funding_fee.clone().into(),
```
da-kami: Here we have to pass in the accumulated fees - see my comment on the representation of `RolloverFee`.

da-kami (commented):
Great start, I think the `RolloverFee` will need to be changed, but overall it looks good.

daemon/src/model.rs
```diff
@@ -635,32 +638,131 @@ impl Default for FundingRate {
...
+            Role::Maker => -amount_sat,
+            Role::Taker => amount_sat,
+        };
+        SignedAmount::from_sat(amount_sat)
+    }
+}
+
+impl fmt::Display for OpeningFee {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.amount.fmt(f)
```
thomaseizinger: Did you test what this does? I think that will `Display` to something like `0 BTC`!

I think we should only store the satoshi value in the DB so you either need to adopt this implementation or do a manual implementation of what the `impl_sqlx_type_display_from_str` macro does.

daemon/src/model/cfd.rs
```diff
@@ -721,7 +735,7 @@ impl Cfd {
         // whether they match
         let hours_to_charge = 1;
 
-        let funding_fee = FundingFee::new(self.margin(), self.funding_rate, hours_to_charge)?;
+        let funding_fee = RolloverFee::new(self.margin(), self.funding_rate, hours_to_charge)?;
```
thomaseizinger: I think this would be simpler if we would calculate `FundingFee` here (which is always for 1 hour) and then simply multiply it by `hours_to_charge`.

thomaseizinger: That should also allow you to get rid of the entire `RolloverFee` concept which - as far as I understand - only exists because we might need to charge for more than one hour.

daemon/src/setup_contract.rs
```diff
@@ -678,25 +677,28 @@ pub async fn roll_over(
...
+            commit: (commit_tx, msg1.commit, commit_desc),
+            cets,
+            refund: (refund_tx, msg1.refund),
+            maker_lock_amount,
+            taker_lock_amount,
+            revoked_commit,
+            settlement_event_id: announcement.id,
+            refund_timelock: rollover_params.refund_timelock,
+        },
+        rollover_params.funding_fee.clone(),
```
thomaseizinger: Don't understand why we need to pass this one out here. If we've passed in a value as a parameter, the caller already knows it so we don't need to return it again!

thomaseizinger (commented):
Thanks for iterating on this. Some comments!

daemon/src/model.rs
klochowicz: I'll store it as satoshis then?

daemon/src/setup_contract.rs
klochowicz: How else would you like me to store it into the event? Perhaps I don't see something obvious here?

daemon/src/setup_contract.rs
klochowicz: We don't do this in this PR. I really do not wish to expand this PR scope even further.

daemon/src/model.rs
klochowicz: The different open ticket tracks these changes - this one is laying the groundwork for storing the fees.

daemon/src/model.rs
klochowicz: We don't have tests with anything but 0s here, so the tests would probably not catch it :)

daemon/src/model.rs
klochowicz: I opted for this for consistency, `Amount` everywhere is serialized this way - however, as @thomaseizinger pointed out, might need to change to sat anyway.

daemon/src/model/cfd.rs
klochowicz: > That should also allow you to get rid of the entire `RolloverFee` concept which - as far as I understand - only exists because we might need to charge for more than one hour.

The `RolloverFee` also stores the funding rate, which needs to to be presented in the projection ultimately.

daemon/src/setup_contract.rs
thomaseizinger: `rollover_params` is a parameter to this function. Meaning the caller has access to this value already because they pass it in. Wherever you use this return value, you should also have access to the parameters that you used so you can pass it into the event there.

daemon/src/model.rs
klochowicz: > Great start, I think the `RolloverFee` will need to be changed, but overall it looks good.

Does it need to be changed in this PR or can be done as a follow-up?

daemon/src/setup_contract.rs
klochowicz: `rollover_params` has a tight scope and all the fields private, would you  prefer me to break this abstraction and allow fishing out that field?

The way I see this is a smaller trade-off, but apart from a slight awkwardness of passing it back, it at least doesn't leak the abstraction. Happy to change it if you consider it not a worthy trade-off.

daemon/src/model.rs
klochowicz: as spoken offline, changed to a newtype storing `u64` satoshis.

daemon/src/setup_contract.rs
klochowicz: changed to allow `funding_fee()` accessor on the RolloverParams

daemon/src/model/cfd.rs
klochowicz: I removed one of the "Fee" structs, hopefully it's cleaner now. Now `FundingFee` is the main thing, it can optionally be initialised with `OpeningFee` on top of that, and it can tally up the fees (implements `Add` trait)

daemon/src/setup_contract.rs
klochowicz: I'll pass it here after all now,  but we might need to sit together to tie the calculations down :)

daemon/src/model.rs
klochowicz: it's just `u64` wrapper now, as per other comments - problem went away :)

klochowicz:
@da-kami @thomaseizinger I tried to address the feedback above the best I can, in a way that was coherent with what we want to do with the funding rates! hopefully it's closer to be merged now :)
I force-pushed, as it really warrants a new review....

Notable changes:
- storing initial funding rate in the DB - that's probably we didn't see anything after trying to set it up in the maker 😓 
- `FundingFee` (formerly `RolloverFee`) can tally up all the fees that happened, making it easier to pass into rollover/setup params
- no awkward `FundingFee` struct that was a common denominator between `OpeningFee` and `RolloverFee` (now you can build `FundingFee` with `.with_opening_fee()` instead).

da-kami (approved):
Let's get it in :)

thomaseizinger (approved):

PR #1187, opened 2022-01-21
Use patched version of xtra

thomaseizinger:
This version includes a patch that logs long running message handlers.
That should be useful in detecting deadlocks in our software.


da-kami (approved):
Let's get it in and learn more about what is going on in the actor system.

PR #1197, opened 2022-01-24
Maker validates settlement amounts according to payout curve

da-kami:
I noticed that this is missing. The maker has to validate the amounts sent by the taker.

Note: At the moment we don't report a failure of settlement back to the taker, but just finish with `Failed` I am not sure that is correct. The assumption is that we don't fail the collaborative settlement. I think this changes *should* not lead to failures as long as the taker acts sane. Tested locally without problems.


thomaseizinger (approved):

bonomat (approved):
👍

PR #1198, opened 2022-01-24
Downgrade block-fixup-merge action to 1.1.2

thomaseizinger:
That version supports to be run on `push` events, which means we can add
it to our bors configuration.

Something seems to be not working with bors' batch merges and GitHub's
required checks.


.github/workflows/ci.yml
```diff
@@ -26,6 +26,14 @@ jobs:
       - uses: Swatinem/rust-cache@v1.3.0
       - run: cargo clippy --workspace --all-targets -- -D warnings
 
+  block-fixup:
```
klochowicz, line 29: why in the `CI`? it's more of a formal git check?

klochowicz (approved):

.github/workflows/ci.yml
thomaseizinger: It should be triggered by exactly the same events as the other checks so it made sense to me to put it here.

More generally though, I see it as a integration check, same as our clippy, test etc. If this check doesn't pass, you are not allowed to integrate with master. That is pretty much the definition of CI :)

PR #1205, opened 2022-01-24
Minor connection message fixes

da-kami:


bonomat (approved):
👍

klochowicz (approved):
much clearer! :)

da-kami:
users will only see it in the next version, but still. Better fix it when we see it :)

PR #1209, opened 2022-01-24
Display `Setup Failed` and `Rejected` in the same way

da-kami:
This makes it easier to "ignore" in the closed positions section.

image

image


klochowicz (approved):
looks nicer indeed! :)

da-kami:
For now making it easy to ignore should do. For a more long term solution I think the "closed positions" section will have to change altogether. But once we tackle accumulated positions I expect the interface to more radically change so I don't think it's worth investing a lot of time into this. I just keep posting small fixes when I spot something that's easy to change :)

PR #1211, opened 2022-01-25
Avoid concurrent connections to database

### LINKED ISSUE #1168: Attempt to retry saving of event in case database is locked
We seem to be sometimes hitting the following error:

> error returned from database: database table is locked: EVENTS: database table is locked: EVENTS

https://github.com/itchysats/itchysats/runs/4876080983?check_suite_focus=true#step:6:266

It should be fairly straight-forward to retry saving the event in case of that error.

thomaseizinger:
By limiting the pool to a size of 1, we can avoid opening concurrent
connections to the database at all so from SQlite's perspective, there
is always just one connection.

Fixes #1168.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -14,6 +14,7 @@ use futures::future::BoxFuture;
 use futures::FutureExt;
 use sqlx::migrate::MigrateError;
 use sqlx::pool::PoolConnection;
+use sqlx::pool::PoolOptions;
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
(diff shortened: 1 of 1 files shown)


klochowicz:
do you expect any performance as a result of this? do the tests run the same time? (BTW: it would be amazing if we could benchmark how long the tests run in debug/release modes easily....)

thomaseizinger:
> do you expect any performance as a result of this? do the tests run the same time? (BTW: it would be amazing if we could benchmark how long the tests run in debug/release modes easily....)

I don't expect any performance problems as a result of this. SQlite usually has the entire DB in-memory so it is pretty fast. If we wanted to squeeze performance out of this, I'd go down the route of allowing to plug in a postgres DB or something like that.

klochowicz (approved):

Issue #1125, opened 2022-01-18
Charge funding fee for "missed" rollovers


klochowicz:
Currently, the funding rate will be charged per hour at every next rollover. However, a taker could technically go offline for extended period of time to "miss out" on few rollovers and not be charged for them, as the next rollover they participate in will extend the period for another 24h. 
At the rollover period we should account for the last rollover time to check how

Note: on umbrel, potential for such exploit is somewhat limited, as the taker stay online continuously. A maker can also easily spot such behaviour, refusing to accept a delayed rollover or even blacklist takers exploiting this behaviour.

PR #1217, opened 2022-01-25
Fund fund fund

### LINKED ISSUE #1164: Margin and Cost relationship in the Taker UI
This is a follow-up of #1066 

Given that we cannot add to lock at the moment our margin is not constant but changes with every funding event (i.e. we never add "fresh" money to keep the margin constant or reduce it).

The relationship of margin and fees is currently not clear enough. I will see to display the margin as current margin (i.e. initial margin - fees) and add a tooltip that shows initial margin and costs. That way we also don't show the costs in the tile itself.

This includes keeping track of the initial margin in the internal model and exposing it in the projection. It should be trivial to keep the initial margin that was used in the contract setup internally.

### LINKED ISSUE #1165: Profit/Loss & payout should take fees into account
Currently our profit/loss and payout display does not take fees into account in the UI.

Storing fees in the internal `Cfd` model is tracked in:  #1131 
This ticket should be tackled after or together with #1131 

The fee actually has to make sure to know it's sign internally and on the projection API, because fees for one party is a positive interest collection for the other. 

We have to keep track of the fees internally and use them in the profit calculation. The tests should be adapted to reflect the loss / gain in terms of fees for the parties. 
Once the projection calc for profit takes fees into account we should align the display in the UI and add a tooltip to the `payout` that makes it clear that `payout = initial_margin + profit - recurring_fees` (note: the opening fee is not part of this equation).

A note on calculating profit and loss based on what is outlined in #1164:
With reducing the margin on every rollover, our profit and loss will take recurring fees into account automatically. The payout curve will define the payout (taking all fees into account; recurring + opening). The profit and loss amount is then `p/l = margin + opening_fee - payout`.

### LINKED ISSUE #1175: Payout curve should get all the accumulated funding fees
One thing I noticed that is not directly related to this PR - but since we are adding the accumulated fees here it might be relevant:

Currently we pass the funding fee for *one* interval into rollover:

https://github.com/itchysats/itchysats/blob/28f7fa1836c12716a0595b862574204669dffbe8/daemon/src/model/cfd.rs#L724-L737

Is this correct in the context of accumulated fees - should it be `current_funding_fee + accumulated_previous_funding_fees` instead? 
How is the fee bucket managed - does the payout curve expect the accumulated fees, or just the fee for the last interval?

_Originally posted by @da-kami in https://github.com/itchysats/itchysats/pull/1170#pullrequestreview-858902780_

da-kami:
resolves https://github.com/itchysats/itchysats/issues/1165
resolves https://github.com/itchysats/itchysats/issues/1164
resolves https://github.com/itchysats/itchysats/issues/1175

Needs some more testing and rebase


daemon/src/model/cfd.rs
```diff
@@ -721,8 +721,8 @@ impl Cfd {
         // TODO: Taker should take this from the maker, optionally calculate and verify
         // whether they match
         let hours_to_charge = 1;
-
-        let funding_fee = FundingFee::new(self.margin(), funding_rate, hours_to_charge)?;
+        let funding_fee =
+            calculate_funding_fee(self.counterparty_margin(), funding_rate, hours_to_charge)?;
```
da-kami: At the moment we always use the the short margin (at a leverage of 1) to apply the funding fee. Does that make sense for now?

daemon/src/model.rs
```diff
@@ -647,43 +647,41 @@ impl str::FromStr for FundingRate {
...
-        }
+    pub fn new(fee: Amount) -> Self {
+        Self { fee }
     }
 
-    /// Value in satoshis for the taker
-    pub fn to_u64(&self) -> u64 {
-        self.amount_sat
-    }
+    fn into_signed_amount(self, role: Role) -> SignedAmount {
```
thomaseizinger: I think we can simplify the code in settle etc if we don't have this indirection.

I'd be okay with just accessing the private fields of the other newtypes directly. We are in the same module so we can bundle all the invariants together.

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+
+        (absolute, flow)
+    }
+
+    pub fn balance(&self) -> SignedAmount {
+        self.balance
+    }
+
+    pub fn add_opening_fee(self, opening_fee: OpeningFee) -> Self {
+        let sum = self.balance + opening_fee.into_signed_amount(self.role);
```
thomaseizinger: This would be easier to understand if it depicted the logic of: opening fee always flows from maker to taker.

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+        Self {
+            balance: sum,
+            position: self.position,
+            role: self.role,
+        }
+    }
+
+    pub fn add_funding_fee(self, funding_fee: FundingFee) -> Self {
+        let sum = self.balance + funding_fee.into_signed_amount(self.position);
```
thomaseizinger: I think the logic should be inlined here, potentially with some early returns.

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+        };
+
+        SignedAmount::from_sat(signed)
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum FeeFlow {
+    LongPaysShort,
+    ShortPaysLong,
```
thomaseizinger: Should these hold an Amount to make it clear that the amount is irrelevant for the `Nein` variant?

daemon/src/projection.rs
```diff
@@ -212,6 +214,9 @@ pub struct Cfd {
 
     #[serde(skip)]
     aggregated: Aggregated,
+
+    #[serde(skip)]
+    fee_account: FeeAccount,
```
thomaseizinger: Should this be part of Aggregated maybe?

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+        Self { fee: Amount::ZERO }
     }
 }
 
-impl_sqlx_type_display_from_str!(OpeningFee);
+impl From<OpeningFee> for Amount {
+    fn from(opening_fee: OpeningFee) -> Self {
+        opening_fee.fee
+    }
+}
```
thomaseizinger: Why are we exposing this? Not ideal IMO.

thomaseizinger (commented):
Some more ideas!

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+
+        SignedAmount::from_sat(signed)
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum FeeFlow {
+    LongPaysShort,
+    ShortPaysLong,
+    Nein,
```
bonomat, line 718: 😅

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+        } else {
+            -fee
+        };
+
+        SignedAmount::from_sat(signed)
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum FeeFlow {
```
bonomat, line 715: I wonder if we could model this differently: Instead of having `LongPaysShort` and vise-verca we have
`FeeFlow::Incoming` and `FeeFlow::Outgoing`

When you initial `FeeAccount` in `Cfd::new` we already know if it the fees will be positive (incoming) or negative (outgoing). 
Maybe we don't have to deal with a `SignedAmount` and back and forth conversion neither.

The reason why I'm proposing is is because you use `FeeFlow` on the `FeeAccount` and not on the `FundingRate`. 
If we have something like `LongPaysShort` I would assume that this is valid globally, but the way how it is modeled right now is that the `FeeFlow` is only valid for the `FeeAccount` for a single position. 

What do you think?

daemon/src/model/cfd.rs
bonomat: I think this is fair for now even though it is not inline with how funding fees are calculated on Bitmex (and probably others) 
 
We should think about how we can document/tell this to the user though.

daemon/src/model/cfd.rs
```diff
@@ -731,7 +728,8 @@ impl Cfd {
                 self.quantity,
                 self.leverage,
                 self.refund_timelock_in_blocks(),
-                tx_fee_rate,
+                tx_fee_rate, // TODO: Where should I get the fee rate from?
```
bonomat: Maybe best is if the maker can provide this when accepting rollover?

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+    ShortPaysLong,
+    Nein,
+}
+
+/// Our own accumulated fees
+///
+/// The counterparty fee is the inverse of this.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct FeeAccount {
+    balance: SignedAmount,
```
bonomat, line 728: So, `balance` is positive means you are owed this amount or you owe this amount to the counterparty?

daemon/src/model/cfd.rs
```diff
@@ -1958,20 +1997,32 @@ mod tests {
...
+
+        let taker_long = FeeAccount::new(Position::Long, Role::Taker)
+            .add_opening_fee(opening_fee)
+            .add_funding_fee(funding_fee);
+
+        let maker_short = FeeAccount::new(Position::Short, Role::Maker)
+            .add_opening_fee(opening_fee)
+            .add_funding_fee(funding_fee);
+
+        println!("{:?}, {:?}", maker_short, taker_long);
```
bonomat: You can probably remove this

bonomat (commented):
Thanks for tackling this complex topic. I've added some minor comments and a thought on a slightly different approach. Let me know what you think

daemon/src/model.rs
da-kami: > I wonder if we could model this differently: Instead of having LongPaysShort and vise-verca we have
FeeFlow::Incoming and FeeFlow::Outgoing

Interesting and possible, but not sure it actually makes it easier to understand. `Incoming` and `Outgoing` are quite high level again which makes it harder to interpret than `LongPaysShort` and `ShortPaysLong`.

One thing that makes this hard to understand is that we reasoning about *fees*. I thought about going away from the notion of `fee` and come up with something that is easier understandable, e.g. interest. But in the end the problem is always: One party receives the other pays; and the value has to incorporate both. I am not sure negative and positive interest is easier to understand. 

I think in the end it should always be about the position - the fact that we have the opening fee that always goes from taker to maker could be seen as implementation detail. I would hope that this can actually be depicted as a separate output on `lock` at some point, so we don't have to care about it when dealing with the payout curve.

> Maybe we don't have to deal with a SignedAmount and back and forth conversion neither.

How would you achieve that? The accumulated fees are the sum of negative *or* positive fees. I don't see a possibility  to come up with the sum without being able to subtract, add and go into negative for the data type. We could theoretically create a datatype that turns an underflow into an amount "in the other direction" and add a sign to it. But I don't think that would make it easier to understand.

--- 

That said: What do you think @thomaseizinger and @klochowicz? 
I can change this, but like the current version slightly better.

daemon/src/model.rs
da-kami: Since the account is about `fees`, the balance being positive means you owe this amount to the other party.

da-kami:
One more note: 

I think my changes to `OpeningFee` are another breaking API change and we will have to adapt the automation when creating an offer.

daemon/src/model.rs
da-kami: I agree. I already had the feeling it would be better to inline after being done, but was not sure about it. Will inline and keep the `OpeningFee` and `FundingFee` slim.

daemon/src/model.rs
thomaseizinger: The primary driver of this design was to create an _objective_ view onto how the fees need to be paid. `Incoming` and `Outgoing` are relative to the user. The main user of this API is the payout curve which doesn't care about the parties but only about long and short. That is where we need to make the main calculation so we opted to make the interpretation of the value there as easy as possible.

With `Incoming` and `Outgoing`, we wouldn't even be able to reason about the amount in the payout curve without passing in more information like role & position at which point you are just re-writing what is in settle to figure out how much you need to deduct from long and add to short or vice versa.

daemon/src/model.rs
thomaseizinger: This is something we should document on the struct!

daemon/src/model.rs
thomaseizinger: Independent of that, the whole fee market in CFDs seems to talk about fees being paid from long to shorts or vice versa so this seems like a natural modelling of this problem?

bonomat: No worries, it's just an idea because I had a hard time following the logic of `LongPaysShort` vs `ShortPaysLong` and the meaning I personally attach to these terms: _long pays short_ is to be expected in a bull market while _short pays long_ is to be expected in a bear market. Having these enum variants here confuses me as the can be different to each position. 


regarding no `SignedAmount`: 
My thinking was that the value is always positive but it is either _incoming_ (i.e. you get paid) or _outgoing_ (i.e. you have to pay). Based on this you can tell if you have to add or subtract it. 

// edit // didn't see your reply Thomas.

da-kami: > regarding no SignedAmount:
> My thinking was that the value is always positive but it is either incoming (i.e. you get paid) or outgoing (i.e. you have to pay). Based on this you can tell if you have to add or subtract it.

`FundingFee` and `OpeningFee` are depicted as always positive.
But for the `FeeAccount` we have to come up with the *sum* of the fees. We can create a construct where we keep the sum always positive and add something that defines the "sign", but I am not sure it makes it easier. I think in the end the sum calculation would still require a signed amount, but we could store the balance as absolute and add the flow to the internal representation. Is that what you mean?
(note: I kept the balance as `SignedAmount` so i can use it directly in the profit calc, but yeah, it's a bit weird that we use it directly there and for the payout curve we use `settle`)

daemon/src/model/cfd.rs
da-kami: I think he already does, this is a rebase issue where I re-introduced a comment that should be gone :)

klochowicz:
> One more note:
> 
> I think my changes to `OpeningFee` are another breaking API change and we will have to adapt the automation when creating an offer.

if it's still optional we won't have too adapt automation, unless we want to use it :) I don't think the automation uses opening fees yet (they've landed quite recently, after recent adjustments there!)

daemon/src/model/cfd.rs
```diff
@@ -465,7 +465,10 @@ impl Cfd {
     ) -> Self {
         let long_initial_margin = calculate_long_margin(initial_price, quantity, leverage);
 
-        // TODO: Evaluate if we want to do that
+        // TODO: we need more information from the offer to decide if we should charge this here or
+        // not. We would have to at least specify the interval. TODO: This HAS to be aligend
+        // to how we hidrate the cfd in projection which makes me believe that we should move this
```
klochowicz: nit: spelling *aligned*, *hydrate*

daemon/src/model/cfd.rs
```diff
@@ -678,15 +681,8 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
-        // TODO: Adjust for total paid fees - do we still need to do that if we
-        // tally up the fees???
-        let margin_minus_total_fees = self.margin();
-
-        // TODO: Calculate the time from the last rollover, don't just assume
```
klochowicz, line 685: IMHO this TODO is pretty crucial - is it tracked by a separate issue already? in case it is, I guess it's fine to remove it.

daemon/src/model.rs
```diff
@@ -983,97 +998,130 @@ mod tests {
...
-            );
-        }
+        assert_eq!(long_taker, (Amount::from_sat(1000), FeeFlow::ShortPaysLong));
+        assert_eq!(
+            short_maker,
+            (Amount::from_sat(1000), FeeFlow::ShortPaysLong)
+        );
     }
+
+    // TODO: Write a proptest for the fee account
```
klochowicz: are you planning to deal with this now or as a follow-up? shall we create an issue for it?

daemon/src/model.rs
```diff
@@ -608,11 +608,11 @@ pub struct FundingRate(Decimal);
 impl FundingRate {
     pub fn new(rate: Decimal) -> Result<Self> {
         anyhow::ensure!(
```
klochowicz, line 610: nit: it would be somehow clearer with a single check:

```rust
anyhow::ensure!(rate.abs() <= Decimal::ONE, "Funding rate can't be higher than 100%");
```

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+            position,
+            role,
+            balance: SignedAmount::ZERO,
+        }
+    }
+
+    pub fn settle(&self) -> (Amount, FeeFlow) {
+        let flow = if self.balance == SignedAmount::ZERO {
+            FeeFlow::Nein
+        } else if (self.position == Position::Long && self.balance.is_positive())
```
klochowicz, line 748: something seems not quite right if you need to expose the similar conditions in multiple places - can you use the same one from FundingFee or have a common abstraction? This seems like an implementation detail that shouldn't necessary leak to here.

klochowicz (commented):
thanks for pushing this one further!

daemon/src/model.rs
thomaseizinger: > No worries, it's just an idea because I had a hard time following the logic of `LongPaysShort` vs `ShortPaysLong` and the meaning I personally attach to these terms: _long pays short_ is to be expected in a bull market while _short pays long_ is to be expected in a bear market. Having these enum variants here confuses me as the can be different to each position.

Correct, that is what is depicted here. In a bull market, settlement of the fee account will result in `LongPaysShort`. In a bear market, it will be `ShortPaysLong`.

If a CFD is kept open for long enough, the fees might swing: Going from `LongPaysShort` to `ShortPaysLong`. The opening fee always flows from maker to taker, so depending on the their position, either long or short starts with a credit or debit in fees.

What we've modeled here is the overall accounting of fees in a CFD. In general, that will follow the bull/bear market rule but because we have an opening fee as well (and maybe in the future a closing fee), we also need to take that into account.

> (note: I kept the balance as `SignedAmount` so i can use it directly in the profit calc, but yeah, it's a bit weird that we use it directly there and for the payout curve we use `settle`)

Tracking the balance as absolute is correct from what I understand. The difficult part is the reasoning about what the sign means. Basic arithmetic defines that adding a negative value is subtraction for example. Because of this, handling signed values can be quite hard to reason about. In the payout curve, we only deal with unsigned values because we are splitting a pot of money into two parts. The assumption was that is easier to understand to pass an unsigned amount to that and separately specify, how the fees are meant to be handled (short to long or long to short).

IMO, we should also use `settle` in the P/L calc and simply also use our own position and role there to reason what that means for us.

daemon/src/model.rs
thomaseizinger: I would be in favor of moving all the decision logic into `FeeAccount` and have `OpeningFee` and `FundingFee` be relatively simple new types.

daemon/src/model.rs
da-kami: why?

daemon/src/model.rs
```diff
@@ -789,7 +764,16 @@ impl FeeAccount {
...
     pub fn add_opening_fee(self, opening_fee: OpeningFee) -> Self {
-        let sum = self.balance + opening_fee.into_signed_amount(self.role);
+        let fee: i64 = opening_fee.fee.as_sat().try_into().expect("not to overflow");
+
+        let signed_fee = match self.role {
+            Role::Maker => -fee,
+            Role::Taker => fee,
+        };
+
+        let signed_fee = SignedAmount::from_sat(signed_fee);
```
thomaseizinger: To remove this boilerplate, you could add an extension trait for `Amount` that has an `invert` method and returns us a `SignedAmount`.

daemon/src/payout_curve.rs
```diff
@@ -109,9 +108,9 @@ fn calculate_payout_parameters(
 
             let long_amount = to_sats(long_amount_btc)?;
```
thomaseizinger, line 107: If you move this line further down, you can avoid converting between `Amount` and `u64`.

daemon/src/model.rs
thomaseizinger: Because it is a public API that allows you to unpack the newtype, thus circumventing any invariants that are attached to it. What is the point of using a newtype in the first place then? The newtype conveys what this value means, if you unpack it, that meaning is removed and you can do whatever you want with it, thus removing the ability of the compiler to assist you in implementing the logic correctly.

This shouldn't exist and we should instead call more high-level APIs for the usecase where this is needed.

daemon/src/model/cfd.rs
da-kami: If I'm not mistaken it is tracked in https://github.com/itchysats/itchysats/issues/1125

daemon/src/model.rs
```diff
@@ -647,43 +647,41 @@ impl str::FromStr for FundingRate {
...
-    fn from(amount: Amount) -> Self {
-        Self::new(amount)
+        SignedAmount::from_sat(signed)
     }
 }
 
 impl fmt::Display for OpeningFee {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.amount_sat.fmt(f)
+        self.fee.as_sat().fmt(f)
```
luckysori, line 674: ❓ So the intention is to display an absolute value? Just confirming.

daemon/src/model.rs
da-kami: yeah, the fact that we still need the logic in two places is a bit odd. This goes towards what @bonomat pointed out. We could actually already keep the `FeeFlow` internally in `FeeAccount` and only use settle. When adding we could still use a signed amount, but internally we could store it as absolute.

daemon/src/model.rs
da-kami: I think yes, because this is just the internal representation. For the projection we use a signed amount at the moment (for the complete fee).

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+
+/// Funding fee as defined by position and rate
+///
+/// Position and rate define if the funding rate is to be added or subtracted when summing up the
+/// fees.
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct FundingFee {
+    #[serde(with = "bdk::bitcoin::util::amount::serde::as_sat")]
+    fee: Amount,
+    rate: FundingRate,
```
luckysori, line 705: 🔧 I would prefer it if we just stored the sign (using an `enum` or even a `bool`) instead of the funding rate. Otherwise we're storing more information than what we need and this data type is a bit confusing (i.e. I needed to verify that the `rate` field is only used for the sign). Perhaps we can use the `FundingRate` type only in the constructor?

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+
+fn calculate_funding_fee(
+    margin: Amount,
+    funding_rate: FundingRate,
+    hours_to_charge: i64,
+) -> Result<FundingFee> {
+    let fraction_of_funding_period = if hours_to_charge as i64 == SETTLEMENT_INTERVAL.whole_hours()
+    {
+        Decimal::ONE
+    } else {
```
luckysori, line 818: ❓ Do we really need a special case for division of `x/x`?

daemon/src/payout_curve.rs
da-kami: not really, because `long_amount_btc` is actually an `f64`.
Or am I missing something else?

daemon/src/model.rs
```diff
@@ -692,17 +690,145 @@ impl str::FromStr for OpeningFee {
...
+            position: self.position,
+            role: self.role,
+        }
+    }
+}
+
+fn calculate_funding_fee(
+    margin: Amount,
+    funding_rate: FundingRate,
+    hours_to_charge: i64,
```
luckysori, line 813: ❓ Why is this an `i64`? Negative hours?

daemon/src/model.rs
da-kami: I removed the todos for now and added https://github.com/itchysats/itchysats/issues/1227

daemon/src/model/cfd.rs
```diff
@@ -463,14 +464,14 @@ impl Cfd {
         initial_tx_fee_rate: TxFeeRate,
     ) -> Self {
         let long_initial_margin = calculate_long_margin(initial_price, quantity, leverage);
-        // TODO: Use FundingFee::default() if we don't want to charge funding fees
-        // based on quantity for the first settlement interval.
-        let funding_fees = FundingFee::new(
+
+        // TODO: Evaluate if we want to do that
```
luckysori: ❓ Can you elaborate on this TODO?

daemon/src/model/cfd.rs
```diff
@@ -463,14 +464,14 @@ impl Cfd {
...
+        let initial_funding_fee = calculate_funding_fee(
             long_initial_margin,
             initial_funding_rate,
             SETTLEMENT_INTERVAL.whole_hours(),
         )
-        .expect("values stored in db to be sane");
+        .expect("values stored in db to be sane"); // TODO: This expect is a bit weird
```
luckysori: I've looked at `calculate_funding_fee` and I think the expect makes sense:

- Division by zero can't happen given that `SETTLEMENT_INTERVAL` is the dividend.
- I think overflow can't happen on the first division since we're dividing by a number >= 1.0.
- We should be able to represent the funding fee with a `u64`. Otherwise our `margin * funding_fee * proportion > 18446744073709551615`, which is a lot of bitcoin.

daemon/src/model/cfd.rs
luckysori: > At the moment we always use the the short margin 

Is that true? The result of `counterparty_margin` changes depending on who is calling this. It would seem that each party would be calculating a different funding fee here.

da-kami:
I added two more tests that show a rountrip so one can see complete examples. It helps my understanding.

luckysori (commented):
I started reviewing by commit and then I realised that I probably wasn't meant to do so, so hopefully some of my comments didn't just get resolved later on!

daemon/src/model.rs
da-kami: `OpeningFee` is the internal representation. In `projection::CfdOrder` we pass it out as `Amount` - which is not incorrect I think. The projection "flattens" this, it is a Bitcoin amount after all. 

Instead of implementing `From` should I then just add something like `into_amount()` on `OpeningFee`?

daemon/src/model/cfd.rs
da-kami: > > At the moment we always use the the short margin
> 
> Is that true? The result of `counterparty_margin` changes depending on who is calling this. It would seem that each party would be calculating a different funding fee here.

It actually does *not* change. We keep an internal book of fees. The margins should be stable - unless I'm missing something. 
At some point we said that the margin would change, but I think that was a misunderstanding. Differently said: The payout curve calculates the payout based on the given parameters, but the given margin will always be the same, while the bookkeeping of the fees changes and that in turn changes the payout.

daemon/src/model.rs
```diff
@@ -647,43 +643,30 @@ impl str::FromStr for FundingRate {
...
-        let amount_sat: i64 = self.amount_sat.try_into().expect("to convert");
-        let amount_sat = match role {
-            Role::Maker => -amount_sat,
-            Role::Taker => amount_sat,
-        };
-        SignedAmount::from_sat(amount_sat)
-    }
+/// Opening fee is always payed from taker to maker
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct OpeningFee {
```
thomaseizinger, line 657: We might want to serialize this using [`serde(transparent)`](https://serde.rs/container-attrs.html#transparent) to avoid unnecessary nesting.

daemon/src/model.rs
klochowicz: this is to avoid lots of expects and potential panics everywhere - it will go away when we introduce a newtype for `Hours`. `Duration::whole_hours()` actually returns hours in `i64` format :/ (so you can encode time in the past)

daemon/src/model.rs
da-kami: We agreed on `to_inner` that consumes `self`

daemon/src/model/cfd.rs
klochowicz: yes, that's exactly why I added an `expect()` in the first place.

daemon/src/model.rs
thomaseizinger: @da-kami @klochowicz Thoughts?

I could be convinced that a bool or enum is better.

Looking at https://doc.rust-lang.org/std/primitive.i64.html#method.is_positive and https://doc.rust-lang.org/std/primitive.i64.html#method.is_negative, using the sign might be problematic because it returns `false` in both cases for `0`, meaning we should actually forbid constructing a `FundingFee` with a `FundingRate` of `0` or at least special case this in the constructor to handle it correctly.

Can we check with a test on what happens if we use a `FundingRate` of `0`?

Regarding your concern @luckysori, one we figured it might be useful for the future to record the actual funding rates that were used to calculate the fees.

daemon/src/model/cfd.rs
luckysori: All fair points.

In any case, I was wrong. I thought we were calling `calculate_funding_fee` with the result of `counterparty_margin` as an argument, for both the taker and the maker, which would be wrong. After checking out the PR I can see that it is not the case, so all is well.

daemon/src/model.rs
klochowicz: I guess the confusing part is that these `Display` traits in the model are for the sqlx macro to be able to store it as a string (and `FromStr` to parse it back) - it took me a while to realise this!

daemon/src/model/cfd.rs
klochowicz: indeed, it is! I forgot I've actually opened that issue! 🤦

daemon/src/model.rs
thomaseizinger: How is storing it as absolute better?

A fee account is a subjective view (in contains a role and a position). Storing the signed amount is consistent with that. `settle` provides a different, objective view onto this data. Storing already that view would be weird.

daemon/src/model.rs
klochowicz: I've actually have a half-finished rollout of that, will see to post it later today, as it will clear a few things.

daemon/src/model.rs
thomaseizinger: > I guess the confusing part is that these `Display` traits in the model are for the sqlx macro to be able to store it as a string (and `FromStr` to parse it back) - it took me a while to realise this!

Yeah it wasn't intended to be used for numbers really ...

daemon/src/payout_curve.rs
thomaseizinger: Ah fair point. Can we change `to_sats` to return `Amount` instead?

daemon/src/model.rs
luckysori: > Regarding your concern [...] we figured it might be useful for the future to record the actual funding rates that were used to calculate the fees.

That sounds fair enough. With that in mind I am not against keeping the field, even though we're not using it that way yet.

da-kami: The `FundingRate` is directly related to the `FundingFee`. I think we should store this information together, even if we don't "use" it at the moment.

daemon/src/model.rs
da-kami: @klochowicz do you remember why this was implemented like this?

daemon/src/model/cfd.rs
thomaseizinger: > > At the moment we always use the the short margin
> 
> Is that true? The result of `counterparty_margin` changes depending on who is calling this. It would seem that each party would be calculating a different funding fee here.

Very good note @luckysori ! cc @da-kami @bonomat 

Can we have a unit test for this that ensures we compute the same funding rate on both sides?

daemon/src/model/cfd.rs
da-kami: I removed the todo and the logic into a separate function that is used in the projection and the model to make it more clear that this is duplicated code.

daemon/src/model/cfd.rs
da-kami: you mean on `model::Cfd`? 
The logic of that is in `accept_rollover_proposal` which uses `self.margin()` and `handle_rollover_accepted_taker` which uses `self.counterparty_margin`. 
This is not totally correct, because this assumes that the maker is always short and the taker always long!

In order to make this clearer (and more correct) I added `model::Cfd::short_margin()` that returns the `short_margin` based on the Cfd's position. That should make it clear and easier to read. 

We should still add a unit test... but adding tests for the `Cfd` logic needs some more time because we have to build up the events. OK to tackle that in a follow up?

daemon/src/model.rs
da-kami: given that the opening fee is not part of the json events, but lives in the `Cfd` that makes sense.

daemon/src/model.rs
klochowicz: we actually had that last week!

daemon/src/model/cfd.rs
thomaseizinger: > We should still add a unit test... but adding tests for the `Cfd` logic needs some more time because we have to build up the events. OK to tackle that in a follow up?

I was wondering if we can do this in a test that solely operates on `FeeAccount` etc.

daemon/src/model.rs
da-kami: Happy to see this in a follow-up

daemon/src/model.rs
klochowicz: not sure, maybe a remnant of different impl, or for readability. you can change it if you wish, and the tests still pass :)

daemon/src/model.rs
```diff
@@ -692,17 +680,155 @@ impl str::FromStr for OpeningFee {
...
+        let sum = self.balance + signed_fee;
+
+        Self {
+            balance: sum,
+            position: self.position,
+            role: self.role,
+        }
+    }
+
+    pub fn add_funding_fee(self, funding_fee: FundingFee) -> Self {
```
thomaseizinger, line 783: Nit: I think this is a good place to add some domain knowledge about funding fees!

thomaseizinger (approved):

PR #1225, opened 2022-01-27
Move wallet info into table

bonomat:
Part of some ui-changes I was working on the plane: 

I think it's nicer if the wallet balance is part of the table and not at the top.

image

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -22,7 +22,7 @@ import Footer from "./components/Footer";
 import History from "./components/History";
 import Nav from "./components/NavBar";
 import Trade from "./components/Trade";
-import { Wallet, WalletInfoBar } from "./components/Wallet";
+import { Wallet } from "./components/Wallet";
 import { BXBTData, Cfd, ConnectionStatus, intoCfd, intoOrder, isClosed, Order, WalletInfo } from "./types";
 import { useEventSource } from "./useEventSource";
 import useLatestEvent from "./useLatestEvent";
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
Looks nice!

klochowicz:
I like it being placed in the table 👍  if anything, I'd only move it either to the top or bottom of it, it's a bit awkward to the user to have it wedged in between the margin and fees (it's unrelated to that, balance matters whether you have total fees summed up / displayed (and then seeing whether you can afford it)

thomaseizinger:
> I like it being placed in the table +1 if anything, I'd only move it either to the top or bottom of it, it's a bit awkward to the user to have it wedged in between the margin and fees (it's unrelated to that, balance matters whether you have total fees summed up / displayed (and then seeing whether you can afford it)

Agree to that, maybe do:

- Required margin
- Opening fee
- Total opening cost
- Available balance
- Remaining balance

Then it looks like you are doing a calculation:

- margin + opening fee = total opening cost
- available balance - total opening cost = remaining balance

bonomat:
I think there is already too much information going on here, so I have a counter proposal: 

image

The opening fee is part of the popup already, I'd just moved things around a bit: 

image

klochowicz (approved):
love the counterproposal! 

f people would like to have more discoverability, we could always someday add a tooltip that breaks down the required margin. However, as said above, we already break it down before the confirmation, so it's pretty much covered!

thomaseizinger (approved):
Much clearer now :)

PR #1232, opened 2022-01-27
Make `monitor::Actor` adhere to ADR013

### LINKED ISSUE #1149: Monitor actor does not follow ADR013 for blockchain events
[ADR013](https://github.com/itchysats/itchysats/blob/master/docs/adr/013-all-actors-can-hydrate-cfd.md) states that a CFD should be loaded and hydrated wherever needed.

The monitor actor does not follow this as it sends blockchain events to the `maker_cfd` / `taker_cfd` actor. Instead, it should use the `command::Executor` to call the corresponding command (`lock_tx_confirmed` etc) on the corresponding CFD.

The difficulty with making this happen is that our current test framework relies on sending the events directly to the `maker_cfd` and `taker_cfd` actor. This will have to be changed to interact with the monitor mock instead.

thomaseizinger:
No rush in merging this.

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -216,8 +216,6 @@ impl Maker {
     }
 
     pub async fn publish_order(&mut self, new_order_params: maker_cfd::NewOrder) {
-        self.mocks.mock_monitor_oracle_attestation().await;
-
         self.system
             .cfd_actor
             .send(new_order_params)
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -166,7 +166,7 @@ impl Maker {
 
         let db = db::memory().await.unwrap();
 
-        let mut mocks = mocks::Mocks::default();
+        let mut mocks = mocks::Mocks::new();
         let (oracle, monitor, wallet, price_feed) = mocks::create_actors(&mocks);
 
         let mut tasks = Tasks::default();
@@ -260,7 +260,7 @@ impl Taker {
 
         let db = db::memory().await.unwrap();
 
-        let mut mocks = mocks::Mocks::default();
+        let mut mocks = mocks::Mocks::new();
         let (oracle, monitor, wallet, price_feed) = mocks::create_actors(&mocks);
 
         let mut tasks = Tasks::default();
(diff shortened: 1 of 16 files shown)


daemon/tests/happy_path.rs
```diff
@@ -51,6 +50,90 @@ macro_rules! wait_next_state {
...
+            .confirm_close_transaction($id)
+            .await;
+    };
+    (cet, $id:expr, $maker:expr, $taker:expr) => {
+        $maker.mocks.monitor().await.confirm_cet($id).await;
+        $taker.mocks.monitor().await.confirm_cet($id).await;
+    };
+}
+
+macro_rules! expiry {
```
klochowicz: nit: `expire`? (to align with macro above? 

Both could also use some simple doc comments, ie. "trigger expiry event in the monitor actor" or something like that

klochowicz (approved):
great refactor, thank you! 👍

daemon/tests/harness/mocks/monitor.rs
```diff
@@ -40,6 +42,78 @@ impl MonitorActor {
...
+impl MockMonitor {
+    pub fn set_executor(&mut self, executor: command::Executor) {
+        self.executor = Some(executor);
+    }
+
+    pub async fn confirm_lock_transaction(&mut self, id: OrderId) {
+        self.executor
+            .as_ref()
+            .expect("executor to be set during test setup")
+            .execute(id, |cfd| Ok(cfd.handle_lock_confirmed()))
```
da-kami: I find it a bit weird that we have to mock all of these. I wonder if we could come up with a design where we can use production code to achieve this, i.e. arrange the production code in a way that we trigger the executor, but don't use the state management in the production monitor.

da-kami (approved):

daemon/tests/harness/mocks/monitor.rs
thomaseizinger: Very interesting idea. We would have to make parts of the original monitor actor swappable.

The internal logic is already split into the IO component, updating the state sending out events. We would have to plug the mock in before the last bit.

Interesting to explore but also, not that important I feel? Esp. keeping monitor::Event private is something I'd like to keep.

rishflab (approved):

da-kami (approved):

thomaseizinger:
@da-kami Is this an actual re-approval where you went through the code again or just a drive-by one? :P

It is quite the code change so I'd appreciate if people have a closer look!

PR #1233, opened 2022-01-27
Handle negative rate in fee calculation

luckysori:
Since the `FundingRate` is constructed from a signed decimal type, we can have a negative `FundingRate`. With that in mind and since we are modelling the `FundingFee` as an absolute value, we must use the absolute value of the `FundingRate` when computing the `FundingFee`'s `fee` field.

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -821,11 +821,13 @@ pub fn calculate_funding_fee(
             .context("can't establish a fraction")?
     };
 
-    let funding_fee =
-        (Decimal::from(margin.as_sat()) * funding_rate.to_decimal() * fraction_of_funding_period)
-            .round_dp_with_strategy(0, rust_decimal::RoundingStrategy::AwayFromZero)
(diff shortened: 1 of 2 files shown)


daemon/src/model.rs
```diff
@@ -821,11 +821,12 @@ pub fn calculate_funding_fee(
...
     };
 
-    let funding_fee =
-        (Decimal::from(margin.as_sat()) * funding_rate.to_decimal() * fraction_of_funding_period)
-            .round_dp_with_strategy(0, rust_decimal::RoundingStrategy::AwayFromZero)
-            .to_u64()
-            .context("Failed to represent as u64")?;
+    let funding_fee = (Decimal::from(margin.as_sat())
+        * funding_rate.to_decimal().abs()
+        * fraction_of_funding_period)
```
thomaseizinger: Can we extract this part into a variable to avoid the `()`?

thomaseizinger (approved):

PR #1234, opened 2022-01-27
Don't expose internal model on the maker API

da-kami:
We don't have tests for the maker's API to create new offers.
We should not expose internal types that can change, but treat the HTTP API as a separate interface, otherwise we might accidentally break the API.

I really think we should stop exposing our internal model to the outside and keep proper API boundaries. 
Some of these changes might be controversion, feel free to come up with other ideas if you have better ones.

### DIFF
--- a/daemon/src/routes_maker.rs
+++ b/daemon/src/routes_maker.rs
@@ -1,3 +1,4 @@
+use crate::Amount;
 use anyhow::Result;
 use bdk::bitcoin::Network;
 use daemon::auth::Authenticated;
@@ -25,6 +26,7 @@ use rocket::response::stream::EventStream;
 use rocket::response::Responder;
 use rocket::serde::json::Json;
 use rocket::State;
+use rust_decimal::Decimal;
 use rust_embed::RustEmbed;
 use serde::Deserialize;
 use std::borrow::Cow;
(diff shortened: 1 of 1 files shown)


daemon/src/routes_maker.rs
```diff
@@ -112,14 +111,29 @@ pub async fn post_sell_order(
     maker: &State<Maker>,
     _auth: Authenticated,
 ) -> Result<(), HttpApiProblem> {
+    let funding_rate = match order.funding_rate {
+        None => None,
+        Some(funding_rate) => Some(FundingRate::new(funding_rate).map_err(|e| {
+            HttpApiProblem::new(StatusCode::BAD_REQUEST)
+                .title("Invalid funding rate")
+                .detail(format!("{e:#}"))
+        })?),
```
thomaseizinger: Can we do the expression with the `?` on a new line please? This took a while to parse in my head :D

thomaseizinger (approved):
Makes sense to me, we should check that the constructors for our newtypes enforce a certain precision (if we want that).

da-kami:
Blocked until the new version is out because it is a breaking API change for the automation. Trivial to change but let's wait nonetheless.

luckysori (approved):

thomaseizinger:
> Blocked until the new version is out because it is a breaking API change for the automation.

What part of this is a breaking change?

da-kami:
Closing this because the API changed.

PR #1237, opened 2022-01-27
Margin for funding rate instead of always short margin

da-kami:
If the funding rate is positive long pays short, i.e. long borrows from short meaning that the rate should be applied to the short margin, because that is what long borrows.
If the funding rate is negative short pays long, i.e. short borrows from long meaning that the rate should be applied to the long margin, because that is what short borrows.

I think this makes more sense than always using the short margin to apply the rate. 
It is more in line with what we discussed here: https://github.com/itchysats/itchysats/discussions/1167#discussioncomment-2013291

### DIFF
--- a/daemon/src/model.rs
+++ b/daemon/src/model.rs
@@ -1,4 +1,6 @@
 use crate::impl_sqlx_type_display_from_str;
+use crate::model::cfd::calculate_long_margin;
+use crate::model::cfd::calculate_short_margin;
 use crate::olivia;
 use crate::SETTLEMENT_INTERVAL;
 use anyhow::Context;
@@ -618,6 +620,10 @@ impl FundingRate {
     pub fn to_decimal(&self) -> Decimal {
         self.0
     }
+
+    pub fn short_pays_long(&self) -> bool {
+        self.0.is_sign_negative()
+    }
 }
 
 impl Default for FundingRate {
@@ -808,10 +814,22 @@ impl FeeAccount {
 }
 
 pub fn calculate_funding_fee(
-    margin: Amount,
+    price: Price,
+    quantity: Usd,
+    leverage: Leverage,
     funding_rate: FundingRate,
     hours_to_charge: i64,
 ) -> Result<FundingFee> {
+    if funding_rate.0.is_zero() {
+        return Ok(FundingFee::new(Amount::ZERO, funding_rate));
+    }
+
+    let margin = if funding_rate.short_pays_long() {
+        calculate_long_margin(price, quantity, leverage)
+    } else {
+        calculate_short_margin(price, quantity)
+    };
+
     let fraction_of_funding_period = if hours_to_charge as i64 == SETTLEMENT_INTERVAL.whole_hours()
     {
         Decimal::ONE
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -463,8 +463,14 @@ impl Cfd {
         initial_funding_rate: FundingRate,
         initial_tx_fee_rate: TxFeeRate,
     ) -> Self {
-        let initial_funding_fee =
-            initial_funding_fee(initial_price, quantity, leverage, initial_funding_rate);
+        let initial_funding_fee = calculate_funding_fee(
+            initial_price,
+            quantity,
+            leverage,
+            initial_funding_rate,
+            SETTLEMENT_INTERVAL.whole_hours(),
+        )
+        .expect("values from db to be sane");
 
         Cfd {
             version: 0,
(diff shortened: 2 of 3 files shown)


daemon/src/model/cfd.rs
```diff
@@ -2272,6 +2290,52 @@ mod tests {
...
+        let quantity = Usd::new(dec!(200));
+        let leverage = Leverage::new(2).unwrap();
+
+        let funding_rate = dec!(-1);
+
+        let cfd = Cfd::dummy_not_open_yet()
+            .with_position(Position::Long)
+            .with_margin(price, quantity, leverage);
+
+        let margin = cfd.margin_for_funding_rate(FundingRate::new(funding_rate).unwrap());
```
thomaseizinger: Could extract creation of `FundingRate` to reduce duplication.

daemon/src/model/cfd.rs
```diff
@@ -2272,6 +2290,52 @@ mod tests {
...
+        assert_eq!(margin, Amount::from_btc(0.01).unwrap());
```
thomaseizinger: I'd call unwrap on the operation (`margin_for_funding_rate`) and not in the assertion.

daemon/src/model/cfd.rs
thomaseizinger: Also, you could chain `margin_for_funding_rate` onto the `cfd` variable to reduce some noise in here.

daemon/src/model/cfd.rs
```diff
@@ -2272,6 +2290,52 @@ mod tests {
...
+
+        let cfd = Cfd::dummy_not_open_yet()
+            .with_position(Position::Short)
+            .with_margin(price, quantity, leverage);
+
+        let margin = cfd.margin_for_funding_rate(FundingRate::new(funding_rate).unwrap());
+        assert_eq!(margin, Amount::from_btc(0.01).unwrap());
+    }
```
thomaseizinger: Arrange - act- assert :)

```suggestion
    #[test]
    fn given_negative_funding_rate_then_long_margin() {
        let price = Price::new(dec!(10000)).unwrap();
        let quantity = Usd::new(dec!(200));
        let leverage = Leverage::new(2).unwrap();
        let funding_rate = FundingRate::new(dec!(-1)).unwrap();

        let margin_on_long_cfd = Cfd::dummy_not_open_yet()
            .with_position(Position::Long)
            .with_margin(price, quantity, leverage)
            .margin_for_funding_rate(funding_rate)
            .unwrap();
        let margin_on_short_cfd = Cfd::dummy_not_open_yet()
            .with_position(Position::Short)
            .with_margin(price, quantity, leverage)
            .margin_for_funding_rate(funding_rate)
            .unwrap();
            
        assert_eq!(margin_on_long_cfd, Amount::from_btc(0.01).unwrap());
        assert_eq!(margin_on_short_cfd, Amount::from_btc(0.01).unwrap());
    }
```

daemon/src/model.rs
```diff
@@ -618,6 +618,10 @@ impl FundingRate {
     pub fn to_decimal(&self) -> Decimal {
         self.0
     }
+
+    pub fn short_pays_long(&self) -> bool {
+        self.0.is_sign_negative()
```
thomaseizinger, line 625: This will be `false` on 0, not sure what the implication is!

thomaseizinger (approved):

daemon/src/model.rs
da-kami: I am not sure what to do with the zero case - for the zero case the fee should be zero. It is irrelevant what the fee flow then is. I did not depict it here. If you want I can also use the fee-flow instead of just returning `bool`?

daemon/src/model/cfd.rs
da-kami: I added a fixup - not exactly what you propose, but you will like it I think :)

daemon/src/model.rs
thomaseizinger: `FeeFlow` now holds an `Amount` so not ideal, I think we can leave it for now!

Perhaps add a test that covers a rate of 0?

thomaseizinger (approved):

bonomat (approved):
looks good

da-kami:
For the `initial_funding_rate` this is actually still different. We always use the long margin there... 
I will see to change that as well.

da-kami:
It's not optimal that we duplicate logic. I see `initial_funding_fee` as something that should go away soon.

da-kami:
@thomaseizinger @bonomat I came up with a better solution. The logic is the same, but in a different place.

luckysori (approved):

thomaseizinger (approved):

PR #1241, opened 2022-01-27
Restore proptest for incrementally collecting funding fees

klochowicz:
Check that we're rounding up when collecting the funding fees pro rata every
hour.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -2268,6 +2268,40 @@ mod tests {
         assert_eq!(funding_fee.rate, funding_rate);
     }
 
+    use proptest::prelude::*;
+    proptest! {
+        #[test]
+        fn rollover_funding_fee_collected_incrementally_should_not_be_smaller_than_collected_once_per_settlement_interval(quantity in 1u64..100_000u64) {
+            let funding_rate = FundingRate::new(dec!(0.01)).unwrap();
+            let price = Price::new(dec!(10_000)).unwrap();
+            let quantity = Usd::new(Decimal::from(quantity));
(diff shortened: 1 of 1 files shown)


daemon/src/model/cfd.rs
```diff
@@ -2272,6 +2272,31 @@ mod tests {
...
+            let fee_account = FeeAccount::new(Position::Long, Role::Taker);
+
+            let fee_account_whole_interval = fee_account.add_funding_fee(funding_fee_for_whole_interval);
+            let fee_account_one_hour = fee_account.add_funding_fee(funding_fee_for_one_hour);
+
+            let total_balance_when_collected_hourly = fee_account_one_hour.balance().checked_mul(SETTLEMENT_INTERVAL.whole_hours()).unwrap();
+            let total_balance_when_collected_for_whole_interval = fee_account_whole_interval.balance();
+
+            prop_assert!(
+                total_balance_when_collected_hourly >= total_balance_when_collected_for_whole_interval,
```
thomaseizinger, line 2295: Better to subtract values from each other and assert that the diff is smaller than some small number.

thomaseizinger (changes requested):

thomaseizinger (approved):

daemon/src/model/cfd.rs
klochowicz: I derived some value by tried and error (if we'd like to make it tighter, we can always adjust the implementation). 30 sat seems good for now.

klochowicz:
bors retry

thomaseizinger:
> bors retry

Don't think a clippy error will be solved be retrying :)

PR #1246, opened 2022-01-28
Enable more lints across the workspace

### LINKED ISSUE #951: Enable more lints
These lints look interesting to enable as `warn`:

- https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-qualifications
- https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-import-braces
- https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies

The latter one might cause problems with us having a library crate because some dependencies are only used in the binary or library. One example that comes to my mind is `clap`. Thus, as part of enabling this, we probably need to split our workspace into:

- `maker`
- `taker`
- _some-common-library_

To enable the lints across all crates in our workspace, take some inspiration from this PR: https://github.com/EmbarkStudios/rust-ecosystem/pull/68.

thomaseizinger:


rishflab (approved):

klochowicz:
lol, OCD lint :) looks good!

klochowicz (approved):

daemon-tests/src/mocks/monitor.rs
```diff
@@ -37,7 +37,7 @@ impl MonitorActor {
 
     async fn handle(&mut self, _: monitor::StartMonitoring) {}
 
-    async fn handle(&mut self, _: monitor::CollaborativeSettlement) {}
+    async fn handle(&mut self, _: monitor::MonitorCollaborativeSettlement) {}
```
luckysori, line 40: 🙃 I like the fact that we're being consistent and also that we're no longer reusing the name of the module here to give the correct meaning to this message. But `monitor::Monitor` is still a bit of an eyesore. What about changing the verb to `Watch` for all these messages that are currently prepended with `Monitor`?

luckysori (approved):
😃

daemon-tests/src/mocks/monitor.rs
thomaseizinger: More than happy to make such a change but I'd like to get the lint in first and then iterate on such ideas so that they are applied consistently!

thomaseizinger: The whole public API of `monitor` is pretty garbage TBH.

klochowicz (approved):

PR #1252, opened 2022-01-28
Make things more open

da-kami:
Because the other states are too detailed.
We could change that in the daemon at some point.
Also shortened some state names to make sure they don't break the tile layout and are more in line with our terminology.

### DIFF
--- a/taker-frontend/src/types.ts
+++ b/taker-frontend/src/types.ts
@@ -104,7 +104,7 @@ export class State {
             case StateKey.REJECTED:
                 return "Rejected";
             case StateKey.PENDING_OPEN:
-                return "Pending Open";
+                return "Open";
             case StateKey.OPEN:
                 return "Open";
             case StateKey.PENDING_COMMIT:
@@ -114,11 +114,11 @@ export class State {
             case StateKey.INCOMING_SETTLEMENT_PROPOSAL:
                 return "Close Proposed";
             case StateKey.OUTGOING_SETTLEMENT_PROPOSAL:
-                return "Close Proposed";
+                return "Closing";
             case StateKey.INCOMING_ROLLOVER_PROPOSAL:
-                return "Rollover Proposed";
+                return "Open";
             case StateKey.OUTGOING_ROLLOVER_PROPOSAL:
-                return "Rollover Proposed";
+                return "Open";
             case StateKey.PENDING_REFUND:
                 return "Refunding";
             case StateKey.REFUNDED:
@@ -141,6 +141,9 @@ export class State {
 
         switch (this.key) {
             case StateKey.OPEN:
+            case StateKey.PENDING_OPEN:
+            case StateKey.INCOMING_ROLLOVER_PROPOSAL:
+            case StateKey.OUTGOING_ROLLOVER_PROPOSAL:
                 return green;
 
             case StateKey.REJECTED:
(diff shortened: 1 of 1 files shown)


luckysori (approved):

thomaseizinger (approved):

taker-frontend/src/types.ts
```diff
@@ -114,11 +114,11 @@ export class State {
...
                 return "Close Proposed";
             case StateKey.OUTGOING_SETTLEMENT_PROPOSAL:
-                return "Close Proposed";
+                return "Closing";
             case StateKey.INCOMING_ROLLOVER_PROPOSAL:
-                return "Rollover Proposed";
+                return "Open";
             case StateKey.OUTGOING_ROLLOVER_PROPOSAL:
-                return "Rollover Proposed";
+                return "Open";
```
klochowicz, line 121: I would prefer to say "renewing" or "funding", as a user I would like to see when things are being renewed. There's no need to hide that from the user, especially that it results in fees changing - otherwise it would be a surprise for no good reason.

klochowicz (changes requested):

taker-frontend/src/types.ts
bonomat: Why would you like to see it?

klochowicz: I'd like to see things "working", ie. action happening as it happens. Otherwise it's pretty hard to find positive confirmation, as we currently do not display closing time or anything that would show that we got rolled over. this could be better solved  with some "inbox" that describes actions that happened, but until then, IMHO it's best to keep at least the current state instead of voluntarily making in a bit worse.

bonomat: Mhmm... I'm not convinced 😬 For me most of the states here are a technical detail and it won't be clear to users what they mean. For example, if you show the user for 5 seconds where the actual rollover is happening that the position is `renewing` or `funding` and then right afterwards it goes back to `open`. What did he gain? I think he will get more confused by that.

luckysori: Yeah, I think the user doesn't care about the existence of `rollover`. We need to tell them when and what fees are charged, but they should not need to care about transactions being signed for a few seconds.

klochowicz: the gain is assisting the user with building an accurate mental model of how the software works (and, as written above, positive confirmation that *something* happened to it, the user is more confident that the software works - note that often GUIs are "faking" work just to please the user, e.g. download/copy progress bars ;D )

Asking from the other POV - what's the harm in showing renewing the contract / funding interval?

bonomat: > Asking from the other POV - what's the harm in showing renewing the contract / funding interval?

Potential confusion? And how likely is it that a user is looking at his open positions when a rollover happens?

But I like your your line of thinking of building up trust. Imho the before mentioned `inbox` is more suitable for that.

thomaseizinger: Such "inbox" could be married with the order history of a CFD assuming combined positions.

You would see things like:

- order placed / filled
- funding fee charged (that is a rollover happening)

klochowicz (approved):

taker-frontend/src/types.ts
thomaseizinger: Until then, I wouldn't be opposed to remove technical details for now until we have a better model.

Yes it is a step back but I'd rather display less and figure out what is missing than to try and retain the status quo for the sake of it!

klochowicz:
@da-kami did you by any chance run into #1260 ? Under normal operation it should only take a few seconds to transition back, did it get stuck on "Rollover proposed"?

PR #1254, opened 2022-01-28
Fix fee taker UI

### LINKED ISSUE #1242: Opening fee is not shown in `Confirm` dialog

da-kami:
Still needs some testing.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -26,6 +26,7 @@ use crate::xtra_ext::SendAsyncSafe;
 use crate::Order;
 use crate::Tasks;
 use crate::SETTLEMENT_INTERVAL;
+use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::Amount;
@@ -47,7 +48,6 @@ use std::time::Duration;
 use time::OffsetDateTime;
 use tokio::sync::watch;
 use xtra::prelude::MessageChannel;
-use xtra::Context;
 use xtra_productivity::xtra_productivity;
 
 /// Store the latest state of `T` for display purposes
@@ -670,7 +670,18 @@ impl Actor {
     }
 
     fn handle(&mut self, msg: Update<Option<Order>>) {
-        let _ = self.tx.order.send(msg.0.map(|x| x.into()));
+        let order = match msg.0 {
+            None => None,
+            Some(order) => match TryInto::<CfdOrder>::try_into(order) {
+                Ok(order) => Some(order),
+                Err(e) => {
+                    tracing::warn!("Unable to convert order: {e:#}");
+                    None
+                }
+            },
+        };
+
+        let _ = self.tx.order.send(order);
     }
 
     fn handle(&mut self, msg: Update<Option<bitmex_price_feed::Quote>>) {
@@ -686,7 +697,7 @@ impl Actor {
 
 #[async_trait]
 impl xtra::Actor for Actor {
-    async fn started(&mut self, ctx: &mut Context<Self>) {
+    async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("we just started");
 
         // this will make us load all cfds from the DB
@@ -775,6 +786,9 @@ pub struct CfdOrder {
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
     pub margin_per_parcel: Amount,
 
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    pub initial_funding_fee_per_parcel: Amount,
+
     pub leverage: Leverage,
     #[serde(with = "round_to_two_dp")]
     pub liquidation_price: Price,
@@ -783,11 +797,13 @@ pub struct CfdOrder {
     pub settlement_time_interval_in_secs: u64,
 }
 
-impl From<Order> for CfdOrder {
-    fn from(order: Order) -> Self {
+impl TryFrom<Order> for CfdOrder {
+    type Error = anyhow::Error;
+
+    fn try_from(order: Order) -> std::result::Result<Self, Self::Error> {
         let parcel_size = Usd::new(dec!(100)); // TODO: Have the maker tell us this.
 
-        Self {
+        Ok(Self {
             id: order.id,
             trading_pair: order.trading_pair,
             position: order.position,
@@ -810,12 +826,21 @@ impl From<Order> for CfdOrder {
                 .settlement_interval
                 .whole_seconds()
                 .try_into()
-                .expect("settlement_time_interval_hours is always positive number"),
+                .context("unable to convert settlement interval")?,
             opening_fee: Some(order.opening_fee.to_inner()),
-            funding_rate_annualized_percent: AnnualisedFundingRate::from(order.funding_rate)
+            funding_rate_annualized_percent: AnnualisedFundingPercent::from(order.funding_rate)
                 .to_string(),
-            funding_rate_hourly_percent: HourlyFundingRate::from(order.funding_rate).to_string(),
-        }
+            funding_rate_hourly_percent: HourlyFundingPercent::from(order.funding_rate).to_string(),
+            initial_funding_fee_per_parcel: calculate_funding_fee(
+                order.price,
+                parcel_size,
+                order.leverage,
+                order.funding_rate,
+                SETTLEMENT_INTERVAL.whole_hours(),
+            )
+            .context("unable to calcualte initial funding fee")?
+            .to_inner(),
+        })
     }
 }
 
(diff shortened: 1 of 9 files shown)


daemon/src/projection.rs
```diff
@@ -812,9 +815,18 @@ impl From<Order> for CfdOrder {
...
-            funding_rate_hourly_percent: HourlyFundingRate::from(order.funding_rate).to_string(),
+            funding_rate_hourly_percent: HourlyFundingPercent::from(order.funding_rate).to_string(),
+            initial_funding_fee_per_parcel: calculate_funding_fee(
+                order.price,
+                parcel_size,
+                order.leverage,
+                order.funding_rate,
+                SETTLEMENT_INTERVAL.whole_hours(),
+            )
+            .expect("to be able to calculate funding fee")
```
da-kami: Should probably use a `TryFrom` to avoid this expect.

daemon/tests/happy_path.rs
```diff
@@ -77,7 +77,7 @@ fn assert_eq_order(mut published: CfdOrder, received: CfdOrder) {
 
     // Hard-coded to match the dummy_new_order()
     assert_eq!(received.opening_fee.unwrap(), Amount::from_sat(2));
-    assert_eq!(received.funding_rate_hourly_percent, "0.001");
+    assert_eq!(received.funding_rate_hourly_percent, "0.00100");
```
da-kami, line 80: Note to self: Why do we need these two additional zeros here? We should not fix the percentage but fix it in the UI...

daemon/src/projection.rs
da-kami: I added https://github.com/itchysats/itchysats/pull/1254/commits/55b7b1dca664fa1cdb5725321caa354f2ded0667 to achieve this. Probably would be even better to report failure back to the caller. I think it *can* happen that this fails - if it fails, it will fail on the maker and taker side, so the order would be invalid on both sides.
The projection is the fist time we trigger this calculation for the per-parcel initial funding fee. Maybe a better strategy would be to actually add this to the `Order`  - or at least ensure it can be calculated when creating it. Happy for suggestions.

taker-frontend/src/App.tsx
```diff
@@ -53,8 +53,11 @@ export const App = () => {
     const parcelSize = parseOptionalNumber(order?.parcel_size) || 0;
     const liquidationPrice = parseOptionalNumber(order?.liquidation_price);
     const marginPerParcel = order?.margin_per_parcel || 0;
-    const openingFee = order?.opening_fee || 0;
+    const leverage = order?.leverage || 0;
+    let openingFee = order?.opening_fee || 0;
```
bonomat:
```suggestion
    const openingFee = order?.opening_fee || 0;
```

daemon/src/projection.rs
```diff
@@ -760,6 +760,9 @@ pub struct CfdOrder {
     /// The funding rate fluctuates with market movements.
     pub funding_rate_hourly_percent: String,
 
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
+    pub initial_funding_fee_per_parcel: Amount,
```
bonomat: Can we move this below `pub parcel_size: Usd,` please? I think this would make it easier to read the file top down

taker-frontend/src/components/Trade.tsx
```diff
@@ -109,11 +110,9 @@ export default function Trade({
 
     const margin = (quantity / parcelSize) * marginPerParcel;
 
-    const openingFee = (quantity / parcelSize) * openingFeePerParcel;
+    let feeForFirstSettlementInterval = (quantity / parcelSize) * fundingFeePerParcel;
```
bonomat:
```suggestion
    const feeForFirstSettlementInterval = (quantity / parcelSize) * fundingFeePerParcel;
```

taker-frontend/src/components/History.tsx
```diff
@@ -154,8 +154,12 @@ const CfdDetails = ({ cfd, connectedToMaker, displayCloseButton }: CfdDetailsPro
                 <Table size="sm" variant={"unstyled"}>
                     <Tbody>
                         <Tr>
-                            <Td><Text as={"b"}>Total fees</Text></Td>
-                            <Td textAlign="right"><BitcoinAmount btc={cfd.accumulated_fees} /></Td>
+                            <Td><Text as={"b"}>Estimated fees</Text></Td>
+                            <Td textAlign="right">
+                                <Tooltip label={"includes transaction fees"} placement={"right"}>
```
bonomat: 🚧 Not sure why but this tooltip seems to break everything. Removing it makes the side load again.

```
Uncaught Error: React.Children.only expected to receive a single React element child.
    onlyChild React
    Tooltip chakra-ui-tooltip.esm.js:290
    React 12
    unstable_runWithPriority scheduler.development.js:468
    React 5
    listener useLatestEvent.ts:13
    useLatestEvent useLatestEvent.ts:16
    React 5
    unstable_runWithPriority scheduler.development.js:468
    React 4
    unstable_runWithPriority scheduler.development.js:468
    React 6
    <anonymous> main.tsx:9
react.development.js:1139:12
```

If you move the tooltip around `Td` it works.

bonomat (dismissed):
One thing to fix please.

taker-frontend/src/components/History.tsx
da-kami: Oh, interesting. On my machine it works, but I generally had some trouble with chakra tooltips causing errors; so far it was not permanent errors though, i.e. the page loaded, but the logs showed an error when hovering over the tooltip. 

I'll move it around the `Td` if that helps.

daemon/src/projection.rs
```diff
@@ -810,12 +826,21 @@ impl From<Order> for CfdOrder {
                 .settlement_interval
                 .whole_seconds()
                 .try_into()
-                .expect("settlement_time_interval_hours is always positive number"),
+                .context("unable to convert settlement interval")?,
```
thomaseizinger, line 829: Previously, we solved this by making the specific field that failed an `Option` rendering it as a `Skeleton` in the UI.

daemon/src/projection.rs
```diff
@@ -810,12 +826,21 @@ impl From<Order> for CfdOrder {
...
-        }
+            funding_rate_hourly_percent: HourlyFundingPercent::from(order.funding_rate).to_string(),
+            initial_funding_fee_per_parcel: calculate_funding_fee(
+                order.price,
+                parcel_size,
+                order.leverage,
+                order.funding_rate,
+                SETTLEMENT_INTERVAL.whole_hours(),
+            )
+            .context("unable to calcualte initial funding fee")?
```
thomaseizinger, line 841: Do we have test cases that show when this can fail?

taker-frontend/src/App.tsx
```diff
@@ -53,8 +53,11 @@ export const App = () => {
     const parcelSize = parseOptionalNumber(order?.parcel_size) || 0;
     const liquidationPrice = parseOptionalNumber(order?.liquidation_price);
     const marginPerParcel = order?.margin_per_parcel || 0;
-    const openingFee = order?.opening_fee || 0;
+    const leverage = order?.leverage || 0;
```
thomaseizinger, line 56: If you default this to 0 now, you should change the `Props` type of `Trade` to no longer accept an optional `leverage` to get more type safety.

daemon/src/projection.rs
```diff
@@ -670,7 +670,18 @@ impl Actor {
     }
 
     fn handle(&mut self, msg: Update<Option<Order>>) {
-        let _ = self.tx.order.send(msg.0.map(|x| x.into()));
+        let order = match msg.0 {
+            None => None,
+            Some(order) => match TryInto::<CfdOrder>::try_into(order) {
+                Ok(order) => Some(order),
+                Err(e) => {
+                    tracing::warn!("Unable to convert order: {:#}", e);
```
thomaseizinger:
```suggestion
                    tracing::warn!("Unable to convert order: {e:#}");
```

taker-frontend/src/components/History.tsx
da-kami: If I wrap it around the `Td` I get this error (that is not fatal though):

```
react-dom.development.js:67 Warning: validateDOMNesting(...): <span> cannot appear as a child of <tr>.
    at span
    at DefaultPortal2 (http://localhost:3000/node_modules/.vite/@chakra-ui_react.js?v=8421fb55:1401:36)
    at Portal (http://localhost:3000/node_modules/.vite/@chakra-ui_react.js?v=8421fb55:1464:28)
    at PresenceChild (http://localhost:3000/node_modules/.vite/chunk-QRD5N5TB.js?v=8421fb55:6477:21)
    at AnimatePresence (http://localhost:3000/node_modules/.vite/chunk-QRD5N5TB.js?v=8421fb55:6556:21)
    at http://localhost:3000/node_modules/.vite/@chakra-ui_react.js?v=8421fb55:20820:17
...
```

It seems Chakra somehow has a `span` for the tooltip which is not allowed as sub-element in html tables.
When navigating through the dom-tree in the browser I was unable to find the span that causes the problem though. 

The chakra documentation states

> If the children of Tooltip is a string, we wrap with in a span with tabIndex set to 0, to ensure it meets the accessibility requirements.

https://chakra-ui.com/docs/overlay/tooltip#usage

But I don't think that is the case here. I think we should report this as bug upstream.

thomaseizinger (commented):
Some comments, haven't tested it.

Would be nice to have unit tests for `calculate_funding_fee` that show us, when exactly this can fail so we can make a better judgement on whether the complexity of handling the error is worthwhile.

daemon/src/projection.rs
da-kami: https://github.com/itchysats/itchysats/pull/1255 should cover this, but will need to be adapted to my changes.
I'd say this is not critical for the release and can be done later.

daemon/src/projection.rs
da-kami: I'd say this is not critical for the release and can be done later.

PR #1255, opened 2022-01-28
Add unit tests for `calculate_funding_fee`

luckysori:


daemon/src/model.rs
```diff
@@ -1275,4 +1276,47 @@ mod tests {
 
         assert_eq!(balance, SignedAmount::from_sat(1000))
     }
+
+    #[test]
+    fn proportional_funding_fees_if_sign_of_funding_rate_changes() {
+        let price = Price::new(Decimal::from_u32(35_000).unwrap()).unwrap();
```
thomaseizinger: Why not use the `dec!` macro?

daemon/src/model.rs
```diff
@@ -1275,4 +1276,47 @@ mod tests {
...
+        assert_eq!(
+            long_pays_short_fee.fee,
+            short_pays_long_fee.fee * long_leverage.get().into()
+        )
+    }
+
+    #[test]
+    fn zero_funding_fee_if_zero_funding_rate() {
+        let fee = calculate_funding_fee(
+            Price::new(Decimal::from_u32(35_000).unwrap()).unwrap(),
```
thomaseizinger: Perhaps use `dummy_price` if you don't care?

daemon/src/model.rs
```diff
@@ -1275,4 +1276,47 @@ mod tests {
...
+            Usd::new(Decimal::from_u32(100).unwrap()),
+            Leverage::new(2).unwrap(),
+            FundingRate::new(Decimal::ZERO).unwrap(),
```
thomaseizinger: Extract into variable and make it the "arrange" part to make it more obvious what the influential variable is in the test?

thomaseizinger (approved):
Some ideas!

daemon/src/model.rs
luckysori: I forgot its existence!

thomaseizinger (approved):

daemon/src/model.rs
```diff
@@ -1275,4 +1275,62 @@ mod tests {
...
+        let epsilon = (long_pays_short_fee.fee.as_sat() as i64)
+            - (short_pays_long_fee.fee.as_sat() as i64) * (long_leverage.get() as i64);
+        assert!(epsilon.abs() < 5)
+    }
+
+    #[test]
+    fn zero_funding_fee_if_zero_funding_rate() {
+        let zero_funding_rate = FundingRate::new(Decimal::ZERO).unwrap();
+
+        let dummy_leverage = Leverage::new(1).unwrap();
```
thomaseizinger, line 1312: We might want to introduce constants for this at some point.

luckysori: Do you mean a crate-wide `DUMMY_LEVERAGE` constant, for example?

thomaseizinger: Uhm, no I was thinking of just constants like `Leverage::ONE` etc

luckysori: Okay, that makes more sense. But the purpose would not be for tests specifically then, right?

PR #1262, opened 2022-01-28
Improve log output

bonomat:
Imho old log message was unreadable : 

```
2022-01-28 21:34:41  WARN daemon::connection: Disconnecting due to lack of heartbeat. Last heartbeat: Some(OffsetDateTime { utc_datetime: PrimitiveDateTime { date: Date { year: 2022, ordinal: 28 }, time: Time { hour: 21, minute: 34, second: 21, nanosecond: 367012000 } }, offset: UtcOffset { hours: 0, minutes: 0, seconds: 0 } })
```

New log says something like: 

```
2022-01-28 21:50:25  WARN daemon::connection: Disconnecting due to lack of heartbeat. Last heartbeat was 22 seconds ago. At 2022-01-28 21:50:02.99517 +00:00:00
```

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -116,9 +116,14 @@ impl State {
             return false;
         }
 
-        tracing::warn!(
-            "Disconnecting due to lack of heartbeat. Last heartbeat: {:?}",
-            self.last_heartbeat()
+        let heartbeat_timestamp = self
+            .last_heartbeat()
+            .map(|heartbeat| heartbeat.to_string())
+            .unwrap_or_else(|| "None".to_owned());
+        let seconds_since_heartbeat = duration_since_last_heartbeat.as_secs();
+        tracing::warn!(%seconds_since_heartbeat,
+            %heartbeat_timestamp,
+            "Disconnecting due to lack of heartbeat",
         );
 
         *self = State::Disconnected;


daemon/src/connection.rs
```diff
@@ -116,9 +116,13 @@ impl State {
             return false;
         }
 
+        let heart_beat = self
+            .last_heartbeat()
+            .map(|heart_beat| format!("{heart_beat}"))
```
thomaseizinger: Curious why clippy doesn't flag this as unnecessary format call. You can use to_string here.

daemon/src/connection.rs
```diff
@@ -116,9 +116,13 @@ impl State {
...
+            .unwrap_or_else(|| "None".to_string());
```
thomaseizinger: Nit: I usually use to_owned for converting from str to String because it captures the semantics better (you are only going from borrowed -> owned). ToString works based on Display which can be called on all sorts of types.

daemon/src/connection.rs
```diff
@@ -116,9 +116,13 @@ impl State {
...
+        let seconds_since_heart_beat = duration_since_last_heartbeat.as_secs();
         tracing::warn!(
-            "Disconnecting due to lack of heartbeat. Last heartbeat: {:?}",
-            self.last_heartbeat()
+            "Disconnecting due to lack of heartbeat. Last heartbeat was {seconds_since_heart_beat} seconds ago. At {heart_beat}",
```
thomaseizinger: Perhaps using tracing fields for the duration and the timestamp would be even better?

I find multi-sentence log messages to always be a bit of a code smell.

thomaseizinger (approved):

bonomat:
@thomaseizinger : incorporated your feedback. New log message looks like this: 

```
{"timestamp":"2022-01-29T22:58:47.867326Z","level":"WARN","fields":{"message":"Disconnecting due to lack of heartbeat","seconds_since_heartbeat":"21","heartbeat_timestamp":"2022-01-29 22:58:47.867062 +00:00:00"},"target":"daemon::connection"}
```

thomaseizinger (approved):

PR #1264, opened 2022-01-30
Set CET timelock to 6

bonomat:
This bugged me for a while. I think we can use a smaller number here. 

This timelock is a relative timelock in regards to the commit transaction expressed in blocks.
It is applied to CETs and prevents users from immediately publishing a CET transaction after the
commit transaction. This is needed for stopping users from publishing an outdated commit
transaction and "cheating" the other user. The longer this timelock is the more time the other
user has time to detect this malicious attempt and punish him.
The downside of a longer timelock is that users won't get their money fast. They have to wait
for this timelock to expire.

A timelock of 6 equals ~60 minutes which should be sufficient.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -54,7 +54,14 @@ use time::OffsetDateTime;
 use uuid::adapter::Hyphenated;
 use uuid::Uuid;
 
-pub const CET_TIMELOCK: u32 = 12;
+/// This timelock is a relative timelock in regards to the commit transaction expressed in blocks.
+/// It is applied to CETs and prevents users from immediately publishing a CET transaction after the
+/// commit transaction. This is needed for stopping users from publishing an outdated commit
+/// transaction and "cheating" the other user. The longer this timelock is the more time the other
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Interestingly, `dlcspecs` uses a CET timelock even without the presence of a channel. It is set to a time just after the expected oracle attestation time. Doing that would require an absolute timelock of course. Could/should we combine this shortened relative timelock (to enable punishment) and a separate absolute timelock (to protect against collusion with an oracle)?

luckysori (commented):
Another comment 😛

I'm not against this change, but I would personally err on the side of security with non-collaborative closure. Given how we've designed this protocol, closing using CETs should be rare because we encourage collaboration. If that is the case I think it would make sense to favour safety over convenience, meaning a _longer_ relative timelock would be better. 

With a longer relative timelock we also relax the "often-online" requirement.

bonomat:
convert to draft to indicate that this is not ready to merge

bonomat:
Closing this as it's a bit outdated

PR #1273, opened 2022-01-31
Add logs for reliably tracking established and gone connections

thomaseizinger:


daemon/src/maker_cfd.rs
```diff
@@ -140,51 +127,10 @@ impl<O, T, W> Actor<O, T, W> {
...
-where
-    T: xtra::Handler<maker_inc_connections::TakerMessage>,
-{
-    async fn handle_taker_connected(&mut self, taker_id: Identity) -> Result<()> {
-        self.takers
-            .send_async_safe(maker_inc_connections::TakerMessage {
-                taker_id,
-                msg: wire::MakerToTaker::CurrentOrder(self.current_order.clone()),
-            })
-            .await?;
```
thomaseizinger: Hmm, I forgot about this ...

bonomat (approved):
👍
Note: we need to update our metrics app

thomaseizinger:
> +1 Note: we need to update our metrics app

Yep. It might be good to merge the first commit as a non-breaking change then adapt the metrics, see how they behave and later remove the other code. I'll split the PR!

thomaseizinger:
This is a harmless change now, nominating to include in release so we can update our metrics asynchronously. cc @bonomat

PR #1274, opened 2022-01-31
Improve "No liquidity" alert from the maker

klochowicz:
Make it clear it's the *maker's* liquidity that has a problem, and fix the
grammar error by rephrasing.


bonomat (approved):
👍

thomaseizinger (approved):

PR #1276, opened 2022-01-31
Avoid log spam for wire messages

thomaseizinger:
In case an xtra message handler hangs, it logs the message that has
been processing. For messages containing types from within the `wire`
module, these can be huge. So big that processing the logs becomes
impossible and/or exceeds certain limits of downstream software.

We already have specific Display implementations for these types, use
those for the Debug representation.

### DIFF
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -58,7 +58,7 @@ impl fmt::Display for Version {
 pub mod taker_to_maker {
     use super::*;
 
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
I wonder whether we'll lose any important debug info for some messages due to that, but the pain of printing out the whole CET justifies this change. 👍

bonomat (approved):
👍

PR #1281, opened 2022-01-31
Bump xtra to latest HEAD

thomaseizinger:
The latest commit relaxes the `Debug` requirement and instead uses `std::any::type_name` to log the offending message.

This is a more systematic solution than https://github.com/itchysats/itchysats/pull/1276.


luckysori (approved):
Woo, less code!

klochowicz (approved):

Issue #1278, opened 2022-01-31
Make rocket's logs more useful


thomaseizinger:
They are noise in grafana due to the excessive use of colours and emojis:

![image](https://user-images.githubusercontent.com/5486389/151882016-2a66f1e6-711f-4e3b-bae6-922ec6b5290d.png)

On top of that, most of the logs emitted by rocket are quite useless IMO. We don't need to print all of the start-up configuration and one log per invoked HTTP handler also seems unnecessary.

PR #1287, opened 2022-02-01
Reduce built-in rocket logs

thomaseizinger:
Not sure if we want to put it into the release but the logs in grafana are unreadable with the CLI colors enabled. See https://github.com/itchysats/itchysats/issues/1278.


bonomat (approved):
I would get it in this release. More readable logs are definitely a plus:)

PR #1291, opened 2022-02-01
Cache expensive CFD events to avoid deserialising them from JSON

klochowicz:
Examining the maker with a flamegraph showed that we have been using ~50% CPU on
deserialising CFD events from JSON.

Using `cached` crate reduced that number significantly (to ~8%). However, the
cache size constraint would become a problem for long-running CFDs - to minimise
this issue, selective dispatch for expensive events was used.

ContractSetupCompleted and RolloverCompleted store the whole DLC, which takes a
lot of time to deserialise; these events were chosen to be cached.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -33,6 +33,7 @@ use bdk::bitcoin::Transaction;
 use bdk::bitcoin::Txid;
 use bdk::descriptor::Descriptor;
 use bdk::miniscript::DescriptorTrait;
+use cached::proc_macro::cached;
 use maia::finalize_spend_transaction;
 use maia::secp256k1_zkp;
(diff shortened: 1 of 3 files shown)


daemon/src/model/cfd.rs
```diff
@@ -56,6 +57,9 @@ use uuid::Uuid;
 
 pub const CET_TIMELOCK: u32 = 12;
 
+pub const CONTRACT_SETUP_COMPLETED_EVENT: &str = "ContractSetupCompleted";
+pub const ROLLOVER_COMPLETED_EVENT: &str = "RolloverCompleted";
```
thomaseizinger: Could be module private.

thomaseizinger (approved):

PR #1294, opened 2022-02-01
Apply timeout to all TCP operations 

thomaseizinger:


daemon/src/lib.rs
```diff
@@ -362,7 +362,7 @@ where
             + 'static,
         n_payouts: usize,
         maker_heartbeat_interval: Duration,
-        connect_timeout: Duration,
+        socket_timoeut: Duration,
```
luckysori: 🔧 `socket_timeout`.

daemon/src/maker_inc_connections.rs
```diff
@@ -457,28 +457,28 @@ impl Actor {
 ///
 /// Both IO operations, upgrading to noise and checking the version are gated by a timeout.
 async fn upgrade(
-    mut stream: TcpStream,
+    stream: TcpStream,
     noise_priv_key: x25519_dalek::StaticSecret,
     this: xtra::Address<Actor>,
 ) -> Result<()> {
+    let socket_timeout = Duration::from_secs(10);
```
luckysori, line 473: ❓ Any reason why this value has changed from 20 to 10 seconds?

luckysori (approved):

daemon/src/maker_inc_connections.rs
thomaseizinger: Copy-paste error :)

thomaseizinger:
So I am not fully sure if this is beneficial. In the debugging last week, I think we saw it a couple of times that a handler which wanted to write to a socket got blocked essentially forever for some reason. Applying a timeout should solve this but doing `.timeout` everywhere is tedious.

I'll leave this as a draft and keep testing it for a while to see how it behaves.

thomaseizinger:
Hmm, the tests are breaking because we are now imposing an overall 10s timeout. Probably not such a good idea :D

Issue #1288, opened 2022-02-01
Dissect `Dlc` struct according to new usage in event sourcing


thomaseizinger:
Previously, the `Dlc` struct needed to aggregate information internally to represent a newly setup but also rolled over CFD.

With event sourcing this is no longer the case.

- The lock transaction will remain the same across all rollovers so it shouldn't need to be stored within the data structure that is stored in `RolloverCompleted`.
Same thing for the refund transaction.
- The revoked commit transactions don't need to be kept within the `Dlc` struct but should instead be tracked as part of the `RolloverCompleted`.
- The stored CETs store extremely redundant data.
- We might be able to get rid of the `n_bits` field of `Cet`: https://github.com/itchysats/itchysats/issues/1301.

thomaseizinger:
As part of this, try to make as many fields of `Dlc` private as possible.

thomaseizinger:
Tagging this as performance because a big part of the deserialization problem might be the redundant storage of data.

We need to figure out, what the absolute minimal data is that we need to store and compute the rest on-demand.

bonomat:
This project is unmaintained

PR #1295, opened 2022-02-01
Do not store whole redundant CET in CfdEvents

luckysori:
Instead we recompute it when we need it.

We still have to keep the TXID around for monitoring purposes. We would like to be able to recompute it, but that would make applying `CfdEvent`s fallible, which is forbidden.

This should help with https://github.com/itchysats/itchysats/issues/1288.

It's only a draft PR because it's just an attempt at doing this quickly. I considered adding some of this functionality to `maia` so that the protocol knowledge remains in that crate (e.g. the order of the outputs in the CET), but this was simpler.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -30,6 +30,8 @@ use bdk::bitcoin::PublicKey;
 use bdk::bitcoin::Script;
 use bdk::bitcoin::SignedAmount;
 use bdk::bitcoin::Transaction;
+use bdk::bitcoin::TxIn;
+use bdk::bitcoin::TxOut;
 use bdk::bitcoin::Txid;
 use bdk::descriptor::Descriptor;
 use bdk::miniscript::DescriptorTrait;
@@ -1352,13 +1354,61 @@ pub fn calculate_profit_at_price(
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct Cet {
-    pub tx: Transaction,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub maker_amount: Amount,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub taker_amount: Amount,
     pub adaptor_sig: EcdsaAdaptorSignature,
 
     // TODO: Range + number of digits (usize) could be represented as Digits similar to what we do
     // in the protocol lib
     pub range: RangeInclusive<u64>,
     pub n_bits: usize,
+
+    pub txid: Txid,
+}
+
+impl Cet {
+    /// Build an actual `Transaction` out of the payout information
+    /// stored in `Self`, together with the input and the output
+    /// addresses.
+    ///
+    /// The order of the outputs matters.
+    ///
+    /// We verify that the TXID of the resulting transaction matches
+    /// the TXID with which `Self` was constructed.
+    pub fn to_tx(
+        &self,
+        (commit_tx, commit_descriptor): (&Transaction, &Descriptor<PublicKey>),
+        maker_address: &Address,
+        taker_address: &Address,
+    ) -> Result<Transaction> {
+        let tx = Transaction {
+            version: 2,
+            input: vec![TxIn {
+                previous_output: commit_tx.outpoint(&commit_descriptor.script_pubkey())?,
+                sequence: CET_TIMELOCK,
+                ..Default::default()
+            }],
+            lock_time: 0,
+            output: vec![
+                TxOut {
+                    value: self.maker_amount.as_sat(),
+                    script_pubkey: maker_address.script_pubkey(),
+                },
+                TxOut {
+                    value: self.taker_amount.as_sat(),
+                    script_pubkey: taker_address.script_pubkey(),
+                },
+            ],
+        };
+
+        if tx.txid() != self.txid {
+            bail!("Reconstructed wrong CET");
+        }
+
+        Ok(tx)
+    }
 }
 
 /// Contains all data we've assembled about the CFD through the setup protocol.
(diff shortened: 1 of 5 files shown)


thomaseizinger (commented):
Can we measure the improvement somehow?

luckysori:
> Can we measure the improvement somehow?

I created some temporary benchmarks for the deserialization:

```rust
use criterion::criterion_group;
use criterion::criterion_main;
use criterion::Criterion;

fn cet(c: &mut Criterion) {
    let cet = r#"
{
  "adaptor_sig": "02a6f941bd0664e8ba981d8fe9b65ebbc819eb30f598d51c81c27db5ad1b4b62850365b9a9958d9a4f872ef487d0a34981715dc2e14cfca2c86d66fa1dc345703f644b04f48bd4b166cac34d32af98170cbf020169004628dc4c6f87cf31669595bf0acd56131dc6aef714d0da28619ac972672c8773f383ec4fa76673e1a7990206808e8aa007aa83996e45f7b0dc24261f30dda0bc0a269bde809fae2137acd4bd",
  "n_bits": 17,
  "range": {
    "end": 67407,
    "start": 67400
  },
  "tx": {
    "input": [
      {
        "previous_output": "a2106b92b978d047ff2577a3423ca2ea15b56e6419c85878d8710dda296d7e7b:0",
        "script_sig": "",
        "sequence": 12,
        "witness": []
      }
    ],
    "lock_time": 0,
    "output": [
      {
        "script_pubkey": "0014d495f792f79dc99d4994ab5c91363e1ddc3239b0",
        "value": 147945
      },
      {
        "script_pubkey": "0014d7339690e745f7f84067546892a15c13b556a6f9",
        "value": 247867
      }
    ],
    "version": 2
  }
}"#;

    c.bench_function("old_cet_repr", |b| {
        b.iter(|| serde_json::from_str::<daemon::model::cfd::Cet>(cet).unwrap())
    });
}

criterion_group!(benches, cet);
criterion_main!(benches);

```

We see an almost 2x speed boost after trimming down the `Cet` struct:

```
old_cet_repr            time:   [2.3720 us 2.3859 us 2.4018 us]                          
new_cet_repr            time:   [1.2532 us 1.2546 us 1.2560 us]
```

I used [this online tool](https://www.debugbear.com/json-size-diff) to compare the size of the JSON stored in the db  and got these results:

image

luckysori:
There's at least one bug in the implementation, so I'll work on a fix tomorrow if we think we should use it.

thomaseizinger:
Nice benchmark! Thanks. Twice as fast sounds good, so lets get it in!

daemon/src/model/cfd.rs
```diff
@@ -1352,13 +1354,61 @@ pub fn calculate_profit_at_price(
...
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub maker_amount: Amount,
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
+    pub taker_amount: Amount,
     pub adaptor_sig: EcdsaAdaptorSignature,
 
     // TODO: Range + number of digits (usize) could be represented as Digits similar to what we do
     // in the protocol lib
     pub range: RangeInclusive<u64>,
     pub n_bits: usize,
```
thomaseizinger, line 1366: It is just one number but isn't this also duplicates across all CETs for one attestation?

luckysori: Actually, I was wrong when I thumbed up before. This number represents the number of  _relevant_ bits, which changes depending on the range being represented. It's used so that we can zip with the right nonces and attestations from the oracle.

Looking at it now, it seems to me like an optimisation (although I'm not sure I'm remembering everything). We should be able to multiply nonces/attestations by zero without affecting the total sum. In any case, I wouldn't tackle this now.

thomaseizinger: > Looking at it now, it seems to me like an optimisation (although I'm not sure I'm remembering everything). We should be able to multiply nonces/attestations by zero without affecting the total sum. In any case, I wouldn't tackle this now.

Okay, can you record this as an idea for later optimization?

luckysori: What I was trying to say is that we appear (not 100% sure) to have optimised for fewer operations (by removing multiplications by zero). But we might want to reconsider that choice in favour of storing less information. I'll create an issue.

daemon/src/model/cfd.rs
```diff
@@ -1352,13 +1354,61 @@ pub fn calculate_profit_at_price(
...
+    pub fn to_tx(
+        &self,
+        (commit_tx, commit_descriptor): (&Transaction, &Descriptor<PublicKey>),
+        maker_address: &Address,
+        taker_address: &Address,
+    ) -> Result<Transaction> {
+        let tx = Transaction {
+            version: 2,
+            input: vec![TxIn {
+                previous_output: commit_tx.outpoint(&commit_descriptor.script_pubkey())?,
```
thomaseizinger, line 1389: I am curious whether or not we can make this function infallible if we pass the outpoint in and make all fields private. Assuming you construct the struct correctly, the output of this function would be what you want. That could shave off the storage of the txid.

Let's explore this at some point.

daemon/src/model/cfd.rs
luckysori: See https://github.com/itchysats/itchysats/issues/1301.

thomaseizinger (approved):
Thanks!

This breaks the database so once we merge this, all old CFDs are not going to load.

daemon/src/model/cfd.rs
luckysori: I would very much like to not store the `Txid`, which is kind of redundant info. Since you seem to have a good idea, maybe record this somewhere :D

daemon/src/monitor.rs
```diff
@@ -722,7 +723,7 @@ impl MonitorParams {
         MonitorParams {
             lock: (dlc.lock.0.txid(), dlc.lock.1),
             commit: (dlc.commit.0.txid(), dlc.commit.2),
-            cets: map_cets(dlc.cets),
+            cets: map_cets(dlc.cets, &dlc.maker_address),
```
da-kami, line 726: Is it correct to always use the `maker` address here? 
I suppose it should not matter, but am not sure.

da-kami (approved):

daemon/src/monitor.rs
thomaseizinger: I think the thinking is that the maker's script will always be present whereas the taker's address might not be in the case of a liquidation. This deserves a comment though :)

luckysori: Nice catch, @da-kami. I think we should just solve this problem by looking at the payout and using the address of the first party with a non-zero payout.

PR #1299, opened 2022-02-01
Remove metric for tracking of connected takers

thomaseizinger:


thomaseizinger:
This should not be merged until our metrics are updated to use https://github.com/itchysats/itchysats/pull/1273.

daemon/src/maker_inc_connections.rs
```diff
@@ -102,6 +97,11 @@ pub struct Actor {
     settlement_actors: AddressMap<OrderId, collab_settlement_maker::Actor>,
     rollover_actors: AddressMap<OrderId, rollover_maker::Actor>,
     tasks: Tasks,
+
+    /// Tracks the last order that we sent out to our connected takers.
+    ///
+    /// Storing this here allows us to send it to newly connected takers.
+    last_order: Option<Order>,
```
luckysori: 🤔 This seems okay for now, but should we consider creating a little `order::Actor` which takes care of this state? It seems like the order doesn't belong in the "connection" actor.

luckysori (approved):

daemon/src/maker_inc_connections.rs
thomaseizinger: But then we have to send requests back and forth again which lead to a deadlock in the previous version. I am unsure!

luckysori: I get your point and there's no need to change this with this PR anyway. Some thoughts (from now on I will talk about offers, since I think that's the right term we agreed on):

If the maker's `offer::Actor` can only receive a new offer from the HTTP API and respond to a `GetOffer` request from the `maker_inc_connections::Actor`, is there potential for a deadlock?

bonomat (changes requested):
Please don't merge yet. Removing this will break some of the metrics we are recording, e.g. how many takers have connected and how many have opened a position.

thomaseizinger:
> Please don't merge yet. Removing this will break some of the metrics we are recording, e.g. how many takers have connected and how many have opened a position.

Yep! I think we will want to wait for #1331 actually before we remove this functionality.

daemon/src/maker_inc_connections.rs
thomaseizinger: I think there can be if there is a circular dependency in general.

luckysori: Do you see a request with a direct response (in the context of `xtra`, a `Message` with a `Result` other than `()`) as a circular dependency?

thomaseizinger: If both actors send messages to each other, that is a circular dependency.

thomaseizinger: Perhaps a dedicated `offer::Actor` would not create a circular dependency, I haven't fully thought it through :)

luckysori (approved):

bonomat (approved):
Note: this is a breaking API change and we need to adopt maker-metrics accordingly.

thomaseizinger:
> Note: this is a breaking API change and we need to adopt maker-metrics accordingly.

Yes. I am already testing this metric on testnet and it is working as expected there.

thomaseizinger:
> Note: this is a breaking API change and we need to adopt maker-metrics accordingly.

We can wait with merging this until after the next deployment and adopt the dashboard first!

bonomat:
> > Note: this is a breaking API change and we need to adopt maker-metrics accordingly.
> 
> We can wait with merging this until after the next deployment and adopt the dashboard first!

To be clear: I was talking about the maker-metrics binary we built on top of the daemon which uses this HTTP API.

thomaseizinger:
> > > Note: this is a breaking API change and we need to adopt maker-metrics accordingly.
> > 
> > 
> > We can wait with merging this until after the next deployment and adopt the dashboard first!
> 
> To be clear: I was talking about the maker-metrics binary we built on top of the daemon which uses this HTTP API.

Yes I know. We should adopt everything else first and only merge this once nothing else is using this API.

thomaseizinger:
It would be nice to remove this but `maker-metrics` has not been updated and it is not in the current focus, hence closing this.

PR #1307, opened 2022-02-02
Cleanup setup actors gone

da-kami:


da-kami:
This *can* happen for the taker as well, because if we restart the taker and we were in a contract setup then we would still have the problem of not having the actor anymore and have the CFD forever in the list. I will see to clean that up as well.

thomaseizinger:
> This _can_ happen for the taker as well, because if we restart the taker and we were in a contract setup then we would still have the problem of not having the actor anymore and have the CFD forever in the list. I will see to clean that up as well.

Wouldn't that clean-up require a housekeeping-like task that either runs on startup or continuously?

daemon/src/routes_maker.rs
```diff
@@ -154,6 +154,8 @@ pub async fn post_cfd_action(
     };
 
     result.map_err(|e| {
+        tracing::warn!(order_id=%id, %action, "Processing action failed: {e:#}");
```
thomaseizinger, line 157: I think we could do this on a more general level with an [`on_response`](https://docs.rs/rocket/0.5.0-rc.1/rocket/fairing/struct.AdHoc.html#method.on_response) rocket fairing:

daemon/src/setup_taker.rs
```diff
@@ -62,6 +67,11 @@ impl Actor {
             executor: command::Executor::new(db, process_manager),
         }
     }
+
+    /// Returns whether the maker has accepted our rollover proposal.
```
thomaseizinger: Typo.

daemon/src/setup_taker.rs
```diff
@@ -189,6 +230,21 @@ impl xtra::Actor for Actor {
...
+            async move {
+                tokio::time::sleep(MAKER_RESPONSE_TIMEOUT).await;
+
+                this.send(MakerResponseTimeoutReached {
+                    timeout: MAKER_RESPONSE_TIMEOUT,
+                })
+                .await
+                .expect("can send to ourselves");
+            }
+        };
```
thomaseizinger, line 245: Note: An extension trait on `Address` for a delayed send would allow us to remove this duplication between the actors.

da-kami:
> > This _can_ happen for the taker as well, because if we restart the taker and we were in a contract setup then we would still have the problem of not having the actor anymore and have the CFD forever in the list. I will see to clean that up as well.
> 
> Wouldn't that clean-up require a housekeeping-like task that either runs on startup or continuously?

Yeah, I guess it does, because we won't trigger anything for CFDs that are in such a state, so we will never run into the actor disconnected error. 
Not sure I have to deal with this now if we assume that takers only rarely restart. I'll add a ticket.

thomaseizinger (approved):

thomaseizinger:
Please either record `tech-debt` tickets or put it on your own TODO list if you don't want to action the comments :)

daemon/src/setup_taker.rs
da-kami: I think we should go further and harmonize those actors. They have the same overall corset for taker and maker, i.e. all taker protocols should be aligned and all maker protocols should be aligned.
All these protocols should follow the same pattern of taker proposes, maker accepts/rejects, message exchange, done.

da-kami:
> Wouldn't that clean-up require a housekeeping-like task that either runs on startup or continuously? 

That made me think. I think we should actually have housekeeping again for both taker and maker and add a `xxxFailed` event if our last event is an event from where we can't continue (i.e. any protocol started / accepted without completion)
This makes me think about what our events represent. We have events from which we can't continue at the moment; I wonder if that is a modelling flaw.

daemon/src/setup_taker.rs
thomaseizinger: I am not sold on this.

1. As we are discussing in https://github.com/itchysats/itchysats/pull/1169, rollover 2.0 will likely no longer have accept/reject.
2. Collaborative close might be replaced with "placing an order" and is thus the same as opening a position.
3. With combined positions, placing an order will need to be decoupled from protocol that runs on-chain. I think we will need some logic that figures out stuff like "If I am 1000 contracts long and place a sell order for 500 contract, I am only going to be 500 contracts long." I think this kind of book-keeping will need to be decoupled from what we are actually doing on-chain.

Long story short (pun intended): These protocols only have the same structure because we invented them like that. I am no longer convinced that they should have this structure so I am hesitant to invest into abstractions that would make them be the same.

On the other hand, a generic extension trait like `Address::send_delayed` can be useful in many cases, will reduce the code duplication here and is unopinionated about our domain.

bonomat (approved):
I think this looks good.

daemon/src/setup_taker.rs
da-kami: https://github.com/itchysats/itchysats/issues/1316

da-kami:
> > Wouldn't that clean-up require a housekeeping-like task that either runs on startup or continuously?
> 
> That made me think. I think we should actually have housekeeping again for both taker and maker and add a `xxxFailed` event if our last event is an event from where we can't continue (i.e. any protocol started / accepted without completion) This makes me think about what our events represent. We have events from which we can't continue at the moment; I wonder if that is a modelling flaw.

https://github.com/itchysats/itchysats/issues/1317

PR #1308, opened 2022-02-02
Performance improvements for projection

thomaseizinger:
Instead of rehydrating all CFDs whenever any CFD or the quote changes we now:

- Only rehydrate the specific CFD whenever it changes
- Keep the hydrated CFD around (this is basically an in-memory persistent read model)
- Only apply the new quote to the CFDs whenever the quote changes

As a follow-up, we can eventually take apart the `projection::Cfd` struct into two structs:

- The aggregate hydrated from all events
- The struct that is actually serialized over the wire

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -22,7 +22,6 @@ use crate::model::Price;
 use crate::model::Timestamp;
 use crate::model::TradingPair;
 use crate::model::Usd;
-use crate::xtra_ext::SendAsyncSafe;
 use crate::Order;
 use crate::Tasks;
 use crate::SETTLEMENT_INTERVAL;
@@ -54,9 +53,8 @@ use xtra_productivity::xtra_productivity;
 /// (replaces previously stored values)
 pub struct Update<T>(pub T);
 
-/// Message indicating that the Cfds in the projection need to be reloaded, as at
-/// least one of the Cfds has changed.
-pub struct CfdsChanged;
+/// Indicates that the CFD with the given order ID changed.
+pub struct CfdChanged(pub OrderId);
 
 pub struct Actor {
     db: sqlx::SqlitePool,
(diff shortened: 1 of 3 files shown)


klochowicz (approved):
great change! Projection actor will thank you 😁

daemon/src/projection.rs
```diff
@@ -540,6 +498,59 @@ impl Cfd {
...
+
+                    None
+                }
+            }
+        }).map(|(in_btc, in_percent, payout)| (Some(in_btc), Some(in_percent.round_dp(1).to_string()), Some(payout)))
+            .unwrap_or_else(|| {
+                tracing::debug!(order_id = %self.order_id, "Unable to calculate profit/loss without current price");
+
+                (None, None, None)
+            });
```
bonomat, line 502: Imho you can improve the readability even further if split this logic up and use early returns with `None`, e.g. if `latest_price` is none, you don't have to do half of the function.

daemon/src/projection.rs
```diff
@@ -635,6 +635,21 @@ impl Tx {
...
         let _ = self.quote.send(quote.map(|q| q.into()));
     }
+
+    fn update_order(&self, quote: Option<Order>) {
+        let order = match quote {
+            None => None,
+            Some(order) => match TryInto::<CfdOrder>::try_into(order) {
+                Ok(order) => Some(order),
+                Err(e) => {
+                    tracing::warn!("Unable to convert order: {e:#}");
```
bonomat, line 638: Do we care about which order failed to convert? If so, we could print the `order_id` .

daemon/src/projection.rs
```diff
@@ -711,9 +711,11 @@ impl Actor {
 
     fn handle(&mut self, msg: Update<Option<bitmex_price_feed::Quote>>) {
         self.state.update_quote(msg.0);
-        self.tx.update_quote(msg.0);
 
-        self.refresh_cfds().await;
+        let hydrated_cfds = self.state.cfds.clone();
+
+        self.tx.update_quote(msg.0);
+        self.tx.update_cfds(hydrated_cfds, msg.0);
```
bonomat: Nice 👍

daemon/src/projection.rs
```diff
@@ -631,6 +631,10 @@ impl Tx {
 
         let _ = self.cfds.send(cfds_with_quote);
     }
+
+    fn update_quote(&self, quote: Option<bitmex_price_feed::Quote>) {
+        let _ = self.quote.send(quote.map(|q| q.into()));
```
bonomat, line 629: Should we care about the receiver potentially not being there anymore?  
I think it should not happen as both, the sender and receiver are both managed in `Actor`.

bonomat (approved):
Quite a few good improvements. I think this should help us get the cpu load down a bit :) Nice job

daemon/src/projection.rs
thomaseizinger: I don't think we should care. I've only copied this code so didn't want to change any assumptions here!

daemon/src/projection.rs
thomaseizinger: Yes, didn't see that when doing the other changes. Do it in a follow-up?

daemon/src/projection.rs
thomaseizinger: Do you want to add it as a follow-up PR? I don't want to rebase again and wait for CI :)

daemon/src/projection.rs
bonomat: Sure, nw

PR #1310, opened 2022-02-02
Treat RevokeConfirmed as CommitConfirmed

### LINKED ISSUE #1300: `Not yet implemented: deal with revoke ` Error
Version: latest master: 10ae1fb98c152e1d323edd22913d6b5378368287

I have 5 positions open, the taker and maker were running over night, restarted the taker and get these errors: 

```
thread 'rocket-worker-thread' panicked at 'not yet implemented: Deal with revoked', daemon/src/monitor.rs:277:32
thread 'rocket-worker-thread' panicked at 'not yet implemented: Deal with revoked', daemon/src/monitor.rs:277:32
thread 'rocket-worker-thread' panicked at 'not yet implemented: Deal with revoked', daemon/src/projection.rs:501:32
```

This kills some workers and the CFDs don't get published via SSE events anymore.

bonomat:
Fixes #1300

The problem was that we monitor for the publishment of revoked commit transaction but don't act on it yet. If there is a `todo!` in the code, we will panic and cannot recover from this.

While if the other party publishes a revoked commit transaction we should punish them this logic is not yet implemented.

For now, we just pretend that everything is fine and treat it as if it was the expected commit transaction.
The downside is that we will receive our money according to an old price.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -487,7 +487,10 @@ impl Cfd {
 
                 self.state = CfdState::PendingCommit;
             }
-            RevokeConfirmed => todo!("Deal with revoked"),
+            RevokeConfirmed => {
+                tracing::error!(order_id = %self.order_id, "Revoked logic not implemented");
+                self.state = CfdState::OpenCommitted;
+            }
             RolloverStarted { .. } => match role {
                 Role::Maker => {
                     self.state = CfdState::IncomingRolloverProposal;
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -1237,7 +1237,12 @@ impl Cfd {
                 // commands
             }
             ManualCommit { tx } => self.commit_tx = Some(tx),
-            RevokeConfirmed => todo!("Deal with revoke"),
+            RevokeConfirmed => {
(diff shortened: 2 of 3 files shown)


daemon/src/projection.rs
```diff
@@ -487,7 +487,10 @@ impl Cfd {
 
                 self.state = CfdState::PendingCommit;
             }
-            RevokeConfirmed => todo!("Deal with revoked"),
+            RevokeConfirmed => {
+                tracing::error!(order_id = %self.order_id, "Revoked logic not implemented");
+                self.state = CfdState::PendingCet;
```
thomaseizinger: For `CommitConfirmed`, we set this to `OpenCommitted`.

`PendingCet` is only meant for when we actually published the CET and it is pending in mempool.

thomaseizinger:
Minor typo in the commit message (I fixed it in the PR title)

daemon/src/projection.rs
bonomat: Ah ok. I'll change it.

bonomat:
> Minor typo in the commit message (I fixed it in the PR title)

rebased, squashed and fixed

thomaseizinger (approved):

PR #1311, opened 2022-02-02
Allow closing position without lock confirmation

### LINKED ISSUE #1305: Some of the open CFDs can be closed, some not
All the CFDs have the same state assigned in the UI, where clearly there is something different about them, as not all of them can be closed. This is super confusing:

image

da-kami:
fixes https://github.com/itchysats/itchysats/issues/1305

At some point we only allowed close if lock is confirmed.
That is an unnecessary restriction, as we can chain transactions together and close with lock being unconfirmed.
With the recent change of only displaying `Open` (no `Pending Open`) the behaviour of not being able to click the close button became confusing.
Thus, we allow closing right after the setup finished.

Note: We don't *have* to merge this one but we can.
I tested it locally by closing a position where lock was not confirmed yet and it worked.

### DIFF
--- a/taker-frontend/src/components/CloseButton.tsx
+++ b/taker-frontend/src/components/CloseButton.tsx
@@ -27,7 +27,7 @@ interface Props {
 /// maker is online.
 export default function CloseButton({ cfd, request, status, buttonTitle, isForceCloseButton }: Props) {
     const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED
-        || !(cfd.state.key === StateKey.OPEN);
+        || ![StateKey.OPEN, StateKey.PENDING_OPEN].includes(cfd.state.key);
 
     let popoverBody = <>
         <Text>
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -644,7 +644,7 @@ impl Cfd {
     }
 
     fn can_settle_collaboratively(&self) -> bool {
-        self.lock_finality && !self.commit_finality && !self.is_final() && !self.is_attested()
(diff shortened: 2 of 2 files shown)


bonomat (approved):
Definitely an UX improvement

taker-frontend/src/components/CloseButton.tsx
```diff
@@ -27,7 +27,7 @@ interface Props {
 /// maker is online.
 export default function CloseButton({ cfd, request, status, buttonTitle, isForceCloseButton }: Props) {
     const disableCloseButton = cfd.state.getGroup() === StateGroupKey.CLOSED
```
thomaseizinger, line 29: This button being enabled or not should really be controlled by the backend, based on whether or not the action is returned. That is the point of a RESTful design, that you don't need to duplicate this logic in the frontend :)

thomaseizinger (approved):

taker-frontend/src/components/CloseButton.tsx
thomaseizinger: It is an absolute anti-pattern to reason about the state in the frontend for what you can and cannot do!

thomaseizinger:
Happy to include this in the release.

taker-frontend/src/components/CloseButton.tsx
da-kami: yeah, we should clean this up. I did not want to go down the rabbit hole of cleaning up the projection cfd states now 😬
I'll create a ticket

da-kami: https://github.com/itchysats/itchysats/issues/1314

PR #1320, opened 2022-02-02
A rollover being accepted is not `ContractSetup`

da-kami:
While on an abstract level this is not incorrect, displaying `ContractSetup` after a rollover was accepted is super confusing in the UI.
Recently we said we should just treat all of rollover as `Open`, so that's what we do now.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -431,7 +431,7 @@ impl Cfd {
                 }
             },
             RolloverAccepted => {
-                self.state = CfdState::ContractSetup;
+                self.state = CfdState::Open;
             }
         };
 


thomaseizinger (changes requested):
We have to be careful doing this in the backend because the actions depend on this!

If you want to temporarily fix this, I recommend doing it in the frontend.

da-kami:
> We have to be careful doing this in the backend because the actions depend on this!
> 
> If you want to temporarily fix this, I recommend doing it in the frontend.

Can't do it in the frontend - how would I know if I am actually in contract setup or it's a rollover that is marked as contract setup? We should not re-use the contract setup state here...

da-kami:
If this causes friction I'd say we just don't merge it and accept that users might see weird `Contract Setup` states. 
We can also go back to displaying `Rollover Proposed` or similar, so that `Contract Setup` does not come out of nowhere...?

da-kami:
We should potentially have a separate state for rollover setup sometimes, closing for now.

PR #1321, opened 2022-02-02
Don't force close if collaborative settlement fails

da-kami:
There are multiple different ways how collaborative settlement can fail at the moment, and we don't distinguish between "cannot do it atm" and "it failed in a bad way we should force close".
Since the different error scenarios require some more design work, we remove force closing on a failed collaborative settlement for now.
The only side effect that this can have is, that the taker runs into a failure, but the maker still publishes the collab close transaction, which will only eventually be picked up by the taker.
This behavior is totally acceptable and way better than randomly running into a force close because of e.g. and ongoing rollover or a collab settlement already in progress for the cfd.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -313,11 +313,9 @@ pub enum CfdEvent {
         #[serde(with = "hex_transaction")]
         commit_tx: Transaction,
     },
-    // TODO: What does "failed" mean here? Do we have to record this as event? what would it mean?
-    CollaborativeSettlementFailed {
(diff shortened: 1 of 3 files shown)


da-kami:
Another side effect: The user won't notice that it did not work in the UI because we don't report failed any different. But I think this is OK, the user will try again if it does not work. In the worst case (we are during a rollover) the user might try again several times (until we display `ContractSetup` after rollover accepted).

thomaseizinger (approved):

thomaseizinger (approved):

PR #1323, opened 2022-02-02
Display CFD state "Funding" while issuing rollover proposal

klochowicz:
It should make it easier to understand why things jump to "Contract Setup"
suddenly.


thomaseizinger:
Can we also just say "Funding" while rollover is active?

klochowicz:
> 

we can't easily, we might look into it post-release.

da-kami:
> Can we also just say "Funding" while rollover is active?

That's the problem of mapping it to `ContractSetup` unless I missed somerhing

thomaseizinger (approved):

PR #1333, opened 2022-02-03
Run `clippy` separately on every crate on CI

luckysori:
Cargo appears to accumulate all the features enabled by all the crates in the workspace when building with the `--workspace` flag. Because of this, when running commands with `--workspace` we can fail to detect one of the crates not being able to compile because of a missing feature.


thomaseizinger (changes requested):
Nice!

I am impressed that it actually picks up the cache correctly.
You will have to adjust the `bors.toml` as well to include these new jobs though.

I wonder if there is a way of making sure we don't forget to add one here.

luckysori:
> I am impressed that it actually picks up the cache correctly.

How did you verify that? We definitely don't seem to be slowing down CI with this change.

thomaseizinger:
> > I am impressed that it actually picks up the cache correctly.
> 
> How did you verify that? We definitely don't seem to be slowing down CI with this change.

https://github.com/itchysats/itchysats/runs/5046929348?check_suite_focus=true#step:4:29

thomaseizinger (approved):

bors.toml
```diff
@@ -1,7 +1,16 @@
...
+  "clippy (daemon)",
+  "clippy (rocket-basicauth)",
+  "clippy (rust-embed-rocket)",
+  "clippy (taker)",
+  "clippy (maker)",
+  "clippy (shared-bin)",
+  "clippy (daemon-tests)",
+  "clippy (tokio-tasks)",
+  "clippy (xtras)",
+  "block-fixup",
```
thomaseizinger: Must have been a rebase errr, this check no longer exists!

That is why bors is timing out.

luckysori: Thanks for the fix!

PR #1334, opened 2022-02-03
Extract bitmex-price-feed into dedicated actor

thomaseizinger:
This allows us to reuse this component in other codebases.


daemon/src/bitmex_price_feed.rs
```diff
@@ -9,6 +8,7 @@ use rust_decimal::Decimal;
 use std::convert::TryFrom;
 use std::time::Duration;
 use time::OffsetDateTime;
+use tokio_tasks::Tasks;
```
luckysori, line 9: 👍 Interesting that these use statements are available from `crate`. There are a lot of `use crate::Tasks` in this crate :(

daemon/src/supervisor.rs
```diff
@@ -1,261 +0,0 @@
-use crate::Tasks;
```
luckysori: Ugh, thanks for finding this. The rebase must have reintroduced it.

daemon/src/projection.rs
```diff
@@ -484,7 +484,7 @@ impl Cfd {
         let latest_quote = match latest_quote {
             Some(latest_quote) => latest_quote,
             None => {
-                tracing::debug!(order_id = %self.order_id, "Unable to calculate profit/loss without current price");
+                tracing::trace!(order_id = %self.order_id, "Unable to calculate profit/loss without current price");
```
luckysori, line 499: 🤩

daemon/src/projection.rs
```diff
@@ -962,6 +975,12 @@ mod round_to_two_dp {
         }
     }
 
+    impl ToDecimal for Decimal {
```
luckysori, line 995: 😄

daemon/src/projection.rs
```diff
@@ -502,7 +501,9 @@ impl Cfd {
         let latest_price = match Price::new(latest_price) {
             Ok(latest_price) => latest_price,
             Err(e) => {
-                tracing::warn!("Failed to compute profit/loss because latest price is invalid: {e}");
+                tracing::warn!(
+                    "Failed to compute profit/loss because latest price is invalid: {e}"
+                );
```
luckysori, line 519: ❓ Why do these diffs keep appearing? As in nothing has changed, but these lines are being reformatted.

luckysori (approved):
LGTM

daemon/src/projection.rs
thomaseizinger: I don't know but I will run the formatter again on each commit to check!

daemon/src/bitmex_price_feed.rs
thomaseizinger: I'll see to pack them all in this commit!

luckysori: Thanks!

daemon/src/projection.rs
thomaseizinger: Needed that to make the serde module work :D

PR #1335, opened 2022-02-03
Dynamically create matrix from workspace members

thomaseizinger:


luckysori (approved):
You must have fixed a bug, but I can't tell! This is a nice idea!

thomaseizinger:
> You must have fixed a bug, but I can't tell! This is a nice idea!

What do you mean? I don't think I fixed any bug :)

luckysori:
> What do you mean? I don't think I fixed any bug :)

I meant with your force push, a bug with the original implementation of this PR.

thomaseizinger:
Now we only need a lint that checks that our `bors.toml` file is update to date.

PR #1336, opened 2022-02-03
Clean up a few stale TODOs

thomaseizinger:


luckysori (approved):
A much needed culling imo. Thanks!

klochowicz (approved):
thank you for the cleanup!

daemon/src/model/cfd.rs
```diff
@@ -893,15 +886,7 @@ impl Cfd {
...
     pub fn setup_contract(self, completed: SetupCompleted) -> Result<Event> {
-        // Version 1 is acceptable, as it means that we started contract setup
-        // TODO: Use self.during_contract_setup after introducing
-        // ContractSetupAccepted event
-        if self.version > 1 {
-            bail!(
-                "Complete contract setup not allowed because cfd in version {}",
-                self.version
-            )
-        }
```
klochowicz: I'm afraid you need to keep that as the tests are failing. At this stage we're not yet `during_contract_setup` - I have once tried to align this protocol with others and use that flag more, but the PR never got in due to awkwardness of calling it "contract setup proposed". 

The TODO should go away though, as it was actively misleading - sorry for that! 😭 It should say "once it's aligned with other protocols"

klochowicz (dismissed):

Issue #1150, opened 2022-01-19
Connection actor in taker binary does not follow ADR006


thomaseizinger:
[ADR006](https://github.com/itchysats/itchysats/blob/master/docs/adr/006-actors-for-resources.md) defines that resources should be tied to the lifetime of an actor.

The connection actor in the taker binary currently does not follow this decision. There is a draft PR here that might contain worthwhile commits: https://github.com/itchysats/itchysats/pull/851

luckysori:
This connection actor doesn't exist anymore 🎉

PR #1337, opened 2022-02-03
WIP: Move taker/maker specific actors into their crates

thomaseizinger:
- [ ] Reduce API surface of `connection` actor by sending all addresses at once in `Connect` (Perhaps address https://github.com/itchysats/itchysats/issues/1150 altogether which allows us to express the re-connect logic through supervisor restarts)
- [ ] Finish https://github.com/itchysats/itchysats/pull/1294 to either get rid of `FutureExt::timeout` dependency in `connection` and `maker_inc_connection` actor so we don't have to expose it


thomaseizinger:
Closed due to inactivity.

PR #1342, opened 2022-02-04
Make sure price feed actor is actually restarted

thomaseizinger:


bonomat (approved):
Really good find. I'm happy to stay on websockets as this seems to be the recommend way to stay within the connection limit.

daemon/src/bitmex_price_feed.rs
```diff
@@ -93,6 +93,10 @@ impl xtra::Actor for Actor {
             let _: Result<(), Disconnected> = this.send(e).await;
         });
     }
+
+    async fn stopping(&mut self, _: &mut xtra::Context<Self>) -> xtra::KeepRunning {
```
da-kami, line 97: Just to make sure I understand this correctly: We do always expect that we call `Context::stop` from within this actor to trigger the wanted behavior? 

---

Xtra docs:

> Note: this method will only be called when [Context::stop](https://docs.rs/xtra/0.5.1/xtra/struct.Context.html#method.stop) is called from this actor. If the last strong address to the actor is dropped, or [Context::stop](https://docs.rs/xtra/0.5.1/xtra/struct.Context.html#method.stop) is called from another actor on the same address, this will not be called. Therefore, Other, general destructor behaviour should be encapsulated in the [Actor::stopped](https://docs.rs/xtra/0.5.1/xtra/trait.Actor.html#method.stopped) method.

https://docs.rs/xtra/0.5.1/xtra/trait.Actor.html

da-kami (approved):

luckysori (approved):

daemon/src/bitmex_price_feed.rs
thomaseizinger: Yes, if you want the actor to stop, we need to call `Context::stop`. In our codebase, we never have more than one instance of each actor running. This is quite confusing so I am tempted to patch this entire feature away in our `xtra` fork.

da-kami: What we return here in `stopping` is the propagated to `stopped` as far as I understand, that should be fine, just wondering if we always trigger `Context::stop` in a context that will actually trigger `stopping` or if there are other cases.

PR #1343, opened 2022-02-04
Poll for latest XBTUSD BitMEX quote

luckysori:
Instead of using a WebSocket which is a connection that needs to be managed.

Yet to do:

- Rename `bitmex_price_feed` to `bitmex_quote`.
- Separate `wire` type `Quote` from `bitmex_quote::Quote`.

### DIFF
--- a/daemon/src/bitmex_price_feed.rs
+++ b/daemon/src/bitmex_price_feed.rs
@@ -1,165 +1,30 @@
 use crate::model::Price;
-use crate::model::Timestamp;
-use crate::Tasks;
+use anyhow::bail;
 use anyhow::Result;
 use async_trait::async_trait;
-use futures::SinkExt;
-use futures::TryStreamExt;
-use rust_decimal::Decimal;
-use std::convert::TryFrom;
+use serde::Deserialize;
+use serde::Serialize;
 use std::time::Duration;
+use time::ext::NumericalDuration;
 use time::OffsetDateTime;
-use tokio_tungstenite::tungstenite;
-use xtra::Disconnected;
+use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
-use xtras::supervisor;
+use xtras::SendInterval;
 
-pub const QUOTE_INTERVAL_MINUTES: i64 = 1;
+pub const QUOTE_INTERVAL_MINUTES: u64 = 1;
+const QUOTE_SYNC_INTERVAL_SECONDS: u64 = (QUOTE_INTERVAL_MINUTES * 60) - 15;
 
-pub struct Actor {
-    tasks: Tasks,
-    latest_quote: Option<Quote>,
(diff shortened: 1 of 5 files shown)


daemon/src/bitmex_price_feed.rs
```diff
@@ -36,184 +65,172 @@ impl Actor {
...
-        pub bid_size: u64,
-        pub ask_size: u64,
-        pub bid_price: f64,
-        pub ask_price: f64,
-        pub symbol: String,
-        pub timestamp: String,
+    async fn get_latest_quote(&self) -> Result<Quote> {
+        let mut url = reqwest::Url::parse(BITMEX_QUOTE_URL).expect("valid URL");
+        let query = serde_urlencoded::to_string(GetQuoteBucketedRequest::latest_xbt_usd())
+            .expect("valid query");
```
thomaseizinger, line 169: I'd probably just hardcode this URL with is query parameters, same as we did before!

thomaseizinger: You should then be able to just use `reqwest::get`.

daemon/src/bitmex_price_feed.rs
```diff
@@ -36,184 +65,172 @@ impl Actor {
...
+                    bail!("Failed to fetch latest quote from BitMEX: {name}, {message}")
+                }
+                Err(deser_err) => {
+                    tracing::warn!("Failed to deserialize BitMEX error: {}", deser_err);
+                    bail!("Failed to fetch latest quote from BitMEX: {}", content)
+                }
+            }
+        }
+
+        let quotes = match serde_json::from_str::<Vec<Quote>>(&content) {
```
thomaseizinger, line 196: Could deserialize into a 1-element array.

daemon/src/bitmex_price_feed.rs
```diff
@@ -36,184 +65,172 @@ impl Actor {
...
+        if !status.is_success() {
+            match serde_json::from_str::<BitmexErrorResponse>(&content) {
+                Ok(BitmexErrorResponse {
+                    error: BitmexErrorMessage { message, name },
+                }) => {
+                    bail!("Failed to fetch latest quote from BitMEX: {name}, {message}")
+                }
+                Err(deser_err) => {
+                    tracing::warn!("Failed to deserialize BitMEX error: {}", deser_err);
+                    bail!("Failed to fetch latest quote from BitMEX: {}", content)
```
thomaseizinger, line 191: I wouldn't bother doing `resp.text()` just for the log message.

thomaseizinger (commented):
Some ideas!

luckysori:
Closed because [we think we have figured out](https://github.com/itchysats/itchysats/pull/1342) how to reliably restart the WebSocket-based price feed actor, so we don't need to rewrite it using HTTP requests.

PR #1344, opened 2022-02-04
Re-write supervisor pattern to work with `xtra::Actor::Stop`

thomaseizinger:
With this patch, we are adopting a proposed patch of xtra that
allows actors to return a value from `stopped`. We utilize this
communicate the reason why an actor stopped back to the supervisor.

This has multiple benefits:

a) We no longer need the supervisor's address within the supervised actor.
b) It is impossible to "forget" to report the stop reason to the supervisor.
c) It allows us to easily supervise ANY actor because they no longer
need to have "knowledge" of the `supervisor` module.

The fork of xtra that we are bumping in this patch also contains another
patch that is of interest to the supervisor pattern: We change the default
of `Actor::stopping` to `KeepRunning::StopSelf` which makes our supervisor
pattern work in the first place.

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -175,7 +175,9 @@ impl Maker {
         tasks.add(wallet_fut);
 
         let (price_feed_addr, price_feed_fut) = price_feed.create(None).run();
-        tasks.add(price_feed_fut);
+        tasks.add(async move {
+            let _ = price_feed_fut.await;
+        });
 
         let settlement_interval = SETTLEMENT_INTERVAL;
 
@@ -289,7 +291,7 @@ impl Taker {
 
                 Ok(monitor)
             },
-            move |_| price_feed.clone(),
+            move || price_feed.clone(),
             config.n_payouts,
             config.heartbeat_interval,
             Duration::from_secs(10),
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -143,14 +143,17 @@ where
 #[async_trait]
 impl<O> xtra::Actor for Actor<O>
 where
-    O: 'static,
-    Self: xtra::Handler<AutoRollover>,
+    O: xtra::Handler<oracle::GetAnnouncement> + 'static,
 {
+    type Stop = ();
+
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("we are alive");
         self.tasks
             .add(this.send_interval(Duration::from_secs(5 * 60), || AutoRollover));
     }
+
+    async fn stopped(self) -> Self::Stop {}
 }
 
 /// Message sent to ourselves at an interval to check if rollover can
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -240,10 +240,14 @@ async fn main() -> Result<()> {
         p2p_socket,
     )?;
 
-    let (supervisor, price_feed) = supervisor::Actor::new(
-        bitmex_price_feed::Actor::new,
-        |_| true, // always restart price feed actor
-    );
+    let (supervisor, price_feed) =
+        supervisor::Actor::with_policy(bitmex_price_feed::Actor::default, |e| match e {
+            bitmex_price_feed::Error::FailedToParseQuote { .. }
+            | bitmex_price_feed::Error::Failed { .. }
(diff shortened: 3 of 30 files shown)


daemon/src/auto_rollover.rs
```diff
@@ -146,11 +146,17 @@ where
...
+    type Stop = ();
+
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("we are alive");
         self.tasks
             .add(this.send_interval(Duration::from_secs(5 * 60), || AutoRollover));
     }
+
+    async fn stopped(self) -> Self::Stop {
+        todo!()
```
thomaseizinger: Whops, that is not good.

daemon-tests/src/lib.rs
```diff
@@ -175,7 +175,9 @@ impl Maker {
         tasks.add(wallet_fut);
 
         let (price_feed_addr, price_feed_fut) = price_feed.create(None).run();
-        tasks.add(price_feed_fut);
+        tasks.add(async move {
+            let _ = price_feed_fut.await;
+        });
```
luckysori, line 180: 🙃 This works too:
```suggestion
        use futures::FutureExt;
        tasks.add(price_feed_fut.map(|_| ()));
```

daemon/src/auto_rollover.rs
```diff
@@ -143,14 +143,17 @@ where
 #[async_trait]
 impl<O> xtra::Actor for Actor<O>
 where
-    O: 'static,
-    Self: xtra::Handler<AutoRollover>,
+    O: xtra::Handler<oracle::GetAnnouncement> + 'static,
```
luckysori, line 146: ❓ What's this change about?

maker/src/main.rs
```diff
@@ -240,10 +240,14 @@ async fn main() -> Result<()> {
...
-        bitmex_price_feed::Actor::new,
-        |_| true, // always restart price feed actor
-    );
+    let (supervisor, price_feed) =
+        supervisor::Actor::with_policy(bitmex_price_feed::Actor::default, |e| match e {
+            bitmex_price_feed::Error::FailedToParseQuote { .. }
+            | bitmex_price_feed::Error::Failed { .. }
+            | bitmex_price_feed::Error::FailedToConnect { .. }
+            | bitmex_price_feed::Error::Unspecified
+            | bitmex_price_feed::Error::StreamEnded => true, // always restart price feed actor
```
luckysori, line 249: ❓ I guess we're explicitly matching because we want to reconsider our policy every time we modify these `Error`s?

xtras/src/supervisor.rs
```diff
@@ -28,18 +28,58 @@ struct Metrics {
...
+where
+    T: xtra::Actor<Stop = ()>,
+{
+    /// Construct a new supervisor for an [`Actor`] with an [`xtra::Actor::Stop`] value of `()`.
+    ///
+    /// The actor will always be restarted if it stops. If you don't want this behaviour, don't use
+    /// a supervisor. If you want more fine-granular control in which circumstances the actor
+    /// should be restarted, set [`xtra::Actor::Stop`] to a more descriptive value and use
+    /// [`Actor::with_policy`].
+    pub fn new(ctor: impl (Fn() -> T) + Send + 'static) -> (Self, Address<T>) {
```
luckysori, line 55: 👍 I like the design with `new` and `with_policy`. Just in case you didn't know, `new` is unused outside of this crate. But I think that's okay.

luckysori (approved):
LGTM!

xtras/src/supervisor.rs
thomaseizinger: Yep, we still need to adopt it!

daemon/src/auto_rollover.rs
thomaseizinger: Uhm I got weird compile-errors without it. Honestly don't understand why :see_no_evil:

maker/src/main.rs
thomaseizinger: Yes but also the introduction of `UnitReason` required me to distinguish between the stop value `S` and the reason of the policy `R`. There is an `Into` requirement that only works if we know what the reason `R` is. By matching explicitly, we give the compiler a type-hint that `R` == `S` in this case and it doesn't whinge about unknown types.

PR #1354, opened 2022-02-05
Instrument log with taker_id

bonomat:
This allows for easier log analysis

Would be nice to backport this to 0.4.x so that we get better logs for the maker

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -194,7 +194,7 @@ where
         taker_id: Identity,
         ctx: &mut xtra::Context<Self>,
     ) -> Result<()> {
-        tracing::info!(%order_id, %taker_id,  "Received proposal from taker");
+        tracing::info!(%order_id, %taker_id,  "Received rollover proposal from taker");
         let this = ctx.address().expect("acquired own address");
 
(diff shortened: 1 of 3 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -194,7 +194,7 @@ where
         taker_id: Identity,
         ctx: &mut xtra::Context<Self>,
     ) -> Result<()> {
-        tracing::info!(%order_id, "Received proposal from taker {taker_id}");
+        tracing::info!(%order_id, %taker_id,  "Received rollover proposal from taker");
```
thomaseizinger, line 197: I think this one is duplicated with the protocol actor.

thomaseizinger (approved):

daemon/src/maker_cfd.rs
bonomat: Can't find it.

thomaseizinger: https://github.com/itchysats/itchysats/blob/6e4d2da18c2bd29528fd9379223880d66a153519/daemon/src/rollover_maker.rs#L236-L239

PR #1356, opened 2022-02-05
Extend Cfd test model to be able to test collab settlement

da-kami:
This should help us to get way more confidence in the internal `Cfd` logic and test edge cases.
So far I only added one test (the asserts are still garbage in that test). Once that test passes we can take it further.

Goals:
- Be able to assert on the actual payout to make sure it is sane.
- Be able to test edge cases and various combinations.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -415,6 +415,7 @@ fn from_json_inner(name: String, data: String) -> Result<CfdEvent> {
 /// `version`). To bring a cfd into a certain state version we load all events from the
 /// database and apply them in order (order by version).
 #[derive(Debug, PartialEq)]
+#[cfg_attr(test, derive(Clone))]
 pub struct Cfd {
     version: u64,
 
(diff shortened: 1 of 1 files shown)


daemon/src/model/cfd.rs
```diff
@@ -2517,6 +2766,26 @@ mod tests {
...
+            amount_maker: Amount,
+            taker_address: Address,
+            maker_address: Address,
+            identity_sk: SecretKey,
+        ) -> Self {
+            self.lock = (lock_tx.extract_tx(), descriptor);
+            self.taker_lock_amount = amount_taker;
+            self.maker_lock_amount = amount_maker;
+            self.identity = identity_sk;
+            self.taker_address = taker_address;
```
da-kami, line 2874: I don't think the addresses should matter here, but since we use them to get the script for the respective party I was not sure and added them from the build params.

daemon/src/model/cfd.rs
```diff
@@ -2456,18 +2533,165 @@ mod tests {
...
+                fee_rate: Default::default(),
+            };
+
+            let taker_params = build_party_params(taker_params).unwrap();
+            let maker_params = build_party_params(maker_params).unwrap();
+
+            let taker_address = taker_params.address.clone();
+            let maker_address = maker_params.address.clone();
+
+            let (lock_tx, descriptor) = build_lock_tx(maker_params, taker_params);
```
da-kami, line 2676: Could the problem be that the lock transaction stored in the Dlc is not signed? 

I can sign it as well, but I thought that should not matter as long as we don't publish on chain 🤔

daemon/src/model/cfd.rs
da-kami: Answer myself: The addresses are definitely needed because we use the script pubkey of the address when building the close transaction.

da-kami:
This PR just does not show my latest force pushed changes... This is really weird. When opening a new one it works.

PR #1357, opened 2022-02-05
Extend Cfd test model to be able to test collab settlement

da-kami:
Goals:
- Be able to assert on the actual payout to make sure it is sane.
- Be able to test edge cases and various combinations.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -415,6 +415,7 @@ fn from_json_inner(name: String, data: String) -> Result<CfdEvent> {
 /// `version`). To bring a cfd into a certain state version we load all events from the
 /// database and apply them in order (order by version).
 #[derive(Debug, PartialEq)]
+#[cfg_attr(test, derive(Clone))]
 pub struct Cfd {
     version: u64,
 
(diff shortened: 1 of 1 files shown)


daemon/src/model/cfd.rs
```diff
@@ -2361,6 +2373,140 @@ mod tests {
...
+        // 0 rollover
+        assert_collab_settlement_taker_long_maker_short(
+            quantity,
+            leverage,
+            opening_price,
+            closing_price,
+            positive_funding_rate,
+            1000,
+            0,
+            Amount::from_sat(49583),
```
da-kami: Interestingly, if there is no price change at all and no fees we still end up in an interval that can be in our favor or not. I am not sure this can be avoided with a model the dicretizes evenly, but I never really took it into account when I thought about the values displayed in the UI.
In this case it's "in favor" or the maker who wins some sats.

daemon/src/model/cfd.rs
```diff
@@ -2361,6 +2373,140 @@ mod tests {
         assert_eq!(funding_fee.rate, funding_rate);
     }
 
+    #[test]
+    fn ensure_collaborative_settlement_takes_rollover_fees_into_account() {
```
thomaseizinger, line 2377: Split these into multiple (i.e. one assertion per test).

daemon/src/model/cfd.rs
```diff
@@ -2361,6 +2373,140 @@ mod tests {
...
+            calculate_long_margin(opening_price, quantity, leverage),
+            Amount::from_sat(50000)
+        );
+        assert_eq!(
+            calculate_short_margin(opening_price, quantity),
+            Amount::from_sat(100000)
+        );
+
+        // 0 rollover
+        assert_collab_settlement_taker_long_maker_short(
```
thomaseizinger: Please return the two `Amount`s and assert on the outside.

thomaseizinger (changes requested):

daemon/src/model/cfd.rs
da-kami: Could we do that in future iterations once we have more tests? 
I think it really helps grouping them to see the behavior of the specific scenario in comparison (i.e. all of these tests are related to the number of rollovers). If I split this and then add more tests on different scenarios I think this will become messy and hard to see for the reader what is related. 

Is there a possibility for grouping tests together in rust? use a module? 🤔

da-kami:
Additionally to @thomaseizinger comments this is what we discussed as follow ups:

- Don't create an actual lock transaction, but just plug a transaction in that is `margin_taker + margin_maker = fees` - this also allows us to simulate fees and simplify the setup.
- https://github.com/itchysats/itchysats/issues/1362

thomaseizinger (approved):

PR #1315, opened 2022-02-02
Make rollover and collab settlement mutually exclusive

da-kami:
fixes #1305

Additional fix for "weird" close button behavior in the UI.

Issue #206, opened 2021-10-05
State transitions upon witnessing tx `in mempool`


da-kami:
## TLDR

Add support to transition the CFD of the party that did not trigger the tx publication into pending state upon witnessing a tx in mempool. 

## TLR

At the moment, when one party publishes a transaction we only transition this party into a `pending` state until the transaction was confirmed. 
The other party only notices a change in the UI once we reach finality - which can be significant time. 
This is not an issue per-se, because if the other party tries to publish *the same* action it will transition. However, it can have weird side effects. E.g. actions that should not be a available anymore are still available and may lead to errors. 
Example: Maker publishes commit. Taker tries to go for collaborative settlement which will be declined. The taker could have already known that that's not possible anymore.

DeliciousHair:
Incorporating into #357

da-kami:
I think this ticket should not be in scope at the moment. We should tackle this once the state machine was refactored. 

This ticket is not super important - it does give the user more immediate feedback, but without that feedback there is only few bugs or side effects (like triggering actions that are potentially not available anymore...). This becomes even less important as we removed the actions in question (rollover) in the new UI. The new UI only has `Close` and the maker is assumed to not force close at the moment. 

This ticket has quite a rat-tail that is: If we transition to "in mempool" states that don't involve confirmation we also have to make sure to retry publishing transations  in such states (...)

bonomat:
This project is unmaintained

PR #1359, opened 2022-02-06
Define more restrictively what it means that a cfd is closed

### LINKED ISSUE #1355: Don't commit if already in settlement
We should prevent the publishing of a commit transaction if there is already a settlement ongoing. 

At the moment, the mempool takes care that this does: these two transactions are competing and  only one can be included at a time in the mempool. Since we pay the same fee, the minors don't prefer one over the other. 

If we were, a commit transaction could RBF (replace by fee) a settlement transaction. 

Which, imho, we should prevent.

da-kami:
Fixes https://github.com/itchysats/itchysats/issues/1355

Previously we mostly used `is_final` as a guard, but it can take significant time until e.g. a collab close transaction gets confirmed which leads to other logic being triggered that is actually not possible anymore.
After a collab close transaction was published, other spend transactions will fail when publishing them, but since we don't handle errors in post-processing additional events will be appended so the `Cfd` was seemingly force-close even though it was already collaboratively closed.

---

Note that there are some other issues attached to this problem:

- We don't deal with in-mempool which is an old problem: #206 - this only changes the problem but not solves it. Instead of asserting that we *have* a transaction available (and assume that is was/will be published) we could assert on finding it in mempool.
- We are not dealing with post-processing failures. Effectively e.g. non-collab settlement *cannot* happen once the collab settlement tx is publsihed, but we store the event nontheless and then the post processing fails. Same for rollover (where the problem becomes visible in the UI more clearly, that's why we added this condition: #1315 )

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -651,8 +651,6 @@ impl Cfd {
 
     fn can_settle_collaboratively(&self) -> bool {
         !self.commit_finality && !self.is_final() && !self.is_attested()
-            // Rollover and collaborative settlement are mutually exclusive, if we are currently rolling over we cannot settle
-            && !self.during_rollover
     }
 
     fn is_attested(&self) -> bool {
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -237,12 +237,10 @@ pub enum NoRolloverReason {
     NotLocked,
     #[error("Cannot roll over when CFD is committed")]
     Committed,
-    #[error("Cannot roll over when CFD is attested by oracle")]
-    Attested,
     #[error("Cannot roll over while CFD is in collaborative settlement")]
     InCollaborativeSettlement,
-    #[error("Cannot roll over when CFD is final")]
-    Final,
+    #[error("Cannot roll over when CFD is already closed")]
+    Closed,
 }
 
 /// Errors that can happen when handling the expiry of the refund
@@ -313,10 +311,12 @@ pub enum CfdEvent {
     // TODO: The monitoring events should move into the monitor once we use multiple
     // aggregates in different actors
     LockConfirmed,
-    /// The lock transaction is confirmed after CFD is already final
+    /// The lock transaction is confirmed after CFD was closed
     ///
     /// This can happen in cases where we publish a settlement transaction while the lock
     /// transaction is still pending and they end up in the same block.
+    /// We include cases where we already have a transaction spending from lock, but it might not
+    /// be final yet.
     LockConfirmedAfterFinality,
     CommitConfirmed,
     CetConfirmed,
@@ -610,6 +610,10 @@ impl Cfd {
         self.settlement_proposal.is_some()
     }
 
+    fn is_in_force_close(&self) -> bool {
+        self.commit_tx.is_some()
+    }
+
     pub fn can_auto_rollover_taker(&self, now: OffsetDateTime) -> Result<(), NoRolloverReason> {
         let expiry_timestamp = self.expiry_timestamp().ok_or(NoRolloverReason::NoDlc)?;
         let time_until_expiry = expiry_timestamp - now;
@@ -623,12 +627,8 @@ impl Cfd {
     }
 
     fn can_rollover(&self) -> Result<(), NoRolloverReason> {
-        if self.is_final() {
-            return Err(NoRolloverReason::Final);
-        }
-
-        if self.is_attested() {
-            return Err(NoRolloverReason::Attested);
+        if self.is_closed() {
+            return Err(NoRolloverReason::Closed);
         }
 
         if self.commit_finality {
@@ -639,10 +639,13 @@ impl Cfd {
             return Err(NoRolloverReason::NotLocked);
         }
 
+        if self.is_in_force_close() {
+            return Err(NoRolloverReason::Committed);
+        }
+
         // Rollover and collaborative settlement are mutually exclusive, if we are currently
         // collaboratively settling we cannot roll over
-        if self.is_in_collaborative_settlement() || self.collaborative_settlement_spend_tx.is_some()
-        {
+        if self.is_in_collaborative_settlement() {
             return Err(NoRolloverReason::InCollaborativeSettlement);
         }
 
@@ -650,7 +653,7 @@ impl Cfd {
     }
 
     fn can_settle_collaboratively(&self) -> bool {
-        !self.commit_finality && !self.is_final() && !self.is_attested()
+        !self.is_closed() && !self.commit_finality && !self.is_attested()
     }
 
     fn is_attested(&self) -> bool {
@@ -659,6 +662,21 @@ impl Cfd {
 
     fn is_final(&self) -> bool {
         self.collaborative_settlement_finality || self.cet_finality || self.refund_finality
+    }
+
+    fn is_collaboratively_closed(&self) -> bool {
+        self.collaborative_settlement_spend_tx.is_some()
+    }
+
+    fn is_refunded(&self) -> bool {
+        self.refund_tx.is_some()
+    }
+
+    fn is_closed(&self) -> bool {
+        self.is_final()
+            || self.is_attested()
+            || self.is_collaboratively_closed()
+            || self.is_refunded()
     }
 
     pub fn start_contract_setup(&self) -> Result<(Event, SetupParams)> {
@@ -975,8 +993,11 @@ impl Cfd {
     }
 
     /// Given an attestation, find and decrypt the relevant CET.
+    ///
+    /// In case the Cfd was already closed we return `Ok(None)`, because then the attestation is not
+    /// relevant anymore. We don't treat this as error because it is not an error scenario.
     pub fn decrypt_cet(self, attestation: &oracle::Attestation) -> Result<Option<Event>> {
-        if self.is_final() {
+        if self.is_closed() {
             return Ok(None);
         }
 
@@ -1031,7 +1052,7 @@ impl Cfd {
     pub fn handle_refund_timelock_expired(self) -> Result<Event, RefundTimelockExpiryError> {
         use RefundTimelockExpiryError::*;
 
-        if self.is_final() {
+        if self.is_closed() {
             return Err(AlreadyFinal);
         }
 
@@ -1045,7 +1066,7 @@ impl Cfd {
 
     pub fn handle_lock_confirmed(self) -> Event {
         // For the special case where we close when lock is still pending
-        if self.is_final() {
+        if self.is_closed() {
             return self.event(CfdEvent::LockConfirmedAfterFinality);
         }
 
@@ -1073,7 +1094,7 @@ impl Cfd {
     }
 
     pub fn manual_commit_to_blockchain(&self) -> Result<Event> {
-        anyhow::ensure!(!self.is_final());
+        anyhow::ensure!(!self.is_closed());
 
         let dlc = self.dlc.as_ref().context("Cannot commit without a DLC")?;
 
@@ -1864,6 +1885,7 @@ mod tests {
     use crate::seed::Seed;
     use crate::wallet;
     use crate::wallet::BuildPartyParams;
+    use crate::Attestation;
     use crate::N_PAYOUTS;
     use bdk::bitcoin::util::psbt::Global;
     use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
@@ -2338,10 +2360,7 @@ mod tests {
 
         let cannot_roll_over = cfd.can_rollover().unwrap_err();
 
-        assert!(matches!(
-            cannot_roll_over,
-            NoRolloverReason::Attested { .. }
-        ))
+        assert!(matches!(cannot_roll_over, NoRolloverReason::Closed))
     }
 
     #[test]
@@ -2352,7 +2371,7 @@ mod tests {
 
         let cannot_roll_over = cfd.can_rollover().unwrap_err();
 
-        assert!(matches!(cannot_roll_over, NoRolloverReason::Final))
+        assert!(matches!(cannot_roll_over, NoRolloverReason::Closed))
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


daemon/src/model/cfd.rs
```diff
@@ -317,7 +317,7 @@ pub enum CfdEvent {
     ///
     /// This can happen in cases where we publish a settlement transaction while the lock
     /// transaction is still pending and they end up in the same block.
-    LockConfirmedAfterFinality,
+    LockConfirmedAfterClose,
```
klochowicz: would changing the existing event cause a lot of disruptions at this stage? perhaps we should append a new one instead of changing this one?

daemon/src/model/cfd.rs
```diff
@@ -610,6 +608,10 @@ impl Cfd {
         self.settlement_proposal.is_some()
     }
 
+    fn is_in_non_collaborative_settlement(&self) -> bool {
```
klochowicz: `is_in_force_close()`? not sure whether we've reached full agreement on the naming yet, but last I checked we were leaning towards "close" and "force-close"

daemon/src/model/cfd.rs
```diff
@@ -241,8 +241,8 @@ pub enum NoRolloverReason {
     Attested,
     #[error("Cannot roll over while CFD is in collaborative settlement")]
     InCollaborativeSettlement,
-    #[error("Cannot roll over when CFD is final")]
-    Final,
```
klochowicz, line 245: as per your commit message: an alternative way we could deal with appending events after "closing" cfd would be just treating the *first* closure as the one that got in (in other words, it could be handled in during rehydration phase - it doesn't need to look at the last event in all cases)

daemon/src/model/cfd.rs
da-kami: That's a very good point!

daemon/src/model/cfd.rs
da-kami: I think this goes against the paradigm of even-sourcing. We would start "ignoring" events. I think it's better to keep the even history correct. We could use it as a workaround to deal with `Cfds` that already are in a "wrong" state, but I think those should fix itself, once we get a confirmation on chain, because the confirmation will actually close it through finality.

daemon/src/model/cfd.rs
da-kami: I opted for keeping the old name `LockConfirmedAfterFinality` which is not entirely correct, but I think acceptable. I extended the doc comment.

daemon/src/model/cfd.rs
klochowicz: I think we can do both things eventually; trying not to get into a weird state in the first place is of course preferred as it makes our lives easier.
However, if we appended an event of e.g. closing a closed cfd, this could not be necessary bad, as long as this would lead to another event "closing failed" etc. Of course it goes outside of the scope of this PR, but it's possible :)

daemon/src/model/cfd.rs
```diff
@@ -978,6 +996,9 @@ impl Cfd {
 
     /// Given an attestation, find and decrypt the relevant CET.
     pub fn decrypt_cet(self, attestation: &oracle::Attestation) -> Result<Option<Event>> {
+        // TODO: Should this be is_closed instead? i.e. should we *not* decrypt the CET if we have
```
klochowicz: it would make sense for me to change it, as trying to publish the transaction will fail anyway down the track?
Shall we add a test to cover this scenario, so we're explicit about desired handling of it? (not in this PR of course)

daemon/src/model/cfd.rs
```diff
@@ -654,8 +654,6 @@ impl Cfd {
 
     fn can_settle_collaboratively(&self) -> bool {
         !self.is_closed() && !self.commit_finality && !self.is_attested()
-            // Rollover and collaborative settlement are mutually exclusive, if we are currently rolling over we cannot settle
```
klochowicz, line 654: thanks for writing detailed reasoning for removing these lines! much appreciated!

daemon/src/model/cfd.rs
```diff
@@ -2577,6 +2707,30 @@ mod tests {
             ]
         }
 
+        fn dummy_ongoing_collab_settlement(order_id: OrderId) -> Vec<Self> {
```
klochowicz: nit: I had a hard time reading what would "ongoing" mean - did you try say that it got started? maybe it's easier to say:
`apply_dummy_collab_settlement_started_event` or something similar?

daemon/src/model/cfd.rs
```diff
@@ -2389,6 +2390,135 @@ mod tests {
...
+
+        let taker_keys = crate::keypair::new(&mut rand::thread_rng());
+        let maker_keys = crate::keypair::new(&mut rand::thread_rng());
+
+        let taker_long = Cfd::taker_long()
+            .with_id(order_id)
+            .with_quantity(quantity)
+            .with_opening_price(opening_price)
+            .with_leverage(leverage)
+            .dummy_open(dummy_event_id());
```
klochowicz: this boilerplate makes these tests not that readable.... I'll post a quick PR on top of it if you don't mind, I have something in mind for it 😀

daemon/src/model/cfd.rs
```diff
@@ -2416,6 +2416,83 @@ mod tests {
...
+
+        let _ = assert!(cfd.start_rollover().is_ok());
+
+        let (cfd, _, _, _) = cfd
+            .with_lock(taker_keys, maker_keys)
+            .dummy_collab_settlement_taker(opening_price);
+
```
klochowicz, line 2438: should we test with the ongoing collab settlement here? (or maybe in yet another test?) I have to admit still am pretty hazy in regards to the distinction "in collab settlement" vs "in ongoing collab settlement".

is this mapping right?
- "in collab settlement" - collab settlement proposed
- "in ongoing collab settlement" -  collab settlement got accepted and we're doing the protocol

klochowicz (approved):
I haven't found anything blocking, most of the comments I added are just questions - we can also clean it up a bit further later to minimise boilerplate, for which I'm happy to submit a PR :)

daemon/src/model/cfd.rs
da-kami: definitely! If you have something in mind already that's very much appreciated. I wanted to get them done with what's currently there and then do a round of refactoring myself. If you already start / do with that's even better :)

daemon/src/model/cfd.rs
da-kami: I'm happy to see work toward that in a follow up PR! I think it's an interesting idea

daemon/src/model/cfd.rs
da-kami: The test below covers the `ongoing` case. I wanted to always cover both collab settlement finished (and we have the tx) and ongoing (where we have the settlement only). It's quite a lot of boilerplate which is annoying.

As for the mapping

> "in collab settlement" - collab settlement proposed
> "in ongoing collab settlement" - collab settlement got accepted and we're doing the protocol

that's not how I defined it, my definition is:

1. ongoing: we have a proposal but no collab settlement transaction yet (i.e. the settlement protocol is not finished yet)
2. completed: we have the collab settlement transaction (the settlement protocol completed successfully) but we don't know anything about finality yet
3. final: we reached finality of the transaction

We could document this somewhere, but not sure where...

daemon/src/model/cfd.rs
da-kami: yeah, the `Event` extension should not know about how this is used in the test so started makes more sense, because that's the only terminology the `Event` knows. 

From the test's perspective started might not be enough to understand why the test was written. I tried to convey "started but not finished" with ongoing... Not sure it actually helps.

It might actually be better to produce the dummy events more fine granular and then chain more together instead of bundling multiple events into one dummy scenario. We could give that a try in a refactor on top.

daemon/src/model/cfd.rs
da-kami: Yeah, in a later commit I actually did 😅 - should probably squash these.

daemon/src/model/cfd.rs
```diff
@@ -2371,6 +2390,212 @@ mod tests {
         assert_eq!(funding_fee.rate, funding_rate);
     }
 
+    #[test]
+    fn given_collab_settlement_then_cannot_start_rollover() {
+        let quantity = Usd::new(dec!(10));
+        let leverage = Leverage::new(2).unwrap();
+        let opening_price = Price::new(dec!(10000)).unwrap();
```
thomaseizinger, line 2397: These values are not important for the test right? Would be good to hide them then! Would `Cfd::taker_long().dummy_open()` be sufficient?

daemon/src/model/cfd.rs
```diff
@@ -2371,6 +2390,212 @@ mod tests {
...
+        let taker_keys = crate::keypair::new(&mut rand::thread_rng());
+        let maker_keys = crate::keypair::new(&mut rand::thread_rng());
+
+        let cfd = Cfd::taker_long()
+            .with_quantity(quantity)
+            .with_opening_price(opening_price)
+            .with_leverage(leverage)
+            .dummy_open(dummy_event_id());
+
+        let _ = assert!(cfd.start_rollover().is_ok());
```
thomaseizinger: The test name doesn't mention that we are testing the scenario of "rollover started - collab settlement started - rollover started".

Remove this?

daemon/src/model/cfd.rs
```diff
@@ -2371,6 +2390,212 @@ mod tests {
...
+
+        let (cfd, _, _, _) = cfd
+            .with_lock(taker_keys, maker_keys)
+            .dummy_collab_settlement_taker(opening_price);
+
+        let err = cfd.start_rollover().unwrap_err();
+        let no_rollover_reason = err.downcast::<NoRolloverReason>().unwrap();
+        assert_eq!(no_rollover_reason, NoRolloverReason::Closed);
```
thomaseizinger: Nit: Could do better use of whitespace here:

```suggestion
	// act
        let (cfd, _, _, _) = cfd
            .with_lock(taker_keys, maker_keys)
            .dummy_collab_settlement_taker(opening_price);
        let result = cfd.start_rollover();
        
        // assert
        let no_rollover_reason = result.unwrap_err().downcast::<NoRolloverReason>().unwrap();
        assert_eq!(no_rollover_reason, NoRolloverReason::Closed);
```

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+    /// over the rollover. Upon finishing the rollover we fail because the cfd was already
+    /// settled.
+    #[test]
+    fn given_collab_settlement_then_cannot_finish_rollover() {
+        let quantity = Usd::new(dec!(10));
+        let leverage = Leverage::new(2).unwrap();
+        let opening_price = Price::new(dec!(10000)).unwrap();
+
+        let taker_keys = crate::keypair::new(&mut rand::thread_rng());
+        let maker_keys = crate::keypair::new(&mut rand::thread_rng());
```
thomaseizinger, line 2429: Would be nice to get rid of the keys on this level if we don't need them for anything else beyond the `with_lock` invocation.

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+    }
+
+    /// Cover scenario where trigger a collab settlement during ongoing rollover
+    ///
+    /// In this scenario the collab settlement finished before the rollover finished.
+    /// If we trigger a collaborative settlement during rollover, the settlement will have priority
+    /// over the rollover. Upon finishing the rollover we fail because the cfd was already
+    /// settled.
+    #[test]
+    fn given_collab_settlement_then_cannot_finish_rollover() {
```
```suggestion
    fn given_collab_settlement_finished_then_cannot_finish_rollover() {
```

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+        let _ = assert!(cfd.start_rollover().is_ok(), "Could not start rollover");
+
+        let cfd = cfd.dummy_start_collab_settlement();
+
+        let no_rollover_reason = cfd
+            .roll_over(RolloverCompleted::succeeded(
+                OrderId::default(),
+                Dlc::dummy(None),
+                FundingFee::dummy(),
+            ))
```
thomaseizinger: This is the "act" part of the test. Would be nice to indicate this with good use of whitespace!

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+        let leverage = Leverage::new(2).unwrap();
+        let opening_price = Price::new(dec!(10000)).unwrap();
+
+        let cfd = Cfd::taker_long()
+            .with_quantity(quantity)
+            .with_opening_price(opening_price)
+            .with_leverage(leverage)
+            .dummy_open(dummy_event_id());
+
+        let _ = assert!(cfd.start_rollover().is_ok(), "Could not start rollover");
```
thomaseizinger: I don't think this should be here according to the test desc.

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+
+        let cfd = cfd.dummy_start_collab_settlement();
+
+        let err = cfd.start_rollover().unwrap_err();
+        let no_rollover_reason = err.downcast::<NoRolloverReason>().unwrap();
+        assert_eq!(
+            no_rollover_reason,
+            NoRolloverReason::InCollaborativeSettlement
```
thomaseizinger: This is a completely separate test, isn't it? (the maker side of things)

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+            .with_id(order_id)
+            .with_quantity(quantity)
+            .with_opening_price(opening_price)
+            .with_leverage(leverage)
+            .dummy_open(dummy_event_id());
+
+        let (taker_long, proposal, taker_sig, _) = taker_long
+            .with_lock(taker_keys, maker_keys)
+            .dummy_start_rollover()
+            .dummy_collab_settlement_taker(opening_price);
```
thomaseizinger: It is pretty confusing that the "act" part of this test is chained together with the "arrange". Can we split this please?

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+    }
+
+    #[test]
+    fn given_collab_settlement_then_cannot_force_close() {
+        let quantity = Usd::new(dec!(10));
+        let leverage = Leverage::new(2).unwrap();
+        let opening_price = Price::new(dec!(10000)).unwrap();
+
+        let taker_keys = crate::keypair::new(&mut rand::thread_rng());
+        let maker_keys = crate::keypair::new(&mut rand::thread_rng());
```
thomaseizinger, line 2551: So. Much. Noise. :grimacing:

daemon/src/model/cfd.rs
```diff
@@ -2373,6 +2390,219 @@ mod tests {
...
+
+        let (cfd, _, _, _) = cfd
+            .with_lock(taker_keys, maker_keys)
+            .dummy_collab_settlement_taker(opening_price);
+
+        // TODO: Potentially change the API so we can assert the actual error
+        assert!(
+            cfd.manual_commit_to_blockchain().is_err(),
+            "Manual commit to blockchain did not error"
+        );
```
thomaseizinger: Why not `unwrap_err` and assert on the error, i.e. call `to_string` on it and assert on the reason?

thomaseizinger (dismissed):
Lots of noise to be removed but overall good effort, thanks!

I think there are a couple of occasions where we start a rollover although it shouldn't be part of the test. Would like to discuss that otherwise ACK.

daemon/src/model/cfd.rs
da-kami: it's more a sanity check that we are able to start rollover *before* we do whatever we do. But I can remove them. This goes towards "a test should only test one thing". Sometimes I find it useful to show to the reader that "at this point we can, then we do this, then we can't..." 

Besides that for some tests it was just copy paste and completely irrelevant for the test 😅
I removed all of them because I don't think they add that much value alltogether. We should potentially test that we can start rollover in separate tests.

daemon/src/model/cfd.rs
da-kami: For tests that actually do a collaborative settlement (for either role) they are important because otherwise we run into different errors. For other tests I can try to remove them.

daemon/src/model/cfd.rs
da-kami: yeah, but for *most* cases we need them outside because they have to be the same for taker and maker. I think this should go away once we get your changes for a simplified lock in that we don't build collaboratively.

daemon/src/model/cfd.rs
da-kami: Yes, but the same test context. I could split it up in `_taker` and `_maker` test but that goes towards the same discussion of being able to group.

PR #1207, opened 2022-01-24
Smash apart the workspace into more focused crates

thomaseizinger:
TODO:

- [x] Ensure each commit builds, updates lockfile correctly and is formatted
- [x] Extract test suite into dedicated crate
- [x] Extract a `tokio-tasks` crate for the `Tasks` and `tokio_ext` stuff
- [x] Extract `xtras` crate
- [x] Write up document explaining crate structure and incentives?

PR #1375, opened 2022-02-07
Extract model crate

luckysori:
Continuing the efforts started with #1207, which was inspired by #226.

The extracted `model` crate still has some suspicious dependencies such as `sqlx`, which we should strive to remove (in the future).

We also extract a `bdk-ext` crate, which is mostly used for tests.

TODO:

- [x] Fix `sqlx-data.json`.
- [x] Separate `bdk-ext` extraction into independent commit.
- [x] Ensure that there are no unused dependencies in `daemon`.
- [x] Update `bors.toml`.


payout-curve/src/lib.rs
```diff
@@ -0,0 +1,827 @@
+use anyhow::Context;
```
thomaseizinger: You seem to have attempted another crate here already? :)

bdk-ext/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "bdk-ext"
```
thomaseizinger, line 2: Can this specialise as `bdk-test` perhaps?

model/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+chrono = { version = "0.4", features = ["serde"] }
+hex = "0.4"
+itertools = "0.10"
+maia = "0.1.0"
+nalgebra = { version = "0.30", default-features = false, features = ["std"] }
+ndarray = "0.15.4"
+ndarray_einsum_beta = "0.7.0"
+num = "0.4.0"
+parse-display = "0.5.3"
+reqwest = { version = "0.11", default-features = false }
```
thomaseizinger: Why?

model/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+rust_decimal = "1.21"
+rust_decimal_macros = "1.21"
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+serde_with = { version = "1", features = ["macros"] }
+sqlx = { version = "0.5", features = ["sqlite", "runtime-tokio-rustls"] }
```
thomaseizinger, line 25: Couple of open questions regarding this:

- Should we trim the feature list here? Does that negatively (poisoned cache) or positively (less code) influence compile times?
- Can/Should we perhaps get rid of it entirely by mapping on the way out of the DB layer?

model/src/cfd.rs
```diff
@@ -1737,13 +1815,13 @@ pub enum Completed<P, E> {
     },
 }
 
-impl<P, E> xtra::Message for Completed<P, E>
```
thomaseizinger, line 1748: I think we may be able to get rid of this enum if we instead have separate commands for these:

- `setup_contract_failed`
- etc

That was the original plan at least. Certainly, we should no longer be sending this as a message so this can probably go away. Would be nice to do these cleanups in a patch at the front of the PR:

- git reset --soft HEAD^ to remove the commit but keep the changes
- git stash
- make the small patch
- git stash pop

model/src/lib.rs
```diff
@@ -29,13 +24,27 @@ use std::ops::Sub;
...
+/// Hence, this constant defines how long a cfd is open (until it gets either settled or rolled
+/// over).
+///
+/// Multiple code parts align on this constant:
+/// - How the oracle event id is chosen when creating an order (maker)
+/// - The sliding window of cached oracle announcements (maker, taker)
+/// - The auto-rollover time-window (taker)
+pub const SETTLEMENT_INTERVAL: time::Duration = time::Duration::hours(24);
+
+pub const N_PAYOUTS: usize = 200;
```
thomaseizinger: This is an application decision IMO and should maybe stay in `daemon`?

model/src/cfd.rs
```diff
@@ -1474,7 +1552,7 @@ impl Dlc {
     /// Create a close transaction based on the current contract and a settlement proposals
     pub fn close_transaction(
         &self,
-        proposal: &crate::model::cfd::SettlementProposal,
+        proposal: &crate::cfd::SettlementProposal,
```
thomaseizinger, line 1563: As a folllow-up, I think we should flatten this crate and then extract more meaningful modules. This `cfd` submodule is annoying me 🤣

model/src/cfd.rs
```diff
@@ -2982,7 +3058,9 @@ mod tests {
     }
 
     pub fn dummy_identity() -> Identity {
-        Identity::new(RandomSeed::default().derive_identity().0)
+        Identity::new(x25519_dalek::PublicKey::from(
+            *b"hello world, oh what a beautiful",
+        ))
```
thomaseizinger, line 3487: Stuff like this would be nice in a separate patch. Then the last commit could all just be "move files".

model/src/lib.rs
```diff
@@ -484,87 +493,6 @@ pub struct WalletInfo {
     pub last_updated_at: Timestamp,
 }
 
-#[derive(
-    Debug, Clone, Copy, SerializeDisplay, DeserializeFromStr, PartialEq, Eq, Hash, PartialOrd, Ord,
-)]
-pub struct BitMexPriceEventId {
```
thomaseizinger, line 489: This could be moved to the olivia module first as well to split the diff.

model/Cargo.toml
luckysori: I was unhappy about this too. We use the `Url` type. I wanted to depend on the `url` crate, but I wasn't sure if it would result in annoying incompatibilities between crates. What do you think?

thomaseizinger: I think you can depend on `url`, `reqwest` just re-exports it in its latest version.

model/src/lib.rs
luckysori: Yeah, you're right. I'll return it to `daemon` and add a similar constant in the `model` tests.

daemon/src/db.rs
```diff
@@ -389,29 +389,34 @@ mod tests {
...
+            "FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"
+                .parse()
+                .unwrap(),
+            OpeningFee::new(Amount::from_sat(2000)),
+            FundingRate::default(),
+            TxFeeRate::default(),
+        )
+    }
 
+    #[async_trait::async_trait]
```
luckysori: Separate commit?

daemon/src/oracle.rs
```diff
@@ -344,15 +340,30 @@ impl xtra::Actor for Actor {
...
+
+    pub fn into_inner(self) -> olivia::Attestation {
+        self.0
+    }
+
+    pub fn id(&self) -> BitMexPriceEventId {
+        self.0.id
+    }
+}
+
```
luckysori: Separate commit?

model/src/cfd.rs
```diff
@@ -3089,4 +3168,19 @@ mod tests {
...
+
+    pub fn new_keypair<R>(rng: &mut R) -> (secp256k1::SecretKey, bitcoin::PublicKey)
+    where
+        R: RngCore + CryptoRng,
+    {
+        let sk = secp256k1::SecretKey::new(rng);
+        let pk = bitcoin::PublicKey::new(secp256k1::PublicKey::from_secret_key(SECP256K1, &sk));
+
+        (sk, pk)
+    }
```
luckysori: Separate commit?

model/src/olivia.rs
```diff
@@ -10,10 +10,18 @@ use time::macros::format_description;
...
 pub const EVENT_TIME_FORMAT: &[FormatItem] =
     format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]");
 
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct Attestation {
+    pub id: BitMexPriceEventId,
+    pub price: u64,
+    pub scalars: Vec<SecretKey>,
+}
+
```
luckysori: Separate commit?

daemon/src/db.rs
thomaseizinger: I think we could also just solve this differently by making it a free function :)

daemon/Cargo.toml
```diff
@@ -16,10 +16,8 @@ hex = "0.4"
 hkdf = "0.12"
 itertools = "0.10"
 maia = "0.1.0"
+model = { path = "../model" }
 nalgebra = { version = "0.30", default-features = false, features = ["std"] }
```
thomaseizinger: What about `nalgebra`? Can this not be moved?

daemon/prepare_db.sh
```diff
@@ -15,4 +15,4 @@ trap 'rm -f $TEMPDB' EXIT
 DATABASE_URL=sqlite:$TEMPDB cargo sqlx migrate run
 
 # prepare the sqlx-data.json rust mappings
-DATABASE_URL=sqlite:./$DAEMON_DIR/$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare -- --bin taker
+DATABASE_URL=sqlite:./$DAEMON_DIR/$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare
```
thomaseizinger, line 18: Nit: Separate commit? :)

daemon/src/db.rs
```diff
@@ -1,20 +1,20 @@
...
-use crate::model::OpeningFee;
-use crate::model::Position;
-use crate::model::Price;
-use crate::model::TxFeeRate;
-use crate::model::Usd;
 use anyhow::Context;
 use anyhow::Result;
 use futures::future::BoxFuture;
 use futures::FutureExt;
+use model;
```
thomaseizinger: That should be unnecessary when it is a dedicated crate.

daemon/src/oracle.rs
thomaseizinger: It is motivated by creating `olivia::Attestation` right? Could move the entire creation of this dedicated model into a separate commit. We are starting to gold-plate here though :D

model/src/olivia.rs
thomaseizinger: Perhaps yeah. Ties in with what I commented above. Up to you!

thomaseizinger (approved):

thomaseizinger:
I am merging https://github.com/itchysats/itchysats/pull/1335. Don't forget to update the bors toml file until we have a check that verifies that it is up-to-date.

daemon/src/db.rs
luckysori: Done. I wanted to see if I could keep a similar API with a trait, but needing `async_trait` is just stupid.

bors.toml
```diff
@@ -11,6 +11,7 @@ status = [
   "clippy (xtras)",
   "clippy (bdk-ext)",
   "clippy (model)",
+  "clippy (xtra-bitmex-price-feed)",
```
thomaseizinger, line 14: Ups! Thank you :)

thomaseizinger (approved):

PR #1377, opened 2022-02-07
Display closing price instead liquidation price once known

da-kami:
The liquidation price becomes irrelevant once we have a closing price (i.e. once we are closing the cfd), so we display the closing price then.
Since we did not previously expose the closing price in the projection we do now.


bonomat (approved):
LGTM

PR #1379, opened 2022-02-08
Fix problem where we transition to `Open` when force closing before lock transaction reached finality

da-kami:


daemon/src/model/cfd.rs
```diff
@@ -2570,6 +2570,22 @@ mod tests {
...
+            .dummy_open(dummy_event_id())
+            .dummy_commit();
+
+        let maker_short = Cfd::maker_short()
+            .dummy_open(dummy_event_id())
+            .dummy_commit();
+
+        let taker_event = taker_long.handle_lock_confirmed();
+        let maker_event = maker_short.handle_lock_confirmed();
+        assert_eq!(taker_event.event, CfdEvent::LockConfirmedAfterFinality);
```
thomaseizinger, line 2653: Nit: Add a newline here.

thomaseizinger (approved):
Thank you!

klochowicz (approved):
Thanks, covering more corner-cases is always good :)

PR #1390, opened 2022-02-09
Respect the initial funding fee for tests where we assert on the values

da-kami:
We distinguish just creating a rummy, or creating taker_long / maker_short from an `Order` which allows us defining the initial funding rate.
Adds a test with values from production to ensure sanity.

### DIFF
--- a/daemon/src/model/cfd.rs
+++ b/daemon/src/model/cfd.rs
@@ -2254,7 +2254,7 @@ mod tests {
         // --|----|<--------------------rollover------------------->|--
         //                                                          now
 
-        let cfd = Cfd::taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
+        let cfd = Cfd::dummy_taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
             datetime!(2021-11-19 10:00:00).assume_utc(),
         ));
         let result = cfd.can_auto_rollover_taker(datetime!(2021-11-19 10:00:00).assume_utc());
@@ -2269,7 +2269,7 @@ mod tests {
         // --|----|<--------------------rollover------------------->|--
         //        now
 
-        let cfd = Cfd::taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
+        let cfd = Cfd::dummy_taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
             datetime!(2021-11-19 10:00:00).assume_utc(),
         ));
 
@@ -2285,7 +2285,7 @@ mod tests {
         // --|----|<--------------------rollover------------------->|--
         //    now
 
-        let cfd = Cfd::taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
+        let cfd = Cfd::dummy_taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
             datetime!(2021-11-19 10:00:00).assume_utc(),
         ));
         let cannot_roll_over = cfd
@@ -2302,7 +2302,7 @@ mod tests {
         // --|----|<--------------------rollover------------------->|--
         //  now
 
-        let cfd = Cfd::taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
+        let cfd = Cfd::dummy_taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
             datetime!(2021-11-19 10:00:00).assume_utc(),
         ));
         let cannot_roll_over = cfd
@@ -2319,7 +2319,7 @@ mod tests {
         // --|----|<--------------------rollover------------------->|--
         //       now
 
-        let cfd = Cfd::taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
+        let cfd = Cfd::dummy_taker_long().dummy_open(BitMexPriceEventId::with_20_digits(
             datetime!(2021-11-19 10:00:00).assume_utc(),
         ));
         let cannot_roll_over = cfd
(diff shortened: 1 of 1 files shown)


daemon/src/model/cfd.rs
```diff
@@ -2663,77 +2651,99 @@ mod tests {
...
-        );
+        // Expected payout at closing-price-interval defined by payout curve
+        let expected_payout_taker = Amount::from_sat(49583);
+        let expected_payout_maker = Amount::from_sat(100247);
+        // Expected initial funding fee based on the funding rate and short-margin (because the rate
+        // is positive meaning long pays short)
+        let expected_initial_funding_fee = Amount::from_sat(10);
+
+        // Note that transaction fees ARE zero for these tests because the order's fee-rate is set
+        // to 0!
```
thomaseizinger: Wouldn't it be better to set it so a non-zero value? (Can be done in a follow-up ofc).

daemon/src/model/cfd.rs
```diff
@@ -2663,77 +2651,99 @@ mod tests {
...
             closing_price,
             positive_funding_rate,
-            1000,
             0,
+            0,
+        );
+        assert_eq!(
+            taker_payout,
+            expected_payout_taker - expected_initial_funding_fee
+        );
```
thomaseizinger: Can we please split this test?

This would be much easier to read if these variables were declared right above the assertion. Esp. with the tests further down, having to scroll up and down to understand what these values are makes it really hard to understand.

thomaseizinger: It would also make the tests a lot more concise because we could inline certain variables in the function call to `collab_settlement_taker_long_maker_short` and reserve the "arrange" part of the test for the values that you really care about. This test could for example be:

```rust
    #[test]
    fn given_collab_close_at_same_price_with_no_rollovers_pays_funding_fee_to_maker() {
        let opening_price = Price::new(dec!(10000)).unwrap();
        let closing_price = opening_price;
	let positive_funding_rate = dec!(0.0001);
	let num_rollovers = 0;

        let (taker_payout, maker_payout) = collab_settlement_taker_long_maker_short(
            Usd::new(dec!(10)),
            opening_price,
            closing_price,
            positive_funding_rate,
            0,
            num_rollovers,
        );

        let long_pays_short_funding_fee = 10;
        assert_eq!(taker_payout, Amount::from_sat(49583 - long_pays_short_funding_fee));
        assert_eq!(maker_payout, Amount::from_sat(100247 + long_pays_short_funding_fee));
}
```

thomaseizinger: We could also make `collab_settlement_taker_long_maker_short` return a `u64` so we can remove the noise of creating `Amount`s.

thomaseizinger (approved):

daemon/src/model/cfd.rs
da-kami: I still think that the test should be one because it is about different number of rollovers. But since this keeps bothering you more than me and it has been coming up twice I'll split them. The initial values will be duplicated in each test then.

da-kami: @bonomat @klochowicz do you have an opinion about this?

bonomat: > @bonomat @klochowicz do you have an opinion about this?

I don't have strong feelings about this.

thomaseizinger: > The initial values will be duplicated in each test then.

There are ways to mitigate this :)

> But since this keeps bothering you more than me

It is less about bothering me or you. There are good reasons for it! Some background knowledge about testing:

The goal to always strive for with tests is:

- **one** statement for arrange
- **one** expression for act
- **one** statement for assert

Having more than one statement for arrange is fairly tolerable.
Having more than one expression for act is more smelly but tolerable if it just takes multiple (production code) calls to get something done. Can easily be mitigate though with something like you did above where you extract a helper for the test
Having more than one assert is tolerable IF they are a single block although ideally, we can assert the behaviour on a single outcome.

Multiple asserts per tests, intermixed with even more "act"s has numerous downsides:

1. An assert will early exit the test, meaning if a test fails (on CI for example) it is basically impossible to know, whether the other asserts would have passed. This is valuable information when debugging a problem to be able to narrow down, where the problem is. If you have one assert per test, you will learn things like: "Interesting, when doing no rollovers, things break but 2 rollovers are fine?".
2. An assert always states an expectation and a test is a form of documentation, how the system behaves. If you have more than one assert per test, it takes a good chunk focus and brainpower, to understand what part of the system is actually being tested here!

klochowicz: I feel like we went on a tangent a bit with the tests:

As for tests, I personally find them often much harder to read after refactoring into the goal above - so, at least as the downside 2) I feel that my brainpower is fragmented in trying to simultaneously parse complicated setup functions & similar things scattered across different tests.
As with many other things, I perceive it as a trade-off - and I'd pick readability, speed & convenience vs a bit easier time debugging (the debugging can be quickly narrowed down if you re-run the test with debugger / stdout and look at it, so I don't think it's a big issue for us!)

daemon/src/model/cfd.rs
da-kami: When starting to get `TxFeeRate` into the tests I noticed that we are actually not taking transaction fees into account for the payout calculation. Should the output of the transaction include this? If I configure the test with a `TxFeeRate` then the amount in the collab settlement tx remains the same.

da-kami: When do the transaction fees come into affect? Would we have to extract the payout amount differently from the transaction so fees are taken into account?

da-kami: After discussion, here is my understanding:
Collab close tx fee is hardcoded to `1 sat/vbyte` resulting in `85` sats per party to be paid. 
The lock tx fee is irrelevant because it is paid on top of the lock amount when publishing lock.
We construct lock together and when it is build `amount + fees` is constructed, where the `fee` is collaboratively paid by both parties.

I tried to reflect the settlement tx fee with a constant to make this clearer.

thomaseizinger (approved):
Thank you!

PR #1391, opened 2022-02-09
Accurately describe Cfd state in the projection

klochowicz:
See commit messages for more details.

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -325,6 +325,37 @@ async fn maker_rejects_rollover_of_open_cfd() {
 }
 
 #[tokio::test]
+async fn maker_rejects_rollover_after_commit_finality() {
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement()).await;
+
+    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    next_with(taker.quote_feed(), |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
+
+    taker.trigger_rollover(order_id).await;
+
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
+        CfdState::IncomingRolloverProposal,
+        CfdState::OutgoingRolloverProposal
+    );
+
+    confirm!(commit transaction, order_id, maker, taker);
+    // Cfd would be in "OpenCommitted" if it wasn't for the rollover
+
+    maker.system.reject_rollover(order_id).await.unwrap();
+
+    // After rejecting rollover, we should display where we were before the
+    // rollover attempt
+    wait_next_state!(order_id, maker, taker, CfdState::OpenCommitted);
+}
+
+#[tokio::test]
 async fn open_cfd_is_refunded() {
     let _guard = init_tracing();
     let oracle_data = OliviaData::example_0();
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -195,6 +195,14 @@ struct Aggregated {
 
     commit_published: bool,
     refund_published: bool,
+
+    /// Keep track of persistent state in case a protocol fails and we need to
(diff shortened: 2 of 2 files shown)


daemon/src/projection.rs
```diff
@@ -226,6 +237,34 @@ impl Aggregated {
...
+        // Rollover is applied after collab settlement for the UI, as it's a
+        // time sensitive and short-lived transition
+        if let Some(rollover_state) = self.rollover_state {
+            state = match rollover_state {
+                ProtocolNegotiationState::Started => match role {
+                    Role::Maker => CfdState::IncomingRolloverProposal,
+                    Role::Taker => CfdState::OutgoingRolloverProposal,
+                },
+                ProtocolNegotiationState::Accepted => CfdState::ContractSetup,
+            }
```
thomaseizinger: Could do this first and just back out of the function early?

daemon-tests/tests/happy_path.rs
```diff
@@ -324,6 +324,37 @@ async fn maker_rejects_rollover_of_open_cfd() {
...
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
+        CfdState::IncomingRolloverProposal,
+        CfdState::OutgoingRolloverProposal
+    );
+
+    deliver_event!(maker, taker, Event::CommitFinality(order_id));
+    // If it wasn't for the collab settlement, it would be "OpenCommitted"
```
thomaseizinger: What does this comment refer to?

daemon/src/projection.rs
klochowicz: makes sense - I could do that, and if I return from both if lets I don't need to have the `mut` variable either!

daemon-tests/tests/happy_path.rs
```diff
@@ -324,6 +324,37 @@ async fn maker_rejects_rollover_of_open_cfd() {
     wait_next_state!(order_id, maker, taker, CfdState::Open);
 }
 
+#[tokio::test]
+async fn maker_rejects_rollover_after_commit_finality() {
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement()).await;
```
thomaseizinger, line 332: Should we inline this if we don't need the variable again?

thomaseizinger (approved):

daemon-tests/tests/happy_path.rs
klochowicz: I'm actually aiming at storing an open cfd in a DB for the tests, because this is the biggest contributor for our CI / local testing to take so long! it's like 1 min per test at least... In other words, it will soon go away 😁

daemon-tests/tests/happy_path.rs
klochowicz: oops, I started with the other protocol, then changed my mind, and forgot to update the comment.... fixed!

daemon-tests/tests/happy_path.rs
thomaseizinger: Interesting, looking forward to it!

I think we should also pair this with benchmarking. If there are a few simple optimisations we can do to cut down the setup time, then I'd take that over a persistent test DB I think.

klochowicz: the biggest difference seem to be compiler optimisations, e.g. 10s sqlx query in rust debug buids. I agree that we benchmarking should be done & implemented, but I think persistent test DB would trim these numbers even further anyway (+ would allow us to paint a more accurate picture of failure, we already covered this pathway by a separate test, and redoing it for other tests gives no value)

da-kami (approved):
This implementation might have unwanted side effects. At the moment it can happen that a rollover / settlement is started but not finished (we never add failed).
In such a scenario I think the state would remain in rollover / settlement forever in the projection, because we never set it to `None` again. 

I am no sure we can easily solve this problem in the projection, but in order for this logic to not lock the UI in a state that is incorrect, I think we will have to tackle our housekeeping task to ensure that we always complete a protocol. 
I'm still happy to merge this as it is an improvement, but want to raise awareness of this behaviour.

thomaseizinger:
> but want to raise awareness of this behaviour.

See https://github.com/itchysats/architecture/pull/7.

PR #1200, opened 2022-01-24
Make process_manager::Actor use send_async_safe

thomaseizinger:

PR #1395, opened 2022-02-09
Trigger CET finality monitoring using process manager

### LINKED ISSUE #1147: Monitor actor do not follow ADR010 for monitoring CET upon attestation
[ADR010](https://github.com/itchysats/itchysats/blob/master/docs/adr/010-process-manager-for-orchestration.md) states that orchestration of behaviour following domain events should happen through the process manager.

The monitor actor does not follow this decision as it gets directly notified by the oracle actor about a new attestation. Instead, the monitor actor should get told by the process manager to monitor for a CET when the respective events happen, i.e. once we decrypted the CET and know which one will be used.

### LINKED ISSUE #1148: Oracle actor does not follow ADR013 for handling new attestation
[ADR013](https://github.com/itchysats/itchysats/blob/master/docs/adr/013-all-actors-can-hydrate-cfd.md) states that a CFD should be loaded and hydrated wherever needed.

The oracle actor does not follow this as it sends the `Attestation` to the `maker_cfd` / `taker_cfd` actor. Instead, it should use the `command::Executor` to call the `decrypt_cet` command on the corresponding CFD.

The difficulty with making this happen is that our current test framework relies on sending the attestation directly to the `maker_cfd` and `taker_cfd` actor. This will have to be changed to interact with the oracle mock instead.

rishflab:
We can trigger CET finality monitoring in the process manager when the oracle attestation events occur. This removes the need for a channel between oracle and monitor actor.

NOTE: Will probably squash all commits! Probably not worth reviewing it commit by commit as rebase may of messed things up.

### DIFF
--- a/daemon-tests/src/mocks/oracle.rs
+++ b/daemon-tests/src/mocks/oracle.rs
@@ -1,7 +1,8 @@
 use crate::maia::OliviaData;
 use async_trait::async_trait;
+use daemon::command;
 use daemon::oracle;
-use mockall::*;
+use model::cfd::OrderId;
 use model::olivia;
 use model::olivia::BitMexPriceEventId;
 use std::sync::Arc;
@@ -16,8 +17,8 @@ pub struct OracleActor {
 }
 
 impl OracleActor {
-    pub fn new() -> (Self, Arc<Mutex<MockOracle>>) {
-        let mock = Arc::new(Mutex::new(MockOracle::new()));
+    pub fn new(executor: command::Executor) -> (Self, Arc<Mutex<MockOracle>>) {
+        let mock = Arc::new(Mutex::new(MockOracle::new(executor)));
         let actor = Self { mock: mock.clone() };
 
         (actor, mock)
@@ -30,7 +31,6 @@ impl xtra::Actor for OracleActor {
 
     async fn stopped(self) -> Self::Stop {}
 }
-impl Oracle for OracleActor {}
 
 #[xtra_productivity(message_impl = false)]
 impl OracleActor {
@@ -38,33 +38,40 @@ impl OracleActor {
         &mut self,
         msg: oracle::GetAnnouncement,
     ) -> Result<olivia::Announcement, oracle::NoAnnouncement> {
-        self.mock.lock().await.get_announcement(msg)
+        self.mock
+            .lock()
+            .await
+            .announcement
+            .clone()
+            .ok_or(oracle::NoAnnouncement(msg.0))
     }
 
-    async fn handle(&mut self, msg: oracle::MonitorAttestation) {
-        self.mock.lock().await.monitor_attestation(msg)
-    }
+    async fn handle(&mut self, _msg: oracle::MonitorAttestation) {}
 
-    async fn handle(&mut self, msg: oracle::Sync) {
-        self.mock.lock().await.sync(msg)
-    }
+    async fn handle(&mut self, _msg: oracle::Sync) {}
+}
+pub struct MockOracle {
+    executor: command::Executor,
+    announcement: Option<olivia::Announcement>,
 }
 
-#[automock]
-pub trait Oracle {
-    fn get_announcement(
-        &mut self,
-        _msg: oracle::GetAnnouncement,
-    ) -> Result<olivia::Announcement, oracle::NoAnnouncement> {
-        unreachable!("mockall will reimplement this method")
+impl MockOracle {
+    fn new(executor: command::Executor) -> Self {
+        Self {
+            executor,
+            announcement: None,
+        }
     }
 
-    fn monitor_attestation(&mut self, _msg: oracle::MonitorAttestation) {
-        unreachable!("mockall will reimplement this method")
+    pub async fn simulate_attestation(&mut self, id: OrderId, attestation: oracle::Attestation) {
+        self.executor
+            .execute(id, |cfd| cfd.decrypt_cet(&attestation.into_inner()))
+            .await
+            .unwrap();
     }
 
-    fn sync(&mut self, _msg: oracle::Sync) {
-        unreachable!("mockall will reimplement this method")
+    pub fn set_announcement(&mut self, announcement: olivia::Announcement) {
+        self.announcement = Some(announcement);
     }
 }
 
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -2,7 +2,6 @@ use daemon::bdk::bitcoin::Amount;
 use daemon::connection::ConnectionStatus;
 use daemon::projection::CfdOrder;
 use daemon::projection::CfdState;
-use daemon_tests::deliver_event;
 use daemon_tests::dummy_new_order;
 use daemon_tests::dummy_quote;
 use daemon_tests::flow::is_next_none;
@@ -13,6 +12,7 @@ use daemon_tests::flow::one_cfd_with_state;
 use daemon_tests::init_tracing;
 use daemon_tests::maia::OliviaData;
 use daemon_tests::mocks::oracle::dummy_wrong_attestation;
+use daemon_tests::simulate_attestation;
 use daemon_tests::start_both;
 use daemon_tests::wait_next_state;
 use daemon_tests::Maker;
@@ -196,9 +196,6 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
     maker.mocks.mock_party_params().await;
     taker.mocks.mock_party_params().await;
 
-    maker.mocks.mock_oracle_monitor_attestation().await;
-    taker.mocks.mock_oracle_monitor_attestation().await;
-
     maker.mocks.mock_wallet_sign_and_broadcast().await;
     taker.mocks.mock_wallet_sign_and_broadcast().await;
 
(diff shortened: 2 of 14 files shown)


daemon/src/process_manager.rs
```diff
@@ -119,8 +124,22 @@ impl Actor {
                     })
                     .await?;
             }
-            OracleAttestedPostCetTimelock { cet, .. }
-            | CetTimelockExpiredPostOracleAttestation { cet } => {
+            CetTimelockExpiredPostOracleAttestation { cet } => {
```
thomaseizinger, line 127: Why not monitor the CET in this case?

daemon/src/process_manager.rs
```diff
@@ -119,8 +124,22 @@ impl Actor {
...
+                    .send_async_safe(monitor::TryBroadcastTransaction {
+                        tx: cet,
+                        kind: TransactionKind::Cet,
+                    })
+                    .await?;
+            }
+            OracleAttestedPostCetTimelock { cet, .. } => {
+                let _ = self
+                    .monitor_cet_finality
+                    .send(MonitorCetFinality {
```
thomaseizinger: Please use `send_async_safe` here. For why, please see https://github.com/itchysats/itchysats/pull/1200, especially the last commit.

daemon/src/oracle.rs
```diff
@@ -252,14 +259,37 @@ impl Actor {
...
+                    .send(process_manager::Event::new(event.clone()))
+                    .await?
+                {
+                    tracing::error!("Sending event to process manager failed: {:#}", e);
+                }
+            }
+        }
+
         let _: Result<(), xtra::Disconnected> =
             self.attestation_channel.send_async_safe(attestation).await;
```
thomaseizinger: If you are already calling `decrypt_cet` above, sending the attestation through this channel _as well_ ends up calling `decrypt_cet` a second time. Remember that this channel is really just the `{maker,taker}_cfd::Actor` and all they are doing with the attestation is calling the `cfd_actors::handle_oracle_attestation` which does exactly the loop above.

However, simply removing this channel is what causes tests to break (because they rely on being able to send an attestation to the `{maker,taker}_cfd::Actor`. (That is what https://github.com/itchysats/itchysats/issues/1148 is about.)

From the description, I take it that your intent was to do https://github.com/itchysats/itchysats/issues/1147 first. For that, this code doesn't need to change and this change should go into your other PR!

thomaseizinger (changes requested):
Some comments. Overall a step a in the right direction but I think there was a slight mixup between the two issues! :)

daemon-tests/src/mocks/oracle.rs
```diff
@@ -30,41 +31,53 @@ impl xtra::Actor for OracleActor {
...
+        _msg: oracle::GetAnnouncement,
     ) -> Result<olivia::Announcement, oracle::NoAnnouncement> {
-        self.mock.lock().await.get_announcement(msg)
+        let announcement = self
+            .mock
+            .lock()
+            .await
+            .announcement
+            .clone()
+            .expect("announcement has been set");
```
thomaseizinger: Why panic when we can return an error?

daemon-tests/src/mocks/oracle.rs
```diff
@@ -30,41 +31,53 @@ impl xtra::Actor for OracleActor {
...
+    fn new(executor: command::Executor) -> Self {
+        Self {
+            executor,
+            announcement: None,
+        }
     }
 
-    fn monitor_attestation(&mut self, _msg: oracle::MonitorAttestation) {
-        unreachable!("mockall will reimplement this method")
+    pub async fn decrypt_cet(&mut self, id: OrderId, attestation: oracle::Attestation) {
```
thomaseizinger: I find the name of this function a bit strange. Can we call this `simulate_attestation` instead?

daemon-tests/tests/happy_path.rs
```diff
@@ -261,13 +280,15 @@ async fn force_close_an_open_cfd() {
     // After CetTimelockExpired, we're only waiting for attestation
     expire!(cet timelock, order_id, maker, taker);
 
-    // Delivering the wrong attestation does not move state to `PendingCet`
-    deliver_event!(maker, taker, dummy_wrong_attestation());
+    // Decrypting the wrong attestation does not move state to `PendingCet`
+    decrypt_cet(&mut maker, &mut taker, order_id, dummy_wrong_attestation()).await;
```
thomaseizinger: Any particular reason we are moving away from using a macro for this? The macro makes it a bit more concise because you don't need to `await` it and the mutable reference is not visible.

daemon/src/oracle.rs
```diff
@@ -56,11 +55,11 @@ struct NewAnnouncementFetched {
     nonce_pks: Vec<schnorrsig::PublicKey>,
 }
 
-/// A module-private message to allow parallelization of fetching attestations.
-#[derive(Debug)]
-struct NewAttestationFetched {
+/// A message to allow parallelization of fetching attestations.
+#[derive(Debug, Clone)]
+pub struct NewAttestationFetched {
```
thomaseizinger: Why is this suddenly public?

daemon/src/oracle.rs
```diff
@@ -186,11 +185,12 @@ impl Actor {
                         .await
                         .context("Failed to deserialize as Attestation")?;
 
-                    this.send(NewAttestationFetched {
-                        id: event_id,
-                        attestation: Attestation(attestation),
-                    })
-                    .await?;
+                    let _ = this
```
thomaseizinger: Why are we dropping this error here?

daemon/src/process_manager.rs
```diff
@@ -141,10 +182,17 @@ impl Actor {
...
-                commit_tx: None, ..
+                commit_tx: None,
+                timelocked_cet: cet,
+                ..
             } => {
-                // Nothing to do: The commit transaction has already been published but the timelock
-                // hasn't expired yet. We just need to wait.
+                let _ = self
+                    .monitor_cet_finality
+                    .send(MonitorCetFinality {
```
thomaseizinger: `send_async_safe` please.

thomaseizinger (commented):
Some minor improvements / questions but overall this looks good, thank you!

daemon-tests/src/mocks/oracle.rs
rishflab: yep makes sense

daemon-tests/tests/happy_path.rs
rishflab: I like plain functions because they work better with IDE tooling and are easier to read if they have to be modified.

the trade off here is concise vs easier-to-work-on and I guess my preference is easier-to-work-on but happy to revert back to a macro

daemon/src/oracle.rs
rishflab: Its used in the tests so I had to make it pub

daemon/src/oracle.rs
rishflab: woops, must of been because the function now returns an error

daemon-tests/src/mocks/oracle.rs
rishflab: I need to return `NoAnnouncement` Error here but i dont have the `BitmexPriceId`. I think it should be ok to panic since it is a test mock.

thomaseizinger: The ID is in the message that you are handling here.

daemon-tests/tests/happy_path.rs
thomaseizinger: Yes please keep the macro for consistency with the current code.

daemon/src/oracle.rs
thomaseizinger: Yes, you added the handler the test but that one is not actually needed. I just tried it locally and it can be deleted without problems! :)

We only send this message to ourselves so it doesn't really make sense that anyone else needs to know about this message!

daemon/src/oracle.rs
```diff
@@ -245,14 +244,26 @@ impl Actor {
...
-            self.attestation_channel.send_async_safe(attestation).await;
+        let mut conn = self.db.acquire().await?;
+
+        for id in db::load_all_cfd_ids(&mut conn).await? {
+            if let Err(err) = self
+                .executor
+                .execute(id, |cfd| cfd.decrypt_cet(&attestation))
+                .await
+            {
+                tracing::warn!("Failed to decrypt CET using attestation: {}", err)
```
thomaseizinger:
```suggestion
                tracing::warn!(order_id = %id, "Failed to decrypt CET using attestation: {}", err)
```

This should probably have the ID.

daemon/src/oracle.rs
```diff
@@ -187,7 +187,7 @@ impl Actor {
 
                     this.send(NewAttestationFetched {
                         id: event_id,
-                        attestation,
+                        attestation: Attestation(attestation),
```
rishflab, line 190: This is to reduce the diff

daemon/src/oracle.rs
```diff
@@ -254,7 +254,7 @@ impl Actor {
         for id in db::load_all_cfd_ids(&mut conn).await? {
             if let Err(err) = self
                 .executor
-                .execute(id, |cfd| cfd.decrypt_cet(&attestation))
+                .execute(id, |cfd| cfd.decrypt_cet(&attestation.0))
```
rishflab, line 257: also to reduce the diff

thomaseizinger (approved):
Thanks!

Squash and go :)

daemon/src/process_manager.rs
```diff
@@ -103,11 +103,11 @@ impl Actor {
                 match self.role {
                     Role::Maker => {
                         self.try_broadcast_transaction
-                            .send(monitor::TryBroadcastTransaction {
+                            .send_async_safe(monitor::TryBroadcastTransaction {
```
thomaseizinger, line 106: Nice, thanks for fixing this!

PR #1397, opened 2022-02-10
Move ADRs to different repository

thomaseizinger:


klochowicz:
What about decisions that only concern one of the projects? like just the automation, just the backtesting, just the itchysats main repo?

klochowicz (approved):
I don't mind them being moved as it makes sense to have them separate if the decisions are global. However,  it would be good to keep around the timestamps etc so I wouldn't move them around too often :) (or used git subtree command to graft the parts)

thomaseizinger:
> What about decisions that only concern one of the projects? like just the automation, just the backtesting, just the itchysats main repo?

I like the appeal of all of these being in one place. In reality, we will make these decisions within the context of all the previous ones so having them all in one spot makes that easier.

thomaseizinger:
> I don't mind them being moved as it makes sense to have them separate if the decisions are global. However, it would be good to keep around the timestamps etc so I wouldn't move them around too often :) (or used git subtree command to graft the parts)

If you want to, knock yourself out. I don't think the commit history of these is particularly useful so I didn't bother :)

klochowicz:
> > I don't mind them being moved as it makes sense to have them separate if the decisions are global. However, it would be good to keep around the timestamps etc so I wouldn't move them around too often :) (or used git subtree command to graft the parts)
> 
> If you want to, knock yourself out. I don't think the commit history of these is particularly useful so I didn't bother :)

at this stage it would probably not worthwhile, they were written in short succession by a small amount of people :-) If we end up doing it again, I volunteer for doing it with a subtree!

PR #1405, opened 2022-02-11
Add timeout on sending TCP packets from taker

klochowicz:
An attempt at mitigating production issue where the connection message
handler got stuck for a prolonged period of time.

Unfortunately, it's not easy to cover it with unit tests so the fix could not be
verified.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -354,7 +354,10 @@ impl Actor {
         };
 
         let our_version = Version::current();
-        write.send(TakerToMaker::Hello(our_version.clone())).await?;
+        write
+            .send(TakerToMaker::Hello(our_version.clone()))
+            .timeout(Duration::from_secs(10))
+            .await??;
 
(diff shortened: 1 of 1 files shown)


thomaseizinger:
Thanks!

I don't think we need a PR for this atm, better to branch of a `release/0.4.3` branch and just put that commit on top of it (i.e. just rename the current branch).

Once we have all the fixes for `release/0.4.3` in place, we can tag the HEAD of that branch for the actual release (also need to bump the version number in the manifest for that).

klochowicz:
> Thanks!
> 
> I don't think we need a PR for this atm, better to branch of a `release/0.4.3` branch and just put that commit on top of it (i.e. just rename the current branch).
> 
> Once we have all the fixes for `release/0.4.3` in place, we can tag the HEAD of that branch for the actual release (also need to bump the version number in the manifest for that).

OK, I'll do that!

PR #1406, opened 2022-02-11
Add timeout on sending TCP packets from taker

### LINKED ISSUE #1401: Make sure the taker actually reconnects and does not get stuck in reconnecting
1. Taker Hotfix on top of 0.4.2  
	1. Add timeouts to all statements related to establishing the connection
	1. Add unit test to test things
	1. Introduce some randomness so that not all takers re-connect at exactly the same point in time

klochowicz:
An attempt at mitigating production issue where the connection message
handler got stuck for a prolonged period of time.

Unfortunately, it's not easy to cover it with unit tests so the fix could not be
verified.

Partially fixes #1401

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -355,7 +355,10 @@ impl Actor {
         };
 
         let our_version = Version::current();
-        write.send(TakerToMaker::Hello(our_version.clone())).await?;
+        write
+            .send(TakerToMaker::Hello(our_version.clone()))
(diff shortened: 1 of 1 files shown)


klochowicz:
I've pushed this change to `origin/release/0.4.3`, but it still needs to get in to the master (and other refactors are coming to it too).

thomaseizinger (approved):

thomaseizinger:
Lets hope it helps 😁

PR #1410, opened 2022-02-13
Add missing timeout to noise initiator handshake

### LINKED ISSUE #1401: Make sure the taker actually reconnects and does not get stuck in reconnecting

klochowicz:
Fixes #1401. Will need to be backported to `0.4.3`.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -42,6 +42,8 @@ use xtras::SendInterval;
 /// Time between reconnection attempts
 pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
 
+const TCP_TIMEOUT: std::time::Duration = Duration::from_secs(10);
+
 /// The "Connected" state of our connection with the maker.
 #[allow(clippy::large_enum_variant)]
 enum State {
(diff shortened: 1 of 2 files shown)


daemon/src/connection.rs
```diff
@@ -611,13 +616,16 @@ pub async fn connect(
                 }
 
                 let num_addresses = maker_addresses.len();
-                let seconds = CONNECT_TO_MAKER_INTERVAL.as_secs();
+
+                // Apply a random number of seconds between the reconnection
+                // attempts so that all takers don't try to reconnect at the same time
+                let seconds = thread_rng().gen_range(5, MAX_RECONNECT_INTERVAL_SECONDS);
```
thomaseizinger, line 622: Wondering if a smaller jitter + exponential backoff with a max of like 5minutes wouldn't be better. But perhaps that would be over-optimising.

thomaseizinger (approved):

daemon/src/connection.rs
da-kami, line 622: Given that it might take some time to the daemon to reconnect to the maker we could think about showing more information (e.g. the seconds) in the UI.

da-kami (approved):

daemon/src/connection.rs
klochowicz: happy to consider it as a part of further changes on the master branch (the one here is indended for backporting, so wanted to keep it as small as possible)

klochowicz: interesting idea! will consider it for the follow-up (for backporting, I'd prefer not to touch the UI)

PR #1411, opened 2022-02-14
Display Cfd tiles in reverse chronoligical order

### LINKED ISSUE #1328: The order of the CFD tiles display in the UI is not always ensured
We added an `order by` when loading from the db at some point

https://github.com/itchysats/itchysats/blob/79fa34250b237f7b0971c29f39a508734a9646e1/daemon/src/db.rs#L267-L277

but that does not seem to be enough anymore with the recent changes of sending the cfds loaded by id async:

https://github.com/itchysats/itchysats/blob/eaad7c3bfaf3c28c7dda70ed53153c40aae3fef8/daemon/src/projection.rs#L731-L733

I think we have to ensure the sorting in the UI - but for that we need an indicator that allows sorting chronologically; i.e. we need to expose the CFD's opening timestamp.

rishflab:
Confirmed that the Cfd tiles are displayed in a random order on master.

Tested PR by creating bunch of orders on a local taker/maker and checking they were in reverse chronological order after multiple restarts

Close #1328

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -11,6 +11,7 @@ use bdk::bitcoin::Transaction;
 use bdk::bitcoin::Txid;
 use bdk::miniscript::DescriptorTrait;
 use core::fmt;
+use itertools::Itertools;
 use maia::TransactionExt;
 use model::calculate_funding_fee;
 use model::cfd::calculate_long_liquidation_price;
@@ -168,6 +169,9 @@ pub struct Cfd {
 
     #[serde(skip)]
     aggregated: Aggregated,
+
+    #[serde(skip)]
+    creation_timestamp: Timestamp,
 }
 
 /// Bundle all state extracted from the events in one struct.
@@ -352,6 +356,7 @@ impl Cfd {
             counterparty: counterparty_network_identity,
             pending_settlement_proposal_price: None,
             aggregated: Aggregated::new(fee_account),
+            creation_timestamp: Timestamp::now(),
         }
     }
 
@@ -361,6 +366,7 @@ impl Cfd {
         match event.event {
             ContractSetupStarted => {
                 self.aggregated.state = CfdState::ContractSetup;
+                self.creation_timestamp = event.timestamp;
             }
             ContractSetupCompleted { dlc } => {
                 self.expiry_timestamp = Some(dlc.settlement_event_id.timestamp());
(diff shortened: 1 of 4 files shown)


daemon/src/projection.rs
```diff
@@ -316,6 +321,7 @@ impl Cfd {
         match event.event {
             ContractSetupStarted => {
                 self.state = CfdState::ContractSetup;
+                self.creation_timestamp = event.timestamp;
```
thomaseizinger, line 369: I guess this is ok for now, alternatively, we could get the timestamp of the first event. Would need to check the version of the aggregate for that.

daemon/src/projection.rs
```diff
@@ -667,6 +673,7 @@ impl Tx {
         let cfds_with_quote = cfds
             .into_iter()
             .map(|(_, cfd)| cfd.with_current_quote(quote))
+            .sorted_by(|a, b| Ord::cmp(&a.creation_timestamp, &b.creation_timestamp))
```
thomaseizinger: Did you verify that this sorts descending, i.e. latest timestamp first?

thomaseizinger (commented):
Concept ACK

daemon/src/projection.rs
rishflab: its the other way, ill swap it

daemon/src/projection.rs
rishflab: Wouldn't this event be the first event (that counts) all the time. `OfferRejected` is the only other event that could come before this but It won't show in the tiles.

thomaseizinger: > Wouldn't this event be the first event (that counts) all the time. `OfferRejected` is the only other event that could come before this but It won't show in the tiles.

Yep, that is correct. It is a question of how to best express the intent.

If we have a bit of code that does:

```rust
if self.version == 0 {
	self.creation_timestamp = event.timestamp;
}
```

Then this expresses more clearly that we are defining the concept of the creation timestamp as the timestamp of the first event.
If we tie the logic to the specific event `ContractSetupStarted`, then the reader has to know that this is the first event to understand why that logic is there and not somewhere else.

I don't mind much either way :)

thomaseizinger (approved):
This works, thanks!

@da-kami Do you think we should remove the `order by` statement from the SQL query again now that we are fixing this in projection?

daemon/src/projection.rs
da-kami: I think it is more resilient to express this with `self.version == 0` and not use `ContractSetupStarted`.

thomaseizinger: I think I agree. @rishflab do you mind doing this in a follow-up? Also I think we now can/should remove the "ORDER BY" statement here: https://github.com/itchysats/itchysats/blob/78cae83c66ff2b7f9f7cb8f10019c7baaec2e04c/daemon/src/db.rs#L274

PR #1418, opened 2022-02-15
Exclude final CFDs from processing

da-kami:
For auto-rollover and monitoring final CFDs are irrelevant and should not be processed.
We see a lot of warnings because we currently process all CFDs all the time.
This simple SQL statements reduces this problem by just not loading final CFDs for processing.
We only load all CFDs in the projection.

This change is very non-intrusive. We don't need to change anything on the model or anything else. 
I think this is a change that will help with multiple issues (warnings upon startup, long startup times, ...), processing monitoring and rollovers.

We *can* obviously decide to optimize loading CFDs as well, but I see that as a completely orthogonal thing to tackle.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -283,6 +283,52 @@ pub async fn load_all_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<O
     Ok(ids)
 }
 
+/// Loads all CFDs where we are still able to append events
+///
+/// This function is to be called when we only want to process CFDs where events can still be
+/// appended, but ignore all other CFDs.
+/// Open in this context means that the CFD is not final yet, i.e. we can still append events.
+/// In this context a CFD is not open anymore if one of the following happened:
+/// 1. Event of the confirmation of a payout (spend) transaction on the blockchain was recorded
+///     Cases: Collaborative settlement, CET, Refund
+/// 2. Event that fails the CFD early was recorded, meaning it becomes irrelevant for processing
+///     Cases: Setup failed, Taker's take order rejected
+pub async fn load_open_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<OrderId>> {
+    let ids = sqlx::query!(
+        r#"
+            select
+                id as cfd_id,
+                uuid as "uuid: model::cfd::OrderId"
+            from
+                cfds
+            where not exists (
+                select id from EVENTS as events
+                where cfd_id = cfds.id and
+                (
+                    events.name = $1 or
+                    events.name = $2 or
+                    events.name= $3 or
+                    events.name= $4 or
+                    events.name= $5
+                )
+            )
+            order by cfd_id desc
+            "#,
+        CfdEvent::COLLABORATIVE_SETTLEMENT_CONFIRMED,
+        CfdEvent::CET_CONFIRMED,
+        CfdEvent::REFUND_CONFIRMED,
+        CfdEvent::CONTRACT_SETUP_FAILED,
+        CfdEvent::OFFER_REJECTED
+    )
+    .fetch_all(&mut *conn)
+    .await?
+    .into_iter()
+    .map(|r| r.uuid)
+    .collect();
+
+    Ok(ids)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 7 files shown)


da-kami:
Sidenote: We could decide to *not* treat `SetupFailed` as final, because it can happen at the moment, that one party succeeds with a contract setup while the other party does not. I am not sure we handle state transitions for that correctly anyway, so I included it for now. I don't think we should optimize for this edge case, because it is unexpected.

daemon/src/db.rs
```diff
@@ -283,6 +283,56 @@ pub async fn load_all_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<O
...
 }
 
+/// Loads cfd that are not final yet
+///
+/// Finality can happen by:
+/// 1. Recording the confirmation of a payout (spend) transaction on the blockchain
+///     Cases: Collaborative settlement, CET, Refund
+/// 2. The CFD failing early meaning it becomes irrelevant for processing
+///     Cases: Setup failed, Taker's take order rejected
+pub async fn load_non_final_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<OrderId>> {
```
thomaseizinger: Any reason we use the wording "non_final" here instead of "open"?

daemon/src/db.rs
```diff
@@ -380,6 +430,121 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn given_collaborative_close_confirmed_then_do_not_load_non_final_cfd() {
+        let mut conn = setup_test_db().await;
+
+        let cfd_not_final = insert(dummy_cfd(), &mut conn).await;
+        append_event(lock_confirmed(&cfd_not_final), &mut conn)
+            .await
+            .unwrap();
```
thomaseizinger: Do we need to re-test this in every test?

I think these would be more concise if we had 1 test that ensures a non-final CFD is included and 5 tests that show, which "final" CFDs are excluded.

daemon/src/db.rs
```diff
@@ -283,6 +283,56 @@ pub async fn load_all_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<O
...
+///     Cases: Collaborative settlement, CET, Refund
+/// 2. The CFD failing early meaning it becomes irrelevant for processing
+///     Cases: Setup failed, Taker's take order rejected
+pub async fn load_non_final_cfd_ids(conn: &mut PoolConnection<Sqlite>) -> Result<Vec<OrderId>> {
+    let (collaborative_settlement_confirmed, _) =
+        CfdEvent::CollaborativeSettlementConfirmed.to_json();
+    let (cet_confirmed, _) = CfdEvent::CetConfirmed.to_json();
+    let (refund_confirmed, _) = CfdEvent::RefundConfirmed.to_json();
+    let (setup_failed, _) = CfdEvent::ContractSetupFailed.to_json();
+    let (rejected, _) = CfdEvent::OfferRejected.to_json();
```
thomaseizinger: I think I'd prefer if we have these names as constants somewhere and reference them directly instead of computing this every time. We did this already for the caching, have a look at that!

thomaseizinger (approved):

daemon/src/db.rs
da-kami: I just tried to be more precise because we use `Open` in a very ambiguous way. E,g. in the projection we can have multiple states after `Open`, until we finally reach a final state.

da-kami: I am happy to change it to `Open` if you think it helps...?

daemon/src/db.rs
da-kami: I think it is super cheap here because non of these events actually has data, but happy to use constants, that's even better.

daemon/src/db.rs
thomaseizinger: I think I'd prefer `Open`. Other opinions?

klochowicz: IMHO "non_final" is a bit better due to issue outlined by @da-kami earlier - there are multiple states around "Open" (some *before* and some *after*, which makes it even more confusing), the current naming is a bit more unambiguous.
At the end of the day though, we're in a different context here, so one could plausibly ignore projection naming here (on the database level there are no "states").

model/src/cfd.rs
```diff
@@ -340,6 +337,14 @@ pub enum CfdEvent {
 }
 
 impl CfdEvent {
+    pub const CONTRACT_SETUP_COMPLETED_EVENT: &'static str = "ContractSetupCompleted";
+    pub const ROLLOVER_COMPLETED_EVENT: &'static str = "RolloverCompleted";
+    pub const COLLABORATIVE_SETTLEMENT_CONFIRMED: &'static str = "CollaborativeSettlementConfirmed";
+    pub const CET_CONFIRMED: &'static str = "CetConfirmed";
+    pub const REFUND_CONFIRMED: &'static str = "RefundConfirmed";
+    pub const CONTRACT_SETUP_FAILED: &'static str = "ContractSetupFailed";
+    pub const OFFER_REJECTED: &'static str = "OfferRejected";
```
klochowicz, line 346: nice, arguably a bit nicer place for these consts in this giant file! 👍

klochowicz (approved):

daemon/src/db.rs
da-kami: In the end I opted for renaming it to `open` to avoid introducing yet another terminology. 
I adapted the rustdoc of the function to be very specific what `open` means in this context.

I think our discussion shows that we should cleanup the states in the projection API because it adds confusion :)

thomaseizinger: > I think our discussion shows that we should cleanup the states in the projection API because it adds confusion :)

Yes please <3

PR #1419, opened 2022-02-15
Upgrade maia to include latest performance improvements

thomaseizinger:
This version of maia includes performance fixes that make
contract-setup 3x as fast!


klochowicz:
> This version of maia includes performance fixes that make contract-setup 3x as fast!

wow, the CI times have improved indeed! :)

klochowicz (approved):

PR #1424, opened 2022-02-16
Allow maker going long

klochowicz:
Automation / GUI should be able to create the offers going the other way from
now on.

Add tests covering most common scenarios: close, force-close,
rollover (contract setup is covered as part of each of them).

TODO: Allow simultaneous offers both ways

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -24,6 +24,7 @@ use model::cfd::calculate_long_margin;
 use model::cfd::OrderId;
 use model::olivia;
 use model::Identity;
+use model::Position;
 use model::Usd;
 use rust_decimal_macros::dec;
 use std::time::Duration;
@@ -120,7 +121,7 @@ async fn taker_receives_order_from_maker_on_publication() {
 
     assert!(is_next_none(taker.order_feed()).await.unwrap());
 
-    maker.publish_order(dummy_new_order()).await;
+    maker.publish_order(dummy_new_order(Position::Short)).await;
 
     let (published, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
@@ -150,7 +151,7 @@ async fn taker_takes_order_and_maker_rejects() {
     // TODO: Why is this needed? For the cfd stream it is not needed
     is_next_none(taker.order_feed()).await.unwrap();
 
-    maker.publish_order(dummy_new_order()).await;
+    maker.publish_order(dummy_new_order(Position::Short)).await;
 
     let (_, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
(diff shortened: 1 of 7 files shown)


da-kami:
Pushed a quick fixup for fixing clippy. We might wanna change the commit message :D

klochowicz:
> Pushed a quick fixup for fixing clippy. We might wanna change the commit message :D

lol, indeed :D I was rushing to post it before my training!

daemon-tests/tests/happy_path.rs
```diff
@@ -211,93 +212,109 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
...
+    let _guard = init_tracing();
+    collaboratively_close_an_open_cfd(Position::Long).await;
+}
 
+async fn collaboratively_close_an_open_cfd(maker_position: Position) {
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(OliviaData::example_0().announcement(), maker_position).await;
     taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
     maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    next_with(taker.quote_feed(), |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
```
thomaseizinger, line 221: Personally, I prefer in-line comments being on the line they are meant for. All too often, code is inserted between a code line and a comment :)

daemon-tests/tests/happy_path.rs
```diff
@@ -211,93 +212,109 @@ async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
...
-    let (mut maker, mut taker, order_id) =
-        start_from_open_cfd_state(OliviaData::example_0().announcement()).await;
+    collaboratively_close_an_open_cfd(Position::Short).await;
+}
+
+#[tokio::test]
+async fn collaboratively_close_an_open_cfd_maker_going_long() {
+    let _guard = init_tracing();
+    collaboratively_close_an_open_cfd(Position::Long).await;
+}
```
thomaseizinger, line 224: Really? This just works? haha

I am suspicuous!

thomaseizinger (approved):

daemon-tests/tests/happy_path.rs
klochowicz: yes, we are suspicious too! that's why we wanted to allow further testing with the automation ASAP! But all the common scenarios have worked so far!

daemon-tests/tests/happy_path.rs
klochowicz: ouch, that was rust-analyzer function extraction refactor! perhaps it's safer to put them above sometimes - I agree that they shouldn't be put *below*, that's just bonkers. will move them back.

klochowicz:
As discussed in the daily, more work needs to be done for it to function properly, but this is a step in the right direction, especially that it doesn't break anything.

PR #1435, opened 2022-02-17
Taker leveraged short

da-kami:
The leverage is now handled correctly, internally we store long and short leverage and calcuate the margin accordingly. In the database we only store the taker leverage (that's what is in the `Order` and is currently always 2). The maker leverage is always assumed to be `1` no matter if the maker is short or long.

Note that the `Order` should be changed to be an `Offer`, and the position of said offer should *not* be the maker's (as it currently is) but the taker's. It is super confusing that the Order's position is the maker's whereas the order's leverage is the one of the taker.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -15,10 +15,10 @@ use itertools::Itertools;
 use maia::TransactionExt;
 use model::calculate_funding_fee;
 use model::calculate_long_liquidation_price;
-use model::calculate_long_margin;
+use model::calculate_margin;
 use model::calculate_profit;
 use model::calculate_profit_at_price;
-use model::calculate_short_margin;
+use model::long_and_short_leverage;
 use model::CfdEvent;
 use model::Dlc;
 use model::EventKind;
@@ -119,7 +119,9 @@ pub struct Cfd {
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
     pub accumulated_fees: SignedAmount,
 
-    pub leverage: Leverage,
+    /// The taker leverage
+    #[serde(rename = "leverage")]
+    pub leverage_taker: Leverage,
     pub trading_pair: TradingPair,
     pub position: Position,
     #[serde(with = "round_to_two_dp")]
@@ -292,7 +294,7 @@ impl Cfd {
             id,
             position,
             initial_price,
-            leverage,
+            taker_leverage,
             quantity_usd,
             counterparty_network_identity,
             role,
@@ -301,19 +303,27 @@ impl Cfd {
             ..
         }: db::Cfd,
     ) -> Self {
-        let long_margin = calculate_long_margin(initial_price, quantity_usd, leverage);
-        let short_margin = calculate_short_margin(initial_price, quantity_usd);
-
-        let (margin, margin_counterparty) = match position {
-            Position::Long => (long_margin, short_margin),
-            Position::Short => (short_margin, long_margin),
+        let (our_leverage, counterparty_leverage) = match role {
+            Role::Maker => (Leverage::ONE, taker_leverage),
+            Role::Taker => (taker_leverage, Leverage::ONE),
         };
-        let liquidation_price = calculate_long_liquidation_price(leverage, initial_price);
+
+        let margin = calculate_margin(initial_price, quantity_usd, our_leverage);
+        let margin_counterparty =
+            calculate_margin(initial_price, quantity_usd, counterparty_leverage);
+
+        // TODO: Change this to reflect that we can be short! - does this make sense for short? What
+        // should the taker UI display?
+        let liquidation_price = calculate_long_liquidation_price(our_leverage, initial_price);
+
+        let (long_leverage, short_leverage) =
+            long_and_short_leverage(taker_leverage, role, position);
 
         let initial_funding_fee = calculate_funding_fee(
             initial_price,
             quantity_usd,
-            leverage,
+            long_leverage,
+            short_leverage,
             initial_funding_rate,
             SETTLEMENT_INTERVAL.whole_hours(),
         )
@@ -333,7 +343,7 @@ impl Cfd {
             order_id: id,
             initial_price,
             accumulated_fees: fee_account.balance(),
-            leverage,
+            leverage_taker: taker_leverage,
             trading_pair: TradingPair::BtcUsd,
             position,
             liquidation_price,
@@ -569,11 +579,15 @@ impl Cfd {
             }
         };
 
+        let (long_leverage, short_leverage) =
+            long_and_short_leverage(self.leverage_taker, self.role, self.position);
+
         let (profit_btc, profit_percent, payout) = match calculate_profit_at_price(
             self.initial_price,
             latest_price,
             self.quantity_usd,
-            self.leverage,
+            long_leverage,
+            short_leverage,
             self.aggregated.fee_account,
         ) {
             Ok((profit_btc, profit_percent, payout)) => {
@@ -879,7 +893,9 @@ pub struct CfdOrder {
     pub id: OrderId,
 
     pub trading_pair: TradingPair,
-    pub position: Position,
+
+    #[serde(rename = "position")]
+    pub position_maker: Position,
 
     /// The maker's price for opening a position
     #[serde(with = "round_to_two_dp")]
@@ -920,7 +936,8 @@ pub struct CfdOrder {
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
     pub initial_funding_fee_per_parcel: Amount,
 
-    pub leverage: Leverage,
+    #[serde(rename = "leverage")]
+    pub leverage_taker: Leverage,
     #[serde(with = "round_to_two_dp")]
     pub liquidation_price: Price,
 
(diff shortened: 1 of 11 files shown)


klochowicz:
I guess you can rebase it already on master, the other one got merged!

da-kami:
> I guess you can rebase it already on master, the other one got merged!

done.

model/src/cfd.rs
```diff
@@ -1361,23 +1382,30 @@ impl AsBlocks for Duration {
...
 ///
-/// The short margin is represented as the quantity of the contract given the
-/// initial price. The short side can currently not leverage the position but
-/// always has to cover the complete quantity.
-pub fn calculate_short_margin(price: Price, quantity: Usd) -> Amount {
-    quantity / price
+/// The margin is the initial margin and represents the collateral both parties have to come up with
+/// to satisfy the contract. Here we calculate the initial margin = quantity / (initial_price *
+/// leverage)
+pub fn calculate_margin(price: Price, quantity: Usd, leverage: Leverage) -> Amount {
```
da-kami, line 1361: I assume this is correct for long and short margins. @bonomat would be cool if you could have a look

daemon/src/projection.rs
```diff
@@ -119,6 +119,8 @@ pub struct Cfd {
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
     pub accumulated_fees: SignedAmount,
 
+    // TODO: This is currently always the taker leverage! This should be clear
+    /// The taker leverage
     pub leverage: Leverage,
```
luckysori: 🙃 You could rename the struct field to `taker_leverage` and use `#[serde(rename = "leverage")]` if you don't want to break the API.

model/src/lib.rs
```diff
@@ -153,6 +152,7 @@ impl InversePrice {
     }
 }
 
+// TODO: Consider representating leverage as own and counterparty right here
```
luckysori: 🙃 A TODO that starts with the word "consider" is a strong candidate for becoming stale imo. I say this because I've done this myself.

model/src/cfd.rs
```diff
@@ -129,13 +129,23 @@ pub struct Order {
     pub id: OrderId,
 
     pub trading_pair: TradingPair,
+
+    /// The maker's position
+    ///
+    /// Since the maker is the creator of this order this always reflects the maker's position.
+    /// When we create a Cfd we change it to be the position as seen by each party, i.e. flip the
+    /// position for the taker.
```
luckysori, line 140: I appreciate the documentation but I still think using `serde rename` would help here (as I commented elsewhere) and below.

model/src/cfd.rs
```diff
@@ -1361,23 +1382,30 @@ impl AsBlocks for Duration {
...
 
-// Make a `Margin` newtype and call `Margin::long`
-/// Calculates the long's margin in BTC
-///
-/// The margin is the initial margin and represents the collateral the buyer
-/// has to come up with to satisfy the contract. Here we calculate the initial
-/// long margin as: quantity / (initial_price * leverage)
-pub fn calculate_long_margin(price: Price, quantity: Usd, leverage: Leverage) -> Amount {
-    quantity / (price * leverage)
+/// Detemine the leverage based on role and position
```
luckysori: 🔧 
```suggestion
/// Determine the leverage based on role and position
```

model/src/cfd.rs
```diff
@@ -1361,23 +1382,30 @@ impl AsBlocks for Duration {
...
+            (Leverage::ONE, taker_leverage)
+        }
+        (Role::Maker, Position::Short) | (Role::Taker, Position::Long) => {
+            (taker_leverage, Leverage::ONE)
+        }
+    }
 }
 
-/// Calculates the shorts's margin in BTC
+// Make a `Margin` newtype and call `Margin::long`
```
luckysori: ❓ Is this meant to be a TODO?

model/src/cfd.rs
```diff
@@ -1361,23 +1382,30 @@ impl AsBlocks for Duration {
...
+/// Calculates the margin in BTC
 ///
-/// The short margin is represented as the quantity of the contract given the
-/// initial price. The short side can currently not leverage the position but
-/// always has to cover the complete quantity.
-pub fn calculate_short_margin(price: Price, quantity: Usd) -> Amount {
-    quantity / price
+/// The margin is the initial margin and represents the collateral both parties have to come up with
+/// to satisfy the contract. Here we calculate the initial margin = quantity / (initial_price *
+/// leverage)
```
luckysori: 🔧 Don't think we should duplicate the formula used in the code in the docstring. It's redundant and could easily get stale.

model/src/cfd.rs
```diff
@@ -1361,23 +1382,30 @@ impl AsBlocks for Duration {
...
-/// Calculates the shorts's margin in BTC
+// Make a `Margin` newtype and call `Margin::long`
+/// Calculates the margin in BTC
 ///
-/// The short margin is represented as the quantity of the contract given the
-/// initial price. The short side can currently not leverage the position but
-/// always has to cover the complete quantity.
-pub fn calculate_short_margin(price: Price, quantity: Usd) -> Amount {
-    quantity / price
+/// The margin is the initial margin and represents the collateral both parties have to come up with
```
luckysori: 🔧 
```suggestion
/// The initial margin represents the collateral both parties have to come up with
```

"The margin is the initial margin" is a bit redundant imo.

luckysori (commented):
A few small comments.

model/src/lib.rs
da-kami: Uh, that's actually an outdated one that should be removed! If anything, we should pack long/short and not own/counterparty in there :D

model/src/cfd.rs
thomaseizinger: Looks at least equivalent to me :)

model/src/lib.rs
```diff
@@ -166,6 +165,10 @@ impl Leverage {
     pub fn get(&self) -> u8 {
         self.0
     }
+
+    pub const ONE: Self = Self { 0: 1 };
```
thomaseizinger: Why this weird syntax of using a named field syntax when you have a tuple struct?

```suggestion
    pub const ONE: Self = Self(1);
```

That doesn't work?

thomaseizinger (approved):

klochowicz (approved):

da-kami:
Fun fact: My changes actually hardcoded the `long_leverage` for both sides, because the `model::Cfd::leverage()` (the getter for leverage) always returned `long_leverage` 😅 - So when we write into the db we write whatever long is for both.
Fixed that by adapting `model::Cfd::leverage()`. I wonder if it would be better to store `own_leverage` internally to make that more clear?

Note that previously we stored only the taker leverage in the db (for maker and taker), because we needed that for the payout curve (which is always long). Now, we have the `long_leverage`, so I am storing the own leverage in the db which makes more sense. I don't think that should have side effects, but we will need more testing...

Did not find time to add tests yet.

model/src/cfd.rs
da-kami: It is, but I wonder if that is correct in the new context :D

da-kami:
> Note that previously we stored only the taker leverage in the db (for maker and taker), because we needed that for the payout curve (which is always long). Now, we have the `long_leverage`, so I am storing the own leverage in the db which makes more sense. I don't think that should have side effects, but we will need more testing...

Note for Fixthat Tomorrow Daniel: The constructor still expects the `taker_leverage`, so that's what has to go into the db.

da-kami:
Some notes:

In the past we always saved the taker leverage into the db. I don't really want to change the db for mainnet-stable, so I'd say we keep that. 
I thought I could change it to always save our own leverage (so the db field would make more sense...), but then old positions would not be loaded correctly, because we always expect the taker leverage in the constructor of `model::Cfd`. I would like to change that constructor to just take `long_leverage` and `short_leverage`, but given backwards compatibility (...) this is not easy to achieve. 

So, for now I opted to just save the taker leverage into the db - as before - but be more explicit about it in code (I did not rename the db field though).
At least that's the conclusion I had - if you think we can solve this better at this point please let me know.

thomaseizinger:
> I don't really want to change the db for mainnet-stable, so I'd say we keep that.

Writing a migration that only changes the `cfds` table is easy enough. We could definitely do that!

klochowicz:
> > I don't really want to change the db for mainnet-stable, so I'd say we keep that.
> 
> Writing a migration that only changes the `cfds` table is easy enough. We could definitely do that!

yes!

are we talking about `maker_leverage` field? this one's easy, because we can default to 1 in the migration script!

da-kami:
> > I don't really want to change the db for mainnet-stable, so I'd say we keep that.
> 
> Writing a migration that only changes the `cfds` table is easy enough. We could definitely do that!

True, but it's not so much about renaming the field, but about storing what makes more sense... The fact that we only store the taker leverage is just weird - but sufficient for not (and it can depict configurable taker leverage, which we might want to tackle on top of mainnet stable).
Once we go multiple makers I think the model should properly depict long and short leverage in the db.

klochowicz:
> > > I don't really want to change the db for mainnet-stable, so I'd say we keep that.
> > 
> > 
> > Writing a migration that only changes the `cfds` table is easy enough. We could definitely do that!
> 
> True, but it's not so much about renaming the field, but about storing what makes more sense... The fact that we only store the taker leverage is just weird - but sufficient for not (and it can depict configurable taker leverage, which we might want to tackle on top of mainnet stable). Once we go multiple makers I think the model should properly depict long and short leverage in the db.

From the db perspective, I'd opt for storing `taker_leverage` (renamed from `leverage` in migration) and `maker_leverage` (defaulted to 1). then the model can read as as they please, as we know the origin. IMHO it will be less confusing, because of consistency with old model and the fact that "maker" and "taker" are 1st class concepts on the database level (ie. we for sure know who the db belongs to).

da-kami:
Note that when creating a leveraged `short` position we panic at the moment:

```
thread 'rocket-worker-thread' panicked at 'attempt to subtract with overflow', model/src/payout_curve.rs:115:30
stack backtrace:
   0: rust_begin_unwind
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/02072b482a8b5357f7fb5e5637444ae30e423c40/library/core/src/panicking.rs:48:5
   3: model::payout_curve::calculate_payout_parameters::{{closure}}

...
```

TODOs:

- Make sure the actor tests catch this, why are we not getting a test failure there?
- Adapt the API for using the payout curve to be able to handle leveraged short. @luckysori  is already working on that.

note: I will still merge this work because it does not cause any trouble at the moment.

model/src/cfd.rs
bonomat: I think this is correct

Issue #1430, opened 2022-02-17
Clean-up model crate a bit


thomaseizinger:
1. Flatten module structure (get rid of `cfd.rs` module)
2. Extract more sensible modules
3. Move tests out into "integration tests" so we can organise those into different files

thomaseizinger:
@luckysori Can we close this as a result of #1439 and continue that work in a different issue?

luckysori:
Sure, let's do that!

PR #1439, opened 2022-02-17
Extract more meaningful modules from `model::cfd`

luckysori:
I set out to start tackling #1430, but I got carried away and started trying to remove things that didn't make sense any longer to me (such as the `Completed` enum) rather than just moving them around.

I intend to continue with #1430 in follow-up PRs. If reviewers already have some ideas for meaningful modules, that would be helpful, although I would action that feedback in the follow-up PRs.


model/src/hex_transaction.rs
```diff
@@ -0,0 +1,50 @@
+use super::*;
+use bdk::bitcoin;
+use bdk::bitcoin::Transaction;
+use serde::Deserializer;
+use serde::Serializer;
+
```
thomaseizinger, line 6: We should usptream this.

daemon/src/setup_maker.rs
```diff
@@ -117,24 +115,45 @@ impl Actor {
 
         self.tasks.add(async move {
             let _: Result<(), xtra::Disconnected> = match contract_future.await {
-                Ok(dlc) => this.send(SetupSucceeded { order_id, dlc }).await,
-                Err(error) => this.send(SetupFailed { order_id, error }).await,
+                Ok(dlc) => this.send(SetupSucceeded { dlc }).await,
+                Err(error) => this.send(SetupFailed { error }).await,
```
thomaseizinger, line 119: Tiny nit: This could have been a separate commit.

thomaseizinger (approved):
Nice work!

klochowicz (approved):
great refactor! thanks for giving much needed TLC to the `model` <3

rishflab (approved):
Thanks, will be much easier to navigate the codebase !

model/src/hex_transaction.rs
luckysori: Tracking this locally.

PR #1452, opened 2022-02-18
Delay broadcast of refund tx on startup

### LINKED ISSUE #1230: Prioritise CETs over `RefundTransaction` on restart
Consider the following situation:

1. Parties open the CFD.
2. Either party starts force-close, publishing the `CommitTransaction`.
3. Both parties go offline.
4. Olivia publishes relevant attestation.
5. The `refund_timelock` on the `CommitTransaction` expires.
6. Either party comes back online.

Assuming that the attestation correctly unlocks one of the CETs, both the unlocked CET and the `RefundTransaction` will be accepted by the miners.  In this situation, I think our software should prioritise the CET, since the `RefundTransaction` is just a fallback mechanism when all else fails.

Recent testing has revealed that we are unintentionally prioritising the `RefundTransaction`. We're probably too eager to publish the `RefundTransaction` as soon as we notice the `refund_timelock` expiry, not giving ourselves enough time to find the attestation, unlock the CET and publish it. This theory seems to be backed by the state of the database:

```
4500d180-c56c-4bea-bab3-d35959036152|1642731266|CommitConfirmed
4500d180-c56c-4bea-bab3-35959036152|1642737793|OracleAttestedPriorCetTimelock

// restart

4500d180-c56c-4bea-bab3-35959036152|1643257379|OracleAttestedPriorCetTimelock
4500d180-c56c-4bea-bab3-d35959036152|1643257392|RefundTimelockExpired
4500d180-c56c-4bea-bab3-d35959036152|1643257392|CetTimelockExpiredPostOracleAttestation
4500d180-c56c-4bea-bab3-d35959036152|1643257718|RefundConfirmed
```

We just happen to process the expiry of the `refund_timelock` first.

rishflab:
Close #1230

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -19,6 +19,7 @@ use model::CfdEvent;
 use model::Dlc;
 use model::EventKind;
 use model::OrderId;
+use model::Timestamp;
 use model::CET_TIMELOCK;
 use serde_json::Value;
 use sqlx::SqlitePool;
@@ -134,6 +135,7 @@ pub struct Actor {
     tasks: Tasks,
     state: State,
     db: sqlx::SqlitePool,
+    initialized_at: Option<Timestamp>,
 }
 
 /// Internal data structure encapsulating the monitoring state without performing any IO.
@@ -300,6 +302,7 @@ impl Actor {
             state: State::new(BlockHeight::try_from(latest_block)?),
             tasks: Tasks::default(),
             db,
+            initialized_at: None,
         })
     }
 }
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Would it make more sense to just delay the entire `ReinitMonitoring` process?

That gives the oracle actor time to modify the CFD in case the attestation has been processed and we simply start monitoring a bit later.

rishflab:
> Would it make more sense to just delay the entire `ReinitMonitoring` process?
> 
> That gives the oracle actor time to modify the CFD in case the attestation has been processed and we simply start monitoring a bit later.

I am not sure of the implications of delaying all monitoring. I can look into it. I went with small fix to avoid breaking stuff.

thomaseizinger:
> > Would it make more sense to just delay the entire `ReinitMonitoring` process?
> > That gives the oracle actor time to modify the CFD in case the attestation has been processed and we simply start monitoring a bit later.
> 
> I am not sure of the implications of delaying all monitoring. I can look into it. I went with small fix to avoid breaking stuff.

I cannot think of any severe consequences in delaying all monitoring on startup but that doesn't mean there aren't any. However, delaying all is a simpler change to reason about than what we have here so I would prefer that.

Opinions @klochowicz @da-kami ?

rishflab:
> > > Would it make more sense to just delay the entire `ReinitMonitoring` process?
> > > That gives the oracle actor time to modify the CFD in case the attestation has been processed and we simply start monitoring a bit later.
> > 
> > 
> > I am not sure of the implications of delaying all monitoring. I can look into it. I went with small fix to avoid breaking stuff.
> 
> I cannot think of any severe consequences in delaying all monitoring on startup but that doesn't mean there aren't any. However, delaying all is a simpler change to reason about than what we have here so I would prefer that.
> 
> Opinions @klochowicz @da-kami ?

Hmm I think we can get away with delaying `ReInitMonitoring` since it is a only a 5 second delay. Cant see any severe implications either.

Pushed #1470 which implements the solution @thomaseizinger proposed.

PR #1454, opened 2022-02-18
Write proptest to test the limits of `payout_curve` module

luckysori:
The intention is to define ranges of values for `price`, `n_contracts`, `taker_leverage`, `n_payouts` and `fee_flow` which
produce sane payouts.

After defining this `proptest` we have noticed that:

- Our payout curve cannot handle an opening price over ~340_000 USD.   Values above that can run into a `Cannot invert singular matrix` error.
- The fee flow cannot be negative (i.e. `ShortPaysLong`), or else we run into an overflow error.

The former can be seen as a limitation of the current implementation. We can probably live with this for the foreseeable future.

The latter is a bug which is solved in https://github.com/itchysats/itchysats/pull/1454/commits/e3a83f8bcd6d8be9250dc2cc1f5af39d315bdf32.

### DIFF
--- a/model/src/payout_curve.rs
+++ b/model/src/payout_curve.rs
@@ -474,6 +474,7 @@ fn create_long_payout_function(
 mod tests {
     use super::*;
     use bdk::bitcoin::Amount;
+    use proptest::prelude::*;
     use rust_decimal_macros::dec;
     use std::ops::RangeInclusive;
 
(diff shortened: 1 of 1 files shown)


luckysori:
Okay, introducing the _fee_ is what causes overflows, @da-kami.

model/src/payout_curve.rs
```diff
@@ -820,6 +821,83 @@ mod tests {
...
+                .zip(payouts.iter().skip(1))
+                .map(|(a, b)| {
+                    (
+                        (a.long_amount + a.short_amount),
+                        (b.long_amount + b.short_amount),
+                    )
+                })
+                .all(|(a, b)| (a as i64 - b as i64).abs() <= 1);
+
+            prop_assert!(are_payout_totals_similar)
```
thomaseizinger, line 862: So with this you are asserting that given a window of two ranges, the payouts are the same?

luckysori: Basically that contiguous payouts are close. Obviously not what I originally intended because I wanted to use `all_equal`. How would you define an `all_almost_equal` combinator?

thomaseizinger: The definition looks good to me, I was just surprised that we want to test this property here!

luckysori: I suppose we don't need an assertion. Originally I set out to verify that we were not panicking or failing when generating the payouts.

This assertion proved useful though, since one of the error cases saw us overflowing and generating some payouts with insanely high values.

thomaseizinger: Another one could be that we are always outputting the same sum of sats.

luckysori: > Another one could be that we are always outputting the same sum of sats.

I think that's what I was trying to approximate with the assertion above.

model/src/payout_curve.rs
```diff
@@ -105,27 +104,35 @@ fn calculate_payout_parameters(
...
-                FeeFlow::Nein => long_amount,
+            // We use `saturating_sub` when deducting fees because the
+            // adjusted payout cannot go below zero. If the original
+            // payout is close or equal to zero and the fee is
+            // sufficiently large we would overflow otherwise.
+            let (short_amount, long_amount) = match fee {
+                FeeFlow::LongPaysShort(fee) => {
+                    let long_minus_fee = long_amount.saturating_sub(fee.as_sat());
+                    let long_fee_deduction = (long_amount as i64) - (long_minus_fee as i64);
+                    let short_plus_fee = (short_amount as i64) + long_fee_deduction;
```
da-kami, line 117: We do a saturating sub so we have a bound for the lower end (cannot go below zero), but we don't have an upper bound when adding fees. 

Don't we have to cap the maximum payout to `long_margin` + `short_margin` here? 
I feel with this change it would be possible to have a higher payout then the actual amount available.

luckysori: Math checks out for me:

```markdown
pool = long + short

long_minus_fee = max(long - fee, 0)                     // capped because we can't go below 0

effective_fee = long - long_minus_fee
long_minus_fee = long - effective_fee                   // after rearranging

short_plus_fee = short + effective_fee

pool = long_minus_fee + short_plus_fee 
     = long - effective_fee + short + effective_fee     // substituting
     = long + short                                     // cancelling
```

At the end we get the same equation we started with.

thomaseizinger (approved):

model/src/payout_curve.rs
da-kami: You are right! - I did not think it through correctly. 
I'll do it one more time in words :)

In a `LongPaysShort` scenario the fee goes from long to short, i.e. the fee will always be deducted from the long margin. This is what `long_minus_fee` represents, which is bounded to `[0,long_amount-1sat]`.

That means `long_fee_deduction` is bounded to `[long_amount,1sat]`.
That means `short_plus_fee` is bounded to `[short_amount+long_amount,short_amount+1sat]`.

Checks out :)

da-kami (approved):

PR #1455, opened 2022-02-18
Create another offer if one gets taken

klochowicz:
See details in commit messages.

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -3,7 +3,7 @@ use daemon::connection::ConnectionStatus;
 use daemon::connection::MAX_RECONNECT_INTERVAL_SECONDS;
 use daemon::projection::CfdOrder;
 use daemon::projection::CfdState;
-use daemon_tests::dummy_new_order;
+use daemon_tests::dummy_offer_params;
 use daemon_tests::dummy_quote;
 use daemon_tests::flow::is_next_none;
 use daemon_tests::flow::next;
@@ -121,7 +121,9 @@ async fn taker_receives_order_from_maker_on_publication() {
 
     assert!(is_next_none(taker.order_feed()).await.unwrap());
 
-    maker.publish_order(dummy_new_order(Position::Short)).await;
+    maker
+        .set_offer_params(dummy_offer_params(Position::Short))
+        .await;
 
     let (published, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
@@ -154,7 +156,9 @@ async fn taker_takes_order_and_maker_rejects() {
     // TODO: Why is this needed? For the cfd stream it is not needed
     is_next_none(taker.order_feed()).await.unwrap();
 
-    maker.publish_order(dummy_new_order(Position::Short)).await;
+    maker
+        .set_offer_params(dummy_offer_params(Position::Short))
+        .await;
 
     let (_, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
@@ -183,7 +187,9 @@ async fn another_offer_is_automatically_created_after_taker_takes_order() {
     // TODO: Why is this needed? For the cfd stream it is not needed
     is_next_none(taker.order_feed()).await.unwrap();
 
-    maker.publish_order(dummy_new_order(Position::Short)).await;
+    maker
+        .set_offer_params(dummy_offer_params(Position::Short))
+        .await;
 
     let (_, received) = next_order(maker.order_feed(), taker.order_feed())
         .await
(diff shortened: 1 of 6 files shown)


daemon-tests/tests/happy_path.rs
```diff
@@ -175,6 +175,37 @@ async fn taker_takes_order_and_maker_rejects() {
     wait_next_state!(received.id, maker, taker, CfdState::Rejected);
 }
 
+#[tokio::test]
+async fn another_offer_is_automatically_created_aftertaker_takes_order() {
```
thomaseizinger:
```suggestion
async fn another_offer_is_automatically_created_after_taker_takes_order() {
```

daemon-tests/tests/happy_path.rs
```diff
@@ -175,6 +175,37 @@ async fn taker_takes_order_and_maker_rejects() {
...
+    let _guard = init_tracing();
+    let (mut maker, mut taker) = start_both().await;
+
+    // TODO: Why is this needed? For the cfd stream it is not needed
+    is_next_none(taker.order_feed()).await.unwrap();
+
+    maker.publish_order(dummy_new_order(Position::Short)).await;
```
thomaseizinger: We might want to rename this to something like `set_offer_parameters`?

thomaseizinger (approved):

daemon-tests/tests/happy_path.rs
klochowicz: sounds good, changed this and also changed `NewOrder` to `OfferParams`

thomaseizinger (approved):
I like this change but it might be controversial so please gather some more approvals before merging :)

klochowicz:
> I like this change but it might be controversial so please gather some more approvals before merging :)

no worries! will wait till Monday, where I'll mention it on the daily.

For other reviewers, a short summary: 
 - this change adjusts the mechanism of creating the offers in the maker
 - instead of manually creating "each" offer, we essentially send a feed of latest parameters from which offers are constructed
 - this change is a preparation for the plan of sending `price_long` as one more parameter (a price for the offers where makers goes long), which will allow for simultaneous creation of offers "the other way". 
 - as a side effect, we'll be able to pump more offers with the params we want more quickly 😃 

a bit more on future direction:
- I'd like to explore allowing setting the `price_long` and `price_short` as `Option<>`, so that the maker could have a means to choose whether they want to go both long and short, just one way, or none (and just allow rolling over currently active CFDs).

thomaseizinger:
I think this goes back on ADR003. Can you please draft a short ADR that explains your reasoning?

bonomat (approved):
Let's give this a try +1

klochowicz:
> I think this goes back on ADR003. Can you please draft a short ADR that explains your reasoning?

I had a re-read of ADR003 and I don't think it goes against it. The same amount of control of the business logic will be available (including stopping new orders from happening, when I introduce `Option<Price>` shortly). is there a specific part that raises your concerns?

edit: perhaps more broadly, I understand this change like a difference in "ergonomics" of using the API, not as a change of the business logic. exactly the same strategies are available to whoever operates the maker.

thomaseizinger:
> > I think this goes back on ADR003. Can you please draft a short ADR that explains your reasoning?
> 
> I had a re-read of ADR003 and I don't think it goes against it. The same amount of control of the business logic will be available (including stopping new orders from happening, when I introduce `Option<Price>` shortly). is there a specific part that raises your concerns?
> 
> edit: perhaps more broadly, I understand this change like a difference in "ergonomics" of using the API, not as a change of the business logic. exactly the same strategies are available to whoever operates the maker.

The decision logic for sending out an offer changes. Previously it was "every time a new offer gets created on the API" and now it is "every time an offer gets taken". It doesn't contract the original ADR but what we are doing now was not the original idea so I think it is worth clarifying that. It is not a big decision but it would be good to record it.

It is also a learning for the future in how such a system should/could be designed.

klochowicz:
> > > I think this goes back on ADR003. Can you please draft a short ADR that explains your reasoning?
> > 
> > 
> > I had a re-read of ADR003 and I don't think it goes against it. The same amount of control of the business logic will be available (including stopping new orders from happening, when I introduce `Option<Price>` shortly). is there a specific part that raises your concerns?
> > edit: perhaps more broadly, I understand this change like a difference in "ergonomics" of using the API, not as a change of the business logic. exactly the same strategies are available to whoever operates the maker.
> 
> The decision logic for sending out an offer changes. Previously it was "every time a new offer gets created on the API" and now it is "every time an offer gets taken". It doesn't contract the original ADR but what we are doing now was not the original idea so I think it is worth clarifying that. It is not a big decision but it would be good to record it.
> 
> It is also a learning for the future in how such a system should/could be designed.

I take your point in a valuable learning opportunity. I've drafted a quick ADR here:
https://github.com/itchysats/architecture/pull/14

Will merge this in the meantime as I consider that I got enough buy-in to try this approach (that will unblock taker going short too).

PR #1465, opened 2022-02-20
Allow skipping one heartbeat in taker UI

### LINKED ISSUE #1128: Taker UI keeps "flickering" because we show for a split second that we are unconnected from the daemon
This happens really frequently for me and is annoying because it keeps "moving" the UI up and down because the warning appears at the top for a split second and then is gone again. 

Maker and taker running locally. The logs don't show any connection issues so I assume the problem is the user interface.
I could not see any relationship between what's happening in the daemon logs and the message appearing.

(don't get distracted by the red dots appearing, that's just me clicking ;)

da-kami:
By setting `interval_msecs * 2` we basically say that if we skip one, the next one would have to be exactly at `interval_msecs * 2` which might cause trouble if the daemon is not running on the same machine.
By increasing the to `interval_msecs * 2` we have more resilience in skipping one heartbeat.
Note that this does not explain why we would miss a heartbeat in the first place.

Fixes #1128

### DIFF
--- a/taker-frontend/src/useEventSource.ts
+++ b/taker-frontend/src/useEventSource.ts
@@ -26,7 +26,7 @@ export function useEventSource(url: string): [EventSource | null, boolean] {
 
             const hearbeat: Heartbeat = JSON.parse((event as EventSourceEvent).data);
             const interval_msecs = hearbeat.interval * 1000;
-            const buffered_interval_msecs = interval_msecs * 2;
+            const buffered_interval_msecs = interval_msecs * 2.2;
             timer = setTimeout(() => {
                 setIsConnected(false);
             }, buffered_interval_msecs);


klochowicz:
do you really think flickering is caused by that? are there any logs indicating that? Can this be the issue in how we restart the timer?

thomaseizinger:
Instead of `* 2`, just an additional 10% might be good enough. I don't think we are missing hearbeats but if we send them out exactly every 5 sec and we only have a buffer of 5 sec then that can indeed cause a race of the two timers, whichever fires first wins which can lead to flickering.

da-kami:
> do you really think flickering is caused by that? are there any logs indicating that? Can this be the issue in how we restart the timer?

When I removed `*2` I saw the message more often (which is quite obvious). I would have to run the daemon on a different machine to ensure this works. Locally it is almost impossible to reproduce the flickering anyway. 

I stared at the code for a while and came to the conclusion that it should not do any harm to change this, and it might help. Could not find other code pathes that would cause the problem. 
I'm happy to reason about it together sometimes. The only thing I can tell: The flickering in production is super annoying. The most annoying thing is, that sometimes I actually *do* lose connection and sometimes not. When the message is only shown for short time then I usually assume that there is no problem, but that's still super annoying that I don't know if I should refresh or not.

thomaseizinger (approved):

klochowicz (approved):
we might as well give it a go.

klochowicz:
@da-kami if it doesn't do the trick, I'm quite sure that the initial implementation (that used a `timeoutHandle`) did not do the flickering.

da-kami:
> @da-kami if it doesn't do the trick, I'm quite sure that the initial implementation (that used a `timeoutHandle`) did not do the flickering.

How do you know it does not do the trick? Did you test with this version and still saw it happen?

klochowicz:
> > @da-kami if it doesn't do the trick, I'm quite sure that the initial implementation (that used a `timeoutHandle`) did not do the flickering.
> 
> How do you know it does not do the trick? Did you test with this version and still saw it happen?

only anecdotal evidence - no-one reported the flickering issue before that refactor. could not be the case, but after looking at the code, it seems plausible to me.

PR #1470, opened 2022-02-21
Delay refund tx broadcast by delaying reinit of monitoring

rishflab:
Different way of doing #1452 

Merge either PR.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -30,6 +30,7 @@ use std::convert::TryInto;
 use std::fmt;
 use std::ops::Add;
 use std::time::Duration;
+use tokio::time::sleep;
 use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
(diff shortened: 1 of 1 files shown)


luckysori (approved):

daemon/src/monitor.rs
```diff
@@ -808,6 +809,16 @@ impl xtra::Actor for Actor {
                             Some(params) => params,
                         };
 
+                        // NOTE: this is a band-aid fix.
+                        // It is possible for an attestation to be available when the refund
+                        // timelock has expired, for example, when the daemon goes down and is
+                        // restarted. In this case we want to prioritise
+                        // broadcasting the cet over the refund transaction.
+                        // We wait at least 5 seconds after the monitor actor is initialised before
```
thomaseizinger: How did we choose this number? I would have thought that we need to at least put 30 seconds here to avoid the race condition.

rishflab: Our sync interval is 5 seconds and task is added before reinitialising monitoring so we would get at least one 1 check to see if there is a pending attestation. I feel 30 secs might be too high as the user will experience a fairly long delay before seeing an updated state on restart. Maybe 10 seconds?

thomaseizinger: On the other hand, if we pick a too short interval, we might not actually fix the problem!

Let's go with 30s. If we want to optimise this, we should fix it properly by not making it time-dependent.

thomaseizinger (approved):

PR #1471, opened 2022-02-21
Add an API to calculate payouts when the taker goes short

### LINKED ISSUE #1432: Add API to generate a payout curve to allow the taker to go short

luckysori:
Fixes #1432.

### DIFF
--- a/model/src/payout_curve.rs
+++ b/model/src/payout_curve.rs
@@ -41,7 +41,10 @@ mod utils;
 /// * price: BTC-USD exchange rate used to create CFD contract
 /// * quantity: Interger number of one-dollar USD contracts contained in the
 /// CFD; expressed as a Usd amount
-/// * leverage: Leveraging used by the taker
+/// * long_leverage: leverage used by the party with the long position
+/// * short_leverage: leverage used by the party with the short position
+/// * n_payouts: number of segments into which the payout curve is discretised
+/// * fee: offset applied to the curve representing a fee paid between parties
 ///
 /// ### Returns
 ///
--- a/model/src/payout_curve.rs
+++ b/model/src/payout_curve.rs
@@ -850,16 +850,16 @@ mod tests {
         fn payout_totals_are_similar(
             price in arb_price(1.0, 340_000.0),
             n_contracts in arb_contracts(1, 10_000_000),
-            taker_leverage in arb_leverage(1, 200),
-            maker_leverage in arb_leverage(1, 200),
+            long_leverage in arb_leverage(1, 200),
+            short_leverage in arb_leverage(1, 200),
             n_payouts in 10usize..2000,
             fee_flow in arb_fee_flow(-100_000_000, 100_000_000),
         ) {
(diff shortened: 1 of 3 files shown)


rishflab (approved):

model/src/payout_curve.rs
```diff
@@ -158,13 +182,21 @@ struct PayoutParameter {
...
-    fn into_payouts(self) -> Result<Vec<Payout>> {
+    fn into_payouts_long_taker(self) -> Result<Vec<Payout>> {
         generate_payouts(
             self.left_bound..=self.right_bound,
             bitcoin::Amount::from_sat(self.short_amount),
             bitcoin::Amount::from_sat(self.long_amount),
         )
     }
+
+    fn into_payouts_short_taker(self) -> Result<Vec<Payout>> {
```
thomaseizinger: Why is this necessary? If the caller know the role of the position, wouldn't that be good enough?

luckysori: It is not necessary but convenient. Before we had an `into_payouts` function which knew that the taker was always long. 

I can remove both functions and inline this logic.

thomaseizinger: Shouldn't we change `maia` to care about `long` and `short` instead of `maker` and `taker`?

I am happy to merge this for now but we should fix this in maia I believe.

thomaseizinger: I created a ticket for it: https://github.com/comit-network/maia/issues/16

luckysori: > Shouldn't we change `maia` to care about `long` and `short` instead of `maker` and `taker`?

Ultimately yes.

model/src/payout_curve.rs
```diff
@@ -43,29 +41,63 @@ mod utils;
...
+/// * long_leverage: leverage used by the party with the long position
+/// * short_leverage: leverage used by the party with the short position
+/// * n_payouts: number of segments into which the payout curve is discretised
+/// * fee: offset applied to the curve representing a fee paid between parties
 ///
 /// ### Returns
 ///
 /// The list of [`Payout`]s for the given price, quantity and leverage.
-pub fn calculate(
+pub fn calculate_long_taker(
```
da-kami: Is the expectation that there will be a `calculate_short_taker` as well? 
Will there be separate functions for `calculate_long_maker` and `calculate_short_maker`? 

I would actually find it easier if we would move this logic on the `Cfd`. The cfd knows position and role and can decide how to call the `calculate_payout_parameters` parameters. I find the indirection with `calculate_long_taker` in between confusing.

klochowicz (approved):
I'm happy for this to go in, as it's an improvement. removing the "maker" and "taker" from this part of the code should probably be tackled along with `maia` refactor.

model/src/payout_curve.rs
luckysori: Fair, I can try to move these APIs to the `model` crate.

thomaseizinger (approved):

Issue #1468, opened 2022-02-21
Taker gets stuck in rollover protocol and position got closed


bonomat:
It looks like the rollover failed and the taker was never able to recover. The position was force closed eventually: 

```
2022-02-19 16:00:04  WARN daemon::model::cfd: Rollover failed: Failed to get announcement: Announcement /x/BitMEX/BXBT/2022-02-20T17:00:00.price?n=20 not found order_id=c90270e7-d11c-4482-a409-4e8ffb891a90
2022-02-19 16:00:14  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 10 seconds
2022-02-19 16:00:15 DEBUG daemon::bitmex_price_feed: Received new quote: Quote { timestamp: Timestamp(1645286400), bid: Price(39966), ask: Price(39966.5) }
2022-02-19 16:00:24  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 20 seconds
2022-02-19 16:00:34  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 30 seconds
2022-02-19 16:00:44  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 40 seconds
2022-02-19 16:00:54  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 50 seconds
2022-02-19 16:01:04  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 60 seconds
2022-02-19 16:01:14  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 70 seconds
...

2022-02-21 02:25:48  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 123940 seconds
2022-02-21 02:25:58  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 123950 seconds
2022-02-21 02:26:08  WARN xtra::context: Actor daemon::connection::Actor has been processing message daemon::connection::MakerStreamMessage for 123960 seconds
```

da-kami:
[...]

PR #1429, opened 2022-02-16
Time out messages forwarded to ephemeral actors

luckysori:
Fixes #1423.

I didn't want to add timeouts everywhere because I believe we did so in a PR which we ended up closing because it broke the tests. Should we add more timeouts? Is 60 seconds reasonable?

PR #1472, opened 2022-02-21
Timeouts for sending to actors for maker and taker

da-kami:
We had a scenario where the taker's connection actor was blocked (forever) upon a failed rollover, where the maker was sending `Msg0` (because for the maker the rollover protocol did not fail. Upon receiving rollover `Msg0` the taker was waiting for the delivery of this message to the rollover actor indefinitely.
Since this blocked the connection actor for processing messages this is a critical bug.
In order to ensure, that the connection actor does not get stuck when processing messages and sending them to actors, we add timeouts to all messages that are sent out to actors. The timeout is set to 60 seconds.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -42,6 +42,7 @@ use xtras::SendInterval;
 pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
 
 const TCP_TIMEOUT: std::time::Duration = Duration::from_secs(10);
+const HANDLE_MESSAGE_TIMEOUT: std::time::Duration = Duration::from_secs(5);
 
 /// The "Connected" state of our connection with the maker.
 #[allow(clippy::large_enum_variant)]
@@ -449,6 +450,7 @@ impl Actor {
                 if self
                     .setup_actors
                     .send_fallible(&order_id, setup_taker::Accepted)
+                    .timeout(HANDLE_MESSAGE_TIMEOUT)
                     .await
                     .is_err()
                 {
@@ -459,6 +461,7 @@ impl Actor {
                 if self
                     .setup_actors
                     .send_fallible(&order_id, setup_taker::Rejected::without_reason())
+                    .timeout(HANDLE_MESSAGE_TIMEOUT)
                     .await
                     .is_err()
                 {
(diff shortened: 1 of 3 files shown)


da-kami:
This is a patch fix for https://github.com/itchysats/itchysats/issues/1468 until we handle this more complete in master

luckysori:
We already did basically the same thing with this PR merged into master: https://github.com/itchysats/itchysats/pull/1429. It probably doesn't matter that we don't use that specific patch for the backported fix, but letting you know in case you didn't know.

da-kami:
> We already did basically the same thing with this PR merged into master: #1429. It probably doesn't matter that we don't use that specific patch for the backported fix, but letting you know in case you didn't know.

Thanks for pointing that out @luckysori.
I noticed that there is one `send` that we did not wrap with a `timeout` on master - that is the one of the current order. Maybe that was on purpose. What do you think @thomaseizinger shall I backport https://github.com/itchysats/itchysats/pull/1429 or is it OK to just roll with this? 
Don't think it will make much difference. Might be faster to just merge this to trigger a release and create a PR on Umbrel.

thomaseizinger:
> backport #1429 or is it OK to just roll with this?

There is no difference other than the variable name, yes?

However, I **don't** want to merge this. If you want to open a PR for a backport, I recommend the following flow:

- Make a new branch `release/0.4.4` that branches of `0.4.3`
- Create a PR from another branch towards `release/0.4.4`
- Merge PR
- Release top of `release/0.4.4` branch

klochowicz:
> > backport #1429 or is it OK to just roll with this?
> 
> There is no difference other than the variable name, yes?
> 
> However, I **don't** want to merge this. If you want to open a PR for a backport, I recommend the following flow:
> 
>     * Make a new branch `release/0.4.4` that branches of `0.4.3`
> 
>     * Create a PR from another branch towards `release/0.4.4`
> 
>     * Merge PR
> 
>     * Release top of `release/0.4.4` branch

I will be doing a release tomorrow or the day after anyway to align the API changes in the maker (so that a single automation can serve both makers), I'm happy to help out with backporting (or even just cut a release myself, as I'll be doing it anyway)

luckysori (approved):

PR #1484, opened 2022-02-22
Demote failure log message to debug

thomaseizinger:
A taker disconnecting is not something we can do much about. The
network connection failing when it shouldn't is a problem but we
cannot tell that from this code path.

Thus, logging this on error just clutters the logs and makes looking
for actual errors much harder.

@klochowicz Might be worth considering to include this in the point release.


bonomat (approved):
👍

PR #1487, opened 2022-02-22
Allow setting both short and long prices when creating an offer

klochowicz:
The price for maker going long is unused for now and only provided for the
future compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.3] - 2022-02-14
+
+- Fix a potential deadlock in taker after losing connection to a maker by introducing timeouts for reading and writing to TCP socket.
+
 ## [0.4.2] - 2022-02-08
 
 ## [0.4.1] - 2022-02-04
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -670,7 +670,7 @@ dependencies = [
 
 [[package]]
 name = "daemon"
-version = "0.4.2"
+version = "0.4.3"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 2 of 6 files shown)


thomaseizinger (approved):
ACK but why do we need to backport this?

klochowicz:
> ACK but why do we need to backport this?

 soon we'll have to have one automation servicing two different maker versions and this will make sure that the API it interacts with can remain the same

thomaseizinger:
> > ACK but why do we need to backport this?
> 
> soon we'll have to have one automation servicing two different maker versions and this will make sure that the API it interacts with can remain the same

Makes sense, thanks for the explanation!

da-kami:
I am confused - this was pushed onto `release/0.4.4` by you @klochowicz or did I misunderstand something? 
Can this just be closed or pointed to master now?

thomaseizinger:
> I am confused - this was pushed onto `release/0.4.4` by you @klochowicz or did I misunderstand something? Can this just be closed or pointed to master now?

Yes I think it was included :)

This can be re-targeted to master now AFAIK.

da-kami:
I changed it to point to master

klochowicz:
> > I am confused - this was pushed onto `release/0.4.4` by you @klochowicz or did I misunderstand something? Can this just be closed or pointed to master now?
> 
> Yes I think it was included :)
> 
> This can be re-targeted to master now AFAIK.

shouldn't we just merge the 0.4.4 changes into master instead? so that the changelog, version numbers etc all agrees?

da-kami:
I expect the next release from `master` to be `0.5.0` - I think we will have breaking changes, so I would not bother too much aligning the version.

klochowicz:
> I expect the next release from `master` to be `0.5.0` - I think we will have breaking changes, so I would not bother too much aligning the version.

I don't think that's an argument for not merging things done on a side branch into master?

klochowicz:
I've opened a merge back from the released side branch: https://github.com/itchysats/itchysats/pull/1490

PR #1491, opened 2022-02-22
Gather metrics about the UTXOs of our wallet

### LINKED ISSUE #1331: Expose initial prometheus metrics from binaries
- Consider whether the metrics should be per actor (i.e. fields that are owned by an actor) or global static metrics
- Expose via HTTP endpoint (reuse rocket, don't open a 2nd port)

thomaseizinger:
curling the endpoint looks something like this:

```
# HELP balance The sum of available UTXOs in the wallet.
# TYPE balance gauge
balance 25196304
# HELP max_utxo_value The largest UTXO.
# TYPE max_utxo_value gauge
max_utxo_value 22628441
# HELP mean_utxo_value The mean UTXO.
# TYPE mean_utxo_value gauge
mean_utxo_value 1938177.230769231
# HELP median_utxo_value The median UTXO.
# TYPE median_utxo_value gauge
median_utxo_value 229999
# HELP min_utxo_value The smallest UTXO.
# TYPE min_utxo_value gauge
min_utxo_value 7385
# HELP num_utxos The number of available UTXOs in the wallet.
# TYPE num_utxos gauge
num_utxos 13
# HELP stddev_utxo_value The standard deviation across all UTXOs.
# TYPE stddev_utxo_value gauge
stddev_utxo_value 6217081.219581583
```

I opted to use the global registry because it makes collecting the metrics a lot more concise. If we would have more wallets for example and we wanted to collect metrics per wallet, I would suggest using `vec` metrics: https://docs.rs/prometheus/0.13.0/prometheus/macro.register_gauge_vec.html

Fixes #1331.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -22,12 +22,49 @@ use maia::TxBuilderExt;
 use model::Timestamp;
 use model::TxFeeRate;
 use model::WalletInfo;
+use statrs::statistics::*;
 use std::collections::HashSet;
 use std::time::Duration;
 use tokio::sync::watch;
 use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
+
+static BALANCE_GAUGE: conquer_once::Lazy<prometheus::Gauge> = conquer_once::Lazy::new(|| {
+    prometheus::register_gauge!(
+        "balance",
+        "The sum of available UTXOs in the wallet in satoshis."
+    )
+    .unwrap()
+});
+static NUM_UTXO_GAUGE: conquer_once::Lazy<prometheus::Gauge> = conquer_once::Lazy::new(|| {
+    prometheus::register_gauge!("num_utxos", "The number of available UTXOs in the wallet.")
+        .unwrap()
+});
+static MEDIAN_UTXO_VALUE_GAUGE: conquer_once::Lazy<prometheus::Gauge> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_gauge!("median_utxo_value", "The median UTXO, in satoshis.").unwrap()
+    });
+static MIN_UTXO_VALUE_GAUGE: conquer_once::Lazy<prometheus::Gauge> =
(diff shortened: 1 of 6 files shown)


daemon/src/wallet.rs
```diff
@@ -22,13 +22,49 @@ use maia::TxBuilderExt;
...
+static BALANCE_GAUGE: conquer_once::Lazy<prometheus::Gauge> = conquer_once::Lazy::new(|| {
+    prometheus::register_gauge!("balance", "The sum of available UTXOs in the wallet.").unwrap()
+});
+static NUM_UTXO_GAUGE: conquer_once::Lazy<prometheus::Gauge> = conquer_once::Lazy::new(|| {
+    prometheus::register_gauge!("num_utxos", "The number of available UTXOs in the wallet.")
+        .unwrap()
+});
+static MEDIAN_UTXO_VALUE_GAUGE: conquer_once::Lazy<prometheus::Gauge> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_gauge!("median_utxo_value", "The median UTXO.").unwrap()
```
bonomat: The second field of this macro is the unit. Can we add that this is in satoshi?

bonomat (approved):
nice, we will need to adapt the prometheus scraping settings to cover this

thomaseizinger:
> we will need to adapt the prometheus scraping settings to cover this

It is only additive though right? It just another metrics endpoint we need to add, nothing existing should be breaking!

daemon/src/wallet.rs
thomaseizinger: I understand that it is a general help text but yes can certainly add in the unit!

bonomat:
> > we will need to adapt the prometheus scraping settings to cover this
> 
> It is only additive though right? It just another metrics endpoint we need to add, nothing existing should be breaking!

yes, that's what I meant. Tell prometheus to also scrape this endpoint.

da-kami (approved):
Cool!

klochowicz (approved):

PR #1500, opened 2022-02-23
Track active connections as a first-class prometheus gauge

thomaseizinger:
Once this is merged we can pick up https://github.com/itchysats/itchysats/pull/1299 again.

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -143,6 +143,8 @@ impl Actor {
         heartbeat_interval: Duration,
         p2p_socket: SocketAddr,
     ) -> Self {
+        NUM_CONNECTIONS_GAUGE.set(0);
+
         Self {
             connections: HashMap::new(),
             taker_connected_channel: taker_connected_channel.clone_channel(),
@@ -165,6 +167,8 @@ impl Actor {
                 .send_async_safe(maker_cfd::TakerDisconnected { id: *taker_id })
                 .await;
         }
+
+        NUM_CONNECTIONS_GAUGE.set(self.connections.len() as i64);
     }
 
     async fn send_to_taker(
@@ -384,6 +388,8 @@ impl Actor {
             },
         );
 
+        NUM_CONNECTIONS_GAUGE.set(self.connections.len() as i64);
+
         tracing::info!(taker_id = %identity, "Connection is ready");
     }
 
@@ -530,3 +536,12 @@ impl xtra::Actor for Actor {
 
     async fn stopped(self) -> Self::Stop {}
 }
+
+static NUM_CONNECTIONS_GAUGE: conquer_once::Lazy<prometheus::IntGauge> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_int_gauge!(
+            "p2p_connections_total",
+            "The number of active p2p connections."
+        )
+        .unwrap()
+    });


daemon/src/maker_inc_connections.rs
```diff
@@ -33,6 +33,12 @@ use xtras::AddressMap;
 use xtras::SendAsyncSafe;
 use xtras::SendInterval;
 
+static NUM_CONNECTIONS_GAUGE: conquer_once::Lazy<prometheus::IntGauge> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_int_gauge!("num_connections", "The number of active connections.")
```
bonomat: Proposal: rename to number of connected takers or similar. Number of connections might too ambiguous

bonomat (approved):
Not sure if we want to keep the pattern, but in maker-metrics I prefixed itchysats metrics with 'itchysats_' which makes it easy to search for in grafana

daemon/src/maker_inc_connections.rs
thomaseizinger: We can do that here but in the near future (once we have a network layer based on libp2p), maker and taker will run the same network actor so we can't easily distuingish the metric based on the role.

Is there a way in prometheus to take the source (i.e. which HTTP endpoint was polled) of a metric into account in its name?

thomaseizinger:
> Not sure if we want to keep the pattern, but in maker-metrics I prefixed itchysats metrics with 'itchysats_' which makes it easy to search for in grafana 

Good idea! Can we do this through configuration based on where the metric is imported from? Having to do that on every metric seems excessive and makes it harder to write reusable code that comes with metrics but is unaware of which app it will be embedded in.

daemon/src/maker_inc_connections.rs
bonomat: > We can do that here but in the near future (once we have a network layer based on libp2p), maker and taker will run the same network actor so we can't easily distuingish the metric based on the role.

That's fine, I meant the comby of `num` and `connections` is ambigious. 
Maybe call it p2p connection or incomming connections? When I hear `connections` it could mean various things: `http_connections`, `outgoing_connections`, ... I feel like we need to add one more thing to this field :D 

> Is there a way in prometheus to take the source (i.e. which HTTP endpoint was polled) of a metric into account in its name?

yes, each _scrape_ target has a name (http endpoint), which will end up as a label to this metric.

bonomat:
> > Not sure if we want to keep the pattern, but in maker-metrics I prefixed itchysats metrics with 'itchysats_' which makes it easy to search for in grafana
> 
> Good idea! Can we do this through configuration based on where the metric is imported from? Having to do that on every metric seems excessive and makes it harder to write reusable code that comes with metrics but is unaware of which app it will be embedded in.

Fair. Not sure if this is possible. We can for sure add labels which are also easy to filter for.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config

daemon/src/maker_inc_connections.rs
thomaseizinger: Okay. What do you think of `num_p2p_connections`?

thomaseizinger: > > Is there a way in prometheus to take the source (i.e. which HTTP endpoint was polled) of a metric into account in its name?
> 
> yes, each _scrape_ target has a name (http endpoint), which will end up as a label to this metric.

Nice, we can add `maker` and `taker `labels there then!

thomaseizinger:
> > > Not sure if we want to keep the pattern, but in maker-metrics I prefixed itchysats metrics with 'itchysats_' which makes it easy to search for in grafana
> > 
> > 
> > Good idea! Can we do this through configuration based on where the metric is imported from? Having to do that on every metric seems excessive and makes it harder to write reusable code that comes with metrics but is unaware of which app it will be embedded in.
> 
> Fair. Not sure if this is possible. We can for sure add labels which are also easy to filter for.
> 
> [prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config)

Based on your other reply, what about adding the label `itchysats` based on the scrape target?

thomaseizinger:
@bonomat Thanks for the pointer in regards to naming!

I read through https://prometheus.io/docs/practices/naming/ and decided that a more appropriate name would be: `p2p_connections_total`. What do you think?

bonomat:
> @bonomat Thanks for the pointer in regards to naming!
> 
> 
> 
> I read through https://prometheus.io/docs/practices/naming/ and decided that a more appropriate name would be: `p2p_connections_total`. What do you think?

Yea, I think that's a good name for this metric.

daemon/src/maker_inc_connections.rs
thomaseizinger: I recorded an issue for this here: https://github.com/itchysats/cloud-conf/issues/18

luckysori:
bors retry

PR #1503, opened 2022-02-24
Prioritise collab settlement over rollover in projection

klochowicz:


thomaseizinger (approved):
Can merge this but hopefully, we should not actually run into this conflict anymore with the changes that @da-kami is preparing!

da-kami:
> Can merge this but hopefully, we should not actually run into this conflict anymore with the changes that @da-kami is preparing!

Fun fact: We are actually not appending `...Started` events if a condition to start a protocol is violated. In such a case the `Executor` would return an error which leads to just appending a `...Failed` event without started. 
We *could* potentially not do that, but a `Failed` event should actually not do any harm.

So, what I was working on was pretty useless I think because it is basically already in place (just took me some time to understand that...).

@klochowicz found some other model problems that are currently being fixed and should resolve the issues we encountered. We'll see to add more tests for that (actor and/or model tests) to ensure that the behavior in the future.

da-kami (approved):

thomaseizinger:
> In such a case the `Executor` would return an error which leads to just appending a `...Failed` event without started.

What? I don't think we should append events for failures in the executor.

bonomat:
> > In such a case the `Executor` would return an error which leads to just appending a `...Failed` event without started.
> 
> What? I don't think we should append events for failures in the executor.

we should document this if this is a design choice.

PR #1486, opened 2022-02-22
WIP: Activate short button in taker UI

rishflab:
Screenshot from 2022-02-24 10-49-07

PR #1504, opened 2022-02-24
Add short taker UI 2

### LINKED ISSUE #1422: Improve the Taker UI to allow going short
Initial version of going short in the taker's UI. 

 - tiles should show the Position (Long / Short)
 - at the very least we need 2 different prices (perhaps price on the button again?)

open questions hopefully solved by this task:
  - does it make sense to be able to go long and short at the same time?
  - is the UI still understandable with the new additions or we need some bigger redesign?

note: this task is not blocked by the backend, could initially show a dummy price in the UI.

rishflab:
Finalised/cleaned up version of #1486

### DIFF
--- a/taker-frontend/src/components/ConfirmOrderModal.tsx
+++ b/taker-frontend/src/components/ConfirmOrderModal.tsx
@@ -24,6 +24,7 @@ import DollarAmount from "./DollarAmount";
 
 interface Props {
     orderId: string;
+    position: string;
     isOpen: boolean;
     onClose: any;
     onSubmit: (req: CfdOrderRequestPayload) => void;
@@ -40,6 +41,7 @@ interface Props {
 
 export default function ConfirmOrderModal({
     orderId,
+    position,
     isOpen,
     onClose,
     onSubmit,
@@ -129,6 +131,7 @@ export default function ConfirmOrderModal({
                                 onSubmit({
                                     order_id: orderId,
                                     quantity,
+                                    position,
                                 });
 
                                 onClose();
--- a/taker-frontend/src/components/ConfirmOrderModal.tsx
+++ b/taker-frontend/src/components/ConfirmOrderModal.tsx
@@ -1,0 +1,144 @@
+import {
+    Button,
+    HStack,
+    Modal,
+    ModalBody,
+    ModalCloseButton,
+    ModalContent,
+    ModalFooter,
+    ModalHeader,
+    ModalOverlay,
+    Table,
+    TableCaption,
+    Tbody,
+    Td,
+    Text,
+    Tooltip,
+    Tr,
+} from "@chakra-ui/react";
+import * as React from "react";
+import { useRef } from "react";
+import { CfdOrderRequestPayload } from "../types";
+import BitcoinAmount from "./BitcoinAmount";
+import DollarAmount from "./DollarAmount";
+
+interface Props {
+    orderId: string;
+    isOpen: boolean;
+    onClose: any;
+    onSubmit: (req: CfdOrderRequestPayload) => void;
+    isSubmitting: boolean;
+    quantity: number;
+    askPriceAsNumber?: number;
+    margin: number;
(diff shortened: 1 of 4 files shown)


taker-frontend/src/components/ConfirmOrderModal.tsx
```diff
@@ -0,0 +1,153 @@
...
+    confirmRef: any;
+    quantity: number;
+    minQuantity: number;
+    askPriceAsNumber?: number;
+    margin: number;
+    leverage: number;
+    liquidationPriceAsNumber: number | undefined;
+    feeForFirstSettlementInterval: number;
+    fundingRateHourly: string;
+    fundingRateAnnualized: string;
```
thomaseizinger: This seems a bit excessive, do we really need all of these?

`setQuantity` for example is only used to set the input field back to `minQuantity`. This component doesn't even know about the input field so I'd set the call to `setQuantity` should happen in the callback that the parent passes in for `onClose`.

thomaseizinger: We should do a similar analysis for all other props.

taker-frontend/src/types.ts
```diff
@@ -235,6 +235,7 @@ export enum StateGroupKey {
 export interface CfdOrderRequestPayload {
     order_id: string;
     quantity: number;
+    position: string;
```
thomaseizinger, line 238: I think I'd have a slight preference for a 2nd endpoint instead of adding a field here.

thomaseizinger (changes requested):
Thanks! Nice commit structure. I got no opinion on the colour change but I think the `ConfirmOrderModal` component has too many props and we should be able to simplify some of those!

da-kami:
I am a bit confused, is it supposed to look like this?

image

Several points  addressed in https://github.com/itchysats/itchysats/issues/1422 were not actioned. 
I don't think this closes this ticket. 

- There have to be two prices (one for short and one for long). I only see one price (the long price) - am I missing something?
- The blinking price should likely go away. The idea was to simplify the two prices and put them on the button for now. We can experiment with two blinking prices, but am not sure about that.
- The required margin for short and long is different (because the prices are different), so we have to show short and long margin

The position colors can be improved, but that is not super important and can be done later. I was thinking about changing the color palette a bit more.

rishflab:
> I am a bit confused, is it supposed to look like this?
> 
> image
> 
> Several points addressed in #1422 were not actioned. I don't think this closes this ticket.
> 
>     * There have to be two prices (one for short and one for long). I only see one price (the long price) - am I missing something?
> 
>     * The blinking price should likely go away. The idea was to simplify the two prices and put them on the button for now. We can experiment with two blinking prices, but am not sure about that.
> 
>     * The required margin for short and long is different (because the prices are different), so we have to show short and long margin
> 
> 
> The position colors can be improved, but that is not super important and can be done later. I was thinking about changing the color palette a bit more.

I think you are right, removed the closes PR tag. I can add a two prices in this PR (use a dummy price for the short). Did not understand how the price worked (backend changes were not complete and I was looking at the backend for guidance).

thomaseizinger:
Even though it does not complete the ticket, I think we can still merge these commits independently I guess?

What do you think @da-kami ?

taker-frontend/src/types.ts
klochowicz: I don't think we need 2 endpoints at all - the `order_id` unambiguously denotes the order. different positions will need different order_ids.

rishflab: yep I don't see the need, the position is in the payload.

thomaseizinger (approved):
concept ACK

The refactoring can be squashed into the original commit that creates the modal.

klochowicz (approved):
cool, let's squash this and let's get this in! thanks :)

thomaseizinger:
@rishflab Some typescript errors are failing the build.

PR #1505, opened 2022-02-24
Fix errors around collab settlement and rollover

klochowicz:
There were multiple cases where we would stop trying to finish a protocol instead of appending a "failed" event.
See commits for details (and new tests).

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -428,6 +428,41 @@ async fn maker_rejects_rollover_after_commit_finality() {
 }
 
 #[tokio::test]
+async fn maker_accepts_rollover_after_commit_finality() {
+    let _guard = init_tracing();
+    let oracle_data = OliviaData::example_0();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(oracle_data.announcement(), Position::Short).await;
+
+    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    next_with(taker.quote_feed(), |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
+
+    taker.trigger_rollover(order_id).await;
+
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
+        CfdState::IncomingRolloverProposal,
+        CfdState::OutgoingRolloverProposal
+    );
+
+    confirm!(commit transaction, order_id, maker, taker);
+
+    maker.system.accept_rollover(order_id).await.unwrap(); // This should fail
+
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
+        // FIXME: Maker wrongly changes state even when rollover does not happen
+        CfdState::ContractSetup,
+        CfdState::OpenCommitted
+    );
+}
+
+#[tokio::test]
 async fn maker_rejects_collab_settlement_after_commit_finality() {
     let _guard = init_tracing();
     let (mut maker, mut taker, order_id) =
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -428,6 +428,64 @@ async fn maker_rejects_rollover_after_commit_finality() {
 }
 
 #[tokio::test]
+async fn maker_rejects_collab_settlement_after_commit_finality() {
+    let _guard = init_tracing();
+    let (mut maker, mut taker, order_id) =
+        start_from_open_cfd_state(OliviaData::example_0().announcement(), Position::Short).await;
+    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
+    next_with(taker.quote_feed(), |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
+
+    taker.system.propose_settlement(order_id).await.unwrap();
+
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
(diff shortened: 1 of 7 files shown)


daemon-tests/tests/happy_path.rs
```diff
@@ -427,6 +427,66 @@ async fn maker_rejects_rollover_after_commit_finality() {
...
+    wait_next_state!(
+        order_id,
+        maker,
+        taker,
+        CfdState::IncomingSettlementProposal,
+        CfdState::OutgoingSettlementProposal
+    );
+
+    confirm!(commit transaction, order_id, maker, taker);
+    // Cfd would be in "OpenCommitted" if it wasn't for the rollover
```
thomaseizinger: Is this comment outdated?

daemon-tests/tests/happy_path.rs
```diff
@@ -427,6 +427,66 @@ async fn maker_rejects_rollover_after_commit_finality() {
...
+
+    maker.system.reject_settlement(order_id).await.unwrap();
+    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
+
+    // After rejecting rollover, we should display where we were before the
+    // rollover attempt
```
thomaseizinger, line 490: Outdated?

thomaseizinger (approved):
Nice fixes, thank you!

model/src/cfd.rs
```diff
@@ -1036,26 +1036,16 @@ impl Cfd {
         }))
     }
 
-    pub fn reject_collaborative_settlement(self, reason: anyhow::Error) -> Result<CfdEvent> {
-        anyhow::ensure!(
-            self.can_settle_collaboratively(),
-            "Cannot reject collaborative settlement"
-        );
```
luckysori, line 1044: ❓ Do we still intend to eventually check the version here? Or does removing this mean that we want to be able to unconditionally emit these events?

luckysori (approved):
LGTM. Just one question.

model/src/cfd.rs
da-kami: I think these should be unconditional for now. Version checks would be important if we want to make sure that "nothing else happened in between" (or similar). 
Given, that we have started an ephemeral protocol we always have to be able to fail/reject it (unconditionally), otherwise we might not allow other protocols (or even the same one) to start.

daemon-tests/tests/happy_path.rs
```diff
@@ -427,6 +427,66 @@ async fn maker_rejects_rollover_after_commit_finality() {
...
+    wait_next_state!(order_id, maker, taker, CfdState::OpenCommitted);
+}
+
+#[tokio::test]
+async fn maker_accepts_collab_settlement_after_commit_finality() {
```
da-kami, line 495: I think it would help if we have longer test names:

suggestions:

`given_maker_committed_when_accept_collab_settlement_then_remains_committed`
`given_maker_committed_when_rejects_collab_settlement_then_remains_committed`

It took me some time to understand why these tests are there and what is the difference.

daemon-tests/tests/happy_path.rs
```diff
@@ -427,6 +427,42 @@ async fn maker_rejects_rollover_after_commit_finality() {
     wait_next_state!(order_id, maker, taker, CfdState::OpenCommitted);
 }
 
+#[tokio::test]
+async fn maker_accepts_rollover_after_commit_finality() {
```
da-kami, line 431:
```suggestion
async fn given_maker_committed_when_accept_rollover_then_rolls_over() {
```

I think it should actually be:

```suggestion
async fn given_maker_committed_when_accept_rollover_then_remains_comitted() {
```

and fail...

model/src/cfd.rs
klochowicz: I'd love to see some of these checks (like the one that `Completed/Failed` comes only after `Started`, but rather as `debug_assert!`s or just warnings when we rehydrate the Cfds - projection now deals gracefully with these events. 

Removing checks in this PR is orthogonal to that problem - I'm removing checks that are supposed to check for *starting* a protocol that are being re-purposed later on.

daemon-tests/tests/happy_path.rs
da-kami, line 431: This test showcases that something is wrong at the moment.
Are you planning to fix this bug as part of this PR?

I think it would be better if we write a failing test and then fix this issue. - Or ignore the test for now. 
I found it a bit confusing to have a test that tests that something is a bug.

da-kami (approved):
I am not sure about the test that shows the rollover bug... I think we should rather have a failing test (and potentially ignore it, if you don't want to fix it here...).

daemon-tests/tests/happy_path.rs
klochowicz: This test actually shows that I at least partially fixed a problem in this PR! (it was failing before the changes).
I am planning to tackle the remaining issue today (in the meantime, with merged PR)

daemon-tests/tests/happy_path.rs
klochowicz: I actually would have harder time parsing the suggested names. :/
I'll think about the naming a bit more today, whether it can be meaningfully improved in a way that works for everyone. I'll reach out to you.

daemon-tests/tests/happy_path.rs
klochowicz: happy to have a further chat about these - the test names you proposed are IMHO less legible (at least for me)

klochowicz:
> I am not sure about the test that shows the rollover bug... I think we should rather have a failing test (and potentially ignore it, if you don't want to fix it here...).

I added the test as it fixes a concrete problem (that's why I wanted to keep it active) - the fact that it uncovered a different weirdness is orthogonal to this PR.

PR #1509, opened 2022-02-24
Update sqlx to 0.5.11

### LINKED ISSUE #1433: Upgrade sqlx dependency to released version
Step 1: Remove dependency on fork, now that https://github.com/launchbadge/sqlx/pull/1658 is merged.
Step 2: Update to released version once available.

rishflab:
closes #1433

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -609,16 +609,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ccaeedb56da03b09f598226e25e80088cb4cd25f316e6e4df7d695f0feeb1403"
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e54ea8bc3fb1ee042f5aace6e3c6e025d3874866da222930f70ce62aceba0bfa"
-dependencies = [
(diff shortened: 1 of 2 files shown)


klochowicz:
perhaps we could pin the version of sqlx to `0.5.11` in both model and daemon? (not sure whether strictly needed due to `Cargo.lock` committed, but it would be more explicit)

thomaseizinger (approved):

thomaseizinger:
> perhaps we could pin the version of sqlx to `0.5.11` in both model and daemon? (not sure whether strictly needed due to `Cargo.lock` committed, but it would be more explicit)

You mean in the `Cargo.toml`? I think we should set that to `0.5` to make it clear that cargo will bump patch versions.

klochowicz:
> > perhaps we could pin the version of sqlx to `0.5.11` in both model and daemon? (not sure whether strictly needed due to `Cargo.lock` committed, but it would be more explicit)
> 
> You mean in the `Cargo.toml`? I think we should set that to `0.5` to make it clear that cargo will bump patch versions.

it's already 0.5. However, as we specifically wanted versions at least 0.5.11, I thought it could do no harm (dependabot would open a PR anyway about it). Not a blocker though :)

klochowicz (approved):

klochowicz:
bors retry

PR #1515, opened 2022-02-28
Dispatch wire messages to actors async from taker connection

da-kami:
Instead of sending messages to actors sync, we send them async, so we don't block the connection actor when dispatching messages.
This removes the timeouts, because we no longer block the connection actor when dispatching messages.

Note that contract setup currently dispatches messages fallible, which does not make any sense in this context.
We should refactor this to handle protocol errors inside of contract setup instead of returning the error back to the connection.
Due to the fallible dispatch, the messages to contract setup where not changed to async yet.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -33,6 +33,7 @@ use tokio_util::codec::Framed;
 use xtra::prelude::MessageChannel;
 use xtra::KeepRunning;
 use xtra_productivity::xtra_productivity;
+use xtras::address_map::NotConnected;
 use xtras::AddressMap;
 use xtras::LogFailure;
 use xtras::SendInterval;
@@ -43,9 +44,6 @@ pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
 const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 
 const HANDLE_CONTRACT_SETUP_TIMEOUT: Duration = Duration::from_secs(10);
-const HANDLE_ROLLOVER_MESSAGE_TIMEOUT: Duration = Duration::from_secs(10);
-const HANDLE_CURRENT_ORDER_TIMEOUT: Duration = Duration::from_secs(10);
-const HANDLE_SETTLEMENT_MESSAGE_TIMEOUT: Duration = Duration::from_secs(120);
 const HANDLE_PROTOCOL_MESSAGE_TIMEOUT: Duration = Duration::from_secs(60);
 
 /// The "Connected" state of our connection with the maker.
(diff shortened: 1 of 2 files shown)


daemon/src/connection.rs
```diff
@@ -449,58 +444,63 @@ impl Actor {
...
-                    .send_fallible(&order_id, msg)
-                    .timeout(HANDLE_PROTOCOL_MESSAGE_TIMEOUT)
-                    .await
-                    .is_err()
-                {
-                    tracing::warn!(%order_id, "No active contract setup");
+                // TODO: sending to the setup actor should not be fallible
+                match self.setup_actors.send_fallible(&order_id, msg).await {
+                    Ok(Ok(())) => (),
+                    Ok(Err(e)) => tracing::warn!(%order_id, "Contract setup not started {e:#}"),
```
thomaseizinger: Stale message.

xtras/src/address_map.rs
```diff
@@ -63,6 +66,21 @@ where
         Ok(())
     }
 
+    pub async fn send_async_safe<M>(&self, key: &K, msg: M) -> Result<(), NotConnected>
```
thomaseizinger: I think we can drop the `_safe` here. That only exists on the trait to differentiate it from `xtra`'s `send_async` function.

xtras/src/address_map.rs
```diff
@@ -49,6 +50,8 @@ where
         self.inner.insert(key, address);
     }
 
+    // TODO: Support messages with a return value, but that might be out of scope for this (add
+    // ticket)
```
klochowicz: perhaps we could just add a ticket for this one? (this can then be discussed/acted on/dropped separately). Not sure if it's a part of this changeset at all.

xtras/src/address_map.rs
```diff
@@ -63,6 +66,21 @@ where
...
+        self.get(key)?
+            .send_async_safe(msg)
+            .await
+            .map_err(|_| NotConnected::new::<A>())?;
+
+        Ok(())
+    }
+
+    // TODO: Consider removing this, because we should not have fallible messages to protocol
+    // actors, the actors should handle errors internally and finish the protocol accordingly
```
thomaseizinger: The `AddressMap` does not concern itself with protocol actors so this reasoning is not quite appropriate.

thomaseizinger (changes requested):

daemon/src/connection.rs
```diff
@@ -43,12 +44,6 @@ pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
 
 const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 
-const HANDLE_CONTRACT_SETUP_TIMEOUT: Duration = Duration::from_secs(10);
```
klochowicz, line 45: should this timeout be removed as a part of this PR? After all, you haven't converted the contract setup to send things in an async way.

xtras/src/address_map.rs
da-kami: true, I created this one for now: https://github.com/itchysats/itchysats/issues/1517

daemon/src/connection.rs
da-kami: I re-introduced the timeouts for contract setup for now. Does not make it prettier, but yeah, that error path of getting stuck should be covered now.

klochowicz (approved):
thanks for tackling this!

thomaseizinger (approved):
LGTM! Thanks!

PR #1524, opened 2022-03-01
Reduce props of modal

thomaseizinger:


rishflab:
Thanks! I was getting stuck on how to pass the pass the closure in. Tested it works, fixed lint and cherry picked onto the branch.

PR #1527, opened 2022-03-01
Set the creation timestamp based on the first event

da-kami:
Otherwise, e.g. `SetupFailed` and `Rejected` are not handled correctly which is annoying in the list of closed positions:

image

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -171,9 +171,6 @@ pub struct Cfd {
 
     #[serde(skip)]
     aggregated: Aggregated,
-
-    #[serde(skip)]
-    creation_timestamp: Timestamp,
 }
 
 /// Bundle all state extracted from the events in one struct.
@@ -208,6 +205,9 @@ struct Aggregated {
     /// Negotiated states of protocols
     rollover_state: Option<ProtocolNegotiationState>,
     settlement_state: Option<ProtocolNegotiationState>,
+
+    version: u64,
+    creation_timestamp: Timestamp,
 }
 
 impl Aggregated {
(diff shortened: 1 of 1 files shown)


da-kami:
Given that the creation timestamp is not an option it is hard to do this logic in a different way. I thought about turning it around, initially setting it to `None` and then just say `if None then set` in `apply`, otherwise set it afterwards to now... but I don't think it should be an option.

Alternatively, if we want to have it in `apply` this does the trick.:

```
if event.timestamp < self.creation_timestamp {
   self.creation_timestamp = event.timestamp;
}
```

Then we would basically set the first one and then the condition would always be false... That's a bit stupid though, because we would check this condition for every event, although we know we want the one of the first one in the list. Could argue that it is more resilient, because then the list of event would not have to be sorted by timestamp - but I think that's a given we would not change anyway.
Not sure it's easier to understand, what do you think?

thomaseizinger (changes requested):
You could to the following:

- `self.creation_timestamp = now()` in the constructor
- `if version == 0 { self.creation_timestamp = event.timestamp }` in apply

da-kami:
> You could to the following:
> 
>     * `self.creation_timestamp = now()` in the constructor
> 
>     * `if version == 0 { self.creation_timestamp = event.timestamp }` in apply

We do not have a version in `projection::Cfd` - seems I forgot to point that out (only wrote it on Elements at some point...). Should I introduce it - should it be added to `Aggregated`?

da-kami:
> > You could to the following:
> > ```
> > * `self.creation_timestamp = now()` in the constructor
> > 
> > * `if version == 0 { self.creation_timestamp = event.timestamp }` in apply
> > ```
> 
> We do not have a version in `projection::Cfd` - seems I forgot to point that out (only wrote it on Elements at some point...). Should I introduce it - should it be added to `Aggregated`?

I'm already doing it :D

thomaseizinger (approved):
That is pretty clean now, thank you :)

klochowicz (approved):
LGTM!

rishflab (approved):

PR #1528, opened 2022-03-01
Unlock wallet UTXOs after sufficient time has passed

### LINKED ISSUE #1135: Unlock temporarily locked UTXOs on failed contract setup
1. Include failed transaction in event
2. Extract inputs from transaction
3. Ping wallet actor to unlock UTXOs

luckysori:
Fixes #1135.

Most of what was changed with the first 3 patches ended up being unnecessary, so we can drop it if it's unwanted. The main benefit of those changes was to be able to write the test in ff69a50196254056f6690755933fe53c1222943d.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -9,6 +9,7 @@ use bdk::bitcoin::Amount;
 use bdk::bitcoin::OutPoint;
 use bdk::bitcoin::PublicKey;
 use bdk::bitcoin::Txid;
+use bdk::blockchain::Blockchain;
 use bdk::blockchain::ElectrumBlockchain;
 use bdk::blockchain::NoopProgress;
 use bdk::database::BatchDatabase;
@@ -66,8 +67,8 @@ static STD_DEV_UTXO_VALUE_GAUGE: conquer_once::Lazy<prometheus::Gauge> =
         .unwrap()
     });
 
-pub struct Actor {
-    wallet: bdk::Wallet<ElectrumBlockchain, bdk::database::MemoryDatabase>,
+pub struct Actor<B> {
+    wallet: bdk::Wallet<B, bdk::database::MemoryDatabase>,
     used_utxos: HashSet<OutPoint>,
     tasks: Tasks,
     sender: watch::Sender<Option<WalletInfo>>,
@@ -77,7 +78,7 @@ pub struct Actor {
 #[error("The transaction is already in the blockchain")]
 pub struct TransactionAlreadyInBlockchain;
 
-impl Actor {
+impl Actor<ElectrumBlockchain> {
     pub fn new(
         electrum_rpc_url: &str,
         ext_priv_key: ExtendedPrivKey,
@@ -113,7 +114,9 @@ impl Actor {
 
         Ok((actor, receiver))
     }
+}
 
+impl<B: Blockchain> Actor<B> {
     fn sync_internal(&mut self) -> Result<WalletInfo> {
         self.wallet
             .sync(NoopProgress, Some(1000))
@@ -150,7 +153,7 @@ impl Actor {
 }
 
 #[xtra_productivity]
-impl Actor {
+impl Actor<ElectrumBlockchain> {
     pub fn handle_sync(&mut self, _msg: Sync) {
         let wallet_info_update = match self.sync_internal() {
             Ok(wallet_info) => Some(wallet_info),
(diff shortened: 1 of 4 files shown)


daemon/src/wallet.rs
```diff
@@ -345,17 +356,99 @@ where
...
+    inner: HashSet<(Instant, OutPoint)>,
+    time_to_lock: Duration,
+}
+
+impl LockedUtxos {
+    fn new(network: Network) -> Self {
+        let time_to_lock = match network {
+            Network::Bitcoin => Duration::from_secs(20 * 60),
+            Network::Testnet | Network::Signet | Network::Regtest => Duration::from_secs(2 * 60),
+        };
```
luckysori: 👀 I didn't invest time into figuring out good values for these in case we decide to solve this whole issue differently. Suggestions are welcome.

thomaseizinger (approved):
Nice work!

I think we could have gotten away without a test on the actor level although it is nice to test the feature through the actor API :)

daemon/src/wallet.rs
```diff
@@ -354,17 +356,70 @@ where
...
 
+struct LockedUtxos {
+    inner: HashSet<(Instant, OutPoint)>,
+    time_to_lock: Duration,
+}
+
+impl LockedUtxos {
+    fn new(network: Network) -> Self {
+        let time_to_lock = match network {
+            Network::Bitcoin => Duration::from_secs(20 * 60),
```
da-kami: ❓ At the moment we only lock the UTXOs when creating a new DLC, i.e. during contract setup (but not rollover because we use the original lock). This is irrelevant for all spending transactions from lock.
Is what I am stating correct?

daemon/src/wallet.rs
```diff
@@ -354,17 +356,70 @@ where
...
+
+        self.inner.extend(utxos);
+    }
+
+    /// Return the list of locked UTXOs.
+    ///
+    /// Before creating the list, it removes all elements which should
+    /// no longer be part of the set of locked UTXOs.
+    fn list(&mut self) -> Vec<OutPoint> {
+        self.remove_expired();
```
da-kami, line 393: ❓ This removes the UTXOs that were actually spent (i.e. where a tx was published that spends them)?

da-kami (approved):
Nice! Just a few questions to make sure I understand it correctly.

daemon/src/wallet.rs
luckysori: Yes, you're right! We only use wallet UTXOs for the lock transaction, which is only constructed during contract setup.

daemon/src/wallet.rs
luckysori: Not quite. The code doesn't actually reason about whether a UTXO was spent or not. 

UTXOs are locked from the perspective of our wallet up to a time based on a configurable value `time_to_lock`. This `LockedUtxos` data structure will eventually unlock them if enough time has passed. Some of those locked UTXOs may have been used in transactions, but enough time should have passed that the wallet doesn't attempt to use them again (and if the wallet does it would just fail). Others were locked but never used because contract setup failed, so it is desirable to unlock them (this is the feature we're adding with this PR).

thomaseizinger: We may want to couple this unlock time with our sync interval in some form.

luckysori: Can you describe what you're suggesting in more detail, @thomaseizinger?

daemon/src/wallet.rs
```diff
@@ -345,17 +356,113 @@ where
...
+    ///
+    /// We have chosen a value of 20 minutes to allow for 1 or 2
+    /// confirmations if the UTXO was actually used. This way we are
+    /// unlikely to attempt to reuse the same UTXO and compete with
+    /// one of our transactions in the mempool.
+    const TIME_TO_LOCK_MAINNET: Duration = Duration::from_secs(20 * 60);
+
+    /// The minimum time that must pass before a locked UTXO is
+    /// released on any of the Bitcoin test networks.
+    const TIME_TO_LOCK_TEST: Duration = Duration::from_secs(2 * 60);
```
luckysori: @thomaseizinger @da-kami I didn't end up changing these values but I documented the reason why the mainnet values were chosen. Does this make sense to you?

daemon/src/wallet.rs
thomaseizinger: Whether or not it is safe to discard locked UTXOs depends on our sync picking up the new transactions (and thus marking it internally as spent. `LockedUtxos` only exists because we need to reliably bridge the time gap between signing a transaction and the UTXOs showing up during the sync.

If we set `time_to_lock` to something like 4 * sync_internal, we should be reasonable safe.

thomaseizinger: Note that we do pick up UTXOs used by transactions in the mempool so we don't actually need to wait for a confirmation.

luckysori: Thanks for the explanation, that makes sense to me now.

Basically, by setting the `time_to_lock` field to `4 * sync_interval` we're saying that we don't expect any locked UTXOs to end up being used if they were not picked up by the  wallet after 4 `sync_interval`s.

thomaseizinger: > Basically, by setting the `time_to_lock` field to `4 * sync_interval` we're saying that we don't expect any locked UTXOs to end up being used if they were not picked up by the wallet after 4 `sync_interval`s.

Correct.

daemon/src/wallet.rs
```diff
@@ -95,12 +99,20 @@ impl Actor {
             ElectrumBlockchain::from(client),
         )?;
 
+        // UTXOs chosen after coin selection will only be locked for a
+        // few wallet sync intervals. UTXOs which were actually
+        // included in published transactions should be marked as
+        // spent by the internal bdk wallet by then. UTXOs which ended
+        // up not being used are expected to be safe to be reused by
+        // then without incurring in double spend attempts.
+        let time_to_lock = SYNC_INTERVAL * 4;
```
luckysori, line 108: @thomaseizinger: From what we previously discussed. I removed all logic based on Bitcoin networks accordingly.

luckysori:
bors retry

PR #1529, opened 2022-03-01
Do not return Result after calculating next announcement

klochowicz:
Based on the documentation of the function passed in, there is no possibility of
this function to fail. Tests cover the edge case (past 11pm, over midnight).

Cleaning this API makes reasoning about the code clearer - before we defensively
wrapped this code around `try_continue!` macro.


thomaseizinger (approved):
Nice!

PR #1539, opened 2022-03-02
Name wallet metrics according to prometheus guidelines

thomaseizinger:
Metrics should:

- Start with a "namespace" of what they describe (in our case wallet)
- End with the unit of the metric in plural form
- Unit-less metrics should use `_total`

More here: https://prometheus.io/docs/practices/naming.


luckysori (approved):
Nice to figure that one out!

bonomat (approved):
Cool, let's keep this naming in mind 👍

PR #1541, opened 2022-03-02
Introduce metrics for broadcast transactions

thomaseizinger:


daemon/src/monitor.rs
```diff
@@ -1000,6 +1018,48 @@ impl Actor {
...
+            "blockchain_refund_transactions_broadcast_total",
+            "The number of refund transactions broadcast."
+        )
+        .unwrap()
+    });
+static COLLABORATIVE_CLOSE_TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<
+    prometheus::IntCounter,
+> = conquer_once::Lazy::new(|| {
+    prometheus::register_int_counter!(
+        "blockchain_collaborative_close_transactions_broadcast_total",
```
luckysori: ❓ Is the `blockchain_` prefix necessary?

daemon/src/monitor.rs
```diff
@@ -1000,6 +1018,48 @@ impl Actor {
...
+    });
+static REFUND_TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<prometheus::IntCounter> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_int_counter!(
+            "blockchain_refund_transactions_broadcast_total",
+            "The number of refund transactions broadcast."
+        )
+        .unwrap()
+    });
+static COLLABORATIVE_CLOSE_TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<
```
luckysori: ❓ Wasn't it collaborative settlement? Nitty, but it would be good to use consistent terminology.

luckysori (approved):

bonomat (commented):
Why do you think this metric is useful?

daemon/src/monitor.rs
thomaseizinger: You are meant to have a "namespace" prefix and I couldn't come up with a better one. Inputs welcome!

daemon/src/monitor.rs
thomaseizinger: Uhm, this file calls it collaborative close :sweat_smile:

thomaseizinger:
> Why do you think this metric is useful?

This will allow us to answer questions like "How many CFDs got force-closed in time-period X". Same for refunded or new positions opened!

thomaseizinger:
> Together with a query powered by `increase()`, we can accurately measure these counters even across restarts of the service (which resets the counter to 0).
> 
> See [siimon/prom-client#364](https://github.com/siimon/prom-client/issues/364).

With my learnings in https://github.com/itchysats/itchysats/discussions/1542, this statement is actually bogus but the metric is still useful!

daemon/src/monitor.rs
```diff
@@ -1000,6 +1018,48 @@ impl Actor {
     }
 }
 
+static LOCK_TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<prometheus::IntCounter> =
```
thomaseizinger: This should probably use an `IntCounterVec`.

thomaseizinger:
> > Why do you think this metric is useful?
> 
> This will allow us to answer questions like "How many CFDs got force-closed in time-period X". Same for refunded or new positions opened!

We can do the same similar with a log query:

```
{container="maker",setup="mainnet"} |= "Refund transaction published on chain"
```

The problem with this is that it takes _really long_ to sieve through our logs so recording the dedicated events that we care about using a counter is a more performant.

bonomat (approved):
Ok, I can think of how this is useful: 
REcentely I was wondering why so many positions were closed. With a look at this metric it would have been obvious that they were collab-closed and all is good without having to check the logs.

thomaseizinger:
> Ok, I can think of how this is useful: REcentely I was wondering why so many positions were closed. With a look at this metric it would have been obvious that they were collab-closed and all is good without having to check the logs.

We can do even more things with this, we can define growth rates by dividing broadcast lock / close for example.

thomaseizinger:
I think there is a bug still and it is not picked up by CI because we don't broadcast transactions. Testing locally atm.

thomaseizinger:
> I think there is a bug still and it is not picked up by CI because we don't broadcast transactions. Testing locally atm.

Works as expected now:

```
# HELP blockchain_transactions_broadcast_total The number of transactions broadcast.
# TYPE blockchain_transactions_broadcast_total counter
blockchain_transactions_broadcast_total{kind="collaborative-close"} 1
blockchain_transactions_broadcast_total{kind="lock"} 1
```

thomaseizinger:
When defining this metrics, it is essential that we declare all labels ahead of time in the metric definition, otherwise `.with()` will panic at runtime.

da-kami (approved):
Thanks! I am very much in favor of any metrics that can give us the status without having to go through the logs to find out what's going on.

PR #1544, opened 2022-03-02
Update wire protocol to support sending both long and short offers

klochowicz:
This is a refactor, which aimed to preserve the current functionality - the only
difference from API perspective is the nature of choosing whether the maker
wants to go short or long (now it's done by setting `price_long` if maker
wants to go long, before it was choosing `position_maker`).

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -238,23 +238,23 @@ impl Maker {
 
     pub async fn set_offer_params(&mut self, offer_params: maker_cfd::OfferParams) {
         let maker_cfd::OfferParams {
-            price,
+            price_long,
+            price_short,
             min_quantity,
             max_quantity,
             tx_fee_rate,
             funding_rate,
             opening_fee,
-            position_maker,
         } = offer_params;
         self.system
             .set_offer_params(
-                price,
+                price_long,
+                price_short,
                 min_quantity,
                 max_quantity,
                 Some(tx_fee_rate),
                 Some(funding_rate),
                 Some(opening_fee),
-                Some(position_maker),
             )
             .await
             .unwrap();
@@ -423,16 +423,22 @@ pub fn dummy_quote() -> Quote {
     }
 }
 
+// Offer params allowing a single position, either short or long
 pub fn dummy_offer_params(position_maker: Position) -> maker_cfd::OfferParams {
+    let (price_long, price_short) = match position_maker {
+        Position::Long => (Some(Price::new(dummy_price()).unwrap()), None),
+        Position::Short => (None, Some(Price::new(dummy_price()).unwrap())),
+    };
+
     maker_cfd::OfferParams {
-        price: Price::new(dummy_price()).unwrap(),
+        price_long,
+        price_short,
         min_quantity: Usd::new(dec!(5)),
         max_quantity: Usd::new(dec!(100)),
         tx_fee_rate: TxFeeRate::new(1),
         // 8.76% annualized = rate of 0.0876 annualized = rate of 0.00024 daily
         funding_rate: FundingRate::new(dec!(0.00024)).unwrap(),
         opening_fee: OpeningFee::new(Amount::from_sat(2)),
-        position_maker,
     }
 }
 
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -16,9 +16,12 @@ use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
+use model::olivia::BitMexPriceEventId;
+use model::pick_single_offer;
 use model::Cfd;
 use model::FundingRate;
 use model::Identity;
+use model::MakerOffers;
 use model::OpeningFee;
 use model::Order;
 use model::OrderId;
@@ -44,6 +47,11 @@ const HANDLE_ACCEPT_ROLLOVER_MESSAGE_TIMEOUT: std::time::Duration =
     std::time::Duration::from_secs(10);
 const HANDLE_ACCEPT_SETTLEMENT_MESSAGE_TIMEOUT: std::time::Duration =
     std::time::Duration::from_secs(120);
+
+#[derive(Clone, Copy)]
+pub struct NewOffers {
+    pub params: OfferParams,
+}
 
 #[derive(Clone, Copy)]
 pub struct AcceptOrder {
@@ -76,17 +84,6 @@ pub struct RejectRollover {
 }
 
 #[derive(Clone, Copy)]
-pub struct OfferParams {
-    pub price: Price,
-    pub min_quantity: Usd,
-    pub max_quantity: Usd,
-    pub tx_fee_rate: TxFeeRate,
-    pub funding_rate: FundingRate,
-    pub opening_fee: OpeningFee,
-    pub position_maker: Position,
-}
-
-#[derive(Clone, Copy)]
 pub struct TakerConnected {
     pub id: Identity,
 }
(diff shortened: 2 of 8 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -75,25 +83,76 @@ pub struct RejectRollover {
...
-#[derive(Clone, Copy)]
-pub struct TakerConnected {
-    pub id: Identity,
+impl OfferParams {
+    fn pick_oracle_event_id(settlement_interval: Duration) -> BitMexPriceEventId {
+        oracle::next_announcement_after(time::OffsetDateTime::now_utc() + settlement_interval)
+    }
+
+    pub fn create_long_order(&self, settlement_interval: Duration) -> Option<Order> {
+        self.price_long.map(|price_long| {
```
thomaseizinger, line 113: You can `?` `self.price_long` here then you can spare yourself the `map`.

model/src/cfd.rs
```diff
@@ -109,6 +109,62 @@ impl From<Uuid> for OrderId {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let long_order_id = self.long.as_ref().map(|o| o.id);
+        let short_order_id = self.long.as_ref().map(|o| o.id);
+        let tx_fee_rate = self.tx_fee_rate;
+        let funding_rate = self.funding_rate;
+
+        write!(
+            f,
+            "short order_id={short_order_id:?}, long order_id={long_order_id:?}, tx_fee_rate={tx_fee_rate:?}, funding_rate={funding_rate}"
+        )
```
thomaseizinger: Why not use the usual structure of `Debug` here? You can create it with the helper functions on `Formatter`.

daemon/src/maker_cfd.rs
```diff
@@ -75,25 +83,76 @@ pub struct RejectRollover {
...
 
+#[derive(Clone, Copy)]
+pub struct TakerConnected {
+    pub id: Identity,
+}
+
+#[derive(Clone, Copy)]
+pub struct TakerDisconnected {
+    pub id: Identity,
+}
```
thomaseizinger, line 94: Nit: The diff would be a bit more minimal if you wouldn't move these!

model/src/cfd.rs
```diff
@@ -109,6 +109,62 @@ impl From<Uuid> for OrderId {
...
+    }
+}
+
+impl MakerOffers {
+    pub fn take_offer_by_order_id(&self, id: OrderId) -> Option<Order> {
+        if let Some(long) = self.long {
+            if long.id == id {
+                self.long.expect("be still valid").replicate();
```
thomaseizinger: You are dropping the return value of `replicate` here, yes? That seems like a mistake :)

thomaseizinger: I think `take_offer_by_order_id` should not become `&mut self`. We are doing the replicate so we shouldn't need to do it in here.

daemon/src/maker_cfd.rs
```diff
@@ -261,42 +320,43 @@ where
...
-            .send_async_safe(maker_inc_connections::BroadcastOrder(self.current_order))
+            .send_async_safe(maker_inc_connections::BroadcastOffers(self.current_offers))
             .await?;
 
-        #[allow(unused_qualifications)]
-        // Need to fully qualify `Option` because we have more than one `Update` message that
-        // contains an `Option<T>`.
-        self.projection
-            .send(projection::Update(self.current_order))
-            .await?;
```
thomaseizinger, line 299: Lol, I appear to not even read / comprehend my own comments.

thomaseizinger (commented):
Did a first pass, will have a closer look tomorrow morning!

daemon/src/maker_cfd.rs
```diff
@@ -75,25 +83,76 @@ pub struct RejectRollover {
...
+                self.opening_fee,
+            )
+        })
+    }
 }
 
-#[derive(Clone, Copy)]
-pub struct TakerDisconnected {
-    pub id: Identity,
+// If I define an extension trait, can I still send the vanilla one?
```
da-kami: I think this comment can be removed

da-kami (approved):
nice! looking forward to plugging it into projection and the UI

model/src/cfd.rs
klochowicz: > You are dropping the return value of `replicate` here, yes? That seems like a mistake :)

you're right! it used to be `&mut self` for a bit, but this was really hard to comprehend and didn't work well! I forgot to remove it here, removing it now :)

daemon/src/maker_cfd.rs
klochowicz: true, but then the impl block & factory function would be separated. I assumed I picked the lesser evil ;-) I will keep in mind minimising the diffs though for next time.

model/src/cfd.rs
klochowicz: > Why not use the usual structure of `Debug` here? You can create it with the helper functions on `Formatter`.

I'm afraid I don't know about the usual structure of `Debug` - would you mind throwing a link / explaining in more detail? I've only done it to keep our trace log that used to rely on order id working.

daemon/src/maker_cfd.rs
klochowicz: `map` was clippy's suggestion - I made a "manual map implementation" in the first place 😅 

I guess I could, but this way I've got only one expression, and it's more clear IMHO - happy to change it if the other way is more often used around here! :)

rishflab (approved):
Sweet, lgtm. We need to also update the SSE feed to send both orders to the maker and taker. We can add that in a second PR.

model/src/cfd.rs
thomaseizinger: This function is your friend: https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct :)

daemon/src/maker_cfd.rs
thomaseizinger: I was suggesting to do this:

```rust
let price_long = self.price_long?;

Some(Order::new(
	  Position::Long,
	  price_long,
	  self.min_quantity,
	  self.max_quantity,
	  Origin::Ours,
	  Self::pick_oracle_event_id(settlement_interval),
	  settlement_interval,
	  self.tx_fee_rate,
	  self.funding_rate,
	  self.opening_fee,
))
```

thomaseizinger: `?` on an `Option` will bubble up the `None` variant.

klochowicz: Yes, I know - and I said that I prefer sparing the unnecessary assignment, and having a single expression if possible.

model/src/cfd.rs
klochowicz: wow, this is cool - thanks! :)

daemon/src/maker_cfd.rs
thomaseizinger: Okay :)

Using `?` is simpler for my head but I don't feel strongly about it.

daemon-tests/src/lib.rs
```diff
@@ -423,16 +423,22 @@ pub fn dummy_quote() -> Quote {
     }
 }
 
+/// Offer params allowing a single position, either short or long
```
thomaseizinger: Nit: triple-slash is rustdoc and that one has a convention that the first line is a summary description of the documentation: https://brson.github.io/2012/04/14/how-to-rustdoc#the-rules-of-rustdoc

A comment like this I would expect to be double-slash.

maker/src/routes.rs
```diff
@@ -94,10 +94,11 @@ pub async fn maker_feed(
     }
 }
 
-/// The maker POSTs this to create a new CfdOrder
+/// The maker POSTs this to create new Offer params
```
thomaseizinger:
```suggestion
/// The maker POSTs this to set the offer params
```

thomaseizinger (approved):

PR #1546, opened 2022-03-02
Add prometheus counter for sent messages

thomaseizinger:
This should make it easier to create metrics like "orders sent in the last x minutes".


bonomat (approved):
cool, very useful 👍

klochowicz (approved):
this indeed looks useful, thank you! :)

PR #1552, opened 2022-03-03
Finish taker going short

da-kami:
Basically plugged it together and got it working :)

There is still a few things we have to make beautiful in the UI.

On top of this work there is some things we can (and should) refactor. 
The objective was to get it working and then iterate.
We can decide if we want to merge something at some point or refactor some parts (around UI, projection, ...) first.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -944,7 +944,7 @@ pub struct CfdOrder {
     pub margin_per_parcel: Amount,
 
     #[serde(with = "::bdk::bitcoin::util::amount::serde::as_btc")]
-    pub initial_funding_fee_per_parcel: Amount,
+    pub initial_funding_fee_per_parcel: SignedAmount,
 
     #[serde(rename = "leverage")]
     pub leverage_taker: Leverage,
@@ -970,6 +970,22 @@ impl TryFrom<Order> for CfdOrder {
 
         let (long_leverage, short_leverage) =
             long_and_short_leverage(order.leverage_taker, order.origin.into(), own_position);
+
+        let initial_funding_fee_per_parcel = calculate_funding_fee(
+            order.price,
+            parcel_size,
+            long_leverage,
+            short_leverage,
+            order.funding_rate,
+            SETTLEMENT_INTERVAL.whole_hours(),
+        )
+        .context("unable to calculate initial funding fee")?;
+
+        // Use a temporary fee account to define the funding fee's sign
+        let temp_fee_account = FeeAccount::new(own_position, order.origin.into());
+        let initial_funding_fee_per_parcel = temp_fee_account
+            .add_funding_fee(initial_funding_fee_per_parcel)
+            .balance();
 
         Ok(Self {
             id: order.id,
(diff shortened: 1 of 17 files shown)


da-kami:
note to self:

Adapt the tests to reflect the initial funding fee correctly. Previously we did not take the rate's sign into account when sending out the funding fee in projection. Now we do, but the tests don't assert accordingly it seems. 

```
 thread 'taker_receives_order_from_maker_on_publication' panicked at 'assertion failed: `(left == right)`
  left: `CfdOrder { id: OrderId(Hyphenated(4dae1c31-b0d5-469e-a292-67766f33642f)), trading_pair: BtcUsd, position_maker: Short, price: Price(50000), opening_fee: Some(Amount(0.00000002 BTC)), funding_rate_annualized_percent: "8.76", funding_rate_hourly_percent: "0.00100", min_quantity: Usd(5), max_quantity: Usd(100), parcel_size: Usd(100), margin_per_parcel: Amount(0.00100000 BTC), initial_funding_fee_per_parcel: SignedAmount(-0.00000048 BTC), leverage_taker: Leverage(2), liquidation_price: Price(33333.333333333333333333333333), creation_timestamp: Timestamp(1646296591), settlement_time_interval_in_secs: 86400 }`,
 right: `CfdOrder { id: OrderId(Hyphenated(4dae1c31-b0d5-469e-a292-67766f33642f)), trading_pair: BtcUsd, position_maker: Short, price: Price(50000), opening_fee: Some(Amount(0.00000002 BTC)), funding_rate_annualized_percent: "8.76", funding_rate_hourly_percent: "0.00100", min_quantity: Usd(5), max_quantity: Usd(100), parcel_size: Usd(100), margin_per_parcel: Amount(0.00100000 BTC), initial_funding_fee_per_parcel: SignedAmount(0.00000048 BTC), leverage_taker: Leverage(2), liquidation_price: Price(33333.333333333333333333333333), creation_timestamp: Timestamp(1646296591), settlement_time_interval_in_secs: 86400 }`', daemon-tests/tests/happy_path.rs:144:5
```

thomaseizinger (approved):
Nice work!

rishflab (approved):

daemon/src/projection.rs
```diff
@@ -898,6 +921,15 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
     }
 }
 
+/// Maker offers represents the offers as cerated by the maker
```
klochowicz, line 924: nit: `created`

model/src/cfd.rs
```diff
@@ -143,6 +146,27 @@ impl MakerOffers {
         None
     }
 
+    /// Takes the order if available
+    ///
+    /// Resets the order that was taken to None.
```
klochowicz, line 151: instead of resetting it to `None`, this would be more useful if it just called `replicate()` on an Order to create a new one! 

this `None` is not used anywhere, and if we replicated it here, we wouldn't need to replicate the whole lot (I did replication of both just to speed up) . 

I'm happy to do it as a follow-up, it's non-blocking change - just let me know if you disagree :)

klochowicz (approved):
nice one, thank you for holding on for my feedback :) nothing blocking.

model/src/cfd.rs
da-kami: This function is called by the taker when taking an order. The taker cannot replicate orders, because we currently take orders by id, and if the taker replicates the maker would not know anything about the ids being created (the maker would always reject an order replicated by the taker). 
The taker knows that that a take request was sent and hence should set to `None` - but cannot replicate. Eventually we should re-visit the old discussion if it's a good idea to take by order-id...

klochowicz: thanks for the explanation!

PR #1553, opened 2022-03-03
Extract `bitmex_stream` crate

thomaseizinger:
Whilst being tiny, this crate encapsulates the logic for a stable
websocket connection to BitMex in the form of an ergonomic Stream API.

### DIFF
--- a/bitmex-stream/examples/xbt_usd_quote.rs
+++ b/bitmex-stream/examples/xbt_usd_quote.rs
@@ -1,0 +1,17 @@
+use anyhow::Result;
+use futures::TryStreamExt;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    tracing_subscriber::fmt()
+        .with_env_filter("info,bitmex_stream=trace")
+        .init();
+
+    let mut stream = bitmex_stream::subscribe(["quoteBin1m:XBTUSD".to_owned()]);
+
+    while let Some(result) = stream.try_next().await? {
+        tracing::info!("{result}");
+    }
+
+    Ok(())
+}
--- a/bitmex-stream/src/lib.rs
+++ b/bitmex-stream/src/lib.rs
@@ -1,0 +1,65 @@
+use async_stream::stream;
+use futures::SinkExt;
+use futures::Stream;
+use futures::StreamExt;
+use std::time::Duration;
+use tokio_tungstenite::tungstenite;
+
+pub use tokio_tungstenite::tungstenite::Error;
+
+/// Connects to the BitMex websocket API, subscribes to the specified topics (comma-separated) and
+/// yields all messages.
+///
+/// To keep the connection alive, a websocket `Ping` will be sent every 5 seconds in case no other
+/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
+pub fn subscribe<const N: usize>(
+    topics: [String; N],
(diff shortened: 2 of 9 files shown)


xtra-bitmex-price-feed/src/lib.rs
```diff
@@ -33,53 +30,28 @@ impl xtra::Actor for Actor {
...
                             }
-                        },
+                        }
+                        None => {
+                            continue;
+                        }
                     }
                 }
+
+                Ok(())
```
thomaseizinger: This should return `StreamEnded`!

bitmex-stream/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+use tokio_tungstenite::tungstenite::client::IntoClientRequest;
+
+pub use tokio_tungstenite::tungstenite::Error;
+
+/// Connects to the specified BitMex websocket URL and yields all messages.
+///
+/// To keep the connection alive, a websocket `Ping` will be sent every 5 seconds in case no other
+/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
+pub fn new(
+    request: impl IntoClientRequest + Unpin + Send + 'static,
```
thomaseizinger: The API probably makes more sense if we hardcode the `wss://bitmex.com/...` part.

bitmex-stream/examples/xbt_usd_quote.rs
```diff
@@ -0,0 +1,17 @@
+use anyhow::Result;
+use futures::TryStreamExt;
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
luckysori, line 5: 👍 Thanks for adding an example.

bitmex-stream/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub use tokio_tungstenite::tungstenite::Error;
+
+/// Connects to the BitMex websocket API, subscribes to the specified topics (comma-separated) and
+/// yields all messages.
+///
+/// To keep the connection alive, a websocket `Ping` will be sent every 5 seconds in case no other
+/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
+pub fn subscribe(
+    topics: impl AsRef<str> + Send + 'static,
```
luckysori: ❓ Why not a list of strings?

luckysori (approved):
LGTM

bitmex-stream/src/lib.rs
thomaseizinger: Could be a list yeah! I iterated on that API a lot, had a brief thought of making it a list but opted not to because we don't have a use for it at the moment.

You also thinking of it might be a good enough reason to do it though!

thomaseizinger: @luckysori What do you think of the last commit? Worth the extra dependency or should I just go with using `String` and require other callsites to use `to_owned` if they can't produce a `&'static str`?

luckysori: I like the fixup commit! It's true that dependency is not gonna be used much, but if it doesn't have a significant effect on our dependency graph then I'm okay with it.

thomaseizinger: It is another proc-macro that has be compiled before all other dependencies so I think I am gonna go with the allocation instead and remove it again.

luckysori (approved):

PR #1543, opened 2022-03-02
Extract btsieve crate

thomaseizinger:
This time for real.

The overall motivation for this is to improve compile-times across our workspace, test things in isolation, allow for a narrower focus when debugging issues and potential reuse in other / future projects (we have re-written transaction monitoring a couple of times now).

PR #1554, opened 2022-03-03
Make CI more resilient

thomaseizinger:
I encountered some problems with our smoke tests here: https://github.com/itchysats/itchysats/pull/1543

Posting what I think is the fix up as a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,13 +95,15 @@ jobs:
       - name: Setup rust toolchain
         run: rustup show
       - uses: Swatinem/rust-cache@v1.3.0
-      - run: cargo build --bins --tests
+      - run: cargo build --tests --workspace
       - run: cargo test --workspace
       # Ignored tests that should be run on the CI - reenable if there are any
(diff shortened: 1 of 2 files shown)


da-kami (approved):
I think the effort is good and am happy to merge it, but we might still see failures every now and then.

We start the binaries and then wait 10 secs until they are "started" (i.e. rocket is started so we can curl the endpoints). 
If, for whatever reason, rocket does not come up within those 10 seconds we won't be able to curls the endpoint of the binary that is not ready. Can we, and do we want to ensure that rocket always becomes available within 10 seconds in both binaries?

Given that we have our testnet setup now with auto-deploy of master - do we even need the smoke tests? Isn't it good enough to find problems where the binary does not start up in our later test stages?

thomaseizinger:
> Given that we have our testnet setup now with auto-deploy of master - do we even need the smoke tests? Isn't it good enough to find problems where the binary does not start up in our later test stages?

That is a fair point. We can discuss removing them, they have always been a bit of an annoyance :D

bors r+

PR #1555, opened 2022-03-03
Don't set gauges to zero on startup

thomaseizinger:
This produces spikes in the actual graph until we do the first sync.
Grafana can be taught to interpolate between two missing datapoints
which is the right thing to do in this context.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -128,14 +128,6 @@ impl Actor<ElectrumBlockchain> {
             used_utxos: LockedUtxos::new(time_to_lock),
         };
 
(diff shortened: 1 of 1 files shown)


klochowicz (approved):

klochowicz:
nit: did you mean to say Prometheus instead of Grafana?

bonomat (approved):
👍

thomaseizinger:
> nit: did you mean to say Prometheus instead of Grafana?

Nope, I meant Grafana. Prometheus only stores the data, grafana displays it (and interpolates where necessary).

PR #1558, opened 2022-03-04
Create new /offer endpoint in the maker

klochowicz:
Evolve the API to allow setting `Option<Price>` for both short and long price.

The deprecated endpoint ( /order/sell) will be removed after automation upgrade.

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -95,9 +95,10 @@ pub async fn maker_feed(
     }
 }
 
+// TODO: Delete after upgrading automation to prefer /offer endpoint
 /// The maker POSTs this to set the offer params
 #[derive(Debug, Clone, Deserialize)]
-pub struct CfdNewOfferParamsRequest {
+pub struct CfdNewOfferParamsRequestDeprecated {
     #[serde(rename = "price")]
     // TODO: Make this Option<Price> when we don't want to create other side
     pub price_short: Price,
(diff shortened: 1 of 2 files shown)


maker/src/routes.rs
```diff
@@ -140,6 +141,46 @@ pub async fn post_sell_order(
...
 
+/// The maker POSTs this to set the offer params
+#[derive(Debug, Clone, Deserialize)]
+pub struct CfdNewOfferParamsRequest {
+    pub price_long: Option<Price>,
+    pub price_short: Option<Price>,
+    pub min_quantity: Usd,
+    pub max_quantity: Usd,
+    pub tx_fee_rate: Option<TxFeeRate>,
+    pub funding_rate: Option<FundingRate>,
```
thomaseizinger, line 157: Can we rename this to `daily_funding_rate` as per https://github.com/itchysats/itchysats/pull/1559?

maker/src/routes.rs
```diff
@@ -140,6 +141,46 @@ pub async fn post_sell_order(
...
+    // TODO: This is not inline with other parts of the API! We should not expose internal types
+    // here. We have to specify sats for here because of that.
+    pub opening_fee: Option<OpeningFee>,
+}
+
+#[rocket::post("/offer", data = "<offer_params>")]
```
thomaseizinger: I think I would have a mild preference of `PUT` over `POST`.

Conventionally, `PUT` is a _replace_ action. `PUT` is also required to be idempotent which this operation is!

thomaseizinger (approved):

maker/src/routes.rs
klochowicz: I'll give it a shot, thanks for reminding about it!

PR #1559, opened 2022-03-04
Be explicit about the daily funding rate

thomaseizinger:
The API contract states that the value supplied here is the daily
funding rate. We make this clear by renaming the variable but apply
`#[serde(rename]` to avoid breaking the API.


klochowicz (approved):
I'll copy the doc across to #1558.

PR #1561, opened 2022-03-04
Log Rocket requests using tracing macro on debug level

### LINKED ISSUE #1447: Make rocket logs more useful
Currently, rocket logs each invocation of a request:

image

Unfortunately, it uses ANSI escape codes for this. I thought I had fixed this in https://github.com/itchysats/itchysats/pull/1287 but apparently not.

We should at least disable the ANSI escape codes here. Perhaps also consider to either remove this logs entirely and/or add our own request handler goes through our existing `tracing` infrastructure. Can be done using an [`on_request`](https://docs.rs/rocket/0.5.0-rc.1/rocket/fairing/struct.AdHoc.html#method.on_request) rocket fairing.

klochowicz:
Reduce the log spam and allow for easier parsing (Rocket logs used escape codes).

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -282,6 +282,14 @@ async fn main() -> Result<()> {
                 tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
             })
         }))
+        .attach(AdHoc::on_request(
+            "Rocket HTTP request",
+            |request, _data| {
+                Box::pin(async move {
+                    tracing::debug!(%request, "HTTP");
+                })
+            },
+        ))
         .launch()
         .await?;
 
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -315,6 +315,12 @@ async fn main() -> Result<()> {
                 tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
             })
         }))
+        .attach(AdHoc::on_request("Rocket request", |req, _data| {
+            Box::pin(async move {
+                let request = req;
+                tracing::debug!(%request,  "rocket");
+            })
+        }))
         .launch()
         .await?;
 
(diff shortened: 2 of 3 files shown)


maker/src/main.rs
```diff
@@ -281,6 +281,12 @@ async fn main() -> Result<()> {
                 tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
             })
         }))
+        .attach(AdHoc::on_request("Rocket request", |req, _data| {
+            Box::pin(async move {
+                let request = req;
+                tracing::debug!(%request,  "rocket");
```
thomaseizinger: What does this look like in the logs? Perhaps we can use a slightly more generic log message like "HTTP request? or something? I think we are at the point where the usage of rocket is basically an implementation detail from the outside which I am quite fond of.

maker/src/main.rs
```diff
@@ -281,6 +281,12 @@ async fn main() -> Result<()> {
                 tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
             })
         }))
+        .attach(AdHoc::on_request("Rocket request", |req, _data| {
+            Box::pin(async move {
+                let request = req;
```
thomaseizinger: You could rename the closure parameter as well and save yourself this assignment.

thomaseizinger (approved):

luckysori (approved):

maker/src/main.rs
klochowicz: I don't remember, I think it was something like "rocket Request" . I don't mind changing it to "HTTP" though, so I'll just push a fixup.

taker/src/main.rs
```diff
@@ -315,6 +315,12 @@ async fn main() -> Result<()> {
                 tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
             })
         }))
+        .attach(AdHoc::on_request("Rocket request", |req, _data| {
+            Box::pin(async move {
+                let request = req;
+                tracing::debug!(%request,  "rocket");
```
thomaseizinger, line 321: My comment would have applied to both fairings :sweat_smile:

klochowicz: > My comment would have applied to both fairings 😅

indeed! we should come up with a scheme that avoids this unfortunate duplication :( doing a follow-up now!

klochowicz: https://github.com/itchysats/itchysats/pull/1565 - I should have done it properly in the first place ;-)

PR #1562, opened 2022-03-06
Minor UI updates

bonomat:
Commits are not perfect unfortunately. 

image

TODO: fix funding rate. The funding rate in the nav bar might not yet be offer dependent.

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -19,7 +19,7 @@ import relativeTime from "dayjs/plugin/relativeTime";
 import utc from "dayjs/plugin/utc";
 import * as React from "react";
 import { useEffect, useState } from "react";
-import { Navigate, Route, Routes, useLocation } from "react-router-dom";
+import { Navigate, Outlet, Route, Routes, useLocation } from "react-router-dom";
 import { Link as ReachLink } from "react-router-dom";
 import useWebSocket from "react-use-websocket";
 import { useLocalStorage } from "usehooks-ts";
@@ -175,9 +175,93 @@ export const App = () => {
 
     const [hideDisclaimer, setHideDisclaimer] = useLocalStorage<boolean>("hideDisclaimer", false);
 
+    return (
+        <>
+            {!hideDisclaimer && <Disclaimer setHideDisclaimer={setHideDisclaimer} />}
+            <Nav
+                walletInfo={walletInfo}
+                connectedToMaker={connectedToMaker}
+                fundingRate={globalTradeParams?.fundingRateHourly}
+                nextFundingEvent={nextFundingEvent}
+                referencePrice={referencePrice}
+            />
+            <Box textAlign="center" padding={3} bg={useColorModeValue("gray.50", "gray.800")}>
+                <Center marginTop={20}>
+                    <Routes>
+                        <Route path="/">
+                            <Route
+                                path="wallet"
+                                element={<>
+                                    <Wallet walletInfo={walletInfo} />
+                                </>}
+                            />
+                            <Route
+                                element={// @ts-ignore: ts-lint thinks that {children} is missing but react router is taking care of this for us
+                                <PageLayout cfds={cfds} connectedToMaker={connectedToMaker} />}
+                            >
+                                <Route
+                                    path="long"
+                                    element={<>
+                                        <Trade
+                                            order={longOrder}
+                                            globalTradeParams={globalTradeParams}
+                                            connectedToMaker={connectedToMaker}
+                                            walletBalance={walletInfo ? walletInfo.balance : 0}
+                                            isLong={true}
+                                        />
+                                    </>}
+                                />
+                                <Route
+                                    path="short"
+                                    element={<>
+                                        <Trade
+                                            order={shortOrder}
+                                            globalTradeParams={globalTradeParams}
+                                            connectedToMaker={connectedToMaker}
+                                            walletBalance={walletInfo ? walletInfo.balance : 0}
+                                            isLong={false}
+                                        />
+                                    </>}
+                                />
+                            </Route>
+                            <Route index element={<Navigate to="long" />} />
+                        </Route>
+                        <Route
+                            path="/*"
+                            element={<>
+                            </>}
+                        />
+                    </Routes>
+                </Center>
+            </Box>
+            <Footer />
+        </>
+    );
+};
+
+interface PageLayoutProps {
+    children: JSX.Element;
+    cfds: Cfd[];
+    connectedToMaker: ConnectionStatus;
+}
+
+function PageLayout({ children, cfds, connectedToMaker }: PageLayoutProps) {
+    return (<VStack w={"100%"}>
+        <NavigationButtons />
+        <Outlet />
+        <HistoryLayout cfds={cfds} connectedToMaker={connectedToMaker} />
+    </VStack>);
+}
+
+interface HistoryLayoutProps {
+    cfds: Cfd[];
+    connectedToMaker: ConnectionStatus;
+}
+
+function HistoryLayout({ cfds, connectedToMaker }: HistoryLayoutProps) {
     const closedPositions = cfds.filter((cfd) => isClosed(cfd));
 
-    const history = <VStack padding={3} w={"100%"}>
+    return (<VStack padding={3} w={"100%"}>
         <History
             connectedToMaker={connectedToMaker}
             cfds={cfds.filter((cfd) => !isClosed(cfd))}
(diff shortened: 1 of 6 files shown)


da-kami:
Some comments on the idea and design. 
Note: Some ideas are unrelated to this PR but address problems that already existed before. I just thought I'll collect them here and we can potentially address them as part of this PR.

## Collection of small issues

- Displaying the choice of long/short trade screen as Tabs and not Buttons with some color indication corresponding to long and short might look better and make the high level choice clearer. 
- "Position" display on the tile can be improved (could be a badge, ...?)
- The spacing (margin) between the long/short button and the tiles is too small now. The distance should be as least as much as the spacing between the tiles (horizontally). see:

image

## Error handling / Toasts

I am not sure I am a fan of displaying errors as toast in the upper right corner. This might overlap things in the UI
What is the motivation for changing this - What is the advantage of this over bottom middle? 

If we keep this it should be at least displayed with some margin, so it is always below the header bar.

## Colors

I think it is confusing to use green/red for long/short *and* profit/loss. 

Suggestion: We could try making long/short as blue/orance. I think the association with profit and loss as being green and red is stronger. In Ambrosia we had buy/sell as blue/orange and it looked good and was understandable. I don't think we have to use green/red for the position.

This was the looks in Ambrosia:
We can potentially make the colors less dominant ;)

image

image

bonomat:
Hard to discuss in here but I give it a try :) 


> Some comments on the idea and design. Note: Some ideas are unrelated to this PR but address problems that already existed before. I just thought I'll collect them here and we can potentially address them as part of this PR.
> ## Collection of small issues
> 
>     * Displaying the choice of long/short trade screen as Tabs and not Buttons with some color indication corresponding to long and short might look better and make the high level choice clearer.

I went with buttons because I wanted to have static routes (http://localhost:3001/long and http://localhost:3001/short). If it is only for the styling, I think we can style the buttons as well. :) 
We can also go for tabs, but then no routes (I think). 


>     * "Position" display on the tile can be improved (could be a badge, ...?)
> 

No opinion :) 

>     * The spacing (margin) between the long/short button and the tiles is too small now. The distance should be as least as much as the spacing between the tiles (horizontally). see:
> 


better? [932d5177b3ef8d5c9156ea33f9e232ac256af787](https://github.com/itchysats/itchysats/pull/1562/commits/932d5177b3ef8d5c9156ea33f9e232ac256af787)
image



> ## Error handling / Toasts
> 
> I am not sure I am a fan of displaying errors as toast in the upper right corner. This might overlap things in the UI What is the motivation for changing this - What is the advantage of this over bottom middle?

The problem with these alert boxes is that they always move all components around. I find this ugly. 
We can of course have the toast in the bottom middle as well. 

> 
> If we keep this it should be at least displayed with some margin, so it is always below the header bar.
> ## Colors
> 
> I think it is confusing to use green/red for long/short _and_ profit/loss.

happy to go for another color but meh on blue and orange 😬

da-kami:
I played with gradients to indicate the position here: https://github.com/itchysats/itchysats/pull/1564

da-kami:
On colors: I had a look at some other UIs, one pattern that is used there is, that the `green/red` used for `long/short` has a bit more of blue (still looks green) and a bit less saturation than the `green/red` used for `profit/loss`.

We could try that as well. We could also define our own color palette for these; I already started working on that in `theme.tsx` in the past. Not sure if it's worth it to do it now though ;)

da-kami:
One thing I noticed outside my review:

image

 - I am going long, why does the dialog say `Market sell ...` ?
 - Why is the price 0?

taker-frontend/src/App.tsx
```diff
@@ -157,49 +193,46 @@ export const App = () => {
...
+                                        <Wallet walletInfo={walletInfo} />
+                                    </>}
+                                />
+                                <Route
+                                    path="long"
+                                    element={<VStack>
+                                        <NavigationButtons/>
                                         <Trade
                                             longOrder={longOrder}
                                             shortOrder={shortOrder}
```
da-kami: Why would we pass long and short order into the `Trade` component if that component is now either long or short? I think it should just be order now...

taker-frontend/src/App.tsx
```diff
@@ -210,3 +243,51 @@ export const App = () => {
...
+                    size="lg"
+                    h={10}
+                    w={"40"}
+                >
+                    <Text fontSize={"md"} color={buttonText}>Short BTC</Text>
+                </Button>
+            </ButtonGroup>
+        </Center>
+    </HStack>);
+}
```
da-kami: nit: newline missing. Surprised that `dprint` did not fix that.

taker-frontend/src/App.tsx
```diff
@@ -159,7 +156,7 @@ export const App = () => {
 
     useEffect(() => {
         const id = "maker-connection-toast";
-        if (!connectedToMaker.online && !toast.isActive(id)) {
+        if (connectedToMakerOrUndefined && !connectedToMakerOrUndefined.online && !toast.isActive(id)) {
```
da-kami, line 161: Doesn't this mean, if we start the application and never connect to a maker, that our `connectedToMaker` status is set to `{ online: false }` but we never display the toast?

Is that wanted behavior?

taker-frontend/src/App.tsx
```diff
@@ -182,7 +182,7 @@ export const App = () => {
         />
 
         {closedPositions.length > 0
-            && <Accordion allowToggle width={"100%"}>
+            && <Accordion allowToggle width={"80%"}>
```
da-kami: I am not sure I can follow why this is better, but I don't mind much.

da-kami (changes requested):
Thanks for tackling the UI issues and adding improvements. 

---

Some issues I found:

Should be investigated and fixed within this PR:
- It seems the confirmation screen is not correctly initialized, see my comment above

Can be done in a follow-up, but is necessary to ship the taker short feature:

- Given that we have two separate funding rates now, the funding rate has to be applied per order and cannot be part of the global trading params anymore but part of the order. We will likely have to display it inside `Trade` and not in the top-bar.

---

Some overall feedback: 

The commits are not atomic and it was confusing to review this by commit. 
There are quite a lot of changes in commits that are not part of what the commit message describes. Better squash then, otherwise it's just harder to review.

Additionally, some things grew on the way, but are in separate commits. Especially the change from long-/short-order to just order in `Trade` should have been one commit. I wrote down that this should be cleaned up and only noticed later that it was. Better squash this next time, or create a cleaner commit history, otherwise reviewing is a pain.

taker-frontend/src/App.tsx
bonomat: yes, because otherwise we get this flickering behavior: 
1. start: `connectedToMakerOrUndefined = undefined `
2. loaded: connectedToMakerOrUndefined = online = true`

taker-frontend/src/App.tsx
bonomat: In this commit we do not have yet have the split.

taker-frontend/src/App.tsx
bonomat: This is just a small UI change making the accordion not using 100% of the width.

bonomat:
> It seems the confirmation screen is not correctly initialized, see my comment above

fun fact: this was introduced by your branch :P Fixed now.



> Can be done in a follow-up, but is necessary to ship the taker short feature:
> 
>     * Given that we have two separate funding rates now, the funding rate has to be applied per order and cannot be part of the global trading params anymore but part of the order. We will likely have to display it inside `Trade` and not in the top-bar.

let's do this in a follow-up PR: question is, do we want to keep it in the top bar or move into trade?



> The commits are not atomic and it was confusing to review this by commit. There are quite a lot of changes in commits that are not part of what the commit message describes. Better squash then, otherwise it's just harder to review.
> 
> Additionally, some things grew on the way, but are in separate commits. Especially the change from long-/short-order to just order in `Trade` should have been one commit. I wrote down that this should be cleaned up and only noticed later that it was. Better squash this next time, or create a cleaner commit history, otherwise reviewing is a pain.

Imho the commits are good enough to keep. Please have a look and let me know if you want them squashed.

taker-frontend/src/App.tsx
da-kami: I think that was expected (and wanted) behavior. Otherwise you would not see the toast if the maker is offline and you open the UI.

Either way, I think given that we have the the maker's status in color on the top, and we show skeletons for price, etc. if we don't have an offer it might not be necessary to show the toast if we *initially* don't have a connection but only show the toast upon losing the connection.

@klochowicz any opinion on that?

taker-frontend/src/components/ConfirmOrderModal.tsx
```diff
@@ -69,9 +73,10 @@ export default function ConfirmOrderModal({
                 <ModalHeader>
                     <HStack>
                         <Text>
-                            Market sell <b>{quantity}</b> of BTC/USD @
+                            Market {buy_or_sell}
```
da-kami:
```suggestion
                            Market {buy_or_sell}&nbsp;
```

I think the linter / dprint removed that space for you - probably better to add an explicit one with `&nbsp;` otherwise it reads like this:

image

da-kami (approved):
Thanks!

taker-frontend/src/App.tsx
bonomat: > I think that was expected (and wanted) behavior. Otherwise you would not see the toast if the maker is offline and you open the UI.

that's wrong. You will see the toast if you are online and the maker is offline. But you won't see the toast if you are offline. 

you're online meaning: your daemon is running or and you have seen the event. 
If you are not online (your daemon is not running) or have not seen the event yet, you can't reason about the maker being here or not.

taker-frontend/src/components/ConfirmOrderModal.tsx
bonomat: Good catch

taker-frontend/src/App.tsx
bonomat: speaking about this: the same is wrong with the online status in the navbar :)

PR #1563, opened 2022-03-07
Long/short funding rate 

da-kami:
This makes the funding rate dependent on the position, so the maker has more control on what rate apply to rollover.

TODO (not part of this PR):

- use both rates in the UI - when we go for https://github.com/itchysats/itchysats/pull/1562 this is trivial
- Adapt the automation accordingly

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -1,4 +1,5 @@
 use crate::cfd_actors::insert_cfd_and_update_feed;
+use crate::cfd_actors::load_cfd;
 use crate::collab_settlement_maker;
 use crate::command;
 use crate::future_ext::FutureExt;
@@ -99,7 +100,8 @@ pub struct OfferParams {
     pub min_quantity: Usd,
     pub max_quantity: Usd,
     pub tx_fee_rate: TxFeeRate,
-    pub funding_rate: FundingRate,
+    pub funding_rate_long: FundingRate,
+    pub funding_rate_short: FundingRate,
     pub opening_fee: OpeningFee,
 }
 
@@ -119,7 +121,7 @@ impl OfferParams {
                 Self::pick_oracle_event_id(settlement_interval),
                 settlement_interval,
                 self.tx_fee_rate,
-                self.funding_rate,
+                self.funding_rate_long,
                 self.opening_fee,
             )
         })
@@ -136,7 +138,7 @@ impl OfferParams {
                 Self::pick_oracle_event_id(settlement_interval),
                 settlement_interval,
                 self.tx_fee_rate,
-                self.funding_rate,
+                self.funding_rate_short,
                 self.opening_fee,
             )
         })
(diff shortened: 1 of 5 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -478,13 +479,30 @@ impl<O, T, W> Actor<O, T, W> {
...
+                    .funding_rate
+            }
+            Position::Short => {
+                current_offers
+                    .short
+                    .context("Short funding rate not available")?
+                    .funding_rate
+            }
+        };
+
```
klochowicz: IMHO these details should not go from a particular order. you can have order `None`, but still apply funding rate, allowing already established Cfds to keep rolling over.

klochowicz (approved):
LGTM! 👍 

I'll now focus on making the automation actually prefer the new endpoint , as we add new feature finally :)

daemon/src/maker_cfd.rs
klochowicz: as per conversations in element, we moved to having 2 global funding rates (long and short). I'll remove the daily_funding_rate from within `Order` in a follow-up commit.

maker/src/routes.rs
```diff
@@ -108,12 +108,10 @@ pub struct CfdNewOfferParamsRequestDeprecated {
...
-    /// The _daily_ funding rate as decided upon by the caller.
-    ///
-    /// The fact that this is the _daily_ funding rate is part of the API contract between this
-    /// application and its caller. Changing this would be a breaking change.
-    #[serde(rename = "funding_rate")]
-    pub daily_funding_rate: Option<FundingRate>,
+    /// The current _daily_ funding rate for the maker's long position
+    pub daily_funding_rate_long: Option<FundingRate>,
+    /// The current _daily_ funding rate for the maker's short position
+    pub daily_funding_rate_short: Option<FundingRate>,
```
thomaseizinger: Don't change this in the deprecated one please, it will break the automation.

model/src/cfd.rs
```diff
@@ -123,7 +124,8 @@ impl fmt::Debug for MakerOffers {
             .field("long_order_id", &self.long.as_ref().map(|o| o.id))
             .field("short_order_id", &self.long.as_ref().map(|o| o.id))
             .field("tx_fee_rate", &self.tx_fee_rate)
-            .field("funding_rate", &self.funding_rate)
+            .field("funding_rate", &self.funding_rate_long)
+            .field("funding_rate", &self.funding_rate_short)
```
thomaseizinger: That overrides the field, you need to use 2 different field names!

thomaseizinger (changes requested):
Found two bugs, otherwise LGTM!

maker/src/routes.rs
da-kami: Right, the depricated API can only add, but not change values that were previously there.

thomaseizinger (approved):

PR #1564, opened 2022-03-07
Use thin gradient to indicate the position of a tile

da-kami:
Just wanted to play a bit with that idea, this is how it looks:

image

image

### DIFF
--- a/taker-frontend/src/components/History.tsx
+++ b/taker-frontend/src/components/History.tsx
@@ -18,7 +18,7 @@ import {
     VStack,
 } from "@chakra-ui/react";
 import * as React from "react";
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
I like the idea but I am not sure if it should replace the textual part. Unless you explain this to users somehow, it won't be very discoverable.

klochowicz:
> I like the idea but I am not sure if it should replace the textual part. Unless you explain this to users somehow, it won't be very discoverable.

yes, the discoverability should be changed. I also think that if you want to encode position in the gradient, you could try using different colours - with this gradient, I immediately was thinking that you're depicting positions that are "winning" or "losing" (as similar colours were used to denote that).

da-kami:
closing this for now until I have a better design :)

PR #1567, opened 2022-03-07
Add libp2p-xtra crate

### LINKED ISSUE #1126: Use a multiplexing library to simplify networking code
For example yamux: https://docs.rs/yamux

General idea:

1. Connection actor on each side tracks the TCP connection
2. Wait for incoming streams
3. Negotiate use of new stream via `multistream-select`, same as in https://github.com/thomaseizinger/xtra-quinn-prototype
4. Spawn new actors based on negotiated protocol and hand both ends of the stream to it

thomaseizinger:
This crate provides a single actor called `Endpoint` which implements a
simple, yet powerful abstraction for managing network connections. Built
on top of libp2p-core, every connection is fully encrypted and
authenticated with noise and multiplexed with yamux.

Closes #45. (Still needs to be adopted but the base work is done.)

The code has been written and tested in isolation and I think it is ready for (temporary) inclusion here. Once we've adopted it for the network layer and found it sufficiently mature enough, we should extract it into its own repository again!

### DIFF
--- a/libp2p-xtra/src/upgrade.rs
+++ b/libp2p-xtra/src/upgrade.rs
@@ -1,0 +1,152 @@
+use crate::verify_peer_id::VerifyPeerId;
+use crate::Connection;
+use futures::channel::mpsc;
+use futures::future;
+use futures::AsyncRead;
+use futures::AsyncWrite;
+use futures::FutureExt;
+use futures::SinkExt;
+use futures::StreamExt;
+use libp2p_core::identity::Keypair;
+use libp2p_core::transport::timeout::TransportTimeout;
+use libp2p_core::transport::Boxed;
+use libp2p_core::upgrade;
+use libp2p_core::upgrade::Version;
+use libp2p_core::Endpoint;
+use libp2p_core::Transport;
+use libp2p_noise as noise;
+use multistream_select::NegotiationError;
+use std::time::Duration;
+use void::Void;
+use yamux::Mode;
+
+/// Upgrades the given [`Transport`].
+///
+/// We apply:
+/// - Noise encryption and authentication
+/// - PeerID verification for each connection
+/// - Yamux multiplexing
+/// - Connection upgrade timeout
+pub fn transport<T>(
+    transport: T,
+    identity: &Keypair,
+    supported_inbound_protocols: Vec<&'static str>,
+    connection_timeout: Duration,
+) -> Boxed<Connection>
+where
+    T: Transport + Clone + Send + Sync + 'static,
+    T::Output: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    T::Error: Send + Sync,
+    T::Listener: Send + 'static,
+    T::Dial: Send + 'static,
+    T::ListenerUpgrade: Send + 'static,
+{
+    let identity = noise::Keypair::<noise::X25519Spec>::new()
+        .into_authentic(identity)
+        .expect("ed25519 signing does not fail");
+
+    let authenticated = transport.and_then(|conn, endpoint| {
+        upgrade::apply(
+            conn,
+            noise::NoiseConfig::xx(identity).into_authenticated(),
+            endpoint,
+            Version::V1,
+        )
+    });
+
+    let peer_id_verified = VerifyPeerId::new(authenticated);
+
+    let multiplexed = peer_id_verified.and_then(|(peer_id, conn), endpoint| {
+        upgrade::apply(
+            conn,
+            upgrade::from_fn::<_, _, _, _, _, Void>(
+                b"/yamux/1.0.0",
+                move |conn, endpoint| async move {
+                    Ok(match endpoint {
+                        Endpoint::Dialer => (
+                            peer_id,
+                            yamux::Connection::new(conn, yamux::Config::default(), Mode::Client),
+                        ),
+                        Endpoint::Listener => (
+                            peer_id,
+                            yamux::Connection::new(conn, yamux::Config::default(), Mode::Server),
+                        ),
+                    })
+                },
+            ),
+            endpoint,
+            Version::V1,
+        )
+    });
+
+    let protocols_negotiated = multiplexed.map(move |(peer, mut connection), _| {
+        let control = connection.control();
+
+        let (mut sender, receiver) = mpsc::channel(5); // Use a bounded channel to allow caller to exercise back-pressure.
+
+        let worker = async move {
+            while let Ok(Some(stream)) = connection.next_stream().await {
+                match future::poll_fn(|cx| sender.poll_ready(cx)).await {
+                    Ok(()) => {}
+                    Err(e) => {
+                        if e.is_disconnected() {
+                            break; // Consumer is no longer interested
+                        }
+
+                        debug_assert!(
+                            !e.is_full(),
+                            "We checked that there is space in the queue with `poll_ready`"
+                        )
+                    }
+                }
+
+                match sender.send(stream).await {
+                    Ok(()) => {}
+                    Err(e) => {
+                        if e.is_disconnected() {
+                            break; // Consumer is no longer interested
+                        }
+
+                        debug_assert!(
+                            !e.is_full(),
+                            "We checked that there is space in the queue with `poll_ready`"
+                        )
+                    }
+                }
+            }
+        }
+        .boxed();
+
+        let incoming = receiver
+            .then(move |stream| {
+                let supported_protocols = supported_inbound_protocols.clone();
+
+                async move {
+                    let result = tokio::time::timeout(
+                        connection_timeout,
+                        multistream_select::listener_select_proto(stream, &supported_protocols),
+                    )
+                    .await;
+
+                    match result {
+                        Ok(Ok((protocol, stream))) => Ok(Ok((stream, *protocol))),
+                        Ok(Err(e)) => Ok(Err(Error::NegotiationFailed(e))),
+                        Err(_timeout) => Ok(Err(Error::NegotiationTimeoutReached)),
+                    }
+                }
+            })
+            .boxed();
+
+        (peer, control, incoming, worker)
+    });
+
+    TransportTimeout::new(protocols_negotiated, connection_timeout).boxed()
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Timeout in protocol negotiation")]
+    NegotiationTimeoutReached,
+    #[error("Failed to negotiate protocol")]
+    NegotiationFailed(#[from] NegotiationError),
+}
--- a/libp2p-xtra/src/verify_peer_id.rs
+++ b/libp2p-xtra/src/verify_peer_id.rs
@@ -1,0 +1,208 @@
+use crate::multiaddress_ext::MultiaddrExt;
+use futures::future::BoxFuture;
+use futures::stream::BoxStream;
+use futures::FutureExt;
+use futures::StreamExt;
+use futures::TryFutureExt;
+use futures::TryStreamExt;
+use libp2p_core::transport::ListenerEvent;
+use libp2p_core::transport::TransportError;
+use libp2p_core::Multiaddr;
+use libp2p_core::PeerId;
+use libp2p_core::Transport;
+use std::fmt;
+use std::fmt::Debug;
+
+/// A [`Transport`] wrapper that ensures we only connect to the expected peer.
+///
+/// Multi-addresses can contain a `/p2p` segment that specifies, which peer we want to connect to,
+/// for example: `/memory/10000/p2p/12D3KooWSLdEVWR1rjrnimdX3KwTvRT8uxNs8q7keREr6MsuizJ7`
+/// However, the `/p2p` part is not actually necessary to establish the physical connection. As
+/// such, it can happen that a different peer sits at the end of an address like `/memory/10000`.
+/// `libp2p-core` does not prevent this situation from happening. It is the responsibility of the
+/// caller of [`Transport::dial`] to verify that we connected to the expected peer.
+///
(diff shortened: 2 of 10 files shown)


libp2p-xtra/src/upgrade.rs
```diff
@@ -0,0 +1,125 @@
...
+                    })
+                },
+            ),
+            endpoint,
+            Version::V1,
+        )
+    });
+
+    let protocols_negotiated = multiplexed.map(move |(peer, mut connection), _| {
+        let control = connection.control();
```
rishflab, line 83: question: what is a connection control?

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// 1. Connection upgrades (i.e. noise handshake, yamux upgrade, etc)
+    /// 2. Protocol negotiations
+    ///
+    /// The provided substream handlers are actors that will be given the fully-negotiated
+    /// substreams whenever a peer opens a new substream for the provided protocol.
+    pub fn new<T, const N: usize>(
+        transport: T,
+        identity: Keypair,
+        connection_timeout: Duration,
+        inbound_substream_handlers: [(
```
rishflab: Would it be worth making this argument a hashmap to let the user know each handler should be unique. It has to be converted to a hashmap anyway?

libp2p-xtra/src/upgrade.rs
thomaseizinger: This is where we enter `yamux` land!

Here are some docs:

- https://docs.rs/yamux/latest/yamux/struct.Control.html
- https://github.com/hashicorp/yamux/blob/master/spec.md

libp2p-xtra/src/node.rs
thomaseizinger: I've chosen to make it an array because I found it more ergonomic to construct. But yes, I may need some documentation that points out that these have to be unique mappings.

thomaseizinger: Should probably fail if we find a duplicate one.

thomaseizinger: Added a check and a test for it.

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        let stream = control.open_stream().await?;
+
+        let result = tokio::time::timeout(self.connection_timeout, async {
+            let (protocol, stream) =
+                multistream_select::dialer_select_proto(stream, protocols, Version::V1).await?;
+
+            Ok((protocol, stream))
+        })
+        .await;
```
thomaseizinger: Can be shortened.

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,504 @@
...
+
+        let (protocol, stream) = match result {
```
thomaseizinger: Use map_err.

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,504 @@
...
+                    .await;
+            },
+        );
+    }
+
+    async fn handle(&mut self, msg: OpenSubstream<Single>) -> Result<Substream, Error> {
+        let peer = msg.peer;
+        let protocols = msg.protocols;
+
+        let (_, stream) = self.open_substream(peer, protocols).await?;
```
thomaseizinger: Add a debug assert on the protocol.

thomaseizinger:
Tagging everyone as reviewer, please ask questions for anything that needs more documentation!

libp2p-xtra/src/verify_peer_id.rs
```diff
@@ -0,0 +1,195 @@
...
+    async fn fails_dial_with_unexpected_peer_id() {
+        let _alice = MemoryTransport::default()
+            .map(simulate_auth_upgrade)
+            .listen_on("/memory/10000".parse().unwrap())
+            .unwrap();
+        let bob = VerifyPeerId::new(MemoryTransport::default().map(simulate_auth_upgrade));
+
+        let result = bob
+            .dial(
+                "/memory/10000/p2p/12D3KooWSLdEVWR1rjrnimdX3KwTvRT8uxNs8q7keREr6MsuizJ7" // Hardcoded PeerId will unlikely be equal to random one that is simulated in
```
luckysori: 😄 "unlikely"

libp2p-xtra/src/verify_peer_id.rs
```diff
@@ -0,0 +1,195 @@
...
+                    // auth upgrade.
```
luckysori: ❓ Is this meant to be part of the previous line?

libp2p-xtra/src/verify_peer_id.rs
```diff
@@ -0,0 +1,195 @@
...
+use libp2p_core::transport::ListenerEvent;
+use libp2p_core::transport::TransportError;
+use libp2p_core::Multiaddr;
+use libp2p_core::PeerId;
+use libp2p_core::Transport;
+use std::fmt;
+use std::fmt::Debug;
+
+#[derive(Clone)]
+pub struct VerifyPeerId<TInner> {
```
luckysori, line 29: 🙃 I think it would be good if we had some documentation on this struct.

libp2p-xtra/src/upgrade.rs
```diff
@@ -0,0 +1,125 @@
...
+
+        let (mut sender, receiver) = mpsc::unbounded();
+
+        let worker = async move {
+            while let Ok(Some(stream)) = connection.next_stream().await {
+                let _ = sender.send(stream).await; // ignore error for now.
```
luckysori: ❓ Why is this okay? Is this something that we're supposed to deal with at some point?

libp2p-xtra/src/upgrade.rs
```diff
@@ -0,0 +1,125 @@
...
+        }
+        .boxed();
+
+        let incoming = receiver
+            .then(move |stream| {
+                let supported_protocols = supported_inbound_protocols.clone();
+
+                async move {
+                    let result = tokio::time::timeout(
+                        connection_timeout,
```
luckysori, line 126: ❓ Is it intentional that we're using the same timeout here and for the `TransportTimeout` constructor?

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,508 @@
...
+use std::time::Duration;
+use thiserror::Error;
+use tokio_tasks::Tasks;
+use xtra::message_channel::StrongMessageChannel;
+use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncSafe;
+
+/// An actor for managing multiplexed connections over a given transport.
+///
+/// The actor does not inflict any policy on connection and/or protocol management.
```
luckysori: 🙃 I think "inflict" doesn't quite fit here. We could for example use:

```suggestion
/// The actor does not impose any policy on connection and/or protocol management.
```

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,508 @@
...
+    peer: PeerId,
+    protocols: Vec<&'static str>,
+    marker_num_protocols: PhantomData<P>,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum Single {}
+
+#[derive(Clone, Copy, Debug)]
+pub enum Multiple {}
```
luckysori: ❓ I would expect these to be `struct`s. Any reason why that is not the case?

luckysori (approved):
Amazing work!

libp2p-xtra/src/verify_peer_id.rs
thomaseizinger: Yes.

libp2p-xtra/src/verify_peer_id.rs
thomaseizinger: Can document yes :)

libp2p-xtra/src/upgrade.rs
thomaseizinger: I think we should actually:

a) use a bounded channel to apply back-pressure
b) break from this loop if we fail to send (which can only fail because the receiver is down)

libp2p-xtra/src/upgrade.rs
thomaseizinger: Yes, I've tried to simplify the API by combining these here.

libp2p-xtra/src/node.rs
thomaseizinger: Because they should never be instantiated really. These is the poor-man's version of dependent types which Rust does not have (yet) :D

Ideally, I'd directly tie the return value of the `OpenSubstream` message to the length of `protocols` array: If only 1 protocol, then I don't need to return the negotiated protocol, if more protocols, I need to say which one was negotiated.

libp2p-xtra/src/verify_peer_id.rs
```diff
@@ -0,0 +1,208 @@
...
+use libp2p_core::transport::TransportError;
+use libp2p_core::Multiaddr;
+use libp2p_core::PeerId;
+use libp2p_core::Transport;
+use std::fmt;
+use std::fmt::Debug;
+
+/// A [`Transport`] wrapper that ensures we only connect to the expected peer.
+///
+/// Multi-addresses can contain a `/p2p` segment that specifies, which peer we want to connect to,
```
luckysori, line 18: 🙃 Writing nit-pick. I'm pretty certain you should not separate the verb from its subject with a comma. Random source: https://www.grammarly.com/blog/comma-with-subjects-and-verbs/.
```suggestion
/// Multi-addresses can contain a `/p2p` segment that specifies which peer we want to connect to,
```

This pause can appear orally of course.

luckysori (approved):
Thanks for incorporating some of my suggestions and for the extra improvements!
I like the way you handled the last commit with the use of `poll_ready` and the `debug_assert`s.

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,508 @@
...
+struct FailedToConnect {
+    peer: PeerId,
+    error: anyhow::Error,
+}
+
+#[derive(Debug)]
+struct ConnectionFailed {
+    peer: PeerId,
+    error: anyhow::Error,
+}
```
klochowicz: would you mind adding some comments here, and possibly improve naming? they are almost the same, but not quite - after careful look at the handler, one seems to be concerned with inflight requests.

libp2p-xtra/src/node.rs
```diff
@@ -0,0 +1,508 @@
...
+/// The combination of the above should make it possible to implement a fairly large number of
+/// policies. For example, to maintain a connection to a specific node, you can regularly check if
+/// the connection is still established by sending [`GetConnectionStats`] and react accordingly
+/// (f.e. sending [`Connect`] in case the connection has disappeared).
+///
+/// Once a connection with a peer is established, both sides can open substreams on top of the
+/// connection. Any incoming substream will - assuming the protocol is supported by the node -
+/// trigger a [`NewInboundSubstream`] message to the actor provided in the constructor.
+/// Opening a new substream can be achieved by sending the [`OpenSubstream`] message.
+pub struct Node {
```
klochowicz: would you be open to renaming this super-generic actor to something better? `ConnectionManager`, `Connector`, something along these lines? Open to other ones too, the current name makes it hard to reason IMHO

klochowicz (approved):
thanks, this review took me into a rabbit hole of trying to understand libp2p a bit better, as I haven't used it. After reading their docs, this PR was a bit more manageable :)

seems like a great piece of work! I'm excited to try it out! 👍

libp2p-xtra/src/node.rs
thomaseizinger: Fair criticism. Looking at some prior art, QUIC implemenations name this thing an `Endpoint`. Thoughts on that?

I think `ConnectionManager` is not a great name because it does much more than manage connections for you. IMO, the core feature of this component is the ability to obtain a multiplexed, virtual stream to the other party. The fact that a physical connection is required for that is explicitly moved into the background.

Because of this abstraction, I named it `Node`. Similar to how a Bitcoin Node gives you a relatively easy to use interface to the Bitcoin network. Node also has this co-notation of a p2p network which I wanted to express.

Some names that come to my mind (none of which I am super happy with, but maybe they help the brain-storming).

- `StreamPortal` (a bit funky)
- `Network` (too broad)
- `Endpoint` (as suggested above)
- `NetworkStreamManager` (a bit long)
- `PeerManager` (I am not a massive fan of "Manager" names haha)

klochowicz: I like `Endpoint` much more! I can get convinced by the prior art argument. 
`StreamPortal` is also good, as it describes essentially what it does. 

The upside value over `Network` and `Node` is that these names are not overloaded by Bitcoin-related terms

thomaseizinger: Renamed to `Endpoint`!

libp2p-xtra/src/node.rs
thomaseizinger: They are pretty different in fact:

- `FailedToConnect`: We failed to connect (i.e. establish a connection).
- `ConnectionFailed`: An (existing) connection failed.

It is pretty obvious to me. Given that they are internal messages, I wouldn't worry too much but feel free to add comments to it!

klochowicz: would you mind if I renamed `s/ConnectionFiled/ExistingConnectionFailed` ? then they wouldn't need comments at all, and would be more self-descriptive. Happy to push a fixup if you agree. :)

thomaseizinger: It is redundant in my opinion, especially in the presence of `FailedToConnect` :man_shrugging:

I'd like to merge it at this point, feel free to send a follow-up :)

klochowicz: sure. incoming a follow-up.

PR #1570, opened 2022-03-08
Add a new shared_bin log target

klochowicz:
Recent refactor moved the HTTP request logging into a different crate, which
requires a new log target defined for the logs to be visible.


da-kami (approved):

shared-bin/src/logger.rs
```diff
@@ -23,6 +23,7 @@ pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
         .add_directive("rocket::launch_=off".parse()?) // disable rocket startup logs
         .add_directive("rocket::shield=off".parse()?) // not sure what rocket shield is but it seems unnecessary
         .add_directive("rocket::server=off".parse()?) // disable rocket request logs (along with potentially other logs)
+        .add_directive(format!("shared_bin={level}").parse()?)
```
thomaseizinger, line 26: A different solution to this would be to define the `target` together with the event:

```rust
tracing::info!(target: "http", "HTTP interface ready");
```

I think I'd mildly prefer that over `shared_bin`. With this, we are leaking our crate structure - which is purely a source code level concern - to a runtime level because the log-parsing etc will have to know about these targets.

PR #1577, opened 2022-03-09
Update maia after introduction of long-short terminology

### LINKED ISSUE #1482: Transition maker/taker to long/short where possible
Requires https://github.com/comit-network/maia/issues/16.

### LINKED ISSUE #1580: Calculate the payout correcly when the taker is short
As @luckysori pointed out we are not using https://github.com/itchysats/itchysats/blob/long-story-short/model/src/cfd.rs#L1900-L1922 in the code, which means that the payout for taker=short/maker=long is calculated wrongly. 

Proposal for fix:
- Create failing test(s): Extend the tests to cover taker=short/maker=long and assert on the values so we know that this is wrong 
- Create another abstraction on top of `calculate_payouts_long_taker` and `calculate_payouts_short_taker` that knows about the role and then delegate to the functions as necessary. I thought about actually having this functionality on the `model::Cfd` - might be worth exploring - but the contract_setup currently does not know about the `model::Cfd`

luckysori:
Fixes #1580.
Fixes #1482.

Blocked by https://github.com/comit-network/maia/issues/16. (We have to amend the last commit and depend on latest `maia` after https://github.com/comit-network/maia/pull/21 is merged).

With this change we also fix the fact that we were (wrongly) only using the `calculate_payouts_long_taker` API. With the new semantics, we can always call the generic `calculate_payouts` API since it accepts arguments based on long/short.

The main drawback of this change is that we've elected to introduce the `Position` into the `Dlc` which is now duplicated across `Cfd` and `Dlc`. This may not be the final solution, but it helps in being able to use ours/theirs and long/short semantics with the data inside of `Dlc`.

The hardest pill to swallow might be the changes to the tests in `cfd.rs`, where a few test APIs have been modified to be able to pass through the `Position`.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -918,11 +918,15 @@ impl Cfd {
         proposal: SettlementProposal,
         sig_taker: Signature,
     ) -> Result<CollaborativeSettlement> {
+        anyhow::ensure!(
+            self.role == Role::Maker,
+            "Only the maker can create complete collaborative settlement signing"
+        );
+
         let dlc = self
             .dlc
             .as_ref()
-            .context("dlc has to be available for collab settlemment")?
-            .clone();
+            .context("Collaborative close without DLC")?;
 
         let (tx, sig_maker) = dlc.close_transaction(&proposal)?;
         let spend_tx = dlc.finalize_spend_transaction((tx, sig_maker), sig_taker)?;
@@ -1309,6 +1313,11 @@ impl Cfd {
         &self,
         proposal: &SettlementProposal,
     ) -> Result<(Transaction, Signature, Script)> {
+        anyhow::ensure!(
+            self.role == Role::Taker,
+            "Only the taker can start collaborative settlement signing"
+        );
+
         let dlc = self
             .dlc
             .as_ref()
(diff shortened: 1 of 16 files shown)


model/src/cfd.rs
```diff
@@ -960,8 +960,8 @@ impl Cfd {
         let proposal = SettlementProposal {
             order_id: self.id,
             timestamp: Timestamp::now(),
-            taker: *payout.taker_amount(),
-            maker: *payout.maker_amount(),
+            taker: *payout.long_amount(),
+            maker: *payout.short_amount(),
```
thomaseizinger: Well, this is not going to be accurate now is it? :D

luckysori: Hehe it looks wrong, but it is in the context of having called:

https://github.com/itchysats/itchysats/pull/1577/files/e75b61cdf4f0857a389e67e00c415d97bd54615d#diff-f078e5601c18fcb29f773d04d16902354c8f66a57b41aedaea122e11a6f8fd3dR943-R950

So it would actually work here.

In any case, I'm still working on this PR because there are other mistakes.

thomaseizinger (changes requested):
I think this is a good change but we cannot merge this because it breaks deserialization of old events and we haven't done https://github.com/itchysats/architecture/issues/3 yet.

Unless we figure out, how to make such changes we will need to band-aid the payout selection problem in a way that doesn't touch events.

PR #1578, opened 2022-03-09
Allow price to be outdated by 2 * quote-interval

thomaseizinger:
Putting the threshold directly at the interval refresh rate is prone
to being triggered erroneously when the quote just hasn't updated yet.

Allowing for twice the refresh interval should fix this.

To allow for easier monitoring of this problem, we also print a log
including the timestamp of the quote.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -421,17 +421,25 @@ where
             .context("Price feed not available")?
             .context("No quote available")?;
 
-        if latest_quote.is_older_than(QUOTE_INTERVAL_MINUTES.minutes()) {
+        let current_price = Price::new(latest_quote.for_taker())?;
+        let quote_timestamp = latest_quote
+            .timestamp
+            .format(&time::format_description::well_known::Rfc3339)
+            .context("Failed to format timestamp")?;
+
+        tracing::debug!(%order_id, %current_price, %quote_timestamp, "Proposing settlement of contract");
+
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Okay, I believe in this solution.

daemon/src/lib.rs
```diff
@@ -421,17 +421,25 @@ where
...
+            .format(&time::format_description::well_known::Rfc3339)
+            .context("Failed to format timestamp")?;
+
+        tracing::debug!(%order_id, %current_price, %quote_timestamp, "Proposing settlement of contract");
+
+        if latest_quote.is_older_than(QUOTE_INTERVAL_MINUTES.minutes() * 2) {
             anyhow::bail!(
                 "Latest quote is older than {} minutes. Refusing to settle with old price.",
-                QUOTE_INTERVAL_MINUTES
+                QUOTE_INTERVAL_MINUTES * 2
```
klochowicz, line 435: nit: this sort of duplication is prone to rot (if we ever decide to change the solution to change the multiplier, there's high likelyhood that the log won't get updated and will be actively misleading.

I'd suggest extracting `let quote_threshold = QUOTE_INTERVAL_MINUTES * 2;` and then using it in both places

klochowicz (approved):
this should please the affected users & the auto-taker too :) good one!

daemon/src/lib.rs
thomaseizinger: Agreed. Will send a follow-up but merge this now so autotaker can close positions again!

PR #1581, opened 2022-03-09
Some improvements to our logs

thomaseizinger:
Motivated as part of going through the logs with my production hat on.

### DIFF
--- a/shared-bin/src/fairings.rs
+++ b/shared-bin/src/fairings.rs
@@ -11,16 +11,20 @@ pub fn log_launch() -> impl Fairing {
                 rocket.config().port
             );
 
-            tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
+            tracing::info!(target: "http", endpoint = %http_endpoint, "HTTP interface is ready");
         })
     })
 }
 
 /// Attach this fairing to enable logging Rocket HTTP requests
 pub fn log_requests() -> impl Fairing {
-    AdHoc::on_request("Rocket HTTP request", |request, _data| {
+    AdHoc::on_response("Log status code for request", |request, response| {
         Box::pin(async move {
-            tracing::debug!(%request, "HTTP");
+            let method = request.method();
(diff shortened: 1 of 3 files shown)


shared-bin/src/fairings.rs
```diff
@@ -11,16 +11,20 @@ pub fn log_launch() -> impl Fairing {
...
-            tracing::info!(endpoint = %http_endpoint, "HTTP interface is ready");
+            tracing::info!(target: "http", endpoint = %http_endpoint, "HTTP interface is ready");
         })
     })
 }
 
 /// Attach this fairing to enable logging Rocket HTTP requests
 pub fn log_requests() -> impl Fairing {
-    AdHoc::on_request("Rocket HTTP request", |request, _data| {
+    AdHoc::on_response("Log status code for request", |request, response| {
```
luckysori, line 21: ❓ Is this meant to remain as `on_request`?

luckysori (approved):
LGTM except for one thing that might be wrong.

shared-bin/src/fairings.rs
thomaseizinger: Nope, needs to be `on_response` so we can get the status code!

luckysori: You're right, I got confused with the documentation.

klochowicz, line 21: nice one! I didn't notice that fairing that gives the status code!

nit: you might want to adjust the comment and perhaps the name to make it clear that it also logs response codes :)

klochowicz (approved):
nice improvements!

shared-bin/src/fairings.rs
thomaseizinger: > nit: you might want to adjust the comment and perhaps the name to make it clear that it also logs response codes :)

Yeah, the current one is not wrong per so, just not as complete as it could be. I don't think it matters much so I'll leave it!

PR #1582, opened 2022-03-09
Two funding rates taker UI

da-kami:
image

image

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -29,26 +29,24 @@ import History from "./components/History";
 import Nav from "./components/NavBar";
 import Trade from "./components/Trade";
 import { Wallet } from "./components/Wallet";
-import { BXBTData, Cfd, ConnectionStatus, intoCfd, intoOrder, isClosed, Order, WalletInfo } from "./types";
+import { BXBTData, Cfd, ConnectionStatus, intoCfd, intoMakerOffer, isClosed, MakerOffer, WalletInfo } from "./types";
 import { useEventSource } from "./useEventSource";
 import useLatestEvent from "./useLatestEvent";
 
-export interface SomeOrder {
+export interface Offer {
     id?: string;
     price?: number;
     marginPerParcel?: number;
     initialFundingFeePerParcel?: number;
-}
+    liquidationPrice?: number;
+    fundingRateAnnualized?: number;
+    fundingRateHourly?: number;
 
-export interface GlobalTradeParams {
+    // defaulted for display purposes
     minQuantity: number;
     maxQuantity: number;
     parcelSize: number;
     leverage: number;
-    openingFee: number;
-    fundingRateAnnualized: string;
-    fundingRateHourly: string;
-    liquidationPrice?: number;
 }
 
 export const App = () => {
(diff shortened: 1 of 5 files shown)


da-kami:
fun fact: https://github.com/itchysats/itchysats/issues/1580 is visible here, see the payout of the short position ;)

da-kami:
The "funding rate" display and behavior will definitely require some more documentation in the UI. 
It is not clear how it behaves from just seeing the values. It is also not clear that is actually can be - or + and fluctuates over time. Furthermore, we should somehow explain why there are two different ones for long and short. 
With @bonomat we thought about calling it interest rate instead, but that will obviously not explain the details ;)

taker-frontend/src/App.tsx
```diff
@@ -181,7 +154,7 @@ export const App = () => {
             <Nav
                 walletInfo={walletInfo}
                 connectedToMaker={connectedToMaker}
-                fundingRate={globalTradeParams?.fundingRateHourly}
+                fundingRate={undefined}
```
thomaseizinger: Why not just delete the property?

thomaseizinger (approved):
Nice!

Commits should be squashed IMO.

thomaseizinger:
> The "funding rate" display and behavior will definitely require some more documentation in the UI.
> It is not clear how it behaves from just seeing the values. It is also not clear that is actually can be - or + and fluctuates over time. Furthermore, we should somehow explain why there are two different ones for long and short.
> With @bonomat we thought about calling it interest rate instead, but that will obviously not explain the details ;)

A combined position would give you a natural place to display the funding rate!

klochowicz (approved):
it looks actually quite natural there! 👍

klochowicz:
> The "funding rate" display and behavior will definitely require some more documentation in the UI. It is not clear how it behaves from just seeing the values. It is also not clear that is actually can be - or + and fluctuates over time. Furthermore, we should somehow explain why there are two different ones for long and short. With @bonomat we thought about calling it interest rate instead, but that will obviously not explain the details ;)

I like the name "interest rate", as this would be consistent with bitmex vocabulary. This is something we could write in our glossary or FAQ for the users (or at least via the tooltip!)

thomaseizinger:
> > The "funding rate" display and behavior will definitely require some more documentation in the UI. It is not clear how it behaves from just seeing the values. It is also not clear that is actually can be - or + and fluctuates over time. Furthermore, we should somehow explain why there are two different ones for long and short. With @bonomat we thought about calling it interest rate instead, but that will obviously not explain the details ;)
> 
> I like the name "interest rate", as this would be consistent with bitmex vocabulary. This is something we could write in our glossary or FAQ for the users (or at least via the tooltip!)

Semantically as far as I understand. CFDs deal with **two** interest rates: one for the base currency and one for the quote currency. The combination of both yields the "funding rate".
Thus it would not be correct to call it "interest rate" because it would not be clear, which one it is.

taker-frontend/src/App.tsx
```diff
@@ -69,40 +67,26 @@ export const App = () => {
...
-                openingFee: order.opening_fee || 0,
-                fundingRateAnnualized: order.funding_rate_annualized_percent || "0",
-                fundingRateHourly: order
-                    ? Number.parseFloat(order.funding_rate_hourly_percent).toFixed(5)
-                    : "0",
-                liquidationPrice: parseOptionalNumber(order.liquidation_price),
+                id: offer.id,
+                initialFundingFeePerParcel: offer.initial_funding_fee_per_parcel,
+                marginPerParcel: offer.margin_per_parcel,
+                price: Number.parseFloat(offer.price),
```
bonomat: not particular related to your PR: 

why are `offer.price`, `offer.min_/max_quantity` strings in the first place? Couldn't we just say these are numbers and avoid `Number.parseFloat`?

da-kami: I think we could do that - we are probably a bit too "correct" there. In the daemon they are represented as `Decimal`, and we send them out as string. We *could* use a `Decimal` library in the UI code as well, but don't at the moment, so for now we could just just make them numbers I support :)

da-kami: I gave this a try, please have another look

klochowicz: I think it was a historical decision to control some of the precision-handling in the projection (which exists solely for the purpose of the UI)...

klochowicz (approved):
LGTM!

taker-frontend/src/App.tsx
```diff
@@ -69,40 +67,26 @@ export const App = () => {
...
-                fundingRateHourly: order
-                    ? Number.parseFloat(order.funding_rate_hourly_percent).toFixed(5)
-                    : "0",
-                liquidationPrice: parseOptionalNumber(order.liquidation_price),
+                id: offer.id,
+                initialFundingFeePerParcel: offer.initial_funding_fee_per_parcel,
+                marginPerParcel: offer.margin_per_parcel,
+                price: offer.price,
+                liquidationPrice: offer.liquidation_price,
+                fundingRateAnnualized: offer?.funding_rate_annualized_percent,
```
bonomat: Nit: `?` looks outdated. 

do we still need this function then? :)

bonomat (approved):
Looks good

PR #1585, opened 2022-03-09
Remove `cfd_actors` module

thomaseizinger:
See commit messages.


daemon/src/auto_rollover.rs
```diff
@@ -108,7 +107,9 @@ where
             .expect("actor to be able to give address to itself");
         for id in cfd_ids {
             try_continue!(async {
-                let cfd = load_cfd(id, &mut conn).await?;
+                #[allow(deprecated)] // This checking for rollover eligibility does not fit into the command abstraction (yet). Figure out a way for fixing it.
```
klochowicz, line 110: a missing `TODO`?

daemon/src/rollover_maker.rs
```diff
@@ -137,7 +139,8 @@ impl Actor {
         let order_id = self.order_id;
         let AcceptRollover {
             tx_fee_rate,
-            funding_rate,
+            long_funding_rate,
+            short_funding_rate,
```
klochowicz, line 143: note that we are paying a price by using the "command" abstraction here - we are passing an unused funding rate, which can be confusing to the users.
Perhaps we could document this in `AcceptRollover`? (that only one of them gets picked, depending on position)

On the plus side, we are removing one additional cfd load, which is good. :-)

daemon/src/command.rs
```diff
@@ -98,3 +145,11 @@ impl<TOne, TTwo, TThree> ExtractEventFromTuple for (CfdEvent, TOne, TTwo, TThree
         (Some(self.0), (self.1, self.2, self.3))
     }
 }
+
+impl<TOne, TTwo, TThree, TFour> ExtractEventFromTuple for (CfdEvent, TOne, TTwo, TThree, TFour) {
```
klochowicz, line 149: I wonder how far this will go ;D

klochowicz (approved):
nice cleanup! we're one step closer to just having "commands" 👍

daemon/src/auto_rollover.rs
thomaseizinger: The fact that we are using `allow(deprecated)` should be enough of a hint that this needs fixing, no? :D

daemon/src/rollover_maker.rs
thomaseizinger: > Perhaps we could document this in `AcceptRollover`? (that only one of them gets picked, depending on position)

Yes can document that!

daemon/src/auto_rollover.rs
klochowicz: I like making such comments clear with `TODO` or `FIXME` (the latter being disfavoured here). Easier to grep or display within the IDE (not sure whether CLion supports that, but magit in emacs shows all the TODOs - so they're harder to ignore). 

I can see how `allow(deprecated)` can be looked the same way - perhaps it's because I'm still quite fresh in this ecosystem....Perhaps I should just add a filter for `allow(deprecated)` in my magit-todos 🤔

No need to address it if you wish to keep it as-is :)

daemon/src/command.rs
```diff
@@ -98,3 +98,11 @@ impl<TOne, TTwo, TThree> ExtractEventFromTuple for (CfdEvent, TOne, TTwo, TThree
...
```
luckysori, line 149: Hehe I added the same thing in my PR.

luckysori (approved):
Very happy with this. Nice use of `deprecated` flag too!

PR #1590, opened 2022-03-09
Calculate liquidation price for short position

### LINKED ISSUE #1589: Liquidation price for short is wrong
Looks like there is an error when calculating the short liquidation price : 

image

says: 

```
Liquidation | $27,635
```

but it should be something very high.

bonomat:
Resolves #1589

Note: the order in the UI has the wrong liquidation price since it still relies on global params which only takes long position into account. This will be fixed in #1582

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -268,11 +268,8 @@ impl Order {
         let leverage_choices_for_taker = Leverage::TWO;
 
         let liquidation_price = match position_maker {
-            Position::Short => calculate_long_liquidation_price(leverage_choices_for_taker, price),
-            Position::Long => {
-                // TODO: Should be different
-                calculate_long_liquidation_price(leverage_choices_for_taker, price)
-            }
+            Position::Short => calculate_short_liquidation_price(leverage_choices_for_taker, price),
+            Position::Long => calculate_long_liquidation_price(leverage_choices_for_taker, price),
         };
 
         Order {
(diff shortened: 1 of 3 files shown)


bonomat:
tests are correct: there is an issue :) converted to draft and will work on it

bonomat:
Ready for review @da-kami / @klochowicz

model/src/cfd.rs
```diff
@@ -1435,6 +1432,17 @@ pub fn calculate_long_liquidation_price(leverage: Leverage, price: Price) -> Pri
     price * leverage / (leverage + 1)
 }
 
+/// calculates short liquidation price
+///
+/// Note: if leverage == 1, then the liquidation price will go towards infinity.
+/// This is represented as Price::INFINITE
+pub fn calculate_short_liquidation_price(leverage: Leverage, price: Price) -> Price {
+    if (leverage - 1) == 0 {
```
da-kami: Any reason why you are not using the `Leverage::ONE` constant? 
To make the division by `0` more obvious? I think it would still read better using the constant...

da-kami (approved):

model/src/cfd.rs
bonomat: `Leverage - Leverage` does not make much sense because (imho) `Leverage-Leverage` should return `Leverage` but `Leverage::new` does not allow `0`, which means, a naive implementation of `impl Sub<Leverage> for Leverage` would not work.

da-kami: I am confused by your answer.
Why not this condition:

```suggestion
    if leverage == Leverage::ONE {
```

bonomat: ah, yea, that would work 😅

PR #1591, opened 2022-03-10
Calculate payouts correctly

### LINKED ISSUE #1580: Calculate the payout correcly when the taker is short

luckysori:
Fixes #1580.

### DIFF
--- a/daemon/src/rollover_maker.rs
+++ b/daemon/src/rollover_maker.rs
@@ -154,7 +154,7 @@ impl Actor {
 
         tracing::debug!(%order_id, "Maker accepts a rollover proposal");
 
-        let (rollover_params, dlc, interval, funding_rate) = self
+        let (rollover_params, dlc, position, interval, funding_rate) = self
             .executor
             .execute(self.order_id, |cfd| {
                 let funding_rate = match cfd.position() {
@@ -162,10 +162,10 @@ impl Actor {
                     Position::Short => short_funding_rate,
                 };
 
-                let (event, params, dlc, interval) =
+                let (event, params, dlc, position, interval) =
                     cfd.accept_rollover_proposal(tx_fee_rate, funding_rate)?;
 
-                Ok((event, params, dlc, interval, funding_rate))
+                Ok((event, params, dlc, position, interval, funding_rate))
             })
             .await?;
 
@@ -208,6 +208,7 @@ impl Actor {
             (self.oracle_pk, announcement),
             rollover_params,
             Role::Maker,
+            position,
             dlc,
             self.n_payouts,
         );
--- a/daemon/src/rollover_taker.rs
+++ b/daemon/src/rollover_taker.rs
@@ -97,7 +97,7 @@ impl Actor {
         }: RolloverAccepted = msg;
         let order_id = self.id;
 
-        let (rollover_params, dlc) = self
+        let (rollover_params, dlc, position) = self
             .executor
             .execute(self.id, |cfd| {
                 cfd.handle_rollover_accepted_taker(tx_fee_rate, funding_rate)
@@ -128,6 +128,7 @@ impl Actor {
             (self.oracle_pk, announcement),
             rollover_params,
             Role::Taker,
+            position,
             dlc,
             self.n_payouts,
         );
--- a/daemon/src/setup_maker.rs
+++ b/daemon/src/setup_maker.rs
@@ -81,7 +81,7 @@ impl Actor {
         // the spawned contract setup task
         self.setup_msg_sender = Some(sender);
 
-        let setup_params = self
+        let (setup_params, position) = self
             .executor
             .execute(order_id, |cfd| cfd.start_contract_setup())
             .await?;
@@ -101,6 +101,7 @@ impl Actor {
             self.build_party_params.clone_channel(),
             self.sign.clone_channel(),
             Role::Maker,
+            position,
             self.n_payouts,
         );
 
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -64,8 +64,6 @@ use uuid::Uuid;
 
 pub const CET_TIMELOCK: u32 = 12;
 
-pub const SHORT_LEVERAGE: u8 = 1;
-
 #[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, sqlx::Type)]
 #[sqlx(transparent)]
 pub struct OrderId(Hyphenated);
@@ -814,6 +812,7 @@ impl Cfd {
                 self.initial_price,
                 self.quantity,
                 self.long_leverage,
+                self.short_leverage,
                 self.refund_timelock_in_blocks(),
                 self.initial_tx_fee_rate(),
                 self.fee_account,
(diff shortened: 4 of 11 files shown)


model/src/cfd.rs
```diff
@@ -918,11 +918,15 @@ impl Cfd {
         proposal: SettlementProposal,
         sig_taker: Signature,
     ) -> Result<CollaborativeSettlement> {
+        anyhow::ensure!(
+            self.role == Role::Maker,
+            "Only the maker can create complete collaborative settlement signing"
+        );
```
klochowicz: I'd go even further by using `assert_debug` or `assert` perhaps - it's not a runtime error, it's a bug in our code if this ever happens, therefore I'd avoid gracefully handling this.

I'm happy also to settle on this one for the time being as it's a giant improvement guarding us from actually doing it when we shouldn't

daemon/src/maker_cfd.rs
```diff
@@ -271,8 +271,13 @@ where
     ) -> Result<()> {
         tracing::info!(%order_id, %taker_id,  "Received rollover proposal from taker");
 
+        let mut conn = self.db.acquire().await?;
+        let cfd = load_cfd(order_id, &mut conn).await?;
+        let position = cfd.position();
```
klochowicz: this would conflict with: #1585 - suggest syncing with @thomaseizinger

daemon-tests/src/lib.rs
```diff
@@ -361,10 +361,10 @@ impl Taker {
         }
     }
 
-    pub async fn trigger_rollover(&self, id: OrderId) {
+    pub async fn trigger_rollover(&self, order_id: OrderId, position: Position) {
```
klochowicz: I thought for a while that this is pretty weird API to need to pass the position, but then realised it's just for the tests..... 😌

klochowicz (approved):
nice job! both with cleanups & features 👍

daemon-tests/src/lib.rs
luckysori: To be fair it's still pretty weird to me, but I couldn't come up with anything better at this stage.

daemon/src/rollover_taker.rs
```diff
@@ -128,6 +132,7 @@ impl Actor {
             (self.oracle_pk, announcement),
             rollover_params,
             Role::Taker,
+            self.position,
```
thomaseizinger: I'd suggest returning the position from `handle_rollover_accepted_taker` if you need it here.

daemon/src/maker_cfd.rs
thomaseizinger: Yes, lets not do this please.

model/src/cfd.rs
```diff
@@ -1890,11 +1897,18 @@ pub fn calculate_payouts_long_taker(
...
+            .collect(),
+        (Position::Short, Role::Taker) | (Position::Long, Role::Maker) => payouts
+            .into_iter()
+            .map(|payout| generate_payouts(payout.range, payout.long, payout.short))
+            .flatten_ok()
+            .collect(),
+    }
 }
 
 pub fn calculate_payouts_short_taker(
```
thomaseizinger: Do we still need this function now?

daemon-tests/src/lib.rs
thomaseizinger: I think this might go away if we return the position together with the event from the command handler?

thomaseizinger (changes requested):
Thanks for bringing over the refactoring commits. One minor change I'd like to see otherwise LGTM!

daemon/src/rollover_taker.rs
luckysori: Thanks for the suggestion, that makes things a lot easier.

model/src/cfd.rs
luckysori: I think you're right. I went for `debug_assert`.

daemon/src/rollover_maker.rs
```diff
@@ -52,6 +53,7 @@ struct RolloverFailed {
 
 pub struct Actor {
     order_id: OrderId,
+    position: Position,
```
thomaseizinger: This actor should also retrieve the position by returning it from `accept_rollover_proposal`.

daemon/src/setup_maker.rs
```diff
@@ -101,6 +101,7 @@ impl Actor {
             self.build_party_params.clone_channel(),
             self.sign.clone_channel(),
             Role::Maker,
+            self.order.position_maker,
```
thomaseizinger: To be consistent, you could also get the position from `start_contract_setup` here instead of passing it in through the `order.` Not sure what is cleaner.

daemon/src/rollover_maker.rs
luckysori: You're right!!!!

daemon/src/setup_maker.rs
luckysori: It's simpler and more consistent that way because we don't have to pass any extra data to the `setup_taker.rs`.

model/src/cfd.rs
```diff
@@ -940,7 +952,9 @@ impl Cfd {
             "Failed to propose collaborative settlement"
         );
 
-        let payout_curve = calculate_payouts_long_taker(
+        let payout_curve = calculate_payouts(
+            self.position,
+            self.role,
             self.initial_price,
             self.quantity,
             self.long_leverage,
```
da-kami, line 961: was unable to comment on the line below...

We still always hard-code `Leverage::new(SHORT_LEVERAGE)` - this has to be change to `self.short_leverage` if I understand this correctly. 

Note that this is in multiple places!
I'd recommend removing the `SHORT_LEVERAGE` constant. The maker always being locked in to `Leverage::ONE` should already be correctly handled when constructing the `Cfd`.

da-kami (changes requested):

thomaseizinger (approved):
Cool, happy with that! Thank you :)

model/src/cfd.rs
thomaseizinger: Nice catch!

luckysori: Makes sense! I'll add a commit on top because this is not strictly related to assigning the payouts to the correct parties.

da-kami: Would be great if you could remove the constant `SHORT_LEVERAGE` so we ensure that we don't use it anymore.

luckysori: > Would be great if you could remove the constant `SHORT_LEVERAGE` so we ensure that we don't use it anymore.

Pretty sure I did!

da-kami: Ah, yeah, sorry. Somhow did not see it in the commit 😅

da-kami (approved):

PR #1594, opened 2022-03-10
Update maker UI to allow maker going long

### LINKED ISSUE #1516: Maker UI changes for allowing the taker to go short
The API already supports going long, but it's not enabled in the maker UI.

We need to:
- allow setting up both long and short offer prices instead of a single one,
- change the button name from 'create/update sell order' to 'create/update offer'

klochowicz:


rishflab (approved):

maker-frontend/src/MakerApp.tsx
```diff
@@ -137,12 +154,13 @@ export default function App() {
...
                         <GridItem colSpan={2} textAlign="center">
                             <Button
-                                disabled={isCreatingNewCfdOrder || orderPrice === "0"}
+                                disabled={isCreatingNewCfdOrder || shortPrice === "0"}
                                 variant={"solid"}
                                 colorScheme={"blue"}
                                 onClick={() => {
                                     let payload: CfdSellOrderPayload = {
-                                        price: Number.parseFloat(orderPrice),
+                                        price: Number.parseFloat(shortPrice),
```
bonomat, line 174: I think the field is called: `price_short`

maker-frontend/src/MakerApp.tsx
```diff
@@ -49,12 +49,13 @@ export default function App() {
```
bonomat, line 51: I'm confused about this commit. Is this logic maybe a bit outdated?

maker-frontend/src/MakerApp.tsx
klochowicz: not in the old endpoint! I'll be doing a follow-up on top of this that uses the new endpoint (same with the `funding_rate`...)

bonomat: ah, got it. 👍

maker-frontend/src/MakerApp.tsx
```diff
@@ -173,7 +173,7 @@ export default function App() {
                                     makeNewCfdSellOrder(payload);
                                 }}
                             >
-                                {order ? "Update Sell Order" : "Create Sell Order"}
+                                {order ? "Update Maker Offers" : "Create Maker Offers"}
```
bonomat: Nit: we are in the maker frontend, so no need to mention this :) 

```suggestion
                                {order ? "Update Offers" : "Create Offers"}
```

klochowicz: github has a bug and won't let me "commit suggestion" :( - I'll push a fixup!

maker-frontend/src/MakerApp.tsx
klochowicz: yeah, this was the starting point from @rishflab's PR. I have created 2 separate spread values in one of the further commits - it doesn't really matter much in the maker UI apart from actually seeing 2 different prices IMHO, as it's more of an example and dev testing env

maker-frontend/src/MakerApp.tsx
```diff
@@ -165,6 +165,8 @@ export default function App() {
                                         max_quantity: Number.parseFloat(maxQuantity),
                                         // TODO: Populate funding rate from the UI
                                         funding_rate: (0.00002283 * 24), // annualized 20% by default to have some values
+                                        // TODO: Populate funding rate from the UI
+                                        daily_funding_rate_long: (-0.00002283 * 24), // annualized 20% by default to have some values
```
bonomat, line 181: On purpose negative?

klochowicz: yes, wanted to visualise the maker profiting both ways - happy to remove the negative sign if it's more misleading than helping!

maker-frontend/src/MakerApp.tsx
```diff
@@ -105,20 +115,39 @@ export default function App() {
                             value={format(maxQuantity)}
                         />
 
-                        <Text>Offer Price:</Text>
+                        <Text>Maker Short Price:</Text>
```
bonomat, line 118: I'm still super confused about the naming... 
Maker short price means, that's the price the maker offers to a taker who goes long?

bonomat (approved):
👍
I think we need to do something about the namings... It's not very intuitive...

maker-frontend/src/MakerApp.tsx
bonomat: no worries, this is just for testing anyways :)

maker-frontend/src/MakerApp.tsx
klochowicz: Yes. That's why I tried do be more elaborate - I could go even more verbose "Taker going long price" instead if that helps.

PR #1607, opened 2022-03-10
Update xtra

thomaseizinger:
The diff can be seen here: https://github.com/Restioson/xtra/compare/95508e6a...41a5dc21

This introduces a metric for measuring the execution time message handlers in actors.


bonomat (approved):
Let's give this a try. 👍

thomaseizinger:
Idiot me did not activate the feature flag ...

PR #1610, opened 2022-03-11
Introduce facilities for deprecating endpoints

thomaseizinger:
Breaking APIs causes friction for clients. Where possible we try to avoid breaking APIs and deprecate them instead whilst adding alternative ones or new versions.

Once deprecated, we need to migrate clients off the deprecated endpoint. To make this easier, we implement some basic rocket facilities for adding a `Deprecation` header. Such header can be parsed by clients and be used to alert monitoring systems that there is code in place which hits deprecated resources. Once alerted, developers can go in and transition to a new endpoint.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1565,6 +1565,7 @@ dependencies = [
  "prometheus",
  "rocket",
  "rocket-basicauth",
+ "rocket-deprecation",
  "rust-embed",
  "rust-embed-rocket",
  "serde",
@@ -2654,6 +2655,14 @@ dependencies = [
  "hex",
  "rocket",
  "void",
+]
+
+[[package]]
+name = "rocket-deprecation"
+version = "0.1.0"
+dependencies = [
+ "rocket",
+ "time 0.3.7",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "bdk-ext",
   "model",
   "btsieve",
+  "rocket-deprecation",
 ]
 resolver = "2"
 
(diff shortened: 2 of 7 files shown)


luckysori (approved):
Cool! We should merge this at some point!

thomaseizinger:
I think I'll actually move the last commit to a separate repository.

klochowicz (approved):
looks really useful, let's get it in!

bonomat (commented):
146b401b02f2e811dd773d42a4f3bbd2efeee847 is only useful if me make use if it. Headers are not something I usually look at so do we plan to make use of this in our clients (automation/metrics/frontend)?

thomaseizinger:
> [146b401](https://github.com/itchysats/itchysats/commit/146b401b02f2e811dd773d42a4f3bbd2efeee847) is only useful if me make use if it. Headers are not something I usually look at so do we plan to make use of this in our clients (automation/metrics/frontend)?

Correct. I built https://github.com/itchysats/reqwest-deprecation as well but we would obviously need to make use of it. I depends a bit on the outcome of https://github.com/itchysats/architecture/pull/25. This kind of infrastructure is useful if you want to evolve a distributed system. I think it would be preferable to not build a distributed system unless really necessary. This work would not be required then.

thomaseizinger:
I've extracted the `rocket-deprecation` crate here: https://github.com/itchysats/rocket-deprecation so this can be safely closed.

As @bonomat pointed out, it doesn't add much value unless we actually use this information and with https://github.com/itchysats/architecture/pull/25 being on the brink, it is not totally clear that this will be the case.

PR #1611, opened 2022-03-11
Calculate `hours_to_charge` for rollover correctly

### LINKED ISSUE #1125: Charge funding fee for "missed" rollovers

luckysori:
Fixes #1125.

Previously we always assumed that rollover would consistently happen every hour. Obviously parties can go offline, so hard-coding a value of 1 hour was naive.

By looking at the timestamp of the the `settlement_event_id` (the time at which the oracle will non-collaboratively settle the CFD) we can figure out how long is left of the time-to-live of the CFD and therefore by how many hours we need to extend it to restore a time-to-live equal to the `SETTLEMENT_INTERVAL`. We can use the latter number to determine how many hours we should charge for the rollover.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -841,15 +841,14 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
-        let hours_to_charge = 1;
-
+        let hours_to_charge = self.hours_to_extend_in_rollover()?;
         let funding_fee = calculate_funding_fee(
             self.initial_price,
             self.quantity,
             self.long_leverage,
             self.short_leverage,
             funding_rate,
-            hours_to_charge,
+            hours_to_charge as i64,
         )?;
 
         Ok((
@@ -885,16 +884,14 @@ impl Cfd {
 
         self.can_rollover()?;
 
-        // TODO: Taker should take this from the maker, optionally calculate and verify
-        // whether they match
-        let hours_to_charge = 1;
+        let hours_to_charge = self.hours_to_extend_in_rollover()?;
         let funding_fee = calculate_funding_fee(
             self.initial_price,
             self.quantity,
             self.long_leverage,
             self.short_leverage,
             funding_rate,
-            hours_to_charge,
+            hours_to_charge as i64,
         )?;
 
         Ok((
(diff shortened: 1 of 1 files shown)


model/src/cfd.rs
```diff
@@ -1336,6 +1333,36 @@ impl Cfd {
...
+
+        if !hours_left.is_positive() {
+            tracing::warn!("Rolling over a contract that can be settled non-collaboratively");
+
+            return Ok(SETTLEMENT_INTERVAL.whole_hours() as u64);
+        }
+
+        let time_to_extend = SETTLEMENT_INTERVAL
+            .checked_sub(hours_left)
+            .context("Cannot rollover if time-to-live of contract is < {SETTLEMENT_INTERVAL}")?;
```
thomaseizinger, line 1356: This will not take the variable's value but just render `{SETTLEMENT_INTERVAL}`.

model/src/cfd.rs
```diff
@@ -1336,6 +1333,36 @@ impl Cfd {
...
+    fn hours_to_extend_in_rollover(&self) -> Result<u64> {
+        let dlc = self.dlc.as_ref().context("Cannot roll over without DLC")?;
+        let settlement_time = dlc.settlement_event_id.timestamp();
+
+        let hours_left = settlement_time - OffsetDateTime::now_utc();
+
+        if !hours_left.is_positive() {
+            tracing::warn!("Rolling over a contract that can be settled non-collaboratively");
+
+            return Ok(SETTLEMENT_INTERVAL.whole_hours() as u64);
```
thomaseizinger, line 1351: This might be debatable but my vote goes for this behaviour. As long as we are not committed on-chain, I think we should try and rollover.

model/src/cfd.rs
```diff
@@ -2921,6 +2948,64 @@ mod tests {
         assert_eq!(is_liquidation_price, should_liquidation_price);
     }
 
+    #[test]
+    fn correctly_calculate_hours_to_extend_in_rollover() {
+        let settlement_interval = SETTLEMENT_INTERVAL.whole_hours();
+        for hour in 0..settlement_interval {
+            let event_id_in_x_hours = BitMexPriceEventId::with_20_digits(
+                OffsetDateTime::now_utc() + Duration::new(hour * 60 * 60, 0),
```
thomaseizinger, line 2956:
```suggestion
                OffsetDateTime::now_utc() + hour.hours(),
```

Might be more readable.

model/src/cfd.rs
```diff
@@ -2921,6 +2948,64 @@ mod tests {
...
+                maker.hours_to_extend_in_rollover().unwrap(),
+                (settlement_interval - hour) as u64
+            );
+        }
+    }
+
+    #[test]
+    fn rollover_extends_time_to_live_by_settlement_interval_if_cfd_can_be_settled() {
+        let event_id_1_hour_ago = BitMexPriceEventId::with_20_digits(
+            OffsetDateTime::now_utc() - Duration::new(60 * 60, 0),
```
thomaseizinger, line 2977:
```suggestion
            OffsetDateTime::now_utc() - 1.hours(),
```

model/src/cfd.rs
luckysori: Hehe thought so.

luckysori:
Ooops, renaming a branch closes the associated PR. Reopening this with @thomaseizinger's review feedback included.

PR #1618, opened 2022-03-14
Calculate `hours_to_charge` for rollover correctly

### LINKED ISSUE #1125: Charge funding fee for "missed" rollovers

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/1125.

Previously we always assumed that rollover would consistently happen every hour. Obviously parties can go offline, so hard-coding a value of 1 hour was naive.

By looking at the timestamp of the the `settlement_event_id` (the time at which the oracle will non-collaboratively settle the CFD) we can figure out how long is left of the time-to-live of the CFD and therefore by how many hours we need to extend it to restore a time-to-live equal to the `SETTLEMENT_INTERVAL`. We can use the latter number to determine how many hours we should charge for the rollover.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -48,6 +48,7 @@ use maia::secp256k1_zkp::SECP256K1;
 use maia::spending_tx_sighash;
 use maia::Payout;
 use maia::TransactionExt;
+use num::Zero;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
 use serde::de::Error as _;
@@ -841,15 +842,14 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
-        let hours_to_charge = 1;
-
+        let hours_to_charge = self.hours_to_extend_in_rollover()?;
         let funding_fee = calculate_funding_fee(
             self.initial_price,
             self.quantity,
             self.long_leverage,
             self.short_leverage,
             funding_rate,
-            hours_to_charge,
+            hours_to_charge as i64,
         )?;
 
         Ok((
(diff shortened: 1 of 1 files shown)


model/src/cfd.rs
```diff
@@ -1336,6 +1333,41 @@ impl Cfd {
...
+                format!(
+                    "Cannot rollover if time-to-live of contract is < {} hours",
+                    SETTLEMENT_INTERVAL.whole_hours()
+                )
+            })?;
+        let hours_to_extend = time_to_extend.whole_hours();
+
+        if !hours_to_extend.is_positive() {
+            bail!("Rollover must extend time-to-live by at least 1 hour");
+        }
```
thomaseizinger: Shouldn't we just default to `1` here?

Do we have a test that hits this path?

thomaseizinger (approved):

model/src/cfd.rs
luckysori: I thought it would be undesirable to perform a rollover if we already have a big enough time-to-live.

To make it less prone to failures I've made it so that a CFD with a time-to-live of 24-25 hours actually rolls over for an extra hour. Beyond that I think it's okay to prevent the rollover from happening. Let me know what you think about https://github.com/itchysats/itchysats/pull/1618/commits/3a672f26d090e45eb39a68dbc16b19371e4af075. I've also added a test.

thomaseizinger: Yes sounds good to me!

PR #1624, opened 2022-03-14
Remove reactr-scripts

bonomat:
Resolves vulnerability: https://github.com/itchysats/itchysats/security/dependabot/19


thomaseizinger (approved):
Lol, we are not even using react-scripts. Why is that even in the list of dependencies!

PR #1638, opened 2022-03-15
Print entire cause chain in case of actor restart

thomaseizinger:
Previously, the stop reason of a supervised actor only needed to
implement `fmt::Display`. In practise however, we are always using
an enum here that implements `std::error::Error`. The default impl of
`Display` for an `Error` only prints the top-level cause. If the error
contains a `cause`, it is the responsibility of the caller to walk the
chain of causes and concatenate all the errors.

Anyhow implements this logic for its own error type and it can be used
by specifying the `alternate` modifier of `Display`: `#`.

To construct an anyhow::Error, we change the trait requirements for the
stop reason to implement `std::error::Error`.

### DIFF
--- a/xtras/src/supervisor.rs
+++ b/xtras/src/supervisor.rs
@@ -2,6 +2,7 @@ use crate::ActorName;
 use async_trait::async_trait;
 use futures::FutureExt;
 use std::any::Any;
+use std::error::Error;
 use std::fmt;
 use std::panic::AssertUnwindSafe;
 use tokio_tasks::Tasks;
@@ -16,7 +17,7 @@ pub struct Actor<T, R> {
     context: Context<T>,
     ctor: Box<dyn Fn() -> T + Send + 'static>,
     tasks: Tasks,
-    restart_policy: Box<dyn FnMut(R) -> bool + Send + 'static>,
+    restart_policy: Box<dyn FnMut(&R) -> bool + Send + 'static>,
     metrics: Metrics,
 }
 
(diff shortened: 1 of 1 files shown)


klochowicz (commented):
out of curiosity - how does this look like? have you managed to capture some examples is the logs? has it proven more useful?

thomaseizinger:
> out of curiosity - how does this look like? have you managed to capture some examples is the logs? has it proven more useful?

It looks like other cause chains that we print out in the logs. They are concatenated with `:`. Something like:

```
Failed to deserialize long_offer event: invalid type: map, expected unit at line 1 column 0
```

xtras/src/supervisor.rs
```diff
@@ -144,13 +148,13 @@ where
...
-    R: fmt::Display + Send + 'static,
+    R: Error + Send + Sync + 'static,
     S: Into<R> + Send + 'static,
 {
     pub fn handle(&mut self, msg: Stopped<R>, ctx: &mut Context<Self>) {
         let actor = T::name();
-        let reason_str = msg.reason.to_string();
-        let should_restart = (self.restart_policy)(msg.reason);
+        let should_restart = (self.restart_policy)(&msg.reason);
+        let reason_str = format!("{:#}", anyhow::Error::new(msg.reason)); // Anyhow will format the entire chain of errors when using `alternate` Display (`#`)
```
luckysori, line 157: 👍 Nice trick. It's interesting that it's not easy to print the error chain without converting to `anyhow::Error`.

luckysori (approved):

xtras/src/supervisor.rs
thomaseizinger: It is in the works: https://github.com/rust-lang/project-error-handling/issues/10#issuecomment-720652610

PR #1639, opened 2022-03-15
Send itchysats version in Hello message from taker and maker

klochowicz:


daemon/src/wire.rs
```diff
@@ -80,7 +80,9 @@ pub mod taker_to_maker {
 #[serde(tag = "type", content = "payload")]
 #[allow(clippy::large_enum_variant)]
 pub enum TakerToMaker {
-    Hello(Version),
+    Hello {
+        wire_version: Version,
```
thomaseizinger, line 84: I am not sure we should break _this_ message as well. With a blue/green deployment, we should be fine doing this but for older clients that try to connect to this version, the version negotiation _itself_ will break.

Normally that is a no-no in terms of compatibility. You shouldn't break the message that establishes whether or not you are compatible.

daemon/Cargo.toml
```diff
@@ -48,4 +48,9 @@ xtras = { path = "../xtras" }
 pretty_assertions = "1"
 serde_test = "1"
 time = { version = "0.3", features = ["std"] }
+
```
thomaseizinger: What is this newline doing here?

daemon/src/version.rs
```diff
@@ -0,0 +1,3 @@
+pub fn version() -> &'static str {
+    env!("VERGEN_GIT_SEMVER_LIGHTWEIGHT")
```
thomaseizinger, line 2: Why is this within the `daemon` crate? If we want to centralise it, I think it should go into `shared-bin`.

thomaseizinger (commented):
Two concerns, otherwise LGTM.

daemon/src/version.rs
klochowicz: we need to have the access to the `version()` inside the daemon in order to send it across (the additional usage in is the 3rd commit, if we didn't plan to send it, then of course `shared-bin` would be the natural target).

thomaseizinger: Right. Would it make sense to pass it in as a parameter instead? I guess that would add a lot of friction.

daemon/src/wire.rs
klochowicz: Fair enough, I have relied implicitly on the blue/green deployment strategy we're about to use to get over the big protocol break, and that could have cause short-term breakage if it wasn't used :/

I could add a whole additional message, but this will complicate the protocol and bake in a temporary awkwardness forever.

What would you consider to be the best path forward?

thomaseizinger: If we wanted to be backwards-compatible, we can just make the `daemon_version` field optional. I would do that like this:

1. Write a test that cements the serialization of the current `Hello` message
2. Introduce a newtype for the payload of the `Hello` message
3. Add an optional, 2nd field to the payload struct and write more tests against it to ensure we can still deserialize it correctly

Probably need to make use of `serde(default)` or something like that.

thomaseizinger: If you wanted to introduce the named fields as well, I you could always write a custom deserializer for the payload struct as well.

thomaseizinger: Probably can actually skip the newtype if you don't want a custom serializer.

klochowicz: Thank you for these comments, they are extremely valuable (copied to my API design notes).

Before I go off on this path, I'd like to confirm - do we actually want/need to be backwards compatible in this case, given we're doing the blue-green deployment this time?

thomaseizinger: I'd say give how simple it seems to be, I'd try and implement this in a time-box. If you can't achieve it, then just break the API :)

daemon/src/version.rs
da-kami: Wouldn't it be better to use the version from `taker` and `maker` crate (can pass it into the wire actors if necessary)?
I don't think it's important that the versions match, we just want to know what version a taker is running on the maker side.

daemon/src/wire.rs
da-kami: > Before I go off on this path, I'd like to confirm - do we actually want/need to be backwards compatible in this case, given we're doing the blue-green deployment this time?

I don't think it matters at this point, given the changes when receiving `Offers`. However, as Thomas pointed out, the user will not get an error messages that can be interpreted if we break `Hello`. However, I don't think many users look at the logs. I'm OK with breaking changes to `Hello` at this point, but agree with Thomas that we can try to do this different if easily possible.

daemon/src/version.rs
klochowicz: This is the version that is printed on the other side when the logs say "running version XXX" - IMHO it would cause a lot of confusion if we suddenly decided to print a different kind of version. We can later adjust that if there are strong opinions - what matters in this PR most is that we have a string that gets printed :)

Note that the versions from `maker` and `taker` crate actually match this one - with the small addition that if we're running from master, we append SHA hash to the version too.

klochowicz: > Right. Would it make sense to pass it in as a parameter instead? I guess that would add a lot of friction.

it would cause more friction and I don't think it's worth it :( can do it later if we decide to.

daemon/src/wire.rs
klochowicz: Timeboxing myself on doing this now, will try to address it before midday :-)

daemon/src/version.rs
thomaseizinger: > Note that the versions from `maker` and `taker` crate actually match this one - with the small addition that if we're running from master, we append SHA hash to the version too.

They don't match. The maker and taker crate versions are currently set to `0.1.0`.
As long as we are cutting _repository-wide_ releases (as in we are releasing all binaries based on a single tag), it does not make any sense to version these binaries differently.

There is a [proposed ADR](https://github.com/itchysats/architecture/pull/16) for changing the versioning scheme to move away from semantic versioning because exactly what we are doing at the moment is super confusing because it may mean that we are releasing a version of the `taker` binary that has 0 functional changes but because we are cutting repository wide releases, we have to put a release out for `taker` as well even if only `maker` changes.

klochowicz: @thomaseizinger , let's not dive into versioning debate in this issue here. To me the comment above just supports that I'm printing the correct one (the one that matters here). I am yet to see any usefulness from having multiple versions - but the problems highlighted by me in the ADR are clearly visible already :(

please note that this is the `daemon_version`, not the binary version, and they actually *do* match.

klochowicz: @da-kami would you prefer me to print out `0.1.0` here?

daemon/Cargo.toml
klochowicz: fixed, thanks for spotting!

daemon/src/wire.rs
klochowicz: Didn't manage to finish it within the timebox - as spoken with @thomaseizinger, the message will get broken soon anyway with the introduction of libp2p, so I don't think it's worth investing more time into it nor delaying the release.

daemon/src/version.rs
thomaseizinger: > @thomaseizinger , let's not dive into versioning debate in this issue here. To me the comment above just supports that I'm printing the correct one (the one that matters here). I am yet to see any usefulness from having multiple versions - but the problems highlighted by me in the ADR are clearly visible already :(

I totally agree! I don't see us moving away from a single repository any time soon and until then, repository-wide releases make total sense which also means, everything should have one version number.

What I was pointing out is that printing the _crate_ version of the maker and taker binary is pointless because it is set to `0.1.0` and that is _different_ from the version we tag.

thomaseizinger (approved):

thomaseizinger:
Newline is still there but not criticial!

da-kami (approved):

klochowicz:
Closed in favour of #1648 - we changed the approach (we want to avoid blue/green deployment), and now the compatibility is critical.

daemon/src/wire.rs
klochowicz: update: I tried using advanced features of serde, but making it work in the way outlined above turned out to be impossible as serde would would serialise things as an array (`[]`) the moment we add a second field, even into a tuple struct or inside a newtype :(

PR #1640, opened 2022-03-15
Actually activate `prometheus` feature flag in xtra

thomaseizinger:
Otherwise the metrics are not going to show up.

Tested locally this time that they are actually there.


thomaseizinger:
borsing on the premise that https://github.com/itchysats/itchysats/pull/1607 was approved.

bors r+

PR #1642, opened 2022-03-15
Correct liquidation price in the UI

### LINKED ISSUE #1641: Make sure the liquidation price in the Taker UI is correct
It appears the liquidation price is flipped. Fix.

da-kami:
Fixes https://github.com/itchysats/itchysats/issues/1641

The liquidation price is only relevant for the UI:
- Remove the liquidation price from the `Order` (finally...)
- Send the liquidation price according to own position in the projection (the only thing that makes sense)

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -169,6 +169,9 @@ fn assert_eq_orders(mut published: CfdOrder, received: CfdOrder) {
     // align initial_funding_fee_per_lot so we can assert on the order
     published.initial_funding_fee_per_lot = received.initial_funding_fee_per_lot;
 
+    // align liquidation price so we can assert on the order
+    published.liquidation_price = received.liquidation_price;
+
     assert_eq!(published, received);
 
     // Hard-coded to match the dummy_new_order()
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -230,9 +230,6 @@ pub struct Order {
     #[serde(rename = "leverage")]
     pub leverage_taker: Leverage,
 
-    // TODO: Remove from order, can be calculated
-    pub liquidation_price: Price,
-
     pub creation_timestamp: Timestamp,
 
     /// The duration that will be used for calculating the settlement timestamp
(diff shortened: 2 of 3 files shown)


daemon-tests/tests/happy_path.rs
```diff
@@ -169,6 +169,9 @@ fn assert_eq_orders(mut published: CfdOrder, received: CfdOrder) {
     // align initial_funding_fee_per_lot so we can assert on the order
     published.initial_funding_fee_per_lot = received.initial_funding_fee_per_lot;
 
+    // align liquidation price so we can assert on the order
+    published.liquidation_price = received.liquidation_price;
```
da-kami, line 173: Note: I thought about if we can easily assert on the actual values, but it cannot easily be achieved and I don't think the actor tests are responsible for that so I did not invest time.

model/src/cfd.rs
```diff
@@ -266,19 +263,13 @@ impl Order {
...
-        };
-
         Order {
             id: OrderId::default(),
             price,
             min_quantity,
             max_quantity,
             leverage_taker: leverage_choices_for_taker,
             trading_pair: TradingPair::BtcUsd,
-            liquidation_price,
```
luckysori, line 281: ❓ Is this a breaking change for our API or database?

luckysori (approved):

model/src/cfd.rs
da-kami: good point, the `Order` is still part of `MakerOffers` on the wire, but since that API changed anyway I think it should be fine to change this. 
My understanding is, that we will have two makers running and old takers are going against the old maker, new takers against the new maker. That should be fine. The automation is fine because the projection API did not change.

Or am I missing something @klochowicz  @thomaseizinger ? 🤔


For the db it does not matter. We are not saving the `Order` anymore.

thomaseizinger: Yep, this is all correct AFAIK. No relevant breaking changes here!

thomaseizinger (approved):

model/src/cfd.rs
bonomat: The automation does not use it but it might still need that field for deserialization. So better double check that it does not break anything.

PR #1645, opened 2022-03-15
Demote warn to debug

thomaseizinger:
It is worth monitoring this log because it may mean that a user has a
weird setup where they try to connect with the same seed from different
machines or something like that.

However, it is not worth printing a `warn` because it is in no way
critical to the application's health.

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -348,7 +348,7 @@ impl Actor {
         let this = ctx.address().expect("we are alive");
 
         if self.connections.contains_key(&identity) {
-            tracing::warn!(
+            tracing::debug!(
                 taker_id = %identity,
                 "Refusing to accept 2nd connection from already connected taker!"
             );


bonomat (approved):

daemon/src/maker_inc_connections.rs
```diff
@@ -348,7 +348,7 @@ impl Actor {
         let this = ctx.address().expect("we are alive");
 
         if self.connections.contains_key(&identity) {
-            tracing::warn!(
```
klochowicz, line 351: I'm not sure about this change - I feel like this should be `WARN` as it is abnormal operation. Addressing the spam right now should be dealt with by fixing the error, which really affects the takers in question :(

klochowicz (changes requested):

daemon/src/maker_inc_connections.rs
thomaseizinger: I tried to address the reasoning in my commit message! I think it is worthwhile tracking and hence keeping the log but logging it on `warn` is too much IMO. `warn` should be reserved for things that are really, really weird, shit is about to blow up but for now things are still somewhat working.

The only level up from there is then `error` which should need immediate attention.

If we apply this thinking consistently, grepping for `warn` and `error` in the logs will yield much more useful things!

klochowicz: As I see it:
- `error` is for things that are not working
- `warn` is for something that shouldn't happen, that is not yet an error, but could result in weirdness.
- `info` - really useful info that it confirms things are 
- `debug` - something that's not even worth logging unless you enable the level explicitly.
- 
Looking at these, IMHO it is much closer to `warn` than `debug` - as it is abnormal operation and we would like to know about it every time it happens.

note after this happens, if the taker in question had any open positions, we will fail rollovers - this leads to real problems, and having it on `warn` would make it easier to spot when the weirdness happened.

I feel that this bug should be dealt with as soon as we can, and hiding it under the carpet serves not purpose. :( If we fix the bug, these log messages will stop.

thomaseizinger: > note after this happens, if the taker in question had any open positions, we will fail rollovers

That is not true, if this happens, then the taker still has an operational connection. That is the whole reason why the log is being printed! Rollovers will work over that initial connection.

klochowicz: > > note after this happens, if the taker in question had any open positions, we will fail rollovers
> 
> That is not true, if this happens, then the taker still has an operational connection. That is the whole reason why the log is being printed! Rollovers will work over that initial connection.

oh, I stand corrected then! seems like I have misunderstood the seriousness of the message.

klochowicz (approved):
After the discussion I got convinced this might be the better log level for this message.

PS It would be good to amend the commit message to state  "Demote 2nd connection warning to debug" or something like that, otherwise it gives a false impression as we're moving them all or sth!

thomaseizinger:
> After the discussion I got convinced this might be the better log level for this message.
> 
> PS It would be good to amend the commit message to state "Demote 2nd connection warning to debug" or something like that, otherwise it gives a false impression as we're moving them all or sth!

Amended.

thomaseizinger:
@klochowicz feel free to bors once re-approved.

klochowicz (approved):

PR #1646, opened 2022-03-16
Restore compatibility of wire protocol with version 2.0.0

thomaseizinger:
By sending out _short_ offers to all takers as well, we can retain
compatibility with the old wire protocol. An older taker will
receive both messages and fail to deserialize one of them. This
is not a hard error so the app will still function. A new taker
will also receive both messages but simply drop the old message.

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -570,6 +570,9 @@ impl Actor {
                     .log_failure("Failed to forward current order from maker")
                     .await;
             }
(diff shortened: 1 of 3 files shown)


klochowicz (approved):
Looks good! 👍 i

let's see how our testnet setup will cope with the 0.4.7 taker against master maker :)

luckysori (approved):
LGTM.

PR #1647, opened 2022-03-16
Distinguish major and minor wire versions when checking for compatibility

da-kami:
- Major version has to match exactly.
- Taker cannot be ahead of maker for the minor or patch version. -> I am not entirely sure about this one. We could also allow that, but it would be quite odd of a taker runs a newer version than the maker. That would mean that the maker was not updated correctly and the taker should potentially contact us.

### DIFF
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -70,11 +70,6 @@ impl Version {
             return false;
         }
 
-        if taker.patch > maker.patch {
-            // taker's patch version is ahead of taker, that is not allowed
-            return false;
-        }
-
         true
     }
 }
@@ -620,41 +615,9 @@ mod tests {
     }
 
     #[test]
-    fn given_maker_patch_ahead_of_taker_then_versions_compatible() {
-        let maker_version = Version(semver::Version::new(2, 0, 1));
-        let taker_version = Version(semver::Version::new(2, 0, 0));
-
-        assert!(Version::is_compatible(&taker_version, &maker_version));
-    }
-
-    #[test]
-    fn given_maker_minor_and_patch_ahead_of_taker_then_versions_compatible() {
-        let maker_version = Version(semver::Version::new(2, 1, 1));
-        let taker_version = Version(semver::Version::new(2, 0, 0));
-
-        assert!(Version::is_compatible(&taker_version, &maker_version));
-    }
-
-    #[test]
     fn given_taker_minor_ahead_of_maker_then_versions_incompatible() {
         let maker_version = Version(semver::Version::new(2, 0, 0));
         let taker_version = Version(semver::Version::new(2, 1, 0));
-
-        assert!(!Version::is_compatible(&taker_version, &maker_version));
-    }
-
-    #[test]
-    fn given_taker_patch_ahead_of_maker_then_versions_incompatible() {
-        let maker_version = Version(semver::Version::new(2, 0, 0));
-        let taker_version = Version(semver::Version::new(2, 0, 1));
-
-        assert!(!Version::is_compatible(&taker_version, &maker_version));
-    }
-
-    #[test]
-    fn given_taker_minor_and_patch_ahead_of_maker_then_versions_incompatible() {
-        let maker_version = Version(semver::Version::new(2, 0, 0));
-        let taker_version = Version(semver::Version::new(2, 1, 1));
 
         assert!(!Version::is_compatible(&taker_version, &maker_version));
     }
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -50,8 +50,32 @@ pub type Write<D, E> = SplitSink<Framed<TcpStream, EncryptedJsonCodec<D, E>>, E>
 pub struct Version(semver::Version);
 
 impl Version {
+    const CURRENT: semver::Version = semver::Version::new(2, 0, 0);
+
     pub fn current() -> Self {
-        Self(semver::Version::new(2, 0, 0))
+        Self(Self::CURRENT)
+    }
+
+    pub fn is_compatible(taker: &Self, maker: &Self) -> bool {
+        let taker = &taker.0;
+        let maker = &maker.0;
+
+        if taker.major != maker.major {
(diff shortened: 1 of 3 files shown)


daemon/src/wire.rs
```diff
@@ -50,8 +50,32 @@ pub type Write<D, E> = SplitSink<Framed<TcpStream, EncryptedJsonCodec<D, E>>, E>
...
+            // major versions don't match
+            return false;
+        }
+
+        if taker.minor > maker.minor {
+            // taker's minor version is ahead of taker, that is not allowed
+            return false;
+        }
+
+        if taker.patch > maker.patch {
```
klochowicz: I would consider relaxing the patch requirement - the patch versions ought to be compatible if we're daring to use SemVer here

daemon/src/wire.rs
```diff
@@ -50,8 +50,32 @@ pub type Write<D, E> = SplitSink<Framed<TcpStream, EncryptedJsonCodec<D, E>>, E>
...
 
 impl Version {
+    const CURRENT: semver::Version = semver::Version::new(2, 0, 0);
+
     pub fn current() -> Self {
-        Self(semver::Version::new(2, 0, 0))
+        Self(Self::CURRENT)
+    }
+
+    pub fn is_compatible(taker: &Self, maker: &Self) -> bool {
```
klochowicz, line 59: This could be also a method that could be used like this:

`our_version.is_compatible_with(&taker_version)` 

I'm not sure whether that would be better though (perhaps a bit more discoverable in the IDE?)

klochowicz (approved):
LGTM!

thanks for covering it extensively with tests 👍

daemon/src/wire.rs
da-kami: That's what I had initially and writing the tests became more cumbersome so I opted for this one.

daemon/src/wire.rs
da-kami: I agree that we should relax it for patch. I added a fixup that removes the check for patch.

We gave `semver`s `VersionReq` (as in the requirement for comparison) a try as well, but I feel it just made the code more complex. I feel the current logic is quite easy to understand and we can quite easily change the conditions.

daemon/src/wire.rs
```diff
@@ -619,43 +614,11 @@ mod tests {
...
-    }
-
-    #[test]
-    fn given_maker_minor_and_patch_ahead_of_taker_then_versions_compatible() {
-        let maker_version = Version(semver::Version::new(2, 1, 1));
-        let taker_version = Version(semver::Version::new(2, 0, 0));
-
-        assert!(Version::is_compatible(&taker_version, &maker_version));
-    }
-
```
klochowicz: I would probably keep these tests and modify the other ones! having them won't hurt, we're cementing explicit support for patch version changes :)

klochowicz (approved):

thomaseizinger:
Thanks for the extensive effort that went into this! I did outline another idea on matrix that may make this unnecessary, reposting here:

> Instead of seeing it as "I am sending my version", we can see it as a "version negotiation" protocol.
> Taker->Maker: I am proposing we speak version 2.0.0.
> Maker->Taker: Confirming to speak version 2.0.0.
> Taker->Maker: I am proposing we speak version 2.1.0.
> Maker->Taker: Confirming to speak version 2.1.0.
> That is actually not hacky at all. Just a version negotiation protocol.

What do you think?

da-kami:
> Thanks for the extensive effort that went into this! I did outline another idea on matrix that may make this unnecessary, reposting here:
> 
> > Instead of seeing it as "I am sending my version", we can see it as a "version negotiation" protocol.
> > Taker->Maker: I am proposing we speak version 2.0.0.
> > Maker->Taker: Confirming to speak version 2.0.0.
> > Taker->Maker: I am proposing we speak version 2.1.0.
> > Maker->Taker: Confirming to speak version 2.1.0.
> > That is actually not hacky at all. Just a version negotiation protocol.
> 
> What do you think?

At the moment we send the version over to each other irregardless of actually accepting the connection. What you are proposing would mean, that the maker would just close the connection to the taker in case the protocol version cannot be confirmed? 
What is the advantage of that over sending versions to each other and let each party decide if they can go forward? 
The only thing I can think of is, that it would be more resilient against one party accepting while the other party declines, is that what you have in mind?

thomaseizinger:
> > Thanks for the extensive effort that went into this! I did outline another idea on matrix that may make this unnecessary, reposting here:
> > > Instead of seeing it as "I am sending my version", we can see it as a "version negotiation" protocol.
> > > Taker->Maker: I am proposing we speak version 2.0.0.
> > > Maker->Taker: Confirming to speak version 2.0.0.
> > > Taker->Maker: I am proposing we speak version 2.1.0.
> > > Maker->Taker: Confirming to speak version 2.1.0.
> > > That is actually not hacky at all. Just a version negotiation protocol.
> > 
> > 
> > What do you think?
> 
> At the moment we send the version over to each other irregardless of actually accepting the connection. What you are proposing would mean, that the maker would just close the connection to the taker in case the protocol version cannot be confirmed? What is the advantage of that over sending versions to each other and let each party decide if they can go forward? The only thing I can think of is, that it would be more resilient against one party accepting while the other party declines, is that what you have in mind?

Not quite. Checkout https://github.com/itchysats/itchysats/pull/1649.

I am not closing the connection if we don't support the version, I am simply sending over the one I want to speak which is always latest and then closing the connection. We could evolve this to not directly close the connection afterwards but wait for the client to close it in case they don't support that version. In otherwise, in theory you could imagine the following to work:

Taker->Maker: Proposing version 2.0.0
Maker (checks version: nope, 2.0.0 is too old)
Maker->Taker: Setting version 2.2.0
Taker (checks response, yeah that is fine too): Leaves connection open

It doesn't make a whole lot of sense though because with this behaviour, a taker should have just sent 2.2.0 from the beginning if they support it.

da-kami:
Closing this in favor of https://github.com/itchysats/itchysats/pull/1649

PR #1648, opened 2022-03-16
Print taker version number in new message

klochowicz:
Add a HelloV2 message that is compatible with the version 2.0 of the wire protocol.

See commit messages for more details.


da-kami (approved):
Nice!

We would only run into trouble if the taker is already updated, but the maker is not yet. Given that we are always planning to upgrade the maker first that should not be a problem.

daemon/src/maker_cfd.rs
```diff
@@ -682,6 +682,9 @@ where
             wire::TakerToMaker::Hello(_) => {
                 unreachable!("The Hello message is not sent to the cfd actor")
             }
+            wire::TakerToMaker::HelloV2 { .. } => {
+                unreachable!("The HelloV2 message is not sent to the cfd actor")
+            }
```
luckysori, line 687: 😭 Yikes, we still have these.

daemon/src/maker_inc_connections.rs
```diff
@@ -489,17 +489,16 @@ async fn upgrade(
...
-                .send(wire::MakerToTaker::Hello(our_version.clone()))
-                .await?;
-
-            if our_version != taker_version {
-                bail!(
-                    "Network version mismatch, we are on version {our_version} but taker is on version {taker_version}",
-                );
-            }
+        wire::TakerToMaker::Hello(taker_wire_version) => {
+            tracing::info!(taker_id = %taker_id, %taker_wire_version, "Received Hello message from taker (version <=0.4.7");
```
luckysori: ❓ Can't you replace `taker_id = %taker_id` with just `%taker_id`?

daemon/src/wire.rs
```diff
@@ -80,7 +80,12 @@ pub mod taker_to_maker {
 #[serde(tag = "type", content = "payload")]
 #[allow(clippy::large_enum_variant)]
 pub enum TakerToMaker {
+    // Deprecated, used by takers up to v0.4.7
```
luckysori: 🙃 I think this could be a docstring.

luckysori (approved):
LGTM!

daemon/src/maker_inc_connections.rs
klochowicz: changed in all 3 cases (added a separate commit)

thomaseizinger (approved):
Thanks!

Can you - as a follow up - reset the version of the `daemon` crate to `0.1.0` and remove the bumping from the release workflow? We should probably document this in an ADR as well. I find setting the crate version adds too much confusion. We are not releasing any of these crates to crates.io and thus, the version set in the manifest doesn't matter at all.

klochowicz:
@thomaseizinger I don't think this follow-up has anything to do with these changes? We do not want to send 0.1.0, but 0.4.7 here after all. Happy to chat after I get online.

thomaseizinger:
> @thomaseizinger I don't think this follow-up has anything to do with these changes? We do not want to send 0.1.0, but 0.4.7 here after all. Happy to chat after I get online.

Correct. It is related in that the version set in the `Cargo.toml` is not actually used for anything because we always take it out of git every time we compile!

PR #1649, opened 2022-03-16
Remodel `Hello` messages as version negotiation and transform messages based on negotiated version

thomaseizinger:
This is an attempt at relaxing the version requirements of our wire protocol. What we have currently in master **does not** work  (for 0.4.7, sorry for not testing this better first :grimacing: ).

Here is screenshot of this branch in action:

image

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -349,7 +349,7 @@ impl Actor {
             Framed::new(connection, EncryptedJsonCodec::new(noise)).split()
         };
 
-        let proposed_version = Version::current();
+        let proposed_version = Version::latest();
         write
             .send(wire::TakerToMaker::HelloV2 {
                 proposed_wire_version: proposed_version.clone(),
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -349,7 +349,7 @@ impl Actor {
             Framed::new(connection, EncryptedJsonCodec::new(noise)).split()
         };
 
-        let proposed_version = Version::latest();
+        let proposed_version = Version::LATEST;
         write
             .send(wire::TakerToMaker::HelloV2 {
                 proposed_wire_version: proposed_version.clone(),
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -188,9 +188,9 @@ pub enum ConnectionStatus {
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum ConnectionCloseReason {
-    VersionMismatch {
-        taker_version: Version,
-        maker_version: Version,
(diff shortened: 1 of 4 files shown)


da-kami (approved):
Thanks for iterating on this! I think this is a good solution for our current problem.

I like the fact that we are explicit about that we speak an old version and the latest one. 
I think this even scales to next versions if it has to. Not saying that should add more and more version code, but it gives us the flexibility that we need at the moment and it should not cause trouble with future releases so I think it is a good path forward.

klochowicz (commented):
Concept ACK - have only looked briefly from my mobile phone, will look again when I get home happy for it to get through if anyone else approves earlier :).

daemon/src/connection.rs
```diff
@@ -364,19 +364,19 @@ impl Actor {
                 )
             })?
             .with_context(|| format!("Failed to read first message from maker {maker_identity}"))? {
-            Some(wire::MakerToTaker::Hello(maker_version)) => {
-                if our_version != maker_version {
+            Some(wire::MakerToTaker::Hello(actual_version)) => {
```
luckysori, line 371: ❓ Is it a little weird that we get the same regular `Hello` message from the maker whether they agree with our version or not?

daemon/src/maker_inc_connections.rs
```diff
@@ -182,7 +182,7 @@ impl Actor {
         heartbeat_interval: Duration,
         p2p_socket: SocketAddr,
     ) -> Self {
-        NUM_CONNECTIONS_GAUGE.set(0);
+        NUM_CONNECTIONS_GAUGE.reset();
```
luckysori, line 191: ❓ This is equivalent?

luckysori (approved):
Very clear commits, thank you!

daemon/src/connection.rs
thomaseizinger: It is not about agreeing, it is a statement / confirmation of "we will be using this version". It is a bit like HTTP content negotiation:

Client->Server: GET /foo, Accept: application/json
Server->Client: 200 OK, Content-Type: application/xml

The client asks for JSON but in the end, the server has the "power" to just say, nah here is some XML.

daemon/src/maker_inc_connections.rs
thomaseizinger: As far as I know, yes.

https://docs.rs/prometheus/0.13.0/prometheus/core/struct.MetricVec.html#method.reset

luckysori (approved):

da-kami (approved):

klochowicz (approved):

PR #1652, opened 2022-03-17
Separate initialization of Cfds in the projection

da-kami:


daemon/src/projection.rs
```diff
@@ -858,27 +879,9 @@ impl xtra::Actor for Actor {
...
-                    anyhow::Ok(())
-                }
-            },
-            |e| async move {
-                tracing::warn!("Failed to do initial rehydratation of CFDs: {e:#}");
-            },
-        );
+        this.send_async_safe(Initialize)
+            .await
+            .expect("we just started");
```
thomaseizinger, line 884: Can we double check that this `expect` is on the `xtra::Disconnected`?

thomaseizinger (approved):

daemon/src/projection.rs
klochowicz: just checked, this is the case: https://github.com/itchysats/itchysats/blob/master/xtras/src/send_async_safe.rs#L40

daemon/src/projection.rs
```diff
@@ -58,6 +59,10 @@ pub struct Update<T>(pub T);
 #[derive(Clone, Copy)]
 pub struct CfdChanged(pub OrderId);
 
+/// Indicates that all Cfd have to be loaded initially
```
klochowicz: I found this comment more confusing than helpful, we might as well delete it or say something like:

"Perform the bulk initialisation of the CFD feed"

klochowicz (approved):

da-kami:
I tested this locally and it works as expected (release build and big db). The full list of `Cfd`s only becomes available after `1min 44sec` - until then the list of cfds is empty. Once we work on improving the storage / loading performance that value should improve :)

daemon/src/projection.rs
```diff
@@ -858,27 +879,9 @@ impl xtra::Actor for Actor {
...
-                    }
-
-                    anyhow::Ok(())
-                }
-            },
-            |e| async move {
-                tracing::warn!("Failed to do initial rehydratation of CFDs: {e:#}");
-            },
-        );
+        this.send_async_safe(Initialize)
```
luckysori, line 882: ❓ What's the logic behind using `send_async_safe` here? Can we run into trouble by letting the queue of `CfdChanged` messages grow whilst we process the `Initialize` message?

luckysori (approved):
Only one question about the use of `send_async_safe`.

daemon/src/projection.rs
da-kami: Theoretically yes and that *could* lead to failures if e.g. takers trigger rollovers when we start up the maker (which can happen sometimes, if the maker was down for longer or we start right after the full hour), but that should not cause problems. 

Requests would be piling up until the initial loading of the projection is done. Note, that those requests piling up is not really an issue. If the maker does not respond to the taker in time the taker's rollover protocol will just timeout eventually. 
Same for contract setup, and same for collab settlement - it might just gracefully fail on the taker side if the maker does not respond because it is still starting up and is thus not accepting protocol requests.

Of course this is not optimal and our storage and loading performance optimizations should make it in soon. But we are already working on that so I don't see blockers.

PR #1654, opened 2022-03-17
Make rollover 0.4.7 compatible with current master

da-kami:
Introduce `ProposeRolloverV2` message on the wire:

- The taker sends `ProposeRolloverV2`
- The maker can handle both `ProposeRolloverV2` and `ProposeRollover`

Introduce `RolloverVersion` to distinguish code paths in `rollover_maker` actor and `accept_rollover_proposal` for the maker.

I opted for having `RolloverVersion` because the decision logic is simple, and we have to store the version in the `rollover_maker` actor because the proposal does not need the decision logic, only after accepting we decide. It might still be somewhat cleaner to duplicate `accept_rollover_proposal` and deprecate one of the functions - but since I needed the `RolloverVersion` anyway I pulled it in there. 
We could also duplicate all code on the way: the `rollover_maker` actor and  `accept_rollover_proposal` in `model::cfd::Cfd` and mark all duplicated code as deprecated. But I found that a bit of an overkill.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -836,8 +836,14 @@ impl Cfd {
         }
 
         let hours_to_charge = match version {
-            rollover::Version::V1 => 1,
-            rollover::Version::V2 => self.hours_to_extend_in_rollover()?,
+            rollover::Version::V1 => {
+                tracing::debug!("Rollover V1");
+                1
+            }
+            rollover::Version::V2 => {
+                tracing::debug!("Rollover V2");
+                self.hours_to_extend_in_rollover()?
+            }
         };
 
         let funding_fee = calculate_funding_fee(
@@ -860,6 +866,7 @@ impl Cfd {
                 tx_fee_rate,
                 self.fee_account,
                 funding_fee,
+                version,
             ),
             self.dlc.clone().context("No DLC present")?,
             self.position,
@@ -903,6 +910,7 @@ impl Cfd {
                 tx_fee_rate,
                 self.fee_account,
                 funding_fee,
+                rollover::Version::V2,
             ),
             self.dlc.clone().context("No DLC present")?,
             self.position,
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -4,6 +4,7 @@ use crate::hex_transaction;
 use crate::olivia;
 use crate::olivia::BitMexPriceEventId;
 use crate::payout_curve;
+use crate::rollover;
 use crate::rollover::RolloverParams;
 use crate::FeeAccount;
 use crate::FeeFlow;
@@ -824,6 +825,7 @@ impl Cfd {
         self,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        version: rollover::Version,
     ) -> Result<(CfdEvent, RolloverParams, Dlc, Position, Duration)> {
         if !self.during_rollover {
             bail!("The CFD is not rolling over");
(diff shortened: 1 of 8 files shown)


model/src/rollover.rs
```diff
@@ -5,6 +5,12 @@ use crate::Price;
 use crate::TxFeeRate;
 use crate::Usd;
 
+#[derive(Debug, Clone, Copy)]
+pub enum Version {
+    V1_0_4_7,
+    Current,
```
thomaseizinger: Just make these 1 and 2?

I think this should version the rollover protocol itself, not which version of the daemon this is implemented.

Also "current" is not a version :)

thomaseizinger: This enum would also be a good place to document the difference in the 2 versions of the protocol!

thomaseizinger (commented):
Some thoughts, I'll leave you to it :)

model/src/rollover.rs
da-kami: I agree with what you are pointing out and will rename to `V1` and `V2` but probably still mention in rustdoc for what version `V1` was for if that's OK. 

I feel "if this grows", i.e. if we at some point have multiple version it would be great to know for what version we introduced them. That could make it easier to monitor what to remove when we know that there are no takers running 0.4.7 anymore.

model/src/cfd.rs
```diff
@@ -16,6 +17,7 @@ use crate::OpeningFee;
 use crate::Percent;
 use crate::Position;
 use crate::Price;
+use crate::RolloverVersion;
```
luckysori: 🔧 I don't think we need this type. It should be enough to use `rollover::Version` in this file.

model/src/lib.rs
```diff
@@ -33,6 +33,7 @@ mod rollover;
 pub use cfd::*;
 pub use contract_setup::SetupParams;
 pub use rollover::RolloverParams;
+pub use rollover::Version as RolloverVersion;
```
luckysori, line 36: :+1: I agree with exporting without leaking the internal structure of `model`.

model/src/cfd.rs
```diff
@@ -833,7 +836,11 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
-        let hours_to_charge = self.hours_to_extend_in_rollover()?;
+        let hours_to_charge = match version {
+            Version::V1_0_4_7 => 1,
+            Version::Current => self.hours_to_extend_in_rollover()?,
+        };
```
luckysori: 👍 Having discussed the merits of duplicating code vs introducing branching logic I think this is okay. We can always revisit this choice in the future if we notice that we're having to introduce a lot of code like this.

luckysori (approved):
I approve.

model/src/rollover.rs
```diff
@@ -5,6 +5,21 @@ use crate::Price;
 use crate::TxFeeRate;
 use crate::Usd;
 
+#[derive(Debug, Clone, Copy)]
+pub enum Version {
+    /// Version one of the rollover protocol
+    ///
+    /// This version cannot handle charging for "missed" rollovers yet, i.e. the hours to charge is
+    /// always set to 1 hour. This version is needed for clients that are <= daemon version
+    /// `0.4.7`.
```
luckysori, line 14: I understand why you're putting this comment here, but it's a bad sign that our model "knows" about daemon versions. Nothing to action, just a thought.

da-kami: I totally agree, I will see to given our versioning ADR a review later!

model/src/rollover.rs
klochowicz: this sounds like a good idea - in fact, this would be in line with what I've already done for `HelloV2` message (rustdoc comment that `Hello` was used by takers up to 0.4.7).

model/src/rollover.rs
klochowicz: From an utilitarian POV, I'd say that the model ha also 0.4.7 version :)

klochowicz (approved):

da-kami:
I did a test just now, and we are still failing to rollover with `0.4.7`. Investigating if we are triggering the correct code path, but I assume we are.

PR #1657, opened 2022-03-21
Verify taker's collaborative settlement signature

### LINKED ISSUE #1629: Verify collaborative settlement transaction signature before considering it completed
We've recently had a bug where we produced an invalid signature. It is unclear how it happened. Adding an additional check would allow us to abort the protocol early.

Completing the protocol successfully but being unable to publish the transaction "only" results in a log message about a failed transaction but does not change the state which makes it impossible to alter this state again.

Ideally, we would append another event on top if we realize that a transaction that we published never got confirmed but our monitoring component does not have this capability at the moment.

luckysori:
Fixes #1629.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1958,7 +1958,6 @@ mod tests {
     use bdk::bitcoin::util::psbt::Global;
     use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
     use bdk_ext::keypair;
-    use bdk_ext::AddressExt;
     use bdk_ext::SecretKeyExt;
     use maia::lock_descriptor;
     use proptest::prelude::*;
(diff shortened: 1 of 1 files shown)


da-kami:
It appears the tests are actually failing because of the changes. I could not spot the problem so far, will have another thorough look tomorrow morning :)

bonomat (approved):
I don't a the problem. Could be just an issue with our dummy/test setup

luckysori:
@bonomat @da-kami The problem was in the test code like we thought. I added https://github.com/itchysats/itchysats/pull/1657/commits/9733d13e0464d28a9a9aa886428c227bd777ddbe so that the maker and taker in the tests agree on their addresses, and signature verifications (which we weren't doing until this PR) can succeed.

PR #1658, opened 2022-03-21
Taker checks for offer sanity

da-kami:
Motivation: 
Besides our discussion ensuring sanity on the maker side, given that we are in the P2P world, the more important issue is actually the taker ensuring sanity as good as possible. 

Implementation:
First we check that the maker offer's creation timestamp is not outdated. We consider a creation timestamp that is older than 10 minutes as outdated, i.e. if we don't receive a new offer from the maker within 10 minutes the taker will consider the offer as outdated when taking.
I think it is fair to assume that there will be *any* change that triggers the creation of new `MakerOffers` within 10 minutes.
Note: We currently trust the maker to send a sane `creation_timestamp`.

Furthermore, the oracle event id sent by the maker is checked for sanity.
The oracle event id is considered sane if it is within the interval: [25h,23h]
We choose 25h because we always ceil to the next full hour.
The maker cannot go back in time, as soon as the maker is past the full hour the next oracle event ID will be picked (anything else would be insane).
See tests for logic details.

If the taker's sanity check fails the offer will be removed on the taker side and an error is returned that indicates that the offer is outdated.
The taker does not start a contract setup if the sanity check fails, i.e. the taker does not send any messages to the maker if the sanity check fails.

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -142,7 +142,7 @@ impl Connection {
                             min_quantity: order.min_quantity,
                             max_quantity: order.max_quantity,
                             leverage_taker: order.leverage_taker,
-                            creation_timestamp: order.creation_timestamp,
+                            creation_timestamp: order.creation_timestamp_maker,
                             settlement_interval: order.settlement_interval,
                             liquidation_price: model::calculate_long_liquidation_price(
                                 order.leverage_taker,
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -6,7 +6,8 @@ use crate::process_manager;
 use crate::projection;
 use crate::setup_taker;
 use crate::wallet;
-use anyhow::Context as _;
+use anyhow::bail;
+use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
@@ -18,6 +19,7 @@ use model::Origin;
 use model::Price;
 use model::Role;
 use model::Usd;
+use time::OffsetDateTime;
 use tokio_tasks::Tasks;
 use xtra::Actor as _;
 use xtra_productivity::xtra_productivity;
@@ -167,7 +169,18 @@ where
             .take_order(order_id);
 
         let order_to_take = order_to_take.context("Order to take could not be found in current maker offers, you might have an outdated offer")?;
+
+        // The offer we are instructed to take is removed from the
+        // set of available offers immediately so that we don't attempt
+        // to take it more than once
         self.current_maker_offers.replace(maker_offers);
+
+        if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
+            bail!(
+                "The maker's offer appears to be outdated, refusing to take: {:?}",
+                &order_to_take
+            );
+        }
 
         tracing::info!("Taking current order: {:?}", &order_to_take);
 
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -230,7 +230,9 @@ pub struct Order {
     #[serde(rename = "leverage")]
     pub leverage_taker: Leverage,
 
-    pub creation_timestamp: Timestamp,
+    /// The creation timestamp as set by the maker
+    #[serde(rename = "creation_timestamp")]
+    pub creation_timestamp_maker: Timestamp,
 
     /// The duration that will be used for calculating the settlement timestamp
     pub settlement_interval: Duration,
@@ -271,7 +273,7 @@ impl Order {
             leverage_taker: leverage_choices_for_taker,
             trading_pair: TradingPair::BtcUsd,
             position_maker,
-            creation_timestamp: Timestamp::now(),
+            creation_timestamp_maker: Timestamp::now(),
             settlement_interval,
             origin,
             oracle_event_id,
@@ -295,6 +297,41 @@ impl Order {
             self.funding_rate,
             self.opening_fee,
         )
+    }
+
+    /// Defines when we consider an order to be outdated
+    ///
+    /// If the maker's offer creation timestamp is older than `OUTDATED_AFTER_MINS` minutes then we
+    /// consider an order to be outdated.
+    const OUTDATED_AFTER_MINS: i64 = 10;
+
+    /// Defines when we consider the order to be outdated.
+    ///
+    /// This is used as a safety net to prevent the taker from taking an outdated order.
+    pub fn is_safe_to_take(&self, now: OffsetDateTime) -> bool {
+        !self.is_creation_timestamp_outdated(now) && self.is_oracle_event_timestamp_sane(now)
+    }
+
+    /// Check if the the maker's offer creation timestamp is outdated
+    ///
+    /// If the creation timestamp is older than `OUTDATED_AFTER_MINS` minutes the offer is
+    /// considered outdated
+    fn is_creation_timestamp_outdated(&self, now: OffsetDateTime) -> bool {
+        self.creation_timestamp_maker.seconds() + (Self::OUTDATED_AFTER_MINS * 60)
+            < now.unix_timestamp()
+    }
+
+    /// Check the oracle event's timestamp for sanity
+    ///
+    /// An id within [25h, 23h] from now is considered sane.
+    fn is_oracle_event_timestamp_sane(&self, now: OffsetDateTime) -> bool {
+        let event_id_timestamp = self.oracle_event_id.timestamp();
+
+        let settlement_interval_minus_one_hour = now + SETTLEMENT_INTERVAL - Duration::HOUR;
+        let settlement_interval_plus_one_hour = now + SETTLEMENT_INTERVAL + Duration::HOUR;
+
+        event_id_timestamp >= settlement_interval_minus_one_hour
+            && event_id_timestamp <= settlement_interval_plus_one_hour
     }
 }
 
@@ -1962,6 +1999,7 @@ mod tests {
     use bdk_ext::SecretKeyExt;
     use maia::lock_descriptor;
     use proptest::prelude::*;
+    use rand::thread_rng;
     use rust_decimal_macros::dec;
     use std::collections::BTreeMap;
     use std::str::FromStr;
@@ -3069,7 +3107,6 @@ mod tests {
         (taker_payout.as_sat(), maker_payout.as_sat())
     }
 
-    use rand::thread_rng;
     proptest! {
         #[test]
         fn rollover_funding_fee_collected_incrementally_should_not_be_smaller_than_collected_once_per_settlement_interval(quantity in 1u64..100_000u64) {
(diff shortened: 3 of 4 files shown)


da-kami:
Note: This is a a bare minimum of checks. Additionally, we could:

* Add an additional `receival_timestamp` for the taker - this goes towards refactoring the `Order` to be represented differently on each side. The taker should be able to record when the `Order` (which is actually an `Offer`...) was received. We could also store this separately within `taker_cfd` to reason about `outdated`. For now I just used the maker's `creation_timestamp` and assume it is sane. I think eventually it would be good to check both - the taker's receival timestamp would be another layer of validation for the taker to ensure that the maker is acting sane.
* Let the taker check periodically if the offer is outdated, and remove it in case it is. That might result in a better UX, but is somewhat more intrusive.

da-kami:
I'd love to reduce the size of `cfd.rs` a bit more. I might try to move the `Order`, `MakerOffers` to a separate file tomorrow including the related tests. Should be possible without too much hassle. Would you consider this an improvement?

daemon/src/maker_inc_connections.rs
```diff
@@ -142,7 +142,7 @@ impl Connection {
                             min_quantity: order.min_quantity,
                             max_quantity: order.max_quantity,
                             leverage_taker: order.leverage_taker,
-                            creation_timestamp: order.creation_timestamp,
+                            creation_timestamp: order.creation_timestamp_maker,
```
luckysori, line 145: 🙃 It would be helpful if this were part of a different commit.

daemon/src/taker_cfd.rs
```diff
@@ -169,6 +171,17 @@ where
         let order_to_take = order_to_take.context("Order to take could not be found in current maker offers, you might have an outdated offer")?;
         self.current_maker_offers.replace(maker_offers);
 
+        // If the offer is not safe to take we return and error and the offer is removed (by taking
+        // it, see code above), so on the first take of an outdated offer we print the error and
+        // then the offer for the taker will be gone until a new offer arrives from the
+        // maker
+        if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
```
luckysori: 🔧 This could be a bit clearer. Particularly since you don't need to refer back to the "code above".

I don't think the code below needs a comment because it reads nicely: "if the order to take is not safe to take we refuse to take it because it appears to be outdated".

```suggestion
        // the offer we are instructed to take is removed from the
        // set of available offers immediately so that we don't attempt
        // to take it more than once
        self.current_maker_offers.replace(maker_offers);

        if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
```

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
             self.opening_fee,
         )
     }
+
+    const OUTDATED_AFTER_MINS: i64 = 10;
+
+    /// Defines when we consider the order to be sane within the system boundaries
+    ///
+    /// This is a safety net for the taker to prevent the taker from taking an outdated order.
+    pub fn is_safe_to_take(&self, now: OffsetDateTime) -> bool {
```
luckysori, line 311: This seems to be what you're trying to convey and also you can use it without negating it.

```suggestion
    pub fn is_outdated(&self, now: OffsetDateTime) -> bool {
```

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
             self.opening_fee,
         )
     }
+
+    const OUTDATED_AFTER_MINS: i64 = 10;
+
+    /// Defines when we consider the order to be sane within the system boundaries
+    ///
+    /// This is a safety net for the taker to prevent the taker from taking an outdated order.
```
luckysori: 🔧 A bit simpler?

```suggestion
    /// Defines when we consider the order to be outdated.
    ///
    /// This is used as a safety net to prevent the taker from taking an outdated order.
```

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
             self.opening_fee,
         )
     }
+
+    const OUTDATED_AFTER_MINS: i64 = 10;
```
luckysori, line 306: 🔧 Might be worth documenting this constant too!

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
...
+    /// Defines when we consider the order to be sane within the system boundaries
+    ///
+    /// This is a safety net for the taker to prevent the taker from taking an outdated order.
+    pub fn is_safe_to_take(&self, now: OffsetDateTime) -> bool {
+        !self.is_creation_timestamp_outdated(now) && self.is_oracle_event_id_sane_to_take(now)
+    }
+
+    /// Check if the the maker's offer creation timestamp is outdated
+    ///
+    /// If the creation timestamp is older than 10 minutes the offer is considered outdated
```
luckysori: 🔧 You never know when things might change... You could even argue that the docstring shouldn't need to explain what the code does (if it's clear enough).

```suggestion
    /// If the creation timestamp is older than `OUTDATED_AFTER_MINS` minutes the offer is considered outdated
```

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
...
+    fn is_creation_timestamp_outdated(&self, now: OffsetDateTime) -> bool {
+        self.creation_timestamp_maker.seconds() + (Self::OUTDATED_AFTER_MINS * 60)
+            < now.unix_timestamp()
+    }
+
+    /// Check the oracle event ID for sanity
+    ///
+    /// An id within [25h, 23h] from now is considered sane.
+    fn is_oracle_event_id_sane_to_take(&self, now: OffsetDateTime) -> bool {
```
luckysori: 🔧 Maybe a bit clearer.

Talking about event ID is always a bit confusing because it's not immediately obvious that the timestamp is embedded in there.

I'm not too fond of the word "sane" because it's generic, but it's hard to express concisely "far enough into the future, but not too far". I guess you could split this condition, but it might not be worth it.

```suggestion
    fn is_oracle_settlement_time_sane(&self, now: OffsetDateTime) -> bool {
```

model/src/cfd.rs
```diff
@@ -296,6 +298,35 @@ impl Order {
...
+        let event_id_timestamp = self.oracle_event_id.timestamp();
+
+        let in_23_hours = now + SETTLEMENT_INTERVAL - Duration::HOUR;
+        let in_25_hours = now + SETTLEMENT_INTERVAL + Duration::HOUR;
+
+        event_id_timestamp >= in_23_hours && event_id_timestamp <= in_25_hours
```
luckysori: 👍 I'm okay with this, even though I would probably bite the bullet and go with `settlement_interval_minus_one_hour`. This definitely reads better!

model/src/cfd.rs
```diff
@@ -3103,6 +3134,114 @@ mod tests {
     }
     }
 
+    #[test]
+    fn give_order_creation_timestamp_outdated_then_order_outdated() {
```
luckysori: 🔧 

```suggestion
    fn given_order_creation_timestamp_outdated_then_order_outdated() {
```

model/src/cfd.rs
```diff
@@ -3103,6 +3134,114 @@ mod tests {
...
+        let order = Order::dummy_short().with_creation_timestamp(creation_timestamp);
+
+        let now =
+            OffsetDateTime::now_utc() + Duration::seconds(Order::OUTDATED_AFTER_MINS * 60 + 1);
+
+        assert!(order.is_creation_timestamp_outdated(now))
+    }
+
+    #[test]
+    fn give_order_creation_timestamp_not_outdated_then_order_not_outdated() {
```
luckysori:
```suggestion
    fn given_order_creation_timestamp_not_outdated_then_order_not_outdated() {
```

model/src/cfd.rs
```diff
@@ -3103,6 +3134,114 @@ mod tests {
...
+
+        assert!(!order.is_creation_timestamp_outdated(now))
+    }
+
+    #[test]
+    fn given_oracle_event_id_is_24h_in_the_future_then_sane_to_take() {
+        // --|---------|---------|----------------------------------|--> time
+        //   -1h       |         +1h                                24h
+        // --|---------|<--------|--------------------------------->|--
+        //             now
```
luckysori, line 3170: 👍 Nice!

model/src/cfd.rs
```diff
@@ -3103,6 +3134,114 @@ mod tests {
...
+
+    #[test]
+    fn given_oracle_event_id_is_more_than_25h_in_the_future_then_not_sane_to_take() {
+        // --|---------|---------|----------------------------------|--> time
+        //   -1h1s     |         +1h                                24h
+        // --|---------|<--------|--------------------------------->|--
+        //   now
+
+        // Note the maker cannot go back in time, if an offer was created with a timestamp of 10:00
+        // it must have been at least 9:00:00
```
luckysori: 🤔 I'm not sure this comment is necessary here. These tests are for a function called by the taker. The taker doesn't need to know how the maker comes up with oracle event IDs. They just need to verify that what they get from the maker is acceptable to them.

luckysori (approved):
Great stuff, thank you for adding a lot of tests. 

I came up with a lot of small suggestions that could make these changes more readable. Take them or leave them :D

daemon/src/taker_cfd.rs
da-kami: fair enough :)

model/src/cfd.rs
da-kami: That's what the error message returned says, but internally the check touches more than just being outdated.
Initially I had `is_outdated` but I went away from it because I felt it was not precise enough. 
I would keep as is for now.

model/src/cfd.rs
da-kami: I totally get what you mean, but I would still keep it with the specific hours for readability. The main question is: Are we planning to change this constant any time soon? If so we should probably keep this more generic 😬

model/src/cfd.rs
da-kami: I added a doccomment, not sure how much it helps. 
I would question the existence of this constant - I created it so the test reads better. Not sure if it's worth that...

model/src/cfd.rs
da-kami: I removed it :)

model/src/cfd.rs
luckysori: Fair enough. Both conditions are about time, but the one related to the oracle settlement time is not necessarily about it being outdated.

model/src/cfd.rs
luckysori: I like that it exists and that you documented it. But that's just my opinion haha.

model/src/cfd.rs
bonomat: I'm actually tempted to change it ever since we had the issue with the oracle being down to more than >24h. 

Once we have dynamic liquidation we definitely can go longer than 24h again, e.g. 7 days?

model/src/cfd.rs
```diff
@@ -3103,6 +3134,114 @@ mod tests {
     }
     }
 
+    #[test]
+    fn give_order_creation_timestamp_outdated_then_order_outdated() {
+        let creation_timestamp = Timestamp::now();
+        let order = Order::dummy_short().with_creation_timestamp(creation_timestamp);
+
+        let now =
```
bonomat, line 3148: 🙃 This is not `now` anymore :)

model/src/cfd.rs
```diff
@@ -319,7 +324,7 @@ impl Order {
     /// Check the oracle event ID for sanity
     ///
     /// An id within [25h, 23h] from now is considered sane.
-    fn is_oracle_event_id_sane_to_take(&self, now: OffsetDateTime) -> bool {
+    fn is_oracle_settlement_time_sane(&self, now: OffsetDateTime) -> bool {
```
bonomat: Nit: the oracle doesn't know about `settlement` but when it will attest to something. 
Also, it's not about the oracle but about the oracle's event. Maybe better something like: `is_oracle_event_timestamp_sane`

bonomat (approved):
👍

model/src/cfd.rs
luckysori: I suggested this so I'll reply.

I think this is the taker's interpretation of the maker's offer. It reasons about the data in there and knows what certain things mean. One of those things is the `oracle_event_id`, which, in the context of this offer, represents the time at which the oracle will allow the non-collaborative "settlement" of the CFD we are intending to open.

I think your suggestion is more precise, but to understand that function name you need to know more about certain implementation details. At a high level, I think the question we want to answer here is: "does the offer being considered have a satisfactory (sane) time-to-live?". But maybe "oracle settlement time" is not all that clear.

model/src/cfd.rs
da-kami: fair enough, I changed it.

PR #1662, opened 2022-03-22
Bid/ask for closing price instead of average

da-kami:
Instead of using an average price we decide the closing price according to role and position.
This is mostly relevant for the taker, as the taker proposes the closing price. However, since we use the closing price for determining P/L, we have to be able to match the price on the maker side as well for sane values of P/L.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -420,13 +420,10 @@ where
             .context("Price feed not available")?
             .context("No quote available")?;
 
-        let current_price = Price::new(latest_quote.for_taker())?;
         let quote_timestamp = latest_quote
             .timestamp
             .format(&time::format_description::well_known::Rfc3339)
             .context("Failed to format timestamp")?;
-
-        tracing::debug!(%order_id, %current_price, %quote_timestamp, "Proposing settlement of contract");
 
         let threshold = QUOTE_INTERVAL_MINUTES.minutes() * 2;
 
@@ -440,7 +437,9 @@ where
         self.cfd_actor
             .send(taker_cfd::ProposeSettlement {
                 order_id,
-                current_price,
+                bid: Price::new(latest_quote.bid())?,
+                ask: Price::new(latest_quote.ask())?,
+                quote_timestamp,
             })
             .await?
     }
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -10,6 +10,7 @@ use anyhow::Context as _;
 use anyhow::Result;
 use async_trait::async_trait;
 use bdk::bitcoin::secp256k1::schnorrsig;
+use model::closing_price;
 use model::Cfd;
 use model::Identity;
 use model::MakerOffers;
@@ -32,10 +33,12 @@ pub struct TakeOffer {
     pub quantity: Usd,
 }
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 pub struct ProposeSettlement {
     pub order_id: OrderId,
-    pub current_price: Price,
+    pub bid: Price,
+    pub ask: Price,
+    pub quote_timestamp: String,
 }
 
 pub struct Actor<O, W> {
@@ -118,17 +121,30 @@ impl<O, W> Actor<O, W> {
     async fn handle_propose_settlement(&mut self, msg: ProposeSettlement) -> Result<()> {
         let ProposeSettlement {
             order_id,
-            current_price,
+            bid,
+            ask,
+            quote_timestamp,
         } = msg;
+
+        let mut conn = self.db.acquire().await?;
+
+        #[allow(deprecated)]
+        // We need the position and role to decide the price here. Consider re-modelling access to
+        // quote / where to taker the decision to avoid this call.
+        let cfd = crate::command::load_cfd(order_id, &mut conn).await?;
+
+        let proposal_closing_price = closing_price(bid, ask, Role::Taker, cfd.position());
 
         let disconnected = self
             .collab_settlement_actors
             .get_disconnected(order_id)
             .with_context(|| format!("Settlement for order {order_id} is already in progress"))?;
 
+        tracing::debug!(%order_id, %proposal_closing_price, %bid, %ask, %quote_timestamp, "Proposing settlement of contract");
+
         let addr = collab_settlement_taker::Actor::new(
             order_id,
-            current_price,
+            proposal_closing_price,
             self.n_payouts,
             self.conn_actor.clone(),
             self.process_manager_actor.clone(),
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1469,6 +1469,23 @@ pub fn long_and_short_leverage(
     }
 }
 
+/// Decide the closing price according to role and position
+///
+/// Depending on role and position the closing price is determined from bid and ask price of the
+/// maker.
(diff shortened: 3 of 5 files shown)


da-kami:
One note on (backwards) compatibility: Given that the decision logic on the acceptance of the proposed price is not in the daemon, we will have to test that this works in production. As long as there are no changes to other systems, `0.4.7` should be compatible. If we make changes to the decision logic in the automation we will have to make sure compatibility is ensured.

daemon/src/taker_cfd.rs
```diff
@@ -116,10 +118,16 @@ impl<O, W> Actor<O, W> {
...
+        let ProposeSettlement { order_id, bid, ask } = msg;
+
+        let mut conn = self.db.acquire().await?;
+
+        #[allow(deprecated)]
+        // We need the position and role to decide the price here. Consider re-modelling access to
+        // quote / where to taker the decision to avoid this call.
+        let cfd = crate::command::load_cfd(order_id, &mut conn).await?;
+
+        let proposal_closing_price = closing_price(bid, ask, cfd.role(), cfd.position());
```
luckysori: 🔧 You can hard-code the role to `Role::Taker` since this is `taker_cfd.rs`.

daemon/src/lib.rs
```diff
@@ -420,13 +420,12 @@ where
...
             .context("No quote available")?;
 
-        let current_price = Price::new(latest_quote.for_taker())?;
         let quote_timestamp = latest_quote
             .timestamp
             .format(&time::format_description::well_known::Rfc3339)
             .context("Failed to format timestamp")?;
 
-        tracing::debug!(%order_id, %current_price, %quote_timestamp, "Proposing settlement of contract");
+        tracing::debug!(%order_id, %quote_timestamp, "Proposing settlement of contract");
```
bonomat: would be good to keep the price information in there.

daemon/src/taker_cfd.rs
```diff
@@ -116,10 +118,16 @@ impl<O, W> Actor<O, W> {
...
-            current_price,
-        } = msg;
+        let ProposeSettlement { order_id, bid, ask } = msg;
+
+        let mut conn = self.db.acquire().await?;
+
+        #[allow(deprecated)]
+        // We need the position and role to decide the price here. Consider re-modelling access to
+        // quote / where to taker the decision to avoid this call.
+        let cfd = crate::command::load_cfd(order_id, &mut conn).await?;
```
luckysori, line 134: 😞 Another chance for collaborative settlement to access the DB directly. At least it seems like it is the only protocol which doesn't fit into the mold.

luckysori (approved):
LGTM!

model/src/cfd.rs
```diff
@@ -1469,6 +1469,22 @@ pub fn long_and_short_leverage(
...
 
+/// Decide the closing price according to role and position
+///
+/// Depending on role and position the closing price is determined from bid and ask price of the
+/// maker.
+/// Taker is long: When closing the taker sells short to close the position, i.e. the bid price
+/// applies.
+/// Taker is short: When closing the taker buys long to close the position, i.e. the
+/// ask price applies.
+/// For the maker the counter-position's price applies to match the price of the taker.
```
bonomat: I think the last sentence needs some more words. 
We only need this for the projects right? In order to show the same closing price for maker and taker we need to take the counter-position for the maker. 

Is my understanding correct?

daemon/src/taker_cfd.rs
bonomat: I would almost prefer to have two functions for coming up with the closing price. 
- 1. for the projection which needs to know about the position
- 2. for the taker_cfd which doesn't need about the position

We could then enrich the `ProposeSettlement` with what the taker intends to do, e.g. `buy` or `sell` the position. That way we wouldn't need to load from the db here.

daemon/src/taker_cfd.rs
da-kami: I had the same thought. Might be worth exploring some redesign choices to overcome some of these issues. For now I decided to just go forward and not think about refactoring too much 😬

bonomat (approved):
I think the logic is sane. Added an idea which I think could make it a bit nicer.

daemon/src/lib.rs
da-kami: We cannot easily do the decision of bid/ask in here because we don't have access to the database.
I can move the comment to `taker_cfd` and record the proposal price, *or* just record bid and ask here. thoughts?

model/src/cfd.rs
da-kami: That is what I wanted to express and I *think* is is correct :)
I will see to describe it with the sentence you used.

daemon/src/lib.rs
bonomat: I would just print both?

daemon/src/taker_cfd.rs
da-kami: But I think `taker_cfd` *has* to know about the position! The role is implied there, but the position is not. Loading from the db cannot be optimized away (but we can move it into the `collab_settlement_taker` actor if we want).
The projection has to know both, because it cannot imply either.

> We could then enrich the `ProposeSettlement` with what the taker intends to do, e.g. `buy` or `sell` the position. That way we wouldn't need to load from the db here. 

How would you take this decision and where?
Different actions (and routes) would be possible, but that would be a breaking API change.

I am afraid that by doing this in two places we might run into the problem of changing one place (and not the other) and then having different values. It was a conscious decision to put this decision logic into one place.

daemon/src/lib.rs
da-kami: we can, but then we might not know which one was actually used unless we lookup the position of the `Cfd` in the database. If that is OK then I'm happy to add both :)

da-kami: I opted for moving the tracing statement and print all three prices, that should give us the best information.
This adds a bit of indirection, because we pass on the quote timestamp, but I think that is acceptable given that we get a better log statement. 
The log statement is also printed at a more precise point in time, because there were failure scenarios in between the the log statement and the actual settlement start. Now we print right before we actually start the settlement.

daemon/src/taker_cfd.rs
bonomat: You are correct, we can only optimize it away if we enrich `ProposeSettlement` which would involve an API change. Any reason why a breaking API change is bad? We always ship the frontend together with the corresponding HTTP Api.

It is a small function which needs 10+ lines of explanation because it is too complex. So I thought it's worth trying to simplify it. Not worth to debate this much longer. :)

daemon/src/taker_cfd.rs
```diff
@@ -134,6 +140,8 @@ impl<O, W> Actor<O, W> {
             .get_disconnected(order_id)
             .with_context(|| format!("Settlement for order {order_id} is already in progress"))?;
 
+        tracing::debug!(%order_id, %proposal_closing_price, %bid, %ask, %quote_timestamp, "Proposing settlement of contract");
```
bonomat, line 143: Good one :)

bonomat (approved):

daemon/src/taker_cfd.rs
da-kami: > Any reason why a breaking API change is bad? We always ship the frontend together with the corresponding HTTP Api.

I don't think there is anything stopping us from changing the API. Might want to think a bit more before doing this. I might give it a follow-up timebox if I find the time. 😬

PR #1665, opened 2022-03-23
Update the feed after the offer changed

da-kami:
If we don't do that we might run into the problem that the UI does not reflect what actually happened in the daemon.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -199,10 +199,7 @@ where
             .await?;
 
         if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
-            bail!(
-                "The maker's offer appears to be outdated, refusing to take: {:?}",
-                &order_to_take
-            );
+            bail!("The maker's offer appears to be outdated, refusing to take offer",);
         }
 
         tracing::info!("Taking current order: {:?}", &order_to_take);
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -191,6 +191,13 @@ where
         // to take it more than once
         self.current_maker_offers.replace(maker_offers);
 
+        // Cleanup own order feed
+        // Due to the 1:1 relationship between order and cfd we can never create another cfd for the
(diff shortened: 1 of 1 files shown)


daemon/src/taker_cfd.rs
```diff
@@ -200,8 +200,8 @@ where
 
         if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
             bail!(
-                "The maker's offer appears to be outdated, refusing to take: {:?}",
-                &order_to_take
+                "The maker's offer appears to be outdated, refusing to take offer with id {}",
```
bonomat: If the error message is returned to the taker-ui then the order_id is not very useful here.

bonomat (approved):
👍

daemon/src/taker_cfd.rs
da-kami: I think the order-id might be useful for debugging, i.e. if the user runs into the problem often the user might report an order-id back to the maker which might help the maker debug. But I agree, for most users it won't be helpful at all.

I will remove the `order_id`, but note that the maker is not getting any feedback. We could think about a feedback loop back to the maker, where the taker sends a message `SomethingIsFishy`, but that's a different feature. I'll write a ticket for that.

da-kami: https://github.com/itchysats/itchysats/issues/1666

daemon/src/taker_cfd.rs
```diff
@@ -191,6 +191,13 @@ where
         // to take it more than once
         self.current_maker_offers.replace(maker_offers);
 
+        // Cleanup own order feed
+        // Due to the 1:1 relationship between order and cfd we can never create another cfd for the
+        // same order id.
+        self.projection_actor
+            .send(projection::Update(self.current_maker_offers))
+            .await?;
```
luckysori, line 199: Somehow I struggled to understand the commit message together with the change. What you've done is remove the offer taken by the user even if the daemon decides that it cannot go through with taking it. Right? Maybe I can open a follow-up PR that makes these lines a bit clearer.

PR #1667, opened 2022-03-23
Clarify comments around updating offer state after taking one

luckysori:
I think both actions we're taking are related and by grouping them and placing the comment above the block it is clear that they work as a group.

I removed the bit about the 1:1 relationship between order and CFD because I don't know why it matters here.

---

This is just a tiny follow-up to #1665.

### DIFF
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -189,14 +189,12 @@ where
         // The offer we are instructed to take is removed from the
         // set of available offers immediately so that we don't attempt
         // to take it more than once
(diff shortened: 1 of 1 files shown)


da-kami (approved):
Fine with me, any reason for the separate block, or is this just aesthetic?

thomaseizinger (approved):

luckysori:
> Fine with me, any reason for the separate block, or is this just aesthetic?

To make it clearer that the comment is referring to both lines.

I actually dislike how optional scopes like this one look in general, but I think it's harder to misinterpret the comment now.

PR #1674, opened 2022-03-23
Provide some context on why our connection fails

thomaseizinger:
Instead of just sending a `ReadFail` message, we send along an error
that gives us more context on why reading from the socket failed.

Found as part of production hat because the existing log message was fairly poor and didn't help me in diagnosing.

### DIFF
--- a/daemon/src/maker_inc_connections.rs
+++ b/daemon/src/maker_inc_connections.rs
@@ -383,8 +383,9 @@ impl Actor {
     }
 
     async fn handle_read_fail(&mut self, msg: ReadFail) {
-        let taker_id = msg.0;
-        tracing::debug!(%taker_id, "Failed to read incoming messages from taker");
+        let ReadFail { taker_id, error } = msg;
+
+        tracing::debug!(%taker_id, "Failed to read incoming messages from taker: {error:#}");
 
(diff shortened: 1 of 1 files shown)


daemon/src/maker_inc_connections.rs
```diff
@@ -417,26 +418,40 @@ impl Actor {
...
 
         let mut tasks = Tasks::default();
-        tasks.add({
-            let this = this.clone();
-
-            async move {
-                while let Ok(Some(msg)) = read.try_next().await {
-                    let res = this
-                        .send(maker_cfd::FromTaker {
+        tasks.add_fallible(
```
luckysori, line 421: ❓ Is the advantage of using `add_fallible` here that we are more explicit about the error case? Or is there a functional difference?

luckysori (approved):

daemon/src/maker_inc_connections.rs
thomaseizinger: It allows me to use `?` inside the future and handle it all in one spot!

luckysori: Ah, of course! Well worth it.

PR #1679, opened 2022-03-23
Enhance logs a bit

bonomat:
The log message `Rollover V2` or `Rollover V1` is not helpful by itself especially if there are multiple rollovers happening in parallel.

### DIFF
--- a/model/src/rollover.rs
+++ b/model/src/rollover.rs
@@ -20,6 +20,15 @@ pub enum Version {
     V2,
 }
 
+impl std::fmt::Display for Version {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            Version::V1 => write!(f, "V1"),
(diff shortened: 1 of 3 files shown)


luckysori (approved):

model/src/rollover.rs
```diff
@@ -20,6 +20,15 @@ pub enum Version {
     V2,
 }
 
+impl std::fmt::Display for Version {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
```
thomaseizinger, line 24: Nit: In other parts of the codebase we do `use std::fmt;` and refer to these symbols as `fmt::Display` to make things a bit shorter.

thomaseizinger (approved):

model/src/rollover.rs
bonomat: I had it first but then I didn't want to add the extra import for this 🙃

PR #1683, opened 2022-03-24
Remove duplicate rollover received log

thomaseizinger:
This is already logged inside `rollover_maker::Actor`.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -270,8 +270,6 @@ where
         taker_id: Identity,
         version: RolloverVersion,
     ) -> Result<()> {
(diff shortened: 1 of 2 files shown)


luckysori (approved):

bonomat (approved):
👍

bonomat (commented):
Actually, can we remove the other one, because here we have the `taker_id`

thomaseizinger:
> Actually, can we remove the other one, because here we have the `taker_id`

I will add the taker_id to the other one. It makes more sense to have the log in the rollover actor and not in the maker_cfd actor.

bonomat (approved):
thanks :)

PR #1684, opened 2022-03-24
Adjust some logs

luckysori:


model/src/cfd.rs
```diff
@@ -885,6 +885,8 @@ impl Cfd {
             hours_to_charge as i64,
         )?;
 
+        tracing::debug!(%hours_to_charge, funding_fee = %funding_fee.fee, "Accepting rollover proposal");
```
thomaseizinger: This will be a duplicate now with the other one. Shouldn't we drop the other one in favor of this?

bonomat (approved):
👍

luckysori:
@bonomat @thomaseizinger Please have another look since the PR and its scope have changed a bit.

model/src/cfd.rs
```diff
@@ -886,7 +886,7 @@ impl Cfd {
         )?;
 
         tracing::debug!(order_id = %self.id, rollover_version = %version,
-            interval_in_hours = %self.settlement_interval.whole_hours(), "Maker accepting a rollover proposal");
+            interval_in_hours = %self.settlement_interval.whole_hours(), "Accepting rollover proposal");
```
thomaseizinger: Thank you!!! This was bugging me every time I read it because I thought it is emitted by the taker due to the use of "maker" in this message :joy:

model/src/lib.rs
```diff
@@ -728,6 +728,32 @@ impl FundingFee {
         })
     }
 
+    /// Calculate the fee paid or earned for a party in a particular
+    /// position.
+    ///
+    /// A positive sign means that the party in the `position` passed
+    /// as an argument is paying the funding fee; a negative sign
+    /// means that they are earning the funding fee.
+    fn compute_relative(&self, position: Position) -> SignedAmount {
```
thomaseizinger, line 737: Would it make sense to reuse `FeeFlow` here for the return value?

thomaseizinger (approved):

model/src/lib.rs
luckysori: Very possibly! Would you be okay with taking over this PR and making that change if you think it would make things better? Otherwise we can continue with it next week.

bonomat (approved):
Looks good 👍

model/src/lib.rs
thomaseizinger: Yeah there is no rush, we can even merge it now and do it as a follow-up if someone cares enough :D

luckysori: I've been exploring this proposed change a little and I have some doubts.

At the moment this method is only used for logging. My idea was that it could be useful context to know who (role) pays what funding fee during a rollover. `FeeFlow` reflects fee payment between different positions, which is not quite the same. What do you think is more useful? I prefer the current one because it's easier to, for example, say: "the maker is earning 400 satoshi with this rollover which covers 4 hours".

Relatedly, I was considering implementing display on a dedicated type, in order to print something like `earns 400 satoshi` instead of `-400`. But it's just for a minor log, so probably not worth it.

luckysori: I'm gonna go ahead with this merge and we can later on decide if we want to make the change you suggested here.

PR #1686, opened 2022-03-25
Improve projection performance by only applying the event that changed

thomaseizinger:
See commit messages for details. Still needs testing.

### DIFF
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -358,19 +358,19 @@ where
             .send(projection::Update(self.current_offers))
             .await?;
 
-        db::insert_cfd(&cfd, &mut conn).await?;
-        self.projection
-            .send(projection::CfdChanged(cfd.id()))
-            .await?;
-
-        // 4. Try to get the oracle announcement, if that fails we should exit prior to changing any
+        // 3. Try to get the oracle announcement, if that fails we should exit prior to changing any
         // state
         let announcement = self
             .oracle
             .send(oracle::GetAnnouncement(order_to_take.oracle_event_id))
             .await??;
 
-        // 5. Start up contract setup actor
+        db::insert_cfd(&cfd, &mut conn).await?;
+        self.projection
+            .send(projection::CfdChanged(cfd.id()))
+            .await?;
+
+        // 4. Start up contract setup actor
         let addr = setup_maker::Actor::new(
             self.db.clone(),
             self.process_manager.clone(),
--- a/daemon/src/maker_cfd.rs
+++ b/daemon/src/maker_cfd.rs
@@ -366,9 +366,7 @@ where
             .await??;
 
         db::insert_cfd(&cfd, &mut conn).await?;
-        self.projection
-            .send(projection::CfdChanged(cfd.id()))
-            .await?;
+        self.projection.send(projection::NewCfd(cfd.id())).await?;
 
         // 4. Start up contract setup actor
         let addr = setup_maker::Actor::new(
--- a/daemon/src/process_manager.rs
+++ b/daemon/src/process_manager.rs
@@ -19,7 +19,7 @@ use xtras::SendAsyncSafe;
 pub struct Actor {
     db: sqlx::SqlitePool,
     role: Role,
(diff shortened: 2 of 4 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -358,19 +358,19 @@ where
...
             .await?;
 
-        db::insert_cfd(&cfd, &mut conn).await?;
-        self.projection
-            .send(projection::CfdChanged(cfd.id()))
-            .await?;
-
-        // 4. Try to get the oracle announcement, if that fails we should exit prior to changing any
+        // 3. Try to get the oracle announcement, if that fails we should exit prior to changing any
         // state
```
luckysori, line 362: 😆 It's nice when the code contradicts the comment.

daemon/src/process_manager.rs
```diff
@@ -72,7 +72,7 @@ impl Actor {
 
         // 2. Post process event
         use EventKind::*;
-        match event.event {
+        match &event.event {
```
luckysori, line 75: ❓ Why is all the subsequent cloning preferable over matching by value here?

luckysori (approved):
LGTM!

daemon/src/process_manager.rs
thomaseizinger: - I needed the full value to dispatch it to projection.
- If I clone the entire event here I also clone all of the CETs which I never need by value in the match.

thomaseizinger:
What is in principle "dangerous" about this solution is that we assume that events will be sent to `projection` in the order they appeared. Given that we only have one instance of projection and one instance of process manager, this assumption should always hold.

It may still be better to apply a more well studied optimisation technique:

- Load all events initially + remember the current version of the CFD (version being the # of events applied)
- Load all events from current version to now later down the line

da-kami (approved):
As long as we keep the `NewCfdEvent` tied to only one place, which is the process manager, this should be fine.
I am bit afraid that this solution might lead to very weird behavior if we accidentially "move code around" . The constraints around event order and when to send updates are implicitly there and I am afraid that we might oversee this in the future. 

That said, I wonder it it would not be better to invest into optimizing loading from the DB using snapshots and generally optimizing the storage before doing this optimization.

daemon/src/projection.rs
```diff
@@ -855,6 +858,21 @@ impl Actor {
             .send_cfds_update(self.state.cfds.clone(), self.state.quote);
     }
 
+    /// Updates the projection state with the new CFD.
+    ///
+    /// In theory, we could send freshly minted CFD over but it is a different Rust struct and
+    /// mapping it is tedious. Loading a new CFD from the DB is cheap so we just do that
+    /// instead.
+    async fn handle(&mut self, msg: NewCfd) {
```
bonomat, line 883: Maybe we should do it nevertheless? 
Loading a cfd means executing 2 queries sequentially, which also involves loading and applying the events which is not "that" cheap imho. 
Mapping it in code should give us an additional performance boost.

bonomat (approved):
Nice. Maybe mapping in code is still a good idea. Can also be done in a follow-up though.

daemon/src/projection.rs
thomaseizinger: But a new CFD has 0 events! That is what made me in the end go with re-loading because it is essentially just a query to the `cfds` table.

thomaseizinger:
> That said, I wonder it it would not be better to invest into optimizing loading from the DB using snapshots and generally optimizing the storage before doing this optimization.

Those are different optimisations though!

An alternative to this one would be an in-memory read-model where we remember which event we applied last on only load new events upon a `CfdChanged` event. That is what I meant here: https://github.com/itchysats/itchysats/pull/1686#issuecomment-1082492467

thomaseizinger:
I am a bit worried that the change I am proposing here relies on too many assumptions that are not stated very explicitly and are hard to enforce through the compiler which makes it possible that they break further down the line.

thomaseizinger:
Replaced by https://github.com/itchysats/itchysats/pull/1722.

PR #1687, opened 2022-03-25
Reuse reqwest::Client in oracle::Actor

thomaseizinger:
Analysis with flamegraphs show that we spend > 1% of our CPU time with
re-constructing `reqwest::Client`s. There is no need to do this, we can
reuse the same client for multiple requests.

The client uses an `Arc` internally so it is cheap to clone.

Based on my flamegraphing, this reduces the relative CPU time spent on the `ensure_having_announcements` function from **1.61%** to **0.65%**.

[flamegraphs.zip](https://github.com/itchysats/itchysats/files/8347794/flamegraphs.zip)

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -27,6 +27,7 @@ pub struct Actor {
     announcement_lookahead: Duration,
     tasks: Tasks,
     db: SqlitePool,
+    client: reqwest::Client,
 }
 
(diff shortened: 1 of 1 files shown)


luckysori (approved):
It's a great discovery that reconstructing a `reqwest::Client` is relatively expensive. Because I think we have done it a lot in the past.

klochowicz (approved):

Issue #320, opened 2021-10-14
Enable the taker to publish the collaborative settlement tx


da-kami:
We had several discussion about that one in the past, just had another one with @klochowicz  :)

Here is our reasoning why it is more resilient if the maker answers with the signed transaction instead of just telling the taker "accepted, I'll publish it for you..." (which is what we currently do):

The time-component for knowing that it will work out is shorter and less complex if the taker is enabled to publish the transaction as well. 
Currently we have to monitor and then eventually decide to publish commit if we don't see the transactions. 

If the taker's collaborative settlement proposal is answered by the maker with sending the sigend transaction back (and they both try to broadcast) the taker has immediate assurance after the response. Meaning, if the maker does not respond within x time the taker does not have to monitor anything, the taker just publishes commit. 
If the maker responds the taker can ensure that the tx goes in (actually both have to ensure because both have it). That's the same reasoning as for the contract setup. 

## Annoyance counter

- 23/03/2022: The protocol being asymmetric caused https://github.com/itchysats/itchysats/issues/1668 because we don't have a timeout on executing the protocol.
- 24/03/2022: The above problem appeared in production.
- 26/03/2022: The timeout introduced in #1671 appears to be to short which caused a scenarios where the maker timed out while waiting for the `Initiate` message, causing yet another instance of the maker force closing a position while the taker thinks it is being collaboratively settled. 
- 05/04/2022: While working on https://github.com/itchysats/itchysats/pull/1765, @da-kami noticed that without changing this protocol, we cannot republish the collab. settlement transaction.

da-kami:
As part of this ticket we will have to adapt the `monitor` to try re-broadcasting the settlement transaction upon startup. Currently that is not done, which can result in the transaction not being re-broadcasted in case an initial broadcast fails and we restart.

Currently we cannot just re-broadcast the settlement transaction, because the taker's transaction is not fully signed (which should be fixed by this ticket).

thomaseizinger:
As part of looking into https://github.com/itchysats/itchysats/issues/1595, I've thought about how to re-design this protocol. Here are my findings:

1. Collaborative close is _not_ a [2-generals problem](https://en.wikipedia.org/wiki/Two_Generals%27_Problem) because even with the protocol failing on one end, the other party can still proceed.
2. In general, sending messages concurrently (like we do in rollover and contract setup) is a problem. If there is a defined sequence of sending and receiving, receiving a message on A is an implicit ACK from B that the last message was received.

[...]

PR #1691, opened 2022-03-26
Allow 5 minutes for the taker to initiate collaborative settlement

da-kami:
In production, we saw scenarios where the timeout was triggered after 60s, but the taker never triggered a rollover again.
This could mean that the taker did send `Initiate` with the signature, but it did not arrive at the maker's collab settlement actor in time before the timeout was triggered, resulting in the taker to believe the settlement was "finished" (i.e. handled bt the maker), but the maker actually timed out.
As a quick fix this allows for 5 minutes to receive `Initiate` on the maker side instead of 1 minute.

### DIFF
--- a/daemon/src/collab_settlement_maker.rs
+++ b/daemon/src/collab_settlement_maker.rs
@@ -22,7 +22,7 @@ use xtras::address_map::IPromiseIamReturningStopAllFromStopping;
 /// If the taker does not come back with `Initiate` until the timeout is triggered we fail the
 /// settlement. If the taker does come back with `Initiate` before the timeout is reached, we don't
 /// fail the settlement even if the timeout is triggered.
-const INITIATE_TIMEOUT: Duration = Duration::from_secs(60);
+const INITIATE_TIMEOUT: Duration = Duration::from_secs(60 * 5);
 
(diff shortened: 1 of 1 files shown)


da-kami:
Quickfix until we action https://github.com/itchysats/itchysats/issues/320

bonomat (approved):
The timeout is fine with me.

thomaseizinger (approved):
It would be really good to validate such kind of theories. I am happy to make this change but I also have a feeling that there is something else that is not working. 60 seconds should have been plenty.

da-kami:
> It would be really good to validate such kind of theories. I am happy to make this change but I also have a feeling that there is something else that is not working. 60 seconds should have been plenty.

I totally agree with this statement overall, we can think about more feedback loops between taker and maker.

In the current situation we cannot easily verify without the taker coming back to us with information. I think the best way forward is to implement #320 - that should make the problem of the taker trusting the maker to publish go away altogether.

The current problem is, that the taker transitions to a state that is treated as "final" trusting the maker. This will go away if both try to publish the transaction. The taker will then never transition to `Pending Close` unless the taker has the fully signed transaction.

Issue #1692, opened 2022-03-27
Funding rate might confuse people


bonomat:
The way how funding rate is depicted at the moment might be confusing. 
I know that a positive funding rate means, long pays short but this is not really visible nor described anywhere. 
Right now, it looks like that both, longs and shorts will need to pay this rate: 

![image](https://user-images.githubusercontent.com/224613/160261374-e674a076-a49a-485a-a049-e149e22c0993.png)

![image](https://user-images.githubusercontent.com/224613/160261378-f118b0a8-89fc-4d11-be90-f19b85368a6a.png)

klochowicz:
isn't this the case in the above pictures that both long and short need to pay, ie. it's not a negative funding rate? (just double-checking as the rates are quite different in long and short tab)

bonomat:
> isn't this the case in the above pictures that both long and short need to pay, ie. it's not a negative funding rate? (just double-checking as the rates are quite different in long and short tab)

mhm... good that it even confuses us 😅 Imho it means: positive means long pay, negative means short pay. So in this case, when shorting, and it says 0.00026, it means that the shorts would receive this amount.

thomaseizinger:
It appears to me that we should fix this in the daemon and return fields like:

```json
{
	"you_pay_hourly": "0.000026%"
}
```
and

```json
{
	"you_receive_hourly": "0.000026%"
}
```

Actual names subject to bike-shedding. The main point being: The daemon knows, which role it is in and thus, could map this directly to an unambiguous representation. I would suggest to have a think about what we want to display in the UI and then make the API resemble this as closely as possible.

PR #1693, opened 2022-03-27
Add FundingRate tooltip to trade table

bonomat:
image

### DIFF
--- a/taker-frontend/src/components/FundingRateTooltip.tsx
+++ b/taker-frontend/src/components/FundingRateTooltip.tsx
@@ -1,0 +1,23 @@
+import { Tooltip } from "@chakra-ui/react";
+import * as React from "react";
+import { PropsWithChildren } from "react";
+
+type FundingRateTooltipProps = PropsWithChildren<{
+    fundingRateHourly?: number;
+    fundingRateAnnualized?: number;
+    disabled: boolean;
+}>;
+
+export function FundingRateTooltip(
+    { children, fundingRateHourly, fundingRateAnnualized, disabled }: FundingRateTooltipProps,
+) {
+    return (<Tooltip
+        label={`The CFD is rolled over perpetually every hour at ${fundingRateHourly}% (${fundingRateAnnualized}% p.a.). 
+        The cost can fluctuate depending on the market movements.`}
+        hasArrow
+        placement={"right"}
+        isDisabled={disabled}
+    >
+        {children}
+    </Tooltip>);
+}
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -27,6 +27,7 @@ import {
     Tbody,
     Td,
     Text,
+    Tooltip,
     Tr,
     useColorModeValue,
     useDisclosure,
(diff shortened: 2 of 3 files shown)


bonomat:
mhmm.. dprint has an issue, looks like there was a version upgrade but the action was not pulled through.

taker-frontend/src/components/Trade.tsx
```diff
@@ -221,7 +221,7 @@ export default function Trade({
                                     </Td>
                                 </Tr>
                                 <Tr>
-                                    <Td>Funding Rate</Td>
+                                    <Td>Perpetual Cost</Td>
                                     <Td isNumeric>{fundingRateHourly ? fundingRateHourly + "%" : "not available"}</Td>
```
thomaseizinger: This number will flip its sign depending on who gets paid right? Is "cost" the best word in both cases?

Probably subject to a larger discussion, I am okay with moving forward to align it for now and then change it later in multiple places.

taker-frontend/src/components/Trade.tsx
```diff
@@ -221,8 +222,20 @@ export default function Trade({
                                     </Td>
                                 </Tr>
                                 <Tr>
-                                    <Td>Perpetual Cost</Td>
-                                    <Td isNumeric>{fundingRateHourly ? fundingRateHourly + "%" : "not available"}</Td>
+                                    <Td><Text>Perpetual Cost</Text></Td>
+                                    <Tooltip
+                                        label={`The CFD is rolled over perpetually every hour at ${fundingRateHourly}%, annualized that is ${fundingRateAnnualized}%. 
+                                        The funding rate can fluctuate depending on the market movements.`}
```
thomaseizinger:
```suggestion
These costs fluctuate depending on the market movements.
```

Otherwise this is the first mention of "funding rate".

taker-frontend/src/components/Trade.tsx
```diff
@@ -221,8 +222,20 @@ export default function Trade({
...
+                                        hasArrow
+                                        placement={"right"}
+                                        isDisabled={!fundingRateHourly}
+                                    >
+                                        <Td isNumeric>
+                                            {fundingRateHourly
+                                                ? "Hourly @ " + fundingRateHourly + "%"
+                                                : "not available"}
```
thomaseizinger: Seen this pattern a couple of times. You could either use a `Skeleton` here or make your own component that displays "not available" if the value is undefined.

taker-frontend/src/components/FundingRateTooltip.tsx
```diff
@@ -0,0 +1,23 @@
+import { Tooltip } from "@chakra-ui/react";
+import * as React from "react";
+
+interface FundingRateTooltipProps {
+    children: JSX.Element;
```
thomaseizinger: You can achieve this with composition of an existing type: `PropsWithChildren`.

See https://www.mendelowski.com/docs/react/children-props-with-typescript/.

thomaseizinger (approved):
LGTM

Some minor feedback.

taker-frontend/src/components/FundingRateTooltip.tsx
bonomat: cool, I didn't know about this. 👍

taker-frontend/src/components/Trade.tsx
bonomat: I think this should be cleaned up together in #1692

taker-frontend/src/components/Trade.tsx
bonomat: good idea. Changed to using a `skeleton`

taker-frontend/src/components/FundingRateTooltip.tsx
bonomat: works nicely. Need to remember this :)

klochowicz: thanks for sharing this tip, looks really useful! 👍

klochowicz (approved):
nice!

thomaseizinger (approved):

Issue #1701, opened 2022-03-28
Don't reuse `ContractSetup` for when rollingover


bonomat:
We sometimes need to know if a CFD is open or not open. Re-using the state `ContractSetup` for when rolling over is active has weird sideeffects. 

Right now we use `ContractSetup` for the time when the contract is being setup and when it is rolledover. 

We should have a separate state here. 

https://github.com/itchysats/itchysats/blob/5ccb0a6cc6269200eb2652e57ebeaf511abe2473/daemon/src/projection.rs#L391-L393

and

https://github.com/bonomat/hermes/blob/5ccb0a6cc6269200eb2652e57ebeaf511abe2473/daemon/src/projection.rs#L268-L276

thomaseizinger:
I am tempted to close this as a duplicate of https://github.com/itchysats/itchysats/issues/1314. IMO, we should not be exposing these states and instead tell the client whether or not the CFD is open or closed with a boolean flag.

bonomat:
> I am tempted to close this as a duplicate of #1314. IMO, we should not be exposing these states and instead tell the client whether or not the CFD is open or closed with a boolean flag.

We should introduce a new state for rollover while we have states. It's just wrong to use `ContractSetup` here. 
If we go away from states, then this is not needed anymore. 
But given this is a small change, I'd be for adding the new state now.

Issue #1314, opened 2022-02-02
Cleanup the Cfd states in the projection to reduce decision logic in the UI


da-kami:
We keep adding decision logic to the UI based on "old" states that are constructed in the projection. 
We should rather refactor the projection API to only expose Cfd states that are actually relevant for the UI to display.

## Annoyance counter

- 24/03/2022: Having to reason about these states in every client leads to bug where two clients can disagree on when a CFD is "closed". In particular, a CFD with the state `CollaborativeSettlementAccepted` is _not_ yet closed because we don't have the taker's signature yet.
- 29/03/2022: This topic came up again in a different context: https://github.com/itchysats/itchysats/issues/1701
- 29/03/2022: Incrementing counter to account for this duplicate: https://github.com/itchysats/itchysats/issues/960

thomaseizinger:
I know @klochowicz has some opinions on this so please chim in with your view!

When re-designing this we should consider the following:

1. Automated clients care about the CFD's state _to some degree_. Primarily they care about whether or not the CFD is "open" or "closed". I put it in quotes because we haven't unambiguously defined what those terms mean. I'll have a go at that here:

A CFD is closed if the payout ratio has been determined and will never change again. In blockchain terms, this is equivalent to having a fully-signed and ready-to-be-broadcast transaction that spends the joint-output into two individual ones, one owned by each party. In any other case, the CFD is considered open.

I think the best way of representing this is with a boolean flag in the response: `is_closed`.

2. User-facing clients may also care about the CFD's state. However, this is has historically been a point of debate. In an ideal world, I think they shouldn't. Here is why:

- Assuming a "combined" position, a CFD is open if you are either long or short with some number of contracts. In blockchain terms, this means you have an unspent DLC somewhere on the chain. This is in-line with being exactly the inverse of the above definition.
- A user can in principle perform two operations: buy more contracts or sell some of their contracts. A user can always do that, even if they don't currently have a position. Not having any contracts and selling some = going short. Selling all your contracts = closing your position, etc. In this mental model, there is no notion of "state". All actions are available all the time.
[...]

PR #1702, opened 2022-03-28
Don't treat rollover accepted as contract setup

bonomat:
The issue with this is that we need to distinguish between an open / not open CFD. It is not good that the same state is being used for while a CFD is being openend and while it is rolled over.

quick fix for #1701. A better solution is needed though. 

Please check closely if this can have some side effects I'm not aware of

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -271,7 +271,12 @@ impl Aggregated {
                     Role::Maker => CfdState::IncomingRolloverProposal,
                     Role::Taker => CfdState::OutgoingRolloverProposal,
                 },
-                ProtocolNegotiationState::Accepted => CfdState::ContractSetup,
+                // TODO: we should have an own state here indicating that the position is being
+                // rolled over.
+                ProtocolNegotiationState::Accepted => match role {
+                    Role::Maker => CfdState::IncomingRolloverProposal,
+                    Role::Taker => CfdState::OutgoingRolloverProposal,
+                },
             };
         };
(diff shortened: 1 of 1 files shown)


daemon/src/projection.rs
```diff
@@ -271,7 +271,12 @@ impl Aggregated {
                     Role::Maker => CfdState::IncomingRolloverProposal,
                     Role::Taker => CfdState::OutgoingRolloverProposal,
                 },
-                ProtocolNegotiationState::Accepted => CfdState::ContractSetup,
+                // TODO: we should have an own state here indicating that the position is being
+                // rolled over.
+                ProtocolNegotiationState::Accepted => match role {
+                    Role::Maker => CfdState::IncomingRolloverProposal,
+                    Role::Taker => CfdState::OutgoingRolloverProposal,
```
thomaseizinger, line 278: This will have the side effect that we emit the actions again on the maker. We cache emitted actions in the automation so this should be safe for now but can become very brittle very soon. Please amend the commit message to explain why this is ok!

I'd prefer if we can go into the direction of https://github.com/itchysats/itchysats/issues/1314.

thomaseizinger (approved):
ACK modulo an amended commit message that explains this in more detail.

bonomat:
@thomaseizinger : this has been solved differently using the new API endpoint right?

luckysori:
Superseded by https://github.com/itchysats/itchysats/pull/1780.

PR #1715, opened 2022-03-29
Add `/open_cfds` endpoint

### LINKED ISSUE #1711: Introduce new `open_cfds` endpoint that only returns CFDs which are considered "open"
`is_open` means we have a spendable multi-sig on the blockchain or have a transaction to create such joint output.

thomaseizinger:
We define "open" as "has a confirmed, unspent joint output". Our
definition of "confirmed" for these purposes is "having a transaction"
that can create such an output. This is the case after contract-setup
has finished but before we try to broadcast the transaction. Failing to
broadcast the transaction is extremely unlikely because our software is
in control of the wallet and should in normal operation prevent double-
spending.

If we ever extend our monitoring - and in a further step the event model
to account for transactions in mempool - we can potentially redefine
this as "transaction must have been successfully broadcasted".

## Note to reviewers

I set the boolean only where I deemed it necessary. Please pay close attention to whether or not it is set in all places correctly.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -510,7 +510,8 @@ impl Cfd {
 
                 self.aggregated.state = CfdState::PendingCommit;
 
-                // We argue that a decrypted CET being available will eventually lead to the joint output being spent.
+                // We argue that a decrypted CET being available will eventually lead to the joint
+                // output being spent.
                 self.aggregated.has_unspent_joint_output = false;
             }
             OracleAttestedPostCetTimelock { cet, price, .. } => {
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -509,6 +509,9 @@ impl Cfd {
                 self.closing_price = Some(price);
 
                 self.aggregated.state = CfdState::PendingCommit;
+
+                // We argue that a decrypted CET being available will eventually lead to the joint output being spent.
+                self.aggregated.has_unspent_joint_output = false;
             }
             OracleAttestedPostCetTimelock { cet, price, .. } => {
                 self.aggregated.cet = Some(cet);
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -221,6 +221,12 @@ struct Aggregated {
 
     version: u32,
     creation_timestamp: Timestamp,
+
+    /// True if we have an unspent joint output on the blockchain or are about to create one (i.e.
+    /// after contract setup).
+    ///
+    /// This is sometimes referred to as a CFD being "open".
+    has_unspent_joint_output: bool,
 }
 
 impl Aggregated {
@@ -239,6 +245,7 @@ impl Aggregated {
             settlement_state: None,
             version: 0,
             creation_timestamp: Timestamp::now(),
+            has_unspent_joint_output: false,
         }
     }
 
@@ -402,6 +409,7 @@ impl Cfd {
                 self.aggregated.latest_dlc = Some(dlc);
 
                 self.aggregated.state = CfdState::PendingOpen;
+                self.aggregated.has_unspent_joint_output = true;
             }
             ContractSetupFailed => {
                 self.aggregated.state = CfdState::SetupFailed;
@@ -445,6 +453,7 @@ impl Cfd {
                 self.closing_price = Some(price);
 
                 self.aggregated.state = CfdState::PendingClose;
+                self.aggregated.has_unspent_joint_output = false;
             }
             CollaborativeSettlementRejected => {
                 self.aggregated.settlement_state = None;
@@ -466,12 +475,15 @@ impl Cfd {
             }
             CetConfirmed => {
                 self.aggregated.state = CfdState::Closed;
+                self.aggregated.has_unspent_joint_output = false;
             }
             RefundConfirmed => {
                 self.aggregated.state = CfdState::Refunded;
+                self.aggregated.has_unspent_joint_output = false;
             }
             LockConfirmedAfterFinality | CollaborativeSettlementConfirmed => {
                 self.aggregated.state = CfdState::Closed;
+                self.aggregated.has_unspent_joint_output = false;
             }
             CetTimelockExpiredPriorOracleAttestation => {
                 self.aggregated.state = CfdState::OpenCommitted;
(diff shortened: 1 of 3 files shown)


daemon/src/projection.rs
```diff
@@ -439,6 +446,7 @@ impl Cfd {
                 self.closing_price = Some(price);
 
                 self.aggregated.state = CfdState::PendingClose;
+                self.aggregated.has_unspent_joint_output = false;
```
luckysori, line 456: ❓ I would assume that the `ContractSetupFailed` state is only reachable from `ContractSetupStarted`, so setting this to `false` here wouldn't be strictly necessary. Is this change deliberate so that we cover our bases in case we can go "backwards" after `ContractSetupCompleted`?

luckysori (approved):
LGTM! It's not immediately obvious what the state of `has_unspent_joint_output` should be after each event, but I think I agree with your choices.

daemon/src/projection.rs
thomaseizinger: You are commenting on the `CollaborativeSettlementCompleted`. The `ContractSetupFailed` event does not set the `has_unspent_joint_output` variable.

thomaseizinger: GitHub is a bit mean here because it collapsed everything in-between :)

luckysori: Okay, that makes way more sense. Thanks for clarifying!

thomaseizinger:
Needs to be rebased on top of https://github.com/itchysats/itchysats/pull/1717 to incorporate the new failure mode.

da-kami (changes requested):
I think all "normal" cases are covered.

Only `ContractSetupCompleted` sets to true.
*Any* event that decides the closing price (or is after the price was locked in) sets to `false`.

One weird outsider is `RevokeConfirmed` - I am not sure about that one. If we find it on chain I think we should still set it to `false` for now because we won't be able to find the `attestation` / `CET`...? Obviously the price is not locked in for that one, but it is still problematic if we keep it around, because we would consider it as `Open` forever given the current logic, or am I missing something? 😬
I think it's safer to set it to `false` in that case and add a comment to change it once handling for `RevokeConfirmed` is implemented.

klochowicz:
> Needs to be rebased on top of #1717 to incorporate the new failure mode.

rebased (applied a separate commit on top of master, so it's easier for everyone to look at the diff) - this will get squashed.

da-kami (approved):
thank you

daemon/src/projection.rs
```diff
@@ -480,11 +492,13 @@ impl Cfd {
...
             }
             RefundTimelockExpired { .. } => {
                 self.aggregated.refund_published = true;
 
                 self.aggregated.state = CfdState::PendingRefund;
+                self.aggregated.has_unspent_joint_output = false;
             }
             OracleAttestedPriorCetTimelock {
                 timelocked_cet,
```
bonomat, line 504: 📯
Is it correct that we can achieve this if we publish the commit transaction too late and the CET timelock has not expired yet when the oracle attests to the price? 

I think this needs to be false as well. If the oracle attests to the price we know already the closing price.

bonomat (changes requested):
Thanks. I think the `OracleAttestedPriorCetTimelock` should also set unpsnet_joint_output to false

daemon/src/projection.rs
thomaseizinger: Hmm, interesting!

I guess we can say that having a valid but _timelocked_ transaction that can spend the joint output should also already account for the output being spent (which is what the boolean represents).

thomaseizinger: > I think this needs to be false as well. If the oracle attests to the price we know already the closing price.

By itself, that reasoning does not make sense though. This boolean tracks whether we have an unspent join output, not whether we know the closing price. I think this comment highlights that we are introducing an unfit abstraction here.

da-kami:
Thanks for the fix, I think we can merge this!

thomaseizinger:
> Thanks for the fix, I think we can merge this!

Yeah. It think it should be fine although the more I think about this, the less I like the abstraction we are introducing. It is such a mess :(

da-kami:
> > Thanks for the fix, I think we can merge this!
> 
> Yeah. It think it should be fine although the more I think about this, the less I like the abstraction we are introducing. It is such a mess :(

Because the Boolean is not representative enough? 
We can introduce more "states" - this goes back to the discussion what this endpoint should be used for. "Open" is an ambiguous state, for the automation it matters if we have the closing price or not to determine "open". If we agree that this endpoint is for the automation only the current representation is sufficient.

thomaseizinger:
> > > Thanks for the fix, I think we can merge this!
> > 
> > Yeah. It think it should be fine although the more I think about this, the less I like the abstraction we are introducing. It is such a mess :(
> 
> Because the Boolean is not representative enough? 
> We can introduce more "states" - this goes back to the discussion what this endpoint should be used for. "Open" is an ambiguous state, for the automation it matters if we have the closing price or not to determine "open". If we agree that this endpoint is for the automation only the current representation is sufficient.

You are missing the point. It is not about more or less states. Also, open is not ambiguous. Unless we operate on randomness, there is no ambiguity here!

The problem is that we are defining an abstraction (open) based on needs (automation) that the current software (maker) is not and cannot be aware of.

If the maker should exist as an independent piece of software, the abstractions that are provided should make sense in isolation, otherwise they are bad abstractions.

In its current state, there is no decision logic in the maker. Consequently, any abstraction that is introduced needs to make sense in isolation without knowledge of what the APIs are used for.

The only reasonable abstraction I could come up with is to define "open" around the blockchain state of the CFD: whether the joint output is spent or not. This has nothing to do with the price.

The fact that we need to slightly skew this abstraction and consider "joint output will likely be spent in a few blocks" as "has no joint output and is thus not open anymore" is what I call a mess.

I suggest to read https://github.com/itchysats/architecture/pull/25 for some more context!

bonomat (approved):
thanks

da-kami:
> [snip]
> You are missing the point. It is not about more or less states. Also, open is not ambiguous. Unless we operate on randomness, there is no ambiguity here!
> 
> The problem is that we are defining an abstraction (open) based on needs (automation) that the current software (maker) is not and cannot be aware of.

For me it was not the point of this PR to invest time into re-designing the API, but to fix a problem where the automation reacts to a state change that should not be (the scenario where a contract setup fails, we treat `ContractSetup` as open in the automation because rollover uses the same state...).

What are you aiming for with this PR? Fixing this problem? Re-design?

> [snip]
> I suggest to read [itchysats/architecture#25](https://github.com/itchysats/architecture/pull/25) for some more context!

I read through it, will react there. I am not sure we have a decision there yet and it will take time to implement this. Given what we currently have it would be great to land some "messy" fix in the meanwhile. That does not have to be an `open` endpoint, we could still opt for using a different state when rolling over.

luckysori:
Let's close this PR for the time being, since we already introduced a quick fix with https://github.com/itchysats/itchysats/pull/1780.

PR #1717, opened 2022-03-29
Don't make clients think our DB is empty

### LINKED ISSUE #1716: Don't send empty list of CFDs to clients whilst we are initializing
Loading all CFDs from the DB can take quite a while, depending on how big the database is. Optimizations such as https://github.com/itchysats/itchysats/issues/1651 are meant to improve this but it can still be misleading to clients to present them an empty list during startup.

Instead of sending an empty list, we should tell the client to come back later once we have fully started up.

thomaseizinger:
Triggered by discussion: https://github.com/itchysats/maker-automation-rs/pull/124#discussion_r837068347

During startup, it can take a while to load the CFDs from the database.
Instead of lying to our clients during this time and presenting them an
empty list, we:

a) Don't emit the list on the feed
b) We return a 500 from the `/cfds` endpoint

There is nothing the client can do about this, hence it is not a 400.
Once the CFDs are loaded, the request will be fulfilled properly.

Fixes #1716.

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -65,7 +65,9 @@ pub async fn maker_feed(
         yield quote.to_sse_event();
 
         let cfds = rx_cfds.borrow().clone();
-        yield cfds.to_sse_event();
+        if let Some(cfds) = cfds {
+            yield cfds.to_sse_event()
+        }
 
         let takers = rx_connected_takers.borrow().clone();
         yield takers.to_sse_event();
@@ -87,7 +89,9 @@ pub async fn maker_feed(
                 }
                 Ok(()) = rx_cfds.changed() => {
                     let cfds = rx_cfds.borrow().clone();
-                    yield cfds.to_sse_event();
+                    if let Some(cfds) = cfds {
+                        yield cfds.to_sse_event()
+                    }
                 }
                 Ok(()) = rx_quote.changed() => {
                     let quote = rx_quote.borrow().clone();
@@ -300,7 +304,12 @@ pub async fn get_cfds<'r>(
     let rx_cfds = rx.cfds.clone();
     let cfds = rx_cfds.borrow().clone();
 
-    Ok(Json(cfds))
+    match cfds {
+        Some(cfds) => Ok(Json(cfds)),
+        None => Err(HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+            .title("CFDs not yet available")
+            .detail("CFDs are still being loaded from the database. Please retry later.")),
+    }
 }
 
 #[rocket::get("/takers")]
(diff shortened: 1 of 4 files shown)


maker/src/routes.rs
```diff
@@ -300,7 +304,12 @@ pub async fn get_cfds<'r>(
     let rx_cfds = rx.cfds.clone();
     let cfds = rx_cfds.borrow().clone();
 
-    Ok(Json(cfds))
+    match cfds {
+        Some(cfds) => Ok(Json(cfds)),
+        None => Err(HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
```
bonomat, line 309: Maybe `StatusCode::SERVICE_UNAVAILABLE` fits better here? 

https://www.rfc-editor.org/rfc/rfc7231#section-6.6.4

bonomat (approved):
Good idea with sending the event only if we are fully loaded

maker/src/routes.rs
thomaseizinger: We could yeah! That would allow us to add a standards-compliant `Retry-After` header to indiciate, when the client should try again.

I initially opted not to because I didn't want to get hung up on using specific status codes as long as it is the right category (500 instead of 400 because it is not a user-error). However, seeing the `Retry-After` thing now, this may be a good idea although we would also have to make use of it in our clients.

Sending 500 is not wrong, just not as specific as we could be.

luckysori (approved):
I like this approach. It does seem like we introduce a lot more error paths after adding the `None` case, but this is more correct.

maker/src/routes.rs
klochowicz: I like the idea of returning a code, as well as the suggestion of returning `Retry-After! :)

klochowicz (approved):
thank you!

maker/src/routes.rs
thomaseizinger: If you guys don't mind, I would propose doing it in a follow-up. Things to be considered are:

- What should the value of `Retry-After` be?
- How do we see this being used in clients? Most of our clients fetch this on an interval anyway. Would we expect them to honor the value in `Retry-After` and dynamically offset the retry interval?

I like the idea in principal but I am just not sure if it is worth it right now.

thomaseizinger:
> I like this approach. It does seem like we introduce a lot more error paths after adding the `None` case, but this is more correct.

There is likely some potential for making this cleaner! To be honest, I did not bother much because I prioritized the correctness of the outcome over the implementation :)

maker/src/routes.rs
bonomat: > I like the idea in principal but I am just not sure if it is worth it right now.

no need to do this now. Also, we would need to implement that logic in the client to make use if it ;)

PR #1721, opened 2022-03-30
Refactor `model::closing_price` function for clarity 

luckysori:
The diff should show that there are no functional changes, but I've adapted the name of the function, the docstring and even the body to aid understandability.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1523,20 +1523,26 @@ pub fn long_and_short_leverage(
     }
 }
 
-/// Decide the closing price according to role and position
+/// Calculate the closing price used to collaboratively settle a CFD.
+/// This value is akin to the one used for a market close order in a
+/// centralised exchange.
 ///
-/// Depending on role and position the closing price is determined from bid and ask price of the
-/// maker.
-/// Taker is long: When closing the taker sells short to close the position, i.e. the bid price
-/// applies.
-/// Taker is short: When closing the taker buys long to close the position, i.e. the
-/// ask price applies.
-/// In order to show the same closing price for maker and taker we need to take the counter-position
-/// for the maker.
-pub fn closing_price(bid: Price, ask: Price, role: Role, position: Position) -> Price {
-    match (role, position) {
-        (Role::Taker, Position::Long) | (Role::Maker, Position::Short) => bid,
-        (Role::Taker, Position::Short) | (Role::Maker, Position::Long) => ask,
+/// We calculate it from the perspective of the maker (i.e. the market).
+///
+/// If the maker has gone long, when closing their position they sell
+/// short. Therefore, in that case we use the ask price.
+///
+/// If the maker has gone short, when closing their position they buy
+/// long. Therefore, in that case we use the bid price.
+pub fn market_closing_price(bid: Price, ask: Price, role: Role, position: Position) -> Price {
+    let maker_position = match (role, position) {
+        (Role::Maker, maker_position) => maker_position,
+        (Role::Taker, taker_position) => taker_position.counter_position(),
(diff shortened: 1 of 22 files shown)


model/src/cfd.rs
```diff
@@ -1523,20 +1523,26 @@ pub fn long_and_short_leverage(
...
-        (Role::Taker, Position::Long) | (Role::Maker, Position::Short) => bid,
-        (Role::Taker, Position::Short) | (Role::Maker, Position::Long) => ask,
+/// We calculate it from the perspective of the maker (i.e. the market).
+///
+/// If the maker has gone long, when closing their position they sell
+/// short. Therefore, in that case we use the ask price.
+///
+/// If the maker has gone short, when closing their position they buy
+/// long. Therefore, in that case we use the bid price.
+pub fn market_closing_price(bid: Price, ask: Price, role: Role, position: Position) -> Price {
```
bonomat, line 1537: This doc should tell us that the maker should not propose the closing price :)

bonomat (approved):
👍

model/src/cfd.rs
luckysori: Hmm, do you think so? 

I see the fact that the taker suggests the closing price as an implementation detail of the collaborative settlement protocol. By writing the code ourselves we mandate that the price is beneficial for the maker anyway, so it doesn't actually make a difference who calls this function in the context of collaborative settlement.

Another reason why I wouldn't want to talk about proposals here is that we actually also use this function in `projections` to calculate unrealized profit/loss.

thomaseizinger (approved):

luckysori:
Sorry about the unrelated `dprint` commit. I was fighting against `dprint` for a bit and it ended up downloading latest versions of the plugins we use.

I'm encountering unexpected errors on `dprint check` on CI. I can't reproduce them locally. I'm on the latest version (`0.24.1`) of `dprint` and the GitHub Action we use should also [use that one](https://github.com/dprint/check#latest-version). Any theories? cc @thomaseizinger

thomaseizinger:
> Any theories? cc @thomaseizinger

Is prettier perhaps competing with other plugins?

thomaseizinger:
cc @klochowicz Did you test what the interaction between prettier and other plugins is here? I know that prettier can format typescript so maybe they two plugins are fighting against each other?

thomaseizinger:
@luckysori I was able to run dprint and make some formatting changes :shrug:

luckysori:
> @luckysori I was able to run dprint and make some formatting changes 🤷

Thanks for fixing that for me!

model/src/cfd.rs
luckysori: We can merge this as is, and if we continue this conversation and decide to add what you suggested we can do that separately.

bonomat: I didn't want to stop you from merging this, just wanted to point it out: 
you are writing everything from the view of the maker, hence, the taker is guessing what the maker wants. Since the maker has the last call anyway (accept or reject) he should be the one proposing the closing price. 

No need to discuss this here any further and not important to act on this immediately :)

luckysori: Okay, I think I get you now. Originally you said:

> This doc should tell us that the maker should not propose the closing price :)

Did you mean the _taker_? If so I misinterpreted your message as you wanting to add this information to this docstring, rather than learning from it (once again) that we should revise the collaborative settlement protocol.

bonomat: yea 🙈 idiot me... sorry for the confusion

>  rather than learning from it (once again) that we should revise the collaborative settlement protocol.

that's what I meant.

PR #1722, opened 2022-03-30
Introduce aggregate cache in database layer

### LINKED ISSUE #1727: Introduce in-memory aggregate cache
Currently, we load all events of a CFD, every time we want to inspect or manipulate it. This is unnecessary and is starting to cause performance problems for CFDs that are open for a long time.

thomaseizinger:
See commit messages for details. This solution does have _some_ trade-offs but should scale a lot better than the caches that we previously had. Hopefully, this will improve the overall responsiveness of our application. Some of the database calls on mainnet take shockingly long which may be the cause for some of the aborted protocols?

Flamegraphing didn't show much difference but that is somewhat expected. Flamegraphs only show how the given CPI time is spent but only relative to how long things actually take. We still need to deserialize the events at some point, even with this patch set.

Where we _can_ see a difference is in the metrics:

With a 5GB database:

- On master, `projection::Initialize` takes 275 seconds
- On master, 12 `NewAttestationFetched` messages took 593 seconds on startup 
- With this patch-set, `projection::Initialize` went down to 142 seconds
- With this patch-set, 56(!) `NewAttestationFetched` messages took 88 seconds

When testing on master, I thought the app was fully initialized after 12 `NewAttestationFetched` messages but that didn't actually seem to be true. In any case, we can see a massive performance boost here!


thomaseizinger:
Oh god. I think the tests are failing because they poison each others cache ...

I guess I need to actually associate the cache with the DB connection. Thought I could avoid that -.-

daemon/src/db.rs
```diff
@@ -311,23 +313,38 @@ async fn load_cfd_events(
...
+{
+    let stream = async_stream::try_stream! {
+        let ids = sqlx::query!(
+            r#"
             select
                 id as cfd_id,
                 uuid as "uuid: model::OrderId"
             from
                 cfds
+            order by cfd_id desc
```
luckysori: ❓ Did you mean to introduce this line?

daemon/src/db.rs
```diff
@@ -357,6 +357,27 @@ where
 ///     Cases: Collaborative settlement, CET, Refund
 /// 2. Event that fails the CFD early was recorded, meaning it becomes irrelevant for processing
 ///     Cases: Setup failed, Taker's take order rejected
+pub fn load_all_open_cfds<C>(
```
luckysori: ❓ Is it intentional that the docstring above was assigned to this new function?

daemon/src/oracle.rs
```diff
@@ -320,11 +321,17 @@ impl xtra::Actor for Actor {
 
                 async move {
                     let mut conn = db.acquire().await?;
+                    let mut stream = db::load_all_cfds::<Cfd>(&mut conn, ());
```
luckysori: ❓ Isn't this meant to be `db::load_all_open_cfds`?

daemon/src/monitor.rs
```diff
@@ -591,20 +597,24 @@ impl xtra::Actor for Actor {
...
-                            monitor_commit_finality,
-                            monitor_cet_timelock,
-                            monitor_refund_timelock,
-                            monitor_refund_finality,
-                            monitor_revoked_commit_transactions,
-                            monitor_collaborative_settlement_finality,
-                            ..
-                        } = db::load_cfd(id, &mut conn, ()).await?;
-
+                    let mut stream = db::load_all_cfds::<Cfd>(&mut conn, ());
```
luckysori: ❓ Isn't this meant to be `db::load_all_open_cfds`?

daemon/src/monitor.rs
```diff
@@ -531,15 +534,18 @@ impl xtra::Actor for Actor {
...
-
-                    for id in db::load_open_cfd_ids(&mut conn).await? {
-                        let Cfd {
-                            cet,
-                            commit_tx,
-                            lock_tx,
-                            ..
-                        } = db::load_cfd(id, &mut conn, ()).await?;
-
+                    let mut stream = db::load_all_cfds::<Cfd>(&mut conn, ());
```
luckysori: ❓ Isn't this meant to be `db::load_all_open_cfds`?

daemon/src/db.rs
luckysori: I think the answer is yes. I suppose the idea is that this is the public API now. But actually we still rely on `load_open_cfd_ids` in `daemon::oracle`.

daemon/src/auto_rollover.rs
```diff
@@ -100,26 +99,34 @@ where
...
-                let cfd = db::load_cfd::<model::Cfd>(id, &mut conn, ()).await?;
-
-                match cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
-                    Ok(()) => this.send_async_safe(Rollover(id)).await?,
-                    Err(reason) => {
-                        tracing::trace!(order_id = %id, %reason, "CFD is not eligible for auto-rollover");
-                    }
+
+        let mut conn = self.db.acquire().await?;
+        let mut stream = db::load_all_cfds::<model::Cfd>(&mut conn, ());
```
luckysori: ❓ Isn't this meant to be `db::load_all_open_cfds`?

daemon/src/db.rs
```diff
@@ -99,9 +110,12 @@ async fn run_migrations(pool: &SqlitePool) -> Result<()> {
     Ok(())
 }
 
-pub async fn insert_cfd(conn: &mut PoolConnection<Sqlite>, cfd: &model::Cfd) -> Result<()> {
-    let query_result = sqlx::query(
-        r#"
+impl Connection {
```
luckysori, line 125: 👍 Thanks for making the diff easy to review!

daemon/src/db.rs
```diff
@@ -200,14 +212,34 @@ impl Connection {
...
+                let cfd = load_cfd_row(&mut conn, id).await?;
+
+                C::new(args, cfd)
+            }
+            Some(cfd) => {
+                // Got a cache entry: Downcast it to the type at hand.
 
-        let cfd = load_cfd_events(&mut conn, id, 0)
+                *cfd.downcast::<C>()
+                    .expect("we index by type id, must be able to downcast")
```
luckysori: ❓ Could this casting be removed if we use a `CfdAggregateKind` enum where the variants hold the actual types inside? Maybe it doesn't work or is not worth it.

luckysori (approved):
I already had a look yesterday, and today this PR is even easier to review after you reorganised the patches. Thank you!

I really like these changes, I had a few questions and a few things that could be slightly incorrect (or that I may have misunderstood).

daemon/src/db.rs
thomaseizinger: That would require the `db` module to depend on all possible types. At the moment, they all live in the `daemon` crate so it would be possible but really, I think there shouldn't be a need for this dependency.

daemon/src/db.rs
thomaseizinger: I did not, well spotted!

daemon/src/auto_rollover.rs
thomaseizinger: It is, well spotted!

daemon/src/db.rs
thomaseizinger: Yeah, the fact that we rely on that in `oracle` is a bit of an inconvenience!

thomaseizinger:
Force-push only contains @luckysori's feedback: https://github.com/itchysats/itchysats/compare/216f619786a6d966afd247176ac9b369c87676bf..620efb2cb1bce03db2a3650bcee2ecb03a05ed0c

luckysori (approved):
Still looks good!

daemon/Cargo.toml
```diff
@@ -12,7 +12,6 @@ bdk = { version = "0.16", default-features = false, features = ["electrum"] }
 bdk-ext = { path = "../bdk-ext" }
 btsieve = { path = "../btsieve" }
 bytes = "1"
-cached = { version = "0.34.0", default-features = false, features = ["proc_macro"] }
```
klochowicz, line 14: I would prefer if removing deserialisation cache was not part of this PR- I do not agree with the conclusion that slower startup performance is justified by removing well-tested ~20 lines of code.

It would be nice to see some data in isolation (along with data *how* much it degrades performance) in a different PR.

daemon/src/db.rs
```diff
@@ -19,15 +22,37 @@ use sqlx::pool::PoolConnection;
...
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
+use std::any::Any;
+use std::any::TypeId;
 use std::path::PathBuf;
+use std::sync::Arc;
 use time::Duration;
 
+#[derive(Clone)]
+pub struct Connection {
```
klochowicz, line 32: love this wrapper, much less awkward than passing mut-ref to `Connections` to functions!

daemon/src/db.rs
```diff
@@ -161,21 +188,163 @@ pub async fn append_event(
...
+    pub async fn load_cfd<C>(&self, id: OrderId, args: C::CtorArgs) -> Result<C>
+    where
+        C: CfdAggregate,
+    {
+        let mut conn = self.inner.acquire().await?;
 
-    Ok(())
+        let cache_key = (TypeId::of::<C>(), id);
+
+        let cfd = match self.aggregate_cache.remove(&cache_key).await {
```
klochowicz, line 217: does the version always change between this function invocations? Seems like it would be possible to remove and re-insert the same key if version didn't change?

Furthermore, wouldn't it be preferable to use `get()` here, as that would allow concurrent access without possibility of a cache miss?

klochowicz (changes requested):
It is a great changeset, I really appreciate the commit granularity . I am really keen on seeing it in production. 💯 
Really inspired by using async streams usage here.

My only reservations are about removing `cached` crate as part of the same PR - which seems orthogonal to these changes, as it is compatible with it.

daemon/src/db.rs
thomaseizinger: To apply more events I need ownership of the CFD struct so I need to remove it!

daemon/Cargo.toml
thomaseizinger: This cache solved the same problem, just in a more hacky and less scalable way ...

The number of cache hits here will be at most 3 (because we have 4 different aggregates). For all closed CFDs, the number of cache hits will be 0 because we only deserialize them for projection. For the remainder of the program, this cache will just sit around and not do anything because the new aggregate cache will ensure we don't even load the events that are in this cache now.

The reason I want to remove this is because its uselessness is not clear unless you think this through really hard. Useless code that looks useful is dangerous and hinders evolving the codebase.

thomaseizinger: There may be a possibility for more performance optimisation but those need to be re-evaluated in the context of the new functionality. The new performance is pretty good and I'd rather try to evaluate it from here again than keeping around old legacy!

daemon/src/db.rs
klochowicz: fair enough!

klochowicz: an alternative would be probably doing a `clone()` then and only updating later on. 
With the clone() we'd lose some of the benefit of trying to make it faster in the first place. though. 

probably a bit cleaner for me, but I can see how that could be subjective.

daemon/Cargo.toml
klochowicz: I created an issue: #1735

klochowicz (approved):

PR #1759, opened 2022-04-04
Default the maker parameters per network

### LINKED ISSUE #1757: Default all parameters, maker per network
To make it easier to "try the binary" I propose to default the maker parameters per network, and additionally default the network to `mainnet`. That way the `taker` does not have to specify parameters to run the binary.

Proposal: Make maker and maker-id dependent on the network:

./taker [network] [--maker] [--maker-id]

Apply defaults:

- mainnet: `--maker` `mainnet.itchysats.network:10000` `--maker-id` `7e35e34801e766a6a29ecb9e22810ea4e3476c2b37bf75882edf94a68b1d9607`
- testnet: `--maker` `testnet.itchysats.network:9999` `--maker-id` `69a42aa90da8b065b9532b62bff940a3ba07dbbb11d4482c7db83a7e049a9f1e`

(no defaults for signet)

da-kami:
fixes: https://github.com/itchysats/itchysats/issues/1757

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use clap::Parser;
@@ -34,15 +35,27 @@ mod routes;
 
 pub const ANNOUNCEMENT_LOOKAHEAD: time::Duration = time::Duration::hours(24);
 
+const MAINNET_ELECTRUM: &str = "ssl://blockstream.info:700";
+const MAINNET_MAKER: &str = "mainnet.itchysats.network:10000";
+const MAINNET_MAKER_ID: &str = "7e35e34801e766a6a29ecb9e22810ea4e3476c2b37bf75882edf94a68b1d9607";
+
+const TESTNET_ELECTRUM: &str = "ssl://blockstream.info:993";
+const TESTNET_MAKER: &str = "testnet.itchysats.network:9999";
+const TESTNET_MAKER_ID: &str = "69a42aa90da8b065b9532b62bff940a3ba07dbbb11d4482c7db83a7e049a9f1e";
+
 #[derive(Parser)]
 struct Opts {
     /// The IP address or hostname of the other party (i.e. the maker).
+    ///
+    /// If not specified it defaults to the itchysats maker for the mainnet or testnet.
     #[clap(long)]
-    maker: String,
+    maker: Option<String>,
 
     /// The public key of the maker as a 32 byte hex string.
+    ///
+    /// If not specified it defaults to the itchysats maker-id for mainnet or testnet.
     #[clap(long, parse(try_from_str = parse_x25519_pubkey))]
-    maker_id: x25519_dalek::PublicKey,
+    maker_id: Option<x25519_dalek::PublicKey>,
 
     /// The IP address to listen on for the HTTP API.
     #[clap(long, default_value = "127.0.0.1:8000")]
@@ -67,10 +80,45 @@ struct Opts {
     password: Option<rocket_basicauth::Password>,
 
     #[clap(subcommand)]
-    network: Network,
+    network: Option<Network>,
 
     #[clap(short, long, parse(try_from_str = parse_umbrel_seed))]
     umbrel_seed: Option<[u8; 32]>,
+}
+
+impl Opts {
+    fn network(&self) -> Network {
+        self.network.clone().unwrap_or_else(|| Network::Mainnet {
+            electrum: MAINNET_ELECTRUM.to_string(),
+            withdraw: None,
+        })
+    }
+
+    fn maker(&self) -> Result<(String, x25519_dalek::PublicKey)> {
+        let network = self.network();
+
+        let maker_url = match self.maker.clone() {
+            Some(maker) => maker,
+            None => match network {
+                Network::Mainnet { .. } => MAINNET_MAKER.to_string(),
+                Network::Testnet { .. } => TESTNET_MAKER.to_string(),
+                Network::Signet { .. } => bail!("No maker default URL configured for signet"),
+            },
+        };
+
+        let maker_id = match self.maker_id {
+            Some(maker_id) => maker_id,
+            None => match network {
+                Network::Mainnet { .. } => parse_x25519_pubkey(MAINNET_MAKER_ID)?,
+                Network::Testnet { .. } => parse_x25519_pubkey(TESTNET_MAKER_ID)?,
+                Network::Signet { .. } => {
+                    bail!("No maker default public key configured for signet")
+                }
+            },
+        };
+
+        Ok((maker_url, maker_id))
+    }
 }
 
 fn parse_x25519_pubkey(s: &str) -> Result<x25519_dalek::PublicKey> {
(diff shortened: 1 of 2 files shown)


thomaseizinger:
CI is failing because the way this is implemented breaks the command line interface!

I am for this change but can we implement it in a non-breaking way AND include a changelog entry?
Otherwise we might have trouble when we do the next umbrel release.

We can do a follow-up release where we transition to a different structure of passing arguments if that is preferable for some reason! :)

da-kami:
We just have to update the Umbrel parameters for the docker container when we do the next release. There should not be any problem with that.
 
I think the maker should be network specific for defaulting the maker. Otherwise there is no way to specify a default testnet maker. The follow up-commit is possible, but I don't really see why we can't break this - am I missing something? :)

.cargo/config.toml
```diff
@@ -3,7 +3,7 @@ linker = "aarch64-linux-gnu-gcc"
 
 [alias]
 dev-maker = "run --bin maker -- testnet"
-dev-taker = "run --bin taker -- --maker localhost:9999 --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e testnet" # Maker ID matches seed found in `testnet/maker_seed`
```
klochowicz: it's a bit unfortunate that parameter order matters now, and it's not backwards-compatible :/

klochowicz:
> We just have to update the Umbrel parameters for the docker container when we do the next release. There should not be any problem with that.
> 
> I think the maker should be network specific for defaulting the maker. Otherwise there is no way to specify a default testnet maker. The follow up-commit is possible, but I don't really see why we can't break this - am I missing something? :)

there is a way of specifying the default testnet maker - right now we specify different ones, we just need to look up the maker id when providing the maker (and it could easily be done in a non-breaking way too)

da-kami:
> > We just have to update the Umbrel parameters for the docker container when we do the next release. There should not be any problem with that.
> > I think the maker should be network specific for defaulting the maker. Otherwise there is no way to specify a default testnet maker. The follow up-commit is possible, but I don't really see why we can't break this - am I missing something? :)
> 
> there is a way of specifying the default testnet maker - right now we specify different ones, we just need to look up the maker id when providing the maker (and it could easily be done in a non-breaking way too)

Sorry, I cannot follow, could you elaborate more what you mean?
Before my changes the `--maker` and `--maker-id` parameters were global parameters, they were not related to the network at all. If we want to default both (i.e. maker + maker id for mainnet + testnet) we will have to relate them.
My goal is to allow users to start the binary without parameters, while at the same time allowing to configure something different when needed.

thomaseizinger:
> [snip]

> You can still default them in the code based on the network, you just have to move away from `clap`'s `default_value` feature. Make them an `Option` and based on the chosen command, fill in the `Option` if it is empty.

fair enough I can give that a try instead of the current solution

> [snip]

da-kami:
> > [snip]
> 
> > You can still default them in the code based on the network, you just have to move away from `clap`'s `default_value` feature. Make them an `Option` and based on the chosen command, fill in the `Option` if it is empty.
> 
> fair enough I can give that a try instead of the current solution
> 
> > [snip]

I squashed the previous two commits and added a fixup with such a solution. I don't like it much, but if you feel that is better I don't mind much - it does the job either way.

taker/src/main.rs
```diff
@@ -169,6 +184,32 @@ impl Network {
...
+        Some(maker) => maker,
+        None => match network {
+            Network::Mainnet { .. } => MAINNET_MAKER.to_string(),
+            Network::Testnet { .. } => TESTNET_MAKER.to_string(),
+            Network::Signet { .. } => bail!("No maker default URL configured for signet"),
+        },
+    };
+
+    let maker_id = match opts.maker_id {
+        Some(maker) => maker,
```
klochowicz:
```suggestion
        Some(maker_id) => maker_id,
```

nit: copy-paste awkwardness

taker/src/main.rs
```diff
@@ -169,6 +184,32 @@ impl Network {
...
+        None => match network {
+            Network::Mainnet { .. } => parse_x25519_pubkey(MAINNET_MAKER_ID)?,
+            Network::Testnet { .. } => parse_x25519_pubkey(TESTNET_MAKER_ID)?,
```
klochowicz: it would be great if we bailed when someone specified different maker than ours without providing maker id.

what will make it a bit tricky, is that we should match on just the domain, ports can change.

note: this can be done separately as a follow-up, as it won't break anything.

klochowicz (approved):
thanks for reworking it,  I like the second one much better! 👍

thomaseizinger:
@da-kami You edited my comment :joy: 

Thanks for reworking the solution!

taker/src/main.rs
```diff
@@ -169,6 +184,32 @@ impl Network {
...
+            Network::Signet { .. } => bail!("No maker default URL configured for signet"),
```
thomaseizinger:
```suggestion
            Network::Signet { .. } => bail!("No maker default public key configured for signet"),
```

taker/src/main.rs
```diff
@@ -169,6 +184,32 @@ impl Network {
 async fn main() -> Result<()> {
     let opts = Opts::parse();
 
+    let network = match opts.network {
+        Some(network) => network,
+        None => Network::Mainnet {
+            electrum: MAINNET_ELECTRUM.to_string(),
+            withdraw: None,
+        },
+    };
```
thomaseizinger: This could be an `.unwrap_or_else` I believe.

taker/src/main.rs
```diff
@@ -169,6 +184,32 @@ impl Network {
...
+
+    let maker_id = match opts.maker_id {
+        Some(maker) => maker,
+        None => match network {
+            Network::Mainnet { .. } => parse_x25519_pubkey(MAINNET_MAKER_ID)?,
+            Network::Testnet { .. } => parse_x25519_pubkey(TESTNET_MAKER_ID)?,
+            Network::Signet { .. } => bail!("No maker default URL configured for signet"),
+        },
+    };
```
thomaseizinger: All of this defaulting logic could be inside an `impl Opts {}` which would reduce the stuff `main` has to deal with.

thomaseizinger (approved):
Thanks for incorporating the feedback!

klochowicz:
@da-kami are we bringing this in? I've noticed it's still a draft despite approvals.  would you like me to take over and address comments made by @thomaseizinger as a second commit here?

da-kami:
> @da-kami are we bringing this in? I've noticed it's still a draft despite approvals. would you like me to take over and address comments made by @thomaseizinger as a second commit here?

Sorry, just saw this now. Since there was no urgency I did not come back to this one. Will work the rest of the feedback in now and then we can go for merge :)

da-kami:
> @da-kami You edited my comment 😂
> 
> Thanks for reworking the solution!

Hahaha, sry the edit that was unintentional!

taker/src/main.rs
da-kami: yeah, that's a different feature I suppose :)

klochowicz (approved):
awesome! excited to see testnet takers working :)

da-kami:
bors retry

PR #1765, opened 2022-04-04
Re-initialize monitoring CET after a restart

### LINKED ISSUE #1761: CET attached but state not advanced
CET was spent, the correct CET is attached but the state is not updated

```json
{
  "order_id": "ed027e22-7771-4621-bf49-928f75682d82",
  "initial_price": "39607.66",
  "accumulated_fees": -0.00005875,
  "leverage": 2,
  "trading_pair": "BtcUsd",
  "position": "Short",
  "liquidation_price": "21000000",
  "quantity_usd": "500",
  "margin": 0.01262382,
  "margin_counterparty": 0.00631191,
  "role": "Maker",
  "profit_btc": -0.00146863,
  "profit_percent": "-11.63",
  "payout": 0.01115519,
  "closing_price": "45084",
  "state": "PendingCet",
  "actions": [],
  "details": {
    "tx_url_list": [
      {
        "label": "Cet",
        "url": "https://mempool.space/tx/0d1fa922c59812016ad83c8716201d018d6c7f0b801bd48758fcdc65d8c9ac7a:0"
      },
      {
        "label": "Lock",
        "url": "https://mempool.space/tx/24551ed5367a1074e3f8443d16228a2d91dd4d44b9a93bba291afcab45c8588a:0"
      },
      {
        "label": "Commit",
        "url": "https://mempool.space/tx/14282f5fbb5bbc61f323a96917a104c7231e24e049cc57c6e540ec8e0db29b3c"
      }
    ]
  },
  "expiry_timestamp": 1648814400,
  "counterparty": "f10142ad845f2b894df8fee088d1951e7da61cb75f142aca3e5fe77de1c8054a",
  "pending_settlement_proposal_price": null
},
```

da-kami:
If we don't do this it can happen that we never find a CET if we restart before the CET is confirmed.

fixes https://github.com/itchysats/itchysats/issues/1761

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- An issue where CET finality was not monitored if we restarted the daemon before the CET was confirmed.
+  This was fixed by adding CET monitoring to the restart behaviour of the monitor actor.
+
 ## [0.4.11] - 2022-04-04
 
 ### Changed
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -131,6 +131,7 @@ struct Cfd {
 
     // Ideally, all of the above would be like this.
     monitor_collaborative_settlement_finality: Option<(Txid, Script)>,
+    monitor_cet_finality: Option<(Txid, Script)>,
 
     // Rebroadcast transactions upon startup
     lock_tx: Option<Transaction>,
@@ -154,6 +155,7 @@ impl db::CfdAggregate for Cfd {
             monitor_refund_finality: false,
             monitor_revoked_commit_transactions: false,
             monitor_collaborative_settlement_finality: None,
+            monitor_cet_finality: None,
             lock_tx: None,
             cet: None,
             commit_tx: None,
@@ -261,6 +263,7 @@ impl Cfd {
                 monitor_refund_finality: false,
                 monitor_revoked_commit_transactions: false,
                 monitor_collaborative_settlement_finality: None,
+                monitor_cet_finality: None,
                 lock_tx: None,
                 cet: None,
                 commit_tx: None,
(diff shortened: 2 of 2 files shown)


da-kami:
When working on this ticket I noticed this: https://github.com/itchysats/itchysats/issues/320#issuecomment-1087985051
For me this highlights yet again that we should do #320

thomaseizinger (approved):
Concept ACK

Can you add a changelog entry please?
I think there is some duplicated code as well, would be nice to get rid of that!

thomaseizinger:
> When working on this ticket I noticed this: https://github.com/itchysats/itchysats/issues/320#issuecomment-1087985051
> For me this highlights yet again that we should do #320 

@da-kami Our process for this is to increment the annoyance counter (and add an entry why!). It is a tool that should make these "discoveries" more objective and allow us to make data driven decisions!

da-kami:
> Concept ACK
> 
> Can you add a changelog entry please? I think there is some duplicated code as well, would be nice to get rid of that!

Getting rid of the duplicated code is a bit tricky, because for the the handler I think we should report back an error if there is one, while on startup we cannot really handle errors... Is there a non-fallible way to get the script of the CET? 

Besides that I was asking myself the question if the concept of having `cet` *and* `monitor_cet_finality` is needed; maybe `cet` (which is an option as well) is enough. Same for collab settlement. I was not sure because we would potentially have to then set the `cet` to `None` when done, so I implemented it similar to collab settlement.

thomaseizinger:
> Besides that I was asking myself the question if the concept of having `cet` _and_ `monitor_cet_finality` is needed; maybe `cet` (which is an option as well) is enough. Same for collab settlement. I was not sure because we would potentially have to then set the `cet` to `None` when done, so I implemented it similar to collab settlement.

The monitoring code is quite messy I agree. There is a lot of low-hanging fruit to consolidate things and make them easier to understand but it wasn't painful enough yet for someone to pick it up and do it!

Currently, I think `monitor::Actor` should have a much more fine-grained interface for monitoring individual transactions which can then be used for each usecase but I haven't thought about it a whole lot :)

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- An issue where CET finality was not monitored if we restarted the daemon before the CET was confirmed.
+  This was fixed by adding CET monitoring to the restart behaviour of the monitor.
```
thomaseizinger: Nit:

```suggestion
  This was fixed by adding CET monitoring to the restart behaviour of the monitor actor.
```

thomaseizinger (approved):

PR #1779, opened 2022-04-06
Move closed CFDs to separate table

### LINKED ISSUE #1651: Move closed CFDs to separate table and delete all their events
As outlined in https://github.com/itchysats/architecture/blob/main/decision-records/018-permanently-aggregate-closed-cfds.md, we want to move closed CFDs to a new table and delete all their events. This issue outlines a more detailed technical solution in how we want to actually achieve that. Despite the overall consensus on the design, there are still a couple of decision to be made. I'll try and make them explicit in this ticket so that we can actually go ahead an implement them.

1. Ensure closed CFDs are permanently closed

If we go ahead and delete events, we need to ensure that closed CFDs are actually permanently closed. This means that a transaction spending from our joint output is sufficiently confirmed on the blockchain. Hence, one change I am proposing is to change the **confirmations** for collab-close, CET and refund to **3 blocks**.

2. Ensure we keep sufficient information for closed CFDs

I propose to extract and store the following fields for a closed CFD:

- All fields from the `cfds` table
- IDs of all transactions confirmed for this CFD: In theory, the final TX would be enough but then we would have to go to the blockchain to grab the others. It isn't a whole lot of data and storing those IDs will make some features easier to implement (like listing a transaction history etc).
- The final price at which we closed the position
- The total fees we paid / received from this position (P/L should be a function of initial price, final price and fees)
- A list of all event names + their timestamp that happened for this CFD (this should likely be in separate table so we don't have to store another JSON blob in a column)

3. Have a house-keeping inspired task for actually moving the data around

Inspired by the decision in https://github.com/itchysats/architecture/blob/main/decision-records/016-regular-housekeeping.md, I propose to have a scheduled task that loads CFDs from the db, aggregates them into a dedicated struct with the above data and writes this data to the DB. Once completed, all events are deleted and as a final step, the entry in `cfds` is deleted.

4. "Load all open CFD IDs" query can be simplified to stop querying for events

All CFDs in the `cfds` table will now be open because the closed ones are in a different table

5. Projection needs to load closed CFDs from that table

## Annoyance counter

- 29/03/2022: Have to revert something in maia because we cannot upgrade dependencies otherwise: https://github.com/comit-network/maia/pull/25

luckysori:
Fixes #1651.

The original issue clearly outlines several steps that we are supposed to tackle, so here's the list of TODOs:

- [x] Require more confirmations to consider a CFD closed. See b74d0ffe13e492e9d8c43c77b626b462f7ffc248 and c59190f5ce192dcf55acdc3ec43ea5bc5de6440f.
- [x] Save CFDs with sufficient confirmations on "final" transactions to a new `closed_cfds` table.
- [x] Define housekeeping-inspired actor which periodically moves closed CFDs to the new table.
- [x] Actually use new actor in `maker` (and `taker`?).
- [x] Simplify `load_open_cfd_ids`.
- [x] Load closed CFDs in `projection`.


daemon/migrations/20220404113202_create_closed_cfds_table.sql
```diff
@@ -0,0 +1,39 @@
...
+    commit_txid text,
+    -- optional cet data
+    cet_txid text,
+    cet_vout text,
+    cet_payout text,
+    -- optional refund tx data
+    refund_txid text,
+    refund_vout text,
+    refund_payout text
+);
```
thomaseizinger: Hmm, can you explain this design a bit more?

I would have expected something like this:

- `transactions` table with columns `id`, `closed_cfd_id`, `kind`, `transaction_id`, `vout`, `value`
- `closed_cfds` table
- `transactions` would hold entries for `lock`, `close`, `cet`, `commit` and `refund`

You can then `JOIN` together the representation that you currently have in `closed_cfds` (if you wanted to) or also just make a query that only returns you refunded CFDs.

daemon/src/close_cfds.rs
```diff
@@ -0,0 +1,417 @@
...
+            return;
+        }
+
+        use CloseStatus::*;
+        if let Collaborative | NonCollaborative | Refund = new_status {
+            tracing::warn!("Close transaction status changed from {self} to {new_status}");
+        }
+
+        *self = new_status
+    }
```
thomaseizinger: Can we not imply that through the transactions that are set in the aggregate?

daemon/src/close_cfds.rs
```diff
@@ -0,0 +1,417 @@
...
+                self.own_script_pubkey = Some(dlc.script_pubkey_for(self.role));
+            }
+            ContractSetupFailed => {}
+            OfferRejected => {}
+            RolloverStarted => {}
+            RolloverAccepted => {}
+            RolloverRejected => {}
+            RolloverCompleted { dlc, funding_fee } => {
+                self.fee_account.add_funding_fee(funding_fee);
+            }
```
thomaseizinger: Shouldn't we set the commit transaction here?

daemon/src/close_cfds.rs
```diff
@@ -0,0 +1,417 @@
...
+                    vout,
+                    payout,
+                    price,
+                })
+            }
+            CollaborativeSettlementRejected => {}
+            CollaborativeSettlementFailed => {}
+            LockConfirmed => {}
+            LockConfirmedAfterFinality => {}
+            CommitConfirmed => {}
```
thomaseizinger: Mark the commit transaction as actually used here?

daemon/src/db.rs
```diff
@@ -1,33 +1,50 @@
...
 use sqlx::sqlite::SqliteConnectOptions;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
 use std::any::Any;
 use std::any::TypeId;
+use std::fmt;
 use std::path::PathBuf;
+use std::str;
 use std::sync::Arc;
 use time::Duration;
```
thomaseizinger: This newline prevents merging and sorting the imports.

daemon/src/db.rs
```diff
@@ -348,6 +365,444 @@ impl Connection {
...
+            non_collaborative_settlement,
+            refund,
+            close_status,
+            ..
+        } = cfd;
+
+        let (lock_txid, lock_dlc_vout) = match lock {
+            Some(Lock { txid, dlc_vout }) => (txid, dlc_vout),
+            None => bail!("Closed CFD without lock"),
+        };
```
thomaseizinger: I think we should design the API of `close_cfds::Cfd` such that this is impossible. Private field and getter perhaps?

thomaseizinger: We should not pass a `Cfd` into here that is not actually closed.

daemon/src/db.rs
```diff
@@ -348,6 +365,444 @@ impl Connection {
...
+
+        let kind = match close_status {
+            CloseStatus::NotClosed => bail!("Cannot insert open CFD in closed CFDs table"),
+            CloseStatus::Collaborative => ClosedCfdKind::Collaborative,
+            CloseStatus::NonCollaborative => ClosedCfdKind::NonCollaborative,
+            CloseStatus::Refund => ClosedCfdKind::Refund,
```
thomaseizinger: If you split the table layout into `closed_cfds` and `transactions`, this insertion logic should become a lot simpler because you just need to insert all transactions and don't care as much about why it was closed.

That reasoning can be done upon fetching a closed CFD from this table based on which transaction we were able to load.

daemon/src/db.rs
```diff
@@ -348,6 +365,444 @@ impl Connection {
...
+            settlement: Settlement::NonCollaborative {
+                commit_txid: cfd.commit_txid,
+                payout: SettlementPayout::Refund {
+                    txid: cfd.refund_txid,
+                    vout: cfd.refund_vout,
+                    payout: cfd.refund_payout,
+                },
+            },
+        })
+    }
```
thomaseizinger: Hmm, I am not convinced that this is a good abstraction as it is currently implemented. You don't know how a CFD was closed before you load it, hence having the caller choose the correct function is a bit weird?

Can we somehow restructure things so that we have three functions:

- `load_refunded_cfds`
- `load_force_closed_cfds`
- `load_collaboratively_closed_cfds`

Those would all return the same `ClosedCfd` which could have different constructors, based on where it is called from. Not sure if it is too wasteful but we could just reload all closed in projection upon every cycle of closing them.

I guess achieving data consistency there is actually a bit tricky. I think we want a CFD to "vanish" from the `cfds` table and appear in the `closed_cfds` atomically but at the same time, projection needs to update its local state as well.

Thinking out loud further: projection has a local cache so I think the responsibility of maintaining data consistency needs to solved there. The cache entry will still be there even if the `close_cfd::Actor` deletes all the events. Thus, `projection` should probably just regularly load all closed CFDs and update its internal cache to remove the entry and instead, populate it from the `closed_cfds` table.

Alternatively, we could make this completely transparent in the `db` layer where `db::load_cfd` can also return an instance from the `closed_cfds` table but as we discussed offline, this is a bit annoying because not every module needs to deal with closed CFDs. It does seem like a much cleaner abstraction though because it hides those DB details. Thoughts?

thomaseizinger (commented):
Did an initial pass, thanks for pushing this forward! :)

A few things to think about!

daemon/src/close_cfds.rs
luckysori: That's what I assumed at first, but I thought we could run into trouble. For example, what would happen if (somehow) the CFD had a collaborative settlement and commit transaction? I wanted to avoid having to write any logic around odd scenarios like that one.

I think it's safer to assume that only one of the final transactions will get 3 confirmations. So setting the `CloseStatus` based solely on those events should be much safer.

luckysori: Writing that comment does highlight to me that the warning above should be a `debug_assert` or an `expect` though.

daemon/src/close_cfds.rs
luckysori: Good catch!

daemon/migrations/20220404113202_create_closed_cfds_table.sql
luckysori: > Hmm, can you explain this design a bit more?

I opted for something very simple, but I assumed that we could do something better :D I will follow your advice here.

daemon/src/close_cfds.rs
luckysori: Do we care about that in `projection`?

daemon/src/db.rs
luckysori: Yeah, I agree. I wasn't happy with all these `bail`s.

daemon/src/close_cfds.rs
luckysori: Actually, no. Rolling over doesn't publish a commit transaction. The `ClosedCfd` struct (and the corresponding db model) only records transactions that are final.

daemon/migrations/20220404113202_create_closed_cfds_table.sql
da-kami: I agree with @thomaseizinger's proposal, that should bring us (close to) [3NF](https://en.wikipedia.org/wiki/Third_normal_form). We use a relational database so it'd be good to follow the recommended paradigms.

daemon/src/close_cfds.rs
```diff
@@ -0,0 +1,417 @@
...
+            CetConfirmed => self.close_status.set(CloseStatus::NonCollaborative),
+            RefundConfirmed => self.close_status.set(CloseStatus::Refund),
+            RevokeConfirmed => {}
```
da-kami: Until we actually implement revoke logic we might want to consider treating this as closed?
I'm not saying we have to do this as part of this PR. It could be treated as separate "feature" for e.g. housekeeping to clean this up. I found it somewhat related to mentioned it here.

daemon/migrations/20220404113202_create_closed_cfds_table.sql
luckysori: Let me know if the new design makes more sense!

daemon/migrations/20220404184003_create_event_log_table.sql
```diff
@@ -0,0 +1 @@
+
```
da-kami: Might want to remove this file, it appears to be empty :)

daemon/src/close_cfds.rs
```diff
@@ -0,0 +1,48 @@
...
+            .add(this.send_interval(CLOSE_CFDS_INTERVAL, || CloseCfds));
+    }
+
+    async fn stopped(self) -> Self::Stop {}
+}
+
+#[xtra_productivity]
+impl Actor {
+    async fn handle(&mut self, _: CloseCfds) {
+        if let Err(e) = self.db.move_to_closed_cfds().await {
```
da-kami, line 42: 🤔 Would it be possible to let this be handled in a separate task or could that cause trouble with database connections?

Would be cool if we would not block the application on "cleanup".

daemon/src/db.rs
```diff
@@ -272,7 +289,7 @@ impl Connection {
             drop(conn);
 
             for id in ids {
-                let cfd = self.load_cfd(id, args.clone()).await?;
+                let cfd = self.load_open_cfd(id, args.clone()).await?;
```
da-kami, line 294: 🛠 I think the function name `load_all_cfds` does not reflect what this function does. Is this supposed to load all CFDs including closed ones?
As far as  can tell it does not load closed CFDs. What's the difference to `load_all_open_cfds`?

daemon/src/db.rs
```diff
@@ -323,22 +340,204 @@ impl Connection {
...
+                closed_cfds.uuid = $1
+            "#,
+            id
+        )
+        .fetch_one(&mut conn)
+        .await?;
+
+        let expiry_timestamp = OffsetDateTime::from_unix_timestamp(cfd.expiry_timestamp)?;
+
+        let collaborative_settlement = load_collaborative_settlement_tx(&mut conn, id).await?;
```
da-kami, line 482: 🙃 I think it could be more efficient to use `join`s to load all the relevant data in one query, not sure it's worth it though. We could run some performance tests to see how much of a difference it makes.

daemon/src/db.rs
```diff
@@ -463,6 +1053,276 @@ async fn load_cfd_events(
...
+
+    if query_result.rows_affected() < 1 {
+        anyhow::bail!("failed to delete from events");
+    }
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
```
da-kami, line 1511: 🛠 I am not sure if you are planning to do this, but please add tests to the new database functions!

We can test whole roundtrips within the data access layer, e.g. `insert_cfd` -> `load_cfd` -> append closes state -> `move_to_closed_cfds` -> `load_closed_cfd` (... etc.)

daemon/migrations/20220404113202_create_closed_cfds_tables.sql
```diff
@@ -0,0 +1,55 @@
...
+);
+CREATE UNIQUE INDEX IF NOT EXISTS closed_cfds_uuid ON closed_cfds (uuid);
+CREATE TABLE IF NOT EXISTS collaborative_close_txs (
+    id integer PRIMARY KEY autoincrement,
+    cfd_id integer NOT NULL,
+    txid text NOT NULL,
+    vout integer NOT NULL,
+    payout integer NOT NULL,
+    price text NOT NULL,
+    FOREIGN KEY (cfd_id) REFERENCES closed_cfds (id)
```
bonomat, line 25: What do you think of adding a `timestamp` to each table? 
That way we can tell immediately when a certain transaction was closed/committed/refunded.

daemon/migrations/20220404184003_create_event_log_table.sql
bonomat: I guess this file is not needed or did you mean to move the last `Create table` statement from the other table into this file?

daemon/src/monitor.rs
```diff
@@ -30,7 +30,11 @@ use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
 
-const FINALITY_CONFIRMATIONS: u32 = 1;
+const LOCK_FINALITY_CONFIRMATIONS: u32 = 1;
+const CLOSE_FINALITY_CONFIRMATIONS: u32 = 3;
```
da-kami, line 34: ⚠️ Without dealing with in-mempool changing this to `3` will cause the UI not to update as expected. 

My idea would have been to keep finality for the UI at `1` confirmation, but only prune internally when we reach 3. But that is a bigger change because we would have to extend monitoring and add additional events (or change the events we append).
We *can* opt for changing this to `3` but we have to be aware that we will only move the `Cfd` to the closed `Cfd` later then, which will be a somewhat unexpected UX change, because nr of confirmations is not reflected in the UI.

daemon/src/db.rs
```diff
@@ -323,22 +340,204 @@ impl Connection {
...
+            }
+
+            if let Some(refund) = closed_cfd.refund {
+                insert_refund_tx(&mut db_tx, id, refund).await?;
+            }
+
+            delete_from_events_table(&mut db_tx, id).await?;
+            delete_from_cfds_table(&mut db_tx, id).await?;
+
+            db_tx.commit().await?;
```
bonomat: Since you have a `commit` here, I would expect that we need to handle `rollback` as well. 
Did you check what happens if one of the many statements above fails?

daemon/src/db.rs
```diff
@@ -323,22 +340,204 @@ impl Connection {
             from
                 cfds
             where not exists (
+                select id from EVENTS as events
+                where cfd_id = cfds.id and
```
bonomat: Should this be? 

```suggestion
                where events.cfd_id = cfds.id and
```

I'm not sure how smart sql is but above you said `id as cfd_id` and then access it here again with `cfds.id` which means this is always true? :D

daemon/src/projection.rs
```diff
@@ -818,6 +822,159 @@ impl db::CfdAggregate for Cfd {
...
+            taker_leverage,
+            n_contracts,
+            counterparty_network_identity,
+            role,
+            fees,
+            expiry_timestamp,
+            lock,
+            commit,
+            collaborative_settlement,
+            non_collaborative_settlement,
```
da-kami: Any reason why you opted for `non_collaborative_settlement` over `cet` here? 
In the database table it is called `cet`. 

For me `commit` + `cet` and `collaborative_settlement` would be OK as well.

daemon/src/projection.rs
```diff
@@ -818,6 +822,159 @@ impl db::CfdAggregate for Cfd {
...
+
+        let (details, closing_price, payout, state) = {
+            let mut tx_url_list = HashSet::default();
+
+            tx_url_list.insert(
+                TxUrl::new(lock.txid.into(), network, TxLabel::Lock)
+                    .with_output_index(lock.dlc_vout.into()),
+            );
+
+            // TODO: Can we hide this in the DB layer?
```
da-kami: 🤔 Not sure the database layer should know about `CfdState`, that might be only relevant for the projection.
But the database layer could provide payout and price I suppose.

daemon/src/db.rs
```diff
@@ -350,6 +549,393 @@ impl Connection {
     }
 }
 
+async fn load_collaborative_settlement_tx(
```
bonomat: You called the table `collaborative_close_txs`, might be a good idea to align the naming. i.e. `load_collaborative_close_tx` or the otherway around.

daemon/src/projection.rs
```diff
@@ -818,6 +822,159 @@ impl db::CfdAggregate for Cfd {
     }
 }
 
+impl db::ClosedCfdAggregate for Cfd {
+    fn new_closed(network: Self::CtorArgs, closed_cfd: db::ClosedCfd) -> Self {
```
da-kami, line 829: ❓ How are you planning to use this? A separate event for closed CFDs? 

At the moment the UI would just not display any closed CFDs or am I missing something?

daemon/src/db.rs
```diff
@@ -350,6 +549,393 @@ impl Connection {
...
+            txid as "txid: model::Txid",
+            vout as "vout: model::Vout",
+            payout as "payout: model::Payout",
+            price as "price: model::Price"
+        FROM
+            collaborative_close_txs
+        JOIN
+            closed_cfds c on c.id = collaborative_close_txs.cfd_id
+        WHERE
+            uuid = $1
```
bonomat: Maybe it's worth mentioning from which table this uuid is coming from, e.g. `closed_cfds.uuid` or `c.uuid`.  I guess this only works because `uuid` is only available in `closed_cfds`. If we ever add a field into `collaborative_close_txs` named `uuid` this could become a problem.

model/src/lib.rs
```diff
@@ -529,7 +528,7 @@ impl_sqlx_type_display_from_str!(Identity);
 #[derive(Debug, Clone)]
 pub struct WalletInfo {
     pub balance: Amount,
-    pub address: Address,
+    pub address: bdk::bitcoin::Address,
```
da-kami: ❓Why do we fully qualify these?

model/src/lib.rs
```diff
@@ -899,6 +898,152 @@ impl str::FromStr for TxFeeRate {
...
+    fn from(fees: &Fees) -> Self {
+        fees.0.as_sat() as i64
+    }
+}
+
+impl_sqlx_type_integer!(Fees);
+
+/// The number of contracts per position.
+#[derive(Debug, Clone, Copy)]
+pub struct Contracts(u64);
```
da-kami, line 983: ❓ This was introduced on the way, but we only use it in closed cfds so far? I'm OK with that, but I'd expect some follow up tickets to clean this up in open ones as well, otherwise we might have even more confusing terminology mess.

bonomat (commented):
On first sight looks really good. Added some comments for now. I'll need to have a second look at it tomorrow as my brain is getting a bit tired today .

daemon/migrations/20220404113202_create_closed_cfds_tables.sql
bonomat: Thinking about it: we'd need to copy the timestamp from the `event` table.

daemon/src/db.rs
```diff
@@ -380,7 +982,11 @@ pub trait CfdAggregate: Clone + Send + Sync + 'static {
     fn version(&self) -> u32;
 }
 
-async fn load_cfd_row(conn: &mut PoolConnection<Sqlite>, id: OrderId) -> Result<Cfd> {
+pub trait ClosedCfdAggregate: CfdAggregate {
```
da-kami, line 1043: Interesting that these traits live in the database. I feel that should probably be driven by the model. But we don't have to solve this as part of this PR.

da-kami (commented):
This is indeed getting quite big. I think most importantly I would add tests for the db logic before merging this. I added some other comments. The change of confirmations to `3` has to be evaluated because it affects the UX. But I don't see that as a blocker. Same for the UI not displaying closed CFDs at the moment. I suppose you are planning to tackle this in a follow-up :)

daemon/migrations/20220404113202_create_closed_cfds_tables.sql
luckysori: I think you're thinking about debugging, since I don't think this information is needed in `projection`. I see how it could be useful, although it's fairly easy to take the txid and use a block explorer.

> Thinking about it: we'd need to copy the timestamp from the event table.

Incidentally, that timestamp reflects when our system noticed that a particular transaction got sufficient confirmations to consider it final. Is that what you would like to have when debugging?

daemon/migrations/20220404184003_create_event_log_table.sql
luckysori: Nice catch! I originally had two files and I finally opted to remove this one, but forgot to do so haha.

daemon/src/db.rs
luckysori: This is a great question. I need to verify how we're supposed to do this. My intention is to make the "move a CFD to the closed CFDs table operation" atomic, and I think for that we need to use this APIs. But it's very possible I've made a mistake.

daemon/src/db.rs
da-kami: I will work without qualifying the table name here, because `cfd_id` only exists in `events`, but we should probably generally qualify the tables by name to make this cleaner.

daemon/src/db.rs
luckysori: Fair point. I did not achieve consistent naming throughout. I'll align it.

daemon/src/db.rs
luckysori: Good idea. I'll try to be more explicit!

daemon/src/close_cfds.rs
luckysori: I honestly didn't know what to do at this stage D:

daemon/src/close_cfds.rs
luckysori: Hmm, my understanding is that this actor shouldn't block the rest of the application by itself.

It may have some effect on the ability for other parts of the codebase to access the database (I'm not exactly sure how sqlite behaves with concurrent access, but that is something that we had to contend with before this PR anyway), but I don't think spawning separate tasks would make a difference. That is because there are no other handlers in `close_cfds::Actor` that could be blocked. Theoretically it could block itself if the previous call to `move_to_closed_cfds` took longer than 5 minutes, but that would be surprising.

daemon/src/db.rs
luckysori: Good on you for noticing this! It was a mistake. I've amended this with a recent force push.

daemon/src/db.rs
luckysori: I am more than happy to make this more performant if you see any opportunities like the one you mention. If you think the current implementation is good enough in that sense I would rather do it in a follow-up PR though.

daemon/src/db.rs
luckysori: Fair point. I will add tests to this module!

daemon/src/monitor.rs
luckysori: I'm glad you paid attention to this change.

I originally considered introducing a separate event, because I had a similar concern. I was more worried about any business logic that could depend on the `{CollaborativeSettlement, Cet, Refund}Confirmed` events, but it turns out that only the UI would be affected.

I thought it was not incorrect to consider a final transaction pending in the UI until it receives 3 confirmations. But if other people care about this unplanned change I think it's totally reasonable to add some `{CollaborativeSettlement, Cet, Refund}Seen` events so that this behaviour doesn't change. Are we able to introduce new events without creating problems in the database layer?

cc @klochowicz @thomaseizinger @bonomat

daemon/src/projection.rs
luckysori: Yeah, I've annoyed myself with this choice too.

I'm working on a fixup commit that should make this better.

daemon/src/projection.rs
luckysori: The TODO wasn't very clear, sorry. It was referring to the way we expose optional fields and end up needing to deal with an `unreachable`.

I'm fixing this with my next fixup commit.

daemon/src/projection.rs
luckysori: This is used when we call `load_all_cfds` now (if there are any closed CFDs in the database that is).

model/src/lib.rs
luckysori: Good catch. I added an unnecessary `Address` newtype at some point.

model/src/lib.rs
luckysori: You have a point.

I think it's worth introducing this to eventually fully replace `Usd`, so I see this as a first step. It's my understanding that we cannot easily make changes to the original tables at the moment, but once we can I would like to do what you're suggesting.

daemon/src/db.rs
luckysori: I did not think hard about where to put this trait because I followed what was already there i.e. the `CfdAggregate` trait.

I think they belong here because their purpose is to construct aggregates from _database_ types. Which is why they take `db::Cfd` and `db::ClosedCfd` as arguments.

daemon/src/db.rs
da-kami: Definitely not a blocker, follow-up is fine!

daemon/src/monitor.rs
da-kami: If you introduce `...Seen` events I think we should be fine. Old CFDs will not have these events, so we should clean them up correctly. I don't see problems with that.

bonomat: > I originally considered introducing a separate event, because I had a similar concern. I was more worried about any business logic that could depend on the {CollaborativeSettlement, Cet, Refund}Confirmed events, but it turns out that only the UI would be affected.

now that you say that. 
I think for `CollabSettlement` it should be fine as we (the maker) are publishing it
For the CET it should also not be a problem because we transition to _closed_ when the oracle attests as well, meaning, we won't miss it neither. 
For refund: we are f**ked anyway and will likely lose some money 😅

daemon/migrations/20220404113202_create_closed_cfds_tables.sql
bonomat: Yes, I thought about debugging. I use the timestamp in our table quite often to check when a certain event was recorded.

daemon/src/db.rs
bonomat: I think your approach is correct. I just wonder if we need to call `rollback` explicitly or if it is done automatically if `db_tx` gets out of scope and `.commit()` was not called.

daemon/src/db.rs
```diff
@@ -323,22 +340,204 @@ impl Connection {
...
+
+    /// Check if the CFD corresponding to the `id` given is in the
+    /// `closed_cfds` table.
+    async fn is_closed_cfd(&self, id: OrderId) -> Result<bool> {
+        let mut conn = self.inner.acquire().await?;
+
+        let cfd = sqlx::query!(
+            r#"
+            SELECT
+                uuid as "id: model::OrderId"
```
bonomat: Given you don't care about the `uuid` you could just `select 1` which is a bit more performant than selecting the actual item. 

```suggestion
1
```

daemon/src/db.rs
```diff
@@ -350,6 +549,393 @@ impl Connection {
...
+            RefundTimelockExpired { refund_tx } => {
+                let own_script_pubkey = self
+                    .own_script_pubkey
+                    .as_ref()
+                    .expect("DLC to exist if refund timelock has expired");
+                let OutPoint { txid, vout } = refund_tx
+                    .outpoint(own_script_pubkey)
+                    .expect("spend script to be in spend tx");
+
+                let payout = &refund_tx.output[vout as usize];
```
bonomat: will this panic if the index does not exist? I mean, it should not happen, but it might be better to do a `.get()`?

bonomat (approved):
Added a few more comments. I think this is excellent work. Let's get this tested on testnet 🥳

daemon/src/db.rs
luckysori: From the documentation:

> A transaction should end with a call to [`commit`] or [`rollback`]. If neither are called before the transaction goes out-of-scope, [`rollback`] is called. In other words, [`rollback`] is called on `drop` if the transaction is still in-progress.

I think we can rely on the `Transaction` (`db_tx` here) to go out of scope when we fail before reaching the end of the function where we call `commit`.

bonomat: cool. that was what I was hoping for :)

daemon/src/db.rs
luckysori: Interestingly only `c.uuid` works.

daemon/src/db.rs
luckysori: I've opened https://github.com/itchysats/itchysats/issues/1807 to track this.

model/src/lib.rs
luckysori: I opened https://github.com/itchysats/itchysats/issues/1808.

daemon/src/db.rs
luckysori: Thanks for the suggestion! I think this doesn't work because `sqlx` expects a Rust identifier here i.e. it won't compile :(

daemon/src/db.rs
luckysori: It is highly unlikely, but I've tried to remove all possible panics from `apply` by making it fallible. If we fail to apply the events we will get a handled error rather than a panic now.

daemon/src/db.rs
bonomat: stupid sqlx ... thanks for trying though

daemon/src/monitor.rs
luckysori: @da-kami: In order to get this in soon I would opt for accepting this change in UI behaviour for now, record a ticket to introduce the `Seen` events and later on decide whether we want to tackle that or not. Would that be okay?

daemon/src/db.rs
luckysori: I've added a few tests with the last fixup. They're only based on collaborative closed CFDs because it was easiest to get real events for those.

I've added tests that only concern the `db` layer, proving that the data is moved around correctly.

I've also added a test to verify that the only user of the closed CFDs table (i.e. `projection`) can't tell the difference between loading from `cfds` and `closed_cfds`. The test is not perfect for a few reasons: depending on `projection` in `db/tests`; needing to make some things public in `projection`; the `projection::Cfd` type is built differently for both types of CFD, so the comparison has some caveats.

I still think the test is valuable and we can always iterate on the approach.

daemon/migrations/20220404113202_create_closed_cfds_tables.sql
luckysori: I've added this in a fixup commit!

daemon/src/close_cfds.rs
da-kami: fair enough, I think I was not thinking it through in depth when I posted my comment.

daemon/src/db.rs
```diff
@@ -324,21 +366,235 @@ impl Connection {
...
+                }),
+            ) => Settlement::Refund {
+                commit_txid,
+                txid,
+                vout,
+                payout,
+            },
+            _ => {
+                bail!(
+                    "Closed CFD has insane combination of transactions:
```
da-kami, line 546: I hope we'll never see that 😅
(Famous last words...)

daemon/src/db.rs
```diff
@@ -367,6 +623,409 @@ pub struct Cfd {
...
+    refund: Option<Refund>,
+}
+
+/// Auxiliary type used to gradually combine a `Cfd` with its list of
+/// `CfdEvent`s.
+///
+/// Once all the `CfdEvent`s have been applied, we can build a
+/// `ClosedCfdInput` which is used for database insertion.
+#[derive(Debug, Clone)]
+struct ClosedCfdInputAux {
```
da-kami: For me this is the same concept as what we called `...Aggregate` in e.g. the projection.
If you agree we might want to align the name. Not blocking though :)

daemon/src/db.rs
da-kami: If they belong to the `db` I find it a bit odd that we re-use them in other parts of the code. Not super important though. We can drop this discussion for now or discuss somewhere else.

daemon/src/db.rs
```diff
@@ -545,103 +1593,152 @@ mod tests {
...
-        let cfd_not_final = dummy_cfd();
-        db.insert_cfd(&cfd_not_final).await.unwrap();
+        let (cfd, contract_setup_completed, collaborative_settlement_completed) =
+            cfd_collaboratively_settled();
+        let order_id = cfd.id();
 
-        db.append_event(lock_confirmed(&cfd_not_final))
+        db.insert_cfd(&cfd).await.unwrap();
+
+        db.append_event(contract_setup_completed).await.unwrap();
```
da-kami, line 1721: 🛠 I think these two `append_event` here don't make sense in the context of this test. The cfd in question is already settled, so why are we appending `contract_setup_completed` and `collaborative_settlement_completed`?
(It won't make a difference to the database, but it is confusing to read the test...)

daemon/src/monitor.rs
da-kami: fine with me, we don't have to block this.

daemon/src/projection.rs
```diff
@@ -828,7 +974,7 @@ impl State {
     }
 
     async fn update_cfd(&mut self, db: db::Connection, id: OrderId) -> Result<()> {
-        let cfd = db.load_cfd(id, self.network).await?;
+        let cfd = db.load_open_cfd(id, self.network).await?;
```
da-kami, line 979: ⚠️ we have to take `closed` cfds into account here

I think we can expose `load_cfd` and use that here... then we always get the CFD no matter if closed or open.
Otherwise we will need some condition here as well, to be able to load the cfd. 
This is somewhat critical, because otherwise once we moved a cfd to closed we will error here when loading, because the order_id won't be found in "open" anymore.

da-kami (changes requested):
Thank you for incorporating all the feedback and adding some tests! I'd be keen to add some more, but that can be done in a follow up.

Updating the CFDs in projection not taking closed into account is blocking for me.
We could decide to merge and do a follow up, but we can then *not* release until the follow-up is in...

daemon/src/db.rs
luckysori: The helper `cfd_collaboratively_settled` returns a `Cfd` and a couple of events which when applied to the `Cfd` put it in a state where it is _then_ settled.

This is a requirement for the test to pass because:

- moving a CFD from `cfds` to `closed_cfds` fails if contract setup was not completed (we need a lock transaction).
- loading a closed CFD that doesn't have a `cet/collaborative_settlement/refund` transaction fails.

That's why we have to appends those two specific events to test this logic.

daemon/src/db.rs
luckysori: It is not `unreachable`, but I also hope we don't see such an error. The good thing is that the error would not break anything. The CFD would remain in the `cfds` table.

daemon/src/db.rs
luckysori: It is an internal aggregate of the db, true! I can call it `ClosedCfdInputAggregate`.

daemon/src/projection.rs
luckysori: It's good that you noticed this behaviour, but I think I disagree with your conclusion.

We call `update_cfd` when we receive a new `CfdEvent` for a CFD. A closed CFD does not need to be updated any further. We don't expect any more events and, more importantly, we don't want to record any more events.

da-kami: > 

hm... interesting point. So we would just keep the last state in the projection and that CFD will only be loaded again on restart. Sounds legit to me. Let's get it in then and see what happens 😄

da-kami (approved):

daemon/src/monitor.rs
luckysori: I opened https://github.com/itchysats/itchysats/issues/1815.

PR #1780, opened 2022-04-07
Introduce `projection::CfdState::RolloverSetup` variant

### LINKED ISSUE #1701: Don't reuse `ContractSetup` for when rollingover

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/1701.

So that we no longer reuse the `CfdState::ContractSetup` variant after the rollover protocol has been accepted but not yet completed.

This change does not introduce any clever behaviour into the frontend. We simply treat this new state as if the rollover was still
pending (which is technically true).

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -422,7 +422,7 @@ async fn rollover_an_open_cfd(maker_position: Position) {
 
     maker.system.accept_rollover(order_id).await.unwrap();
 
-    wait_next_state!(order_id, maker, taker, CfdState::ContractSetup);
+    wait_next_state!(order_id, maker, taker, CfdState::RolloverSetup);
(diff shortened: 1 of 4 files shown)


thomaseizinger (approved):

klochowicz:
concept ACK. Happy for it to come through as soon as the tests are passing!

PR #1781, opened 2022-04-07
Instrument substream with metrics

thomaseizinger:
As a nice benefit, wrapping the original `Negotiated<yamux::Stream>`
makes the API of `libp2p-xtra` more contained by not exposing details
about `yamux`.


libp2p-xtra/src/substream.rs
```diff
@@ -0,0 +1,158 @@
...
+///
+/// Opening a new substream is cheap compared to say a TCP connection. Additionally, because
+/// substreams are opened on top of an existing connection, they are not affected by NAT or
+/// firewalls. Once a connection is established, either end of the connection (the original dialer
+/// and listener) can open substreams.
+///
+/// Each substream is dedicated to a specific protocol which must be specified upon construction.
+///
+/// Substreams are instrumented with prometheus metrics that track the duration they are alive for
+/// and how many bytes are read from and written to the stream.
```
luckysori, line 29: 👍 Thanks for the detailed docstring.

libp2p-xtra/src/substream.rs
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    fn poll_close(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<std::io::Result<()>> {
+        let this = self.project();
+
+        this.inner.poll_close(cx)
+    }
+}
+
+const PROTOCOL_LABEL: &str = "protocol";
```
luckysori, line 134: ❓ Does this mean that every protocol will have the same label? Shouldn't this be configurable?

luckysori (approved):
I love this idea and how quickly you implemented it!

libp2p-xtra/src/substream.rs
thomaseizinger: This is the label "key"! The labels are a `HashMap`. One label can have as many values (the actual protocol names) as you want but you need to specify the labels upfront!

thomaseizinger:
> I love this idea and how quickly you implemented it!

Thanks! I am quite stoked myself about what we can do with this data :)

thomaseizinger:
Note to self: Include dialer/listener role in metric.

klochowicz (approved):
really cool 👍

PR #1782, opened 2022-04-07
Don't strip crate from actor name

thomaseizinger:
We have actors that are defined in their own crates, like
`xtra_bitmex_price_feed`. Those show up just as `Actor` in the logs.

The crate stripping made sense when all actors where located in the
`daemon` crate but that is no longer the case.


luckysori (approved):
Makes sense. It would be nice if could override this for situations where the name is really long.

PR #1784, opened 2022-04-08
Add xtra-libp2p-ping actor

### LINKED ISSUE #1767: Implement `libp2p-ping` protocol
To ensure our connection does not get closed due to inactivity by intermediary network devices, we need to regularly ping each node. This is the equivalent of our current heartbeat.

libp2p implements this on a _per-substream_ basis where a node that wishes to ping:

- opens a substream with the ID `/ipfs/ping/1.0.0`
- sends random 32 bytes to the party
- expects the same 32 bytes to be sent back
- closes the substream

The protocol is implemented here: https://github.com/libp2p/rust-libp2p/blob/master/protocols/ping/src/protocol.rs

We could actually use these low-level functions if we wanted but depending on `libp2p-ping` will pull in extra dependencies so I don't think it is worth it. We can just copy them into our project. I would suggest to introduce this protocol as a separate crate so we don't increase compile-times too much as we add these new protocols.

A successful ping will also measure the latency to the other party.

Decisions that are yet to be made:

- Should only the taker ping the maker or should both ping each other (to retrieve the latency)?
- Should failure to ping have an impact the connection status?
- What metrics do we want to record about pings?
	- A histogram of ping latencies would be quite useful to get an idea of how well connected a node is.

thomaseizinger:


klochowicz (approved):
LGTM!

xtra-libp2p-ping/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+///
+/// Using this indicates that we are wire-compatible with other libp2p/ipfs nodes.
+pub const PROTOCOL_NAME: &str = "/ipfs/ping/1.0.0";
+
+/// An actor implementing the official ipfs/libp2p ping protocol.
+///
+/// The ping protocol serves two purposes:
+///
+/// 1. To measure the latency to other peers.
+/// 2. To prevent an otherwise little-utilised connection from being closed by intermediary network
```
luckysori: 🙃 
```suggestion
/// 2. To prevent an otherwise seldom-utilised connection from being closed by intermediary network
```

luckysori (approved):
LGTM!

PR #1786, opened 2022-04-08
A few fixes around libp2p-xtra

klochowicz:
Additions and fixes discovered during integration

### DIFF
--- a/libp2p-xtra/src/endpoint.rs
+++ b/libp2p-xtra/src/endpoint.rs
@@ -70,9 +70,11 @@ pub struct OpenSubstream<P> {
     marker_num_protocols: PhantomData<P>,
 }
 
+/// Marker type denominating a single protocol.
 #[derive(Clone, Copy, Debug)]
 pub enum Single {}
 
+/// Marker type denominating multiple protocols.
 #[derive(Clone, Copy, Debug)]
 pub enum Multiple {}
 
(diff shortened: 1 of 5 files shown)


libp2p-xtra/examples/dialer.rs
```diff
@@ -0,0 +1,72 @@
...
+struct Opts {
+    #[clap(long)]
+    multiaddr: Multiaddr,
+
+    #[clap(long, default_value = "ExampleDialer")]
+    name: String,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
```
thomaseizinger: Perhaps we should call the file `hello_world_dialer`.

libp2p-xtra/src/endpoint.rs
```diff
@@ -381,30 +382,48 @@ impl Endpoint {
         let this = ctx.address().expect("we are alive");
         let listen_address = msg.0.clone();
 
-        self.listen_addresses.insert(listen_address.clone()); // FIXME: This address could be a "catch-all" like "0.0.0.0" which actually results in
-                                                              // listening on multiple interfaces.
```
thomaseizinger, line 392: If possible, can we keep this one here and remove the line in the last commit with the rationale that we want to wait until the call to `listen_on` succeeds before we add it?

thomaseizinger (approved):

libp2p-xtra/src/endpoint.rs
klochowicz: I'm afraid I don't follow. are you talking about the  the comment or the whole 2 lines? and remove the whole solution from last commit?

thomaseizinger: I think this commit should not touch these two lines and instead we can move the insert from here to what we have in the last commit!

PR #1791, opened 2022-04-11
Add hello world listener example for libp2p-xtra

klochowicz:
Simple example that can work in conjunction with dialer.


luckysori (approved):
Thanks for adding another example!

PR #1800, opened 2022-04-12
Set commit tx to published when `OracleAttestedPriorCetTimelock`

da-kami:
Whenever we are in `PendingCommit` the link should be available already.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -494,6 +494,7 @@ impl Cfd {
                 self.aggregated.timelocked_cet = Some(timelocked_cet);
                 self.closing_price = Some(price);
 
+                self.aggregated.commit_published = true;
                 self.aggregated.state = CfdState::PendingCommit;
             }
             OracleAttestedPostCetTimelock { cet, price, .. } => {


daemon/src/projection.rs
```diff
@@ -494,6 +494,7 @@ impl Cfd {
                 self.aggregated.timelocked_cet = Some(timelocked_cet);
                 self.closing_price = Some(price);
 
+                self.aggregated.commit_published = true;
```
luckysori, line 497: 👍 I think we can add this to `RefundTimelockExpired` too. Also should it also be part of the `CetTimelockExpiredPostOracleAttestation` and `CetTimelockExpiredPriorOracleAttestation`?

luckysori (approved):
Makes sense to me.

bonomat (approved):
👍

klochowicz (approved):
👍

daemon/src/projection.rs
da-kami: I would only set it in scenarios that are *directly* related to publishing `commit`. That is at least what we have been doing so far. 
`RefundTimelockExpired` sets `self.aggregated.refund_published = true;` because it is about the refund tx and only indirectly implies `commit`. But `RefundTimelockExpired` implies `CommitConfirmed` which sets `self.aggregated.commit_published = true;`.

I thought about `CetTimelockExpiredPostOracleAttestation` and `CetTimelockExpiredPriorOracleAttestation` as well, but they also don't direcly relate to publishing `commit` - they imply `CommitConfirmed` (timelock expiry can only happen if the tx made it on chain...). 
For me publishing `commit` is only directly related to `ManualCommit`, `OracleAttestedPriorCetTimelock` (in a case where `ManualCommit` is not triggered). Additionally `CommitConfirmed` is relevant for cases where the other party published `commit`. These three cases are covered now.
Note that `OracleAttestedPostCetTimelock` implies `CommitConfirmed` as well.

da-kami: We *could* opt for covering cases less strict, but that will make it hard to reason about the code I feel.

luckysori: I'm okay with your reasoning, so no need to change anything.

I find this whole `match` statement hard to follow, but I don't have a better idea a this point.

da-kami:
bors ping

da-kami:
bors retry

da-kami:
bors retry

da-kami:
bors retry

PR #1805, opened 2022-04-13
Change trigger for `add-new-pull-requests-to-project-board` action

### LINKED ISSUE #1804: Dependabot PRs don't have access to secrets
This causes our `add_pull_request_to_project` workflow to fail. For example: https://github.com/itchysats/itchysats/runs/5990696334?check_suite_focus=true.

This issue talks about the same problem: https://github.com/dependabot/dependabot-core/issues/3253. It suggests changing the event that triggers the workflow to https://github.com/itchysats/itchysats/runs/5990696334?check_suite_focus=true.

luckysori:
Fixes #1804.

Triggering this action on `pull_request_target` should grant `dependabot` access to the `secrets` environment variable. See
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target.

### DIFF
--- a/.github/workflows/add-new-pull-requests-to-project-board.yml
+++ b/.github/workflows/add-new-pull-requests-to-project-board.yml
@@ -1,7 +1,7 @@
 name: Add new pull requests to project board
 
 on:
-  pull_request:
+  pull_request_target:
     types:
       - opened
 


klochowicz (approved):
nice find!

it's a bit random that manually re-triggering that action works 😕

PR #1809, opened 2022-04-13
Wire message dispatch cleanup

da-kami:
Stability here we come ... 😬

### DIFF
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -44,9 +44,6 @@ pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
 
 const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 
-const HANDLE_CONTRACT_SETUP_TIMEOUT: Duration = Duration::from_secs(10);
-const HANDLE_PROTOCOL_MESSAGE_TIMEOUT: Duration = Duration::from_secs(60);
-
 /// The "Connected" state of our connection with the maker.
 #[allow(clippy::large_enum_variant)]
 enum State {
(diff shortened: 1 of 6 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -691,20 +691,12 @@ where
...
-                unreachable!("This kind of message should be sent to the `setup_maker::Actor`")
-            }
-            wire::TakerToMaker::Hello(_) => {
-                unreachable!("The Hello message is not sent to the cfd actor")
-            }
-            wire::TakerToMaker::HelloV2 { .. } => {
-                unreachable!("The HelloV2 message is not sent to the cfd actor")
-            }
-            wire::TakerToMaker::Unknown => {
-                // Ignore unknown message to be forwards-compatible.
```
klochowicz, line 707: why don't you want to be forwards-compatible??? you might have gone too far with combining them :)

daemon/src/maker_cfd.rs
```diff
@@ -691,20 +691,12 @@ where
                     tracing::warn!("Failed to handle rollover proposal V2: {:#}", e);
                 }
             }
-            wire::TakerToMaker::RolloverProtocol { .. } => {
-                unreachable!("This kind of message should be sent to the rollover_maker::Actor`")
-            }
-            wire::TakerToMaker::Protocol { .. } => {
-                unreachable!("This kind of message should be sent to the `setup_maker::Actor`")
```
klochowicz, line 698: nit: the previous error messages were better (at least for me), but I don't mind that much

daemon/src/maker_inc_connections.rs
```diff
@@ -435,7 +432,9 @@ impl Actor {
                             msg,
                         })
                         .await
-                        .context("`maker_cfd::Actor` is disconnected")?;
```
klochowicz, line 438: nice catch, fixing the misleading context! 👍

daemon/src/maker_inc_connections.rs
```diff
@@ -494,40 +493,39 @@ impl Actor {
 
         use wire::TakerToMaker::*;
         match msg.msg {
-            Protocol { order_id, msg } => match self.setup_actors.get_connected(&order_id) {
```
klochowicz, line 497: did you mean to remove the `get_connected()` check from here? This doesn't seem like it should be part of the changeset (if anything, you can use `send_async()` inside, after finding the appropriate setup actor, instead of dispatching to all of them).

klochowicz (changes requested):
concept ACK, a few things highlighted in the comments perhaps could be reverted - other than that, I'm keen to try it out!

daemon/src/maker_cfd.rs
luckysori, line 707: 🔧 I would keep the behaviour of the `Unknown` variant as is. We don't want to panic if we receive an unknown message.

da-kami: I think this comment was misleading. 
I think the decision to process / not process a message should *not* be taken in `maker_cfd` but in the connection handling actor, i.e. `maker_inc_connections`. That's why I moved it there and made all message that we don't care about in `maker_cfd` `unreachable`. 
Am I missing something?

daemon/src/maker_cfd.rs
```diff
@@ -691,20 +691,12 @@ where
...
-                unreachable!("The HelloV2 message is not sent to the cfd actor")
-            }
-            wire::TakerToMaker::Unknown => {
-                // Ignore unknown message to be forwards-compatible.
+            wire::TakerToMaker::RolloverProtocol { .. }
+            | wire::TakerToMaker::Protocol { .. }
+            | wire::TakerToMaker::Hello(_)
+            | wire::TakerToMaker::HelloV2 { .. }
+            | wire::TakerToMaker::Unknown => {
+                unreachable!("Message {} is not dispatched to this actor", msg.name())
```
luckysori: 🔧 We consider this unreachable, but it's still risky to have `unreachable`s in production. What do you think about this?

```rust
if cfg!(debug_assertions) {
    unreachable!("Message {} is not dispatched to this actor", msg.name())
}
```

I reckon it's safer to just ignore unexpected messages in general. We won't need to deal with this situation once we migrate to the new network layer, but that's still weeks away.

daemon/src/maker_cfd.rs
da-kami: Same as my comment above - I don't think these different `unreachable` statements helped.
Eventually we could think about moving away from the `enum`. This code should just not be in `maker_cfd` at all - it is not concerned with these messages.

daemon/src/maker_inc_connections.rs
```diff
@@ -530,8 +530,21 @@ impl Actor {
...
+            | ProposeRolloverV2 { .. }
+            | Settlement {
+                order_id: _,
+                msg: taker_to_maker::Settlement::Propose { .. },
+            } => {
+                // dispatch to the maker cfd actor
+                let _ = self.taker_msg_channel.send_async_safe(msg).await;
+            }
+            Hello(_) | HelloV2 { .. } => {
+                // No dispatching because these messages are handled when upgrading the connection
```
luckysori: 🔧 We could follow the same pattern I suggested above here.

daemon/src/maker_inc_connections.rs
da-kami: As far as I can tell this `get_connected` check will be implied when we use `send_async` on the `AddressMap` directly. This `get_connected` check here is legacy code for me. We might even be able to *not* expose `get_connected`. I can have a look at that.

daemon/src/connection.rs
```diff
@@ -439,82 +436,46 @@ impl Actor {
...
-                        tracing::warn!(%order_id, "Contract setup not started: Accepting contract setup encountered an error {e:#}")
-                    }
-                    Ok(Err(NotConnected(_))) => {
-                        tracing::warn!(%order_id, "No active contract setup")
-                    }
-                    Err(_elapsed) => {
-                        tracing::warn!(%order_id, "Contract setup not started: Contract setup took too long to process Accept")
+                    Ok(_) => {}
+                    Err(NotConnected(_)) => {
+                        tracing::error!(%order_id, "No active setup actor");
```
luckysori: 🔧 Can replace with `if let Err(e)`.

daemon/src/maker_cfd.rs
klochowicz: fair enough! I missed this, it makes sense to squash it there, instead of moving it to this actor.

daemon/src/setup_maker.rs
```diff
@@ -219,14 +219,20 @@ impl Actor {
...
+                    "Contract setup task not available, unable to dispatch contract setup message"
+                );
+                return;
+            }
+        };
 
-        Ok(())
+        if let Err(e) = sender.send(msg).await {
+            tracing::error!("Failed to dispatch contract setup message to task: {e}")
+        }
```
luckysori: 🔧 To reduce the number of times you have to explicitly handle errors, you can extract the body of this method into a non-`xtra` method on `Actor` which returns a `Result`. Then you can handle that result here with a single `if let Err(e)`.

And then you can include the whole error stack with `{e:#}`.

daemon/src/setup_maker.rs
```diff
@@ -219,14 +219,20 @@ impl Actor {
...
-        let mut sender = self
-            .setup_msg_sender
-            .clone()
-            .context("Cannot forward message to contract setup task")?;
-        sender.send(msg).await?;
+    fn handle(&mut self, msg: wire::SetupMsg) {
+        let mut sender = match self.setup_msg_sender.clone() {
+            Some(sender) => sender,
+            None => {
+                tracing::error!(
```
luckysori: 🔧 I think we want most of the logs in this PR to remain as `warn`.

daemon/src/maker_cfd.rs
klochowicz: as said above, don't mind too much with this one 🙃  it was just expressing an opinion about the contents of `unreachable` - it actually used to mention which actor handles it.

daemon/src/connection.rs
luckysori: And below.

luckysori (approved):
LGTM except for some details.

daemon/src/maker_inc_connections.rs
klochowicz: you are right! I didn't look much into `AddressMap` before, we're looking at the keys whenever we `send` as well, so that part was not needed there beforehand either....
We can remove that method altogether. :)

daemon/src/maker_cfd.rs
klochowicz: Really like the suggestion - perhaps we could go throughout our codebase and deal with it with one sweep? maybe a macro that at least logs a warning if this happens in production, and panics in debug?

klochowicz (approved):
nice one!!!

daemon/src/maker_cfd.rs
da-kami: Same as in the reply to @klochowicz  - I think the `maker_cfd` actor should *not* be in charge of this decision. This decision should be taken in `maker_inc_connections`. I think this code was actually misleading and is the result of legacy code that was not changed in `maker_inc_connections`.

daemon/src/maker_cfd.rs
luckysori: Hehe it's inspired by your idea of using `debug_assert` 😛

I think there's only one other place where we use `unreachable` in production, but I like the idea of introducing a macro that panics in debug but logs in production.

da-kami: I am happy to remove the `unreachable` here, I am not a fan of it! 
I like the proposal, in addition we could log a `warn` in production.

daemon/src/maker_cfd.rs
luckysori: I'm content with not making this a special case as long as we only call `unreachable`  on `cfg!(debug_assertions)`.

daemon/src/maker_inc_connections.rs
da-kami: I added another commit that removes the `get_connected` check because it is unused now :)

daemon/src/maker_cfd.rs
da-kami: If we do that we assume that a debug build will never encounter previous (i.e. old) messages. We can, but do we want that?

da-kami: OK with the fixup?

luckysori (approved):
Amazing! Thanks for taking care of the suggestions.

klochowicz (approved):
LGTM! Thank you for taking care of this!

PR #1810, opened 2022-04-13
Continuous listening for libp2p-xtra

klochowicz:
See commit msgs for details.

### DIFF
--- a/libp2p-xtra/src/listener.rs
+++ b/libp2p-xtra/src/listener.rs
@@ -1,0 +1,111 @@
+use crate::ConnectionStats;
+use crate::Endpoint;
+use crate::GetConnectionStats;
+use crate::ListenOn;
+use anyhow::anyhow;
+use anyhow::Result;
+use async_trait::async_trait;
+use libp2p_core::Multiaddr;
+use std::time::Duration;
+use tokio_tasks::Tasks;
+use xtra::Address;
+use xtra_productivity::xtra_productivity;
+use xtras::SendInterval;
+
+/// How often we check whether we are still connected
+pub const CHECK_INTERVAL: Duration = Duration::from_secs(5);
+
+/// xtra actor that taker care of listening for incoming connections to the Endpoint.
+///
+/// Periodically polls Endpoint to check whether connection is still active.
+/// Should be used in conjunction with supervisor for continuous and resilient listening.
+pub struct Actor {
+    tasks: Tasks,
+    endpoint: Address<Endpoint>,
+    listen_address: Multiaddr,
+    /// Contains the reason we are stopping.
+    stop_reason: Option<Error>,
+}
+
+impl Actor {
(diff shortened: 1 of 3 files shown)


libp2p-xtra/src/listener.rs
```diff
@@ -0,0 +1,120 @@
...
+                    tracing::error!("Cannot listen on the endpoint");
+                    let _: Result<(), Disconnected> = this.send(e).await;
+                },
+            );
+        };
+
+        // Only start checking the connection after it had enough time to be established
+        tokio::time::sleep(CHECK_INTERVAL).await;
+        self.tasks
+            .add(this.send_interval(CHECK_INTERVAL, || CheckConnection));
```
luckysori: ❓ What about this?

```suggestion
        if let Err(e) = endpoint.send(ListenOn(listen_address)).await {
            self.stop_reason = ErrorFailed { source: anyhow!(e) };
            ctx.stop()
        };
    
        self.tasks
            .add(this.send_interval(CHECK_INTERVAL, || CheckConnection));
```

I think we can stop immediately if we can't listen on the endpoint. That way we don't have to introduce the `sleep` and the logic is simpler.

libp2p-xtra/src/listener.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[error("Endpoint actor is disconnected")]
+    NoEndpoint,
+    #[error("Connection dropped from endpoint")]
+    ConnectionDropped,
+    #[error("Stop reason was not specified")]
+    Unspecified,
+}
+
+#[derive(Clone, Copy)]
+pub struct CheckConnection;
```
luckysori: 🔧 I think these don't need to be `pub`.

luckysori (approved):
LGTM! Just a couple of things.

libp2p-xtra/src/listener.rs
klochowicz: `Error` has to be public, for supervisor needs. The other one doesn't, fixed!

libp2p-xtra/src/listener.rs
klochowicz: we still need `sleep` because the Endpoint takes a while for it to register that, it's not immediately available :(

I reworked the other part, it is simpler indeed!

luckysori: Shame about the sleep! I wonder if sending `ListenOn` should only return after the `Endpoint` is ready to listen. That may not be possible though.

Nice otherwise!

klochowicz: Another thing that I could think of would be setting it up immediately in the hashset in `Endpoint`. This was indeed the original behaviour, but somehow we changed it to be more correct. 
In other words, we could redefine from what it means to be present in the hashset to: "listening or listening requested". It would still be removed if the request failed, so I guess it's more ergonomic for users, especially when polling is the way.

Thoughts?

PR #1816, opened 2022-04-14
Add metrics for position size

bonomat:
Concept ACK request: we want to move metrics from maker-metrics into the daemon for simplicity reasons and reduction of duplicating logic. 
The question is, "where" in the daemon should we derive the metrics from. 

If concept ack I'll remove some duplication and make the metrics function a bit nicer. 

Looks like this. 

```
# HELP position_size_quantity Total quantity of positions on itchsyats.
# TYPE position_size_quantity gauge
position_size_quantity{direction="long",status="closed"} 300
position_size_quantity{direction="long",status="failed"} 0
position_size_quantity{direction="long",status="open"} 0
position_size_quantity{direction="short",status="closed"} 700
position_size_quantity{direction="short",status="failed"} 0
position_size_quantity{direction="short",status="open"} 0
```

### DIFF
--- a/daemon/src/cfd_metrics.rs
+++ b/daemon/src/cfd_metrics.rs
@@ -1,0 +1,116 @@
+use crate::projection::Cfd;
+use model::Position;
+use model::Usd;
+use rust_decimal::prelude::ToPrimitive;
+use std::collections::HashMap;
+
+const DIRECTION_LABEL: &str = "direction";
+const DIRECTION_LONG_LABEL: &str = "long";
+const DIRECTION_SHORT_LABEL: &str = "short";
+const STATUS_LABEL: &str = "status";
+const STATUS_OPEN_LABEL: &str = "open";
+const STATUS_CLOSED_LABEL: &str = "closed";
+const STATUS_FAILED_LABEL: &str = "failed";
+
+static POSITION_SIZE_GAUGE: conquer_once::Lazy<prometheus::GaugeVec> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_gauge_vec!(
+            "position_size_quantity",
+            "Total quantity of positions on itchsyats.",
+            &[DIRECTION_LABEL, STATUS_LABEL]
+        )
+        .unwrap()
+    });
+
+pub(crate) fn update_position_metrics(cfds: &[Cfd]) {
+    {
+        let open_cfds = cfds.iter().filter(|cfd| cfd.is_open());
+        let (open_long, open_short): (Vec<_>, Vec<_>) =
+            open_cfds.partition(|cfd| cfd.position == Position::Long);
+
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_LONG_LABEL),
+                (STATUS_LABEL, STATUS_OPEN_LABEL),
+            ]))
+            .set(
+                sum_amounts(&open_long)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_SHORT_LABEL),
+                (STATUS_LABEL, STATUS_OPEN_LABEL),
+            ]))
+            .set(
+                sum_amounts(&open_short)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+    }
+
+    {
+        let closed_cfds = cfds.iter().filter(|cfd| cfd.is_closed());
+        let (closed_long, closed_short): (Vec<_>, Vec<_>) =
+            closed_cfds.partition(|cfd| cfd.position == Position::Long);
+
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_LONG_LABEL),
+                (STATUS_LABEL, STATUS_CLOSED_LABEL),
+            ]))
+            .set(
+                sum_amounts(&closed_long)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_SHORT_LABEL),
+                (STATUS_LABEL, STATUS_CLOSED_LABEL),
+            ]))
+            .set(
+                sum_amounts(&closed_short)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+    }
+    {
+        let failed_cfds = cfds.iter().filter(|cfd| cfd.is_failed());
+        let (failed_long, failed_short): (Vec<_>, Vec<_>) =
+            failed_cfds.partition(|cfd| cfd.position == Position::Long);
+
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_LONG_LABEL),
+                (STATUS_LABEL, STATUS_FAILED_LABEL),
+            ]))
+            .set(
+                sum_amounts(&failed_long)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+        POSITION_SIZE_GAUGE
+            .with(&HashMap::from([
+                (DIRECTION_LABEL, DIRECTION_SHORT_LABEL),
+                (STATUS_LABEL, STATUS_FAILED_LABEL),
+            ]))
+            .set(
+                sum_amounts(&failed_short)
+                    .into_decimal()
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+    }
+}
+
+fn sum_amounts(cfds: &[&Cfd]) -> Usd {
+    cfds.iter()
+        .fold(Usd::ZERO, |sum, cfd| cfd.quantity_usd + sum)
+}
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -751,6 +751,18 @@ impl Cfd {
 
         in_open_state && not_expired
     }
+
+    fn is_closed(&self) -> bool {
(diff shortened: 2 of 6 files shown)


daemon/src/cfd_metrics.rs
```diff
@@ -0,0 +1,116 @@
+use crate::projection::Cfd;
+use model::Position;
+use model::Usd;
+use rust_decimal::prelude::ToPrimitive;
+use std::collections::HashMap;
+
+const DIRECTION_LABEL: &str = "direction";
```
da-kami, line 7: 🙃 `Direction` as term does not exist in the `itchysats` repo. I know we started using it in `metrics` and `automation`. I think we should agree on what terminology to use to avoid confusion in the future. In `itchysats` it is called `Position`. Is that not correct?

daemon/src/projection.rs
```diff
@@ -729,6 +729,28 @@ impl Cfd {
 
         Some(url)
     }
+
+    fn is_open(&self) -> bool {
```
da-kami: 🛠 This commit seems to introduce a function that is not used in code, but only in the tests. Is this commit incomplete or is this just to showcase what you have planned?

daemon/src/projection.rs
```diff
@@ -751,6 +751,18 @@ impl Cfd {
 
         in_open_state && not_expired
     }
+
+    fn is_closed(&self) -> bool {
```
da-kami: ❓

So basically you are mapping the (aggregated) `projection::CfdState` to an "more aggregated" state of `is_open`, `is_closed` and `is_failed`. Why are we doing this based on the projection states and not the events? 
We *can* do this, but I am not sure it's a good idea. I feel we add one more layer of mapping.

That goes back to the discussion if we should keep the `projection::CfdState` or change / remove it. How do you see this evolve?

---

Are you planning to expose the new "states" (open, closed, failed) on the API?
So far they are only used in tests.

da-kami (commented):
It would have helped to have some kind of goal / why for this PR. Since there is no ticket to read through it's hard to understand the motivation of "if concept is ack...".

My understanding is, that this is an effort to start moving metrics into the daemon. But from what I understand this PR does not actually define the API, but how to create and update metric internally. 

I am not sure about the relation of projection and metrics. Maybe it would actually be better to treat metrics separately, create another Aggregate for them and then derive them from the events.
I am not sure if `internal event model -> projection aggregate -> cfd states -> metrics` will help us in the future. How do you see this evolve? 
Maybe we should discuss how we want to evolve the projection first. If it is urgent to move the metrics, and we *want* to base them on `projection::CfdState` we can move forward with this, but it might only be a temporary solution.

daemon/src/projection.rs
bonomat: I tried to explain it in the commit message, not sure if you saw it: `This will come handy later on when creating metrics.` 
Meaning, this is a separate commit introducing this function which will be used for the metrics later.

daemon/src/projection.rs
bonomat: Yes, the idea is to have a simple function which allows us to determine if a position is `open`, `closed` or `failed`. 

> Why are we doing this based on the projection states and not the events?

Do you think based on the events is better?

> Are you planning to expose the new "states" (open, closed, failed) on the API?

No, it should only be used for the metrics.

daemon/src/cfd_metrics.rs
bonomat: Happy to go with `position` :)

bonomat:
> It would have helped to have some kind of goal / why for this PR. Since there is no ticket to read through it's hard to understand the motivation of "if concept is ack...".
> 

Sorry, I thought it was clear: the concept ack question was about where to place the metric generation. 

> My understanding is, that this is an effort to start moving metrics into the daemon. But from what I understand this PR does not actually define the API, but how to create and update metric internally.
> 
> I am not sure about the relation of projection and metrics. Maybe it would actually be better to treat metrics separately, create another Aggregate for them and then derive them from the events. I am not sure if `internal event model -> projection aggregate -> cfd states -> metrics` will help us in the future. How do you see this evolve? Maybe we should discuss how we want to evolve the projection first. If it is urgent to move the metrics, and we _want_ to base them on `projection::CfdState` we can move forward with this, but it might only be a temporary solution.

My thinking was that the relation of `projection` and `metrics` is very similar. They should never diverge, otherwise we will get confused if we look at the board vs the maker-frontend. 
Hence, I thought deriving it from the projection makes sense. I'm happy to consider otherwise.

da-kami:
> > [snip]
> > I am not sure about the relation of projection and metrics. Maybe it would actually be better to treat metrics separately, create another Aggregate for them and then derive them from the events. I am not sure if `internal event model -> projection aggregate -> cfd states -> metrics` will help us in the future. How do you see this evolve? Maybe we should discuss how we want to evolve the projection first. If it is urgent to move the metrics, and we _want_ to base them on `projection::CfdState` we can move forward with this, but it might only be a temporary solution.
> 
> My thinking was that the relation of `projection` and `metrics` is very similar. They should never diverge, otherwise we will get confused if we look at the board vs the maker-frontend. 

I do see your point, but I still feel we are plugging the metrics on top of the projection without being sure about how to evolve the projection. With tying this together it might become even harder to change / remove `projection::CfdState` but also e.g. creating separate projection(s) for maker and taker because they might have different needs.

> Hence, I thought deriving it from the projection makes sense. I'm happy to consider otherwise.

Let's wait for some other opinions. Currently I am not sure what is "better" - separating metrics from projection or plugging it on top. I just voiced some concerns so we are aware of them :)

daemon/src/projection.rs
da-kami: > [snip]
> > Are you planning to expose the new "states" (open, closed, failed) on the API?

> No, it should only be used for the metrics.

How are you planning to expose the metrics? As another route that then pulls them from the projection?

bonomat:
Maybe worth mentioning: the current state is that we derive the metrics from what the project provides, i.e. we poll the HTTP API and create our metrics based on that.

daemon/src/projection.rs
bonomat: The maker daemon has a metric endpoint under `api/metrics`: 
The _source_ of this endpoint is managed by prometheus and used here: 
https://github.com/itchysats/itchysats/blob/ec30d46f04caaa7cec3dc366e21598545c75dd53/maker/src/routes.rs#L275-L286

bonomat: Maybe worth to mention: the metrics are static variables stored in a global registry. `prometheus.gather()` knows how to find them.

da-kami: Thanks for pointing me to that, I think I never looked into this code in detail. I will have another broader look later to come up with a more complete conclusion.

daemon/src/projection.rs
```diff
@@ -729,6 +729,28 @@ impl Cfd {
...
+            CfdState::OutgoingRolloverProposal,
+            CfdState::OutgoingSettlementProposal,
+            CfdState::RolloverSetup,
+        ]
+        .contains(&self.state);
+
+        let not_expired = match self.expiry_timestamp {
+            None => true,
+            Some(expiry_timestamp) => expiry_timestamp > OffsetDateTime::now_utc(),
+        };
```
luckysori, line 751: 🤔 I suppose we've had this logic elsewhere before, but I'm not sure I agree.

The (confusingly named) `expiry_timestamp` corresponds to the time at which the oracle is expected to attest to an event that would allow non-collaborative settlement. Even if we have reached this time:

- The oracle could fail to attest in time.
- The commit transaction could take a while to be confirmed.

I don't see why we need to use this value to reason about the openness of the CFD

luckysori (commented):
I hear the concern about how to calculate these specific metrics. Perhaps a dedicated aggregate for the `maker` would be cleaner, but I'm not totally against this approach.

daemon/src/projection.rs
bonomat: Yea, we use the same logic in other places outside of this project :)

Now that I think about it a bit more, for metrics this is actually not that important. It is only important for out automation. 
We should derive `is_closed` only from the states... 

Thanks for pointing this out 🙏

daemon/src/projection.rs
```diff
@@ -790,8 +791,9 @@ impl Tx {
                     &a.aggregated.creation_timestamp,
                 )
             })
-            .collect();
+            .collect::<Vec<_>>();
 
+        update_position_metrics(cfds_with_quote.as_slice());
```
thomaseizinger, line 796: `projection` was designed to be the read model that is exposed to the UI. Retro-fitting metrics into here is not ideal.

It shouldn't be too difficult to create a separate actor that periodically loads all CFDs and exposes these metrics. That way, you can extract exactly the information you want and don't have to re-purpose the `projection::Cfd` model.

Reasoning from an existing read-model is likely to produce all sorts of bugs.

I am worried that adding this here will make `projection` significantly harder to maintain as it its model serves yet another purpose now.

thomaseizinger: The problem with a separate actor is that we can't update the metrics in real-time. However, is is more a limitation of the fact that our ES implementation doesn't have event subscriptions (i.e. notifying actors about a new event).

At the moment, we circumvent that by notifying `projection` explicitly at end of `process_manager`. We could easily to the same thing for a `metrics` actor and thus avoid loading things periodically.

daemon/src/projection.rs
```diff
@@ -729,6 +730,40 @@ impl Cfd {
...
+            CfdState::PendingOpen,
+            CfdState::Open,
+            CfdState::PendingCommit,
+            CfdState::OpenCommitted,
+            CfdState::IncomingRolloverProposal,
+            CfdState::IncomingSettlementProposal,
+            CfdState::OutgoingRolloverProposal,
+            CfdState::OutgoingSettlementProposal,
+            CfdState::RolloverSetup,
+        ]
```
thomaseizinger, line 745: Not a big fan of this at all. These states only exist because they were originally exposed on the API and used for the internal model. They _only_ exist in the projection layer for backwards-compatibility reasons.

I think we should take the opportunity when exposing first-class metrics from the daemon and reason about the desired state based on the actual source-of-truth which are the events.

bonomat:
superseded by #1921

PR #1818, opened 2022-04-14
Provide context when failing to load CFDs from DB

luckysori:


luckysori:
This is a trivial change to get more context for debugging, so I'm merging this without approvals.

PR #1819, opened 2022-04-14
Run move-to-close-cfds-table futures sequentially 

luckysori:
Our continuous testing has revealed the following error since #1779:

> Failed to move closed CFDs to corresponding table: error returned from database: database is locked: database is locked

This is an attempt at dealing with this problem, recorded here: #1820.


luckysori:
Hoping to solve #1820 with this, or at least get more context.

PR #1824, opened 2022-04-15
Generate ed25519 keypair from the seed

klochowicz:
Pass it trough to the actor systems.

Unfortunately, it turned out to be impossible to derive peer id from maker id,
so we will need an additional CLI option on the taker side.
After this gets merged, we'll be able to easily add a default argument for the
taker based on maker logs.

### DIFF
--- a/daemon/src/seed.rs
+++ b/daemon/src/seed.rs
@@ -3,10 +3,29 @@ use anyhow::Result;
 use bdk::bitcoin::util::bip32::ExtendedPrivKey;
 use bdk::bitcoin::Network;
 use hkdf::Hkdf;
+use libp2p_core::identity::ed25519;
+use libp2p_core::identity::Keypair;
+use libp2p_core::PeerId;
 use rand::Rng;
 use sha2::Sha256;
 use std::convert::TryInto;
 use std::path::Path;
+
+/// Struct containing keys for both legacy and libp2p connections.
+///
+/// It is located here as all the information is derived from the seed.
+#[derive(Clone)]
+pub struct Identities {
+    pub identity_sk: x25519_dalek::StaticSecret,
+    pub identity_pk: x25519_dalek::PublicKey,
+    pub libp2p: Keypair,
+}
+
+impl Identities {
+    pub fn peer_id(&self) -> PeerId {
+        self.libp2p.public().to_peer_id()
+    }
+}
 
 pub trait Seed {
     fn seed(&self) -> Vec<u8>;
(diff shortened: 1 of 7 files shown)


maker/src/main.rs
```diff
@@ -191,11 +191,13 @@ async fn main() -> Result<()> {
     let auth_username = rocket_basicauth::Username("itchysats");
     let auth_password = seed.derive_auth_password::<rocket_basicauth::Password>();
 
-    let (identity_pk, identity_sk) = seed.derive_identity();
+    let keypair = seed.derive_identities();
```
da-kami: ❓Why was this renamed to `keypair`? I feel identities still expresses better what this is.

maker/src/main.rs
```diff
@@ -191,11 +191,13 @@ async fn main() -> Result<()> {
...
 
-    let hex_pk = hex::encode(identity_pk.to_bytes());
+    let peer_id = keypair.libp2p.public().to_peer_id();
```
da-kami: 🙃 Could also hide the way we access the internals and just have `identities.peer_id()` here.

da-kami (approved):

daemon/src/seed.rs
```diff
@@ -43,6 +56,34 @@ pub trait Seed {
...
+    fn derive_ed25519_keypair(&self) -> Result<ed25519::Keypair> {
+        let mut secret = [0u8; 32];
+
+        Hkdf::<Sha256>::new(None, &self.seed())
+            .expand(b"LIBP2P_IDENTITY", &mut secret)
+            .expect("okm array is of correct length");
+
+        let keypair = ed25519::Keypair::from(
+            ed25519::SecretKey::from_bytes(secret)
+                .context("Failed to create secret key from random bytes")?,
```
luckysori: 🔧 I think this can be an `expect`. 

`SecretKey::from_bytes` only fails if we don't pass exactly 32 bytes, and the hash generated by `SHA256` is 32 bytes.

maker/src/main.rs
luckysori: It seems like destructuring would make sense here (but yes, "keypair" isn't quite right).

daemon/src/seed.rs
```diff
@@ -1,13 +1,26 @@
...
 use bdk::bitcoin::Network;
 use hkdf::Hkdf;
+use libp2p_core::identity::ed25519;
+use libp2p_core::identity::Keypair;
 use rand::Rng;
 use sha2::Sha256;
 use std::convert::TryInto;
 use std::path::Path;
 
+/// Struct containing secrets for both legacy and libp2p connections.
```
luckysori: 🙃 
```suggestion
/// Struct containing keys for both legacy and libp2p connections.
```

luckysori (approved):

maker/src/main.rs
klochowicz: great suggestion!

maker/src/main.rs
klochowicz: good catch! was meant to be renamed to identities indeed. Unfortunately, can't destructure here as it's being passed to MakerActorSystem

daemon/src/seed.rs
klochowicz: sure!

PR #1825, opened 2022-04-15
Add continuous connecting support for libp2p-xtra

klochowicz:
Another piece of the puzzle that can re-establish connections when things go down for whatever reason.

### DIFF
--- a/libp2p-xtra/examples/hello_world_dialer.rs
+++ b/libp2p-xtra/examples/hello_world_dialer.rs
@@ -8,13 +8,14 @@ use libp2p_core::identity::Keypair;
 use libp2p_core::Multiaddr;
 use libp2p_core::PeerId;
 use libp2p_tcp::TokioTcpConfig;
-use libp2p_xtra::Connect;
+use libp2p_xtra::dialer;
 use libp2p_xtra::Endpoint;
 use libp2p_xtra::OpenSubstream;
 use std::time::Duration;
 use tokio::time::sleep;
 use xtra::prelude::*;
 use xtra::spawn::TokioGlobalSpawnExt;
+use xtras::supervisor;
 
 #[derive(Parser)]
 struct Opts {
@@ -37,11 +38,17 @@ async fn main() -> Result<()> {
         .create(None)
         .spawn_global();
 
-    endpoint_addr
-        .send(Connect(opts.multiaddr.clone()))
-        .await
-        .unwrap()
-        .unwrap();
+    let dialer_constructor = {
+        let connect_addr = opts.multiaddr.clone();
+        let endpoint_addr = endpoint_addr.clone();
+        move || dialer::Actor::new(endpoint_addr.clone(), connect_addr.clone())
+    };
+
+    let (supervisor, _dialer_actor) = supervisor::Actor::with_policy(
+        dialer_constructor,
+        |_: &dialer::Error| true, // always restart dialer actor
+    );
+    let _dialer_supervisor = supervisor.create(None).spawn_global();
 
     sleep(Duration::from_secs(1)).await;
 
--- a/libp2p-xtra/src/dialer.rs
+++ b/libp2p-xtra/src/dialer.rs
@@ -1,0 +1,115 @@
+use crate::Connect;
+use crate::ConnectionStats;
+use crate::Endpoint;
+use crate::GetConnectionStats;
+use anyhow::anyhow;
+use async_trait::async_trait;
+use libp2p_core::Multiaddr;
+use std::time::Duration;
+use tokio_tasks::Tasks;
+use xtra::Address;
+use xtra_productivity::xtra_productivity;
+use xtras::SendInterval;
+
+/// How often we check whether we are still connected
+pub const CHECK_INTERVAL: Duration = Duration::from_secs(5);
+
+/// xtra actor that takes care of dialing (connecting) to an Endpoint.
+///
+/// Periodically polls Endpoint to check whether connection is still active.
+/// Should be used in conjunction with supervisor maintaining resilient connection.
+pub struct Actor {
+    tasks: Tasks,
+    endpoint: Address<Endpoint>,
+    connect_address: Multiaddr,
+    stop_reason: Option<Error>,
+}
+
+impl Actor {
+    pub fn new(endpoint: Address<Endpoint>, connect_address: Multiaddr) -> Self {
+        Self {
+            tasks: Tasks::default(),
+            endpoint,
+            connect_address,
+            stop_reason: None,
+        }
+    }
+
+    /// Returns error if we cannot access Endpoint or the connect address is not
+    /// present inside Endpoint.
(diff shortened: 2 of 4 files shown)


libp2p-xtra/examples/hello_world_dialer.rs
```diff
@@ -37,11 +38,22 @@ async fn main() -> Result<()> {
         .create(None)
         .spawn_global();
 
-    endpoint_addr
-        .send(Connect(opts.multiaddr.clone()))
-        .await
-        .unwrap()
-        .unwrap();
+    let dialer_constructor = {
+        // XXX: Can this be done nicer???
```
da-kami: Is this still an open `TODO`?

libp2p-xtra/examples/hello_world_dialer.rs
```diff
@@ -37,11 +39,22 @@ async fn main() -> Result<()> {
...
+        let connect_addr = opts.multiaddr.clone();
+        let endpoint_addr = endpoint_addr.clone();
+        move || {
+            let connect_addr = connect_addr.clone();
+            let endpoint_addr = endpoint_addr.clone();
+            dialer::Actor::new(endpoint_addr, connect_addr)
+        }
+    };
```
luckysori: > Can this be done nicer?

I don't think so. I would inline the second set of clones though, but it's just style.

luckysori (approved):
LGTM!

libp2p-xtra/examples/hello_world_dialer.rs
```diff
@@ -37,11 +39,22 @@ async fn main() -> Result<()> {
...
+
+    let (supervisor, _dialer_actor) = supervisor::Actor::with_policy(
+        dialer_constructor,
+        |_: &dialer::Error| true, // always restart dialer actor
```
da-kami, line 49: 🤔 I am a bit confused by the comment `// always restart dialer actor`. As far as I understand,  with this restart policy we don't always restart, but only when the actor emits a`dialer::Error` - is that the only error that is emitted?

libp2p-xtra/src/dialer.rs
```diff
@@ -0,0 +1,114 @@
...
+
+/// xtra actor that takes care of dialing (connecting) to an Endpoint.
+///
+/// Periodically polls Endpoint to check whether connection is still active.
+/// Should be used in conjunction with supervisor maintaining resilient connection.
+pub struct Actor {
+    tasks: Tasks,
+    endpoint: Address<Endpoint>,
+    connect_address: Multiaddr,
+    /// Contains the reason we are stopping.
```
da-kami:
```suggestion
```

I don't feel this comment adds much value without adding a more detailed description after the sentence. Like this I feel the name explains it and the comment is redundant.

libp2p-xtra/src/dialer.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub fn new(endpoint: Address<Endpoint>, connect_address: Multiaddr) -> Self {
+        Self {
+            tasks: Tasks::default(),
+            endpoint,
+            connect_address,
+            stop_reason: None,
+        }
+    }
+
+    async fn check_connection_active_in_endpoint(&self) -> Result<(), Error> {
```
da-kami, line 40: I am having a hard time following the code in here.
Let me try to summarize what I understand to validate that is is correct:

1. We use a supervisor actor to restart the `dialer::Actor` in case the actor emits a `dialer::Error`.
2. Inside the `dialer::Actor` we monitor the connectionby sending `CheckConnection`  every `5` seconds

So, if handling `CheckConnection` in `check_connection_active_in_endpoint` fails we set the stop reason to `ConnectionDropped` and then stop the actor, which results in the actor being restarted by the supervisor. That's it, right?

I think it would help me to inline the code of `check_connection_active_in_endpoint` in the handler so there is a bit less indirection.

da-kami (approved):

libp2p-xtra/examples/hello_world_dialer.rs
klochowicz: this was a comment to the reviewers whether this can be done nicer. @luckysori said that it can't, so I'm removing it. :)

libp2p-xtra/examples/hello_world_dialer.rs
klochowicz: happy to inline the second set of clones, makes it *a bit* nicer. thanks for looking at this bit :)

libp2p-xtra/src/dialer.rs
klochowicz: sure, happy to remove it. Was just following a pattern seen in other actors.

libp2p-xtra/examples/hello_world_dialer.rs
klochowicz: yes, this is the only error that is emitted! do you think it would be less confusing if I just create a supervisor actor without any explicit policy? (I believe it will then restart on *all* the errors?)

libp2p-xtra/src/dialer.rs
klochowicz: Yes, exactly!

Unfortunately, inlining messes up error handling. I've added a comment, hopefully it will be a bit easier to understand with it.

libp2p-xtra/examples/hello_world_dialer.rs
luckysori: Huh, I thought the type hint was superfluous. The `dialer::Actor` can _only_ emit a `dialer::Error`, so that's the only thing the `supervisor::Actor` can expect here. Therefore, this actor will always be restarted. Any flaws in my logic?

klochowicz: unfortunately, it seems that if we want actors to be flexible, we cannot use the simple constructor anymore :( might look into the supervisor later on.

luckysori: > unfortunately, it seems that if we want actors to be flexible, we cannot use the simple constructor anymore :( might look into the supervisor later on.

What do you mean by flexible?

> Huh, I thought the type hint was superfluous. The dialer::Actor can only emit a dialer::Error, so that's the only thing the supervisor::Actor can expect here. Therefore, this actor will always be restarted. Any flaws in my logic?

I was wrong in the sense that the type hint is needed, but it's true that the error emitted by the supervised actor when stopping is predictable (and always the same).

klochowicz: flexible - meaning that one can use a supervisor with a custom policy (turns out if we define a custom error, we can't use `new()`, we have to rely on `with_policy()` - and the decision should be relying on the users of the API, not baked inside particular actors.

yup, type hint is needed indeed 😮‍💨 Errors are different, depending on the way the actor stops. the consumers of this API could easily say "don't try to restart if you've been disconnected" or perhaps "wait for X seconds before you attempt to re-dial when you see such-or-such error. 
In this example, it just restarts straight away.

PR #1845, opened 2022-04-19
Verify effect on events after moving CFDs to closed

luckysori:
This PR just makes our new `db` tests a bit more complete.

After a successful move from `cfs` to `closed_cfds`, the `events` table should have no entries for the corresponding CFD.

Conversely, after an unsuccessful move, the `events` table should have the same number of events for the corresponding CFD.


luckysori:
This just makes the tests better, so I'm confident to merge without approvals.

bors r+

PR #1846, opened 2022-04-19
Not enough balance improvement

### LINKED ISSUE #1813: Improve the "Your balance is too low" message in the UI
From user feedback:

1. It is somewhat unclear that the message `Your balance is too low. Please deposit more into your wallet` is related to opening a new position
2. It is unclear where the wallet can be found (note: This is only somewhat related, for more elaborate solution I will open a separate ticket)

Proposed solution:

1. Change message to: `Not enough balance to open a new position, please deposit more`
2. Make the "warning" a link to the wallet

da-kami:
fixes https://github.com/itchysats/itchysats/issues/1813

image

Note: I don't specifically like the word `position` in this context. You add to / reduce from your position, not "open a new one", but I did not find a better way to phrase this. Suggestions welcome.

### DIFF
--- a/taker-frontend/src/components/AlertBox.tsx
+++ b/taker-frontend/src/components/AlertBox.tsx
@@ -1,18 +1,34 @@
-import { Alert, AlertDescription, AlertIcon, AlertStatus, AlertTitle } from "@chakra-ui/react";
+import { ExternalLinkIcon } from "@chakra-ui/icons";
+import { Alert, AlertDescription, AlertIcon, AlertStatus, AlertTitle, Link } from "@chakra-ui/react";
 import * as React from "react";
+import { Link as ReachLink } from "react-router-dom";
 
 interface AlertBoxProps {
     title: string;
     description: string;
     status?: AlertStatus;
+    reachLinkTo?: string;
 }
 
-export default function AlertBox({ title, description, status = "error" }: AlertBoxProps) {
+export default function AlertBox({ title, description, status = "error", reachLinkTo }: AlertBoxProps) {
     return (
         <Alert status={status}>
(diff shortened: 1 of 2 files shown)


luckysori (approved):
I don't mind the word position, given that we don't have combined positions yet.

klochowicz (approved):
LGTM!

taker-frontend/src/components/AlertBox.tsx
```diff
@@ -1,18 +1,34 @@
...
 }
 
-export default function AlertBox({ title, description, status = "error" }: AlertBoxProps) {
+export default function AlertBox({ title, description, status = "error", reachLinkTo }: AlertBoxProps) {
     return (
         <Alert status={status}>
             <AlertIcon />
             <AlertTitle mr={2}>{title}</AlertTitle>
-            <AlertDescription>{description}</AlertDescription>
+            {reachLinkTo
```
bonomat, line 18: An alternative option would be to use `{children}` here and let the user of the component decide on what additional elements should be added to the alert box description. 
E.g. like in here: https://github.com/itchysats/itchysats/blob/f3617e51e088265bc4fd891bb9d2a5ff3dbae1b5/taker-frontend/src/components/FundingRateTooltip.tsx#L5-L25

bonomat (approved):
👍

taker-frontend/src/components/AlertBox.tsx
da-kami: yeah, I totally thought about this, but was not sure if I can achieve exactly what I wanted so I did not invest time for now.
Happy to see it in a follow-up.

PR #1849, opened 2022-04-19
Rename libp2p-xtra to xtra-libp2p

### LINKED ISSUE #1785: Rename `libp2p-xtra` to `xtra-libp2p`
The naming pattern in our workspace for actors is to name the `xtra` bit first and everything else after. `libp2p-xtra` does not follow this naming pattern but it should!

klochowicz:
Fixes #1785


luckysori (approved):
👍

PR #1852, opened 2022-04-19
Remove CFDs moved to `closed_cfds` table from aggregate cache

### LINKED ISSUE #1851: Remove CFD from DB cache after moving to `closed_cfds` table

luckysori:
Fixes #1851.

It's a draft because I'm not convinced we need this and I don't love the way the solution is implemented.

Even if we "wrongly" load a CFD in `closed_cfds` from the aggregate cache, nothing bad should happen. This is similar to loading a closed CFD in `cfds`. The difference is that for the former we would be unable to append new superfluous events to the database, and might log some misleading warnings.

The downsides of the changes in this PR are:

- We cannot programmatically ensure that we are removing from all caches (if we add new ones in the future).
- We have to make public types that were implementation details of certain modules.

Perhaps there's a better way to approach this, because cache invalidation is a [famously hard](https://www.martinfowler.com/bliki/TwoHardThings.html) problem.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -438,6 +438,8 @@ impl Connection {
 
                 db_tx.commit().await?;
 
+                self.remove_from_all_aggregate_caches(id).await;
+
                 anyhow::Ok(())
             };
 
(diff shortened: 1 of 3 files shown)


da-kami:
After some more thought I think we don't need this. Happy with closing this PR. If we ever encounter problems with the cache not being cleared in the future I would tackle it then.

klochowicz:
> After some more thought I think we don't need this. Happy with closing this PR. If we ever encounter problems with the cache not being cleared in the future I would tackle it then.

same. I don't think this will cause any issues for us as it currently stands.

PR #1856, opened 2022-04-20
Ignore CFDs that were moved to closed while processing all open CFDs

### LINKED ISSUE #1847: Handle loading errors if a CFD is no longer open (but the actor still thinks it is)
Closing CFDs is a separate task can lead to moving a CFD while another actor thinks it is still open.

We have three actors that use `load_open_cfd_ids` and then delegate to other actors to load the CFD:

1. `monitor::Actor`
2. `oracle::Actor`
3. `auto_rollover::Actor`

Once we have the a list of `open` cfd `id`s we cannot be sure that the CFD can actually be loaded, because it might have been moved to `closed_cfds` after the list was loaded.
For e.g. the `oracle` actor it is likely to run into a loading problem, because we do somewhat long-running tasks (decrypt CET) after we load the CFD ids.

This is not an easy problem to solve. I think we don't want to go into a direction where we give some tasks priority over each other, that would bring another set of problems.
After a discussion with @luckysori  we came to the conclusion that the best way forward is for each actor that loads from the database to be resilient against loading errors. Additionally we might want to have a specific error for "CFD was moved to closed"  that is returned on `load_open_cfd`.
The way I imagine this is: If the CFD cannot be found in `cfds`, check if it can be found in `closed_cfds`, if so return a specific error that the caller can deal with / ignore.

da-kami:
fixes #1847 

Due to loading CFD ids and processing them one by one ew can encounter situations where the task that moves CFDs to the `closed_cfds` table moves a CFD while another actor still processes this CFD as part of the open CFD id list.
We introduce a `db::Error` that models all errors, so we can distinguish the `OpenCfdNotFound` case.
When loading all open CFDs we are ignoring cases where a CFD cannot be found in `cfds` (the table representing open CFDs) and print a trace log.
We don't handle `load_open_cfd` error cases of `OpenCfdNotFound` separately because it is not expected to call this function and fail (because the CFD should already be final if it was moved to `closed_cfds`).

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -222,7 +222,7 @@ impl Connection {
     }
 
     /// Load a CFD in its latest version from the database.
-    pub async fn load_open_cfd<C>(&self, id: OrderId, args: C::CtorArgs) -> Result<C>
+    pub async fn load_open_cfd<C>(&self, id: OrderId, args: C::CtorArgs) -> Result<C, Error>
     where
         C: CfdAggregate,
     {
@@ -349,7 +349,16 @@ impl Connection {
             let ids = self.load_open_cfd_ids().await?;
 
             for id in ids {
-                let cfd = self.load_open_cfd(id, args.clone()).await?;
+                let cfd = match self.load_open_cfd(id, args.clone()).await {
+                    Ok(cfd) => Ok(cfd),
+                    Err(Error::OpenCfdNotFound) => {
+                        tracing::trace!(order_id=%id, target="db", "Ignoring open CFD not found because it was likely moved to the closed table");
+                        continue;
+                    },
+                    Err(e) => {
+                        Err(e)
+                    }
+                }?;
 
                 yield cfd;
             }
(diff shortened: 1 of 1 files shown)


daemon/src/db.rs
```diff
@@ -349,7 +349,17 @@ impl Connection {
             let ids = self.load_open_cfd_ids().await?;
 
             for id in ids {
-                let cfd = self.load_open_cfd(id, args.clone()).await?;
+                let cfd = match self.load_open_cfd(id, args.clone()).await {
+                    Ok(cfd) => Ok(cfd),
+                    Err(Error::OpenCfdNotFound) => {
+                        // Ignore cases where a CFD was not found
```
luckysori: 🔧 Can remove, because it's redundant with respect to the log one line below.

da-kami: done

luckysori (approved):
LGTM! Thanks! I've simplified one part with a fixup commit, but you can take it or leave it.

bonomat (approved):
LGTM

klochowicz (approved):

PR #1857, opened 2022-04-20
Gracefully handle initialization errors in oracle and monitor

### LINKED ISSUE #1848: Fix all fallible CFD loading loops
This is relevant for https://github.com/itchysats/itchysats/issues/1847

When processing loading CFDs after calling `load_all_open_cfds` we have some code parts where we fail the loop if loading the CFD fails. Consequently, the first loading failure will stop processing.

This is critical when starting the oracle, because the oracle actor is not supervised. 

Proposed fix (for now): Make startup of oracle and monitor unfailable. Only log errors.

da-kami:
fixes #1848 

If we fail to process one CFD we should not break the loop because we might not process the other CFDs leading to problems.
We don't fail the application if we run into an error with initializing the monitor or the oracle, so we should handle the initialization errors inside the actor.
I think fallible tasks when starting an actor that is not supervised should be avoided because we don't handle the errors and it is unclear what the behavior of the application would be.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -16,7 +16,7 @@ use bdk::miniscript::DescriptorTrait;
 use btsieve::ScriptStatus;
 use btsieve::State;
 use btsieve::TxStatus;
-use futures::TryStreamExt;
+use futures::StreamExt;
 use model::CfdEvent;
 use model::Dlc;
 use model::EventKind;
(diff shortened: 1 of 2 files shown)


daemon/src/monitor.rs
```diff
@@ -554,77 +554,77 @@ impl xtra::Actor for Actor {
...
+                    };
+
+                    if let Some(tx) = commit_tx {
+                        if let Err(e) = this
+                            .send(TryBroadcastTransaction {
+                                tx,
+                                kind: TransactionKind::Commit,
+                            })
+                            .await
+                            .expect("receiver to outlive the actor")
```
luckysori: ❓ I didn't expect to see this change in this commit. Can we undo it or at least move it to a separate commit?

luckysori (approved):
Good job using `next` instead of `try_next`!

I would like to see the changes from `?` to `expect` in a separate commit.

daemon/src/monitor.rs
da-kami: After having another look I am actually not sure we *should* expect in these cases. 

These are cases where we send to ourselves from a task that we spawn (and for some cases actually react to the result that the handler returns). I think we can assume that the actor should always be alive in such a case (because the task is cleaned up with the actor).
But probably it's still better to treat this as it was. I removed the expects - which means that in `monitor` the tasks added remain fallible.

klochowicz:
clarification for the last sentence commit description: fallible tasks are orthogonal to the supervisor actor - even if the actor was supervised, this fallible task would not have resume processing under any circumstances. They can be used in conjunction when the fallible task does stop the whole actor, causing a restart if it's supervised (which was not the case here!).

IMHO it would be better if we removed the sentence, as it's a bit misleading, but it's not that big of a deal :)

klochowicz (approved):
well spotted! there is no reason to stop loading other CFDs if one fails.

daemon/src/oracle.rs
```diff
@@ -312,37 +312,33 @@ impl xtra::Actor for Actor {
...
-                                .await;
+                    } = match cfd {
+                        Ok(cfd) => cfd,
+                        Err(e) => {
+                            tracing::warn!("Failed to load CFD from database: {e:#}");
+                            continue;
                         }
+                    };
+                    if let Some(pending_attestation) = pending_attestation {
+                        let _: Result<(), xtra::Disconnected> = this
```
bonomat, line 333: Do you need the type hint if you ignore it? I see that it has been there before, but maybe it is unnecessary.

bonomat (commented):
LGTM, thanks for fixing this.

bonomat (approved):

PR #1858, opened 2022-04-20
Startup optimization

bonomat:
When initialize monitoring we used to load cfds from the db twice in parallel. The problem is that the cache might not have been initialized when starting the second loop (they are processed in parallel). Hence, we unnecessarily load cfds twice from the db. This resulted in necessary load on our machine. Combining these two loops results in a huge optimization.

logs showing this is loaded twice: 
```
2022-04-20 00:48:13 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=daemon::monitor::Cfd cfd_version=0 num_events=554
2022-04-20 00:48:13 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=daemon::projection::Cfd cfd_version=0 num_events=554
2022-04-20 00:48:13 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=daemon::oracle::Cfd cfd_version=0 num_events=554
2022-04-20 00:48:13 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=daemon::monitor::Cfd cfd_version=0 num_events=554
```

Fun fact, loading this CFD from the DB in particular took 32seconds!!!

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -25,7 +25,6 @@ use model::CET_TIMELOCK;
 use serde_json::Value;
 use std::collections::HashMap;
 use std::time::Duration;
-use tokio::time::sleep;
 use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
@@ -621,16 +620,6 @@ impl xtra::Actor for Actor {
                             None => continue,
                             Some(params) => params,
                         };
-
-                        // NOTE: this is a band-aid fix.
-                        // It is possible for an attestation to be available when the refund
-                        // timelock has expired, for example, when the daemon goes down and is
-                        // restarted. In this case we want to prioritise
-                        // broadcasting the cet over the refund transaction.
-                        // We wait at least 30 seconds after the monitor actor is initialised before
-                        // reinitialising monitoring to give the daemon time to fetch and decrypt
-                        // the cet from the oracle if it is available.
-                        sleep(Duration::from_secs(30)).await;
 
                         this.send(ReinitMonitoring {
                             id,
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -557,7 +557,6 @@ impl xtra::Actor for Actor {
         self.tasks.add_fallible(
             {
                 let db = self.db.clone();
-                let this = this.clone();
 
                 async move {
                     let mut stream = db.load_all_open_cfds::<Cfd>(());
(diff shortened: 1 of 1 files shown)


luckysori (approved):
I'm okay with 82d97f4cb552c9ac80b4d5dee2701d3347a5e702 because the risk is minimal and the fix as formerly implemented probably didn't work. I would change the commit message title though 😛 

With 5608680292bb0288c1f6ca8cd709dc014a7f6adc I wonder why we decided to keep them separate originally and if combining them could have any adverse effects. I can't come up with anything though.

bonomat:
> I would change the commit message title though 😛

done :)

klochowicz:
if we ever decide that for some reason we'd like to keep them separate, we should ensure that they don't run in parallel - the aggregate cache would have worked if they weren't accessed at the same time, saving us from reaching all the way to db the second time. 

There's not much reason for not combining them though, unless they would be error-prone.

daemon/src/monitor.rs
```diff
@@ -675,7 +640,7 @@ impl xtra::Actor for Actor {
                 }
             },
             |e| async move {
-                tracing::warn!("Failed to re-initialize monitoring: {e:#}");
+                tracing::warn!("Failed to re-broadcast transactions: {e:#}");
```
klochowicz: The error became somewhat misleading, as we folded 2 distinct tasks into one:

```suggestion
                tracing::warn!("Failed to re-broadcast transactions or init the monitoring: {e:#}");
```

klochowicz (approved):

daemon/src/monitor.rs
```diff
@@ -675,7 +651,7 @@ impl xtra::Actor for Actor {
...
```
da-kami: 🙃 I think the message reading `Failed to re-initialize monitoring: {e:#}` given we combine the two blocks. I'd argue that re-broadcasting transactions is part of re-initializing the monitoring (but not the other way round).

da-kami (approved):
Oh my, I overlooked that `sleep` when I touched that code.

I am in favior of removing this, does not look like a good band-aide to me.

One sidenote: If I read the code correctly the 30s `sleep` was per CFD (!). Given that we load(ed) the CFDs in the same order this might have worked (because we would process the same CFD in the oracle, ...), but I still don't think the "gain" of giving priority to CETs over refund justifies the slow-down.

da-kami:
> I wonder why we decided to keep them separate originally and if combining them could have any adverse effects. I can't come up with anything though.

I gave it some hard thought but could not come up with a scenario where the separation is needed. Re-broadcasting transactions should not matter for kicking off monitoring I think.

daemon/src/monitor.rs
klochowicz: I like your suggestion better than mine :)

PR #1859, opened 2022-04-20
Roundtrip insertion of transactions into DB

luckysori:
Adding these tests revealed a bug that is also fixed in this patch:

In the `cets` table, we were inserting the value of the `vout` column in the `payout` column (and vice-versa). This was possible (without hard errors) because those two columns share the same underlying type i.e. integer.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -761,7 +761,7 @@ pub struct LockInput {
     timestamp: Timestamp,
 }
 
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq)]
 struct CollaborativeSettlement {
     txid: Txid,
     vout: Vout,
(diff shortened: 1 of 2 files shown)


da-kami (approved):
thanks for adding these and fixing a bug! 

I think eventually we could pick up some old ideas to go one step further and call `move_to_closed_cfds` for all the different cases. We could use snapshot DLC data for such a test. Not sure it's worth it but it would give us even more confidence I feel.

luckysori:
> thanks for adding these and fixing a bug!
> 
> I think eventually we could pick up some old ideas to go one step further and call `move_to_closed_cfds` for all the different cases. We could use snapshot DLC data for such a test. Not sure it's worth it but it would give us even more confidence I feel.

I totally agree. I started doing something like what you suggest and it was so painful that I gave up for the time being. But we should do it.

Issue #1842, opened 2022-04-19
CET without Commit causes error


da-kami:
Not sure if that's a bug in the way we monitor or the condition for "sanity" on transactions is just too strict:

```
  | {"timestamp":"2022-04-19T01:00:51.851369243Z","level":"ERROR","fields":{"message":"Failed to rehydrate CFD: Failed to load closed CFD 408e589a-655e-4a08-9edc-82960a090741: Closed CFD has insane combination of transactions:\n                       None,\n                       None,\n                       Some(Cet { txid: Txid(754338cf0289ad7e96be8fdb9cccb3824410376040e3cd3a3a51aef764a4aee3), vout: Vout(262754), payout: Payout(Amount(0.00000000 BTC)), price: Price(40590), timestamp: Timestamp(1650074407) })
-- | --
```

luckysori:
### Investigation

TL;DR: I cannot quite explain how we have run into this problem.

The logs reveal that this was a manually committed CFD:

```
{"message":"Appended event to database","event":"ManualCommit","order_id":"408e589a-655e-4a08-9edc-82960a090741"}
```

After the CET was confirmed sufficiently, it was moved along with other closed CFDs:


```
{"message":"Moving CFDs to closed_cfds table: [OrderId(Hyphenated(05bef017-6a09-4057-8074-5f7c42ca99c7)), OrderId(Hyphenated(408e589a-655e-4a08-9edc-82960a090741)), OrderId(Hyphenated(a5710ecb-9103-4821-9f33-a24009e212aa)), OrderId(Hyphenated(eb0365aa-e636-4ab7-aea0-febca938f101)), OrderId(Hyphenated(fa5e7ac6-ea5c-41ad-8d29-d982b5681b1b)), OrderId(Hyphenated(fb54d74b-196a-4a47-8237-bf1e42050305))]"}
```

The `event_log` table contains the `ManualCommit` event, meaning that it was part of the list of events when the CFD was being moved:

```
362|40|ContractSetupStarted|1649978140
363|40|ContractSetupCompleted|1649978145
364|40|LockConfirmed|1649978678
365|40|RolloverStarted|1649984629
366|40|RolloverAccepted|1649984629
367|40|RolloverCompleted|1649984648
368|40|LockConfirmed|1649984651
369|40|ManualCommit|1650074407
370|40|OracleAttestedPriorCetTimelock|1650074407
371|40|CommitConfirmed|1650074797
[...]

PR #1860, opened 2022-04-20
Prevent us from saving insane tx combinations into the db

luckysori:
By adding this we can detect problems _before_ we commit to deleting all the corresponding information in `cfds` and `events`, meaning that we can more easily debug problems related to the insane-tx-combinations error.

As an example, debugging https://github.com/itchysats/itchysats/issues/1842 would have been easier.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -424,22 +424,54 @@ impl Connection {
                 insert_closed_cfd(&mut db_tx, closed_cfd).await?;
                 insert_event_log(&mut db_tx, id, event_log).await?;
 
-                if let Some(collaborative_settlement) = closed_cfd.collaborative_settlement {
-                    insert_collaborative_settlement_tx(&mut db_tx, id, collaborative_settlement)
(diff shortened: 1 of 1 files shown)


daemon/src/db.rs
```diff
@@ -424,22 +424,54 @@ impl Connection {
...
+                        insert_refund_tx(&mut db_tx, id, refund).await?;
+                    }
+                    ClosedCfdInput {
+                        collaborative_settlement,
+                        commit,
+                        non_collaborative_settlement,
+                        refund,
+                        ..
+                    } => bail!(
+                        "CFD to be closed has insane combination of transactions:
```
bonomat, line 469: I can see this error already popping up in our logs 🙈

bonomat (approved):
LGTM

daemon/src/db.rs
luckysori: Hahah yes! Better earlier than later, as @da-kami pointed out offline.

PR #1863, opened 2022-04-20
Set the `commit_tx` in the model if available upon oracle attestation

da-kami:
I am not sure if we actually have edge cases where not setting this caused problems, but it is definitely more correct to apply it in case it only becomes available with the attestation (and not upon manual commit).
Note that we have to apply it only if it was not set previously, otherwise we might overwrite `Some` with `None`.

Note: In the projection this was already handled (without actually using the field, but it is correct). Oracle and monitor don't reason about this event.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1438,8 +1438,15 @@ impl Cfd {
                 self.during_contract_setup = false;
             }
             OracleAttestedPostCetTimelock { cet, .. } => self.cet = Some(cet),
-            OracleAttestedPriorCetTimelock { timelocked_cet, .. } => {
+            OracleAttestedPriorCetTimelock {
+                timelocked_cet,
+                commit_tx,
+                ..
+            } => {
                 self.cet = Some(timelocked_cet);
+                if self.commit_tx.is_none() {
+                    self.commit_tx = commit_tx;
+                }
             }
             ContractSetupFailed { .. } => {
                 self.during_contract_setup = false;


luckysori (approved):

model/src/cfd.rs
```diff
@@ -1438,8 +1438,15 @@ impl Cfd {
...
             OracleAttestedPostCetTimelock { cet, .. } => self.cet = Some(cet),
-            OracleAttestedPriorCetTimelock { timelocked_cet, .. } => {
+            OracleAttestedPriorCetTimelock {
+                timelocked_cet,
+                commit_tx,
+                ..
+            } => {
                 self.cet = Some(timelocked_cet);
+                if self.commit_tx.is_none() {
+                    self.commit_tx = commit_tx;
```
bonomat, line 1448: Don't you have to check if `commit_tx.is_some()`, otherwise you might set it to `None`?

bonomat: Oh, it does not matter, because it was `none` before :D

bonomat (approved):
LGTM

model/src/cfd.rs
da-kami: Yeah, this shows that we should model the `OracleAttestedPriorCetTimelock` event different. The problem is that we put an optional field into an event, but only *some* combinations are valid. Namely, if it was `None` (because we did not have a `ManualCommit`) it *has* to be `Some`. But we cannot enforce that - we cannot just fail `apply` if we don't have a `ManualCommit` and don't have the `commit_tx` in `OracleAttestedPriorCetTimelock`.

I think the correct way to model this is two events, where one contains the `commit_tx` and the other one does not. Depending on the situation we append either the one event or the other. But we cannot easily change events like this.

PR #1864, opened 2022-04-20
Mark `add_opening_fee` and `add_funding_fee` as `must_use`

luckysori:
These methods don't do anything to the original value i.e. they're immutable, so we must use the return value.

Adding the attribute to `add_funding_fee` reveals that we had a bug when moving CFDs to the `closed_cfds` table: we were ignoring the funding fee for every rollover that had occurred.


luckysori:
I want to see this tested on `testnet` and the change is uncontroversial.

bors r+

bonomat (commented):
👍

PR #1868, opened 2022-04-21
Set the commit transaction in the `ClosedInputCfdAggregate` upon `CommitConfirmed`

da-kami:
We have cases where one one side we never have an event that contains `commit_tx` because the other party published it was only picked up from the blockchain upon `CommitConfirmed`.
For such cases we set that `commit_tx` is set upon `CommitConfirmed` based on the latest DLC.

Note: When working on this I came to the conclusion that it might be better to reason differently when applying events on the `ClosedCfdInputAggregate`.
I think the better way will be setting data based on the `xxxConfirmed` events (if possible) and only reason about them. In the end it only matters what was confirmed on the blockchain. This will lilely simplify the logic in `ClosedCfdInputAggregate::apply`. @luckysori and me are looking into that now to see if it leads to a more resilient design. Until then we can still merge this fix so I posted this PR :)

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -13,6 +13,7 @@ use futures::StreamExt;
 use maia::TransactionExt;
 use model::CfdEvent;
 use model::Contracts;
+use model::Dlc;
 use model::EventKind;
 use model::FeeAccount;
 use model::Fees;
@@ -784,6 +785,7 @@ struct ClosedCfdInputAggregate {
     collaborative_settlement: Option<CollaborativeSettlement>,
     cet: Option<Cet>,
     refund: Option<Refund>,
+    latest_dlc: Option<Dlc>,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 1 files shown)


daemon/src/db.rs
```diff
@@ -933,7 +938,26 @@ impl ClosedCfdInputAggregate {
...
+                    None => {
+                        bail!("No DLC after commit confirmed");
+                    }
+                    Some(ref dlc) => {
+                        let script_pubkey = dlc.commit.2.script_pubkey();
+                        let OutPoint { txid, .. } = dlc
+                            .commit
+                            .0
+                            .outpoint(&script_pubkey)
+                            .context("Missing DLC in commit TX")?;
```
bonomat, line 952: I think the error message here is wrong. 
It's more something like: `Could not find output in commit tx`

bonomat (approved):
LGTM

da-kami:
The work that I just prepared with @luckysori is actually better, closing and re-opening.

PR #1869, opened 2022-04-21
Remove timestamp from closed CFD txs in DB

luckysori:
The timestamp column was included to help developers debug problems, but this information is readily available using a block explorer.

It has been removed because deciding which timestamp to use is not obvious. Should it be the one corresponding to the time of publication, inclusion in the mempool or confirmation? Being consistent across all transaction types is difficult because we use different events for each transaction type.

We recommend developers to use the `event_log` table which includes a `created_at` column to reason about transactions and time.

### DIFF
--- a/daemon/migrations/20220404113202_create_closed_cfds_tables.sql
+++ b/daemon/migrations/20220404113202_create_closed_cfds_tables.sql
@@ -10,8 +10,7 @@ CREATE TABLE IF NOT EXISTS closed_cfds (
     fees integer NOT NULL,
     expiry_timestamp integer NOT NULL,
     lock_txid text NOT NULL,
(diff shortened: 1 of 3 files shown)


bonomat (approved):
I like having all the data in 1 place. Cross-checking with the explorer is possible but is always an extra step. Anyway, happy to go with this if this makes things easier :)

luckysori:
> I like having all the data in 1 place. Cross-checking with the explorer is possible but is always an extra step. Anyway, happy to go with this if this makes things easier :)

I realise that, which is why I tried to include it in the original PR. But I couldn't find a way to keep this around without inconsistencies and increasing code complexity, which is why we made this call :(

PR #1871, opened 2022-04-21
Refactor `db::ClosedCfdInputAggregate` to be modelled around confirmations

da-kami:
We decide based on the confirmations and don't try to build up the CFD state as we apply events.
This is more driven by what actually happened on the blockchain.

Additionally we treat `commit` as part of a `cet` or `refund` settlement, which reduces error pathes.
This work fixes a bug where we did not set a commit transaction correctly.


luckysori (approved):
Great changes!

da-kami:
Merging this so we can see it work on the testnet setup over night.

PR #1873, opened 2022-04-21
bdk 0.17

### LINKED ISSUE #1821: Investigate regression after bumping `bdk` to version `0.17`
We want to keep depending on latest `bdk` but it appears that upgrading to `0.17` stopped our wallet from finding UTXOs. Perhaps we need to use the updated APIs differently or open an issue on `bdk` if the APIs no longer support our needs.

---

Original problem: https://github.com/itchysats/itchysats/issues/1814.

Bump to `bdk:0.17.0`: https://github.com/itchysats/itchysats/pull/1655.

Temporary fix: #1817.

bonomat:
Should fix #1821

If we want to ensure that more addresses are cached, then we need to call it explicitly:  https://github.com/bitcoindevkit/bdk/commit/45767fcaf7c65d5020f99fdc35c147acd6e8d037

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -11,13 +11,13 @@ use bdk::bitcoin::PublicKey;
 use bdk::bitcoin::Txid;
 use bdk::blockchain::Blockchain;
 use bdk::blockchain::ElectrumBlockchain;
-use bdk::blockchain::NoopProgress;
 use bdk::database::BatchDatabase;
 use bdk::wallet::tx_builder::TxOrdering;
 use bdk::wallet::AddressIndex;
 use bdk::FeeRate;
 use bdk::KeychainKind;
 use bdk::SignOptions;
+use bdk::SyncOptions;
 use maia::PartyParams;
 use maia::TxBuilderExt;
 use model::Timestamp;
@@ -84,7 +84,8 @@ static STD_DEV_UTXO_VALUE_GAUGE: conquer_once::Lazy<prometheus::Gauge> =
     });
 
 pub struct Actor<B> {
-    wallet: bdk::Wallet<B, bdk::database::MemoryDatabase>,
+    wallet: bdk::Wallet<bdk::database::MemoryDatabase>,
+    blockchain_client: B,
     used_utxos: LockedUtxos,
     tasks: Tasks,
     sender: watch::Sender<Option<WalletInfo>>,
@@ -109,7 +110,6 @@ impl Actor<ElectrumBlockchain> {
             Some(bdk::template::Bip84(ext_priv_key, KeychainKind::Internal)),
             ext_priv_key.network,
             db,
-            ElectrumBlockchain::from(client),
         )?;
 
         // UTXOs chosen after coin selection will only be locked for a
@@ -126,16 +126,19 @@ impl Actor<ElectrumBlockchain> {
             tasks: Tasks::default(),
             sender,
             used_utxos: LockedUtxos::new(time_to_lock),
+            blockchain_client: ElectrumBlockchain::from(client),
         };
 
         Ok((actor, receiver))
     }
 }
 
-impl<B: Blockchain> Actor<B> {
+impl Actor<ElectrumBlockchain> {
     fn sync_internal(&mut self) -> Result<WalletInfo> {
+        self.wallet.ensure_addresses_cached(1000)?;
+
         self.wallet
-            .sync(NoopProgress, Some(1000))
+            .sync(&self.blockchain_client, SyncOptions::default())
             .context("Failed to sync wallet")?;
 
         let balance = self.wallet.get_balance()?;
@@ -224,7 +227,10 @@ impl Actor<ElectrumBlockchain> {
         };
 
         self.wallet.sign(&mut psbt, SignOptions::default())?;
-        let txid = self.wallet.broadcast(&psbt.extract_tx())?;
+
+        let tx = psbt.extract_tx();
+        let txid = tx.txid();
+        self.blockchain_client.broadcast(&tx)?;
 
         tracing::info!(%txid, "Withdraw successful");
 
@@ -337,7 +343,7 @@ trait BuildLockTx {
     ) -> Result<PartiallySignedTransaction>;
 }
 
-impl<B, D> BuildLockTx for bdk::Wallet<B, D>
+impl<D> BuildLockTx for bdk::Wallet<D>
 where
     D: BatchDatabase,
 {
@@ -440,6 +446,7 @@ mod tests {
                     inner: HashSet::default(),
                     time_to_lock,
                 },
+                blockchain_client: (),
             })
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -247,9 +247,9 @@ dependencies = [
 
 [[package]]
 name = "bdk"
-version = "0.16.1"
+version = "0.17.0"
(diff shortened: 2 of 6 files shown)


daemon/src/wallet.rs
```diff
@@ -135,6 +135,8 @@ impl Actor<ElectrumBlockchain> {
 
 impl Actor<ElectrumBlockchain> {
     fn sync_internal(&mut self) -> Result<WalletInfo> {
+        self.wallet.ensure_addresses_cached(1000)?;
```
klochowicz, line 138: perhaps it would be worthwhile to expose this as an optional CLI parameter? In case we run into trouble with calculating wallet balance, we would be able to restart with a little tweak to the parameters instead of cutting a new release

klochowicz (approved):
great find!

luckysori (approved):
Looking at the `bdk` code we thought this wouldn't work for some reason, but I trust your judgement. And obviously the PR you link to indicates that it should. Thanks for looking into this!

daemon/src/wallet.rs
bonomat: I dont' think this is necessary. I'd rather make this dynamic and increase slowly until this maximum is reached.

bonomat:
> Looking at the `bdk` code we thought this wouldn't work for some reason, but I trust your judgement. And obviously the PR you link to indicates that it should. Thanks for looking into this!

Did I miss something? Why wouldn't it work ?

da-kami:
> > Looking at the `bdk` code we thought this wouldn't work for some reason, but I trust your judgement. And obviously the PR you link to indicates that it should. Thanks for looking into this!
> 
> Did I miss something? Why wouldn't it work ?

Because of the cache size being a constant of `100` addresses in `sync`:

https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L1575

and the fact that `ensure_addresses_cached` does not seem to use any other source to set this to a different number:

https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L354-L390

Maybe we missed something though. Following the code there is not easy and we did not evaluate it in detail. But might be worth having another look ;)
You could take the `taker_master`'s `testnet` wallet and try if the balance is being picked up using the code on this branch. We had problems with that recently, that's why we reverted the upgrade.

da-kami (changes requested):
Thanks for having another go at this!
I'd say we should only merge this after we successfully tested that the testnet setup's `taker_master` wallet balance was picked up correctly. I think it should not be too hard to test that locally. Might be a bit annoying, but I really want to avoid that we run into a problem where user's balances are `0` 😬

klochowicz:
> Thanks for having another go at this!
> I'd say we should only merge this after we successfully tested that the testnet setup's `taker_master` wallet balance was picked up correctly. I think it should not be too hard to test that locally. Might be a bit annoying, but I really want to avoid that we run into a problem where user's balances are `0` 😬

I'd say merging after release will give us plenty of time to iterate on the fix if for some reason it won't work. I agree, let's not rush it 👍

bonomat:
> > > Looking at the `bdk` code we thought this wouldn't work for some reason, but I trust your judgement. And obviously the PR you link to indicates that it should. Thanks for looking into this!
> > 
> > 
> > Did I miss something? Why wouldn't it work ?
> 
> Because of the cache size being a constant of `100` addresses in `sync`:
> 
> [https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L1575](https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs?rgh-link-date=2022-04-22T00%3A12%3A37Z#L1575)
> 
> and the fact that `ensure_addresses_cached` does not seem to use any other source to set this to a different number:
> 
> [https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L354-L390](https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs?rgh-link-date=2022-04-22T00%3A12%3A37Z#L354-L390)
> 
> Maybe we missed something though. Following the code there is not easy and we did not evaluate it in detail. But might be worth having another look ;) You could take the `taker_master`'s `testnet` wallet and try if the balance is being picked up using the code on this branch. We had problems with that recently, that's why we reverted the upgrade.

That's why we call it explicitly now which should have the same effect as before no?

da-kami:
> > > > Looking at the `bdk` code we thought this wouldn't work for some reason, but I trust your judgement. And obviously the PR you link to indicates that it should. Thanks for looking into this!
> > > 
> > > 
> > > Did I miss something? Why wouldn't it work ?
> > 
> > 
> > Because of the cache size being a constant of `100` addresses in `sync`:
> > [https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L1575](https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs?rgh-link-date=2022-04-22T00%3A12%3A37Z#L1575)
> > and the fact that `ensure_addresses_cached` does not seem to use any other source to set this to a different number:
> > [https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs#L354-L390](https://github.com/bitcoindevkit/bdk/blob/master/src/wallet/mod.rs?rgh-link-date=2022-04-22T00%3A12%3A37Z#L354-L390)
> > Maybe we missed something though. Following the code there is not easy and we did not evaluate it in detail. But might be worth having another look ;) You could take the `taker_master`'s `testnet` wallet and try if the balance is being picked up using the code on this branch. We had problems with that recently, that's why we reverted the upgrade.
> 
> That's why we call it explicitly now which should have the same effect as before no?

If you look at the code, then you see that `sync` calls `ensure_addresses_cached` with a constant of `100` addresses. I think that our call to `ensure_addresses_cached` with `1000` addresses does not change anything in the code. But maybe I'm mistaken. Did you have a look at the code?
Previously we passed the `1000` addresses to `sync`directly (which is now not possible anymore because the signature of `sync` changed)!

da-kami:
> > Thanks for having another go at this!
> > I'd say we should only merge this after we successfully tested that the testnet setup's `taker_master` wallet balance was picked up correctly. I think it should not be too hard to test that locally. Might be a bit annoying, but I really want to avoid that we run into a problem where user's balances are `0` 😬
> 
> I'd say merging after release will give us plenty of time to iterate on the fix if for some reason it won't work. I agree, let's not rush it 👍

Yeah, I'm fine with testing it on `testnet` after the release. We don't have to invest the time to test locally then, but might still be a good option because the feedback cycle is way shorter.

bonomat:
Yes, I did ;)

when calling `ensure_addresses_cached` we generate `X` addresses and store them in the db. When syncing this db is read and checked against the blockchain. 

Initially `ensure_addresses_cached` we passed in `1000` to the call `wallet.sync` which called `ensure_addresses_cached`. Now, we call `ensure_addresses_cached` with `1000` before and again with `100` inside `wallet.sync`. 

Not sure what you are referring to or usnure about?

da-kami (approved):
might still want to wait until the release is out to avoid potential problems related to the upgrade (not necessarily just the wallet)

bonomat:
btw: 0.18 is out 😅 https://github.com/bitcoindevkit/bdk/releases/tag/v0.18.0

thomaseizinger (approved):
Nice! Missed that in my original commit.

Would you mind squashing so we don't have a broken commit in the history?

PR #1876, opened 2022-04-22
Integrate libp2p with ping actor into daemons

### LINKED ISSUE #1766: Initial integration with libp2p-xtra
1. Use libp2p's tokio TCP transport
2. Create a ed25519 identity from our current network identity
3. Create a compatible listen multiaddress from our current listen address on the maker
4. Transform `maker` domain / ip + port to multiaddress for dialling on the taker
--5. Create a `KeepConnectionAlive` actor for the taker that - on an interval - queries the `ConnectionStats` on the taker side and reinitates a dial in case we are not connected to the maker.--
6. Introduce a `libp2p_connections_total` metric that tracks the number of connections

klochowicz:


xtra-libp2p/src/utils.rs
```diff
@@ -0,0 +1,32 @@
...
+    let ip = socket_addr.ip();
+    let port = socket_addr.port();
+    ensure!(socket_addr.is_ipv4(), "only ipv4 is supported");
+
+    format!("/ip4/{ip}/tcp/{port}/p2p/{peer_id}")
+        .parse::<Multiaddr>()
+        .with_context(|| "failed to construct multiaddr")
+}
+
+/// Creates MultiAddr from SocketAddr and PeerId
```
luckysori: 🙃 This docstring says we're using a `PeerId` to build the listen multiaddress, but we're not.

xtra-libp2p/src/lib.rs
```diff
@@ -24,6 +24,7 @@ pub mod listener;
 mod multiaddress_ext;
 mod substream;
 mod upgrade;
+pub mod utils;
```
luckysori: 🙃 Could do

```rust
mod utils:

pub use utils::*;
```

If we don't want to leak the fact that this crate has a `utils` module.

Very subjective though.

daemon/src/lib.rs
```diff
@@ -140,6 +148,40 @@ where
...
+            TokioTcpConfig::new(),
+            identity.libp2p,
+            ENDPOINT_CONNECTION_TIMEOUT,
+            [(
+                xtra_libp2p_ping::PROTOCOL_NAME,
+                xtra::message_channel::StrongMessageChannel::clone_channel(&ping_address),
+            )],
+        );
+
+        let libp2p_socket = SocketAddr::new(p2p_socket.ip(), p2p_socket.port() + 1);
```
luckysori: 🙃 I would move this line next to where it's used.

daemon/src/lib.rs
```diff
@@ -140,6 +148,40 @@ where
...
+        tasks.add(endpoint_context.run(endpoint));
+
+        let endpoint_listen = xtra_libp2p::utils::create_listen_tcp_multiaddr(&libp2p_socket)
+            .expect("to parse properly");
+
+        let listener_constructor = move || {
+            let endpoint_addr = endpoint_addr.clone();
+            let endpoint_listen = endpoint_listen.clone();
+            listener::Actor::new(endpoint_addr, endpoint_listen)
+        };
```
luckysori: ❓ Can/should we inline this?

taker/src/main.rs
```diff
@@ -298,6 +318,19 @@ async fn main() -> Result<()> {
...
+    let possible_addresses = resolve_maker_addresses(maker_url.as_str()).await?;
+
+    // Assume that the first resolved ipv4 address is good enough for libp2p.
+    let first_maker_address = possible_addresses
+        .iter()
+        .find(|x| x.is_ipv4())
+        .context("Could not resolve maker URL")?;
+
+    let maker_libp2p_address =
+        SocketAddr::new(first_maker_address.ip(), first_maker_address.port() + 1); // libp2p is always the +1 port from legacy networking
```
luckysori: 🔧 Perhaps we should encapsulate this knowledge in a function with a docstring.

klochowicz:
one thing I was not too happy with was the "magically" hard-coding the libp2p port. happy to expose it as another CLI option, and default it on the taker. Might be best in a follow-up

xtra-libp2p/src/dialer.rs
```diff
@@ -35,22 +36,24 @@ impl Actor {
...
-
-        listen_addresses
-            .contains(&self.connect_address)
+        connected_peers
+            .contains(
+                &self
+                    .connect_address
+                    .clone()
+                    .extract_peer_id()
+                    .ok_or(Error::InvalidPeerId)?,
```
luckysori, line 56: 🔧 It's annoying to have to deal with this error when the result will never change from the moment we construct the actor. Can we get around this?

xtra-libp2p/src/dialer.rs
```diff
@@ -35,22 +36,24 @@ impl Actor {
         }
     }
 
-    /// Returns error if we cannot access Endpoint or the connect address is not
-    /// present inside Endpoint.
+    /// Returns error if we cannot access Endpoint or the peer is not connected Endpoint.
```
luckysori: 🙃 Missing words.

```rust
/// Returns error if we cannot access the Endpoint or if the peer is not connected to the Endpoint.
```

xtra-libp2p/src/dialer.rs
luckysori: We could build and save the `PeerId` in `new` and make `new` fallible. Only downside is slight redundancy of data. But I prefer that over having to handle errors every time we check if a connection is active.

luckysori (approved):
Excited to get this in!

daemon/src/lib.rs
klochowicz: 👍

klochowicz: :+1:

xtra-libp2p/src/lib.rs
klochowicz: I don't think there's much value in leaking `utils` module, perhaps a bit of discovery - happy to change it your way. :)

taker/src/main.rs
klochowicz: great idea. Both sides should use the same piece of code at all times, it's a shared convention after all.

xtra-libp2p/src/dialer.rs
klochowicz: unfortunately, the supervisor does not support fallible actor constructors. I had a quick stab in remodelling the arguments to tighten the API, but it's a bit annoying too, and I don't want to use `expect()` anywhere. Will keep it as-is for now, might get back to it later on.

xtra-libp2p/src/utils.rs
```diff
@@ -0,0 +1,32 @@
+use anyhow::ensure;
+use anyhow::Context;
+use anyhow::Result;
+use std::net::SocketAddr;
+
+use libp2p_core::Multiaddr;
+use libp2p_core::PeerId;
+
+/// Creates MultiAddr from SocketAddr and PeerId
+pub fn create_connect_tcp_multiaddr(
```
thomaseizinger, line 10: I reckon these should live outside of `xtra-libp2p`. This utils are super specific to your usecase and having them in here will create friction if you ever want to move this crate into its own repository and release it on crates.io.

daemon/src/lib.rs
```diff
@@ -140,6 +148,38 @@ where
...
+
+        let libp2p_socket = libp2p_socket_from_legacy_networking(&p2p_socket);
+        let endpoint_listen =
+            xtra_libp2p::create_listen_tcp_multiaddr(&libp2p_socket).expect("to parse properly");
+
+        let (supervisor, _listener_actor) = supervisor::Actor::with_policy(
+            move || {
+                let endpoint_addr = endpoint_addr.clone();
+                let endpoint_listen = endpoint_listen.clone();
+                listener::Actor::new(endpoint_addr, endpoint_listen)
```
thomaseizinger, line 177: These also seem app-specific because they hint at a policy of "one app is either dialer or listener". `xtra-libp2p` was meant to be policy-free to make it more applicable to many usecases.

Additionally, having that code in the `xtra-libp2p` crate means each binary needs to compile it even though it only uses a part of it. For now, we could put them into the `daemon` crate but long-term, the actor systems would be better of being in the respective binary crate along-side with these use-case specific actors.

As a rule of thumb, generic code that can be built through composition should likely live in a separate crate, esp. when the goal is to release the building block as its own crate. Smaller API surface means more stable API surface :)

taker/src/main.rs
```diff
@@ -117,7 +126,19 @@ impl Opts {
...
-        Ok((maker_url, maker_id))
+        let maker_peer_id = match self.maker_peer_id {
+            Some(maker_peer_id) => maker_peer_id,
+            None => match network {
+                // TODO: Add defaults for mainnet and testnet
+                Network::Mainnet { .. } => bail!("No maker default peer id configured for mainnet"),
+                Network::Testnet { .. } => bail!("No maker default peer id configured for testnet"),
+                Network::Signet { .. } => {
+                    bail!("No maker default peer id configured for signet")
+                }
```
thomaseizinger, line 137: Signet is shorter than the other two, why does this one wrap?

taker/src/main.rs
```diff
@@ -298,6 +319,17 @@ async fn main() -> Result<()> {
 
     let (projection_actor, projection_context) = xtra::Context::new(None);
 
+    let possible_addresses = resolve_maker_addresses(maker_url.as_str()).await?;
+
+    // Assume that the first resolved ipv4 address is good enough for libp2p.
+    let first_maker_address = possible_addresses
+        .iter()
+        .find(|x| x.is_ipv4())
+        .context("Could not resolve maker URL")?;
```
thomaseizinger, line 328: Instead of doing this, we should wrap the `TokioTcpConfig` in a DNS transport and let it do the name resolution. See https://docs.rs/libp2p-dns/0.32.1/libp2p_dns/.

xtra-libp2p/src/dialer.rs
```diff
@@ -46,6 +46,9 @@ impl Actor {
             .await
             .map_err(|_| Error::NoEndpoint)?;
 
+        dbg!(&listen_addresses);
+        dbg!(&self.connect_address);
```
thomaseizinger: Are these meant to still be here? Do we not have a lint that detects this?

xtra-libp2p/src/dialer.rs
```diff
@@ -36,7 +36,8 @@ impl Actor {
         }
     }
 
-    /// Returns error if we cannot access Endpoint or the peer is not connected Endpoint.
+    /// Returns error if we cannot access the Endpoint or if the peer is not
+    /// connected to the Endpoint.
```
thomaseizinger, line 40: Does `dprint check` pass for all commits before this one? You are not changing the comment here.

thomaseizinger (approved):
A few comments, nothing major :)

Note that `xtra-libp2p-ping` actually has metrics so we will see those as well, not just in the logs!

xtra-libp2p/src/utils.rs
klochowicz: why would this create friction? they don't  bring any dependencies (`SocketAddr` is std lib after all)

taker/src/main.rs
klochowicz: I'll create an issue from this and deal with it in a follow-up, good idea!

daemon/src/lib.rs
klochowicz: I'm happy to extract `listener` and `dialer` to separate crates if that suits you.
Optionally, I could hide them behind feature flags? 
 
The trouble with policy-free crates is that they are less typically less usable, making it cumbersome to be quickly used in a project. I think the building blocks added recently, even if they hint that they are usable only for one party being dialer, the other listener, are still generic enough to be used elsewhere, and as such have nothing to do with itchysats app?

thomaseizinger: > are still generic enough to be used elsewhere, and as such have nothing to do with itchysats app?

Yes but you don't know the requirements of "elsewhere", so really it is premature optimisation to assume that a different usecase will need the same policy.

> I'm happy to extract `listener` and `dialer` to separate crates if that suits you.

Ideally, they would sit in `maker` and `taker` but we can't do that until we move the actor systems out of `daemon` so I would put it in `daemon` for now.

> The trouble with policy-free crates is that they are less typically less usable, making it cumbersome to be quickly used in a project.

We are probably splitting hairs at this point. Policy-free code is typically _more useful_ because it applies to a wider range of usecases (keeping policy out is one of the core guidelines of the linux kernel for example [^1]). Does that make it less usable because you need to write a couple of lines of code for your usecase? Hard to say.

[^1]: As I have learnt from @tcharding, can't quote another source unfortunately :)

tcharding: I couldn't find a kernel reference for you but here is one from 'The Art of UNIX Programming', which is an awesome book in my opinion.

https://www.linuxtopia.org/online_books/programming_books/art_of_unix_programming/ch01s06_3.html

luckysori: I wonder how this copy arrived in my bookshelf!

image

klochowicz: thank you for the comments, I will keep this is in mind when working in this library in the future. Really insightful.

taker/src/main.rs
klochowicz: I have no idea, I guess this is a question to `rustfmt` :)

taker/src/main.rs
klochowicz: tracked by #1896

xtra-libp2p/src/dialer.rs
klochowicz: I tried checking running `dprint` after each commit and couldn't find anything.
Also, I don't know how to refer back to particular commit from github UI, so I can't tell for sure.

xtra-libp2p/src/utils.rs
klochowicz: I'll have a think about better placement as we keep going. It's still early days.

klochowicz:
bors cancel

xtra-libp2p/src/dialer.rs
klochowicz: we have a lint and they're not here as a result of the changeset. clippy doesn't run for each commit separately, if we care enough we could change our github actions, but I don't think it's worth the hassle.

PR #1880, opened 2022-04-22
Fix join on the correct `cfd_id`

da-kami:
`commit_tx.id` points to the auto-generated `id` of the table, we what we have to `join` on is always `closed_cfds.id` which is represented as `cfd_id`, i.e. `commit_tx.cfd_id`.


bonomat (approved):
LGTM. Thanks for the test. Best test ever!

luckysori (approved):

PR #1893, opened 2022-04-26
Open user interface in default browser upon startup

da-kami:
This improves the experience to get the user interface up and running.
The URL includes username and password so the user just gets the UI.

Behaviour: Build it and run `./taker testnet` (or just `./taker` for mainnet) and the UI will just be opened in your default browser ready to use :) - This basically means double click will launch the UI (with daemon on mainnet)

Note: We might be able to access the `Username` and `Password` from the rocket state in the fairing (those two are managed by rocket), but I could not figure that out so I just create a dumb version for now. Making the password `Clone` might be somewhat of a security risk though.

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -22,6 +22,8 @@ use daemon::N_PAYOUTS;
 use model::olivia;
 use model::Identity;
 use model::SETTLEMENT_INTERVAL;
+use rocket::fairing::AdHoc;
+use rocket::fairing::Fairing;
 use shared_bin::fairings;
 use shared_bin::logger;
 use shared_bin::logger::LevelFilter;
@@ -354,6 +356,7 @@ async fn main() -> Result<()> {
         .register("/", rocket::catchers![rocket_basicauth::unauthorized])
         .attach(fairings::log_launch())
         .attach(fairings::log_requests())
+        .attach(ui_browser_launch())
         .launch()
         .await?;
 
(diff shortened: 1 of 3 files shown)


bonomat (approved):
works for me

klochowicz:
how would this behave with docker container? are we sacrificing security of umbrel node or any other remote machine with this PR?

bonomat:
> how would this behave with docker container? are we sacrificing security of umbrel node or any other remote machine with this PR?

it will fail because there is no default browser in the docker container. This only works for binaries.

da-kami:
> > how would this behave with docker container? are we sacrificing security of umbrel node or any other remote machine with this PR?
> 
> it will fail because there is no default browser in the docker container. This only works for binaries.

It will then print on `debug` level that it was unable to start the UI in the default browser. I think this is acceptable at this point. But it's a good point, because I will likely still have to explain where to get the password from when using the container.

taker/src/main.rs
```diff
@@ -374,3 +377,33 @@ async fn resolve_maker_addresses(maker_addr: &str) -> Result<Vec<SocketAddr>> {
     );
     Ok(possible_addresses)
 }
+
+/// Attach this fairing to enable logging Rocket launch
```
thomaseizinger: Stale comment.

taker/src/main.rs
```diff
@@ -374,3 +377,33 @@ async fn resolve_maker_addresses(maker_addr: &str) -> Result<Vec<SocketAddr>> {
...
+pub fn ui_browser_launch(
+    username: rocket_basicauth::Username,
+    password: rocket_basicauth::Password,
+) -> impl Fairing {
+    AdHoc::on_liftoff("ui browser launch", move |rocket| {
+        Box::pin(async move {
+            let http_endpoint = format!(
+                "http://{}:{}@{}:{}",
+                username,
+                password,
```
thomaseizinger, line 399: You can get this from the `Rocket` instance: https://docs.rs/rocket/0.5.0-rc.1/rocket/struct.Rocket.html#method.state

rocket-basicauth/src/lib.rs
```diff
@@ -47,7 +47,7 @@ impl PartialEq<String> for Username {
     }
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
thomaseizinger: We can revert this once we use https://docs.rs/rocket/0.5.0-rc.1/rocket/struct.Rocket.html#method.state.

klochowicz (approved):
happy for it to go in, Rocket state can be rolled out in a follow-up so not to hold the release off.

thomaseizinger (approved):

PR #1894, opened 2022-04-26
Use refund TX to calculate payout in projection

luckysori:
Although exceedingly rare, a CFD's payout can come from a refund transaction

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -261,11 +261,7 @@ impl Aggregated {
         }
 
         let tx = self.cet.or(self.timelocked_cet)?;
-        let dlc = self
-            .latest_dlc
-            .as_ref()
-            .expect("dlc to be present when we have a cet");
-        let script = dlc.script_pubkey_for(role);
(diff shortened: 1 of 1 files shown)


daemon/src/projection.rs
```diff
@@ -252,6 +255,11 @@ impl Aggregated {
             return Some(extract_payout_amount(tx, script));
         }
 
+        if let Some(tx) = self.refund_tx {
+            let script = self.latest_dlc?.script_pubkey_for(role);
+            return Some(extract_payout_amount(tx, script));
+        }
+
         let tx = self.cet.or(self.timelocked_cet)?;
         let dlc = self
```
da-kami: Could change below `.expect("dlc to be present when we have a cet");` to be just `?` as well I suppose.

da-kami (approved):

daemon/src/projection.rs
luckysori: Done: https://github.com/itchysats/itchysats/pull/1894/commits/613cb3c6c4737c5df1403bc16602af75d2ddfb11.

PR #1902, opened 2022-04-27
Use rayon to deserialize CFD events in parallel

thomaseizinger:
Rayon is a parallel data processing library. By replacing `into_iter`
with `into_par_iter`, we are constructing a parallel iterator that is
baked by a thread-pool which defaults to a number of threads equal to
the logical cores of the system.

On my local machine (8 logical cores), this speeds up the
`projection::Initialize` handler from 83 to 28 seconds for a 2.8 GB
database.

### DIFF
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1088,12 +1088,14 @@ async fn load_cfd_row(conn: &mut Transaction<'_, Sqlite>, id: OrderId) -> Result
 ///
 /// The version of a CFD is the number of events that have been applied. If we have an aggregate
 /// instance in version 3, we can avoid loading the first 3 events and only apply the ones after.
+///
+/// Events will be sorted in chronological order.
 async fn load_cfd_events(
     conn: &mut Transaction<'_, Sqlite>,
     id: OrderId,
     from_version: u32,
 ) -> Result<Vec<CfdEvent>> {
-    let events = sqlx::query!(
+    let mut events = sqlx::query!(
         r#"
 
         select
@@ -1122,6 +1124,8 @@ async fn load_cfd_events(
         })
     })
     .collect::<Result<Vec<_>>>()?;
+
+    events.sort_unstable_by(CfdEvent::chronologically);
 
     Ok(events)
 }
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -33,6 +33,7 @@ use model::Txid;
 use model::Usd;
 use model::Vout;
 use model::SETTLEMENT_INTERVAL;
+use rayon::prelude::*;
 use sqlx::migrate::MigrateError;
 use sqlx::pool::PoolConnection;
 use sqlx::sqlite::SqliteConnectOptions;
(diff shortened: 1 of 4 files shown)


luckysori (approved):
Damn, that's amazing! Any downsides to using this?

klochowicz (approved):
awesome idea! 👍

klochowicz:
> Damn, that's amazing! Any downsides to using this?

- you might get slightly different results if a given iterator has side effects (e.g. sending stuff over a channel or writing to disk)
- if you have a handful of elements in a collection it might not be worth the overhead (ie. it might actually be a bit slower).

da-kami (approved):
Great find, happy to give it a try!
Sounds promising! https://github.com/rayon-rs/rayon#no-data-races

thomaseizinger:
Let me double check that this doesn't screw up the order of events :D

da-kami (approved):
I suppose time needed for sorting can be neglected.

daemon/src/db.rs
```diff
@@ -1123,6 +1125,8 @@ async fn load_cfd_events(
     })
     .collect::<Result<Vec<_>>>()?;
 
+    events.sort_unstable_by(CfdEvent::chronologically);
```
bonomat, line 1128: Why don't you use sql for the ordering? That would save us from needing `CfdEvent::chronologically`

thomaseizinger: Because I want to ensure that I do the sorting _after_ I've used rayon to deserialize the events in parallel. I think rayon should preserve the order but as mentioned in the commit message, I think it is better to be explicit about this!

bonomat: Ah, got it.

thomaseizinger:
> I suppose time needed for sorting can be neglected.

I hope so? I think they should already be sorted anyway so ideally, this function call is a no-op.

klochowicz:
> > I suppose time needed for sorting can be neglected.
> 
> I hope so? I think they should already be sorted anyway so ideally, this function call is a no-op.

yup, according to rayon the order should be preserved https://github.com/rayon-rs/rayon/issues/551

thomaseizinger:
> > > I suppose time needed for sorting can be neglected.
> > 
> > 
> > I hope so? I think they should already be sorted anyway so ideally, this function call is a no-op.
> 
> yup, according to rayon the order should be preserved [rayon-rs/rayon#551](https://github.com/rayon-rs/rayon/issues/551)

Thanks! I've referenced the issue in the commit message.

model/src/cfd.rs
```diff
@@ -378,6 +379,32 @@ impl CfdEvent {
...
+    /// ```rust
+    /// # use model::{CfdEvent, Timestamp, EventKind, OrderId};
+    ///
+    /// # let id = OrderId::default();
+    /// # let make_event_at = move |seconds| CfdEvent {
+    /// #    timestamp: Timestamp::new(seconds),
+    /// #    id,
+    /// #    event: EventKind::LockConfirmed,
+    /// # };
+    /// #
```
luckysori, line 396: ❓ Thanks for adding an example! What do the `#`s mean?

thomaseizinger: They make this part of the code not render in rustdoc!

PR #1907, opened 2022-04-27
Increase message timeout for contract setup to `120s`

da-kami:
From incidents where `ContractSetup` failed we saw that the messages from the taker actually did arrive, but it took longer than `60s` to get processed on the maker.
This can be because sending + receiving a message might be slow over slow internet connection, but can also be related to the message load on the maker.
The `maker_inc_connections` actor is a bottleneck at the moment, but that cannot easily be solved.
Given that we know that a dropped connection was *not* the problem in this instance, but the messages were just not processed in time we increase the timeout for contract setup messages.
This should not have a bad impact on the system, because contract setup only happens sporadically.
Rollover messages remain at `60s` because we want to avoid to keep too many rollover actors.
Failing rollovers by stopping actors early is not a problem because rollovers are automatically retried.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -56,8 +56,18 @@ use std::ops::RangeInclusive;
 use std::time::Duration;
 use xtra::prelude::MessageChannel;
 
-/// How long protocol waits for the next message before giving up
-const MSG_TIMEOUT: Duration = Duration::from_secs(60);
+/// How long contract setup protocol waits for the next message before giving up
+///
+/// 120s are currently needed to ensure that we can outlive times when the maker/taker are under
+/// heavy message load. Failed contract setups are annoying compared to failed rollovers so we allow
+/// more time to see them less often.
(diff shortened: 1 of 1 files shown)


daemon/src/setup_contract.rs
```diff
@@ -56,8 +56,18 @@ use std::ops::RangeInclusive;
 use std::time::Duration;
 use xtra::prelude::MessageChannel;
 
-/// How long protocol waits for the next message before giving up
-const MSG_TIMEOUT: Duration = Duration::from_secs(60);
+/// How long rollover protocol waits for the next message before giving up
```
luckysori: 🔧 

```suggestion
/// How long contract setup waits for the next message before giving up
```

klochowicz (approved):
Thank you for detailed write-up. I agree with the reasoning, it matches what we've seen in the logs lately. LGTM!

luckysori (approved):

thomaseizinger (approved):

PR #1910, opened 2022-04-27
Prevent specifying TxFeeRate below min relay fee

klochowicz:


klochowicz:
I'm not sure whether this is the best spot to handle these errors. It also shows that we should have tests / fuzz against wrong values early. Maybe a few rocket tests would be good to have.

luckysori (commented):
I suggest https://github.com/itchysats/itchysats/commit/001e61401396588ec4b5747335a95af434bc9235 instead of https://github.com/itchysats/itchysats/pull/1910/commits/502076c952224c4c4e581f24146717342e88300b.

maker/src/routes.rs
```diff
@@ -124,6 +124,12 @@ pub async fn put_offer_params(
     maker: &State<Maker>,
     _auth: Authenticated,
 ) -> Result<(), HttpApiProblem> {
+    if offer_params.tx_fee_rate.to_u32() < 1 {
+        return Err(HttpApiProblem::new(StatusCode::BAD_REQUEST)
+            .title("Posting offer failed")
```
thomaseizinger, line 129:
```suggestion
            .title("Bad fee rate")
```

thomaseizinger (approved):
I think I'd agree with @luckysori to use https://github.com/itchysats/itchysats/commit/001e61401396588ec4b5747335a95af434bc9235 instead.

It might be worth trying out what the error message is in case someone sends `0` and we try to deserialize it into a `NonZeroU32`.

thomaseizinger:
> It might be worth trying out what the error message is in case someone sends `0` and we try to deserialize it into a `NonZeroU32`.

Quick research suggests a good enough error message: https://docs.rs/serde/1.0.136/src/serde/de/impls.rs.html#99

klochowicz:
> > It might be worth trying out what the error message is in case someone sends `0` and we try to deserialize it into a `NonZeroU32`.
> 
> Quick research suggests a good enough error message: https://docs.rs/serde/1.0.136/src/serde/de/impls.rs.html#99

it seems it's being swallowed by Rocket. We might need to use custom catcher or something, because as it is it's not that obvious at all :(

in the logs:
`Handled request method=PUT path=/api/offer status=422 Unprocessable Entity`

in the UI:
image

klochowicz:
> I suggest [001e614](https://github.com/itchysats/itchysats/commit/001e61401396588ec4b5747335a95af434bc9235) instead of [502076c](https://github.com/itchysats/itchysats/commit/502076c952224c4c4e581f24146717342e88300b).

thank you! I really like leveraging the type system here! the downside is that we'll need to dig deeper into Rocket to allow some sane error messages here, but that seems to be a better way forward

thomaseizinger:
It is probably better to be lenient in the serialization model and try to map into a more type-safe model manually at which point we can then create nice error messages depending on why it fails.

klochowicz:
Superceded by #1927

klochowicz:
> > > It might be worth trying out what the error message is in case someone sends `0` and we try to deserialize it into a `NonZeroU32`.
> > 
> > 
> > Quick research suggests a good enough error message: https://docs.rs/serde/1.0.136/src/serde/de/impls.rs.html#99
> 
> it seems it's being swallowed by Rocket. We might need to use custom catcher or something, because as it is it's not that obvious at all :(
> 
> in the logs: `Handled request method=PUT path=/api/offer status=422 Unprocessable Entity`
> 
> in the UI: image

added a custom catcher in a different PR that at least displays a bit less cryptic message.

PR #1916, opened 2022-04-28
Move failed and rejected CFDs to a separate DB table

### LINKED ISSUE #1881: Make `cfds` the table for *only* `open` cfds
With #1779 we moved CFDs that have a confirmed spend transaction into `closed_cfds`.

With this we wanted to start treating `cfds` as `open` cfds, but this is not possible yet because `cfds` still contains `OfferRejected` and `ContractSetupFailed` entries. 

In order to treat the `cfds` table as `open` cfds we have to:

- Periodically move `OfferRejected` and `ContractSetupFailed` cfds to separate table(s), that way the `cfds` table is truly cleaned up.
- Consider removal of `load_open_cfd_ids` and adapt `load_all_open_cfds` to just load from `cfds` without filtering. At the moment we have the filter in place because we have a big amount of closed CFDs to be moved on the initial run of moving to `closed_cfds`. Once the initial run is done it should be fine to assume that `cfds` is reasonably cleaned up and we can just load from it. Removing those filter-function will simplify code.

luckysori:
Fixes #1881.

Tasks:

- [x] Refactor `db` so that it doesn't keep growing out of control.
- [x] Add `db` APIs to move failed and rejected CFDs to a separate table.
- [x] Test `db::failed` APIs.
- [x] Introduce actor to periodically move failed and rejected CFDs to the corresponding table.
- [x] Consider removing any filters applied to the `cfds` table (since it can now be assumed that it generally only contains open CFDs).


daemon/src/db/failed.rs
```diff
@@ -0,0 +1,179 @@
+//! This module allows us to move faile CFDs to a separate table so
```
thomaseizinger:
```suggestion
//! This module allows us to move failed CFDs to a separate table so
```

daemon/src/db/failed.rs
```diff
@@ -0,0 +1,179 @@
...
+                db_tx.commit().await?;
+
+                anyhow::Ok(())
+            };
+
+            if let Err(e) = fut.await {
+                tracing::warn!(order_id = %id, "Failed to move failed CFD: {e:#}");
+            }
+
+            tracing::debug!(order_id = %id, "Moved CFD to `failed_cfds` table");
```
thomaseizinger: You will print this even in an error case :)

The entire construct should probably be a `match`!

daemon/src/db/closed.rs
```diff
@@ -142,6 +142,8 @@ impl Connection {
             if let Err(e) = fut.await {
                 tracing::warn!(order_id = %id, "Failed to move closed CFD: {e:#}");
             }
+
+            tracing::debug!(order_id = %id, "Moved CFD to `closed_cfds` table");
```
thomaseizinger: This will log-unconditionally, so also in the case of an error above :)

Probably best to use a `match` or early return.

daemon/src/db.rs
```diff
@@ -315,6 +317,28 @@ impl Connection {
...
+
+            let ids = sqlx::query!(
+                r#"
+                SELECT
+                    uuid as "uuid: model::OrderId"
+                FROM
+                    failed_cfds
+                "#
+            )
+            .fetch_all(&mut *conn)
```
thomaseizinger: I recently learned that there is also `fetch` which returns a `Stream`! That would make the whole thing even more "iteratively"!

thomaseizinger (commented):
Few comments but looks like a good direction!

thomaseizinger:
> Consider removing any filters applied to the `cfds` table (since it can now be assumed that it generally only contains open CFDs).

I think that may not be fully true!

There can be a time-window where a CFD fails / gets rejected and has not been moved yet.

daemon/src/db/failed.rs
```diff
@@ -0,0 +1,179 @@
...
+use sqlx::Transaction;
+
+/// A trait for building an aggregate based on a `FailedCfd`.
+pub trait FailedCfdAggregate: CfdAggregate {
+    fn new_failed(args: Self::CtorArgs, cfd: FailedCfd) -> Self;
+}
+
+/// Data loaded from the database about a failed CFD.
+#[derive(Debug, Clone, Copy)]
+pub struct FailedCfd {
```
da-kami, line 53: 🛠 Failed CFDs may have events, i.e. we may have a `ContractSetupStarted` and `ContractSetupFailed` event. I think it might be a good idea to save the timestamp of the first event recorded - that is what we use for sorting the open CFDs.. 
I think a user would still expect to see the CFDs chronologically ordered, but without timestamps we cannot really achieve that. 
I think this is actually already missing when processing closed cfds - I would always save it as `creation_timestamp` (or `first_event_timestamp` it you want). I am not sure we have to add it as another field, we could also load it from the eventlog (I suppose failed will also have eventlog).
I think we should always have at least one event in the database, so we should always be able to come up with a timestamp. For cases where we don't have any events we could actually just delete from `cfds` because this is too incomplete to do anything.

da-kami (commented):
Overall concept ack. 
Would you treat rejected separate or handle them as "failed" where we add a reason why it failed?
I think it might be good enough to treat all "failed" in the same table unless we want to store different data for failed/rejected.

I noticed that we are missing one thing from the previous work already and that is the ability to sort the closed / failed CFDs chronologically. See my related comment

daemon/src/db.rs
luckysori: Thanks for the tip!

How do you think it would interact with the fact that we would be using the same `Connection` to generate this stream and to load CFDs?

luckysori:
> > Consider removing any filters applied to the `cfds` table (since it can now be assumed that it generally only contains open CFDs).
> 
> I think that may not be fully true!
> 
> There can be a time-window where a CFD fails / gets rejected and has not been moved yet.

That is true, which is why I said "generally" 😛 We definitely cannot remove all these filters blindly.

daemon/src/db/failed.rs
luckysori: > 🛠 Failed CFDs may have events, i.e. we may have a ContractSetupStarted and ContractSetupFailed event.

Do you think it's worth saving events for failed CFDs? To me it's even questionable if we should be persisting them! One could imagine that a failed CFD could eventually just become an ephemeral notification for the user.

But with this PR I would persist them since we have no better alternative at the moment.

> I think it might be a good idea to save the timestamp of the first event recorded - that is what we use for sorting the open CFDs..
I think a user would still expect to see the CFDs chronologically ordered, but without timestamps we cannot really achieve that.

Good point: we want these to appear in order. It is a bit annoying to have to process the events at all just to get a timestamp, but I guess we should.

> I think this is actually already missing when processing closed cfds - I would always save it as creation_timestamp (or first_event_timestamp it you want).

It's true that the closed CFDs are not sorted properly (because we don't use the `projection::Cfd::aggregate` field. More importantly, the `profit_percent`, `profit_btc` and `payout` fields are actually wrong. I added a test to verify this, but I did not realise that we were calling `with_current_quote` on all CFDs (doing so on closed CFDs overwrites these values with wrong ones).

> I am not sure we have to add it as another field, we could also load it from the eventlog (I suppose failed will also have eventlog).

I would prefer not to add entries to the `event_log` table for failed or rejected CFDs, because they're always going to be the same. I think this information wouldn't be very useful.

luckysori:
> Would you treat rejected separate or handle them as "failed" where we add a reason why it failed?
I think it might be good enough to treat all "failed" in the same table unless we want to store different data for failed/rejected.

I hadn't decided yet, but I think one table should suffice.

daemon/src/db/failed.rs
da-kami: > I would prefer not to add entries to the event_log table for failed or rejected CFDs, because they're always going to be the same. I think this information wouldn't be very useful.

I am not sure about that. In *most* cases it won't have value, but in some cases it might. There might be error scenarios where something goes into a weird failed state that should be analysed. It's not expensive to save these event log entries. 
If we decide to not keep an eventlog for failed at all we *definitely* need a timestamp in the table, because otherwise we don't know anything about the time the CFD "failed". I would say it might actually be valuable to know when contract setup started and when it failed. (Rejected might be interesting as well, because we have reasons in there - but that does not necessarily have to be stored in an event-log; the reason would not be persisted there...).

daemon/src/db/failed.rs
```diff
@@ -0,0 +1,556 @@
...
+}
+
+impl_sqlx_type_display_from_str!(Kind);
+
+impl Connection {
+    pub async fn move_to_failed_cfds(&self) -> Result<()> {
+        let ids = self.failed_cfd_ids_according_to_events().await?;
+
+        if !ids.is_empty() {
+            tracing::debug!("Moving ContractSetupFailed CFDs to failed_cfds table: {ids:?}");
```
da-kami:
```suggestion
            tracing::debug!("Moving failed CFDs to failed_cfds table: {ids:?}");
```

It includes both `ContractSetupFailed` and `OfferRejected`

daemon/src/db/failed.rs
```diff
@@ -0,0 +1,556 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct FailedCfd {
+    pub id: OrderId,
+    pub position: Position,
+    pub initial_price: Price,
+    pub taker_leverage: Leverage,
+    pub n_contracts: Contracts,
+    pub counterparty_network_identity: Identity,
+    pub role: Role,
+    pub fees: Fees,
```
da-kami, line 61: I remember we discussed this: The `fees` are added because we display them in the UI?
Maybe we should actually re-consider this. It is somewhat confusing and should always be `0`. I think the UI is also confusing for displaying it.

We don't have to block this PR by this change, can do in a follow-up.

daemon/src/db/failed.rs
```diff
@@ -0,0 +1,556 @@
...
+    let row = sqlx::query!(
+        r#"
+        SELECT
+            event_log_failed.created_at
+        FROM
+            event_log_failed
+        JOIN
+            failed_cfds on failed_cfds.id = event_log_failed.cfd_id
+        WHERE
+            failed_cfds.uuid = $1 AND (event_log_failed.name = $2 OR event_log_failed.name = $3)
```
da-kami: 🛠 I read the comment, but I am not sure why we cannot always `select` the first event in the log. I would prefer that over the current solution. I think an `ASC LIMIT 1` should do the job or am I missing something?

daemon/src/projection.rs
```diff
@@ -971,7 +974,83 @@ impl db::ClosedCfdAggregate for Cfd {
...
+        let liquidation_price = match position {
+            Position::Long => calculate_long_liquidation_price(our_leverage, initial_price),
+            Position::Short => calculate_short_liquidation_price(our_leverage, initial_price),
+        };
+
+        // there are no events to apply at this stage for closed CFDs,
+        // which is why this field is mostly ignored
+        let mut aggregated = Aggregated::new(FeeAccount::new(position, role));
+
+        // set the creation_timestamp to be able to sort closed CFDs
```
da-kami:
```suggestion
        // set the creation_timestamp to be able to sort failed CFDs
```

daemon/src/projection.rs
```diff
@@ -971,7 +974,83 @@ impl db::ClosedCfdAggregate for Cfd {
...
+        let margin = calculate_margin(initial_price, quantity_usd, our_leverage);
+        let margin_counterparty =
+            calculate_margin(initial_price, quantity_usd, counterparty_leverage);
+
+        let liquidation_price = match position {
+            Position::Long => calculate_long_liquidation_price(our_leverage, initial_price),
+            Position::Short => calculate_short_liquidation_price(our_leverage, initial_price),
+        };
+
+        // there are no events to apply at this stage for closed CFDs,
```
da-kami:
```suggestion
        // there are no events to apply at this stage for failed CFDs,
```

daemon/src/projection.rs
```diff
@@ -971,7 +974,83 @@ impl db::ClosedCfdAggregate for Cfd {
...
             counterparty: counterparty_network_identity,
             pending_settlement_proposal_price: None,
-            aggregated: empty_aggregated,
+            aggregated,
+            network,
+        }
+    }
+}
+
+impl db::FailedCfdAggregate for Cfd {
```
da-kami, line 982: ❓Can we potentially create a generic aggregate for closed and failed?

da-kami (approved):
Thanks for the great work! 
Nothing blocking per-se, but I think selecting the `creation_timestamp` for failed cfds could be changed.

daemon/src/db/failed.rs
luckysori: Fair point! I just wanted to keep the behaviour the same as it currently is in production for the time being.

daemon/src/db/failed.rs
luckysori: Fair enough, I did not consider taking the first event! I'll try that instead.

daemon/src/projection.rs
luckysori: I don't see it at the moment because we have completely different expectations for closed and failed CFDs. The data that is stored is different, particularly in terms of the transactions that are attached to closed CFDs (always a lock transaction and a choice of settlement transactions) and that are not attached to failed CFDs. 

We could define a model with optional fields, but I'm not sure I would recommend that added complexity.

daemon/src/db/closed.rs
```diff
@@ -0,0 +1,1365 @@
...
+    Ok(())
+}
+
+/// Obtain the time at which the closed CFD was created, according to
+/// the `event_log` table.
+///
+/// Every closed CFD must have gone through contract setup at some
+/// point. Therefore, we base the creation timestamp on the
+/// `EventKind::ContractSetupStarted` variant.
+async fn load_creation_timestamp(
```
da-kami, line 922: We could actually use the logic that is used for `failed_cfds` here as well. Can be done in follow-up.

da-kami (approved):

daemon/src/db/closed.rs
luckysori: Done with https://github.com/itchysats/itchysats/pull/1916/commits/38c75bf4010bccb508edf0a6e0da8ead1c93409e.

luckysori:
The recent rebase made me notice that perhaps there was a tiny bug, @bonomat. See https://github.com/itchysats/itchysats/pull/1916/commits/cb0e8137a974e5c42186d73bdda3a662f6c77e99.

daemon/src/db.rs
thomaseizinger: Yeah you would have to use the same connection likely. One thing that you might be able to do is do the mapping right on the stream that is being returned instead of collecting it all first. Doesn't make much of a difference though I think!

daemon/src/db.rs
```diff
@@ -431,6 +455,37 @@ impl Connection {
...
+            )
+            "#,
+            model::EventKind::OFFER_REJECTED,
+            model::EventKind::CONTRACT_SETUP_FAILED,
+        )
+        .fetch_all(&mut *conn)
+        .await?
+        .into_iter()
+        .map(|r| r.uuid)
+        .collect();
```
thomaseizinger: Here we could omit the `into_iter` part of we would use `fetch`.

daemon/src/db.rs
```diff
@@ -264,23 +264,7 @@ impl Connection {
...
-                "#
-            )
-            .fetch_all(&mut *conn)
-            .await?
-            .into_iter()
-            .map(|r| r.uuid);
-
-            drop(conn);
-
+            let ids = self.load_open_cfd_ids().await?;
```
thomaseizinger, line 267: We should probably document that callers cannot rely on only open IDs being returned here.

thomaseizinger (approved):
Excellent work @luckysori !

daemon/src/db.rs
luckysori: In the interest of getting this merged before I encounter more conflicts I vote for merging this as is and I can look at this suggestion in a follow-up PR.

daemon/src/db.rs
luckysori: I don't seem to be able to return the stream because `conn` is borrowed and dropped as soon as we return :(

```rust
// error[E0515]: cannot return value referencing local variable `conn`
//    --> daemon/src/db.rs:391:9
//     |
// 388 |         .fetch(&mut *conn)
//     |                      ---- `conn` is borrowed here
// ...
// 391 |         Ok(ids)
//     |         ^^^^^^^ returns a value referencing data owned by the current function

async fn closed_cfd_ids_according_to_the_blockchain(
    &self,
) -> Result<impl Stream<Item = Result<OrderId, sqlx::Error>>> {
    let mut conn = self.inner.acquire().await?;

    let ids = sqlx::query!(
        r#"
        select
            id as cfd_id,
            uuid as "uuid: model::OrderId"
        from
            cfds
        where exists (
            select id from EVENTS as events
            where events.cfd_id = cfds.id and
            (
                events.name = $1 or
                events.name = $2 or
                events.name= $3
            )
        )
        "#,
        EventKind::COLLABORATIVE_SETTLEMENT_CONFIRMED,
        EventKind::CET_CONFIRMED,
        EventKind::REFUND_CONFIRMED,
    )
    .fetch(&mut *conn)
    .map(|res| res.map(|r| r.uuid));

    Ok(ids)
}

```

luckysori:
bors r+

Going ahead with this now! I'm still looking at the reviews in case we learn something new.

luckysori:
bors retry

PR #1918, opened 2022-04-28
Fix linebreaks in FAQ

bonomat:
Markdown either needs an empty new line or a trailing `\` to enforce linebreaks

Before: 
image

After: 

image


klochowicz (approved):
thanks, that looks much nicer!

PR #1919, opened 2022-04-28
Introduce symmetric `/itchysats/close-position` protocol based on libp2p

### LINKED ISSUE #320: Enable the taker to publish the collaborative settlement tx

thomaseizinger:
Fixes #320.


da-kami:
After having a look at the current state here are some thoughts:

- When the taker starts the protocol he relies on the "old" `CollaborativeSettlementStarted` event that contains the `SettlementProposal`. Why? I would have seen a new even there. We can keep the logic of the old protocol for backwards compatibility. What is the motivation for re-using the old events? Simpler? I wonder if this does not make it more complicated in the end, because we have to design the new protocol to fit the old events.  
- My understanding is, that on the maker side we are now using one actor per protocol. This is an improvement to before, because protocols don't influence each other, but still that actor handles requests for one protocol by all takers. Would it make sense to go one step further and spin up an actor per protocol-request per taker?
I am not sure the abstraction of `pending_protocols` helps because we have two sources of truth: The CFD which has a state and the `pending_protocols` which has a state that *should* reflect the state of the CFD. Can we achieve to have the CFD as single source of truth and spin up actors per protocol per CFD? This might bring back problems where we spin up the same protocol for one CFD twice, but we could handle that in a different way.

klochowicz:
> After having a look at the current state here are some thoughts:
> 
> * When the taker starts the protocol he relies on the "old" `CollaborativeSettlementStarted` event that contains the `SettlementProposal`. Why? I would have seen a new even there. We can keep the logic of the old protocol for backwards compatibility. What is the motivation for re-using the old events? Simpler? I wonder if this does not make it more complicated in the end, because we have to design the new protocol to fit the old events.

We have the freedom to introduce the new events, but if there's no need to do so, it's best to avoid it. There was no need to introduce one in this case.

> * My understanding is, that on the maker side we are now using one actor per protocol. This is an improvement to before, because protocols don't influence each other, but still that actor handles requests for one protocol by all takers. Would it make sense to go one step further and spin up an actor per protocol-request per taker?

this sounds like a premature optimisation and introducing a whole new level of complexity when it's not needed. I would consider having that as an option if we're pressed for it. Until we actually can confirm that will be a problem I would prefer not to venture that way.

>   I am not sure the abstraction of `pending_protocols` helps because we have two sources of truth: The CFD which has a state and the `pending_protocols` which has a state that _should_ reflect the state of the CFD. Can we achieve to have the CFD as single source of truth and spin up actors per protocol per CFD? This might bring back problems where we spin up the same protocol for one CFD twice, but we could handle that in a different way.

avoiding `pending_protocols` state is an interesting idea - I'll have a think about it today!

klochowicz:
Closing this PR by Thomas, will re-open as myself.

PR #1921, opened 2022-04-28
Add position metrics to maker

bonomat:
Looks like this :
```
# HELP positions_quantity Total quantity of positions on ItchySats.
# TYPE positions_quantity gauge
positions_quantity{position="long",status="closed"} 0
positions_quantity{position="long",status="failed"} 0
positions_quantity{position="long",status="open"} 100
positions_quantity{position="long",status="refunded"} 0
positions_quantity{position="long",status="unknown"} 0
positions_quantity{position="short",status="closed"} 100
positions_quantity{position="short",status="failed"} 0
positions_quantity{position="short",status="open"} 0
positions_quantity{position="short",status="refunded"} 0
positions_quantity{position="short",status="unknown"} 0
# HELP positions_total Total number of positions on ItchySats.
# TYPE positions_total gauge
positions_total{position="long",status="closed"} 0
positions_total{position="long",status="failed"} 0
positions_total{position="long",status="open"} 1
positions_total{position="long",status="refunded"} 0
positions_total{position="long",status="unknown"} 0
positions_total{position="short",status="closed"} 1
positions_total{position="short",status="failed"} 0
positions_total{position="short",status="open"} 0
positions_total{position="short",status="refunded"} 0
positions_total{position="short",status="unknown"} 0
```

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -313,41 +313,28 @@ mod metrics {
     fn set_position_metrics<'a>(cfds: impl Iterator<Item = &'a Cfd>, status: &str) {
         let (long, short): (Vec<_>, Vec<_>) = cfds.partition(|cfd| cfd.position == Position::Long);
 
-        POSITION_QUANTITY_GAUGE
-            .with(&HashMap::from([
-                (POSITION_LABEL, POSITION_LONG_LABEL),
-                (STATUS_LABEL, status),
-            ]))
-            .set(
-                sum_amounts(&long)
-                    .into_decimal()
-                    .to_f64()
-                    .unwrap_or_default(),
-            );
-        POSITION_AMOUNT_GAUGE
-            .with(&HashMap::from([
-                (POSITION_LABEL, POSITION_LONG_LABEL),
-                (STATUS_LABEL, status),
-            ]))
-            .set(long.len() as i64);
+        set_metrics_for(POSITION_LONG_LABEL, status, &long);
+        set_metrics_for(POSITION_SHORT_LABEL, status, &short);
+    }
 
+    fn set_metrics_for(position_label: &str, status: &str, position: &[&Cfd]) {
         POSITION_QUANTITY_GAUGE
             .with(&HashMap::from([
-                (POSITION_LABEL, POSITION_SHORT_LABEL),
+                (POSITION_LABEL, position_label),
                 (STATUS_LABEL, status),
             ]))
             .set(
-                sum_amounts(&short)
+                sum_amounts(position)
                     .into_decimal()
                     .to_f64()
                     .unwrap_or_default(),
             );
         POSITION_AMOUNT_GAUGE
             .with(&HashMap::from([
-                (POSITION_LABEL, POSITION_SHORT_LABEL),
+                (POSITION_LABEL, position_label),
                 (STATUS_LABEL, status),
             ]))
-            .set(short.len() as i64);
+            .set(position.len() as i64);
     }
 
     fn sum_amounts(cfds: &[&Cfd]) -> Usd {
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -3,6 +3,7 @@ use async_trait::async_trait;
 use futures::StreamExt;
 use model::CfdEvent;
 use model::EventKind;
+use model::OrderId;
 use model::Position;
 use model::Usd;
 use rust_decimal::prelude::FromPrimitive;
@@ -70,6 +71,7 @@ struct UpdateMetrics;
 /// Read-model of the CFD for the position metrics actor.
 #[derive(Clone, Copy)]
 pub struct Cfd {
+    id: OrderId,
     position: Position,
     quantity_usd: Usd,
 
@@ -87,6 +89,7 @@ impl db::CfdAggregate for Cfd {
 
     fn new(_: Self::CtorArgs, cfd: db::Cfd) -> Self {
         Self {
+            id: cfd.id,
             position: cfd.position,
             quantity_usd: cfd.quantity_usd,
             is_open: false,
@@ -204,6 +207,7 @@ impl Cfd {
 impl db::ClosedCfdAggregate for Cfd {
     fn new_closed(_: Self::CtorArgs, closed_cfd: db::ClosedCfd) -> Self {
         let db::ClosedCfd {
+            id,
             position,
             n_contracts,
             ..
@@ -213,6 +217,7 @@ impl db::ClosedCfdAggregate for Cfd {
             Usd::new(Decimal::from_u64(u64::from(n_contracts)).expect("u64 to fit into Decimal"));
 
         Self {
+            id,
             position,
             quantity_usd,
 
@@ -243,6 +248,9 @@ mod metrics {
     const STATUS_FAILED_LABEL: &str = "failed";
     const STATUS_REJECTED_LABEL: &str = "rejected";
     const STATUS_REFUNDED_LABEL: &str = "refunded";
+    // this is needed so that we do not lose cfds which are in a weird state, e.g. open & closed.
+    // This should be 0 though but for the time being we add it
+    const STATUS_UNKNOWN_LABEL: &str = "unknown";
 
     static POSITION_QUANTITY_GAUGE: conquer_once::Lazy<prometheus::GaugeVec> =
         conquer_once::Lazy::new(|| {
@@ -276,6 +284,30 @@ mod metrics {
             cfds.iter().filter(|cfd| cfd.is_refunded),
             STATUS_REFUNDED_LABEL,
         );
+
+        set_position_metrics(
+            cfds.iter().filter(|cfd| {
+                let unknown_state = is_unknown(cfd);
+                if unknown_state {
+                    tracing::error!(
+                        is_open = cfd.is_open,
+                        is_closed = cfd.is_closed,
+                        is_refunded = cfd.is_refunded,
+                        is_rejected = cfd.is_rejected,
+                        order_id = %cfd.id,
+                        "CFD is in weird state"
+                    );
+                }
+                unknown_state
+            }),
+            STATUS_UNKNOWN_LABEL,
+        );
+    }
+
+    /// Return true if a CFD is in multiple states
+    fn is_unknown(cfd: &Cfd) -> bool {
+        !(cfd.is_open ^ cfd.is_closed ^ cfd.is_refunded ^ cfd.is_rejected ^ cfd.is_failed)
+            && (cfd.is_open || cfd.is_closed || cfd.is_refunded || cfd.is_rejected || cfd.is_failed)
     }
 
     fn set_position_metrics<'a>(cfds: impl Iterator<Item = &'a Cfd>, status: &str) {
@@ -321,5 +353,102 @@ mod metrics {
     fn sum_amounts(cfds: &[&Cfd]) -> Usd {
         cfds.iter()
             .fold(Usd::ZERO, |sum, cfd| cfd.quantity_usd + sum)
+    }
+
+    #[cfg(test)]
+    mod tests {
+        use super::*;
+
+        fn create_cfd(state_flags: StateFlags) -> Cfd {
+            Cfd {
+                id: Default::default(),
+                position: Position::Long,
+                quantity_usd: Usd::ZERO,
+                is_open: state_flags.open,
+                is_closed: state_flags.closed,
+                is_failed: state_flags.failed,
+                is_refunded: state_flags.refunded,
+                is_rejected: state_flags.rejected,
+                version: 0,
+            }
+        }
+
+        #[derive(Debug, Clone, Copy)]
+        struct StateFlags {
+            open: bool,
+            closed: bool,
+            failed: bool,
+            refunded: bool,
+            rejected: bool,
+        }
+
+        #[test]
+        fn when_cfd_is_in_single_no_state_then_is_not_unknown() {
+            let state_matrix = [
+                [false, false, false, false, false],
+                [true, false, false, false, false],
+                [false, true, false, false, false],
+                [false, false, true, false, false],
+                [false, false, false, true, false],
+                [false, false, false, false, true],
+            ];
+
+            for states in state_matrix {
+                let is_open = states[0];
+                let is_closed = states[1];
+                let is_failed = states[2];
+                let is_refunded = states[3];
+                let is_rejected = states[4];
+                let state_flags = StateFlags {
+                    open: is_open,
+                    closed: is_closed,
+                    failed: is_failed,
+                    refunded: is_refunded,
+                    rejected: is_rejected,
+                };
+                let cfd = create_cfd(state_flags);
+
+                assert!(
+                    !is_unknown(&cfd),
+                    "State combination was unknown. {state_flags:?}"
+                );
+            }
+        }
+
+        #[test]
+        fn when_cfd_is_in_multiple_states_then_is_unknown() {
+            let state_matrix = [
+                [true, true, false, false, false],
+                [true, false, true, false, false],
+                [true, false, false, true, false],
+                [true, false, false, false, true],
+                [false, true, true, false, false],
+                [false, true, false, true, false],
+                [false, true, false, false, true],
+                [false, false, true, true, false],
+                [false, false, true, false, true],
+                [false, false, false, true, true],
+            ];
+            for states in state_matrix {
+                let is_open = states[0];
+                let is_closed = states[1];
+                let is_failed = states[2];
+                let is_refunded = states[3];
+                let is_rejected = states[4];
+
+                let state_flags = StateFlags {
+                    open: is_open,
+                    closed: is_closed,
+                    failed: is_failed,
+                    refunded: is_refunded,
+                    rejected: is_rejected,
+                };
+                let cfd = create_cfd(state_flags);
+                assert!(
+                    is_unknown(&cfd),
+                    "State combination was not unknown. {state_flags:?}"
+                );
+            }
+        }
     }
 }
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -220,6 +220,7 @@ impl db::ClosedCfdAggregate for Cfd {
             is_closed: true,
             is_failed: false,
             is_refunded: false,
+            is_rejected: false,
             version: 0,
         }
     }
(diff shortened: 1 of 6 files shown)


daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,296 @@
...
+        self.apply(event)
+    }
+
+    fn version(&self) -> u32 {
+        self.version
+    }
+}
+
+impl Cfd {
+    fn apply(mut self, event: CfdEvent) -> Self {
```
klochowicz, line 156: not an issue with this review, but a meta-comment:
it is a bit concerning that we're spreading the business logic in so many places, but that's one of the fall-out of using multiple aggregates in our take on ES. not sure what to do about it though.

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+use model::Usd;
+use rust_decimal::prelude::FromPrimitive;
+use rust_decimal::Decimal;
+use std::time::Duration;
+use tokio_tasks::Tasks;
+use xtra_productivity::xtra_productivity;
+use xtras::SendInterval;
+
+// TODO: ideally this would be more often
+pub const UPDATE_METRIC_INTERVAL_SECONDS: Duration = Duration::from_secs(60);
```
klochowicz:
```suggestion
pub const UPDATE_METRIC_INTERVAL: Duration = Duration::from_secs(60);
```
(and roll-out where used)

nit: it's already a `Duration`, so it's not in "seconds"

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+    async fn handle(&mut self, _: UpdateMetrics) {
+        tracing::debug!("Collecting metrics");
+        let mut stream = self.db.load_all_cfds::<Cfd>(());
+
+        let mut cfds = Vec::new();
+        while let Some(cfd) = stream.next().await {
+            let cfd = match cfd {
+                Ok(cfd) => cfd,
+                Err(e) => {
+                    tracing::error!("Failed to rehydrate CFD: {e:#}");
```
klochowicz, line 59: perhaps it would be a useful metric to know how many times we failed to rehydrate a CFD?

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+            ContractSetupCompleted { .. } | LockConfirmed | LockConfirmedAfterFinality => Self {
+                is_open: true,
+                ..self
+            },
+            ContractSetupFailed => Self {
+                is_failed: true,
+                ..self
+            },
+            OfferRejected => Self {
+                is_failed: true,
```
klochowicz: have you considered distinguishing failed vs rejected here?

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+    fn new_closed(_: Self::CtorArgs, closed_cfd: db::ClosedCfd) -> Self {
+        let db::ClosedCfd {
+            id,
+            position,
+            n_contracts,
+            ..
+        } = closed_cfd;
+
+        let quantity_usd =
+            Usd::new(Decimal::from_u64(u64::from(n_contracts)).expect("u64 to fit into Decimal"));
```
klochowicz: nit: it's not the first time this one is popping up, perhaps we could write a `TryFrom<u64> for Decimal` ?

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+    const POSITION_SHORT_LABEL: &str = "short";
+
+    const STATUS_LABEL: &str = "status";
+    const STATUS_OPEN_LABEL: &str = "open";
+    const STATUS_CLOSED_LABEL: &str = "closed";
+    const STATUS_FAILED_LABEL: &str = "failed";
+    const STATUS_REFUNDED_LABEL: &str = "refunded";
+    // this is needed so that we do not lose cfds which are in a weird state, e.g. open & closed.
+    // This should be 0 though but for the time being we add it
+    const STATUS_UNKNOWN_LABEL: &str = "unknown";
```
klochowicz, line 295: makes total sense. we can create separate grafana alert if it's ever not 0.

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+use model::Position;
+use model::Usd;
+use rust_decimal::prelude::FromPrimitive;
+use rust_decimal::Decimal;
+use std::time::Duration;
+use tokio_tasks::Tasks;
+use xtra_productivity::xtra_productivity;
+use xtras::SendInterval;
+
+// TODO: ideally this would be more often
```
klochowicz: do you suspect it might impact performance significantly?

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+        POSITION_AMOUNT_GAUGE
+            .with(&HashMap::from([
+                (POSITION_LABEL, POSITION_SHORT_LABEL),
+                (STATUS_LABEL, status),
+            ]))
+            .set(short.len() as i64);
+    }
+
+    fn sum_amounts(cfds: &[&Cfd]) -> Usd {
+        cfds.iter()
```
klochowicz, line 385: it would be interesting whether it would make sense to employ `rayon` here as we scale the system up, with its `par_iter`...

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+                    .to_f64()
+                    .unwrap_or_default(),
+            );
+        POSITION_AMOUNT_GAUGE
+            .with(&HashMap::from([
+                (POSITION_LABEL, POSITION_SHORT_LABEL),
+                (STATUS_LABEL, status),
+            ]))
+            .set(short.len() as i64);
+    }
```
klochowicz: extracting the repetition made it much easier to follow;

```suggestion
 
        set_metrics_for(POSITION_LONG_LABEL, status, &long);
        set_metrics_for(POSITION_SHORT_LABEL, status, &short);
    }

    fn set_metrics_for(position_label: &str, status: &str, position: &[&Cfd]) {
        POSITION_QUANTITY_GAUGE
            .with(&HashMap::from([
                (POSITION_LABEL, position_label),
                (STATUS_LABEL, status),
            ]))
            .set(
                sum_amounts(position)
                    .into_decimal()
                    .to_f64()
                    .unwrap_or_default(),
            );
        POSITION_AMOUNT_GAUGE
            .with(&HashMap::from([
                (POSITION_LABEL, position_label),
                (STATUS_LABEL, status),
            ]))
            .set(position.len() as i64);
    }
```

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+                "Total quantity of positions on ItchySats.",
+                &[POSITION_LABEL, STATUS_LABEL]
+            )
+            .unwrap()
+        });
+
+    static POSITION_AMOUNT_GAUGE: conquer_once::Lazy<prometheus::IntGaugeVec> =
+        conquer_once::Lazy::new(|| {
+            prometheus::register_int_gauge_vec!(
+                "positions_total",
```
klochowicz:
```suggestion
                "positions_number",
```

`positions_total_number` would work too if you prefer that.

"total" is a bit ambiguous, I actually thought it referred to the quantities.

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+        while let Some(cfd) = stream.next().await {
+            let cfd = match cfd {
+                Ok(cfd) => cfd,
+                Err(e) => {
+                    tracing::error!("Failed to rehydrate CFD: {e:#}");
+                    continue;
+                }
+            };
+            cfds.push(cfd);
+        }
```
klochowicz: I suspect this would be more idiomatic (and faster):

```suggestion
              let stream = self.db.load_all_cfds::<Cfd>(());

        let cfds: Vec<Cfd> = stream
            .inspect(|cfd| {
                if let Err(e) = cfd {
                    tracing::error!("Failed to rehydrate CFD: {e:#}");
                }
            })
            .filter_map(|cfd| async move { cfd.ok() })
            .collect()
            .await;

```

klochowicz (approved):
thanks for reworking this, looks good!

daemon/src/position_metrics.rs
thomaseizinger: Similar to projection, you could ping this actor a message every time an event gets appended to a particular CFD and then update the metrics accordingly.

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,344 @@
...
+                        "CFD is in weird state"
+                    );
+                }
+                unknown_state
+            }),
+            STATUS_UNKNOWN_LABEL,
+        );
+    }
+
+    fn set_position_metrics(cfds: Filter<Iter<Cfd>, fn(&&Cfd) -> bool>, status: &str) {
```
thomaseizinger:
```suggestion
    fn set_position_metrics(cfds: impl Iterator<Item = &Cfd>, status: &str) {
```

This should do the trick as well.

daemon/src/position_metrics.rs
thomaseizinger: I doubt it. The overhead of moving things to a different thread likely exceeds the duration of the actual addition.

thomaseizinger (commented):
Thanks for reworking this!

I think similar to projection, we could have a local cache here where we store all aggregates, only update the one that changes every time something runs through the process manager and then re-compute the metrics every time based on all CFDs. That would remove the need for doing it on an interval.

What do you think?

daemon/src/position_metrics.rs
klochowicz: eventually, it would be nice if the actors that are concerned about this situation could subscribe to the process manager themselves and got notified when it happens (so that process manager doesn't need to be aware of projection / metrics / whatever, just dispatch a message that could be handled by different actors. 
Some sort of Observer pattern, but with actors :)

thomaseizinger: Other than aesthetics, is there a need to decouple the process manager from this knowledge? Making the observation dynamic also has downsides: It is harder to understand, who gets notified when. At least at the moment, one can trace the source code to see where messages land.

If the process manager is merely iterating through a list of `MessageChannel<CfdChanged>`, this won't be as easy.

klochowicz: this would decouple the software components - low-level layers should not know about the high-level ones.
in my opinion it would make things less error-prone and more modular.
As for understanding - is it hard to understand or follow code for classic OO observers? it's quite easy in an IDE to look at usages, in this case, which actors handle the given message.

thomaseizinger: > this would decouple the software components - low-level layers should not know about the high-level ones.

I agree with that in general but it is also important to remember that decoupling per se isn't the end goal! Decoupling is for example useful if two components want/need to change at a different pace. Both, process manager and projection primarily process events meaning they are already semantically coupled.

Plus, the fact that we want to update the projection state every time a new event gets saved couples them from a requirements perspective.

> As for understanding - is it hard to understand or follow code for classic OO observers?

It is certainly hard_er_ than components which are wired together explicitly.

I am not necessarily opposed to the idea of OO-style observers. I am simply pointing out that if the nature of the problem does not require dynamic subscribing and unsubscribing, we are introducing a bit more complexity than fundamentally needed. In other words, there could be a simpler solution and not going for that one might confuse future readers.

There is a quote I like which says: 
["All problems in computer science can be solved by another level of indirection."](https://en.wikipedia.org/wiki/Indirection#cite_note-1)

I think we currently don't have a problem with the process manager knowing about projection so the cost is hard to justify IMO.

daemon/src/position_metrics.rs
bonomat: Totally agree with this statement.

daemon/src/position_metrics.rs
bonomat: I don't think this is possible: you can only implement traits on structs from the same crate.

daemon/src/position_metrics.rs
bonomat: I'm not sure if a metric for this is suitable. We'd need to add a label for the cfd id to the metric which does not scale well (and is not recommended )
We better leave a log here.

daemon/src/position_metrics.rs
bonomat: Not yet, but can add it.

daemon/src/position_metrics.rs
bonomat: mhmm.. not sure about that. 
I tried to follow the naming convention here: https://prometheus.io/docs/practices/naming/

daemon/src/position_metrics.rs
bonomat: Yes, I think so because we go against the db for every call.

daemon/src/position_metrics.rs
klochowicz: how about `positions_number_total` ? that's less ambiguous, and shouldn't really get into the way of the naming practices?

bonomat:
> Thanks for reworking this!
> 
> I think similar to projection, we could have a local cache here where we store all aggregates, only update the one that changes every time something runs through the process manager and then re-compute the metrics every time based on all CFDs. That would remove the need for doing it on an interval.
> 
> What do you think?

Done in [`c9aa83b` (#1921)](https://github.com/itchysats/itchysats/pull/1921/commits/c9aa83ba3da81d241f79886ddd94cd302e3ddf6d). The downside of this is that now the metrics logic gets executed for the taker as well because we have the same process manager.

daemon/src/position_metrics.rs
thomaseizinger: We already started to mitigate this in `FeeAccount` and can try to approach other situations the same way:

Create smaller models that represents little state machines which can be fed a list of values extracted from events which can be shared between different aggregates. For example, you could introduce a separate `CfdState` struct that can build itself up from a list of events and that one can be reused across multiple aggregates.

This is idea is also discussed in https://signalsandthreads.com/state-machine-replication-and-why-you-should-care/.

thomaseizinger:
> Done in [`c9aa83b` (#1921)](https://github.com/itchysats/itchysats/pull/1921/commits/c9aa83ba3da81d241f79886ddd94cd302e3ddf6d). The downside of this is that now the metrics logic gets executed for the taker as well because we have the same process manager.

We could either plug in a dummy actor or make the message channel optional if we don't want that. I don't see it as a big issue though.

Just on Friday I had locally added a metrics endpoint to the taker because I wanted to inspect the ping latency stats gathered by the new `xtra-libp2p-ping` protocol so maybe we will just end up exposing metrics on the taker as well at some point?

daemon/src/position_metrics.rs
```diff
@@ -0,0 +1,301 @@
+use crate::db;
+use async_trait::async_trait;
+use futures::StreamExt;
+use model::CfdEvent;
+use model::EventKind;
+use model::Position;
+use model::Usd;
```
thomaseizinger: Note on modularity: If the `db` ends up being extracted as a separate crate, this actor could also be a separate one (by depending on only `db` and `model`).

The compile-times of the workspace will continue growing if we continue to just add stuff to `daemon` and not try and modularise it into individual crates.

daemon/src/position_metrics.rs
thomaseizinger: `Decimal` as a `From<u64>` implementation: https://docs.rs/rust_decimal/1.23.1/rust_decimal/struct.Decimal.html#impl-From%3Cu64%3E

daemon/src/position_metrics.rs
```diff
@@ -322,4 +354,94 @@ mod metrics {
...
+                [false, false, false, false, true],
+            ];
+
+            for states in state_matrix {
+                let is_open = states[0];
+                let is_closed = states[1];
+                let is_failed = states[2];
+                let is_refunded = states[3];
+                let is_rejected = states[4];
+                let cfd = create_cfd(is_open, is_closed, is_failed, is_refunded, is_rejected);
```
thomaseizinger: I think this would be more readable if you would define a (test-only) struct which has 5 boolean fields that you can use to initialize your `Cfd` aggregate with:

```rust
struct StateFlags {
	open: bool,
	closed: bool,
	failed: bool,
	refunded: bool,
	rejected: bool
}
```

Then your state matrix definition can be:

```rust
let state_matrix = [
	StateFlags { open: false, closed: false, failed: false, refunded: false, rejected: false },
	// ...
];
```

daemon/src/position_metrics.rs
```diff
@@ -8,24 +10,29 @@ use model::Position;
...
 use rust_decimal::Decimal;
+use std::collections::HashMap;
 use std::time::Duration;
-use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
-use xtras::SendInterval;
+use xtras::SendAsyncSafe;
 
 // TODO: ideally this would be more often
 pub const UPDATE_METRIC_INTERVAL: Duration = Duration::from_secs(60);
```
thomaseizinger: This should now be unused right? If it would have been private from the beginning, rustc would probably flag it as `dead_code` now!

thomaseizinger: The TODO is also stale I believe!

bonomat:
> We could either plug in a dummy actor or make the message channel optional if we don't want that. I don't see it as a big issue though.

We could do that. Don't you think this could confuse devs?

> Just on Friday I had locally added a metrics endpoint to the taker because I wanted to inspect the ping latency stats gathered by the new xtra-libp2p-ping protocol so maybe we will just end up exposing metrics on the taker as well at some point?

Fair point. It might not be bad idea after all ...

thomaseizinger (approved):
Thanks!

A few minor comments!
If it doesn't cause too much friction, would you mind squashing the last two commits into the first one?

thomaseizinger:
> > We could either plug in a dummy actor or make the message channel optional if we don't want that. I don't see it as a big issue though.
> 
> We could do that. Don't you think this could confuse devs?

Borderline trade-off for me.

Now that I think of it, if we don't want metrics in the `taker` daemon, we should probably introduce a `metrics` feature-flag in `daemon` and not activate it in the `taker` binary.

daemon/src/position_metrics.rs
thomaseizinger: You could even do things like:

```rust
StateFlags {
	open: true,
	..StateFlags::all_false()
}
```

daemon/src/position_metrics.rs
bonomat: Should be better now that it is updated for each new event

daemon/src/position_metrics.rs
bonomat: Done

daemon/src/position_metrics.rs
bonomat: I just tried it but I think it is easier with a matrix to see that all combinations are covered. I'll do a mix.

bonomat:
> > > We could either plug in a dummy actor or make the message channel optional if we don't want that. I don't see it as a big issue though.
> > 
> > 
> > We could do that. Don't you think this could confuse devs?
> 
> Borderline trade-off for me.
> 
> Now that I think of it, if we don't want metrics in the `taker` daemon, we should probably introduce a `metrics` feature-flag in `daemon` and not activate it in the `taker` binary.

I think taker metrics can come handy for our testnet setup so I created a follow-up ticket for this: #1946 .

klochowicz (approved):
LGTM!

PR #1922, opened 2022-04-28
Add FAQ link to the Footer

### LINKED ISSUE #1898: Add FAQ section to website and add link to taker-ui
# FAQ

- a blog entry explaining most confusing parts of the system
- link in the corner of the app, so it's easy to spot for newcomers
- can get updated "as we go", when users are asking us a question that would be obvious to us, but not to a first-time user
- can also point to some popular knowledge bases where people can learn about CFDs etc.

Example:
Q: Can I use my existing wallet / umbrel wallet?
A: Not yet. We've got it on our roadmap, for the time being the app has its own inbuilt wallet into which you can transfer funds. It is available in the "Wallet" tab.

_Originally posted by @klochowicz in https://github.com/itchysats/itchysats/discussions/1841#discussioncomment-2598915_

klochowicz:
It seems like the Footer is the most natural place after all to place the FAQ link.


da-kami (approved):
Sufficient until we have the "tour" - I might add references to the FAQ in the tour as well, if it makes sense.

PR #1926, opened 2022-04-29
Do not apply current quote to a closed `projection::Cfd`

luckysori:
Fixes a bug mentioned here: https://github.com/itchysats/itchysats/pull/1916#discussion_r860547648.

When the `closed_cfds` DB table was introduced, we added a test to check that the changes had not affected `projection`.

Unfortunately, we missed the fact that all CFDs, whether loaded on the initialisation of the `projection::Actor` or updated on reception of a `CfdChanged` message, actually go through the `with_current_quote` method. Originally we thought that closed CFDs, which can never receive updates, did not go through `with_current_quote`.

This caused a bug because `with_current_quote` overwrites the `profit_btc`, `profit_percent` and `payout` fields. These are already set correctly when closed CFDs are loaded from the database.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -556,6 +556,11 @@ impl Cfd {
     }
 
     pub fn with_current_quote(self, latest_quote: Option<xtra_bitmex_price_feed::Quote>) -> Self {
+        // Closed CFDs should not be modified by the current quote
+        if self.aggregated.state == CfdState::Closed {
+            return self;
+        }
+
         // If we have a dedicated closing price, use that one.
         if let Some(payout) = self.aggregated.clone().payout(self.role) {
             let payout = payout
--- a/daemon/src/db.rs
+++ b/daemon/src/db.rs
@@ -1728,18 +1728,23 @@ mod tests {
             .await
             .unwrap();
 
-        let projection_open = db
-            .load_open_cfd::<crate::projection::Cfd>(order_id, bdk::bitcoin::Network::Testnet)
-            .await
-            .unwrap();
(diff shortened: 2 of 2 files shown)


thomaseizinger (approved):
Putting everything in one model in `projection` was really quite a bad idea back when I implemented it :(

luckysori:
> Putting everything in one model in `projection` was really quite a bad idea back when I implemented it :(

I agree that something doesn't quite work in `projection`.

Do you already have some alternative ideas on how we could refactor this code?

daemon/src/projection.rs
```diff
@@ -556,6 +556,11 @@ impl Cfd {
     }
 
     pub fn with_current_quote(self, latest_quote: Option<xtra_bitmex_price_feed::Quote>) -> Self {
+        // Closed CFDs should not be modifited by the current quote
```
klochowicz: nit: typo

```suggestion
        // Closed CFDs should not be modified by the current quote
```

klochowicz (approved):

thomaseizinger:
> Do you already have some alternative ideas on how we could refactor this code?

Split out the serialization model from the aggregate then you have a much cleaner mapping where the aggregate can first come to a decision on what state it is in and in the 2nd step, you inspect that state and de-normalize it for serialization.

The problem now is that we are doing both in one step which blurs, which state should be used for reasoning and what is meant for the client. For example, we derive a state called `CfdState` which by nature of being derived is less suitable for decision making than the actual source of thruth. Later, we use the derived state to make more decisions. This has led and will lead to more bugs :)

PR #1927, opened 2022-04-29
Disallow sending TxFeeRate=0

klochowicz:
Catch all 422 errors in a bit nicer message.


klochowicz:
the improved error message (it's not perfect, but it's a huge improvement over the one in superceded PR)
image

shared-bin/src/catchers.rs
```diff
@@ -0,0 +1,21 @@
...
+/// Emit HttpApiProblem whenever a Unprocessable Entity error (422) happens
+///
+/// Rocket emits this error when deserialised JSON is syntactically valid but
+/// it's not valid *semantically* (e.g. when passed 0 to a NonZerou32).
+/// Display this as error 400 as 422 is Web-Dav specific.
+#[rocket::catch(422)]
+pub fn unprocessable_entity(req: &Request) -> HttpApiProblem {
+    HttpApiProblem::new(StatusCode::BAD_REQUEST)
+        .title("Bad Request")
+        .detail(format!("{}", req.uri()))
```
thomaseizinger, line 15: Fundamentally, the client should know which URI it was hitting.

If we want to display the URI in the error message, I'd suggest we add that client side or leave it out altogether.

Perhaps we can say "Invalid JSON" or something?

taker/src/main.rs
```diff
@@ -383,9 +384,9 @@ async fn main() -> Result<()> {
                 routes::post_withdraw_request,
             ],
         )
-        .register("/api", rocket::catchers![rocket_basicauth::unauthorized])
+        .register("/api", default_catchers())
```
thomaseizinger, line 387: Does this indirection actually add value?

taker/src/main.rs
```diff
@@ -383,9 +384,9 @@ async fn main() -> Result<()> {
...
         .mount("/", rocket::routes![routes::dist, routes::index])
-        .register("/", rocket::catchers![rocket_basicauth::unauthorized])
+        .register("/", default_catchers())
```
thomaseizinger, line 389: I think this can stay as is, we don't need the 422 catcher to serve the static assets, we only need it for the API.

thomaseizinger (approved):
I am not convinced that the direction of bundling all catchers together actually provides value but that is not a blocking concern.

taker/src/main.rs
klochowicz: I agree with your reasoning that it might not be strictly necessary.
However, it's not wrong either - if 422 will not happen, then nothing gets caught.
It's easier to have them aligned than to always re-think whether it was needed or not.

taker/src/main.rs
klochowicz: yup, this shows that we use the same catchers on both sides, otherwise you need manual inspection to prove it. It is less risk of maintenance burden, less likely to forget to add a catcher on the other side when it gets added in one place.

shared-bin/src/catchers.rs
klochowicz: As a user, it was helpful to see the "/api/offer" there. Eventually I'd like to highlight the problem in the particular field, but we don't have/support field validation at this stage unfortunately. 

Not sure whether 'Invalid JSON' would be better for the clients, showing this in the UI would confuse the users even more?

luckysori (approved):
Thanks for considering https://github.com/itchysats/itchysats/pull/1927/commits/73e67e89eaac1562d26a0d081f17f5f417fbe6e6!

No strong feelings about https://github.com/itchysats/itchysats/pull/1927/commits/f354c9228a51fa5abc45347cb64d53a222a08ce6.

klochowicz:
bors r+

> Thanks for considering [73e67e8](https://github.com/itchysats/itchysats/commit/73e67e89eaac1562d26a0d081f17f5f417fbe6e6)!
> 
> No strong feelings about [f354c92](https://github.com/itchysats/itchysats/commit/f354c9228a51fa5abc45347cb64d53a222a08ce6).

no worries, thank you for contributing it! I hope to add validators in due time, to make it even nicer experience (so that the user knows which field breaks the request...).

Issue #1904, opened 2022-04-27
"Long BTC" and "Short BTC" buttons in taker-frontend reload the entire page


thomaseizinger:
I believe this must be some bug in the UI code. We have client-side routing and thus should not need to reload the entire page.

bonomat:
Was this on master? I can't reproduce this on 0.4.12

thomaseizinger:
I'll try to reproduce this!

bonomat:
Not reproducible

PR #1931, opened 2022-04-29
User interface improvements

da-kami:
Adapted the user interface to be more intuitive and scale better

image

After having trouble with the hamburger menu when working on the guided tour @klochowicz and me took a step back and concluded that it was never intuitive on the first place (at least on desktop), so it was replaced with a switch button  between `home` and `wallet`.
As part of changing that the top-bar was split into two sections where one is for navigation and controlls and the other one is information (i.e. maker status, funding, reference price).

The new layout locks a viewport in the middle. We can consider using the `gray.200` area for a branding pattern (was somewhat inspired by https://www.uxmatters.com/mt/archives/2020/07/designing-mobile-tables.php)

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -51,6 +51,12 @@ export interface Offer {
     leverage: number;
 }
 
+// TODO: Evaluate moving these globals into the theme to make them accessible through that
+export const VIEWPORT_WIDTH = 1000;
+export const VIEWPORT_WIDTH_PX = VIEWPORT_WIDTH + "px";
+export const BG_LIGHT = "gray.50";
+export const BG_DARK = "gray.800";
+
 export const App = () => {
     const toast = useToast();
 
@@ -169,68 +175,78 @@ export const App = () => {
                 nextFundingEvent={nextFundingEvent}
                 referencePrice={referencePrice}
             />
-            <Box textAlign="center" padding={3} bg={useColorModeValue("gray.50", "gray.800")}>
-                <Center marginTop={20}>
-                    <Routes>
-                        <Route path="/">
-                            <Route
-                                path="wallet"
-                                element={
-                                    <>
-                                        <Wallet walletInfo={walletInfo} />
-                                    </>
-                                }
-                            />
-                            <Route
-                                element={
-                                    // @ts-ignore: ts-lint thinks that {children} is missing but react router is taking care of this for us
-
-
-                                        <PageLayout
-                                            cfds={cfds}
-                                            connectedToMaker={connectedToMaker}
-                                            showPromoBanner={isWithinPromoPeriod}
-                                        />
-
-                                }
+            <Box>
+                <Center>
+                    <Box
+                        maxWidth={(VIEWPORT_WIDTH + 200) + "px"}
+                        width={"100%"}
+                        height={"100%"}
+                        bgGradient={useColorModeValue(
+                            "linear(to-r, white 5%, gray.800, white 95%)",
+                            "linear(to-r, gray.800, white, gray.800)",
+                        )}
+                    >
+                        <Center>
+                            <Box
+                                textAlign="center"
+                                padding={3}
+                                bg={useColorModeValue(BG_LIGHT, BG_DARK)}
+                                maxWidth={VIEWPORT_WIDTH_PX}
+                                width={"100%"}
+                                marginTop={"100px"}
                             >
-                                <Route
-                                    path="long"
-                                    element={
-                                        <>
-                                            <Trade
-                                                offer={longOffer}
-                                                connectedToMaker={connectedToMaker}
-                                                walletBalance={walletInfo ? walletInfo.balance : 0}
-                                                isLong={true}
+                                <Routes>
+                                    <Route path="/">
+                                        <Route
+                                            path="wallet"
+                                            element={<Wallet walletInfo={walletInfo} />}
+                                        />
+                                        <Route
+                                            element={
+                                                // @ts-ignore: ts-lint thinks that {children} is missing but react router is taking care of this for us
+
+
+                                                    <PageLayout
+                                                        cfds={cfds}
+                                                        connectedToMaker={connectedToMaker}
+                                                        showPromoBanner={isWithinPromoPeriod}
+                                                    />
+
+                                            }
+                                        >
+                                            <Route
+                                                path="long"
+                                                element={
+                                                    <Trade
+                                                        offer={longOffer}
+                                                        connectedToMaker={connectedToMaker}
+                                                        walletBalance={walletInfo ? walletInfo.balance : 0}
+                                                        isLong={true}
+                                                    />
+                                                }
                                             />
-                                        </>
-                                    }
-                                />
-                                <Route
-                                    path="short"
-                                    element={
-                                        <>
-                                            <Trade
-                                                offer={shortOffer}
-                                                connectedToMaker={connectedToMaker}
-                                                walletBalance={walletInfo ? walletInfo.balance : 0}
-                                                isLong={false}
+                                            <Route
+                                                path="short"
+                                                element={
+                                                    <Trade
+                                                        offer={shortOffer}
+                                                        connectedToMaker={connectedToMaker}
+                                                        walletBalance={walletInfo ? walletInfo.balance : 0}
+                                                        isLong={false}
+                                                    />
+                                                }
                                             />
-                                        </>
-                                    }
-                                />
-                            </Route>
-                            <Route index element={<Navigate to="long" />} />
-                        </Route>
-                        <Route
-                            path="/*"
-                            element={
-                                <>
-                                </>
-                            }
-                        />
-                    </Routes>
+                                        </Route>
+                                        <Route index element={<Navigate to="long" />} />
+                                    </Route>
+                                    <Route
+                                        path="/*"
+                                        element={<Navigate to="long" />}
+                                    />
+                                </Routes>
+                            </Box>
+                        </Center>
+                    </Box>
                 </Center>
             </Box>
             <Footer />
@@ -273,7 +289,7 @@ function HistoryLayout({ cfds, connectedToMaker }: HistoryLayoutProps) {
 
             {closedPositions.length > 0
                 && (
-                    <Accordion allowToggle width={"80%"}>
+                    <Accordion allowToggle maxWidth={VIEWPORT_WIDTH_PX} width={"100%"}>
                         <AccordionItem>
                             <h2>
                                 <AccordionButton>
--- a/taker-frontend/src/components/NavBar.tsx
+++ b/taker-frontend/src/components/NavBar.tsx
@@ -1,26 +1,27 @@
-import { HamburgerIcon, MoonIcon, SunIcon } from "@chakra-ui/icons";
+import { Icon, MoonIcon, SunIcon, WarningIcon } from "@chakra-ui/icons";
 import {
     Box,
     Button,
+    ButtonGroup,
+    Center,
     Divider,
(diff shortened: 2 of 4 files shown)


da-kami:
I know this is not super important yet, but this change also improves mobile display again:

image

da-kami:
Here you see the problems I run into when navigating within the tour.

The initial step is attached correctly according to the item attached. For the second step we navigate to the wallet. That step is already not attached correctly, but is displayed in the middle. 
Before the third step we navigate back to home, but the step is displayed in the left corner (it is actually attached to the `long` button - if the navigation is removed then the one on the long button works).

thomaseizinger:
> Here you see the problems I run into when navigating within the tour.

This might be related: https://github.com/itchysats/itchysats/issues/1904

taker-frontend/src/App.tsx
```diff
@@ -169,68 +175,92 @@ export const App = () => {
...
-                                                isLong={true}
+                                <Routes>
+                                    <Route path="/">
+                                        <Route
+                                            path="wallet"
+                                            element={
+                                                <>
+                                                    <Wallet walletInfo={walletInfo} />
+                                                </>
+                                            }
```
thomaseizinger: The fragment here seems redundant.

```suggestion
                                            element={<Wallet walletInfo={walletInfo} />}
```

taker-frontend/src/App.tsx
```diff
@@ -169,68 +175,92 @@ export const App = () => {
...
+                                                path="long"
+                                                element={
+                                                    <>
+                                                        <Trade
+                                                            offer={longOffer}
+                                                            connectedToMaker={connectedToMaker}
+                                                            walletBalance={walletInfo ? walletInfo.balance : 0}
+                                                            isLong={true}
+                                                        />
+                                                    </>
```
thomaseizinger: Same here.

taker-frontend/src/App.tsx
```diff
@@ -169,68 +175,92 @@ export const App = () => {
...
-                                        <>
-                                            <Trade
-                                                offer={shortOffer}
-                                                connectedToMaker={connectedToMaker}
-                                                walletBalance={walletInfo ? walletInfo.balance : 0}
-                                                isLong={false}
+                                            <Route
+                                                path="short"
+                                                element={
+                                                    <>
```
thomaseizinger: And here.

taker-frontend/src/App.tsx
```diff
@@ -169,68 +175,92 @@ export const App = () => {
...
+                                        </Route>
+                                        <Route index element={<Navigate to="long" />} />
+                                    </Route>
+                                    <Route
+                                        path="/*"
+                                        element={
+                                            <>
+                                            </>
+                                        }
+                                    />
```
thomaseizinger: What does this do? Render nothing? :joy:

taker-frontend/src/components/NavBar.tsx
```diff
@@ -51,100 +55,190 @@ export default function Nav({ walletInfo, connectedToMaker, nextFundingEvent, re
...
+                                            const radio = getRadioProps({ value });
+                                            return (
+                                                <RadioCard key={value} {...radio}>
+                                                    <HStack>
+                                                        {value === "wallet"
+                                                            ? <FaWallet />
+                                                            : <FaHome />}
+                                                        <Text>{value}</Text>
+                                                    </HStack>
+                                                </RadioCard>
```
thomaseizinger: I _think_ that from an accessibility PoV, this is not very good. These should be anchor elements (`<a>`). You can style does as radio buttons but semantically, they should be anchors.

taker-frontend/src/components/NavBar.tsx
```diff
@@ -51,100 +55,190 @@ export default function Nav({ walletInfo, connectedToMaker, nextFundingEvent, re
...
+                                <Flex justifyContent={"flex-left"}>
+                                    <HStack {...group}>
+                                        {options.map((value) => {
+                                            const radio = getRadioProps({ value });
+                                            return (
+                                                <RadioCard key={value} {...radio}>
+                                                    <HStack>
+                                                        {value === "wallet"
+                                                            ? <FaWallet />
+                                                            : <FaHome />}
```
thomaseizinger: This won't scale with more elements in `options` which makes this code brittle to changes elsewhere. It wouldn't be too hard to just do a clean mapping from `value` to icon and it would make this code "complete".

thomaseizinger (commented):
Looking at https://codesandbox.io/s/introjs-react-router-push-vj1jc?file=/pages/Dashboard.jsx, I think you will have to solve the tour thing by:

- Keeping the tour state of "isActive" global, i.e. outside of the navigation (where your current steps are defined)
- Have each page have its own little tour and pass the tour state in

This way, the tour of a sub page will be active only if you are "coming from a tour" but at the same time, the elements should be selected correctly.

klochowicz:
> I know this is not super important yet, but this change also improves mobile display again:
> 
> image

I absolutely love the new layout, you've taken it to a whole new level since we paired up on Friday. much more discoverable.

da-kami:
I decided to make this PR about the user interface changes only. The "tour" still needs discussion.
We can decide on these changes without taking the tour into account.

taker-frontend/src/App.tsx
da-kami: yes:

image

da-kami: Not sure if this should be changed as part of this PR though. 
We would have to decide what behavior we want - maybe just re-route to `/long` if the route cannot be found?

taker-frontend/src/components/NavBar.tsx
da-kami: I am not sure what you mean - wrap elements in an `<a>` instead of `HStack`?

taker-frontend/src/App.tsx
da-kami: I "fixed" it by redirecting to `long` in such a case as well - not sure that is better. alternatively we could add some kind of `404 not found` page.
It's actually interesting that the `Layout` is applied to an undefined route - I would have not expected that given how the code is nested and the [documentation example of `react-router-dom`](https://reactrouter.com/docs/en/v6/examples/basic).

bonomat: I think you can remove it given we redirect to `long` a few lines above.

taker-frontend/src/components/RadioCard.tsx
```diff
@@ -0,0 +1,38 @@
+import { Box, useColorModeValue, useRadio } from "@chakra-ui/react";
+import React from "react";
+
+export function RadioCard(props: any) {
+    const { getInputProps, getCheckboxProps } = useRadio(props);
```
bonomat: Imho you are misusing radio-groups to achieve _highlightable_ buttons. 

I had the same issue when creating the two toggle buttons `Long BTC` and `Short BTC` and opted for buttons as which I believe was recommended by react router. 

Anyway, there is a new example in place where I believe I found the old one. Maybe this comes handy for you: 
https://reactrouter.com/docs/en/v6/examples/custom-link

taker-frontend/src/App.tsx
```diff
@@ -169,68 +175,92 @@ export const App = () => {
...
+                    "linear(to-r, gray.800 20%, white, gray.800 80%)",
+                )}
+            >
+                <Center>
+                    <Box
+                        maxWidth={(VIEWPORT_WIDTH + 400) + "px"}
+                        width={"100%"}
+                        height={"100%"}
+                        bg={useColorModeValue("gray.200", "gray.700")}
+                    >
```
bonomat: I almost love it 😬 Imho we have too many grays. The outer one clashes with the navbar which is always visible.

image

bonomat (commented):
I really like the new navbar, well done. 
Imho the grey in the outer viewport area clashes with the navbar (and footer).

taker-frontend/src/App.tsx
bonomat: oh, and I realized I don't have the fancy gradient ...

klochowicz: maybe it's time to introduce the ItchySats logo's golden colour? :) or the ItchySats wallpaper in the back? ;D

taker-frontend/src/components/RadioCard.tsx
da-kami: Interesting, I have not thought about it from the router's perspective but just had a look at how I can achieve this with Chakra UI https://chakra-ui.com/docs/components/form/radio#custom-radio-buttons

Maybe a stupid question but why are we mis-using radio buttons here? For me this functionality is like a radio button because only one of the routes can be active. 

I am happy to have another crack at this tomorrow, but am still busy with working on the tour, so I might only pick this up a bit later.

taker-frontend/src/App.tsx
da-kami: > I almost love it 😬 Imho we have too many grays.

Totally agree! I saw it more as a placeholder until we have something "designed" (e.g. trading CSS pattern, logo, ...) that we can put there.
Would you say this has to be fixed before we merge?
Alternatively we can also just make the outer area white for now. Not as fancy but better than before :)

> oh, and I realized I don't have the fancy gradient ...

The gradient's problem is the way I use it because I define only one gradient and use a percentage distribution over the whole screen. That won't scale. Can be fixed but will need a bit of time to get it right. Maybe it's better to remove it for now.

> maybe it's time to introduce the ItchySats logo's golden colour? 

colors: I think we should invest some time into our color palette, but I am just not really good at it 😅

taker-frontend/src/components/RadioCard.tsx
klochowicz: from chakra docs: "Please be aware that the example below should only be used if you really need
 a radio button for data collection purposes. If you want to toggle between
 different content on activation of a button use the Tabs component."

perhaps `Tabs` would work best, as this is exactly our use-case?

taker-frontend/src/App.tsx
bonomat: >  Would you say this has to be fixed before we merge?

No, because I wouldn't know how to make it better.

taker-frontend/src/components/RadioCard.tsx
bonomat: I'd use radio buttons only in a form environment where the user has to select one item of a list (or if he can select multiple, use checkboxes). 
`Tabs` still sound a bit weird to me in this use case as it's not really a tab (as in the browser), imho it behaves more like buttons.

da-kami: > from chakra docs: "Please be aware that the example below should only be used if you really need
a radio button for data collection purposes. If you want to toggle between
different content on activation of a button use the Tabs component."

😅 Seems I missed that. 
I'll have a look how it is done for long / short and potentially just re-use that! Thanks for pointing me to the `NavigationButtons` @bonomat, Imissed that we already have this functionality :)

taker-frontend/src/App.tsx
thomaseizinger: > Totally agree! I saw it more as a placeholder until we have something "designed" (e.g. trading CSS pattern, logo, ...) that we can put there.

Run ads? :D

thomaseizinger (approved):

taker-frontend/src/App.tsx
da-kami: I removed the area and changed the gradient to feel more like a shadow:

image

I tried to make the gradient of fixed size, but it's hard to achieve this in the current arrangement, because I cannot easily create a horizontal layout with fixed side elements where the center element is always stretched to the maximum view-port width. 
So, when making the window smaller the gradient eventually disappears at a certain window size. I think this is acceptable for now 🤷‍♂️

PR #1932, opened 2022-05-02
Specify sqlx version as 0.5.13

klochowicz:
This should make it easier to bump versions with dependabot (upcoming 0.5.14
release will have support for uuid 1.0.0)

Am I correct with assuming, that if we don't specify the version fully, dependabot doesn't check for bugfix releases?

### DIFF
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -38,7 +38,7 @@ serde_json = "1"
 serde_with = { version = "1", features = ["macros"] }
 sha2 = "0.10"
 snow = "0.9"
-sqlx = { version = "0.5", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
+sqlx = { version = "0.5.13", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
 statrs = "0.15"
 thiserror = "1"
 time = { version = "0.3", features = ["serde", "macros", "parsing", "formatting", "serde-well-known"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1092,9 +1092,9 @@ dependencies = [
 
 [[package]]
 name = "event-listener"
-version = "2.5.1"
+version = "2.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


daemon/Cargo.toml
```diff
@@ -38,7 +38,7 @@ serde_json = "1"
 serde_with = { version = "1", features = ["macros"] }
 sha2 = "0.10"
 snow = "0.9"
-sqlx = { version = "0.5", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
+sqlx = { version = "0.5.13", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
```
thomaseizinger, line 41:
```suggestion
sqlx = { version = "=0.5.13", features = ["offline", "sqlite", "uuid", "runtime-tokio-rustls"] }
```

To pin it, you will need to use `=`. Without a modifier, `^` is assumed which means update to latest supported version without breaking changes and for < 0.x that means every patch version.

thomaseizinger: See https://doc.rust-lang.org/cargo/reference/resolver.html#semver-compatibility.

klochowicz: > To pin it, you will need to use `=`. Without a modifier, `^` is assumed which means update to latest supported version without breaking changes and for < 0.x that means every patch version.

fair enough! in this case, I think I worded the commit poorly. What I meant, is that this is the minimum version, so that dependabot could submit a PR when new one comes out.

thomaseizinger: Yes that is what it will do! But it will already do that because the lock file specifies that 0.5.13 is the current one.

klochowicz: the thing is it doesn't already do that for some reason - the lock file was stuck on `0.5.11`, despite sqlx having 2 releases in between. I thought this commit would make dependabot raise PRs when the dependency upgrades are available - am I missing something?

klochowicz: I've changed the terminology from "pin" to "specify" to better highlight what I'm doing here. Does that work for you @thomaseizinger ?

thomaseizinger (approved):

PR #1937, opened 2022-05-02
Integrate FeedbackFish into ItchySats taker

### LINKED ISSUE #1936: Use a 3rd-party feedback widget
### Discussed in https://github.com/itchysats/itchysats/discussions/1933

<div type='discussions-op-text'>

<sup>Originally posted by **klochowicz** May  2, 2022</sup>
before we venture into the territory of creating in-house messaging platform or user registration, I thought we could explore some ready-made solutions for react to save us time and allow collecting feedback sooner.

I had a look at 2 platforms that give us ready-made feedback widgets:
- Feedback Fish: https://feedback.fish/get-started/
- SerVoice: https://www.servoice.io/

Feedback Fish seems to be really appealing, and it's free to try it out (first 25 feedbacks). then it's ~$20 per month in case we'd like to continue using it.

As they have a react component ready, pretty much all code that needs to be into the app is trivial: https://feedback.fish/help/react

I'd imagine that feedback is something more critical on early stage, so even if we start paying, we could stop in a couple of months. And if we don't get feedbacks, we won't pay anything.

I'd like to go ahead with this as it's super easy to implement and from briefly looking at the react component it's super easy (looks like we can correlate feedback to a `taker_id`, which might help whether feedback comes from people actually opening positions`).
There's a way of passing in emails (opt-in), in case we'd like to respond to user feedback.</div>

klochowicz:
This way users can share anonymous feedback with us in a hassle-free way.
We'll get notified about new feedback over e-mail.

### DIFF
--- a/taker-frontend/src/components/Footer.tsx
+++ b/taker-frontend/src/components/Footer.tsx
@@ -9,7 +9,11 @@ function TextDivider() {
     return <Divider orientation={"vertical"} borderColor={useColorModeValue("black", "white")} height={"20px"} />;
 }
 
-export default function Footer() {
+interface FooterProps {
+    taker_id: string;
+}
+
+export default function Footer({ taker_id }: FooterProps) {
     return (
         <Box
             bg={useColorModeValue("gray.100", "gray.900")}
@@ -30,8 +34,10 @@ export default function Footer() {
                     <Text fontSize={"20"} fontWeight={"bold"}>Contact us:</Text>
                     <SocialLinks />
                     <TextDivider />
-                    {/* TODO Consider adding takerId */}
-                    <FeedbackFish projectId="c1260a96cdb3d8">
+                    <FeedbackFish
+                        projectId="c1260a96cdb3d8"
+                        metadata={{ position: "footer", customerId: taker_id }}
+                    >
                         <Button fontSize={"20"} color={useColorModeValue("black", "white")}>Send feedback</Button>
                     </FeedbackFish>
                 </HStack>
--- a/taker-frontend/src/components/Footer.tsx
+++ b/taker-frontend/src/components/Footer.tsx
@@ -1,5 +1,6 @@
 import { ExternalLinkIcon } from "@chakra-ui/icons";
-import { Box, Center, Divider, HStack, Link, Text, useColorModeValue } from "@chakra-ui/react";
+import { Box, Button, Center, Divider, HStack, Link, Text, useColorModeValue } from "@chakra-ui/react";
+import { FeedbackFish } from "@feedback-fish/react";
 import * as React from "react";
 import { FAQ_URL, FOOTER_HEIGHT } from "../App";
 import { SocialLinks } from "./SocialLinks";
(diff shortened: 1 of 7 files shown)


taker-frontend/src/components/Footer.tsx
```diff
@@ -27,6 +28,11 @@ export default function Footer() {
                     <TextDivider />
                     <Text fontSize={"20"} fontWeight={"bold"}>Contact us:</Text>
                     <SocialLinks />
+                    <TextDivider />
+                    {/* TODO Consider adding takerId */}
```
bonomat: +1 on adding the `taker_id` right away so that we know if the feedback is from a seasoned trader on ItchySats or a newbie.

bonomat (commented):
Can we make it more prominent? Maybe even add it to the header ? Or a floating bubble in the right lower corner?

klochowicz:
> Can we make it more prominent? Maybe even add it to the header ? Or a floating bubble in the right lower corner?

If the UI revamp by @da-kami comes in first, I'll try to rebase it into a more prominent spot.

taker-frontend/src/components/Footer.tsx
thomaseizinger: Let's just remember that this can be spoofed unless we utilise signatures.

thomaseizinger (approved):
Screenshot would be nice to see what it looks like without having to checkout the PR!

da-kami:
> Screenshot would be nice to see what it looks like without having to checkout the PR!

Yes please!

da-kami:
If you need more space in the top bar (on narrow screen, ...) we could actually make the design responsive and e.g. don't show the text of Home/Wallet button if screen size below certain width. Should not be too hard to achieve.

klochowicz:
the looks: 
image


the sample on feedback widget (including takerId)
image

klochowicz:
> If you need more space in the top bar (on narrow screen, ...) we could actually make the design responsive and e.g. don't show the text of Home/Wallet button if screen size below certain width. Should not be too hard to achieve.

this will probably be the way to go, but I don't think it's in the scope of this PR - perhaps we could work a bit on the UI when we're pairing up in Sydney?

taker-frontend/src/components/Footer.tsx
klochowicz: > +1 on adding the `taker_id` right away so that we know if the feedback is from a seasoned trader on ItchySats or a newbie.

done. I also added a small printout of the taker Id, so that anyone can refer to it when they want to debug some issues with us (if they're reaching out to us in a different way)

klochowicz:
> 

The easiest way would be moving it to the left of the ItchySats logo - I'll see what I can do. The downside would be less space on mobile, for which responsive design mentioned by @da-kami would be an answer. However, we do not explicitly support mobile yet, so it's still probably worth focusing on optimal desktop experience....

bonomat:
> Can we make it more prominent? Maybe even add it to the header ? Or a floating bubble in the right lower corner?

I think this has to be more prominent, Imho people will not see it in the footer.

klochowicz:
> > Can we make it more prominent? Maybe even add it to the header ? Or a floating bubble in the right lower corner?
> 
> I think this has to be more prominent, Imho people will not see it in the footer.

IMHO it's visible, especially when you first start the app (before you add Cfds). 
we can add a suggestion to solicit feedback as part of the tour as well.
I also would prefer to make it a bit less annoying for more seasoned users (for anyone not wanting to give us feedback, this is wasted screen real estate - as long as it's easy to get to it, they will know where to find it if they seen it once).

da-kami (approved):
I agree with @bonomat - I'd add it more prominent. Happy to merge this and do it in a follow-up thought.

PR #1947, opened 2022-05-03
Don't ignore error messages

bonomat:


da-kami (approved):

daemon/src/maker_inc_connections.rs
```diff
@@ -607,7 +607,7 @@ async fn upgrade(
 
     let daemon_version = daemon_version.unwrap_or_else(|| String::from("<= 0.4.7"));
 
-    tracing::trace!(%taker_id, %taker_address, %negotiated_wire_version, %daemon_version, "Connection upgrade successful");
+    tracing::debug!(%taker_id, %taker_address, %negotiated_wire_version, %daemon_version, "Connection upgrade successful");
```
luckysori: 🤔 This was purposefully demoted recently because it is quite low level compared to the "Connection is ready" log which is emitted once the connection has actually been established. It's not a big deal, but I obviously disagree with this change 😛

bonomat: Fair enough :) 
In a recent debugging session we identified the need for more logs, but it might be easier to just enable this using `RUST_LOG`. I'll drop this commit from the PR.

daemon/src/maker_inc_connections.rs
```diff
@@ -236,12 +236,14 @@ impl Actor {
             .get_mut(taker_id)
             .ok_or(NoConnection(*taker_id))?;
 
-        if conn.send(msg).await.is_err() {
-            self.drop_taker_connection(taker_id).await;
-            return Err(NoConnection(*taker_id));
+        match conn.send(msg).await {
+            Ok(_) => Ok(()),
```
thomaseizinger: Is the ignored value here a `()`? Would be good to explicitly match against so it becomes a compile-error in case anyone ever changes the signature of this function.

daemon/src/maker_inc_connections.rs
thomaseizinger: Perhaps we should just drop the `Connection is ready` log instead?

thomaseizinger (approved):

klochowicz (approved):
LGTM!

daemon/src/maker_inc_connections.rs
bonomat: done

daemon/src/maker_inc_connections.rs
bonomat: I'll leave it for now.

PR #1948, opened 2022-05-03
Add time-to-first-position metric to `maker`

### LINKED ISSUE #1895: Record taker-to-open-position-metric
The easiest way to move forward here is to record a metric in the db which records when a taker was seen for the very first time and when the taker opened their first position. 

_Originally posted by @klochowicz in https://github.com/itchysats/itchysats/discussions/1841#discussioncomment-2590202_

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/1895.

Since this metric is obviously only relevant for a maker, I was inspired to attempt to start extracting code that is only used by the `maker` binary into the `maker` crate. These changes aren't strictly necessary, but I think they are in the right direction.


future-timeout/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "future-timeout"
```
thomaseizinger: Smallest crate in the universe :D

Perhaps just duplicate this?

maker/src/cfd.rs
```diff
@@ -0,0 +1,711 @@
+use crate::collab_settlement;
```
thomaseizinger, line 1: I wonder why this isn't marked as moved?

daemon/src/rollover_maker.rs
```diff
@@ -0,0 +1 @@
+
```
thomaseizinger: What is this? :)

daemon/src/db/time_to_first_position.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+async fn load_first_seen_timestamp(
+    conn: &mut PoolConnection<Sqlite>,
+    taker_id: Identity,
+) -> Result<Option<OffsetDateTime>> {
+    let row = sqlx::query!(
+        r#"
+        SELECT
+            time_to_first_position.first_seen_timestamp
```
thomaseizinger:
```suggestion
            first_seen_timestamp
```

I don't think you need the table name here!

daemon/src/db/time_to_first_position.rs
```diff
@@ -0,0 +1,262 @@
...
+}
+
+async fn load_first_position_timestamp(
+    conn: &mut PoolConnection<Sqlite>,
+    taker_id: Identity,
+) -> Result<Option<OffsetDateTime>> {
+    let row = sqlx::query!(
+        r#"
+            SELECT
+                time_to_first_position.first_position_timestamp
```
thomaseizinger: Same here.

daemon/src/db/time_to_first_position.rs
```diff
@@ -0,0 +1,262 @@
...
+    pub async fn try_insert_first_seen(
+        &self,
+        taker_id: Identity,
+        timestamp: OffsetDateTime,
+    ) -> Result<()> {
+        let mut conn = self.inner.acquire().await?;
+
+        if !have_seen_before(&mut conn, taker_id).await? {
+            insert_first_seen(&mut conn, taker_id, timestamp).await?;
+        }
```
thomaseizinger: Could potentially replace this with SQL?

https://stackoverflow.com/a/19337206

daemon/Cargo.toml
```diff
@@ -42,7 +43,6 @@ sqlx = { version = "0.5", features = ["offline", "sqlite", "uuid", "runtime-toki
 statrs = "0.15"
 thiserror = "1"
 time = { version = "0.3", features = ["serde", "macros", "parsing", "formatting", "serde-well-known"] }
-future-timeout = { path = "../future-timeout" }
```
bonomat: did you check if we can remove more dependencies now?

daemon/migrations/20220502172229_create_time_to_first_position_table.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS time_to_first_position (
+    id integer PRIMARY KEY autoincrement,
+    taker_id text NOT NULL,
```
bonomat: Imho this should be `UNIQUE`

daemon/src/db/time_to_first_position.rs
bonomat: nice idea. 👍

daemon/migrations/20220502172229_create_time_to_first_position_table.sql
thomaseizinger: There is a unique constraint further down!

daemon/src/db/time_to_first_position.rs
```diff
@@ -0,0 +1,262 @@
...
+        None => return Ok(None),
+        Some(row) => row.first_seen_timestamp,
+    };
+
+    let timestamp = match timestamp {
+        None => return Ok(None),
+        Some(timestamp) => timestamp,
+    };
+
+    let timestamp = OffsetDateTime::from_unix_timestamp(timestamp)?;
```
bonomat: Might be a usecase for `transpose` https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose :) 

Maybe like this: 

```
timestamp
  .map(|row| OffsetDateTime::from_unix_timestamp(row.first_seen_timestamp))
  .transpose()?
```

daemon/Cargo.toml
bonomat: Saw it happening a few commits later :)

daemon/src/db/time_to_first_position.rs
da-kami: yeah, in statements that only involve one table they are somewhat of an overkill.

daemon/migrations/20220502172229_create_time_to_first_position_table.sql
da-kami: Is there a specific reason / advantage for (only) creating the `UNIQUE` constraint implicitly through adding an index? 
I agree with @bonomat that it would be easier to understand if we'd have (an additional) `UNIQUE` constraint here.

daemon/src/db/time_to_first_position.rs
da-kami: I think this should work for this instance, but might be a good idea to have another look yourself on how sqlite's conflict resolution algorithm works when using `REPLACE`:

https://sqlite.org/lang_conflict.html

You will definitely need a `UNIQUE` constraint on the `taker_id` then, otherwise `REPLACE` might not work as expected.
Given that we can have `NULL` values in the table please add tests that show that this works as intended!

da-kami (commented):
ack metrics concept! 
I like where the extracing code part is going, but it feels almost a bit much for this PR :)

future-timeout/Cargo.toml
luckysori: Hahaha fair enough.

maker/src/cfd.rs
luckysori: That is weird! Maybe the original file wasn't deleted?

daemon/src/rollover_maker.rs
luckysori: Classic! I'll see how many files were not moved properly.

daemon/src/db/time_to_first_position.rs
luckysori: Yep, can do. I think we were trying to be consistent with always referencing the table. But it is overkill.

daemon/migrations/20220502172229_create_time_to_first_position_table.sql
luckysori: I probably just forgot the `UNIQUE` label!

daemon/src/db/time_to_first_position.rs
luckysori: I tried it originally, but it was not very readable at all haha, because it's actually a `Result<Option<Option<OffsetDateTime>>` and I don't think you can pattern match much since `sqlx::query!` outputs "an ad-hoc anonymous struct type".

But do give it a go and see if you can produce something better!

daemon/src/db/time_to_first_position.rs
luckysori: I like this proposal, but `sqlx` doesn't accept the query I've written. Perhaps someone can help me with this.

luckysori:
> I like where the extracing code part is going, but it feels almost a bit much for this PR :)

@da-kami: It's certainly not necessary, but I think it's a good opportunity to start tackling this. Otherwise, the `daemon` crate will keep growing out of control. I wonder what @thomaseizinger thinks, since he started doing similar work but ended up dropping it.

daemon/src/db/time_to_first_position.rs
bonomat: Happy to help out. I'm the master of SQL and killed once a production DB at Siemens using recursive read queries!!!

bonomat: This was bogus...

bonomat: Hold on, this does not work, you might accidentally update the wrong taker's fields 😬

bonomat: So, for inserting the `first_seen_timestamp` you can do this: 

```sql
INSERT OR IGNORE INTO time_to_first_position (taker_id, first_seen_timestamp)
select 'test4', 4
```

It will only insert if it does not exist yet based on the unique key `taker_id`

For updating the first open position you can do: 

```sql
UPDATE time_to_first_position
SET first_position_timestamp = 42
WHERE taker_id = 'test' and first_position_timestamp is null;
```

luckysori: Thanks for giving it a go! I'll try to add it to this PR.

luckysori: Why can't we do the same thing for the `first_position_timestamp`? The approach you propose for the `first_seen_timestamp` seems to work (i.e. tests pass).

bonomat: Afaik, if `taker_id` is already present then `insert or ignore` will ignore it and won't do anything.

luckysori: > Afaik, if `taker_id` is already present then `insert or ignore` will ignore it and won't do anything.

Confirmed!

luckysori:
I had to add https://github.com/itchysats/itchysats/pull/1948/commits/aa72ce9a813f740574c860a823df3e26884cc1f3 because CI was failing. I don't quite understand why a warning about this would appear with this PR, but it's a tiny change anyway.

daemon/src/db/time_to_first_position.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    async fn load_first_seen_timestamp(
+        conn: &mut PoolConnection<Sqlite>,
+        taker_id: Identity,
+    ) -> Result<Option<OffsetDateTime>> {
+        let row = sqlx::query!(
+            r#"
+            SELECT
+                time_to_first_position.first_seen_timestamp
```
bonomat: 🙃
You could remove `time_to_first_position` here

daemon/src/db/time_to_first_position.rs
bonomat: Pushed one commit just for fun :) https://github.com/itchysats/itchysats/pull/1948/commits/a876482ac26d4dd106dec23a8a61e3b7b96059d2

bonomat (approved):
I think this is good to go!

daemon/src/db/time_to_first_position.rs
```diff
@@ -159,19 +159,15 @@ mod tests {
...
-        Ok(Some(timestamp))
+        let timestamp = row
+            .map(|row| {
+                row.first_seen_timestamp
+                    .map(OffsetDateTime::from_unix_timestamp)
+            })
+            .unwrap_or(None)
+            .transpose()?;
+
+        Ok(timestamp)
```
luckysori, line 170: Amazing! Thank you 😍

PR #1957, opened 2022-05-04
Body pushes footer to the bottom

da-kami:
The initially (or when switching to the wallet) "floating" footer has annoyed me for too long :)

There are several different ways to achieve the footer being always at the bottom, but at the same time not being fixed. I played with flexes for a while but could not get it done without weird side effects. I think it would be nice to have flex height for header and footer, but at this point I just want it to be done. 
Solution:

If the body is less than the viewport height (in our case height of the browser window) then we set a minimum of `body-height = viewport-height - header-height - footer-height`.
If the body becomes larger than the viewport height the footer will be pushed down (no fixed footer, only fixed header).

image

When body is higher than viewport the footer is pushed down and we scroll to get to it:

image

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -177,80 +177,78 @@ export const App = () => {
                 nextFundingEvent={nextFundingEvent}
                 referencePrice={referencePrice}
             />
-            <Box height={"100%"}>
-                <Center>
-                    <Box
-                        maxWidth={(VIEWPORT_WIDTH + 200) + "px"}
-                        width={"100%"}
-                        bgGradient={useColorModeValue(
-                            "linear(to-r, white 5%, gray.800, white 95%)",
-                            "linear(to-r, gray.800 5%, white, gray.800 95%)",
-                        )}
-                    >
-                        <Center>
-                            <Box
-                                textAlign="center"
-                                padding={3}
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
LGTM

taker-frontend/src/App.tsx
```diff
@@ -177,80 +177,78 @@ export const App = () => {
...
-                                bg={useColorModeValue(BG_LIGHT, BG_DARK)}
-                                maxWidth={VIEWPORT_WIDTH_PX}
-                                marginTop={`${HEADER_HEIGHT}px`}
-                                minHeight={`calc(100vh - ${FOOTER_HEIGHT}px - ${HEADER_HEIGHT}px)`}
-                                width={"100%"}
-                            >
-                                <Routes>
-                                    <Route path="/">
+            <Center>
+                <Box
```
bonomat, line 181: Can you replace the `Box` with `Center`?

bonomat (approved):
LGTM

taker-frontend/src/App.tsx
da-kami: I played around with it for a bit but it's not straight forward - just moving the arguments to `Center` messes up the design. Am not investing more time at this point :)

PR #1960, opened 2022-05-04
Move away from funding rate to interest rate in projection and ui

### LINKED ISSUE #1877: Re-work funding rate to display positive when the user is getting paid and negative when the user pays
### Discussed in https://github.com/itchysats/itchysats/discussions/1751

After a discussion on github, I think we're pretty aligned how we want to represent the funding rates:
- negative, if the user is paying (funding)
- positive, if the user is getting paid (getting funded)

The proposal in the discussion to just expose the correct API from the daemon (as it knows the role) seems sensible. @thomaseizinger posted some name suggestions in the discussion (https://github.com/itchysats/itchysats/discussions/1751#discussioncomment-2497443)

<div type='discussions-op-text'>

<sup>Originally posted by **bonomat** March 27, 2022</sup>
The way how funding rate is depicted at the moment might be confusing. 
I know that a positive funding rate means, long pays short but this is not really visible nor described anywhere. 
Right now, it looks like that both, longs and shorts will need to pay this rate: 

--- Annoyance Counter --
(1) Discussion started: https://github.com/itchysats/itchysats/discussions/1751
(2) Discussion point 1: team members are confused https://github.com/itchysats/itchysats/discussions/1751#discussioncomment-2497441
(3) Discussion closed and started working on [#1877](https://github.com/itchysats/itchysats/issues/1877)
(4) Wrote an ADR which showed that team members are not aligned on what it actually means
(5) New ticket created which looks like a UI Bug: https://github.com/itchysats/itchysats/issues/2107
(6) It's really hard to figure out if the numbers in our UI are correct.

bonomat:
Resolves #1877 

The "funding rate" so far was rather confusing. The issue with it was that we had multiple funding rates: one for going long and one for going short.
The fact that these two were different says that they are interest rates which need to be paid or are paid out for keeping a position open.

When looking at interest rates it also makes more sense to have negative or positive values and it's easier to see if you have to pay or will receive, i.e. similar to a bank account:
- If the interest rate is positive it means you receive,
- If the interest rate is negative it means you pay.

Combined with the position we can then convert the funding rate into interest rate:
As a reminder:
- A positive funding rate means "long pay short"
- A negative funding rate means "short pay long'

Some minor UI changes: 

This used to say `funding rate`: 
image

This also used to say `funding rate`: 
image

Note: 
the funding rate is positive for shorts because they will receive funding. The tooltip also explains this. 
image


luckysori:
Holding off the review until we agree on https://github.com/itchysats/architecture/pull/29.

klochowicz:
I'm not sure whether interest rate is easier to understand for our users. Banks communicate "interest rate" as a positive number, and it always means paying *to* them.

bonomat:
> I'm not sure whether interest rate is easier to understand for our users. Banks communicate "interest rate" as a positive number, and it always means paying _to_ them.

That's what I say: 

> When looking at interest rates it also makes more sense to have negative or positive values and it's easier to see if you have to pay or will receive, i.e. similar to a bank account:
> 
>     * If the interest rate is positive it means you receive,
>     * If the interest rate is negative it means you pay.

bonomat:
Closing: no consensus found

bonomat:
> I'm not sure whether interest rate is easier to understand for our users. Banks communicate "interest rate" as a positive number, and it always means paying _to_ them.

Depends on the product: for a savings account the interest indicate that you get paid, while negative interest means you have to pay.

PR #1961, opened 2022-05-04
Initial tour through ItchySats

da-kami:
Probably best if you click through yourself and see how it feels. 
I can record a gif tomorrow if wanted.

TODO:

- [x] make sure to start the UI always in `light` mode for the first time


da-kami:
Note: Even though we explicitly set the `initialColorMode` to `light` now, for existing users that already have used ItchySats in `dark` mode before the browser might remember these settings and the initial tour might start up looking weird because it cannot handle the `dark` mode at the moment.
This can be fixed, but I would not invest time into this at this point.

One thing we could do is only start the tour if no CFDs are open, but I am not sure it would work properly because at the point in time when we start the tour we might have not received the CFD list yet.

da-kami:
Complete lifecycle including refresh in the end (from wallet route) and disabling the tour on  the second run:

taker-frontend/src/components/Tour.tsx
```diff
@@ -0,0 +1,182 @@
...
+            position: "right",
+        },
+        {
+            element: "#longButton",
+            intro: (
+                <>
+                    <Text>
+                        Clicking this button will open a confirmation dialog similar to this:
+                        <br />
+                        <img src={confirmationDialog} width={"400px"} alt={"Confirm it already!"} />
```
bonomat, line 99: I would remove this image. It will be outdated too fast, e.g. this PR #1960 makes changes to the UI already.

taker-frontend/src/components/Tour.tsx
```diff
@@ -0,0 +1,182 @@
...
+    };
+
+    const navigate = useNavigate();
+    const location = useLocation();
+
+    const onChange = (nextStepIndex: number) => {
+        // Before the last step we jump to the wallet
+        if (nextStepIndex === (tourSteps.length - 1)) {
+            navigate("/wallet");
+        } else if (!location.pathname.endsWith("long")) {
```
bonomat: This asks for troubles or move steps around.
What do you think about adding an additional field to the toursteps such as `target`. Then you can do something more generic like: 
```
        let target = tourSteps.at(nextStepIndex).target;
        if (!location.pathname.endsWith(target)) {
            navigate(target);
        }
```

bonomat:
Looking really nice already. I stopped with the review for now. Ping me when you are done forcepushing please!

taker-frontend/src/components/Tour.tsx
bonomat: You could probably include the dialog component and fill it with dummy values.

da-kami: > You could probably include the dialog component and fill it with dummy values.

This *might* be hard to achieve given the render lifecycle and the tour lifecycle. I would not try to achieve this, it will be more headache than it helps.
I think the screenshot is fine - even if it gets outdated. It helps the user understand that "something" will pop up, I don't think users will try to understand it. The point is that the user *know* that they can press the `Long` button and it won't open the position immediately. The screenshot helps grasping that - I would not expect users to read the fineprint on it...

taker-frontend/src/components/Tour.tsx
da-kami: I like the proposed idea and am happy to give it a try.

Note: We actually hardcode `long` in several places in the code already, assuming that is the default route 😬

da-kami: Worked out nice! Code is way cleaner!

The behavior did not change.
I wonder if we can even detect the "we are at the start" in a better way, but could not figure out a good way given that we always just get the `next` index `onChange`.

taker-frontend/src/components/Tour.tsx
bonomat: you already mention in the text that a dialog will be opened, if the image is outdated, then it does not add much value. Imho anyone can imagine how a confirmation dialog will look like :D .

taker-frontend/src/components/Tour.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { useLocation, useNavigate } from "react-router-dom";
+import { FAQ_URL } from "../App";
+import cfd101 from "../images/CFD_101_light_bg.svg";
+import confirmationDialog from "../images/confirmation_dialog.png";
+import itchyGuyWelcome from "../images/ItchyGuyWelcome.svg";
+
+export const Tour = () => {
+    const tourSteps = [
+        {
+            title: "G'Day Hodlers!",
```
bonomat: 🙃
A hodler does not trade, he hodls 😬
```suggestion
            title: "G'Day Satoshi!",
```

bonomat (approved):
Nice, imho we should try to change the dialog's width. Looks a bit weird that it is so long. 
Feel free to create a follow-up ticket for that.

taker-frontend/src/components/Tour.tsx
da-kami: Fair enough! I guess strictly hodl really just stands for "keep your money under the pillow".
I saw it more from the perspective of Bitcoin in Bitcoin out, I want to only grow my Bitcoin.

taker-frontend/src/components/Tour.tsx
da-kami: > you already mention in the text that a dialog will be opened, if the image is outdated, then it does not add much value. Imho anyone can imagine how a confirmation dialog will look like :D .

I think the image helps drawing attention to it. Text is easily ignored, an image will draw your attention.
Is this a blocker for you? I still thinks it adds more value than it brings problems 🤷‍♂️

da-kami:
> Nice, imho we should try to change the dialog's width. Looks a bit weird that it is so long. Feel free to create a follow-up ticket for that.

I reckon that's possible by modifying the `intro.js` CSS. I wanted to avoid that - happy to create a follow-up ticket so my annoyance counter does not go through the roof 😬
see: https://github.com/itchysats/itchysats/issues/1968

taker-frontend/src/components/Tour.tsx
klochowicz: this is a good question. under my naive assumption, trading *in* bitcoin, is still HODLing bitcoin, as you don't get rid of it! but I defer to your judgements, as I'm a relative newbie in this space!

klochowicz (approved):
There's certainly room for improvement in the looks of the tour.
I completely haven't noticed that very subtle pointer and felt that almost all the pop-ups were placed randomly. 
That said, it is clearly an improvement over what was here before. Any other CSS/looks changes can be done later.

Issue #1739, opened 2022-04-01
Let the taker choose leverage 1, 2 and 3


luckysori:
Leverage 2 is already the default, but we don't want to hard-code it anymore. See #1573.

Once #1573 is in, we can then let the taker choose a leverage from 1 to 3.

PR #1965, opened 2022-05-05
A selection of leverages

### LINKED ISSUE #1573: Actually send leverage choices to the taker
Hardcode them to only allow "2" for now.

The point of this task is to prepare the underlying infrastructure for configurable leverage.

bonomat:
- maker sends over a list of leverages the taker can chose from
- taker sends over the leverage he wants to use

Resolves #1573 

Please check if I this is how we want to preserve backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Maker only: Add new metric under `{maker_url}/api/metrics` which shows the quantity of total positions
 - Maker only: Add new metric under `{maker_url}/api/metrics` which shows the number of total positions
 - Taker UI: Add initial onboarding tour that steps the taker through the user interface and ends in the wallet.
+- API changes: maker and taker HTTP Api changed to let the taker chose between multiple leverages
+  - maker: `/api/offer`: has an additional field `leverage_choices: Vec<Leverage>`
+  - taker: `/api/cfd/order`: has an additional field `leverage: Leverage`, i.e. the leverage selected by the taker
+  - both: `/api/feed/long_offer` && `/api/feed/short_ofer`: instead of a single leverage it now returns a list of leverages (`leverage_choices`)
+  - network: an additional field `leverage` was added to `wire::MakerToTaker::TakeOffer`.
+  - network: an additional field `leverage_choices` was added to `Order` which is sent from maker to taker when he created a new offer.
+    The field `leverage` was deprecated.
 
 ## [0.4.12] - 2022-04-26
 
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -1255,8 +1255,7 @@ pub struct CfdOrder {
     #[serde(with = "round_to_two_dp")]
     pub lot_size: Usd,
 
-    #[serde(rename = "leverage")]
-    pub taker_leverage_choices: Leverage,
+    pub leverage_choices: Vec<Leverage>,
 
     /// Own liquidation price according to position and leverage
     #[serde(with = "round_to_two_dp")]
@@ -1294,8 +1293,11 @@ impl TryFrom<Order> for CfdOrder {
             Origin::Theirs => order.position_maker.counter_position(),
         };
 
+        // FIXME: this won't work anymore because we don't know what leverage the user wants to
+        // take. I'll fix it to Leverage::2 for now as this is our only choice but this
+        // needs to be fixed!
         let (long_leverage, short_leverage) =
-            long_and_short_leverage(order.leverage_taker, role, own_position);
+            long_and_short_leverage(Leverage::TWO, role, own_position);
 
         let initial_funding_fee_per_lot = FundingFee::calculate(
             order.price,
@@ -1334,7 +1336,7 @@ impl TryFrom<Order> for CfdOrder {
             max_quantity: order.max_quantity,
             lot_size,
             margin_per_lot,
-            taker_leverage_choices: order.leverage_taker,
+            leverage_choices: order.leverage_choices,
             liquidation_price,
             creation_timestamp: order.creation_timestamp_maker,
             settlement_time_interval_in_secs: order
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -1204,7 +1204,7 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
     }
 }
 
-/// Maker offers represents the offers as cerated by the maker
+/// Maker offers represents the offers as created by the maker
 #[derive(Debug, Clone, PartialEq, Serialize)]
 pub struct MakerOffers {
     /// The offer where the maker's position is long
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -92,9 +92,17 @@ pub enum TakerToMaker {
         proposed_wire_version: Version,
         daemon_version: String,
     },
+    /// This is deprecated, use `TakeOrderWithLeverage` instead
+    #[serde(rename = "TakeOrder")]
+    DeprecatedTakeOrder {
+        order_id: OrderId,
+        quantity: Usd,
+    },
+    #[serde(rename = "TakeOrderWithLeverage")]
     TakeOrder {
         order_id: OrderId,
         quantity: Usd,
+        leverage: Leverage,
     },
     ProposeRollover {
         order_id: OrderId,
@@ -123,7 +131,10 @@ pub enum TakerToMaker {
 impl TakerToMaker {
     pub fn name(&self) -> &'static str {
         match self {
-            TakerToMaker::TakeOrder { .. } => "TakerToMaker::TakeOrder",
+            // allowing deprecated use of field `TakeOrder` here for backwards compatibility.
+            #[allow(deprecated)]
+            TakerToMaker::DeprecatedTakeOrder { .. } => "TakerToMaker::TakeOrder",
+            TakerToMaker::TakeOrder { .. } => "TakerToMaker::TakeOrderWithLeverage",
             TakerToMaker::Protocol { msg, .. } => match msg {
                 SetupMsg::Msg0(_) => "TakerToMaker::Protocol::Msg0",
                 SetupMsg::Msg1(_) => "TakerToMaker::Protocol::Msg1",
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -12,6 +12,7 @@ use http_api_problem::HttpApiProblem;
 use http_api_problem::StatusCode;
 use model::FundingRate;
 use model::Identity;
+use model::Leverage;
 use model::OpeningFee;
 use model::OrderId;
 use model::Price;
@@ -102,7 +103,7 @@ pub async fn maker_feed(
 }
 
 /// The maker PUTs this to set the offer params
-#[derive(Debug, Clone, Copy, Deserialize)]
+#[derive(Debug, Clone, Deserialize)]
 pub struct CfdNewOfferParamsRequest {
     pub price_long: Option<Price>,
     pub price_short: Option<Price>,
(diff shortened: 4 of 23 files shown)


daemon/src/maker_cfd.rs
```diff
@@ -618,8 +628,25 @@ where
...
+            wire::TakerToMaker::TakeOrder {
+                order_id,
+                quantity,
+                leverage,
+            } => {
+                // Old clients do not send over leverage. Hence we default to Leverage::TWO which
+                // was the default before. We log a warning though.
+                let leverage = match leverage {
+                    None => {
+                        tracing::warn!(taker_id = %taker_id, "Received take request without leverage. Could be an outdated client but worth to double check the taker's version.");
```
da-kami: 🙃 I think we can downgrade this to `debug`. If it is something that can happen (i.e. is expected to some degree) I don't think it should be a warning.

daemon/src/projection.rs
```diff
@@ -1214,8 +1214,11 @@ impl TryFrom<Order> for CfdOrder {
             Origin::Theirs => order.position_maker.counter_position(),
         };
 
+        // FIXME: this won't work anymore because we don't know what leverage the user wants to
```
da-kami, line 1296: 🙃 Should we record this as a ticket? I understand why this was added, but I am not sure it's very helpful to merge this.

maker/src/routes.rs
```diff
@@ -124,6 +126,12 @@ pub async fn put_offer_params(
     maker: &State<Maker>,
     _auth: Authenticated,
 ) -> Result<(), HttpApiProblem> {
+    // TODO: temporary until the automation can deal with this
```
da-kami: 🙃 Might be better to record a ticket for the TODO that points to the code and explains what is to do.

model/src/cfd.rs
```diff
@@ -225,11 +226,15 @@ pub struct Order {
 
     /// The taker leverage that the maker allows for the taker
     ///
-    /// Currently only leverage of x2 is possible, that's why this is not a range of choices.
-    /// The maker's leverage always set o x1 is currently just implied.
+    /// This is needed for backwards compatibility reasons.
+    #[deprecated(since = "0.13.0", note = "please use `leverage_choices_taker` instead")]
```
da-kami:
```suggestion
    #[deprecated(since = "0.4.13", note = "please use `leverage_choices_taker` instead")]
```

I guess?

da-kami (approved):
noice! thanks for driving this forward!

daemon/src/maker_cfd.rs
bonomat: You are right, it will happen particulary at the beginning when upgrading to a new version...
Happy to downgrade it but then I'm afraid we will just not see this message in the flood of other log messages.

daemon/src/projection.rs
bonomat: Definitely. I wasn't sure how to solve this yet. Could be a bigger task. 
Are you fine with leaving the todo and having the ticket?

maker/src/routes.rs
bonomat: I prefer having both :)

model/src/cfd.rs
bonomat: upsi... close enough 😬

CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Maker only: Add new metric under `{maker_url}/api/metrics` which shows the quantity of total positions
 - Maker only: Add new metric under `{maker_url}/api/metrics` which shows the number of total positions
+- API changes: maker and taker HTTP Api changed to let the taker chose between multiple leverages
+  - maker: `/api/offer`: has an additional field `leverage_choices: Vec<Leverage>`
```
luckysori, line 16: 🤔 I'm still a bit suspicious of the fact that we are cramming a lot of information on the offer which changes infrequently compared to how often we push out new offers.

daemon/src/maker_cfd.rs
```diff
@@ -618,8 +628,25 @@ where
 
     async fn handle(&mut self, FromTaker { taker_id, msg }: FromTaker) {
         match msg {
-            wire::TakerToMaker::TakeOrder { order_id, quantity } => {
-                if let Err(e) = self.handle_take_order(taker_id, order_id, quantity).await {
+            wire::TakerToMaker::TakeOrder {
```
luckysori: ❓ Would it be cleaner to introduce a separate wire message altogether and ensure that the maker handles both versions? That way we don't have to log that awkward warning below. If we received `V1` we know we can imply leverage 2 safely; if we get `V2` the body of the message will always contain the configurable `leverage` field.

daemon/src/projection.rs
```diff
@@ -1175,8 +1175,8 @@ pub struct CfdOrder {
     #[serde(with = "round_to_two_dp")]
     pub lot_size: Usd,
 
-    #[serde(rename = "leverage")]
-    pub taker_leverage_choices: Leverage,
+    #[serde(rename = "leverage_choices")]
+    pub taker_leverage_choices: Vec<Leverage>,
```
luckysori: 👁️ Just pointing out that I think that renaming this field is a breaking change.

daemon/src/projection.rs
luckysori: ❓ What's the plan for this? The obvious solution to me would be to a range of `initial_funding_fee_per_lot` based on the `leverage_choices_taker`. But what's the `initial_funding_fee_per_lot` used for anyway?

daemon/src/wire.rs
```diff
@@ -95,6 +95,9 @@ pub enum TakerToMaker {
     TakeOrder {
         order_id: OrderId,
         quantity: Usd,
+        // This is an option to preserve backwards compatibility as old clients do not send this
+        // value. If NONE, it will default to Leverage::Two
+        leverage: Option<Leverage>,
```
luckysori: 🔧 I think introducing a separate variant would be easier to maintain.

model/src/cfd.rs
```diff
@@ -225,11 +226,15 @@ pub struct Order {
...
     #[serde(rename = "leverage")]
     pub leverage_taker: Leverage,
 
+    /// A selection of leverages that the maker allows for the taker
+    #[serde(rename = "leverage_choices")]
+    pub leverage_choices_taker: Vec<Leverage>,
```
luckysori: ❓ Why do we need to have a different name for the serialisation and the struct field? Is it a stylistic choice?

daemon/src/projection.rs
bonomat: #1983 craeted

maker/src/routes.rs
bonomat: done.

maker/src/routes.rs
```diff
@@ -124,6 +126,12 @@ pub async fn put_offer_params(
     maker: &State<Maker>,
     _auth: Authenticated,
 ) -> Result<(), HttpApiProblem> {
+    // temporary until the automation can deal with this
+    let mut leverage_choices = offer_params.leverage_choices.clone();
+    if leverage_choices.is_empty() {
+        leverage_choices.push(Leverage::TWO)
+    }
+
```
luckysori: 🔧 I don't think this works if I understand your intentions.

This is a breaking change to the automation i.e. the API expects a `leverage_choices` field now. One way to fix this would be to annotate the field as `#[serde(default)]`. That way a missing `leverage_choices` is replaced with the default value for `Vec<Leverage>` i.e. an empty vector.

Even better would be to use https://serde.rs/field-attrs.html#default--path, and you can provide a function which already returns a `vec![Leverage::TWO]`.

taker/src/routes.rs
```diff
@@ -139,8 +140,18 @@ pub async fn post_order_request(
...
+        return Err(HttpApiProblem::new(StatusCode::BAD_REQUEST)
+            .title("Invalid leverage selected. For time being, please go with leverage 2."));
+    }
     taker
-        .take_offer(cfd_order_request.order_id, cfd_order_request.quantity)
+        .take_offer(
+            cfd_order_request.order_id,
+            cfd_order_request.quantity,
+            cfd_order_request.leverage,
+        )
```
luckysori, line 154: ❓ Do we really need this?

luckysori (commented):
Sorry for the delay, I forgot to submit this review last week.

CHANGELOG.md
bonomat: I think in regards to `leverage` this is fine. These things just belong together for me...

daemon/src/projection.rs
bonomat: I thought about and decided this is not a problem. Might double check with the metrics library we built on top.

model/src/cfd.rs
bonomat: Good point. I think this should be aligned 👍

maker/src/routes.rs
bonomat: Works without any problem at the moment 🙃

daemon/src/maker_cfd.rs
bonomat: I think an extra wire message is not worth the complexity and effort given that the code is compatible.

taker/src/routes.rs
bonomat: Why not? Because it goes towards #1739 ?

model/src/cfd.rs
bonomat: aligned in [`40284ed` (#1965)](https://github.com/itchysats/itchysats/pull/1965/commits/40284ed17e5ba34427844ce70f31d4b27e0bf04a)

maker/src/routes.rs
luckysori: You can run this against an old `automaker` without trouble? I'm very surprised.

I still think it would be cleaner to leverage `#[serde(default = "path")]`.

daemon/src/projection.rs
```diff
@@ -1255,8 +1255,8 @@ pub struct CfdOrder {
     #[serde(with = "round_to_two_dp")]
     pub lot_size: Usd,
 
-    #[serde(rename = "leverage")]
-    pub taker_leverage_choices: Leverage,
+    #[serde(rename = "leverage_choices")]
+    pub leverage_choices: Vec<Leverage>,
```
luckysori: 🔧 😛 

```suggestion
    pub leverage_choices: Vec<Leverage>,
```

daemon/src/maker_cfd.rs
luckysori: I would much prefer explicit variants for different versions of the protocol. Reading the code it is harder to understand that the message is used by two different versions: you have to read the docstring, inline comments and warning, and interpret nested branching logic.

But it's not a blocking concern.

taker/src/routes.rs
luckysori: Sorry, I meant the `if` statement specifically. We can't verify on the maker side that the requested leverage is acceptable?

maker/src/routes.rs
bonomat: Done in: [`6ad195d` (#1965)](https://github.com/itchysats/itchysats/pull/1965/commits/6ad195d0f80e4a0220786cea4fb585ab381dfb38)

daemon/src/projection.rs
bonomat: done in [`fc5ddaa` (#1965)](https://github.com/itchysats/itchysats/pull/1965/commits/fc5ddaae1ffb681dc356c07e7af26984900f8ca3)

luckysori (approved):
I mentioned a couple things that I would have done differently, but I'm happy to get this merged! Thanks for considering my review comments.

bonomat:
@luckysori : please check the last commit :) Happy to squash it then

luckysori (approved):
Thanks for the extra effort! I do believe it is the better approach.

PR #1970, opened 2022-05-05
Do not emit log if attestation time is within the minute

### LINKED ISSUE #1966: Don't fetch attestations on the hour
There is a timing issue when fetching attestations from the oracle: 

```
 {"timestamp":"2022-05-05T05:00:00.407204071Z","level":"DEBUG","fields":{"message":"Fetching attestation for /x/BitMEX/BXBT/2022-05-05T05:00:00.price?n=20"},"target":"daemon::oracle"}
...

{"timestamp":"2022-05-05T05:00:01.082597076Z","level":"DEBUG","fields":{"message":"Failed to fetch attestation: Failed to deserialize as Attestation: error decoding response body: attestation missing: attestation missing"},"target":"daemon::oracle"}
```

If you check the explorer the attestation was provided a few seconds later: 

https://h00.ooo/x/BitMEX/BXBT/2022-05-05T05:00:00.price?n=20

```
"time":"2022-05-05T05:00:08"
```

luckysori:
Fixes #1966.

What do you think, @bonomat? Alternatively I would just not bother, but I don't know how often we get misled by this log.

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -208,10 +208,18 @@ impl Actor {
                         anyhow::bail!("GET {url} responded with {code}");
                     }
 
-                    let attestation = response
-                        .json::<olivia::Attestation>()
-                        .await
-                        .context("Failed to deserialize as Attestation")?;
+                    let attestation = match response.json::<olivia::Attestation>().await {
+                        Ok(a) => a,
+                        Err(_)
+                            if OffsetDateTime::now_utc() < event_id.timestamp() + 1.minutes() =>
+                        {
+                            // we give Olivia a grace period of 1
+                            // minute to come up with any given
+                            // attestation
+                            return Ok(());
+                        }
+                        e => e.context("Failed to deserialize as Attestation")?,
+                    };
 
                     this.send(NewAttestationFetched {
                         id: event_id,


bonomat:
I think we overcomplicated our life with what is happening in `olivia.rs`. 
Imho best is to just deserialize to `Response` and check it `attestation` is `Some`, else ignore and fetch again later. 

https://github.com/itchysats/itchysats/blob/b4b41f2bc5d14a21d785544584eea2009c02d3c9/model/src/olivia.rs#L145-L149

luckysori:
> I think we overcomplicated our life with what is happening in `olivia.rs`. Imho best is to just deserialize to `Response` and check it `attestation` is `Some`, else ignore and fetch again later.
> 
> https://github.com/itchysats/itchysats/blob/b4b41f2bc5d14a21d785544584eea2009c02d3c9/model/src/olivia.rs#L145-L149

I see what you mean, but then we would never learn if Olivia has stopped attesting: we would continue to ignore and retry.

bonomat:
I think that's better than now. We are tracking how often we fetch announcements and if this keeps failing, e.g. nothing was successfully fetched in the last 1h, then we know that something is wrong.

bonomat:
> I think that's better than now. We are tracking how often we fetch announcements and if this keeps failing, e.g. nothing was successfully fetched in the last 1h, then we know that something is wrong.

What do you think of returning a custom error if attestation is null in here: 
https://github.com/itchysats/itchysats/blob/b4b41f2bc5d14a21d785544584eea2009c02d3c9/model/src/olivia.rs#L176

Then we can check on this particular error and ignore/debug print it instead.

luckysori:
> I think that's better than now. We are tracking how often we fetch announcements and if this keeps failing, e.g. nothing was successfully fetched in the last 1h, then we know that something is wrong.

I think it's important to note that Olivia can successfully produce announcements and fail to update the corresponding endpoints with an attestation. What you describe would miss this sort of problem if I understand it correctly.

> What do you think of returning a custom error if attestation is null in here:
> [snip]
> Then we can check on this particular error and ignore/debug print it instead.

It's good that you suggest this, because my approach in this PR misses the fact that deserialising into an attestation can fail for multiple reasons. I think we want to emit a custom error as you suggest; match against it when we deserialise; and then decide what to do.

I still think it might be good to silence any logs if the deserialisation error comes within the minute, but we don't need to be that clever.

luckysori:
I'm gonna close this since it's not that important and it should matter less with https://github.com/itchysats/itchysats/issues/1966.

Issue #1622, opened 2022-03-14
Make `ProposeRollover` more explicit


bonomat:
When rolling over both parties need to know from where they are rollingover from and to when/what they are rollingover to. If one of the values is different, the rollover might fail at a very late step e.g. when verifying the signatures. 

Right now, these values are assumed, i.e. both parties think they will rollover for another 24h, from the "last successful" rollover, handle their own feeAccount and subtract the feeAmounts, etc. 

We should make the `ProposeRollover` message more explicit in regards to where do we rollover from and until do we rollover to. That way we can ensure that both parties are talking about the same thing before starting with the actual protocol. 

Differently said, we (either party) should tell the other party the params they are going to use for the rollover. 

More discussion on this topic: https://github.com/itchysats/itchysats/discussions/1726

## Annoyance counter

- 01/04/2022: Rollover not being explicit leads to several rollover failures which leads to unintended publication of old states which triggers unimplemented revoke behaviour. This came up in https://github.com/itchysats/itchysats/pull/1715.
- 01/04/2022: Closed https://github.com/itchysats/itchysats/issues/1039 in favor of this one.
- 01/04/2022: Closed https://github.com/itchysats/itchysats/issues/1617 in favor of this one.

thomaseizinger:
This sounds like a good idea overall but it also highlights, how little design and thought went into the current rollover protocol. It would be good to push https://github.com/itchysats/architecture/pull/8 forward and fix these problems in a more structured way.

PR #1981, opened 2022-05-07
Send over fee account

bonomat:
This is more a band aid than a proper fix but should fix our ever failing rollovers, as now both parties are talking about the same rollover fees. 

The downside of this is that the `maker` could _trick_ the `taker` into paying a higher fee. 
For the time being, where we are the only maker, this should be good enough.

Ideally, we should ensure that a rollover guarantees to succeed for both parties. And hence, both parties always talk about the same rollover. 
Imho his is not easily achievable with the current implementation. 

Do Not Merge! If we want this, we will have to release from this branch!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.4.12.1] - 2022-05-06
+
 ## [0.4.12] - 2022-04-26
 
 ### Changed
@@ -118,7 +120,8 @@ Backport <https://github.com/itchysats/itchysats/pull/924> in an attempt to fix 
 
 Initial release for mainnet.
 
-[Unreleased]: https://github.com/itchysats/itchysats/compare/0.4.12...HEAD
+[Unreleased]: https://github.com/itchysats/itchysats/compare/0.4.12.1...HEAD
+[0.4.12.1]: https://github.com/itchysats/itchysats/compare/0.4.12...0.4.12.1
 [0.4.12]: https://github.com/itchysats/itchysats/compare/0.4.11...0.4.12
 [0.4.11]: https://github.com/itchysats/itchysats/compare/0.4.10...0.4.11
 [0.4.10]: https://github.com/itchysats/itchysats/compare/0.4.9...0.4.10
--- a/daemon/src/rollover_maker.rs
+++ b/daemon/src/rollover_maker.rs
@@ -177,6 +177,27 @@ impl Actor {
 
         let taker_id = self.taker_id;
 
+        // the maker computes the rollover fee and sends it over to the taker so that both parties
+        // are on the same page
+        let complete_fee = match rollover_params.version {
+            RolloverVersion::V1 => {
+                // Note there is actually a bug here, but we have to keep this as is to reach
+                // agreement on the fee for the protocol V1 version.
+                //
+                // The current fee is supposed to be added here, but we never noticed because in V1
+                // the fee is always charged for one hour using a static rate. This
+                // results in applying the fee in the DLC only for the next rollover
+                // (because we do apply the fee in the Cfd when loading the rollover
+                // event). Effectively this means, that we always charged one
+                // rollover too little.
+                rollover_params.fee_account.settle()
+            }
+            RolloverVersion::V2 => rollover_params
+                .fee_account
+                .add_funding_fee(rollover_params.current_fee)
+                .settle(),
+        };
+
         self.send_to_taker_actor
             .send(maker_inc_connections::TakerMessage {
                 taker_id,
@@ -185,6 +206,7 @@ impl Actor {
                     oracle_event_id,
                     tx_fee_rate,
                     funding_rate,
+                    complete_fee: complete_fee.into(),
                 },
             })
             .await
@@ -214,6 +236,7 @@ impl Actor {
             position,
             dlc,
             self.n_payouts,
+            complete_fee,
         );
 
         let this = ctx.address().expect("self to be alive");
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -43,11 +43,11 @@ use model::calculate_payouts;
 use model::olivia;
 use model::Cet;
 use model::Dlc;
+use model::FeeFlow;
 use model::Position;
 use model::RevokedCommit;
 use model::Role;
 use model::RolloverParams;
-use model::RolloverVersion;
 use model::SetupParams;
 use model::CET_TIMELOCK;
 use std::collections::HashMap;
@@ -353,6 +353,7 @@ pub async fn roll_over(
     our_position: Position,
     dlc: Dlc,
     n_payouts: usize,
+    complete_fee: FeeFlow,
 ) -> Result<Dlc> {
     let sk = dlc.identity;
     let pk = PublicKey::new(secp256k1_zkp::PublicKey::from_secret_key(SECP256K1, &sk));
@@ -378,24 +379,6 @@ pub async fn roll_over(
         .with_context(|| format_expect_msg_within("Msg0"))?
         .try_into_msg0()
         .context("Failed to read Msg0")?;
-
-    let complete_fee = match rollover_params.version {
-        RolloverVersion::V1 => {
-            // Note there is actually a bug here, but we have to keep this as is to reach agreement
-            // on the fee for the protocol V1 version.
-            //
-            // The current fee is supposed to be added here, but we never noticed because in V1 the
-            // fee is always charged for one hour using a static rate. This results in applying the
-            // fee in the DLC only for the next rollover (because we do apply the fee in the Cfd
-            // when loading the rollover event). Effectively this means, that we always charged one
-            // rollover too little.
-            rollover_params.fee_account.settle()
-        }
-        RolloverVersion::V2 => rollover_params
-            .fee_account
-            .add_funding_fee(rollover_params.current_fee)
-            .settle(),
-    };
 
     let maker_lock_amount = dlc.maker_lock_amount;
     let taker_lock_amount = dlc.taker_lock_amount;
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -16,6 +16,7 @@ use maia::secp256k1_zkp::SecretKey;
 use maia::CfdTransactions;
 use maia::PartyParams;
 use maia::PunishParams;
+use model::FeeFlow;
 use model::FundingRate;
 use model::Leverage;
 use model::MakerOffers;
@@ -174,6 +175,7 @@ pub enum MakerToTaker {
         oracle_event_id: BitMexPriceEventId,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        complete_fee: CompleteFee,
     },
     RejectRollover(OrderId),
     Settlement {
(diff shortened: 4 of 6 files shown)


bonomat:
Thinking about this more: I think this is actually the correct thing to do :)

- this is related: #1982 
- this PR goes towards #1622

CHANGELOG.md
```diff
@@ -118,7 +120,8 @@ Backport <https://github.com/itchysats/itchysats/pull/924> in an attempt to fix
 
 Initial release for mainnet.
 
-[Unreleased]: https://github.com/itchysats/itchysats/compare/0.4.12...HEAD
```
da-kami, line 121: ❓ Why not `0.4.13`?

daemon/src/wire.rs
```diff
@@ -174,6 +175,10 @@ pub enum MakerToTaker {
         oracle_event_id: BitMexPriceEventId,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        /// Fee to be paid for the rollover.
+        ///
+        /// The maker comes up with this amount so that both parties are on the same page
+        complete_fee: FeeFlow,
```
da-kami: 🛠 I don't think it's a good idea to use the `FeeFlow` directly here. I think it would be cleaner to convert into a struct/enum that is used over the wire only, so we can change the `FeeFlow` without breaking the wire messages.

da-kami: ❓ How does the application react if a different taker version tries to deserialize `ConfirmRollover` with this additional field? Does that work without error or are we breaking compatibility here?

da-kami: I think the cleaner way forward would be to introduce a new version of the `ConfirmRollover` message and distinguish how it will be handled in taker and maker.

daemon/src/rollover_maker.rs
```diff
@@ -177,6 +177,27 @@ impl Actor {
 
         let taker_id = self.taker_id;
 
+        // the maker computes the rollover fee and sends it over to the taker so that both parties
+        // are on the same page
+        let complete_fee = match rollover_params.version {
```
da-kami, line 182: 🛠 I don't think this distinction makes sense if you change the protocol to the maker coming up with the fee. If the maker comes up with the fee the fee should *always* be the complete amount, `RolloverVersion::V1` does not make sense here.

da-kami (changes requested):
I don't disapprove going into the direction of letting the maker come up with the fee for now, but I think this solution will cause trouble. Please have a look at the comments.

CHANGELOG.md
bonomat: We decided for the patch release to go for 0.4.12.1 

But why is this file changed in the first place? 🧐

daemon/src/wire.rs
bonomat: > ❓ How does the application react if a different taker version tries to deserialize ConfirmRollover with this additional field? Does that work without error or are we breaking compatibility here?

Yes, serde ignores additional fields in the json.

bonomat: > I think the cleaner way forward would be to introduce a new version of the ConfirmRollover message and distinguish how it will be handled in taker and maker.

I don't think this is necessary and unnecessarily adds complexity to our code base.

bonomat: Happy to introduce a different struct for that.

daemon/src/rollover_maker.rs
bonomat: Can do.

CHANGELOG.md
bonomat: > I don't disapprove going into the direction of letting the maker come up with the fee for now, but I think this solution will cause trouble. Please have a look at the comments.

What trouble do you see? Can you elaborate please.

bonomat:
> I don't disapprove going into the direction of letting the maker come up with the fee for now, but I think this solution will cause trouble. Please have a look at the comments.

What trouble do you see? Can you elaborate please.

daemon/src/wire.rs
bonomat: Actually, this is not easily to be changed. `FeeFlow` is used all the way down to the `payout_curve`

daemon/src/rollover_maker.rs
bonomat: see [`fa976dd` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/fa976dd8754345c9e9616bc5821bda3f253eb449)

daemon/src/wire.rs
bonomat: It's not easy to introduce a different type here given how `FeeFlow` is implemented. I've opted for just converting it into an enum owned by wire to achieve what you were proposing. 

[`93fdaa2` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/93fdaa2a65b118babeb8d5980bb42ec1392745ae)

CHANGELOG.md
bonomat: This commit is not part of this PR. 
It shows up because I branched away from the tag `0.4.12.1` and not the branch `0.4.12.1`. 
I think this commit should belong into the branch `0.4.12.1`.

cc @klochowicz

daemon/src/wire.rs
```diff
@@ -174,6 +175,10 @@ pub enum MakerToTaker {
         oracle_event_id: BitMexPriceEventId,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        /// Fee to be paid for the rollover.
+        ///
+        /// The maker comes up with this amount so that both parties are on the same page
```
luckysori: 🔧 Probably better if this docstring goes on the type.

model/src/lib.rs
```diff
@@ -760,9 +760,11 @@ impl FundingFee {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum FeeFlow {
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
     LongPaysShort(Amount),
+    #[serde(with = "::bdk::bitcoin::util::amount::serde::as_sat")]
```
luckysori: 🔧 I guess this is not necessary anymore.

daemon/src/wire.rs
```diff
@@ -174,6 +175,10 @@ pub enum MakerToTaker {
         oracle_event_id: BitMexPriceEventId,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        /// Fee to be paid for the rollover.
```
luckysori: 📣 We are not gonna be able to do rollovers with takers on a previous version. I think we need to introduce a new variant.

da-kami:
> > I don't disapprove going into the direction of letting the maker come up with the fee for now, but I think this solution will cause trouble. Please have a look at the comments.
> 
> What trouble do you see? Can you elaborate please.

Previous versions not working correctly given that we use the same message, but different logic internally.

daemon/src/setup_contract.rs
```diff
@@ -379,24 +380,6 @@ pub async fn roll_over(
...
-        RolloverVersion::V1 => {
-            // Note there is actually a bug here, but we have to keep this as is to reach agreement
-            // on the fee for the protocol V1 version.
-            //
-            // The current fee is supposed to be added here, but we never noticed because in V1 the
-            // fee is always charged for one hour using a static rate. This results in applying the
-            // fee in the DLC only for the next rollover (because we do apply the fee in the Cfd
-            // when loading the rollover event). Effectively this means, that we always charged one
-            // rollover too little.
-            rollover_params.fee_account.settle()
```
luckysori, line 392: Are we certain that we're not handling rollovers from `RolloverVersion::V1`? Otherwise removing this means that they will eventually fail to rollover.

luckysori (commented):
I understand the motivation but I'm not sure if we can do it like this.

daemon/src/setup_contract.rs
bonomat: I just checked our logs, we don't have any `ProposeRolloverV1` anymore.

daemon/src/wire.rs
bonomat: Why not? An old taker who does not know about this field will just ignore it when deserializing the message.

daemon/src/setup_contract.rs
luckysori: 14% of takers are on version 0.4.7 or prior according to our metrics 😨

model/src/lib.rs
bonomat: Removed in [`2c1fea6` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/2c1fea64473efb878d23b18c0b8734c66d1277b2)

daemon/src/setup_contract.rs
bonomat: Probably all with no position 🤷‍♂️
I'm happy to keep it here though. Not much harm done.

daemon/src/wire.rs
bonomat: Moved in [`75af806` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/75af8063283ec90b15367747f2601f8dc756f1e0)

bonomat:
> Previous versions not working correctly given that we use the same message, but different logic internally.

The logic was backwards compatible until your proposal of removing `ProposeRolloverV1`  which I did in here: [`fa976dd` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/fa976dd8754345c9e9616bc5821bda3f253eb449)
Before that either the maker sends it over to the taker. Or both parties compute it. 
I don't see any problems here.

daemon/src/wire.rs
luckysori: That is true.

But I think it is difficult to follow  some of the implicit logic introduced with this change. For example:

- Latest maker confirms rollover, sending over `complete_fee`.
- Old taker ignores `complete_fee`, goes into `setup_contract::roll_over`.
- Latest maker also goes into `setup_contract::roll_over` and uses the `complete_fee` passed in.
- Old taker calculates the `complete_fee` differently depending on the version.
- Rollover succeeds or fails depending on whether they agree or not.

I think by removing the rollover version handling code we run the risk of breaking old takers. I would have continued with adding a new version of the rollover protocol and a new confirmation message. That way we are explicitly aware of what is going on depending on what kind of taker we're dealing with. The code might be complex (several branches) but at least it's explicit.

daemon/src/setup_contract.rs
luckysori: We can certainly make the call that we no longer want to support those takers. But we should probably do that in a separate commit.

da-kami:
> > Previous versions not working correctly given that we use the same message, but different logic internally.
> 
> The logic was backwards compatible until your proposal of removing `ProposeRolloverV1` which I did in here: [`fa976dd` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/fa976dd8754345c9e9616bc5821bda3f253eb449) Before that either the maker sends it over to the taker. Or both parties compute it. I don't see any problems here.

I might have missed something in my earlier review. I think you are right and we should keep `ProposeRolloverV1` distinction on the maker. I was wrong, let's remove [`fa976dd` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/fa976dd8754345c9e9616bc5821bda3f253eb449)!

Let me try to think this through one more time:

1. New maker, new taker: 
    - The new maker calculates the fee to be used and sends it over.
    - The new taker takes whatever the maker sends and uses it.
3. New maker, old taker: 
    - The new maker calculates the fee to be used and sends it over. 
    - The old taker ignores it and calculates the fee as part of `setup_contract::rollover`. This should effectively be the same as what happened before - the only difference is that the maker calculates it at a different point.

No other combinations apply.

da-kami (approved):
Sorry for the confusion, once [`fa976dd` (#1981)](https://github.com/itchysats/itchysats/pull/1981/commits/fa976dd8754345c9e9616bc5821bda3f253eb449) was removed this should work! Testnet will reveal problems if there are any.

daemon/src/setup_contract.rs
bonomat: > We can certainly make the call that we no longer want to support those takers. But we should probably do that in a separate commit.

Let's leave it as it, i.e. with `RolloverVersion::V1`, then the version are compatible.

daemon/src/rollover_maker.rs
```diff
@@ -177,6 +177,27 @@ impl Actor {
...
+                // event). Effectively this means, that we always charged one
+                // rollover too little.
+                rollover_params.fee_account.settle()
+            }
+            RolloverVersion::V2 => rollover_params
+                .fee_account
+                .add_funding_fee(rollover_params.current_fee)
+                .settle(),
+        };
+
```
luckysori, line 200: I guess it's intentional to move this logic to the maker-specific code. It makes sense to me.

I would still prefer to keep this code where it was and add another rollover version, but I think this should work (and will be tested) and hopefully we can deprecate this protocol soon anyway.

luckysori (approved):

daemon/src/rollover_maker.rs
bonomat: Yes, it was intentional. How would you keep the code where it was and have the maker come up with the complete fee?

bonomat:
I see we got two approvals by @luckysori and @da-kami : do we want this to also be in the master or just for the path release branching away from 0.4.12.1?

bonomat:
> I see we got two approvals by @luckysori and @da-kami : do we want this to also be in the master or just for the path release branching away from 0.4.12.1?

If we decide to release from `master` then we don't need this PR but go for https://github.com/itchysats/itchysats/pull/1985 instead.

bonomat:
closing this as we merged https://github.com/itchysats/itchysats/pull/1985 into master and released from master

Issue #1974, opened 2022-05-06
rollover happening every two hours


bonomat:
This sounds like a bug in mainnet-v0.4.12 

We found that my taker is rolling over orders every 2h but asks for `hours_to_charge`:1`

```
{"timestamp":"2022-05-06T00:03:04.677445035Z","level":"DEBUG","fields":{"message":"Accepting rollover proposal","order_id":"8a21b8cc-4dc4-4dc8-a4be-3122f762154e","rollover_version":"V2","hours_to_charge":"1","funding_fee":"0.00000001 BTC"},"target":"model::cfd"}
{"timestamp":"2022-05-05T22:03:04.0474127Z","level":"DEBUG","fields":{"message":"Accepting rollover proposal","order_id":"8a21b8cc-4dc4-4dc8-a4be-3122f762154e","rollover_version":"V2","hours_to_charge":"1","funding_fee":"0.00000001 BTC"},"target":"model::cfd"}
```

bonomat:
Additional note: 

My taker is sending **two** rollover proposals at the same time.

luckysori:
### Analysis

After some time investigating this, this is what I have figured out:

- A CFD is eligible for extension via rollover as soon as the settlement time is 23 hours in the future or less:

https://github.com/itchysats/itchysats/blob/3bc975548c05eb141fe61724625393839d134cb4/model/src/cfd.rs#L787-L797

- We check for rollover eligibility every 5 minutes:

https://github.com/itchysats/itchysats/blob/3bc975548c05eb141fe61724625393839d134cb4/daemon/src/auto_rollover.rs#L140-L144

- Combining the first two points, we expect most CFDs to be rolled over when the settlement time is between 23 hours and 22 hours and 55 minutes in the future. Since BitMEX event IDs are based on whole hours, it follows that the timestamp at which rollover is expected to be triggered should have a value of 0-5 for the minutes.

- When we perform a rollover, we set the new settlement time between 24 and 25 hours in the future, as can be seen here: 

https://github.com/itchysats/itchysats/blob/3bc975548c05eb141fe61724625393839d134cb4/maker/src/rollover.rs#L175-L176

where `interval` is always 24 hours.

- The `next_announcement_after` API returns an event ID with a timestamp close to 25 hours whenever the current timestamp (i.e. the value returned by calling `OffsetDateTime::now_utc`) has low minutes.

I think this explains why most rollovers happen every 2 hours.

### Solution

[...]

PR #1984, opened 2022-05-09
Trace time left in CFD when calculating hours to extend in rollover

luckysori:
This can give us extra context when trying to figure out if a CFD is being rolled over for the correct number of hours.

---

Motivated by https://github.com/itchysats/itchysats/issues/1974.

It's important to highlight that if a rollover extends the lifetime of a CFD by 1 hour and 59 minutes, we are currently charging for only 1 hour. We may want to be less generous given that quite a lot of CFDs are not getting rolled over consistently every hour.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1425,6 +1425,8 @@ impl Cfd {
 
         let hours_left = settlement_time - OffsetDateTime::now_utc();
 
+        tracing::trace!(target = "cfd", time_left_in_cfd = %hours_left, "Calculating hours to extend in rollover");
+
         if !hours_left.is_positive() {
             tracing::warn!("Rolling over a contract that can be settled non-collaboratively");
 


model/src/cfd.rs
```diff
@@ -1425,6 +1425,8 @@ impl Cfd {
 
         let hours_left = settlement_time - OffsetDateTime::now_utc();
 
+        tracing::trace!(time_left_in_cfd = %hours_left, "Calculating hours to extend in rollover");
```
da-kami:
```suggestion
        tracing::trace!(target="rollover",time_left_in_cfd = %hours_left, "Hours to extend in rollover");
```

Not sure if `rollover` is a good target as per the definition of how `target` is to be used. But given that this is a `trace` log I think we need a target because we don't want all logs on trace. 

Additionally, as far as I understand the code, this is already calculated at this point.

luckysori: > Not sure if rollover is a good target as per the definition of how target is to be used. But given that this is a trace log I think we need a target because we don't want all logs on trace.

Good point. I'll fix this.

> Additionally, as far as I understand the code, this is already calculated at this point.

`time_left_in_cfd != hours_to_extend`.

model/src/cfd.rs
```diff
@@ -1425,6 +1425,8 @@ impl Cfd {
 
         let hours_left = settlement_time - OffsetDateTime::now_utc();
 
+        tracing::trace!(target = "cfd", time_left_in_cfd = %hours_left, "Calculating hours to extend in rollover");
```
bonomat, line 1428: Where are we printing `hours_to_extend` and can we combine these two log messages?

bonomat (approved):
👍

model/src/cfd.rs
luckysori: That was my original approach, but we would have to calculate the value of `hours_left` again just for logging. I felt like this information is lower level than the other values we print, so outputting it separately on the trace level made sense to me.

luckysori: Here's the other log:

https://github.com/itchysats/itchysats/blob/3d749ea2767a4279b4ed96b0fa1a262f5743c7a7/model/src/cfd.rs#L921-L927

PR #1985, opened 2022-05-09
Maker to send over fee account

bonomat:
This way we can ensure that both parties are talking about the same fee.

Same changes as https://github.com/itchysats/itchysats/pull/1981


bonomat:
This change was already approved in https://github.com/itchysats/itchysats/pull/1981 hence merging this. 

bors r+

PR #1999, opened 2022-05-10
Enable metric endpoint for taker

### LINKED ISSUE #1980: Enable metrics on taker
We should enable the metrics on the taker as well for better monitoring.

They are already collected but the endpoint is not provided.

bonomat:
Fixes #1980 

This allows us to collect metrics for the taker as well.

In a discussion with Thomas we thought about having a feature flag for this but I opted against it because that would mean we need to release 2 versions: one with and one without this feature. 

What do you think? Shall I add a feature flag or maybe a new cli arg which allows to enable the metric endpoint?


luckysori (approved):
A feature flag may be nice, but I don't see a big issue with including this by default.

bonomat:
merging this without feature flag. It's behind the authenticated API so it is not a security issue. 

bors r+

da-kami (approved):

PR #2003, opened 2022-05-10
Add small offset to timestamp before fetch

### LINKED ISSUE #1966: Don't fetch attestations on the hour

bonomat:
We expect the oracle to attest to the price at `timestamp` but due to spacetime distortion we give it another minute before we fetch. This should reduce the error below because we will only fetch the attestation if it is _very_ likely to have occurred.

```json
{"timestamp":"2022-05-10T07:00:03.759931111Z","level":"DEBUG","fields":{"message":"Failed to fetch attestation: Failed to deserialize as Attestation: error decoding response body: attestation missing: attestation missing"},"target":"daemon::oracle"}
```

Resolves #1966 

What do you think @luckysori ?

### DIFF
--- a/model/src/olivia.rs
+++ b/model/src/olivia.rs
@@ -75,7 +75,7 @@ impl BitMexPriceEventId {
     /// Checks whether this event has likely already occurred.
     ///
     /// We can't be sure about it because our local clock might be off from the oracle's clock.
-    pub fn has_likely_occured(&self) -> bool {
+    pub fn has_likely_occurred(&self) -> bool {
         let now = OffsetDateTime::now_utc();
 
         now > self.timestamp + Duration::minutes(1)
@@ -444,6 +444,6 @@ mod tests {
         let past_event =
             BitMexPriceEventId::with_20_digits(datetime!(2021-09-23 10:00:00).assume_utc());
 
-        assert!(past_event.has_likely_occured());
+        assert!(past_event.has_likely_occurred());
     }
 }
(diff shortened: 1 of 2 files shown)


model/src/olivia.rs
```diff
@@ -74,10 +75,10 @@ impl BitMexPriceEventId {
     /// Checks whether this event has likely already occurred.
     ///
     /// We can't be sure about it because our local clock might be off from the oracle's clock.
-    pub fn has_likely_occured(&self) -> bool {
+    pub fn has_likely_occurred(&self) -> bool {
         let now = OffsetDateTime::now_utc();
 
-        now > self.timestamp
+        now > self.timestamp + Duration::minutes(1)
```
luckysori, line 81: 🙃 You can also use this, which is a tiny bit more readable:

```suggestion
        use time::ext::NumericalDuration;
        now > self.timestamp + 1.minutes()
```

luckysori (approved):
I think this is a good idea.

PR #2008, opened 2022-05-11
Make `10000` default maker port on `testnet`

da-kami:
This aligns the default maker port between testnet and mainnet.


bonomat (approved):

da-kami:
This PR is actually pointless because the taker would spam for a different reason (`maker-id` not matching). What we should really do is remove the `--maker` and `--maker-id` from the Umbrel setup!

I am doing that as part of updating the Umbrel PR.
I don't think we have to change the ports.

PR #2010, opened 2022-05-11
Apply timeout to all message handlers of connection actors

### LINKED ISSUE #2005: Configure timeouts for connection actor(s?)
Use the functionality added in https://github.com/comit-network/xtra/pull/4, to ensure that the connection actor(s?) always time out after some time attempting to handle a message.

---
Source issue: https://github.com/itchysats/itchysats/issues/1954.

luckysori:
Fixes #2005.

This is so that we can ensure that no message can block the actor forever.

Things to consider:

- The original issue suggested that we may want to remove some internal timeouts. I have not dared to touch that, but maybe someone else with more context can make a call.
- The value of 120 seconds is completely arbitrary. The way this new `xtra` feature works, the timeout applies to _all_ handlers for a given `Actor`. This suggests to me that we want to be cautious with this value. But 120 seconds might be too much time for it to be useful.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -180,13 +180,17 @@ where
             .checked_mul(2)
             .expect("not to overflow");
 
-        tasks.add(connection_actor_ctx.run(connection::Actor::new(
-            maker_online_status_feed_sender,
-            &cfd_actor_addr,
-            identity.identity_sk,
-            taker_heartbeat_timeout,
-            connect_timeout,
-        )));
+        tasks.add(
+            connection_actor_ctx
+                .with_handler_timeout(Duration::from_secs(120))
+                .run(connection::Actor::new(
+                    maker_online_status_feed_sender,
+                    &cfd_actor_addr,
+                    identity.identity_sk,
+                    taker_heartbeat_timeout,
+                    connect_timeout,
+                )),
+        );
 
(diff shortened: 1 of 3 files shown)


klochowicz (approved):
I like the conservative approach. We can experiment with removing some of the inner timeouts later, but having them should not hurt.

daemon/src/lib.rs
```diff
@@ -180,13 +180,17 @@ where
...
-        tasks.add(connection_actor_ctx.run(connection::Actor::new(
-            maker_online_status_feed_sender,
-            &cfd_actor_addr,
-            identity.identity_sk,
-            taker_heartbeat_timeout,
-            connect_timeout,
-        )));
+        tasks.add(
+            connection_actor_ctx
+                .with_handler_timeout(Duration::from_secs(120))
```
bonomat, line 185: Given 120 is arbitrary, maybe it's worth mentioning this in the code or in the commit message that this is arbitrary. Otherwise we will wonder in the future why this number was chosen.

klochowicz: agreed. it might be worth extracting as a const so that it's easily found at the later point! (then a comment would have a nice spot too).

luckysori: I agree, will do.

da-kami: as follow up please :)

PR #2011, opened 2022-05-11
Allow updating maia in a backwards-compatible fashion

klochowicz:
Build a map of taker daemon versions when they connect to the maker. This allows
us to determine a policy in regards to which maia version to use during payout curve
calculation.

In this case, if the taker version is below 0.4.15 or unknown, use legacy maia version.

### DIFF
--- a/maker/src/cfd.rs
+++ b/maker/src/cfd.rs
@@ -1,5 +1,6 @@
 use crate::collab_settlement;
 use crate::connection;
+use crate::connection::TakerDaemonVersion;
 use crate::contract_setup;
 use crate::future_ext::FutureExt;
 use crate::metrics::time_to_first_position;
@@ -35,6 +36,8 @@ use model::SettlementProposal;
 use model::Timestamp;
 use model::TxFeeRate;
 use model::Usd;
+use semver::VersionReq;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use time::Duration;
 use tokio_tasks::Tasks;
@@ -46,6 +49,16 @@ use xtras::SendAsyncSafe;
 
 const HANDLE_ACCEPT_CONTRACT_SETUP_MESSAGE_TIMEOUT: std::time::Duration =
     std::time::Duration::from_secs(10);
+
+/// Determine which maia version to use in protocol execution
+///
+/// Older version of maia had a bug in which the CETs were calculated incorrectly.
+/// This enum ensures that we can be backwards-compatible with takers running older version.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum MaiaProtocolVariant {
+    Old,
+    New,
+}
 
 #[derive(Clone)]
 pub struct NewOffers {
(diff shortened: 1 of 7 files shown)


klochowicz:
marked it as 'draft' as @bonomat changes that would actually import 2 versions of maia need to come first and be used here :-) (either via duplication of API or via importing 2 versions of the crate).

BTW: should we bump `maia`'s  version? it's a breaking change there!

maker/src/cfd.rs
```diff
@@ -47,6 +50,16 @@ use xtras::SendAsyncSafe;
 const HANDLE_ACCEPT_CONTRACT_SETUP_MESSAGE_TIMEOUT: std::time::Duration =
     std::time::Duration::from_secs(10);
 
+/// Determine which maia version to use in protocol execution
+///
+/// Older version of maia had a bug in which the CETs were calculated incorrectly.
+/// This enum ensures that we can be backwards-compatible with takers running older version.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum MaiaProtocolVariant {
+    Old,
```
da-kami, line 59: 🙃 How about one of these instead:

* `V1` and `V2`
* The old maia version that contains the bug as `x.y.z` and `CURRENT`

maker/src/cfd.rs
```diff
@@ -185,6 +200,7 @@ pub struct Actor<O, T, W> {
     oracle: xtra::Address<O>,
     time_to_first_position: xtra::Address<time_to_first_position::Actor>,
     connected_takers: HashSet<Identity>,
+    taker_versions: HashMap<Identity, Option<TakerDaemonVersion>>,
```
da-kami, line 203: 🙃 This was the simplest way of keeping this information where it is needed, but I am not sure I agree that it should be here. I think it would be better if we would keep the `taker_version` in the connection actor and dispatch different messages there. 
This would be mean more code, but I think it would be cleaner because we encapsulate the decision what version it is in one place. 

I am not totally against the proposed solution, but we scatter decision logic based on network versions over multiple actors which might get harder to maintain and clean up.

da-kami, line 203: 🙃 Could also just be `HashMap<Identity, TakerDaemonVersion>` and we just don't insert in case there is no version. We default to `None` (resulting in `Old`) anyway if we don't find an entry.

da-kami (commented):
I think it would be cleaner to keep the decision logic based on the version in the connection and dispatch different messages. Yes, that means more code but we keep the decision and network version logic in one place and don't have it in multiple actors.
I don't feel strong about this - if you think this is the best way forward we can roll with this, but I am a bit wary of having decision logic around the network version in multiple actors 😬

bonomat:
@klochowicz : is this still needed? If we rollout rollover using libp2p everyone will need to be on the new protocol, meaning, we could make the distinguishing between legacy network stack and libp2p-based network stack

klochowicz:
> @klochowicz : is this still needed? If we rollout rollover using libp2p everyone will need to be on the new protocol, meaning, we could make the distinguishing between legacy network stack and libp2p-based network stack

yes, I like this idea! with libp2p, this mechanism will not be needed anyway. closing this PR.

PR #2012, opened 2022-05-11
Omit 0-amount outputs when reconstructing CETs

luckysori:
This fixes a bug where we were naively adding 0-amount outputs for the party that was getting liquidated when reconstructing "liquidation" CETs.

The network would not accept these transactions, but we were failing as soon as we were checking against the original CET TXID. In `maia` we were already constructing this kind of transaction correctly, by omitting the 0-amount output.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -269,9 +269,12 @@ pub async fn new(
     // need some fallback handling (after x time) to spend the outputs in a different way so the
     // other party cannot hold us hostage
 
-    let maker_script_pubkey = params.maker().address.script_pubkey();
-    let taker_script_pubkey = params.taker().address.script_pubkey();
-    let cets = tokio::task::spawn_blocking(move || {
+    let cets = tokio::task::spawn_blocking({
+        let maker_address = params.maker().address.clone();
+        let taker_address = params.taker().address.clone();
+        let commit_tx = commit_tx.clone();
+        let commit_desc = commit_desc.clone();
+        move || {
         own_cets
             .into_iter()
             .map(|grouped_cets| {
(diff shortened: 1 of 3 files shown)


da-kami (approved):
Would it be worth to add a unit test for this?

luckysori:
> Would it be worth to add a unit test for this?

Fair question.

It is hard to unit test because I think the property we want to test is: can we reconstruct all CETs correctly i.e. the same as `maia` constructs them originally. For that we would have to run a contract setup (and a rollover) and check that the intermediate transactions that we drop match the `Cet` structs that we generate based on them. I think this would be hard to set up.

Perhaps we could add a `debug_assert` in `setup_contract::new/roll_over` to check this property _during_ contract setup.

da-kami:
> > Would it be worth to add a unit test for this?
> 
> Fair question.
> 
> It is hard to unit test because I think the property we want to test is: can we reconstruct all CETs correctly i.e. the same as `maia` constructs them originally. For that we would have to run a contract setup (and a rollover) and check that the intermediate transactions that we drop match the `Cet` structs that we generate based on them. I think this would be hard to set up.
> 
> Perhaps we could add a `debug_assert` in `setup_contract::new/roll_over` to check this property _during_ contract setup.

I get that, but we could still add simple tests to ensure different scenarios - regardless of the CET matching issue:

1. both amounts `>0` will create tx with two outputs.
2. taker amount `0` will create tx with one output.
3. maker amount with `0` will create tx with one output. 

This is more for test coverage to ensure that future changes don't result in trouble.

If you feel that does not add enough value feel free to merge without.

luckysori:
> This is more for test coverage to ensure that future changes don't result in trouble.
>
> If you feel that does not add enough value feel free to merge without.

I think that is a valid approach and it would give us some confidence. What about what I propose in https://github.com/itchysats/itchysats/pull/2012/commits/9bbee9aa00906f3bfcaea94bea5e9868d24bb55b? We pay a small cost in debug builds but we can detect all kinds of problems related to this reconstruction.

daemon/src/setup_contract.rs
```diff
@@ -568,20 +580,28 @@ pub async fn roll_over(
...
                         maker_amount,
                         taker_amount,
                         adaptor_sig: *other_encsig,
                         range: digits.range(),
                         n_bits: digits.len(),
                         txid: tx.txid(),
-                    })
+                    };
+
+                    debug_assert_eq!(
```
da-kami, line 597: I think this is fair, but it would be interesting to have some stats because rollovers are already expensive and we would do that for *each* CET. 
Could you maybe put that into a separate PR and we merge it later and try to get some stats on it on `testnet`?
I'd like to get the fix in without this, because it is not predictable if more rollovers will fail with this change.

da-kami (changes requested):
Please move the last commit into a separate PR!

daemon/src/setup_contract.rs
luckysori: The change only applies to debug builds! So it doesn't affect production. I can put it in a separate PR, but I would merge it straight away 😛

da-kami (approved):
Sry, missed that it was a debug assert!

PR #2013, opened 2022-05-11
Move taker id to modal and some resposiveness

da-kami:
Scales better:

Wide screen:

image

Narrow screen:

image

Mobile (not perfect, but at least all header and footer items are all there):

image

--- 

Taker ID is not important enough to always display it, I added an `info` model where one can find it. We can store some other meta info (e.g. version, links to code, ...) there and layout nicer (did not invest time for now):

image


luckysori (approved):
Fine by me!

PR #2014, opened 2022-05-11
Weird cfd log metrics fixes

da-kami:
That fixes the logs.

@bonomat If you prefer to restrict the `rejected` and `failed` CFD by ID let me know. We could also clean with up in the database, but I am not sure it's worth it. I think the condition I added is good enough.

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -376,6 +376,7 @@ mod metrics {
                         is_closed = cfd.is_closed,
                         is_refunded = cfd.is_refunded,
                         is_rejected = cfd.is_rejected,
+                        is_failed = cfd.is_failed,
                         order_id = %cfd.id,
                         "CFD is in weird state"
                     );
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -168,10 +168,19 @@ impl Cfd {
                 is_open: true,
                 ..self
             },
-            ContractSetupFailed => Self {
-                is_failed: true,
-                ..self
-            },
+            ContractSetupFailed => {
+                // This is needed due to a bug that has since been fixed; `OfferRejected` and
+                // `ContractSetupFailed` are mutually exclusive.
+                // We give `OfferRejected` priority over `ContractSetupFailed`.
+                if self.is_rejected {
+                    Self { ..self }
+                } else {
+                    Self {
+                        is_failed: true,
+                        ..self
+                    }
+                }
+            }
             OfferRejected => Self {
                 is_rejected: true,
                 ..self
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -164,7 +164,7 @@ impl Cfd {
                 is_refunded: false,
                 ..self
             },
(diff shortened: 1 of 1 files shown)


daemon/src/position_metrics.rs
```diff
@@ -168,10 +168,20 @@ impl Cfd {
...
             },
-            ContractSetupFailed => Self {
-                is_failed: true,
-                ..self
-            },
+            ContractSetupFailed => {
+                // This is needed because we have a CFD (bd6221b0-47f6-4b14-ac9b-04cfa355cb9b) in
+                // the database where `OfferRejected` and `ContractSetupFailed` was appended. This
+                // was due to a bug that has since been fixed; `OfferRejected` and
+                // `ContractSetupFailed` are mutually exclusive.
```
luckysori: 🙃 I think this comment should exist without any reference to production. The motivation is to ensure that we only have one state at a time, in case we can emit more than one "final" event.

luckysori (approved):
Thank you for fixing this!

I think we should replace the booleans with an `enum`. The final states are mutually exclusive. There is no need to make the aggregate more complex than it needs to be.

daemon/src/position_metrics.rs
```diff
@@ -168,10 +168,20 @@ impl Cfd {
...
+                if self.is_rejected {
```
bonomat, line 175: What do you think of logging something here? I understand that *this should never happen*, but if it ever happen, it should mean something is wrong.

bonomat (approved):
Thanks for the fixes

daemon/src/position_metrics.rs
luckysori: I'll insist on my earlier point: I think this would be simpler if we just have a `CfdState` enum with all the variants and you don't have to reason about what was there before. We simply overwrite it with the state that corresponds to the last event we process.

daemon/src/position_metrics.rs
da-kami: > 🙃 I think this comment should exist without any reference to production. The motivation is to ensure that we only have one state at a time, in case we can emit more than one "final" event.

I agree! But before tackling this I'd like to take a step back and look at the big picture of aggregates currently in the code-base. We had the idea of crating one aggregate for projection and metrics at some point. I am not sure if it's a good idea but I'd like to evaluate it once we have a look into how to make the projection "better".
I feel the aggregates have been sources of bugs for some times and we keep running into similar/same problems with different aggregates when we change things. For me this points to not being the right abstraction. 
I will merge this fix for now. So we don't forget the refactoring idea I recorded it as a ticket but would not action it until we have a better high level picture :)
https://github.com/itchysats/itchysats/issues/2022

daemon/src/position_metrics.rs
da-kami: > What do you think of logging something here? I understand that this should never happen, but if it ever happen, it should mean something is wrong.

I think this specific case is actually not a big deal. Even if it happens - both of them are treated as failed internally. Since we know that we have one old position that has this issue I would not log anything for now. We already have checks for this in place when we move CFDs into closed / failed - if we want to log I'd do it there.

daemon/src/position_metrics.rs
luckysori: You probably meant to reply to a different message 😛

PR #2019, opened 2022-05-11
Fix payout / P/L of closed CFDs

da-kami:
Use the `payout` field for decision making. If it was set prior to `with_current_quote` we just exit because the `payout` is already set.
Like this we are independent of the state; which is necessary to make sure that `CfdAggregate` is still handled correctly when being closed.

Note that this is a somewhat hacky fix. We will run into trouble with this logic in the projection again once we make changes. 
The `Aggregate` within three `xyzCfdAggregate`s is error prone, we should refactor and simplfy this.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -556,8 +556,9 @@ impl Cfd {
     }
 
     pub fn with_current_quote(self, latest_quote: Option<xtra_bitmex_price_feed::Quote>) -> Self {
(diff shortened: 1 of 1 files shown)


da-kami:
This should fix the problem, refactoring of projection can be scheduled later

luckysori (approved):
The fix makes sense to me! Let's see what the tests think.

PR #2023, opened 2022-05-12
Reduce time to sync wallet and sync less often

bonomat:
We only want to sync for 1000 addresses during the first sync. This is because we might have generated over 1000 addresses in the past and if we would not sync that far ahead we would miss some balance.
For repeatedly sync it is enough to sync for the default 100 addresses because everytime we generate a change address it is automatically included.

This is a minor improvement because generating 1000 addresses and syncing for it takes some time.

The last commit might be *a* reason for our ever growing ram usage.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -135,8 +135,6 @@ impl Actor<ElectrumBlockchain> {
 
 impl Actor<ElectrumBlockchain> {
     fn sync_internal(&mut self) -> Result<WalletInfo> {
-        self.wallet.ensure_addresses_cached(1000)?;
-
         self.wallet
             .sync(&self.blockchain_client, SyncOptions::default())
             .context("Failed to sync wallet")?;
@@ -178,11 +176,9 @@ impl Actor<ElectrumBlockchain> {
             Ok(wallet_info) => Some(wallet_info),
             Err(e) => {
                 tracing::debug!("{:#}", e);
-
                 None
             }
         };
-
         let _ = self.sender.send(wallet_info_update);
     }
 
@@ -285,6 +281,15 @@ impl xtra::Actor for Actor<ElectrumBlockchain> {
     type Stop = ();
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("self to be alive");
+
+        // We only cache the addresses at startup
+        if let Err(e) = self
+            .wallet
+            .ensure_addresses_cached(1000)
+            .with_context(|| "Could not cache addresses")
+        {
+            tracing::warn!("{:#}", e);
+        }
 
         self.tasks.add(this.send_interval(SYNC_INTERVAL, || Sync));
     }
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -32,7 +32,7 @@ use tokio_tasks::Tasks;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
 
-const SYNC_INTERVAL: Duration = Duration::from_secs(10);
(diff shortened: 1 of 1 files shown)


daemon/src/wallet.rs
```diff
@@ -286,6 +305,9 @@ impl xtra::Actor for Actor<ElectrumBlockchain> {
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("self to be alive");
 
+        this.send_async_safe(InitialSync)
+            .await
+            .expect("we just started");
         self.tasks.add(this.send_interval(SYNC_INTERVAL, || Sync));
```
klochowicz, line 296: please be aware this might do `InitialSync` and `Sync` almost at the same time. perhaps we could move adding the task to `InitialSync` handler (or dispatch it synchronously, so the task is only added after the initial sync?). The first option out of the two seems nicer.

klochowicz (approved):
happy to give it a go, let's just avoid syncing twice on startup :)

nice ony with adding the new debug info as well.

daemon/src/wallet.rs
```diff
@@ -134,9 +135,17 @@ impl Actor<ElectrumBlockchain> {
 }
 
 impl Actor<ElectrumBlockchain> {
-    fn sync_internal(&mut self) -> Result<WalletInfo> {
-        self.wallet.ensure_addresses_cached(1000)?;
+    fn initial_sync(&mut self) -> Result<WalletInfo> {
+        match self.wallet.ensure_addresses_cached(1000) {
```
luckysori: 🔧 Use `anyhow::Context` instead of matching?

daemon/src/wallet.rs
```diff
@@ -173,16 +182,26 @@ impl Actor<ElectrumBlockchain> {
 
 #[xtra_productivity]
 impl Actor<ElectrumBlockchain> {
+    pub fn handle_initial_sync(&mut self, _msg: InitialSync) {
+        tracing::trace!(target : "wallet", "Wallet initial sync started");
```
luckysori: 🙃 Could put this on `info` since it happens once per restart.

daemon/src/wallet.rs
```diff
@@ -33,7 +33,7 @@ use xtra_productivity::xtra_productivity;
 use xtras::SendAsyncSafe;
 use xtras::SendInterval;
 
-const SYNC_INTERVAL: Duration = Duration::from_secs(10);
+const SYNC_INTERVAL: Duration = Duration::from_secs(60);
```
luckysori, line 35: 🤡 We could consider basing this value on the network, but it may not matter too much.

luckysori (approved):
I like the idea!

I would consider @klochowicz's suggestion before merging, but other than that it LGTM.

daemon/src/wallet.rs
bonomat: What do you mean with this? Sync more often on testnet?

daemon/src/wallet.rs
bonomat: I know what you mean but I can't get it working. 
Afaik I can't use `this.send(...)` as this blocks the actor. Can you provide a patch please?

klochowicz: I had a stab, and figured out that the easiest way would be to actually block in the `started()` waiting for the cache - if you deem this unacceptable, please drop the commit I added to this PR - note that the actual sync will be executed in the task.
I wrote in the description that we should be OK, as we don't expect anything to work without the cache anyway (so it's a part of the construction phase).

bonomat: that's a good idea. Did you try it and does it run?

PR #2025, opened 2022-05-12
Only allow one rollover at the time

da-kami:
This is a desperate attempt to get more stability into the maker's connection actor.
With this change we only allow one rollover at the time to reduce the load on the maker.


bonomat (approved):
Let's give this a try

PR #2026, opened 2022-05-13
Only allow two rollover at a time

da-kami:
With this change we only allow two rollover at the time to reduce the load on the maker.

The current amount of CFDs requires to allow more than 1, so we opt for 2 for now until we know if that helps with stability.


luckysori (approved):

bonomat (approved):
Let's give this a try :)

bonomat:
bors retry

PR #2027, opened 2022-05-13
Log `OrderId` next to message whenever possible

luckysori:
It's not always available (not all messages are associated with and `OrderId`), which is why we have to use the `tracing` syntax
`?order_id`.

The context of when a message was sent or received, associated with an `OrderId`, should be useful for debugging.

### DIFF
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -159,6 +159,20 @@ impl TakerToMaker {
             TakerToMaker::Unknown => "TakerToMaker::Unknown",
         }
     }
+
+    pub fn order_id(&self) -> Option<OrderId> {
+        use TakerToMaker::*;
+        match self {
+            DeprecatedTakeOrder { order_id, .. }
+            | TakeOrder { order_id, .. }
+            | ProposeRollover { order_id, .. }
+            | ProposeRolloverV2 { order_id, .. }
+            | Protocol { order_id, .. }
+            | RolloverProtocol { order_id, .. }
+            | Settlement { order_id, .. } => Some(*order_id),
+            Hello(_) | HelloV2 { .. } | Unknown => None,
+        }
+    }
 }
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


klochowicz (approved):
cool, didn't know about the ?syntax 👍

luckysori:
> cool, didn't know about the ?syntax 👍

Ha me neither! To be fair it's not exactly what I wanted, because it's just to signify that we want to use the `Debug` implementation. The commit message wasn't clear about this.

It's not ideal so we may decide to revert this change eventually. But I think it could help with debugging our current problems.

luckysori:
I ended up settling for matching against the result of calling `order_id`. That way we don't have to deal with ugly `Debug` implementations of `Option<OrderId>` in our logs.

daemon/src/wire.rs
```diff
@@ -159,6 +159,20 @@ impl TakerToMaker {
...
+        use TakerToMaker::*;
+        match self {
+            DeprecatedTakeOrder { order_id, .. }
+            | TakeOrder { order_id, .. }
+            | ProposeRollover { order_id, .. }
+            | ProposeRolloverV2 { order_id, .. }
+            | Protocol { order_id, .. }
+            | RolloverProtocol { order_id, .. }
+            | Settlement { order_id, .. } => Some(*order_id),
+            _ => None,
```
da-kami: 🛠 Please don't use `_` here - if we add a new message with `order_id` we would ignore the `order_id`, if you mention all choices we will notice when adding the new message to the match.

daemon/src/wire.rs
```diff
@@ -282,6 +296,22 @@ impl MakerToTaker {
...
+            CurrentOrder(Some(DeprecatedOrder047 { id: order_id, .. }))
+            | ConfirmOrder(order_id)
+            | RejectOrder(order_id)
+            | InvalidOrderId(order_id)
+            | Protocol { order_id, .. }
+            | RolloverProtocol { order_id, .. }
+            | ConfirmRollover { order_id, .. }
+            | RejectRollover(order_id)
+            | Settlement { order_id, .. } => Some(*order_id),
+            _ => None,
```
da-kami: 🛠 Same here

da-kami (approved):
ack!
But please be explicit in the match and don't use `_` :)

luckysori:
bors retry

luckysori:
bors retry

PR #2028, opened 2022-05-13
Enable xtra tracing above warn

da-kami:
Also explicitly activates the xtra features on the `maker` crate.


klochowicz (approved):

luckysori (approved):
Good catch!

PR #2041, opened 2022-05-16
Cull old DLC data

luckysori:
I started to implement the proposal outlined here: https://github.com/itchysats/itchysats/discussions/1867#discussioncomment-2670977. In the end I decided to do something a bit different: instead of segregating the data (i.e. separate DLC from events), we make the `dlc` field in the relevant events optional. None of the event-handling logic had to change to accommodate for the introduction of optional `Dlc`s.

~This PR introduces a breaking change to the events we store in the database. We therefore will need to leverage the fact that we _can_ now make breaking changes to this layer provided we only have closed or failed CFDs.~
**EDIT**: See https://github.com/itchysats/itchysats/pull/2041#issuecomment-1129441722 for why we don't think this PR introduces breaking changes to the database.

TODOs:

- [x] Introduce actor which uses the new `db` API.
- [x] Use actor in `MakerActorSystem`.
- [x] Add a couple more tests.
- [x] Refactor `cull_dlcs` slightly.


daemon/src/db/cull_dlcs.rs
```diff
@@ -0,0 +1,187 @@
...
+    name,
+    data
+FROM
+    events
+WHERE
+    name = $1 OR
+    name = $2
+ORDER BY
+    created_at DESC
+LIMIT -1 OFFSET 1
```
bonomat: This should not be necessary, i.e. you want to select all anyway and `limit -1, offset 1` would _limit_ your results from index `-1`, i.e. all :)

daemon/src/db/cull_dlcs.rs
```diff
@@ -0,0 +1,187 @@
...
+
+            let event = match data {
+                EventKind::ContractSetupCompleted { dlc: Some(_), .. } => {
+                    EventKind::ContractSetupCompleted { dlc: None }
+                }
+                EventKind::RolloverCompleted {
+                    dlc: Some(_),
+                    funding_fee,
+                } => EventKind::RolloverCompleted {
+                    dlc: None,
```
bonomat: I'm not sure this is safe: we need the old `revocation` and `publish` keys to be able to punish.

bonomat (commented):
In general I think it's the right way, but I think we need some of those values to be able to punish malicious behavior.

daemon/src/db/cull_dlcs.rs
luckysori: This should all be stored here:

https://github.com/itchysats/itchysats/blob/0a7b03d39dd5c9d6e483bde31d4fae14163a4dd3/model/src/cfd.rs#L1811

https://github.com/itchysats/itchysats/blob/0a7b03d39dd5c9d6e483bde31d4fae14163a4dd3/model/src/cfd.rs#L2000-L2015

https://github.com/itchysats/itchysats/blob/0a7b03d39dd5c9d6e483bde31d4fae14163a4dd3/daemon/src/setup_contract.rs#L637-L644

daemon/src/db/cull_dlcs.rs
luckysori: I will try without `LIMIT -1`, but I read that it was needed for SQLite.

luckysori: I get this error as soon as I remove the `LIMIT -1`:

```
error: error returned from database: near "OFFSET": syntax error
  --> daemon/src/db/cull_dlcs.rs:10:20
   |
10 |           let rows = sqlx::query!(
   |  ____________________^
11 | |             r#"
12 | | SELECT
13 | |     id,
...  |
26 | |             model::EventKind::ROLLOVER_COMPLETED_EVENT,
27 | |         )
   | |_________^
   |
```

Source for the `LIMIT -1` solution: https://stackoverflow.com/a/19676495.

bonomat: You'd need to remove the `offset` as well which should give you the whole table. 

I.e. just this: 

```
SELECT
    id,
    name,
    data
FROM
    events
WHERE
    name = $1 OR
    name = $2
ORDER BY
    created_at DESC
```

daemon/src/db/cull_dlcs.rs
bonomat: Ah, thanks for looking this up.

daemon/src/db/cull_dlcs.rs
luckysori: But the point was to omit the newest rollover/contract-setup event, because we don't want to cull that one.

I can do that outside of course. Originally it mattered more to do it in the query, because I was attempting to modify the `data` field inside of the query. It could not work because the data field is a JSON blob so it seems impossible to tweak a particular field in there.

bonomat: Oh, so sorry then. I did not understand your intention.

daemon/src/db/cull_dlcs.rs
da-kami: @luckysori does that mean we actually stored this information redundantly in the past? 
Comes in handy now 😅

daemon/src/db/cull_dlcs.rs
```diff
@@ -0,0 +1,187 @@
+use crate::db::Connection;
+use anyhow::Result;
+use model::EventKind;
+
+impl Connection {
+    #[allow(dead_code)]
+    async fn cull_old_dlcs(&self) -> Result<()> {
```
da-kami: 🙃 I see that you cull all DLCs but the last one. Can we change this to keeping the last two until rollover is more stable?

daemon/src/db/cull_dlcs.rs
```diff
@@ -0,0 +1,187 @@
...
+    use model::Role;
+    use model::Timestamp;
+    use model::TxFeeRate;
+    use model::Usd;
+    use rust_decimal_macros::dec;
+    use sqlx::Connection;
+    use time::Duration;
+
+    #[tokio::test]
+    async fn given_two_rollovers_when_cull_old_dlcs_only_last_one_has_some_dlc() {
```
da-kami: 🙃 I would add another test for the case where we have only one rollover and cull to ensure that this rollover is not removed.

Additionally I'd add one test with three rollovers to ensure that we remove the first two rollovers - but this one is not as important.

da-kami (commented):
I like the simplicity of this solution as a first step! 
We can work on replacing the json with a properly normalized schema in follow ups.

da-kami:
> This PR introduces a breaking change to the events we store in the database. We therefore will need to leverage the fact that we can now make breaking changes to this layer provided we only have closed or failed CFDs.

Why is this a breaking change?
I feel if we just cull the data it should not be a breaking change. The Aggregate would only set the `Dlc` with the last `Rollover`, but that should not be breaking I think...?

daemon/src/db/cull_dlcs.rs
bonomat: Why are the last two important?

daemon/src/db/cull_dlcs.rs
luckysori: Yes, we had multiple copies of this, growing after each rollover.

daemon/src/db/cull_dlcs.rs
luckysori: I don't know why we would want to save anything other than the last one. What are you thinking about?

daemon/src/db/cull_dlcs.rs
luckysori: Yeah, I will add a few more tests!

I just wanted to prove that it seems to work before continuing.

luckysori:
> Why is this a breaking change? I feel if we just cull the data it should not be a breaking change. The Aggregate would only set the `Dlc` with the last `Rollover`, but that should not be breaking I think...?

The reason why I thought this PR introduced a breaking change to the database was because:

https://github.com/itchysats/itchysats/blob/78c6ee1c272e47390465102319fd5c230f3d2158/model/src/cfd.rs#L426-L428

https://github.com/itchysats/itchysats/blob/78c6ee1c272e47390465102319fd5c230f3d2158/model/src/cfd.rs#L436-L439

But I think you are right. Making those fields optional should not affect deserialisation of old events which always have them present! If this is correct then we don't even have to wait for `cfds`/`events` to be empty before we roll out this change!

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -0,0 +1,308 @@
+use crate::db::Connection;
+use anyhow::Result;
+use model::EventKind;
+use sqlx::pool::PoolConnection;
+use sqlx::Sqlite;
+
+impl Connection {
+    #[allow(dead_code)]
```
bonomat: I assume this is just temporary and will be removed in a later commit. :)

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        Ok(())
+    }
+}
+
+/// Load all old events with DLC data from the database and return
+/// them with empty `dlc` fields.
+///
+/// `EventKind::ContractSetupCompleted` an
+/// `EventKind::RolloverCompleted` are the two kinds of event tha can
```
bonomat:
```suggestion
/// `EventKind::RolloverCompleted` are the two kinds of event that can
```

maker/src/actor_system.rs
```diff
@@ -179,7 +181,9 @@ where
             .create(None)
             .spawn(&mut tasks);
 
-        tasks.add(time_to_first_position_ctx.run(time_to_first_position::Actor::new(db)));
+        tasks.add(time_to_first_position_ctx.run(time_to_first_position::Actor::new(db.clone())));
+
+        let _cull_old_dlcs_actor = cull_old_dlcs::Actor::new(db).create(None).spawn(&mut tasks);
```
bonomat, line 188: I wonder if this will make troubles as it's being started at the same time as other actors who are reading from the db. 

What do you think of running `cull` once before we initialize anything else?

daemon/src/db/cull_old_dlcs.rs
luckysori: Good catch. You're right. I'm gonna undo this because it doesn't achieve much (there are other warnings after my recent refactor anyway).

bonomat (approved):
Good work. I think it would be good to run `cull_old_dlcs` once before we start up anything else to (a) not interfere with other loading (b) profit from the emptied db already during the first startup

maker/src/actor_system.rs
luckysori: > I wonder if this will make troubles as it's being started at the same time as other actors who are reading from the db.

I don't see potential for problems: we are just deleting irrelevant data that we never access otherwise. And I expect `sqlx` and `sqlite` to let us modify rows safely.

> What do you think of running cull once before we initialize anything else?

I see the appeal, but I'm not sure how to do it in a non-hacky way (i.e. without a long sleep before we do anything else in the actor system constructors).

Also we have to consider that after a few restarts most of the old DLC-related events will be culled already. It's only the very first time that we introduce this change that we may benefit significantly from running it first in isolation.

bonomat: We could just call it directly and await the result? 
As you said, the main benefit is for the first start, sub-sequential starts don't matter much which also means, that this call will be fast.

luckysori: > We could just call it directly and await the result? As you said, the main benefit is for the first start, sub-sequential starts don't matter much which also means, that this call will be fast.

Good idea! I've added https://github.com/itchysats/itchysats/pull/2041/commits/548cb5ff824ac0522676c27d86078fa18a8182ca.

bonomat (approved):
Awesome, looks really good!

daemon/src/db/cull_dlcs.rs
da-kami: Given that some logic around revoke commits is still missing it might be useful to have the old DLC so we can "cleanup"  CFDs by going one back in history. 🤷‍♂️
Once we cull and only keep latest we can't use this "cleanup" anymore but would have to just remove the "broken" CFDs. I don't mind much. Feel free to just roll with one. Just wanted to raise awareness that at some point we wanted to transition "broken" CFDs into closed by removing the last rollover and then restart.

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -0,0 +1,307 @@
...
+        ));
+
+        assert!(matches!(
+            &events[2].event,
+            EventKind::RolloverCompleted { dlc: None, .. },
+        ));
+
+        assert!(matches!(
+            &events[3].event,
+            EventKind::RolloverCompleted { dlc: Some(_), .. },
```
da-kami: 🙃 I think it would be better to compare against the actual expected DLC - or at least something that makes sure that this is the DLC we expect (as in the latest one) and not e.g. the one of contract setup. Given the logic this should always work out, but still.

da-kami (approved):
Let's test it on testnet and with the prod db :)

daemon/src/db/cull_dlcs.rs
luckysori: Thanks for raising this!

It's true that we had that idea, but we haven't actually used it despite having a few out-of-sync CFDs. I don't think it's a critical feature (it's only a workaround for a bug) so I wouldn't adapt this new feature for it at this stage.

We can always reconsider before we release this new functionality.

Issue #1769, opened 2022-04-05
Design proposal for `/itchysats/identify` protocol


thomaseizinger:
The proposed design is loosely based on https://github.com/libp2p/specs/tree/master/identify.

Core design choices:

- Only support _pulling_ identify information: Identify information is primarily useful for the peer that receives it. We don't want to push information to peers unless we know they are interested in it.
- Identify information should include:
	- agent-version
	- list of supported protocols
- Use JSON for encoding messages. We can switch a binary format in a different version.

The agent-version is useful for monitoring purposes.

Retrieving the list of supported protocol will allow us to make a runtime decision on which protocols to use to communicate with a peer. For example, a peer can signal that they want to have offers pushed (see https://github.com/itchysats/itchysats/issues/1768) by advertising the `/itchysats/offer` protocol.

Currently, `libp2p-xtra` has a static configuration of which protocols are supported by an `Endpoint`. However, one can imagine making this list dynamic and thus, advertising different protocols to different peers. This will allow for a form of A/B testing where we can offer a specific protocol to a few select peers and/or to a subset (i.e. 10%).

For example, a maker could advertise that it supports a new rollover protocol to select peers, thus allowing for a way of safely testing things in production without rolling them out to all takers at once.

## Open questions

- Use protocol specified in https://github.com/libp2p/specs/tree/master/identify instead? It will require the use of protobuf for the message encoding.

luckysori:
> Use protocol specified in https://github.com/libp2p/specs/tree/master/identify instead? It will require the use of protobuf for the message encoding.

Unless this would speed us up considerably, I wouldn't bother. Is the main advantage compatibility? I don't think we need to worry about it for a long time.

> Only support pulling identify information: Identify information is primarily useful for the peer that receives it. We don't want to push information to peers unless we know they are interested in it.

Does this mean that peers should periodically poll each other for this information? It seems acceptable for a taker to periodically poll the single maker they're connected to. But a maker could be connected to many takers. How do you see this working?

thomaseizinger:
> > Use protocol specified in [https://github.com/libp2p/specs/tree/master/identify](https://github.com/libp2p/specs/tree/master/identify?rgh-link-date=2022-04-07T00%3A47%3A08Z) instead? It will require the use of protobuf for the message encoding.
> 
> Unless this would speed us up considerably, I wouldn't bother. Is the main advantage compatibility? I don't think we need to worry about it for a long time.

[...]

PR #2042, opened 2022-05-16
Store `PeerId` for `Cfd` in the database

### LINKED ISSUE #1993: Store counterparty peer id into the DB
We will need to store this information to be able to successfully finish some protocols, such as collab settlement or rollover.

in case a peer id was not stored, we can fall back to the mainnet/testnet/dev-maker/dev-taker peer id with a simple lookup by counterparty network identity (ie. the old id).

da-kami:
fixes https://github.com/itchysats/itchysats/issues/1993

In the model the `PeerId` is optional because we cannot always predict it.
For the maker it is impossible to map old taker identities to peer-ids.
For the taker we can only map known maker-ids to peer-ids.

In the database we use a `placeholder` `PeerId` to indicate if the peer-id of the counterparty is known.
This is allows to add the `NOT NULL` constraint to the new `counterparty_libp2p_identity` column.
When loading the `counterparty_libp2p_identity` from the database for the taker, if the `placeholder` is loaded we default known maker peer-ids based on a mapping to from the legacy network identity.
This allows new takers to use the new rollover protocol for positions that were opened with a previous taker version.

We introduce `HelloV3` message.
New takers connects to the maker with `HelloV3`.
The maker treats the taker's peer-id as optional field when passing on incoming messages; making the maker backwards-compatible.


daemon/src/connection.rs
```diff
@@ -356,9 +360,10 @@ impl Actor {
 
         let proposed_version = Version::LATEST;
         write
-            .send(wire::TakerToMaker::HelloV2 {
+            .send(wire::TakerToMaker::HelloV3 {
```
luckysori, line 363: ❓ Why is this new message about sharing one's `libp2p` peer ID not sent via `libp2p`?

daemon/src/db.rs
```diff
@@ -39,6 +42,15 @@ mod failed;
...
 pub mod time_to_first_position;
 
+const MAINNET_MAKER_ID: &str = "7e35e34801e766a6a29ecb9e22810ea4e3476c2b37bf75882edf94a68b1d9607";
+const MAINNET_MAKER_PEER_ID: &str = "12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY";
+
+const TESTNET_MAKER_ID: &str = "69a42aa90da8b065b9532b62bff940a3ba07dbbb11d4482c7db83a7e049a9f1e";
+const TESTNET_MAKER_PEER_ID: &str = "12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1";
+
+const DEV_MAKER_ID: &str = "10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e";
+const DEV_MAKER_PEER_ID: &str = "12D3KooWDjzHna3pNi1Bt1DoRfrpsBREykJKXDRDxXvhJNAdDZEk";
```
luckysori: ❓ I guess we don't want to merge this?

daemon/migrations/20220516000538_libp2p_identity.sql
```diff
@@ -0,0 +1,6 @@
+-- The default value is a dummy value that indicates that this is an "old" CFD
```
klochowicz: perhaps add a comment how you obtained this value, otherwise it's pretty random "dummy" value ;D

daemon/src/db.rs
```diff
@@ -160,6 +173,14 @@ impl Connection {
         .bind(&cfd.settlement_time_interval_hours().whole_hours())
         .bind(&cfd.quantity())
         .bind(&cfd.counterparty_network_identity())
+        .bind(&cfd.counterparty_libp2p_identity().unwrap_or({
+            tracing::debug!(
+                order_id=%cfd.id(),
+                taker_id=%cfd.counterparty_network_identity(),
+                "Inserting deprecated CFD with placeholder peer-id!"
+            );
```
luckysori: ❓ Is this expected to happen only once per "old" CFD?

model/src/cfd.rs
```diff
@@ -616,6 +617,7 @@ pub struct Cfd {
     settlement_interval: Duration,
     quantity: Usd,
     counterparty_network_identity: Identity,
+    counterparty_libp2p_identity: Option<Libp2pIdentity>,
```
luckysori: ❓ So as soon as all CFDs have an associated counterparty peer ID we can remove the `Option`, right?

daemon/src/db.rs
```diff
@@ -508,15 +531,40 @@ async fn load_cfd_row(conn: &mut Transaction<'_, Sqlite>, id: OrderId) -> Result
...
+                    DEV_MAKER_ID => Some(DEV_MAKER_PEER_ID),
+                    _ => None,
+                })
+                .map(|peer_id| PeerId::from_str(peer_id).expect("static peer-id to parse"))
+                .map(Libp2pIdentity::from),
+                Role::Maker => None,
+            }
+        } else {
+            Some(cfd_row.counterparty_libp2p_identity)
+        };
```
klochowicz: it would be cleaner to extract this as a free function that taker the row and returns peer_id - then it's easier to remove later, and the consts can be in the scope of the function instead of global.

daemon/src/db.rs
```diff
@@ -160,6 +173,14 @@ impl Connection {
         .bind(&cfd.settlement_time_interval_hours().whole_hours())
         .bind(&cfd.quantity())
         .bind(&cfd.counterparty_network_identity())
+        .bind(&cfd.counterparty_libp2p_identity().unwrap_or({
+            tracing::debug!(
+                order_id=%cfd.id(),
+                taker_id=%cfd.counterparty_network_identity(),
+                "Inserting deprecated CFD with placeholder peer-id!"
```
luckysori: 🙃 The exclamation mark makes me think this is very interesting, but I think this is pretty boring.

```suggestion
                "Inserting deprecated CFD with placeholder peer-id"
```

model/src/lib.rs
```diff
@@ -529,6 +530,55 @@ impl str::FromStr for Identity {
 
 impl_sqlx_type_display_from_str!(Identity);
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Libp2pIdentity(PeerId);
```
klochowicz: Perhaps we could still use the name `PeerId`? ("Libp2pIdentity" is a bit of a mouthful for a common type....)
`peer_identity` in the worst case? :)

I like the newtype though!

model/src/cfd.rs
luckysori: ❓ Does this really need to be optional? You've already set this field to a default value in the database. If we load a CFD from the database, won't it already have this field included? I don't see when we can get a `None` here.

daemon/src/db.rs
```diff
@@ -508,15 +531,40 @@ async fn load_cfd_row(conn: &mut Transaction<'_, Sqlite>, id: OrderId) -> Result
...
+    // The peer-id is defaulted to a placeholder dummy in the database. Based on the dummy we can
+    // decide to default the peer-id according to known makers.
+    let counterparty_libp2p_identity =
+        if cfd_row.counterparty_libp2p_identity == Libp2pIdentity::placeholder() {
+            match role {
+                // Default known maker PeerIds based on the legacy network identity
+                Role::Taker => (match counterparty_network_identity.to_string().as_str() {
+                    MAINNET_MAKER_ID => Some(MAINNET_MAKER_PEER_ID),
+                    TESTNET_MAKER_ID => Some(TESTNET_MAKER_PEER_ID),
+                    DEV_MAKER_ID => Some(DEV_MAKER_PEER_ID),
```
luckysori: 🤔 We have the network available every time we call `load_cfd_row`. We could pass it in and use it here instead of these constants.

daemon/src/db.rs
```diff
@@ -160,6 +173,14 @@ impl Connection {
         .bind(&cfd.settlement_time_interval_hours().whole_hours())
         .bind(&cfd.quantity())
         .bind(&cfd.counterparty_network_identity())
+        .bind(&cfd.counterparty_libp2p_identity().unwrap_or({
```
luckysori: ❓ Could you insert the correct value here (using a similar `match` to the one you use below) if you detect a `Libp2pIdentity::placeholder()`?

daemon/src/db.rs
luckysori: Incorrect, we don't have that information when we archive closed/failed CFDs :(

model/src/lib.rs
```diff
@@ -529,6 +530,55 @@ impl str::FromStr for Identity {
...
+    pub fn peer_id(&self) -> PeerId {
+        self.0
+    }
+
+    // TODO: Change this to be const
+    /// Placeholder used for db entries prior to using libp2p
+    pub fn placeholder() -> Libp2pIdentity {
+        // peer id for key filled up with 0s
+        Self(
+            "12D3KooWE1krEN7utXUFhLmLXLWPSuah5WQ7ChzUFetkMS7TRXsm"
```
luckysori: 🔧 Would be good if you could tie this value to the one you use in the test in `seed.rs`.

luckysori (commented):
Just some thoughts!

daemon/src/connection.rs
da-kami: The problem is knowing the `PeerId` in a `Cfd` that was *not* opened with the libp2p connection. 
For the moment both connections exist and we open a `Cfd` over the legacy connection, but roll over with libp2p. So, when creating the `Cfd` we have to associate it with the `PeerId` as well. We don't have to do this using the `Hello` message, could us `ContractSetup` instead but it was somewhat convenient to do it like this - and we will have it in every message which might come in handy.

daemon/src/db.rs
da-kami: Why not? 
We don't have to but all of these are temporary until we remove the old connection code. Since we know that one we we thought about adding it.

daemon/src/db.rs
da-kami: This is only expected to happen if a taker on a previous version opens a Cfd with a maker on the new version.
Unfortunately we cannot really distinguish because taker and maker share the same db and model.

model/src/cfd.rs
da-kami: Originally I wanted to just use the default, but if we ever log it in a context that is *not* the new protocol it will be weird. 
We could still accept to always set it and be careful when to use this field for logging.

daemon/src/db.rs
da-kami: yep. I really did not like duplicating this, but bare in mind that all of this is temporary. Once the legacy code is removed all of this will go away. I think for rolling libp2p out ("giving rollover a try") it is acceptable to do this.

daemon/src/db.rs
da-kami: I would still need it when loading, because previously opened positions should also be treated with the new protocol on the taker. Given that I just did it on loading. I don't want to update existing CFDs. 

I could even do it in the migration statement using a `CASE`. Maybe doing it as part of the migration is actually better. The current solution is somewhat easier, because I don't try to be smart when migrating, but only when loading and this is only temporary.

model/src/lib.rs
da-kami: I know, but it's hard because we moved the code to different `crates`. I don't really want to expose the `PlaceholderSeed` - I just added the test to showcase that the placeholder is derived deterministically. Could also remove it. The test does not prove anything.

daemon/src/connection.rs
da-kami: Theoretically we could also say that the maker actually does not need to safe the `PeerId` because the taker always starts the protocol with the maker and we know the Cfd's id. But like this we work more towards the future, because eventually both parties will store this information.
For the maker mostly for loggin purposes and validation. For the taker to communicate with the correct maker.

daemon/src/db.rs
klochowicz: It would be good if this lookup was contained inside a function, so that we don't need to expose these constants globally (so it's obvious why they're there) - see other comment. If we won't be able to use the const for the tests, it might be enough to just ensure it's not `None` when a lookup works. 
We'll test all 3 combinations anyway.

model/src/cfd.rs
klochowicz: IMHO it's much cleaner to have it as Optional. in case you upgrade a maker with a CFD open, this field will not be set (and the lookup will fail, as we can't create a mapping on that side). I think it's much cleaner to encode such things with an Option than a magic value (in the DB it makes a bit more sense to have a default if we convert it to an Option when reading; the added value is preventing setting it to `None`, all new CFDs ought to have that field)

model/src/lib.rs
```diff
@@ -1477,6 +1527,25 @@ mod tests {
         assert!(relative.is_positive())
     }
 
+    #[test]
+    fn libp2p_placeholder_snapshot() {
+        let placeholder = Libp2pIdentity::placeholder().peer_id();
+        assert_eq!(
+            placeholder.to_string(),
+            "12D3KooWE1krEN7utXUFhLmLXLWPSuah5WQ7ChzUFetkMS7TRXsm"
```
klochowicz: I'd extract this a public const - `NO_PEER_ID` or something.

klochowicz (approved):
nice one!
concept ACK, hopefully we'll get it in tomorrow!

daemon/src/db.rs
```diff
@@ -39,6 +42,15 @@ mod failed;
 
 pub mod time_to_first_position;
 
+const MAINNET_MAKER_ID: &str = "7e35e34801e766a6a29ecb9e22810ea4e3476c2b37bf75882edf94a68b1d9607";
+const MAINNET_MAKER_PEER_ID: &str = "12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY";
+
+const TESTNET_MAKER_ID: &str = "69a42aa90da8b065b9532b62bff940a3ba07dbbb11d4482c7db83a7e049a9f1e";
+const TESTNET_MAKER_PEER_ID: &str = "12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1";
```
bonomat: ❓These values already exist in `taker::main.rs`. Why did you add them here again?
 
https://github.com/itchysats/itchysats/blob/6dff7129cc13d6230fd09e4e569bf2ba425d6044/taker/src/main.rs#L45-L53

daemon/src/db.rs
da-kami: Happy to do so, but then I have to hardcode the values in the tests instead of using the constants.

daemon/src/seed.rs
```diff
@@ -157,3 +157,30 @@ impl From<[u8; 32]> for UmbrelSeed {
...
+
+        let placeholder_seed = PlaceholderSeed();
+        let placeholder_peer_id = placeholder_seed.derive_identities().peer_id();
+
+        let expected_placeholder = "12D3KooWE1krEN7utXUFhLmLXLWPSuah5WQ7ChzUFetkMS7TRXsm"
+            .parse::<PeerId>()
+            .unwrap();
+
+        assert_eq!(expected_placeholder, placeholder_peer_id)
+    }
```
bonomat: I don't really understand the need for this test. 
Maybe you can add a constant and add a doc-test to show how you derived the constant. That way you can use the constant in different places. You can also annotate the constant with `cfg(test)` if it is only needed for tests.

model/src/lib.rs
bonomat: You could move it into a module `libp2p` and call it `PeerId`.

model/src/lib.rs
bonomat: I've added a comment above: I'd extract it as a constant and add a doc test which shows that this value is derived from `0000..`

bonomat (approved):
My understanding of this change is that this is only needed because `TakeOrder` comes over the old connection where we don't have the libp2p details available but it is needed for rollover which will be on the new connection. 

Couldn't we achieve the same with just defaulting to either the testnet/mainnet maker peer id and leaving the taker peer id empty (or dummy) for the time being. 
Eventually we will move `TakeOrder` to a libp2p protocol and will have this value available. Then we can drop the default values. 

I'm not sure if my line of thinking is fully thought through, and I don't want to block you neither, so please move on :)

daemon/src/seed.rs
da-kami: We don't need this test, but I wanted to showcase where the constant comes from. I will remove it and document it as part of the const.
The problem is that in `model` I can't access the seed from `daemon` unless I add a dependency. I don't think we should add cross-dependencies between these crates.

daemon/src/connection.rs
luckysori: I appreciate all the explanations and context, but I'm still wondering about my original question:

>  Why is this new message about sharing one's `libp2p` peer ID not sent via `libp2p`?

Only takers who upgrade to the new version including these changes will be able to tell the maker about their `libp2p` peer ID. Why extend the legacy networking stack for the purpose of sharing this information if we can define a protocol via `libp2p` for the same purpose?

daemon/src/db.rs
luckysori: The ones I was wondering about specifically were the ones with the `DEV_` prefix. What are they about?

model/src/cfd.rs
luckysori: So you do think this can be `None` then?

daemon/src/connection.rs
da-kami: We *can* use the `identify` protocol for knowing the peer-id, but the question how to associate this `peer-id` with a `Cfd` that was opened on the legacy connection remains. When you open the `Cfd` you don't not know the peer-id unless we know it over the old connection or are able to "look it up".

As an alternative we could somehow store a mapping from legacy identity to peer-id in the database, or we could have some kind of task that updates `Cfd`s based on connections being opened over libp2p2. I feel that these solutions are even more intrusive.

da-kami:
> [snip]

> Couldn't we achieve the same with just defaulting to either the testnet/mainnet maker peer id and leaving the taker peer id empty (or dummy) for the time being. Eventually we will move `TakeOrder` to a libp2p protocol and will have this value available. Then we can drop the default values.

Basically yes, but it might be hard to test it then, and it might be hard to get proper stats from the maker logs.
I bounced several solutions in my head when implementing it and all have up and downsides. I think this solution brings us closest to what it should be in the future, because all newly added `Cfd` know both identities and at the same time the taker is able to rollover previously opened positions with the new protocol.

daemon/src/connection.rs
luckysori: > We can use the identify protocol for knowing the peer-id, but the question how to associate this peer-id with a Cfd that was opened on the legacy connection remains. When you open the Cfd you don't not know the peer-id unless we know it over the old connection or are able to "look it up".

This comment on the `identify` protocol proposal talks about fetching the peer ID for a particular taker ID: https://github.com/itchysats/itchysats/issues/1769#issuecomment-1092515313.

But I understand that the fastest way forward is to do something like what you have done here!

luckysori (approved):
LGTM!

klochowicz:
> My understanding of this change is that this is only needed because `TakeOrder` comes over the old connection where we don't have the libp2p details available but it is needed for rollover which will be on the new connection.
> 
> Couldn't we achieve the same with just defaulting to either the testnet/mainnet maker peer id and leaving the taker peer id empty (or dummy) for the time being. Eventually we will move `TakeOrder` to a libp2p protocol and will have this value available. Then we can drop the default values.
> 
> I'm not sure if my line of thinking is fully thought through, and I don't want to block you neither, so please move on :)

unfortunately we need to pull it through to make existing actor tests pass (they start by taking order) and running makers locally (including dev-maker)....

daemon/src/db.rs
da-kami: I removed them because I actually think what @klochowicz and me had in mind does not really make sense. 
If we open a new CFD using our shared dev seed then that new CFD will have a peer-id anyway. I don't think we have to support rolling over old `Cfd` with the new protocol in the dev environment.

daemon/src/db.rs
da-kami: Note on:

> I could even do it in the migration statement using a CASE

No we can't. I gave this a try but it appears this is not possible in Sqlite.

model/src/lib.rs
da-kami: I did that which caused a bit of refactoring. We use `peer_id` throughout the code base now, except when deriving the identity using the prefix `LIBP2P_IDENTITY`. I can't change that because it would change our identity.

model/src/lib.rs
da-kami: > I've added a comment above: I'd extract it as a constant and add a doc test which shows that this value is derived from 0000..

I cannot easily add the test because `model` does not know about `deamon` crate where the `seed` is located. I could move the `seed` to `model` but I am not sure that is correct. For now I just removed the test.

da-kami:
Handling failed and closed cfds was missing. I added another commit.

luckysori (approved):
Looks good!

Looking at it again, I have a question. Did you consider _not_ letting old CFDs run `libp2p`-based protocols? I wonder if that would make things simpler.

model/src/cfd.rs
klochowicz: yes, it can be `None`, in case you upgrade with an open CFD and you can't look up the counterparty id (which will be the case on the maker).

klochowicz (approved):
LGTM!

da-kami:
bors ping

da-kami:
bors retry

Issue #2002, opened 2022-05-10
Attestation fetching stopped


bonomat:
According to our logs it looks like we stopped fetching attestations between 2022-05-10T08:01:13.144388844Z and 2022-05-10T22:50:21.301540243Z(after a restart).

bonomat:
Appearances of the log message `Fetching attestation` per hour over the last 2 days: 

![image](https://user-images.githubusercontent.com/224613/167741401-c821ed26-49f4-4eac-ad45-067478eb9fba.png)

bonomat:
# Analysis

My interpretation of this issue is that this is not an issue: 

- Each DLC has a `settlement_event_id` which is the event id for when this CFD will expire unless it is rolled over
- During startup we load all CFDs from the DB and add their `settlement_event_id` to a list called `pending_attestation`: 

https://github.com/itchysats/itchysats/blob/b29277675c5fbb89c3f3c24369de35b62b404e4e/daemon/src/oracle.rs#L308-L342

https://github.com/itchysats/itchysats/blob/b29277675c5fbb89c3f3c24369de35b62b404e4e/daemon/src/oracle.rs#L234-L240

- on a regular interval we check if this attestation has _likely_ occurred yet and will try to fetch it from the oracle: 

https://github.com/itchysats/itchysats/blob/b29277675c5fbb89c3f3c24369de35b62b404e4e/daemon/src/oracle.rs#L185-L189

- on every rollover or contract completion, we also add an event id into this list, by sending `MonitorAttestation` to the oracle actor, e.g. here: 

https://github.com/itchysats/itchysats/blob/d6f8c9955631c18f2a31920c39de350049e486a7/daemon/src/process_manager.rs#L95-L99

and here: 

https://github.com/itchysats/itchysats/blob/d6f8c9955631c18f2a31920c39de350049e486a7/daemon/src/process_manager.rs#L198-L202


# Conclusion
- This means that when starting up and with no down time, we likely get a list of events to monitor 24h from now
- On every rollover we add a new id to this list, so that it can be fetched in the future, i.e. one in 24+1h from now
- Because our maker (testnet and production) gets regularly restarted, this list mostly contains events for 24h in the future and none for the next hour
- This means, we won't be fetching any attestation for the next 24h. 
- However, if our application runs for over 24h, we expect to be fetching for new attestations every 1h. 

Long story short, this does not seem to be an issue.

PR #2049, opened 2022-05-17
Small oracle changes

bonomat:
A few minor changes I made while trying to understand the bug in #2002 .

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -91,7 +91,7 @@ impl<O, W, P> TakerActorSystem<O, W, P>
 where
     O: Handler<oracle::MonitorAttestation>
         + Handler<oracle::GetAnnouncement>
-        + Handler<oracle::Sync>
+        + Handler<oracle::SyncAnnouncements>
         + Actor<Stop = ()>,
     W: Handler<wallet::BuildPartyParams>
         + Handler<wallet::Sign>
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -89,10 +89,7 @@ pub struct TakerActorSystem<O, W, P> {
 
 impl<O, W, P> TakerActorSystem<O, W, P>
 where
-    O: Handler<oracle::MonitorAttestation>
-        + Handler<oracle::GetAnnouncement>
-        + Handler<oracle::SyncAnnouncements>
-        + Actor<Stop = ()>,
+    O: Handler<oracle::MonitorAttestation> + Handler<oracle::GetAnnouncement> + Actor<Stop = ()>,
     W: Handler<wallet::BuildPartyParams>
         + Handler<wallet::Sign>
         + Handler<wallet::Withdraw>
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -31,6 +31,9 @@ const REQWEST_TIMEOUT: core::time::Duration = core::time::Duration::from_secs(10
 /// Syncing every 60 seconds might still be an overkill but should not hurt us.
 const SYNC_ANNOUNCEMENTS_INTERVAL: core::time::Duration = std::time::Duration::from_secs(60);
 
+/// We want to sync attestations fast but don't spam our internal actor. Hence, we chose 30 seconds.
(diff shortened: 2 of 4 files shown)


daemon/src/oracle.rs
```diff
@@ -199,6 +199,7 @@ impl Actor {
 
                     let response = client
                         .get(url.clone())
+                        .timeout(core::time::Duration::from_secs(10))
```
luckysori: ❓ Could it hang forever before?

luckysori: 🔧 Maybe extract the `Duration` into a constant so that we can explain the reason, even if it's fairly arbitrary.

daemon/src/oracle.rs
```diff
@@ -310,7 +310,7 @@ impl xtra::Actor for Actor {
         let this = ctx.address().expect("we are alive");
         self.tasks.add(
             this.clone()
-                .send_interval(std::time::Duration::from_secs(5), || Sync),
+                .send_interval(std::time::Duration::from_secs(30), || Sync),
```
luckysori: 🤔 Should we have two intervals?

We know that we only get one attestation per hour, but it is very important to get attestations as soon as possible. We could optimise this further but I think it's acceptable to check for a new attestation every 30 seconds.

But the announcements are updated at a completely different rate. We look 25 hours in the future, but we could look months ahead (e.g. https://outcome.observer/h00.ooo/x/BitMEX/BXBT/2022-07-20T01:00:00.price?n=20).

I would recommend to look 48 hours ahead and only run `ensure_having_announcements` once per hour. Would leave a lot of space for further optimisation, but the gains would be small given that we would already be doing

https://github.com/itchysats/itchysats/blob/5e6d8e63c1636eb6d7de725b6c489622e3da466b/daemon/src/oracle.rs#L136-L141

48 times per hour, which is pretty inexpensive.

daemon/src/lib.rs
```diff
@@ -90,7 +90,7 @@ impl<O, W, P> TakerActorSystem<O, W, P>
 where
     O: Handler<oracle::MonitorAttestation>
         + Handler<oracle::GetAnnouncement>
-        + Handler<oracle::Sync>
+        + Handler<oracle::SyncAnnouncements>
```
luckysori: 🔧 You can remove this altogether (in a separate commit? :). `oracle::Sync` should be a module-private type, but we made it public to use it in `daemon-tests`. The `TakerActorSystem` doesn't need to know about the internal messages `oracle::Actor` sends to itself (verified locally).

maker/src/actor_system.rs
```diff
@@ -55,7 +55,7 @@ impl<O, W> ActorSystem<O, W>
 where
     O: Handler<oracle::MonitorAttestation>
         + Handler<oracle::GetAnnouncement>
-        + Handler<oracle::Sync>
+        + Handler<oracle::SyncAnnouncements>
```
luckysori: Same thing I mentioned for `TakerActorSystem` applies here (also tested).

luckysori (approved):
I left a comment before seeing your last commit 😛 I like the changes, although I would tweak the timing of the internal messages.

daemon/src/oracle.rs
bonomat: > ❓ Could it hang forever before?

My understanding is yes and I couldn't find any evidence that it wouldn't in the reqwest codebase.

daemon/src/oracle.rs
bonomat: I think it is acceptable to check every 30 seconds for now. We can reduce this to check even less often but I'm not sure how much would gain from this.

luckysori: Sure, we don't have to go to 60 minutes now.

But why 30 seconds for announcements and 60 seconds for attestations? It would make more sense the other way around for me.

bonomat: Fair point. what timing would you chose?

luckysori: Maybe:

- Attestations every 30 seconds.
- Announcements every minute.

bonomat: works for me :)

PR #2052, opened 2022-05-17
Define separate `oracle::MonitorAttestations` for start-up

luckysori:
It should be more efficient to send one message instructing the actor to monitor several attestations when initialising the `oracle::Actor`, rather than multiple smaller messages.

Inspired by looking at #2049.

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -247,16 +247,18 @@ impl Actor {
             )
         }
     }
+
+    fn add_pending_attestation(&mut self, event_id: BitMexPriceEventId) {
+        if !self.pending_attestations.insert(event_id) {
+            tracing::trace!("Attestation for {event_id} already being monitored");
+        }
+    }
 }
 
 #[xtra_productivity]
 impl Actor {
     fn handle_monitor_attestation(&mut self, msg: MonitorAttestation) {
-        let price_event_id = msg.event_id;
-
-        if !self.pending_attestations.insert(price_event_id) {
-            tracing::trace!("Attestation {price_event_id} already being monitored");
-        }
+        self.add_pending_attestation(msg.event_id)
     }
 
     fn handle_get_announcement(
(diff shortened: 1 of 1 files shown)


daemon/src/oracle.rs
```diff
@@ -38,6 +38,11 @@ pub struct MonitorAttestation {
     pub event_id: BitMexPriceEventId,
```
bonomat, line 55: Do you think it makes sense to get rid of `MonitorAttestation` in the favor of `MonitorAttestations` and just send around a vec with 1 element?

bonomat (approved):
Nice improvement.

da-kami (approved):

daemon/src/oracle.rs
luckysori: I don't think it's a good idea for two reasons:

- The public API of this module doesn't need to let us send a message to monitor more than one attestation at a time.
- We would be allocating unnecessarily when sending `MonitorAttestations` with one item.

I think keeping both messages is worth it.

luckysori:
Had to modify it slightly after rebasing.

daemon/src/oracle.rs
```diff
@@ -362,11 +362,13 @@ impl xtra::Actor for Actor {
...
-                this.clone()
-                    .send_interval(SYNC_ATTESTATIONS_INTERVAL, || SyncAttestations)
-                    .await;
             }
         });
+
+        self.tasks.add(async {
+            this.send_interval(SYNC_ATTESTATIONS_INTERVAL, || SyncAttestations)
+                .await;
+        })
```
bonomat: This was on purpose in the same task to only start it after we fetched everything from the db.

bonomat (approved):
One comment. I'm not sure if it is a big difference if you keep it where it was.

daemon/src/oracle.rs
luckysori: Ooops, my bad. I thought there was no point and it made rebasing easier. I'll undo it!

PR #2053, opened 2022-05-17
Steal connection

### LINKED ISSUE #2007: Steal connection on 2nd connection attempt
This would be an attempt at solving some of the issues described here: https://github.com/itchysats/itchysats/issues/1954.

bonomat:
Resolves #2007

If the same taker tries to connect twice we prioritize the second connection. This can be useful if the taker thinks he is disconnected from the maker but the maker does not


da-kami (approved):
I find it hard to define all the implications, but given that a taker that reconnects will not use the "old" connection anymore this is the only thing that makes sense so I'm happy to get it on testnet to see how it behaves.

luckysori (approved):

Issue #1545, opened 2022-03-02
Using Postgres as a database


thomaseizinger:
The goal of this ticket is to migrate the maker's db to a Postgres DB but keep the taker with SQLite. 

This includes a docker-based postgres setup for our local dev enivornment. 

Note: before merging this change we also need to upgrade our testnet setup and provide a postgres container

<details>
<summary>Old ticket details ⬇️</summary>
1. Do a PoC:

A rough outline of what is required to make this change. F.e. do we need to adjust the migrations or are they all standard conform. What about authentication? etc

The expected benefits of switching to postgres are:

- Improved performance
- Concurrent connections
- Move reporting out of the daemon into a separate program that queries the database

2. Based on PoC, decide on:

- Use postgres for maker and taker?
- If only for maker, how easy is it to create two database modules?
- Should we use sqlx's [`AnyConnection`](https://docs.rs/sqlx/latest/sqlx/struct.AnyConnection.html)?

## Annoyance counter

- 24/03/2022: Cannot easily access the production database and had to muck around with sshfs to get the sqlite file off the virtual machine.

</details>

rishflab:
I think we can separate this into 2 main categories of issues, production related and sqlite migration to postgres related:

Sqlite to postgres migration:
- [x] If we want to maintain sqlite compatibility how do we ensure this?
- [x] Do our existing queries work
- [x] Do our existing migrations work?
- [ ] Do we want to migrate existing data from sqlite to postgres?

Production:
- [ ] How to deploy production postgres (same vm vs seperate vm vs managed)
- [ ] How to manage production postgres (debugging, logs, upgrades)
- [ ] Authentication/secrets handling

Integration with metrics:
- [ ] grafana

Developer experience
- [ ] postgres dev setup

thomaseizinger:
This is a good start, thanks for writing this up!

The primary focus for now should be on figuring out, which code changes are required for switching to postgres.

thomaseizinger:
PoC was posted here: https://github.com/itchysats/itchysats/pull/1631

rishflab:
ah crap my commit containing the findings did not push. I just pushed it right now. Sorry!

thomaseizinger:
With what I just worked on, I was using multiple git worktrees and thus now have different databases (because they are different directories).

In this case, it would be have nice to also connect the taker to a postgres DB that I can just run locally. Docker makes that very easy.

[...]

PR #2054, opened 2022-05-17
Move sqlite db into own crate

bonomat:
This will allow us to support a second db such as postgres easier as we can duplicate the code and don't have to make it work for various implementations at the same time.

This is based on proposal by Thomas in https://github.com/itchysats/itchysats/issues/1545 and https://github.com/itchysats/itchysats/commit/1b2f851849697f30cd409460c57a468752b726ae


luckysori:
@bonomat: Ping me after the hefty rebase for a review!

bonomat:
> @bonomat: Ping me after the hefty rebase for a review!

I'll only do a rebase if we are willing to get this in, otherwise it's too much work for the fish 🐟

Can I get a concept ACK?

da-kami:
> This will allow us to support a second db such as postgres easier as we can duplicate the code and don't have to make it work for various implementations at the same time.

This makes sense. I think we should roll with this if it helps.
Would you push this through before or after we work on #2084 ?

bonomat:
> > This will allow us to support a second db such as postgres easier as we can duplicate the code and don't have to make it work for various implementations at the same time.
> 
> This makes sense. I think we should roll with this if it helps. Would you push this through before or after we work on #2084 ?

I'd get it in before #2084.

sqlite-db/src/lib.rs
```diff
@@ -30,14 +30,11 @@ use std::path::PathBuf;
...
-pub use failed::*;
+pub use db::closed::*;
+pub use db::failed::*;
 
-mod closed;
-mod event_log;
-mod failed;
-
-pub mod time_to_first_position;
+mod db;
```
luckysori: 🔧 I would flatten the module structure, getting rid of `db`. 

Might be better to do it in a follow-up commit to make the diff easier to review, but not sure.

sqlite-db/src/test_events/contract_setup_completed.json
```diff
@@ -0,0 +1,568 @@
+{
```
luckysori: ❓ Did we forget to delete the originals? I'm surprised these appear as new files.

sqlite-db/src/impls.rs
```diff
@@ -0,0 +1,44 @@
+use model::CfdEvent;
+
+impl crate::CfdAggregate for model::Cfd {
```
luckysori, line 3: ❓ I'm surprised that we had to move this here. Why was it?

I don't think we should have an `impls` module in the database layer. This crate provides some APIs which require the implementation of certain traits to be used. It should not need to know any details about its implementers.

Cargo.toml
```diff
@@ -16,6 +16,7 @@ members = [
   "bitmex-stream",
   "xtra-libp2p",
   "xtra-libp2p-ping",
+  "sqlite-db",
```
luckysori, line 19: 📣 You have to add `"clippy (sqlite-db)"` to `bors.toml` too!

luckysori (approved):
Concept ACK.

The existence of the tests in the first patch already made me unhappy. And having to move `db`-layer tests to `projection.rs` makes me even sadder. But I guess we can live with it for now. I do think it's a bad sign though.

sqlite-db/src/impls.rs
bonomat: We have to move this because we can't implement `CfdAggregate` for `model::CFD` in daemon. 

The alternative would be to add the newly created sqlite db crate as dependency to `model`.

sqlite-db/src/test_events/contract_setup_completed.json
bonomat: I had to duplicate it because this is needed for the tests which were moved to projection.rs

sqlite-db/src/lib.rs
bonomat: I can do that 👍

sqlite-db/src/test_events/contract_setup_completed.json
bonomat: I opted for moving it and doing file traversing to find the file.

sqlite-db/prepare_db.sh
```diff
@@ -15,4 +15,4 @@ trap 'rm -f $TEMPDB' EXIT
 DATABASE_URL=sqlite:$TEMPDB cargo sqlx migrate run
 
 # prepare the sqlx-data.json rust mappings
-DATABASE_URL=sqlite:./$DAEMON_DIR/$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare
+DATABASE_URL=sqlite:./$CRATE_DIR/$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare --merged -- --all-targets
```
da-kami: ❓What does this do `--merged -- --all-targets` do?

sqlite-db/src/impls.rs
da-kami: The database uses the model here not the other way round, so they way this was moved is correct!

da-kami (approved):

sqlite-db/prepare_db.sh
bonomat: Oh, this was a rebase error. It was originally added by @luckysori  and lateron removed. 
Thanks for the good eyes.

da-kami (approved):

klochowicz (approved):
looks good! 👍

sqlite-db/src/impls.rs
luckysori: This is correct. It just feels a little bad to me.

PR #2058, opened 2022-05-18
Use enum instead of booleans for indicating state

### LINKED ISSUE #2022: Simplify the `metrics` `CfdAggregate`
Follow-up of https://github.com/itchysats/itchysats/pull/2014

Instead of using `boolean` values we should depict finality as `enum`.

bonomat:
Note: with this we are losing the capability to see if a CFD is in a _weird_ state.

Resolves #2022

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -168,19 +168,10 @@ impl Cfd {
                 state: AggregatedState::Open,
                 ..self
             },
-            ContractSetupFailed => {
-                // This is needed due to a bug that has since been fixed; `OfferRejected` and
-                // `ContractSetupFailed` are mutually exclusive.
-                // We give `OfferRejected` priority over `ContractSetupFailed`.
-                if let AggregatedState::Rejected = self.state {
-                    self
-                } else {
-                    Self {
-                        state: AggregatedState::Failed,
-                        ..self
-                    }
-                }
-            }
+            ContractSetupFailed => Self {
+                state: AggregatedState::Failed,
+                ..self
+            },
             OfferRejected => Self {
                 state: AggregatedState::Rejected,
                 ..self
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -117,13 +117,20 @@ pub struct Cfd {
     position: Position,
     quantity_usd: Usd,
 
-    is_open: bool,
-    is_closed: bool,
-    is_failed: bool,
-    is_refunded: bool,
-    is_rejected: bool,
+    state: AggregatedState,
 
     version: u32,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum AggregatedState {
+    /// Used when the CFD is new or in contract setup
+    New,
+    Open,
+    Closed,
+    Failed,
+    Refunded,
+    Rejected,
 }
 
 impl db::CfdAggregate for Cfd {
@@ -134,11 +141,7 @@ impl db::CfdAggregate for Cfd {
             id: cfd.id,
             position: cfd.position,
             quantity_usd: cfd.quantity_usd,
-            is_open: false,
-            is_closed: false,
-            is_failed: false,
-            is_refunded: false,
-            is_rejected: false,
+            state: AggregatedState::New,
             version: 0,
         }
     }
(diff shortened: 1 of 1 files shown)


daemon/src/position_metrics.rs
```diff
@@ -158,31 +161,28 @@ impl Cfd {
...
+                state: AggregatedState::Open,
                 ..self
             },
             ContractSetupFailed => {
                 // This is needed due to a bug that has since been fixed; `OfferRejected` and
                 // `ContractSetupFailed` are mutually exclusive.
                 // We give `OfferRejected` priority over `ContractSetupFailed`.
-                if self.is_rejected {
-                    Self { ..self }
+                if let AggregatedState::Rejected = self.state {
```
luckysori: I would just set it to `Failed`. I don't think we gain anything by reasoning about the previous state.

daemon/src/position_metrics.rs
```diff
@@ -211,52 +210,44 @@ impl Cfd {
...
             CetConfirmed => Self {
-                is_open: false,
-                is_closed: true,
+                state: AggregatedState::Closed,
                 ..self
             },
             RefundConfirmed => Self {
-                is_open: false,
-                is_refunded: true,
+                state: AggregatedState::Closed,
```
luckysori: 🔧 

```suggestion
                state: AggregatedState::Refunded,
```

daemon/src/position_metrics.rs
```diff
@@ -211,52 +210,44 @@ impl Cfd {
...
                 ..self
             },
             LockConfirmedAfterFinality => Self {
                 // This event is only appended if lock confirmation happens after we spent from lock
                 // on chain. This is for special case where collaborative settlement is triggered
                 // before lock is confirmed. In such a case the CFD is closed
-                is_open: false,
-                is_closed: true,
+                state: AggregatedState::Closed,
```
luckysori, line 224: 🤔 It seems weird to overwrite anything here. I would rely on whatever happened before and not do anything here.

daemon/src/position_metrics.rs
```diff
@@ -211,52 +210,44 @@ impl Cfd {
...
                 ..self
             },
             CetTimelockExpiredPriorOracleAttestation
             | CetTimelockExpiredPostOracleAttestation { .. } => Self {
-                is_open: false,
-                is_closed: true,
+                state: AggregatedState::Closed,
```
luckysori, line 229: 🤔 Should we even set the `state` to "final" for anything other than `Confirmed` events, rejected and failed? Do we gain anything from reporting this stuff early in the metrics?

luckysori (commented):
Some more suggestions and a little fix.

daemon/src/position_metrics.rs
bonomat: Isn't this the bug what Daniel described?

daemon/src/position_metrics.rs
bonomat: True, we can just remain in whatever state we have been.

daemon/src/position_metrics.rs
bonomat: We would see that a CFD is basically closed at the time the CET expired, not only when the closed transaction was confirmed.

daemon/src/position_metrics.rs
luckysori: Fair enough: wrongly reporting failed can be problematic because we might think there's a problem when looking at the metrics.

On the other hand, this only applies for CFDs in the `cfds` table. As soon as we move CFDs to the `failed_cfds` table, this problem goes away. That combined with the fact that the comment says that the bug has already been fixed makes me think that we don't need any of this code (or the comment).

daemon/src/position_metrics.rs
luckysori: Is that preferable when looking at metrics? I just don't know.

bonomat: Yes, because we act on the same information in services built on top and would see discrepancies here.

daemon/src/position_metrics.rs
bonomat: I'll remove it in a separate commit.

luckysori (approved):
Oops, forgot to give approval.

PR #2059, opened 2022-05-18
Fix latency issues with the ping protocol

### LINKED ISSUE #2050: Libp2p `Ping` is taking longer over time
We had a look at stats and noticed that the metrics on `Ping` messages of the ping actor take significant time to be executed.
We process `Ping` once every 5 seconds for all peers - i.e. there is a for loop that spins off a task for each peer. Inside the task we do access the endpoint, which may take time. Seemingly there is no expensive blocking calls in that handler, so it is awkward that with more takers execution times become longer (yes there is a loop, but the loop only spins off tasks...).

The only thing that is a blocking call when we enter the handler is:

- We dispatch `CheckConnection` which dispatches `GetConnectionStats` when we handle `Ping` and in another actor (check_connection_active_in_interval) on the same interval. There seems to be a relationship between handling `GetConnectionStats` and `Ping`.

In general it would be great if we could try to reproduce the latency we observe on `mainnet` locally by writing a test that spins up multiple takers. Then we could more deterministically test and fix things without very long feeback looks.
The testnet number don't correlate well with mainnet, so these problems seem to be related with multiple takers connecting.

It would be great to also get stats on the metrics - how much overhead are the metrics adding?
We assume that the overhead should be minimal, but we do observe more problems after rolling out a lot more metrics.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2050.

Patch 206bc0a8d802f1c1683ea55475aa782288436b0e might be contentious (and does not need to be merged), but I am fairly sure that it is the way to go.

---

Check out https://github.com/itchysats/itchysats/commit/5882fd5851d3a36b886f651edd4a17b58154c44a to see how I tested it and read some observations.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -67,7 +67,7 @@ pub mod wire;
 pub const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 
 pub const ENDPOINT_CONNECTION_TIMEOUT: Duration = Duration::from_secs(20);
-pub const PING_INTERVAL: Option<Duration> = Some(Duration::from_secs(5));
+pub const PING_INTERVAL: Duration = Duration::from_secs(5);
 
 pub const N_PAYOUTS: usize = 200;
 
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -29,7 +29,6 @@ use xtra_bitmex_price_feed::QUOTE_INTERVAL_MINUTES;
 use xtra_libp2p::dialer;
 use xtra_libp2p::multiaddress_ext::MultiaddrExt;
 use xtra_libp2p::Endpoint;
-use xtra_libp2p_ping::ping;
 use xtra_libp2p_ping::pong;
 use xtras::supervisor;
 
@@ -202,11 +201,7 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
-        ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
-            .create(None)
-            .spawn(&mut tasks);
         let pong_address = pong::Actor::default().create(None).spawn(&mut tasks);
-
         let endpoint = Endpoint::new(
             TokioTcpConfig::new(),
             identity.libp2p,
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -29,6 +29,8 @@ use xtra_bitmex_price_feed::QUOTE_INTERVAL_MINUTES;
 use xtra_libp2p::dialer;
 use xtra_libp2p::multiaddress_ext::MultiaddrExt;
 use xtra_libp2p::Endpoint;
+use xtra_libp2p_ping::ping;
+use xtra_libp2p_ping::pong;
 use xtras::supervisor;
 
 pub use bdk;
@@ -200,9 +202,10 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
-        let ping_address = xtra_libp2p_ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
+        ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
             .create(None)
             .spawn(&mut tasks);
+        let pong_address = pong::Actor::default().create(None).spawn(&mut tasks);
 
         let endpoint = Endpoint::new(
             TokioTcpConfig::new(),
@@ -210,7 +213,7 @@ where
             ENDPOINT_CONNECTION_TIMEOUT,
             [(
                 xtra_libp2p_ping::PROTOCOL_NAME,
-                xtra::message_channel::StrongMessageChannel::clone_channel(&ping_address),
+                xtra::message_channel::StrongMessageChannel::clone_channel(&pong_address),
             )],
         );
 
--- a/xtra-libp2p-ping/src/pong.rs
+++ b/xtra-libp2p-ping/src/pong.rs
@@ -1,0 +1,30 @@
+use crate::protocol;
+use async_trait::async_trait;
+use tokio_tasks::Tasks;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_productivity::xtra_productivity;
+
+#[derive(Default)]
+pub struct Actor {
+    tasks: Tasks,
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, message: NewInboundSubstream) {
+        let NewInboundSubstream { stream, peer } = message;
+
+        let future = protocol::recv(stream);
+
+        self.tasks.add_fallible(future, move |e| async move {
+            tracing::debug!(%peer, "Inbound ping protocol failed: {e}");
+        });
+    }
+}
+
+#[async_trait]
+impl xtra::Actor for Actor {
+    type Stop = ();
+
+    async fn stopped(self) -> Self::Stop {}
+}
--- a/xtra-libp2p-ping/src/protocol.rs
+++ b/xtra-libp2p-ping/src/protocol.rs
@@ -1,0 +1,46 @@
+//! The actual protocol functions for sending ping messages.
+//!
+//! Insbired by https://github.com/libp2p/rust-libp2p/blob/102509afe3a3b984e43a88dbe4de935fde36f319/protocols/ping/src/protocol.rs#L82-L113.
+
+pub const SIZE: usize = 32;
+
+use futures::AsyncReadExt;
+use futures::AsyncWriteExt;
+use rand::distributions;
+use rand::thread_rng;
+use rand::Rng;
+use std::io;
+use std::time::Duration;
+use std::time::Instant;
+use xtra_libp2p::Substream;
+
+/// Sends a ping and waits for the pong.
+pub async fn send(mut stream: Substream) -> io::Result<Duration> {
+    let payload: [u8; SIZE] = thread_rng().sample(distributions::Standard);
+    stream.write_all(&payload).await?;
+    stream.flush().await?;
+
+    let started = Instant::now();
+
+    let mut recv_payload = [0u8; SIZE];
+    stream.read_exact(&mut recv_payload).await?;
+
+    if recv_payload != payload {
+        return Err(io::Error::new(
+            io::ErrorKind::InvalidData,
+            "Ping payload mismatch",
+        ));
+    }
+
+    Ok(started.elapsed())
+}
+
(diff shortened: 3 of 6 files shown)


daemon/src/lib.rs
```diff
@@ -202,11 +201,7 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
-        ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
-            .create(None)
-            .spawn(&mut tasks);
         let pong_address = pong::Actor::default().create(None).spawn(&mut tasks);
-
```
klochowicz: I feel like this should be a separate PR, as you suspect that it's a contentious change - let's chat about it separately.
the main reason is: it's hard to observe the possible congestions you're mitigating in this PR  if you disable that side of the protocol. 

all in all, I'm happy to explore alternatives, but I don't think they'll be easier - the one outlined in the commit message has the same drawback that the old heartbeat (intimate knowledge how pings on the other party are set up, ie. how often do they ping). Pinging from the taker overcomes this. If anything, I'd consider removing the heartbeat from the *maker* side - we don't really do anything with the latency measurements.

Also, we're not using this information *yet*, because we still rely on heartbeat - hopefully not for a long time.

daemon/src/lib.rs
```diff
@@ -200,17 +202,18 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
-        let ping_address = xtra_libp2p_ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
+        ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
```
klochowicz: It's a nifty way of getting around of congestion in ping, but I must admit am not in favour of this commit  - it just hides the underlying problem (congestion on processing external messages and trying to handle internal ones).

I feel like we need a more general solution, as this cannot be used in any protocols that require more interaction (e.g. rollover). 

the ping protocol problems we've seen is only a manifestation of problems that we see in other places, and should not be treated specially. We're looking at it as it's more isolated and easier to reproduce these problems - we can't roll such an approach to cfd actors etc.

We'll likely need a new way of scheduling internal messages in an async way, or a 2-tier queue (internal messages taking priority over external, or something else entirely!).

 out of curiosity: have you tried disabling prometheus metrics for the tests? has that changed anything?

klochowicz (changes requested):
I love the refactor and a few small fixups, but I do not support the 2 congestial changesets:
- splitting the actor (this just hides the problem),
- removing the ping from the taker

Any chance we can move them out of this PR so that we can discuss this is stages?

klochowicz:
I've re-requested review from myself, as I came to the following conclusion re: congestial changes:
- even if this does not solve the general problem, it immediately solves the biggest pain point we have in production re: libp2p (and we can always this old branch to come up with a general solution)
- sending pings from the takers arguably shouldn't be there yet (as per YAGNI - it's not used yet, and as you pointed out, there is a chance of solving it in a different way). I got around to the removal, and we talk about designing makes status inferral in a separate ticket when/if we get around to removing old heartbeats from the taker (to at least determine whether maker is online).

These 2 remaining points can and should be done separately. :)

Thanks heaps for investigating this issue so thoroughly 🥇

klochowicz (approved):

daemon/src/lib.rs
luckysori: > It's a nifty way of getting around of congestion in ping, but I must admit am not in favour of this commit - it just hides the underlying problem (congestion on processing external messages and trying to handle internal ones).

I see no benefit in having one actor for this protocol. The split shows that the `ping` and `pong` actors share no code. They are doing different things. They don't need to share state (the `pong` actor is stateless). It makes no sense to block the processing of pings on processing messages that access internal state that we don't care about!

I think the original design was flawed because it unnecessarily entangled different concerns into a single actor. It may be attractive to put everything in the same actor, but we have to have a strong reason to do so. After all, an actor is just a mechanism to perform actions concurrently based on state.

> I feel like we need a more general solution, as this cannot be used in any protocols that require more interaction (e.g. rollover).
>
> the ping protocol problems we've seen is only a manifestation of problems that we see in other places, and should not be treated specially. We're looking at it as it's more isolated and easier to reproduce these problems - we can't roll such an approach to cfd actors etc.

My learning from this patch is that we should try to build smaller actors. I suspect that we can see performance improvements by applying this idea to other parts of the codebase.

> We'll likely need a new way of scheduling internal messages in an async way, or a 2-tier queue (internal messages taking priority over external, or something else entirely!).

I am absolutely for introducing extensions to `xtra`, either by contributing or building on top! I think there is potential to make performance improvements that way too.

> out of curiosity: have you tried disabling prometheus metrics for the tests? has that changed anything?

I did, but I saw no notable improvements :(

daemon/src/lib.rs
luckysori: > I feel like this should be a separate PR, as you suspect that it's a contentious change - let's chat about it separately.
the main reason is: it's hard to observe the possible congestions you're mitigating in this PR if you disable that side of the protocol.

I am happy to move it. I was just lazy last night.

I just made a lot of observations using [the test](https://github.com/itchysats/itchysats/commit/5882fd5851d3a36b886f651edd4a17b58154c44a) and by far the biggest difference with the `1-to-m` actor setup was to not have the `m`s pinging too.

> all in all, I'm happy to explore alternatives, but I don't think they'll be easier - the one outlined in the commit message has the same drawback that the old heartbeat (intimate knowledge how pings on the other party are set up, ie. how often do they ping). Pinging from the taker overcomes this. If anything, I'd consider removing the heartbeat from the maker side - we don't really do anything with the latency measurements.

This is fair. We don't know what the best way to trigger reconnecting is yet. I still don't think pinging in both directions is wise, but I could be wrong.

daemon/src/lib.rs
klochowicz: > I think the original design was flawed because it unnecessarily entangled different concerns into a single actor. It may be attractive to put everything in the same actor, but we have to have a strong reason to do so. After all, an actor is just a mechanism to perform actions concurrently based on state.

an actor is an abstraction, and in this case I think the original design nailed the level of abstraction - abstracting one protocol per actor. IMHO this should not cause trouble if we didn't have some underlying problems. 

In recent work we started to avoid internal dispatch to actually achieve some concurrency - otherwise, if all the futures need to fan back in 1 by 1 to complete an internal dispatch, we love all the benefits! 

thanks for checking disabling the metrics, at least that tells us it's not the issue :-)

luckysori: > an actor is an abstraction, and in this case I think the original design nailed the level of abstraction - abstracting one protocol per actor. IMHO this should not cause trouble if we didn't have some underlying problems.

Sorry but I think this is wrong.

What is the benefit of putting the dialer and listener protocols in the same actor (other than code organisation, which is subjective), at the cost of fundamentally limiting our ability to perform the listener protocol in a timely manner?

By combining both protocols we are using one message queue when we need two.

Even the design of `xtra_libp2p` guides us towards only registering the listener protocol actor when constructing `Endpoint`: the dialer protocol works without passing its `xtra::Address` to the constructor. And the dialer protocol doesn't even need to be managed by an actor!

xtra-libp2p-ping/src/pong.rs
```diff
@@ -0,0 +1,30 @@
...
+use xtra_productivity::xtra_productivity;
+
+#[derive(Default)]
+pub struct Actor {
+    tasks: Tasks,
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, message: NewInboundSubstream) {
```
da-kami, line 14: ❓Why does pong not return a `pong`? 

Isn't the point of a ping protocol that you get a pong from the other party when sending a ping? Am I missing something?

xtra-libp2p-ping/src/protocol.rs
```diff
@@ -0,0 +1,46 @@
...
+            io::ErrorKind::InvalidData,
+            "Ping payload mismatch",
+        ));
+    }
+
+    Ok(started.elapsed())
+}
+
+/// Waits for a ping and sends a pong.
+pub(crate) async fn recv(mut stream: Substream) -> io::Result<()> {
```
da-kami: ❓ What is the point of separating these functions into `protocol`?

I know that the initial design also did that, but I have still not figure out what the advantage would be. I think it is way easier to understand if we inlined this code in the respective actor. 
For me the actor represents the protocol - there is not really a way (at least I don't see a clear one yet...) to separate "the protocol" (as in the pure message exchange) from the actor. With ping now using two actors it makes even less sense to me to have these functions separate from the actor. I feel it just makes it harder to follow the code.

da-kami, line 39: ❓ What is the point of separating these functions into `protocol`?

I know that the initial design also did that, but I have still not figure out what the advantage would be. I think it is way easier to understand if this is just inlined in the actor. 
For me the actor represents the protocol - there is not really a way (at least I don't see a clear one yet...) to separate "the protocol" (as in the pure message exchange) from the actor. With ping now using two actors it makes even less sense to me to have these functions separate from the actor. I feel it just makes it harder to follow the code.

da-kami (approved):
One comment on how to structure the code of these protocols. I was thinking about this when implementing rollover as well. Looking forward to discussing some of the findings here in context of the rollover impl :)

xtra-libp2p-ping/src/protocol.rs
luckysori: The ping protocol we are performing doesn't have anything to do with the substream that we send messages through.

I've introduced 18dbe20e8c12fcd09fb4743518c5c872ffd740b5 to demonstrate how this doesn't need to depend on `xtra`.

The message protocol is on a different level of abstraction than the actors that call its functions. For example, the `ping` actor calculates latencies based on performing the protocol. The message protocol could also be tested in isolation.

luckysori:
Moving ahead with the merge since what @da-kami mentions could be tackled separately if needed.

xtra-libp2p-ping/src/protocol.rs
da-kami: It will be interesting to "map" this information to other protocols we design. For rollover we somehow have a similar constraint:
The `setup_contract::roll_over` function is the "protocol" and everything else is only scaffolding to *start* that protocol - including the maker's accept and reject. 

But, when it comes to introducing a new version of the protocol then that would be at level of negotiating the substream - and that is done by the actor. So, to have a new protocol version we would need another actor as well, which renders this separation somewhat useless. Or am I missing something?

PR #2064, opened 2022-05-19
Libp2p rollover

da-kami:
Introduces protocol for rollover over libp2p connection.

Open TODO:

- [ ] Ensure to incorporate the changes of https://github.com/itchysats/itchysats/pull/2066 once it merges

---

On top of this PR we need a clearer picture how we see this evolve. Would be good to write an ADR about this
	
- [ ] Discussion about evolving protocol versions
		- Separation of protocol (messaging) and actor (decision making, processing external decisions e.g. accept/reject)
		- Multiple actors for one protocol version?
		- Decision making on the maker side?

- [ ] Discuss about scaling - congestion issues (bottleneck of actors being flooded with messages)
		- Potentially have one actor per protocol *per taker*
		Challanges: Ephemeral actors per connection
		- Potentially introduce bounded inboxes
		Challanges: Bounded inboxes is somewhat of an unknown

Note on congestion: We tried to avoid the pattern of "coming back to the actor from a task within the actor" because it likely leads to more congestion. We dispatch into a task early (on both the maker and taker side) and then finishing up is just done within the task using the executor (i.e. finish in the database) without involving the rollover actor. I think this is a good pattern but open to discussion :)

### DIFF
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -2,11 +2,14 @@ use crate::connection;
 use crate::db;
 use crate::oracle;
 use crate::process_manager;
+use crate::rollover;
+use crate::rollover::taker::ProposeRollover;
 use crate::rollover_taker;
 use anyhow::Result;
 use async_trait::async_trait;
 use futures::StreamExt;
 use maia_core::secp256k1_zkp::schnorrsig;
+use model::libp2p::PeerId;
 use model::OrderId;
 use std::time::Duration;
 use time::OffsetDateTime;
@@ -25,6 +28,7 @@ pub struct Actor<O> {
     conn: Address<connection::Actor>,
     oracle: Address<O>,
     n_payouts: usize,
+    libp2p_rollover: Address<rollover::taker::Actor>,
     rollover_actors: AddressMap<OrderId, rollover_taker::Actor>,
     tasks: Tasks,
 }
@@ -36,6 +40,7 @@ impl<O> Actor<O> {
         process_manager: Address<process_manager::Actor>,
         conn: Address<connection::Actor>,
         oracle: Address<O>,
+        libp2p_rollover: Address<rollover::taker::Actor>,
         n_payouts: usize,
     ) -> Self {
         Self {
@@ -45,6 +50,7 @@ impl<O> Actor<O> {
             conn,
             oracle,
             n_payouts,
+            libp2p_rollover,
             rollover_actors: AddressMap::default(),
             tasks: Tasks::default(),
         }
@@ -66,28 +72,47 @@ where
         }
     }
 
-    async fn handle(&mut self, Rollover(order_id): Rollover) {
-        let disconnected = match self.rollover_actors.get_disconnected(order_id) {
-            Ok(disconnected) => disconnected,
-            Err(_) => {
-                tracing::debug!(%order_id, "Rollover already in progress");
-                return;
-            }
-        };
-
-        let addr = rollover_taker::Actor::new(
+    async fn handle(
+        &mut self,
+        Rollover {
             order_id,
-            self.n_payouts,
-            self.oracle_pk,
-            self.conn.clone(),
-            &self.oracle,
-            self.process_manager.clone(),
-            self.db.clone(),
-        )
-        .create(None)
-        .spawn(&mut self.tasks);
+            maker_peer_id,
+        }: Rollover,
+    ) {
+        if let Some(maker_peer_id) = maker_peer_id {
+            if let Err(e) = self
+                .libp2p_rollover
+                .send(ProposeRollover {
+                    order_id,
+                    maker_peer_id,
+                })
+                .await
+            {
+                tracing::error!(%order_id, "Failed to dispatch proposal to libp2p rollover actor: {e:#}");
+            }
+        } else {
+            let disconnected = match self.rollover_actors.get_disconnected(order_id) {
+                Ok(disconnected) => disconnected,
+                Err(_) => {
+                    tracing::debug!(%order_id, "Rollover already in progress");
+                    return;
+                }
+            };
 
-        disconnected.insert(addr);
+            let addr = rollover_taker::Actor::new(
+                order_id,
+                self.n_payouts,
+                self.oracle_pk,
+                self.conn.clone(),
+                &self.oracle,
+                self.process_manager.clone(),
+                self.db.clone(),
+            )
+            .create(None)
+            .spawn(&mut self.tasks);
+
+            disconnected.insert(addr);
+        }
     }
 }
 
@@ -114,11 +139,17 @@ where
                 }
             };
             let id = cfd.id();
+            let maker_peer_id = cfd.counterparty_peer_id();
 
             match cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
                 Ok(()) => {
-                    let _ = this.send_async_safe(Rollover(id)).await; // If we disconnect, we don't
-                                                                      // care.
+                    // If we disconnect, we don't care.
+                    let _ = this
+                        .send_async_safe(Rollover {
+                            order_id: id,
+                            maker_peer_id,
+                        })
+                        .await;
                 }
                 Err(reason) => {
                     tracing::trace!(order_id = %id, %reason, "CFD is not eligible for auto-rollover");
@@ -155,4 +186,7 @@ pub struct AutoRollover;
 ///
 /// This helps us trigger rollover in the tests unconditionally of time.
 #[derive(Clone, Copy)]
-pub struct Rollover(pub OrderId);
+pub struct Rollover {
+    pub order_id: OrderId,
+    pub maker_peer_id: Option<PeerId>,
+}
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -52,6 +52,7 @@ pub mod oracle;
 pub mod position_metrics;
 pub mod process_manager;
 pub mod projection;
+pub mod rollover;
 pub mod rollover_taker;
 pub mod seed;
 pub mod setup_contract;
@@ -175,12 +176,25 @@ where
         .create(None)
         .spawn(&mut tasks);
 
+        let (endpoint_addr, endpoint_context) = Context::new(None);
+
+        let libp2p_rollover_addr = rollover::taker::Actor::new(
+            endpoint_addr.clone(),
+            executor.clone(),
+            oracle_pk,
+            xtra::message_channel::MessageChannel::clone_channel(&oracle_addr),
+            n_payouts,
+        )
+        .create(None)
+        .spawn(&mut tasks);
+
         let auto_rollover_addr = auto_rollover::Actor::new(
             db.clone(),
             oracle_pk,
             process_manager_addr,
             connection_actor_addr.clone(),
             oracle_addr,
+            libp2p_rollover_addr,
             n_payouts,
         )
         .create(None)
@@ -202,8 +216,6 @@ where
         tasks.add(monitor_ctx.run(monitor_constructor(executor.clone())?));
 
         tasks.add(oracle_ctx.run(oracle_constructor(executor.clone())));
-
-        let (endpoint_addr, endpoint_context) = Context::new(None);
 
         let pong_address = pong::Actor::default().create(None).spawn(&mut tasks);
         let endpoint = Endpoint::new(
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -1,0 +1,288 @@
+use anyhow::Context;
+use anyhow::Result;
+use async_trait::async_trait;
+use asynchronous_codec::Framed;
+use asynchronous_codec::JsonCodec;
+use futures::future;
+use futures::SinkExt;
+use futures::StreamExt;
+use libp2p_core::PeerId;
+use maia_core::secp256k1_zkp::schnorrsig;
+use model::FundingRate;
+use model::OrderId;
+use model::Position;
+use model::Role;
+use model::RolloverVersion;
+use model::TxFeeRate;
+use std::collections::HashMap;
+use tokio_tasks::Tasks;
+use xtra::message_channel::MessageChannel;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_libp2p::Substream;
+use xtra_productivity::xtra_productivity;
+
+use crate::command;
+use crate::oracle;
+use crate::rollover::protocol::*;
+use crate::setup_contract;
+
+use super::protocol;
+
+type ListenerConnection = (
+    Framed<Substream, JsonCodec<ListenerMessage, DialerMessage>>,
+    PeerId,
+);
+
+/// Permanent actor to handle incoming substreams for the `/itchysats/rollover/1.0.0`
+/// protocol.
+///
+/// There is only one instance of this actor for all connections, meaning we must always spawn a
+/// task whenever we interact with a substream to not block the execution of other connections.
+pub struct Actor {
+    tasks: Tasks,
+    oracle_pk: schnorrsig::PublicKey,
+    get_announcement: Box<dyn MessageChannel<oracle::GetAnnouncement>>,
+    n_payouts: usize,
+    pending_protocols: HashMap<OrderId, ListenerConnection>,
+    executor: command::Executor,
+}
+
+impl Actor {
+    pub fn new(
+        executor: command::Executor,
+        oracle_pk: schnorrsig::PublicKey,
+        get_announcement: Box<dyn MessageChannel<oracle::GetAnnouncement>>,
+        n_payouts: usize,
+    ) -> Self {
+        Self {
+            tasks: Tasks::default(),
+            oracle_pk,
+            get_announcement,
+            n_payouts,
+            pending_protocols: HashMap::default(),
+            executor,
+        }
+    }
+}
+
+#[async_trait]
+impl xtra::Actor for Actor {
+    type Stop = ();
+
+    async fn stopped(self) -> Self::Stop {}
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, msg: NewInboundSubstream, ctx: &mut xtra::Context<Self>) {
+        let NewInboundSubstream { peer, stream } = msg;
+        let address = ctx.address().expect("we are alive");
+
+        self.tasks.add_fallible(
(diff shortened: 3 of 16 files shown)


da-kami:
Make sure to incorporate changes of 
https://github.com/itchysats/itchysats/pull/2066
if it merges before

daemon/src/auto_rollover.rs
```diff
@@ -66,28 +72,47 @@ where
...
+                .send(ProposeRollover {
+                    order_id,
+                    maker_peer_id,
+                })
+                .await
+            {
+                tracing::error!(%order_id, "Failed to dispatch proposal to libp2p rollover actor: {e:#}");
             }
-        };
+        } else {
```
luckysori, line 93: ❓ Why support the old protocol on the taker side? I thought this would disappear.

luckysori: Handling this seems to contradict part of the commit message:

> All the CFDs that were open before this changeset will also attempt to rollover using libp2p.

daemon/src/rollover/taker.rs
```diff
@@ -0,0 +1,223 @@
...
+            .send(OpenSubstream::single_protocol(
+                maker_peer_id.inner(),
+                rollover::PROTOCOL,
+            ))
+            .await
+            .context("Endpoint is disconnected")
+            .context("Failed to open substream")
+        {
+            Ok(Ok(substream)) => substream,
+            // TODO: Same error path and add failed
```
da-kami: @klochowicz we have to fix this one :)

daemon/src/lib.rs
```diff
@@ -175,12 +176,25 @@ where
         .create(None)
         .spawn(&mut tasks);
 
+        let (endpoint_addr, endpoint_context) = Context::new(None);
+
+        let libp2p_rollover_addr = rollover::taker::Actor::new(
+            endpoint_addr.clone(),
+            executor.clone(),
+            oracle_pk,
+            Box::new(oracle_addr.clone()),
```
luckysori: 🙃 `oracle_addr.clone_channel()` is what we would usually do.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+impl xtra::Actor for Actor {
+    type Stop = ();
+
+    async fn stopped(self) -> Self::Stop {}
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, msg: NewInboundSubstream, ctx: &mut xtra::Context<Self>) {
+        let NewInboundSubstream { peer, stream } = msg; // TODO: Use `PeerId` for something!
```
luckysori: ❓ Do we have to use the `PeerId`? I'm not sure.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+            propose,
+            framed,
+            peer,
+        } = msg;
+        let order_id = propose.order_id;
+
+        if let Err(e) = self
+            .executor
+            .execute(order_id, |cfd| {
+                // TODO: Validate that the correct peer is invoking this?
```
luckysori, line 121: Does it really matter if it's the wrong `PeerId`. It will never succeed if they don't have the correct secret keys.

We could fail early by checking here, but it's not a huge deal imo.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+                cfd.start_rollover()
+            })
+            .await
+        {
+            tracing::warn!(%order_id, "Rollover failed when being proposed - this can mean that a rollover never finished: {e:#}");
```
luckysori: I don't understand the extra information here. I would say `Rollover failed after handling taker proposal`.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+
+            // We have to append failed to ensure that we can rollover in the future
+            // The cfd logic might otherwise prevent us from starting a rollover if there is still
+            // one ongoing that was not properly ended.
+            append_failed(order_id, e, &self.executor).await;
```
luckysori: 😞  That is annoying.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+            return;
+        }
+
+        // In case we fail to accept/reject some proposals might never get cleaned up. Given that
+        // the taker will retry this should not do much harm because we will replace the proposal
+        // with a new one. This is acceptable for now.
+        self.pending_protocols.insert(order_id, (framed, peer));
+    }
+
+    async fn handle(&mut self, msg: Accept) -> Result<()> {
```
luckysori, line 142: Still hate the fact that we need to wait for external approval to accept a rollover. Makes this actor more complicated when we could configure this stuff differently (and for _all_ rollovers).

Obviously nothing to action. I'm just annoyed at this haha.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+        let Accept {
+            order_id,
+            tx_fee_rate,
+            long_funding_rate,
+            short_funding_rate,
+        } = msg;
+
+        let (mut framed, _) = self
```
luckysori: ❓ We're saving the `PeerId` bu then we don't use it?

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+
+                    let (sink, stream) = framed.split();
+
+                    let dlc = setup_contract::roll_over(
+                        sink.with(|msg| future::ok(ListenerMessage::RolloverMsg(Box::new(msg)))),
+                        Box::pin(stream.filter_map(|msg| async move {
+                            match msg {
+                                Ok(msg) => msg.into_rollover_msg().ok(),
+                                Err(e) => {
+                                    tracing::error!("Conversion error: {e:#}");
```
luckysori: 🔧 Perhaps the error message could point to the rollover protocol.

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+                    framed
+                        .send(ListenerMessage::Decision(Decision::Confirm(Confirm {
+                            order_id,
+                            oracle_event_id,
+                            tx_fee_rate,
+                            funding_rate,
+                            complete_fee: complete_fee.into(),
+                        })))
+                        .await
+                        .context("Failed to send Msg1::Accept")?;
```
luckysori: 🔧 Maybe simpler to say `Failed to send rollover confirmation message`.

klochowicz:
In order to facilitate zooming out and seeing the big picture going forward, this is how much could be removed from the taker alone if this PR gets in:
https://github.com/itchysats/itchysats/pull/2077

daemon/src/rollover/maker.rs
```diff
@@ -0,0 +1,293 @@
...
+        Ok(())
+    }
+
+    async fn handle(&mut self, msg: Reject) -> Result<()> {
+        let Reject { order_id } = msg;
+
+        let (mut framed, ..) = self
+            .pending_protocols
+            .remove(&order_id)
+            .with_context(|| format!("No active protocol for order {order_id}"))?;
```
luckysori: You may want to add context either here or wherever we send the `Reject` message so that when we see this log we quickly realise that this is just about rejecting a rollover that has already disappeared i.e. not critical.

daemon/src/rollover/protocol.rs
```diff
@@ -0,0 +1,149 @@
...
+) {
+    tracing::error!(%order_id, "Rollover failed: {e:#}");
+
+    if let Err(e) = executor
+        .execute(order_id, |cfd| Ok(cfd.fail_rollover(e)))
+        .await
+    {
+        tracing::error!(%order_id, "Failed to execute rollover failed: {e:#}")
+    }
+}
```
luckysori: 🙃 I don't think the `append_` terminology is ideal, because the caller is just executing a command. Maybe just give it the same name as the command it calls.

Also, nitpicky, but I would not put these in the `protocol` submodule. They're just utilities.

daemon/src/rollover/protocol.rs
```diff
@@ -0,0 +1,149 @@
...
+    order_id: OrderId,
+    dlc: Dlc,
+    funding_fee: FundingFee,
+    executor: &command::Executor,
+) {
+    if let Err(e) = executor
+        .execute(order_id, |cfd| Ok(cfd.complete_rollover(dlc, funding_fee)))
+        .await
+    {
+        tracing::warn!(%order_id, "Failed to execute rollover completed: {e:#}")
```
luckysori: 🔧 Probably should be `error` to be consistent with the others.

daemon/src/rollover/protocol.rs
```diff
@@ -0,0 +1,149 @@
...
+    Rejected,
+    #[error("Rollover failed")]
+    Failed { source: anyhow::Error },
+}
+
+impl From<anyhow::Error> for DialerError {
+    fn from(source: anyhow::Error) -> Self {
+        Self::Failed { source }
+    }
+}
```
luckysori: ❓ Can't you just annotate the `Failed` variant with `#[from]` or `#[source]` instead of manually implementing this?

daemon/src/rollover/protocol.rs
```diff
@@ -0,0 +1,149 @@
...
+pub enum DialerMessage {
+    Propose(Propose),
+    RolloverMsg(Box<RolloverMsg>),
+}
+
+impl DialerMessage {
+    pub fn into_propose(self) -> Result<Propose> {
+        match self {
+            DialerMessage::Propose(propose) => Ok(propose),
+            DialerMessage::RolloverMsg(_) => Err(anyhow!("Expected Propose but got RolloverMsg")),
```
luckysori: 🔧 It's more idiomatic to use `bail!` instead of `Err(anyhow!(_))`.

Here and below.

daemon/src/rollover/taker.rs
```diff
@@ -0,0 +1,223 @@
...
+use xtra_libp2p::OpenSubstream;
+use xtra_productivity::xtra_productivity;
+
+/// The duration that the taker waits until a decision (accept/reject) is expected from the maker
+///
+/// If the maker does not respond within `DECISION_TIMEOUT` seconds then the taker will fail the
+/// rollover.
+const DECISION_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// One actor to rule all the rollovers
```
luckysori, line 39: 😄

daemon/src/rollover/taker.rs
```diff
@@ -0,0 +1,223 @@
...
+            Ok(Err(e)) => {
+                tracing::error!("Failed to open connection to maker: {e:#}");
+                return;
+            }
+            Err(e) => {
+                tracing::error!(%order_id, "Failed to open connection to maker for  rollover: {e:#}");
+
+                if let Err(e) = self
+                    .executor
+                    .execute(order_id, |cfd| Ok(cfd.fail_rollover(e)))
```
luckysori: ❓ Why record a failure? Surely nothing has happened yet in the database at this stage.

daemon/src/rollover/taker.rs
```diff
@@ -0,0 +1,223 @@
...
+            .await
+            .context("Endpoint is disconnected")
+            .context("Failed to open substream")
+        {
+            Ok(Ok(substream)) => substream,
+            // TODO: Same error path and add failed
+            Ok(Err(e)) => {
+                tracing::error!("Failed to open connection to maker: {e:#}");
+                return;
+            }
```
luckysori: 🔧 I think it will be more convenient to extract this into an `Actor` method, use `?` and handle all errors with a single `if let Err(e)` in this `xtra` handler.

daemon/src/rollover/taker.rs
```diff
@@ -0,0 +1,223 @@
...
+                        .timeout(DECISION_TIMEOUT)
+                        .await
+                        .with_context(|| {
+                            format!(
+                                "Maker did not accept/reject within {} seconds.",
+                                DECISION_TIMEOUT.as_secs()
+                            )
+                        })?
+                        .context("End of stream while receiving Msg1")?
+                        .context("Failed to decode Msg1")?
```
luckysori: 🔧 Probably should use something more descriptive than `Msg1` in the error message.

maker/src/actor_system.rs
```diff
@@ -146,7 +161,7 @@ where
         let endpoint_listen = daemon::libp2p_utils::create_listen_tcp_multiaddr(&libp2p_socket)
             .expect("to parse properly");
 
-        let (supervisor, _listener_actor) = supervisor::Actor::with_policy(
+        let (supervisor, listener_actor) = supervisor::Actor::with_policy(
```
luckysori, line 163: 🔧 Can you do this in a separate commit? Would be nice to explain the fix separately.

maker/src/cfd.rs
```diff
@@ -525,44 +528,71 @@ impl<O, T, W> Actor<O, T, W> {
...
-            .rollover_actors
-            .send_async(
-                &order_id,
-                rollover::AcceptRollover {
+        let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
+
+        if cfd.counterparty_peer_id().is_some() {
+            // Using send here is fine because we dispatch to a task internally
+            self.libp2p_rollover
+                .send(daemon::rollover::maker::Accept {
```
luckysori, line 537: ❓ Do we need to though?

luckysori (approved):
Great job! Looks so simple 😍

I don't think there's anything blocking!

daemon/src/auto_rollover.rs
da-kami: This was mostly for the case where we cannot default the maker peer id - but in production / testnet this should never be a problem and locally it actually does not matter, so yeah, we should remove it :)

daemon/src/rollover/maker.rs
da-kami: Yes! But I did not want to touch the model / events as part of this.

daemon/src/rollover/maker.rs
da-kami: The listener does not need it. We mostly save it for logging on the maker side - similar to the taker network identity. The taker always connects, but the maker knows the identity so the maker knows how it's talking to.

daemon/src/auto_rollover.rs
klochowicz: There is a follow-up PR that removes the old protocol! #2077  (marked as a draft, as we haven't yet got data whether libp2p will be the silver bullet for all our troubles ;-). It can go in shortly after we decide whether libp2p is good enough to keep it.

maker/src/cfd.rs
da-kami: The original idea was to be able to give *some* error information back to the HTTP endpoint. But I don't think the caller actually cares. 
The only error that is currently returned is an error where we don't have a pending rollover internally. This could be useful for the caller at some point. But for now probably not. 
I'm happy to change this to dispatch async, but then I'd say we should change the `Accept` handler *not* to return a `Result`. It does not make sense to return a `Result` if we don't care about it. @klochowicz what do you think?

daemon/src/rollover/maker.rs
da-kami: Yeah, this is just about failing early. A taker could "attack" a maker by sending bogus rollover proposals to keep the maker busy. No money would be lost, but it could result in a denial of service attack. Not super important yet but if we can easily do it it might be better to just do it now :)

daemon/src/lib.rs
da-kami: Somehow that did not work as easily when using it in the `lib`. Don't really understand why. I think we should move the `TakerActorSystem` into the taker. But we can do that as follow up.

daemon/src/rollover/maker.rs
da-kami: I removed the `TODO` - I think it was outdated.

daemon/src/rollover/protocol.rs
da-kami: changed to `emit` as it was in the legacy protocol

daemon/src/rollover/taker.rs
da-kami: We still have the paradigm of `started` and `failed`. If we only have a `started` event then the logic in `model::Cfd` won't allow triggering a rollover, because it assumes that one is still ongoing. Only if we add `completed`, `rejected` or `failed` the rollover is "finished". 
I don't like this, but I would not change this logic as part of this work.

daemon/src/rollover/taker.rs
da-kami: Given that we are on the taker we could consider it. Generally I would like to avoid the pattern of sending messages to ourselves because it adds unnecessary overhead.

daemon/src/rollover/taker.rs
klochowicz: recording a failure is trace of event that happened (and was significant). can be useful for analysing trends in grafana. 
and as @da-kami pointed out above, we actually currently need to do that to accurately determine ongoing rollover for a Cfd

maker/src/cfd.rs
klochowicz: So far we rely on sending back some error information back to the caller (e.g. if some invariants were not satisfied for accepting; we do that for all the protocols, so I propose not changing it here.
. I think it's the right way to do, because it seems to have more advantages (like being able to sometimes report errors, or ensuring an eager user can't call it twice at the same time).

If we'd like to explore moving away from it, it should rather have it's own discussion / POC to see how it affects the codebase. I don't think it's worth it at this stage.

daemon/src/rollover/taker.rs
klochowicz: I extracted a little helper (I understood @luckysori comment as extracting a helper for a bit nicer error handling). It make things a bit more clear, as it flattened the match statement and made the code more readable by not repeating the code.
I opted for not excluding the whole handler as we still need to `emit_fail` on the error - if it was just logging, I'd return an error from the handler and put everything in the method.

daemon/src/rollover/protocol.rs
klochowicz: awesome! didn't know about it - used `#[source]`! t

maker/src/actor_system.rs
klochowicz: done!

klochowicz (approved):
I've addressed a few more comments from @luckysori and squashed all the commits. Let's test it on testnet over the weekend 💪

PR #2066, opened 2022-05-19
Introduce rollover version 3

### LINKED ISSUE #1974: rollover happening every two hours

luckysori:
Fixes #1974 (we can still rollover every two hours, but we should charge/pay accordingly).

Taker now default to version 3. The maker supports all versions.

This new version calculates the hours to extend the settlement time by during rollover based on the event ID that will actually be proposed. The motivation is to charge for rollovers based on the actual extension to the settlement time of the CFD that they provide.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -909,7 +909,7 @@ impl Cfd {
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
         version: rollover::Version,
-    ) -> Result<(CfdEvent, RolloverParams, Dlc, Position, Duration)> {
+    ) -> Result<(CfdEvent, RolloverParams, Dlc, Position, BitMexPriceEventId)> {
         if !self.during_rollover {
             bail!("The CFD is not rolling over");
         }
@@ -918,9 +918,14 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
+        let candidate_event_id =
+            olivia::next_announcement_after(OffsetDateTime::now_utc() + self.settlement_interval);
         let hours_to_charge = match version {
             rollover::Version::V1 => 1,
             rollover::Version::V2 => self.hours_to_extend_in_rollover()?,
+            rollover::Version::V3 => {
+                self.hours_to_extend_in_rollover_based_on_event(candidate_event_id)?
+            }
         };
 
         let funding_fee = FundingFee::calculate(
@@ -955,7 +960,7 @@ impl Cfd {
             ),
             self.dlc.clone().context("No DLC present")?,
             self.position,
-            self.settlement_interval,
+            candidate_event_id,
         ))
     }
 
@@ -974,7 +979,10 @@ impl Cfd {
 
         self.can_rollover()?;
 
-        let hours_to_charge = self.hours_to_extend_in_rollover()?;
+        let candidate_event_id =
+            olivia::next_announcement_after(OffsetDateTime::now_utc() + self.settlement_interval);
+        let hours_to_charge =
+            self.hours_to_extend_in_rollover_based_on_event(candidate_event_id)?;
         let funding_fee = FundingFee::calculate(
             self.initial_price,
             self.quantity,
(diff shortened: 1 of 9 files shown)


da-kami (dismissed):
Is this important to get in? 
If not then I'd block this until rollover over libp2p is done and then take decisions on how to get this in.
The old rollover protocol should not change anymore but we should evolve the new protocol versions otherwise this becomes unmaintainable.

luckysori:
> Is this important to get in? If not then I'd block this until rollover over libp2p is done and then take decisions on how to get this in. The old rollover protocol should not change anymore but we should evolve the new protocol versions otherwise this becomes unmaintainable.

We talked about this a few days ago and it's part of the iteration.

I think we agreed that maintaining the old protocol is still important until we decide what we end up doing with libp2p. What's the concern with fixing this bug?

model/src/cfd.rs
```diff
@@ -1470,6 +1478,38 @@ impl Cfd {
...
+
+        let hours_left = current_settlement_time - OffsetDateTime::now_utc();
+
+        tracing::trace!(target = "cfd", time_left_in_cfd = %hours_left, "Calculating hours to extend in rollover");
+
+        if !hours_left.is_positive() {
+            tracing::warn!("Rolling over a contract that can be settled non-collaboratively");
+
+            return Ok(SETTLEMENT_INTERVAL.whole_hours() as u64);
+        }
```
bonomat, line 1496: This should not happen right?

model/src/cfd.rs
```diff
@@ -3100,6 +3140,86 @@ mod tests {
...
+            );
+
+            assert_eq!(
+                maker
+                    .hours_to_extend_in_rollover_based_on_event(candidate_event_id)
+                    .unwrap(),
+                (candidate_event_id.timestamp() - current_event_id.timestamp()).whole_hours()
+                    as u64
+            );
+        }
```
bonomat, line 3199: I find it hard to reason about asserts within a for loop. What do you think of adding one additional positive test case for a value between `0` and `settlement_interval`? It might be redundant but can help readability.

bonomat (approved):
Nice. I think we should keep maintainting the old protocol until we know for sure that we want to go with libp2p

model/src/cfd.rs
luckysori: I don't think so given how the application works.

But I don't think the rollover protocol should stop you if it happens.

Mind you this is just copied from the method introduced with version 2.

model/src/cfd.rs
luckysori: Are you suggesting to add a separate test function which is the same as this one but for only a single case? I can do that.

bonomat: yes please.

da-kami:
> > Is this important to get in? If not then I'd block this until rollover over libp2p is done and then take decisions on how to get this in. The old rollover protocol should not change anymore but we should evolve the new protocol versions otherwise this becomes unmaintainable.
> 
> We talked about this a few days ago and it's part of the iteration.
> 
> I think we agreed that maintaining the old protocol is still important until we decide what we end up doing with libp2p. What's the concern with fixing this bug?

Probably I was premature blocking it. I think we can get this in. We just have to be careful upgrading libp2p to use this version as well (or more generally speaking: the libp2p protocol has to include these changes). Given that it's currently based on the old impl that will be easy - we just bump the version in the taker).

We have to be aware that when rolling out libp2p takers that upgrade will use libp2p only. So libp2p protocols have to be at least as much up to date as legacy (or ahead).

da-kami:
Actually, we will have to adapt more than just the version because this touches how we prepare for the rollover. Still not hard, but it involve changing a bit more code.

PR #2072, opened 2022-05-19
Do not cull all but one DLCs 

### LINKED ISSUE #2070: Query loads all but last entry but should load all but last per CFD
I think the query below is not working as expected:

https://github.com/itchysats/itchysats/blob/b36cd30ece958ab861b08032787a41d4baa2dbd5/daemon/src/db/cull_old_dlcs.rs#L37-L54

It should load all events but the last one of _each cfd_ but what it does is it loads all but the _last overall_.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2070.

The `cull_old_dlcs` API was meant to delete the old DLC data of all CFDs. The implementation had a huge bug which instead deleted all DLC data (old or current) for all open CFDs, except for the data of the oldest DLC-related event ever.

We modify the test that was meant to verify that we don't delete unexpected stuff when calling this API to make sure that we don't introduce a similar bug in the future.

### DIFF
--- a/daemon/src/db/time_to_first_position.rs
+++ b/daemon/src/db/time_to_first_position.rs
@@ -66,12 +66,79 @@ impl Connection {
     }
 }
 
+// We cannot hide this under the `test` compilation flag because it
+// makes it much less convenient to call `cargo sqlx prepare`.
+#[allow(dead_code)]
+mod sqlx_test_utils {
+    use super::*;
+    use sqlx::pool::PoolConnection;
+    use sqlx::Sqlite;
+
+    pub(crate) async fn load_first_seen_timestamp(
+        conn: &mut PoolConnection<Sqlite>,
+        taker_id: Identity,
+    ) -> Result<Option<OffsetDateTime>> {
+        let row = sqlx::query!(
+            r#"
+            SELECT
+                first_seen_timestamp
+            FROM
+                time_to_first_position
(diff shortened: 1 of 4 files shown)


daemon/src/db/time_to_first_position.rs
```diff
@@ -66,12 +66,79 @@ impl Connection {
     }
 }
 
+// We cannot hide this under the `test` compilation flag because it
+// makes it much less convenient to call `cargo sqlx prepare`.
+#[allow(dead_code)]
+mod sqlx_test_utils {
```
bonomat, line 72: Thanks for fixing this. I was wondering already but flagged it as my computer doing weird things

bonomat (commented):
Green light for the first commit, maybe extract it into an own PR for the time being. 
I'm working on a different query for cull_old_dlcs situation

daemon/src/db/time_to_first_position.rs
da-kami: Same here!

luckysori: I'm gonna split this into a separate PR!

luckysori:
Superseded by https://github.com/itchysats/itchysats/pull/2073 and https://github.com/itchysats/itchysats/pull/2074.

PR #2073, opened 2022-05-20
Remove unneeded event data in one query

### LINKED ISSUE #2070: Query loads all but last entry but should load all but last per CFD

bonomat:
By using json_set we can manipulate the json within the DB. This allows us to have a single query which removes unneeded event data, i.e. RolloverCompleted events which have a succeeding RolloverCompleted event and ContractSetupCompleted events that also have a succeeding RolloverCompleted event.

Resolves #2070

First patch includes commit from #2072 and can be dropped later.

### DIFF
--- a/daemon/src/db/cull_old_dlcs.rs
+++ b/daemon/src/db/cull_old_dlcs.rs
@@ -1,127 +1,88 @@
 use crate::db::Connection;
 use anyhow::Result;
-use model::EventKind;
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
 
 impl Connection {
+    /// Cull DLC data from old events.
+    ///
+    /// This method modifies the `dlc` field of the JSON blobs stored in the `events`
+    /// table for all the rows which do not correspond to the most recent
+    /// `EventKind::ContractSetupCompleted` or `EventKind::RolloverCompleted` event
+    /// of a CFD.
     pub async fn cull_old_dlcs(&self) -> Result<()> {
         let mut conn = self.inner.acquire().await?;
 
-        let emptied_dlc_events = emptied_dlc_events(&mut conn).await?;
-
-        for (id, event) in emptied_dlc_events {
-            if let Err(e) = update_event_data(&mut conn, id, event).await {
-                tracing::warn!("Failed to overwrite old DLC: {e:#}");
-            }
-        }
+        Connection::cull_rollover_completed(&mut conn).await?;
+        Connection::cull_contract_setup_completed(&mut conn).await?;
 
         Ok(())
     }
-}
 
-/// Load all old events with DLC data from the database and return
-/// them with empty `dlc` fields.
-///
-/// `EventKind::ContractSetupCompleted` an
-/// `EventKind::RolloverCompleted` are the two kinds of event that can
-/// have DLC data associated with them. One of these event is
-/// considered old if it isn't the most recent one related to a CFD.
-///
-/// This method does not modify the database: it just loads rows and
-/// processes them. The returned values can later be used to update
-/// the contents of the database.
-async fn emptied_dlc_events(
-    conn: &mut PoolConnection<Sqlite>,
-) -> Result<impl Iterator<Item = (i64, EventKind)>> {
-    let rows = sqlx::query!(
-        r#"
-        SELECT
-            id,
-            name,
-            data
-        FROM
-            events
-        WHERE
-            name = $1 OR
-            name = $2
-        ORDER BY
-            created_at DESC
-        LIMIT -1 OFFSET 1
-        "#,
-        model::EventKind::CONTRACT_SETUP_COMPLETED_EVENT,
-        model::EventKind::ROLLOVER_COMPLETED_EVENT,
-    )
-    .fetch_all(&mut *conn)
-    .await?;
+    /// Cull old cfd rollover events which are not needed anymore
+    ///
+    /// We set the `dlc` field of events `RolloverCompleted` to null where the event is not the most
+    /// recent `RolloverCompleted` event.
+    async fn cull_rollover_completed(conn: &mut PoolConnection<Sqlite>) -> Result<()> {
+        let affected_rows = sqlx::query!(
+            r#"
+            update EVENTS
+                set
+                    data = (
+                        select json_set(events.data, '$.dlc', null)
+                        from events as inner_events
+                        where inner_events.id = EVENTS.id 
+                    )
+                where name = $1
+                  and id not in (
+                    select max(id) from EVENTS as inner_events
+                    where inner_events.cfd_id = EVENTS.cfd_id
+                      and inner_events.name = EVENTS.name
+                );
+            "#,
+            model::EventKind::ROLLOVER_COMPLETED_EVENT,
+        )
+        .execute(&mut *conn)
+        .await?
+        .rows_affected();
 
-    Ok(rows
-        .into_iter()
-        .filter_map(|row| {
-            let id = row.id;
-            let event = match EventKind::from_json(row.name, row.data) {
-                Ok(event) => Some(event),
-                Err(e) => {
-                    tracing::warn!("Failed to deserialize EventKind from JSON: {e:#}");
-                    None
-                }
-            };
+        tracing::debug!(%affected_rows,"Culled RolloverCompleted events");
 
-            event.map(|event| (id, event))
-        })
-        .filter_map(|(id, event)| {
-            use EventKind::*;
-            let event = match event {
-                ContractSetupCompleted { dlc: Some(_), .. } => {
-                    Some(ContractSetupCompleted { dlc: None })
-                }
-                RolloverCompleted {
-                    dlc: Some(_),
-                    funding_fee,
-                } => Some(RolloverCompleted {
-                    dlc: None,
-                    funding_fee,
-                }),
-                // ignore events that have already been culled
-                ContractSetupCompleted { dlc: None } | RolloverCompleted { dlc: None, .. } => None,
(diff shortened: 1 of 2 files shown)


maker/src/actor_system.rs
```diff
@@ -83,7 +84,10 @@ where
             + Handler<monitor::MonitorCetFinality>
             + Actor<Stop = ()>,
     {
+        let now = Instant::now();
+        tracing::info!("Culling started");
         futures::executor::block_on(db.cull_old_dlcs())?;
+        tracing::info!(sync_time_sec = %now.elapsed().as_secs(), "Culling done");
```
bonomat: Removed in fixup

bonomat (commented):
On the full maker db this query took ~60 seconds on my machine. On the second run it took ~10 seconds.

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -1,127 +1,82 @@
 use crate::db::Connection;
 use anyhow::Result;
-use model::EventKind;
 use sqlx::pool::PoolConnection;
 use sqlx::Sqlite;
 
 impl Connection {
+    /// Cull old cfd events
+    ///
+    /// This function removes old `dlc` data from events which are not needed anymore
```
luckysori: 🙃 

```suggestion
    /// Cull DLC data from old events.
    ///
    /// This method modifies the `dlc` field of the JSON blobs stored in the `events` 
    /// table for all the rows which do not correspond to the most recent 
    /// `EventKind::ContractSetupCompleted` or `EventKind::RolloverCompleted` event 
    /// of a CFD.
```

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -1,127 +1,82 @@
...
-
-            event.map(|event| (id, event))
-        }))
-}
+    /// Cull old cfd rollover events which are not needed anymore
+    ///
+    /// We set the `dlc` field of events `RolloverCompleted` to null where the event is not the most
+    /// recent `RolloverCompleted` event.
+    async fn cull_rollover_completed(conn: &mut PoolConnection<Sqlite>) -> Result<()> {
+        let effected_rows = sqlx::query!(
```
luckysori: 🔧 

```suggestion
        let affected_rows = sqlx::query!(
```

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -1,127 +1,82 @@
...
+                        select json_set(events.data, '$.dlc', null)
+                        from events as inner_events
+                        where inner_events.id = EVENTS.id 
+                    )
+                where name = 'RolloverCompleted'
+                  and id not in (
+                    select max(id) from EVENTS as inner_events
+                    where inner_events.cfd_id = EVENTS.cfd_id
+                      and inner_events.name = EVENTS.name
+                );
```
luckysori: I'm sure you've tried it, but why doesn't something like

```suggestion
                where name = 'RolloverCompleted' OR name = 'ContractSetupCompleted'
                  and id not in (
                    select max(id) from EVENTS as inner_events
                    where inner_events.cfd_id = EVENTS.cfd_id
                      and inner_events.name = EVENTS.name
                );
```

work?

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -1,127 +1,82 @@
...
+                        where EVENTS.id = inner_events.id
+                    )
+                where name = 'ContractSetupCompleted'
+                  and cfd_id in (
+                    select distinct cfd_id from EVENTS as inner_events
+                        where inner_events.cfd_id = EVENTS.cfd_id
+                        and inner_events.name = 'RolloverCompleted'
+                    );
+            "#
+        )
```
luckysori: 🔧 We normally use these constants for this.
```suggestion
                where name = $1
                  and cfd_id in (
                    select distinct cfd_id from EVENTS as inner_events
                        where inner_events.cfd_id = EVENTS.cfd_id
                        and inner_events.name = 'RolloverCompleted'
                    );
            "#,
            model::EventKind::CONTRACT_SETUP_COMPLETED_EVENT,
        )
```

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -149,16 +104,60 @@ mod tests {
...
+        let mut conn = db.inner.acquire().await.unwrap();
+        let mut db_tx = conn.begin().await.unwrap();
+        let events = load_cfd_events(&mut db_tx, cfd.id(), 0).await.unwrap();
+        db_tx.commit().await.unwrap();
+
+        assert_eq!(events[0].event, contract_setup_cloned,);
+    }
 
+    #[tokio::test]
+    async fn given_one_rollovers_when_cull_then_first_event_has_none_dlc_and_second_hassome_dlc() {
```
luckysori: 🔧 
```suggestion
    async fn given_one_rollover_when_cull_then_first_event_has_none_dlc_and_second_has_some_dlc() {
```

bonomat:
size optimization: 

before: 

```
5.3G
```

after
```
30M
```

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -149,16 +104,60 @@ mod tests {
...
+        db_tx.commit().await.unwrap();
+
+        assert!(matches!(
+            &events[0].event,
+            EventKind::ContractSetupCompleted { dlc: None, .. },
+        ));
+
+        assert!(matches!(
+            &events[1].event,
+            EventKind::RolloverCompleted { dlc: Some(_), .. },
```
luckysori: 🔧 Daniel suggested to always check against the original here, like we do in the previous test.

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -180,7 +179,16 @@ mod tests {
             EventKind::RolloverCompleted { dlc: None, .. },
         ));
 
-        assert_eq!(events[2].event, most_recent_rollover);
+        assert!(matches!(
+            &events[2].event,
+            EventKind::RolloverCompleted { dlc: Some(_), .. },
```
luckysori: 🔧 Same as above.

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -180,7 +179,16 @@ mod tests {
...
+        ));
+
+        let event0_timestamp = &events[0].timestamp;
+        let event1_timestamp = &events[1].timestamp;
+        let event2_timestamp = &events[2].timestamp;
+
+        assert!(event0_timestamp < event1_timestamp && event1_timestamp < event2_timestamp);
```
luckysori: 🙃 I wouldn't test this here. This is testing a property of `load_cfd_events`: that the events returned are ordered. I would be happy to introduce a test in `db.rs` for this (in a separate commit if possible).

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -224,24 +229,46 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn given_contract_setup_completed_when_cull_old_dlcs_then_dlc_is_some() {
+    async fn given_two_cfds_one_contract_setup_completed_other_rolled_over_both_culled_correctly() {
```
luckysori: 🙃 
```suggestion
    async fn given_cfds_contract_setup_completed_and_cfd_rolled_over_then_both_culled_correctly() {
```

luckysori (approved):
Thanks for improving things! I had one question and a few minor suggestions. But LGTM!

daemon/src/db/cull_old_dlcs.rs
bonomat: Because if we only have `ContractSetupCompleted` but no `RolloverCompleted` yet we want to keep the dlc data.

luckysori: Have you tried it? My crappy SQL brain compiler seems to think that it could work.

maker/src/actor_system.rs
```diff
@@ -83,7 +84,10 @@ where
             + Handler<monitor::MonitorCetFinality>
             + Actor<Stop = ()>,
     {
+        let now = Instant::now();
+        tracing::info!("Culling started");
```
bonomat: and yet again it appears 🙈

daemon/src/db/cull_old_dlcs.rs
bonomat: I tried, but maybe you can come up with a query that works? 

The second query looks slightly different than the first if you look closely: it checks if the `cfd_id` is  in the list of events where there is at least one `RolloverCompletedEvent`

luckysori (approved):
LGTM, thanks!

daemon/src/db/cull_old_dlcs.rs
luckysori: > The second query looks slightly different than the first if you look closely: it checks if the cfd_id is in the list of events where there is at least one RolloverCompletedEvent

True. My idea would be to do something very similar to what you do in the first one, but slightly modified so that it considers all contract-setup/rollover events and only keeps the most recent one (whether it is contract-setup or rollover). But there must be a flaw in how I imagine it working.

> I tried, but maybe you can come up with a query that works?

I invoke the other SQL expert @da-kami if he's curious 😛 And if this already works we can iterate on it later.

PR #2074, opened 2022-05-20
Revert changes to `prepare_db.sh`

luckysori:
With patch b89983111b2e074169e0d4c25df3a9be352cc8a8 we modified the `prepare_db.sh` script in order to be able to define `sqlx` queries under `cfg(test)`.

This was convenient, but it turns out that the changes cause the script to trigger _complete_ recompilation of the project every time we run it.

There may be an alternative, but for the time being we revert this change. The effect is very minor: we move some test queries to a
`sqlx_test_utils` annotated with `#[allow(dead_code)]`.


bonomat (approved):
👍

PR #2075, opened 2022-05-20
Optimize culling query

bonomat:
We don't have to update events which are already null.

### DIFF
--- a/daemon/sqlx-data.json
+++ b/daemon/sqlx-data.json
@@ -1,15 +1,5 @@
 {
   "db": "SQLite",
-  "10e7b0efb777c853f9576d065863bcf2556a00c12dc8f95b3cc852b0f99c6018": {
-    "query": "\n            update EVENTS\n                set\n                    data = (\n                        select json_set(events.data, '$.dlc', null)\n                        from events as inner_events\n                        where inner_events.id = EVENTS.id \n                    )\n                where name = $1\n                  and id not in (\n                    select max(id) from EVENTS as inner_events\n                    where inner_events.cfd_id = EVENTS.cfd_id\n                      and inner_events.name = EVENTS.name\n                );\n            ",
-    "describe": {
-      "columns": [],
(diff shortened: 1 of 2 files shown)


luckysori (approved):
Good point!

da-kami (commented):
Not sure this improvement is necessary - why would you keep a DLC without cets?

da-kami:
> Not sure this improvement is necessary - why would you keep a DLC without cets?

Ah, it's the whole data field....?

bonomat:
> > Not sure this improvement is necessary - why would you keep a DLC without cets?
> 
> Ah, it's the whole data field....?

it's an optimization to not run the query if the field was already removed in a former run.

da-kami:
> > > Not sure this improvement is necessary - why would you keep a DLC without cets?
> > 
> > 
> > Ah, it's the whole data field....?
> 
> it's an optimization to not run the query if the field was already removed in a former run.

What confuses me is that you check against `dlc.cets` - I thought we are removing the complete previous `dlc` not just the `cets`. Maybe I missed that change.

bonomat:
No, Lucas introduced a change which allowed us to remove the CETs: https://github.com/itchysats/itchysats/commit/1520dbcf408ebfc96068edaee5c181606629b7f3
We still keep the remaining info from the DLC.

luckysori:
> No, Lucas introduced a change which allowed us to remove the CETs: [1520dbc](https://github.com/itchysats/itchysats/commit/1520dbcf408ebfc96068edaee5c181606629b7f3) We still keep the remaining info from the DLC.

Hmmm, we actually remove the entirety of the `dlc` field. We never look at the data once we roll over. I think we should check against `dlc` being null.

bonomat:
What I said was bollocks. We indeed remove the full dlc field. 
Since the field `dlc` does not exist on `null` the query still works. But I you are right. Thanks for poitning it out @da-kami

PR #2076, opened 2022-05-20
fixup check if dlc is not null

bonomat:
even though `.dlc.cets` is not null if `.dlc` is not null, correct is to check if `.dlc` is not null.

Followup PR from #2075


luckysori (approved):
Thanks!

PR #2077, opened 2022-05-20
Remove rollover taker ephemeral actor

klochowicz:
Now that rollovers happen over libp2p, we can remove the unused code.

Auto-rollover actor became *much* simpler.


klochowicz:
Marking this explicitly as a draft, as it's dependent on rolling with libp2p :)

It's more of a showcase how much unnecessary complexity goes away with removing ephemeral actors.

da-kami (approved):
Happy to get this in on top - it only touches the taker code that can be removed. This should be save because in the new taker this is dead code. 

I approved it, but we can wait with merging until we have more confidence in libp2p. My confidence level is high, but let's see what testnet thinks :)

Looking forward to removing more!

luckysori (approved):
Happy to remove unused code!

bonomat:
I think this needs some love because it includes a few commits which are already in.

klochowicz:
Rebased on top of master, now that libp2p-rollover is in.

PR #2085, opened 2022-05-23
Reduce culling log spam

bonomat:


klochowicz (approved):

daemon/src/db/cull_old_dlcs.rs
```diff
@@ -47,7 +47,11 @@ impl Connection {
         .await?
         .rows_affected();
 
-        tracing::debug!(%affected_rows,"Culled RolloverCompleted events");
+        if affected_rows > 0 {
+            tracing::debug!(%affected_rows,"Culled RolloverCompleted events");
```
luckysori, line 51: 🙃 Not a big deal, but otherwise it sounds like we're removing the events!

```suggestion
            tracing::debug!(%affected_rows,"Culled DLCs from RolloverCompleted events");
```

luckysori (approved):

PR #2087, opened 2022-05-23
Add missing changelog entry

bonomat:
This commit was added in 0.4.17 and is missing in master


luckysori (approved):
I think you should add something like

```
[Unreleased]: https://github.com/itchysats/itchysats/compare/0.4.17...HEAD
[0.4.16]: https://github.com/itchysats/itchysats/compare/0.4.16...0.4.17
```

at the bottom.

PR #2088, opened 2022-05-23
Constrain unbounded Tasks growth in rollover protocol

klochowicz:
Instead of appending tasks indefinitely, constrain the rollover actor to only
allow one task per order ID at a time.
We implicitly rely on the invariants that we cannot get two decisions for the same rollover
at the same time.

Note: This is not a general solution, but it removes unbounded growth of the vector.

### DIFF
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -40,6 +40,7 @@ type ListenerConnection = (
 /// task whenever we interact with a substream to not block the execution of other connections.
 pub struct Actor {
     tasks: Tasks,
+    protocol_tasks: HashMap<OrderId, Tasks>,
     oracle_pk: schnorrsig::PublicKey,
     get_announcement: Box<dyn MessageChannel<oracle::GetAnnouncement>>,
     n_payouts: usize,
@@ -56,6 +57,7 @@ impl Actor {
     ) -> Self {
         Self {
             tasks: Tasks::default(),
+            protocol_tasks: HashMap::default(),
             oracle_pk,
             get_announcement,
             n_payouts,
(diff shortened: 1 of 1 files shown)


daemon/src/rollover/maker.rs
```diff
@@ -236,6 +239,7 @@ impl Actor {
                 }
             },
         );
+        self.protocol_tasks.insert(order_id, tasks);
```
bonomat, line 242: What happens with the existing `task` if it is being overwritten by another one?

klochowicz: it gets dropped. that's why it's not a general solution, and not safe to roll out in other cases (e.g. ping, where we periodically add them). Here due to our business rules we can't perform a decision twice for the same rollover. Technically, if the future execution would take over an hour, we would still overwrite it - but at that point, it would probably be a good thing to do anyway :-)

da-kami (approved):

daemon/src/rollover/maker.rs
```diff
@@ -249,7 +253,8 @@ impl Actor {
 
         emit_rejected(order_id, &self.executor).await;
 
-        self.tasks.add_fallible(
+        let mut tasks = Tasks::default();
+        tasks.add_fallible(
```
luckysori, line 257: Not an issue imo, but this lets the maker cancel an ongoing rollover by quickly rejecting it after having accepted.

luckysori (approved):
LGTM. We know it's not ideal, but it should be safe.

daemon/src/rollover/maker.rs
bonomat: Thanks for the clarification

bonomat (approved):

daemon/src/rollover/maker.rs
klochowicz: interesting! this might actually be a feature 😁

PR #2089, opened 2022-05-23
Legacy connection establishment logs `peer_id` if available

da-kami:
This is important to see if peers that run into errors over the libp2p connection are able to establish a legacy connection.
If a peer only tries to connect over libp2p it is likely a peer outside our application realm.

### DIFF
--- a/maker/src/connection.rs
+++ b/maker/src/connection.rs
@@ -495,7 +495,11 @@ impl Actor {
             ]))
             .inc();
 
-        tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, "Connection is ready");
+        if let Some(peer_id) = peer_id {
+            tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, %peer_id, "Connection is ready");
+        } else {
+            tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, "Connection is ready");
+        }
     }
 
     async fn handle_listener_failed(&mut self, msg: ListenerFailed, ctx: &mut xtra::Context<Self>) {


luckysori (approved):

maker/src/connection.rs
```diff
@@ -495,7 +495,11 @@ impl Actor {
             ]))
             .inc();
 
-        tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, "Connection is ready");
+        if let Some(peer_id) = peer_id {
+            tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, %peer_id, "Connection is ready");
+        } else {
+            tracing::debug!(taker_id = %identity, taker_addres = %address, %wire_version, %daemon_version, "Connection is ready");
+        }
```
bonomat, line 502: How can someone connect from outside of our application and provide `wire_version` and `daemon_version` to us?

bonomat (commented):
How can someone connect from outside of our application realm and provide `wire_version` and `daemon_version` to us?

bonomat (approved):

da-kami:
> How can someone connect from outside of our application realm and provide `wire_version` and `daemon_version` to us?

This log has nothing to do with the libp2p connection. The log helps to add context between legacy and libp2p connection because like this we know that a peer with a certain peer-id connected over the legacy connection.

There could be a peer that connects only over libp2p but never uses the legacy connection. If that happens that would be a peer "outsides of our application realm", as in, the peer speaks libp2p but not necessarily our protocols. 
Once we remove the legacy connection we will have to deal with this problem different, i.e. if we can't negotiate a protocol we should just drop the connection. 

Maybe a bit more context. We see this log quite often:

```
2022-05-22 23:31:28	{"timestamp":"2022-05-22T13:31:28.372152793Z","level":"DEBUG","fields":{"message":"Outbound ping protocol failed: Timeout in protocol negotiation","peer":"12D3KooWKYrtqmMUA8agwY7qUrnUb23eAyniWkh3ExeU7GEgFbNW"},"target":"xtra_libp2p_ping"}
```

I am not sure if we see this because someone is just using the libp2p connection. I assume that these peers also use the legacy connection, but at the moment I just don't know if they do because there is no association between legacy and libp2p connection in the logs.

bonomat:
Thanks for the clarification.  I was just confused by this part of the commit message
>  If a peer only tries to connect over libp2p it is likely a peer outside our application realm.

I think you added the peerID to the logs so that you can relate `peerID` and `takerID`.

PR #2094, opened 2022-05-23
Let taker chose leverage

### LINKED ISSUE #1739: Let the taker choose leverage 1, 2 and 3

### LINKED ISSUE #1983: Dynamic computation of `initial_funding_fee`, `margin`, `liquidation_price`
At the moment the project assumes that we always have a leverage of 2 for the taker and 1 for the maker. 
This will change once we have selectable leverages.
At the moment, the projection uses the fixed known leverage to compute
- `initial_funding_fee`, 
- `margin`, 
- `liquidation_price`

If the taker can select a leverage in the UI, then these values will need to be calculated dynamically from the UI (or via an additional HTTP endpoint).

bonomat:
Resolves #1739 
Resolves #1983

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- The API for retrieving offers now returns `leverage_details` which is a set of leverages including pre-computed values for margin, liquidation price and initial funding fee.
+- Taker can choose a leverage provided the maker offers a selection of leverage.
+
 ## [0.4.17] - 2022-05-23
 
 - Culling old DLC data from database, i.e. we remove old DLC data which is not needed anymore. This reduces the db size and is more efficient when loading
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -22,7 +22,6 @@ use daemon_tests::Maker;
 use daemon_tests::MakerConfig;
 use daemon_tests::Taker;
 use daemon_tests::TakerConfig;
-use model::calculate_margin;
 use model::olivia;
 use model::Identity;
 use model::Leverage;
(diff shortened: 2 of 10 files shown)


da-kami (commented):
I think the calculation of the margin and liquidation price according to the leverage is still missing. Merging this without the calculation does not make much sense to me.

bonomat:
> I think the calculation of the margin according to the leverage is still missing. Merging this without the calculation does not make much sense to me.

There is a second ticket for this: https://github.com/itchysats/itchysats/issues/1983

da-kami:
> > I think the calculation of the margin according to the leverage is still missing. Merging this without the calculation does not make much sense to me.
> 
> There is a second ticket for this: #1983

Ok, but I still think that activating this in the UI without implementing https://github.com/itchysats/itchysats/issues/1983 does not make sense. If we merge this the user will be able to select it, but the calculation update is missing.

bonomat:
> > > I think the calculation of the margin according to the leverage is still missing. Merging this without the calculation does not make much sense to me.
> > 
> > 
> > There is a second ticket for this: #1983
> 
> Ok, but I still think that activating this in the UI without implementing #1983 does not make sense. If we merge this the user will be able to select it, but the calculation update is missing.

Fair enough, I'll add the other ticket into this PR to avoid merge conflicts.

bonomat:
@da-kami : I've implemented the other ticket as well :)

da-kami (approved):
Great stuff!

I gave it a try locally also playing with allowing `x5` in the maker orders which results in a somewhat funny "bug" where the slider on the taker side allows `4` but that's actually not available. That's probably the problem of using a slider.

image

If we only plan to use 1, 2, 3 for now then I think we can merge, otherwise we should probably somehow fix this. Not sure the slider allows choices with different step intervals.

bonomat:
> Great stuff!
> 
> I gave it a try locally also playing with allowing `x5` in the maker orders which results in a somewhat funny "bug" where the slider on the taker side allows `4` but that's actually not available. That's probably the problem of using a slider.
> 
> image
> 
> If we only plan to use 1, 2, 3 for now then I think we can merge, otherwise we should probably somehow fix this. Not sure the slider allows choices with different step intervals.

damn you were too fast :D , I noticed this bug as well and wanted to remove the `x5` again from the maker before anyone else notices it 😬
I don't think the slider allows steps with different intervals neither, so the alternative would just be to allow leverage 4 aswell

For now, we only want to support 1,2,3 anyway, so I'll remove the option from the maker.

bonomat:
I've removed leverage `5` from the maker's UI because we don't want to support it

bors r+

PR #2096, opened 2022-05-24
Taker sends `environment` as part of `Hello` message

da-kami:
- Introduce `HelloV4` that contains the environment

Set environment according to:

If `ITCHYSATS_ENVIRONMENT` envvar present: parse the value of the envvar.
    - If value is `umbrel` set to `UMBREL`
    - If value is `raspiblitz` set to `RASPIBLITZ`
    - If value is `docker` set to `DOCKER`
    - Else: set to `UNKNOWN`
Else
    - Set to `BINARY`

For previous `Hello` messages when logging we default environment to `LEGACY`.
For the tests we set the environment to `TEST`.

---

TODO (separate PRs / updates):

- Set `ITCHYSATS_ENY` on raspiblitz to `raspiblitz`
- set `ITCHYSATS_ENV` on umbrel to `umbrel`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM gcr.io/distroless/cc
 
+ENV ITCHYSATS_ENV=docker
+
 USER 1000
 
 LABEL "org.opencontainers.image.source"="https://github.com/itchysats/itchysats"
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -19,6 +19,7 @@ use model::OrderId;
 use model::Price;
 use model::Role;
 use model::Usd;
+use parse_display::Display;
 use seed::Identities;
 use std::time::Duration;
 use time::ext::NumericalDuration;
@@ -116,6 +117,7 @@ where
         projection_actor: Address<projection::Actor>,
         maker_identity: Identity,
         maker_multiaddr: Multiaddr,
+        environment: Environment,
     ) -> Result<Self>
     where
         M: Handler<monitor::StartMonitoring>
@@ -201,6 +203,7 @@ where
                     identity.peer_id(),
                     maker_heartbeat_interval,
                     connect_timeout,
+                    environment,
                 )),
         );
 
@@ -335,5 +338,27 @@ where
                 fee: Some(fee_rate),
             })
             .await?
+    }
+}
+
+#[derive(Debug, Copy, Clone, Display)]
+pub enum Environment {
+    Umbrel,
+    RaspiBlitz,
+    Docker,
+    Binary,
+    Test,
+    Legacy,
+    Unknown,
+}
+
+impl Environment {
+    pub fn from_str_or_unknown(s: &str) -> Environment {
+        match s {
+            "umbrel" => Environment::Umbrel,
+            "raspiblitz" => Environment::RaspiBlitz,
+            "docker" => Environment::Docker,
+            _ => Environment::Unknown,
+        }
     }
 }
--- a/daemon/src/wire.rs
+++ b/daemon/src/wire.rs
@@ -99,6 +99,12 @@ pub enum TakerToMaker {
         daemon_version: String,
         peer_id: PeerId,
     },
+    HelloV4 {
+        proposed_wire_version: Version,
+        daemon_version: String,
+        peer_id: PeerId,
+        environment: Environment,
+    },
     /// This is deprecated, use `TakeOrderWithLeverage` instead
     #[serde(rename = "TakeOrder")]
     DeprecatedTakeOrder {
@@ -168,6 +174,7 @@ impl TakerToMaker {
             TakerToMaker::Hello(_) => "TakerToMaker::Hello",
             TakerToMaker::HelloV2 { .. } => "TakerToMaker::HelloV2",
             TakerToMaker::HelloV3 { .. } => "TakerToMaker::HelloV3",
+            TakerToMaker::HelloV4 { .. } => "TakerToMaker::HelloV4",
             TakerToMaker::Unknown => "TakerToMaker::Unknown",
         }
     }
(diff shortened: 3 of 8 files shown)


daemon/src/lib.rs
```diff
@@ -346,3 +349,43 @@ where
...
+            "docker" => Environment::DOCKER,
+            _ => Environment::UNKNOWN,
+        };
+
+        Ok(environment)
+    }
+}
+
+impl Environment {
+    pub fn name(&self) -> &'static str {
```
klochowicz: why not just a Display trait impl? (whilst at there, we could automatically derive it instead of manually writing it!)

taker/src/main.rs
```diff
@@ -340,6 +341,13 @@ async fn main() -> Result<()> {
         taker_peer_id: identities.libp2p.public().to_peer_id().to_string(),
     };
 
+    let environment = match option_env!("ITCHYSATS_ENVIRONMENT") {
+        Some(environment) => environment
+            .parse()
+            .expect("String that is not processable to default to unknown environment"),
```
klochowicz: I am having a hard time processing this error message, would you mind rephrasing it?. also:  shouldn't this rather fallback to unknown env instead of panicking?

daemon/src/wire.rs
```diff
@@ -664,6 +671,44 @@ impl From<CfdTransactions> for RolloverMsg1 {
     }
 }
 
+#[derive(Debug, Copy, Clone, Serialize, Deserialize)]
+pub enum Environment {
```
klochowicz, line 675: I probably wouldn't bother with the second enum - the risk of someone setting `Legacy` in the envvar is minimal.

Unknown could have been handled by `Option<Environment>`, if it wasn't set it must be unknown (that gets rid of awkwardness when parsing and arguably is more natural, as it's a first-class way to encode that it's unknown).

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM gcr.io/distroless/cc
 
+ENV ITCHYSATS_ENVIRONMENT=docker
```
klochowicz: this would be slightly more conventional:

```suggestion
ENV ITCHYSATS_ENV=docker
```

taker/src/main.rs
luckysori: 🔧 It would be safer if you don't use the fallible method `parse`. Instead of implementing `FromStr` we could introduce a custom method that takes an `&str` and is infallible.

taker/src/main.rs
```diff
@@ -340,6 +341,13 @@ async fn main() -> Result<()> {
         taker_peer_id: identities.libp2p.public().to_peer_id().to_string(),
     };
 
+    let environment = match option_env!("ITCHYSATS_ENVIRONMENT") {
```
klochowicz:
```suggestion
    let environment = match option_env!("ITCHYSATS_ENV") {
```

daemon/src/lib.rs
```diff
@@ -346,3 +349,43 @@ where
             .await?
     }
 }
+
+#[derive(Debug, Copy, Clone)]
+pub enum Environment {
+    UMBREL,
+    RASPIBLITZ,
+    DOCKER,
+    BINARY,
```
klochowicz: would you be open to changing `BINARY` to `DESKTOP_APP`or `EXECUTABLE`  or something? it might be my personal awkwardness, but I never refer to the apps I run as binaries. I get that this might be subjective, so please feel free to ignore this comment :)

klochowicz (approved):
thank you for tackling this  - it will be immensely useful to have info about it.

wrote some nits, but nothing blocking

daemon/src/lib.rs
```diff
@@ -346,3 +349,43 @@ where
...
+    TEST,
+    LEGACY,
+    UNKNOWN,
```
luckysori: 🔧 Enum variants should be `UpperCamelCase`.

daemon/src/lib.rs
```diff
@@ -346,3 +349,43 @@ where
...
+            Environment::UMBREL => "Umbrel",
+            Environment::RASPIBLITZ => "RaspiBlitz",
+            Environment::DOCKER => "Docker",
+            Environment::BINARY => "Binary",
+            Environment::TEST => "Test",
+            Environment::LEGACY => "Legacy",
+            Environment::UNKNOWN => "Unknown",
+        }
+    }
+}
```
luckysori: ❓ Why not implement `Display`?

luckysori (approved):
LGTM, just a couple of suggestions.

taker/src/main.rs
da-kami: It does fallback to `Unknown` on the inside. `FromString` is implemented infallible. I an change it to fallback here as well, I don't mind the redundancy.

daemon/src/wire.rs
da-kami: Setting `Legacy` in the envvar will still default to `UNKNOWN`. But that's not the point. We serialize / deserialize over the wire while we perform other operations (e.g. parse) with `daemon::Environment`. I think we should clearly separate `wire` types to avoid breaking them.

klochowicz: fair enough!

taker/src/main.rs
```diff
@@ -340,6 +341,11 @@ async fn main() -> Result<()> {
         taker_peer_id: identities.libp2p.public().to_peer_id().to_string(),
     };
 
+    let environment = match option_env!("ITCHYSATS_ENV") {
+        Some(environment) => Environment::from_str_or_fallback(environment),
```
luckysori: 🙃 Thanks! Maybe this name would be a little more natural.

```suggestion
        Some(environment) => Environment::from_str_or_unknown(environment),
```

luckysori (approved):

bonomat (approved):
Well done.
We could add a metric for this which would allow us to easily track it on grafana

da-kami:
> Well done. We could add a metric for this which would allow us to easily track it on grafana

Happy to do that in a follow-up.

PR #2100, opened 2022-05-24
Let `xtras::spawner::Actor` manage ping tasks

### LINKED ISSUE #2051: `Tasks` contains an ever growing vec of `RemoteHandle`
When we `add` (or `add_fallible`) a task the `RemoteHandle` remains in the internal `Vec` forever, it will never be removed. 
Given that `RemoteHandle` is just a pointer this should not be expensive, but we should investigate how fast this *can* grow.

In both the legacy connection actor(s) and the libp2p impl  we add tasks upon connection handling. 
I.e. for ping we add a task for each peer every 5 seconds. This can multiply up quite fast over time.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2051.

It does so by spawning the ephemeral ping tasks via the `xtras::spawner::Actor` (as opposed to using `Tasks` directly). This
actor will clean up the task handles internally when they are no longer needed.

We are still able to tie the lifetime of theses tasks to the lifetime of the `ping::Actor` because:

1. Actors are dropped when all their (strong) `xtra::Address`es are dropped.
2. We store the only (strong) `xtra::Address` to the `spawner::Actor` inside of `ping::Actor`.

This means that whenever `ping::Actor` is dropped, the last `xtra::Address<spawner::Actor, Strong>` is dropped, stopping all the ongoing tasks that were spawned through it.

### DIFF
--- a/tokio-tasks/src/task_map.rs
+++ b/tokio-tasks/src/task_map.rs
@@ -1,0 +1,69 @@
+use crate::future_ext::FutureExt;
+use futures::future::RemoteHandle;
+use futures::Future;
+use std::collections::HashMap;
+use std::hash;
+
+/// Struct controlling the lifetime of the async tasks, such as
+/// running actors and periodic notifications. If it gets dropped, all
+/// tasks are cancelled.
+///
+/// Each task is associated with a key, so that it can be removed from
+/// the internal `HashMap` whenever consumers want to stop tracking
+/// the corresponding task. By removing it from the `HashMap` we free
+/// up space.
+pub struct TaskMap<K>(HashMap<K, RemoteHandle<()>>);
+
+impl<K> TaskMap<K>
+where
+    K: Eq + hash::Hash,
+{
+    /// Spawn the task on the runtime and remember the handle.
+    ///
+    /// The task will be stopped if this instance of [`TaskMap`] goes
+    /// out of scope.
+    pub fn add(&mut self, key: K, f: impl Future<Output = ()> + Send + 'static) {
+        let handle = f.spawn_with_handle();
+        self.0.insert(key, handle);
+    }
+
+    /// Spawn a fallible task on the runtime and remember the handle.
+    ///
+    /// The task will be stopped if this instance of [`TaskMap`] goes
+    /// out of scope. If the task fails, the `err_handler` will be
+    /// invoked.
+    pub fn add_fallible<E, EF>(
+        &mut self,
+        key: K,
+        f: impl Future<Output = Result<(), E>> + Send + 'static,
+        err_handler: impl FnOnce(E) -> EF + Send + 'static,
+    ) where
+        E: Send + 'static,
+        EF: Future<Output = ()> + Send + 'static,
+    {
+        let fut = async move {
+            match f.await {
+                Ok(()) => {}
+                Err(err) => err_handler(err).await,
+            }
+        };
+
+        let handle = fut.spawn_with_handle();
+        self.0.insert(key, handle);
+    }
+
+    /// Remove the handle of a task from the internal `HashMap`.
+    ///
+    /// After removing a task which is present in the `HashMap`, the
+    /// `RemoteHandle` will be dropped, effectively ending the task if
+    /// it was still ongoing.
+    pub fn remove(&mut self, key: &K) {
+        self.0.remove(key);
+    }
+}
+
+impl<K> Default for TaskMap<K> {
+    fn default() -> Self {
+        Self(HashMap::new())
+    }
+}
--- a/xtras/src/spawner.rs
+++ b/xtras/src/spawner.rs
@@ -1,0 +1,128 @@
+use crate::SendAsyncSafe;
+use async_trait::async_trait;
+use futures::Future;
+use tokio_tasks::TaskMap;
+use uuid::Uuid;
+use xtra::Context;
+use xtra_productivity::xtra_productivity;
+
+/// A spawner actor controls the lifetime of tasks it is instructed to
+/// spawn. If all (strong) `Address`es to this `xtra::Actor` are
+/// dropped, all the tasks it is tracking will be cancelled.
+///
+/// Furthermore, a spawner actor will learn about any spawned task
+/// that has finished, dropping the corresponding task handle and
+/// freeing up memory.
+pub struct Actor {
+    tasks: TaskMap<Uuid>,
+}
+
+impl Actor {
+    pub fn new() -> Self {
+        Self {
+            tasks: TaskMap::default(),
+        }
+    }
+}
+
+#[xtra_productivity]
+impl<T> Actor
+where
+    T: Future<Output = ()> + Send + 'static,
+{
+    fn handle<T>(&mut self, msg: Spawn<T>, ctx: &mut Context<Self>)
+    where
+        T: Future<Output = ()> + Send + 'static,
+    {
+        let this = ctx.address().expect("we are alive");
+        let id = Uuid::new_v4();
+
+        self.tasks.add(id, async move {
+            msg.task.await;
+            if let Err(e) = this.send_async_safe(Finished(id)).await {
+                tracing::warn!("Failed to tell spawner about finished task: {e:#}");
+            };
+        })
+    }
+}
+
+#[async_trait]
+impl<T, EH, EF, E> xtra::Handler<SpawnFallible<T, EH>> for Actor
+where
+    T: Future<Output = Result<(), E>> + Send + 'static,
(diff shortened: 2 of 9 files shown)


klochowicz:
I don't think that the additional layer of abstraction is helpful (the Spawner). It has somewhat cumbersome API and just hides away the fact that things are being spawned in separate futures. 

The TaskMap could be used directly, with less overhead and same benefits.

That said, thanks for adding TaskMap!

tokio-tasks/src/task_map.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// Struct controlling the lifetime of the async tasks, such as
+/// running actors and periodic notifications. If it gets dropped, all
+/// tasks are cancelled.
+///
+/// Each task is associated with a key, so that it can be removed from
+/// the internal `HashMap` whenever consumers want to stop tracking
+/// the corresponding task. By removing it from the `HashMap` we free
+/// up space.
+pub struct TaskMap<K>(HashMap<K, RemoteHandle<()>>);
```
klochowicz, line 15: Could we use this one directly instead?

luckysori: I think it would be dramatically less scalable because we would have to block the processing of `Ping` messages based on `Finished`.

At this stage I see actors that handle messages that access different data and are completely independent as an antipattern.

klochowicz: Oh, this *was* the added value! I missed the crucial bit! Indeed, the `Spawner` makes sense - but then, do we really need to have the `TaskMap` and the keyed usage?

luckysori: The way I saw it, we need to send ourselves a message informing of a task having finished and we need to be able to know which task handle to drop based on that.

What do you propose other than something like `TaskMap`?

klochowicz: hm. coming at it again, the fact that `Spawner` uses a TaskMap is its implementation detail. We could generate unique id of a future when it's started, instead of passing it in via the message via the public API. This way it would be easier to use and also misuse-resistant (no arbitrary replacement of the same keys, dropping a handle mid-way through.

luckysori: That's a good idea, I like that!

tokio-tasks/src/task_map.rs
```diff
@@ -0,0 +1,70 @@
...
+        let handle = f.spawn_with_handle();
+        self.0.insert(key, handle);
+    }
+
+    /// Spawn a fallible task on the runtime and remember the handle.
+    ///
+    /// The task will be stopped if this instance of [`TaskMap`] goes
+    /// out of scope. If the task fails, the `err_handler` will be
+    /// invoked.
+    pub fn add_fallible<E, EF>(
```
da-kami, line 35:
```suggestion
    pub fn insert_fallible<E, EF>(
```

🙃 Might be more idemoatic

xtras/src/spawner.rs
```diff
@@ -0,0 +1,128 @@
...
+use xtra_productivity::xtra_productivity;
+
+/// A spawner actor controls the lifetime of tasks it is instructed to
+/// spawn. If all (strong) `Address`es to this `xtra::Actor` are
+/// dropped, all the tasks it is tracking will be cancelled.
+///
+/// Furthermore, a spawner actor will learn about any spawned task
+/// that has finished, dropping the corresponding task handle and
+/// freeing up memory.
+pub struct Actor {
```
da-kami, line 16: I think that his pattern might actually be good enough - not sure we need the `TaskMap` if we have this. 
I imagine that we can have multiple of this `spawner::Actor`, we could just scale it up with multiple actors. 
What do you think? Is it worth to have both?

da-kami (approved):
I like it!

klochowicz:
thanks for incorporating these changes,  it looks much nicer! I suspect we could get away without `xtra::Actor` abstraction, I'll try to give it a stab in the morning! (same idea that you came up with  but instead of Actor just modified version of `Tasks` that wraps futures with a send from `mpsc` channel back to some `tokio::select!` macro. 
Main motivation is that it would make the starting easier (no need for `Option<>`) and it would remove the somewhat antipattern of routinely dispatching things back to ourselves - with a lot of tasks, the actor mailbox could become a bottleneck.

I'll write a POC in the  afternoon, and will link it here!

tokio-tasks/src/task_map.rs
klochowicz: insert is an implementation detail, IMHO we shouldn't leak out how we internally deal with futures that finish 😇

xtras/src/spawner.rs
luckysori: How do you see this actor working without the underlying `TaskMap`?

tokio-tasks/src/task_map.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /// Remove the handle of a task from the internal `HashMap`.
+    ///
+    /// After removing a task which is present in the `HashMap`, the
+    /// `RemoteHandle` will be dropped, effectively ending the task if
+    /// it was still ongoing.
+    pub fn remove(&mut self, key: &K) {
+        self.0.remove(key);
+        self.0.shrink_to_fit(); // reap the benefits of removing the handle
```
klochowicz: this line will cause us to re-allocating all the time, moving the memory to a new size that's just the size, then requesting an alloc straight away when get a new task, altogether leading to performance implications that are not worth the call in the first place. I think we could do without it - the size should not grow without end, with periodic tasks we should stabilise with N tasks per taker, rendering these constant re-allocations useless.

luckysori: Makes sense, we don't need to resize it constantly.

We could end up in a situation where we end up adding more often than we are removing, meaning that the capacity of the `HashMap` will keep growing over time. We could call `shrink_to_fit` periodically, but perhaps it's not worth it.

luckysori: I've removed the line with the last force-push.

tokio-tasks/src/task_map.rs
da-kami: But the fact that we replace what was inside is *not* an implementation detail! That is what is indicated by insert. Add would mean that it is added.

xtras/src/spawner.rs
da-kami: Don't know what I was thinking when I wrote this. Somehow I thought we are using `TaskMap` directly in the actor, but what we are doing *is* using this `Actor` to manage tasks, and `TaskMap` is only used by this actor. 
Sorry for the confusion, you can ignore my comment!

tokio-tasks/src/task_map.rs
luckysori: Sorry I missed this discussion.

I chose to use the same names as for `Tasks` because I obviously copy-pasted 99% of this code :rofl:

I think it' s an important comment though, because it highlights something that I hadn't considered. The point of introducing this type is to allow us to `remove` tasks that have finished, not to overwrite tasks that might still be ongoing! With the underlying `HashMap` it seemed natural to allow overwriting (current state of the PR), but we could decide that `add` and `add_fallible` should do nothing if the corresponding key is present in the map. What do you think, @klochowicz @da-kami?

The naming discussion is interesting too. I do think that we don't need to reveal the fact that we use a `HashMap` internally by using the exact same terminology, but it may indicate that we are overwriting in case of a collision, making the API easier to understand. Maybe we can find a name that is both descriptive and not too influenced by the underlying type.

da-kami: I think we should consider *not* allowing overwrites. I think we should ensure that a task finishes (we can add auto-timeouts at some point if we want, but that's a different feature) and not just kill a task.

So yeah, I think we should guard that we only add if the ID is not present - and then `add` also makes sense.

da-kami (approved):
I think we can still action the guard of not replacing and then get this onto the testnet setup :)

klochowicz (approved):
I suspect that there's a potential to simplify this a bit, but let's get this in as it solves the problem ! :)

tokio-tasks/src/task_map.rs
luckysori: As discussed with @klochowicz, this PR doesn't actually allow ephemeral tasks to be overwritten.

The `TaskMap` abstraction does allow the user to overwrite tasks, but because the `spawner::Actor` generates a _unique_ ID for every task, tasks are never actually overwritten.

Let's move forward with this PR and we can come back to the naming discussion later (or not if @klochowicz's alternative solution is adopted eventually).

PR #2101, opened 2022-05-24
Try to fallback to the legacy rollover

da-kami:
This makes this code path more resilient between different maker and taker versions.
The decision of the maker dispatching `accept` and `reject` is based on the `peer_id` being set in the CFD.
Given that we have a version released where the peer_id is set, but the libp2p protocol is not available we ensure that the maker can still fallback to try with the legacy protocol.
We are not expecting to see this message in the logs, but this adds resilience in case we have to rollback changes or takers trying different versions.


luckysori (approved):
LGTM.

PR #2102, opened 2022-05-24
Depend on two versions of maia

klochowicz:
With the advent of libp2p-based protocols, we have the opportunity to clearly
demarcate older and newer takers, making it possible to roll out recent maia bugfixes.

After this change:
- libp2p-based protocols will use maia v0.2.0
- legacy networking will use maia v0.1.0 (`maia_deprecated`), which matches maia
    version shipped with legacy takers

The contents of setup_contract.rs (protocol logic) has been copied wholesale to
`setup_contract_deprecated.rs` with the expectation that the whole file will be
deleted when we decide to drop support for takers using legacy networking layer.

Co-authored-by: Philipp Hoenisch <philipp@hoenisch.at>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,6 @@ resolver = "2"
 
 [patch.crates-io]
 xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
-maia = { git = "https://github.com/comit-network/maia" } # Unreleased
-maia-core = { git = "https://github.com/comit-network/maia" } # Unreleased
+maia = { git = "https://github.com/comit-network/maia", rev = "a963084189e1598ca622abe206a78015ab6b8466" } # Unreleased
+maia-core = { git = "https://github.com/comit-network/maia", rev = "e419cb2cf19bed6ec53eaa3672408a6205a9b705", package = "maia-core" } # Unreleased version, pinned before maia's breaking change
 xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity" } # Unreleased
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1866,7 +1866,10 @@ impl Dlc {
 
             (outpoint, amount)
         };
-        let (tx, sighash) = maia::close_transaction(
+        // In order to preserve backwards compatibility, we are using maia
+        // v0.1.0 always (this code is called from legacy collab settlement protocol)
+        // TODO: Use maia v0.2.0 with libp2p-based collab close
+        let (tx, sighash) = maia_deprecated::close_transaction(
             lock_desc,
             lock_outpoint,
             lock_amount,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -816,7 +816,8 @@ dependencies = [
  "libp2p-core",
  "libp2p-noise",
  "libp2p-tcp",
- "maia",
+ "maia 0.1.0",
+ "maia 0.2.0",
  "maia-core",
  "model",
  "parse-display",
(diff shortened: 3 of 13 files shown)


klochowicz:
depends on: https://github.com/comit-network/maia/pull/36

Cargo.toml
```diff
@@ -21,6 +21,7 @@ resolver = "2"
 
 [patch.crates-io]
 xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
-maia = { git = "https://github.com/comit-network/maia" } # Unreleased
-maia-core = { git = "https://github.com/comit-network/maia" } # Unreleased
+maia = { git = "https://github.com/comit-network/maia", rev = "cf41b907e6eb68245dcb7c561ec7d4e6bcc4ed19" } # FIXME: point back to master after https://github.com/comit-network/maia/pull/36 gets merged
```
klochowicz: this will need to be adjusted after maia changes get upstreamed.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ resolver = "2"
...
+
+maia-core = { git = "https://github.com/comit-network/maia", rev = "e419cb2cf19bed6ec53eaa3672408a6205a9b705", package = "maia-core" } # Unreleased version, pinned before maia's breaking change
```
da-kami, line 25: 🙃 `Unreleased version` this will get outdated as soon as we release. Not sure this helps. 

We can assume that `0.4.19` will be the release. Maybe just hardcode the version here...?

da-kami (commented):
I am a bit confused by the change. 
Where in the legacy code is the decision being made if the old or the new `contract_setup` is being called? How does the maker distinguish between the old and the new code?

Cargo.toml
klochowicz: the "Unreleased" comment refers to `maia`, not itchysats! (and it was there before, I just extended it)

da-kami (approved):

da-kami:
> I am a bit confused by the change. Where in the legacy code is the decision being made if the old or the new `contract_setup` is being called? How does the maker distinguish between the old and the new code?

To answer it myself: The idea is that the upgrade in the protocol is *only* active on the libp2p protocol. This way this is not a breaking change. because if both taker and maker upgrade they user libp2p and maia 2.0, if the taker does not upgrade then the legacy protocol and maia 1.0 is used.

model/src/cfd.rs
```diff
@@ -1866,7 +1866,7 @@ impl Dlc {
 
             (outpoint, amount)
         };
-        let (tx, sighash) = maia::close_transaction(
+        let (tx, sighash) = maia_deprecated::close_transaction(
```
luckysori, line 1872: ❓ Can you remind me why we want to use the buggy version unconditionally? And maybe add the explanation explicitly to the commit message if possible.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ resolver = "2"
 
 [patch.crates-io]
 xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
-maia = { git = "https://github.com/comit-network/maia" } # Unreleased
-maia-core = { git = "https://github.com/comit-network/maia" } # Unreleased
+maia = { git = "https://github.com/comit-network/maia", rev = "cf41b907e6eb68245dcb7c561ec7d4e6bcc4ed19" } # FIXME: point back to master after https://github.com/comit-network/maia/pull/36 gets merged
+
```
luckysori: 🔧 Unwanted newline?

luckysori (approved):
LGTM.

When you fix the tests, could you apply the changes to the first commit?

model/src/cfd.rs
klochowicz: we are not rolling out collab close on libp2p just yet, and this code has changed quite a bit on the branch due to the fact that we've been redesigning the protocol so that both parties publish.
If we end up using the same function, then we'll need to add a condition here, yes. I figured it would be easier to assess on the collab close branch (and for now the crucial bit is preserving the backwards compatibility).

I will put this in the comment here and also in the commit message

PR #2110, opened 2022-05-25
Service not available instead of general HTTP error

da-kami:
Upon startup a `503` indicates in a better way that the service is just not available (yet).
This error can *only* happen on startup, because that is the only time the feed contains `None`.
We only initialize the feed with `None` but always send `Some`.

We could even go further and include a `Retry-After` header, but I did not implement that yet.

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -262,7 +262,7 @@ pub async fn get_cfds<'r>(
 
     match cfds {
         Some(cfds) => Ok(Json(cfds)),
-        None => Err(HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
LGTM!

bonomat (approved):
Fun fact: https://github.com/itchysats/itchysats/pull/1717#discussion_r837974880

👍

klochowicz:
> Fun fact: [#1717 (comment)](https://github.com/itchysats/itchysats/pull/1717#discussion_r837974880)
> 
> 👍

oops, sorry, I don't know why I didn't change it then 🥺

bonomat:
> oops, sorry, I don't know why I didn't change it then 🥺

It wasn't your PR ;)

PR #2111, opened 2022-05-25
Make `now` explicit when calculating rollover hours

da-kami:
In the production code we make it more explicit that we are passing in `now`.
Being explicit can help us when changing this code - if we ever use `now` for derive another value in the same code-path we have to ensure to use the same now.

Test execution was flaky because we used different `now` in the same context.
So, when executing a test right before the full hour we might have diverging `now` values within a test resulting in unwanted behaviour.
In order to make to test this code deterministically we have to make now configurable.

We opt for introducing a function `common_time_boundaries` that returns an event at the full hour, one second after and one second before the full hour.
This should be sufficient.

Note that the proptest appears to have been wrong, we cannot do `+60` if we assume `now` to be the same.
We should not base this test on the fact that we have two different now values where one is slightly newer than the other one, for fixed values the boundary has to be `+59`.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -901,13 +901,14 @@ impl Cfd {
             bail!("Can only accept proposal as a maker");
         }
 
-        let candidate_event_id =
-            olivia::next_announcement_after(OffsetDateTime::now_utc() + self.settlement_interval);
+        let now = OffsetDateTime::now_utc();
+
+        let candidate_event_id = olivia::next_announcement_after(now + self.settlement_interval);
         let hours_to_charge = match version {
             rollover::Version::V1 => 1,
-            rollover::Version::V2 => self.hours_to_extend_in_rollover()?,
+            rollover::Version::V2 => self.hours_to_extend_in_rollover(now)?,
             rollover::Version::V3 => {
-                self.hours_to_extend_in_rollover_based_on_event(candidate_event_id)?
+                self.hours_to_extend_in_rollover_based_on_event(candidate_event_id, now)?
             }
         };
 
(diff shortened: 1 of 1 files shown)


model/src/cfd.rs
```diff
@@ -3131,180 +3134,195 @@ mod tests {
 
     #[test]
     fn given_current_settlement_in_12_hours_and_candidate_in_19_then_7_hour_extension() {
-        let current_event_id =
-            BitMexPriceEventId::with_20_digits(OffsetDateTime::now_utc() + 12.hours());
-        let candidate_event_id =
-            BitMexPriceEventId::with_20_digits(OffsetDateTime::now_utc() + 19.hours());
+        for now in common_time_boundaries() {
```
bonomat, line 3137: I'm personally not a fan of having loops in tests. 
I do prefer dedicated tests per values because if one of the asserts fails, you wouldn't know immediately which one it is. 

If you want to avoid code duplication you can extract parts into an own function and just pass in now. 

Not a blocking comment.

bonomat (approved):

model/src/cfd.rs
da-kami: I totally agree - I went for this as a "quick patch". I don't assume these tests will change much, but I wanted to make sure we test the boundaries. 

> If you want to avoid code duplication you can extract parts into an own function and just pass in now. 

I can do that, but the assert would still have to be by test to have value. 
I think the proper way forward might be configuring proptests for these.

I can add more tests, but I am not happy with this ever growing file 😅
I might just write a ticket and come back to this after some more thought - will decide tomorrow :)

da-kami: As solution for now I added `now` to be printed in the assert so we know with what value a test failed if it does.

Additionally I added https://github.com/itchysats/itchysats/issues/2116 as follow up

PR #2122, opened 2022-05-26
No decision based on `peer_id` in maker rollover decision making

da-kami:
The maker cannot base his decision on the CFD's peer_id being set, because previously opened positions will not contain the peer_id and the maker cannot default it.
The idea was that the maker would use the `peer_id` known in the connection to decide, but accept/reject are completely independent of the connection.

To keep this logic simple and avoid passing on state to the cfd actor we just try to dispatch to libp2p and if that fails (because there is no pending rollover) we fallback to the legacy connection.
This removes the need of load the CFD which is nice.

### DIFF
--- a/maker/src/cfd.rs
+++ b/maker/src/cfd.rs
@@ -529,31 +529,45 @@ impl<O, T, W> Actor<O, T, W> {
 
         let order_id = msg.order_id;
 
-        let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
+        // We try to dispatch to libp2p rollover first
+        // Using send here is fine because we dispatch to a task internally
+        match self
+            .libp2p_rollover
+            .send(daemon::rollover::maker::Accept {
+                order_id,
+                tx_fee_rate: current_offers.tx_fee_rate,
+                long_funding_rate: current_offers.funding_rate_long,
+                short_funding_rate: current_offers.funding_rate_short,
+            })
+            .await
+        {
+            // Return early if dispatch to libp2p rollover worked
+            Ok(Ok(())) => return Ok(()),
+            Ok(Err(error)) => {
+                let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
+                if can_use_libp2p(&cfd) {
+                    tracing::error!("Failed to accept rollover for libp2p: {error:#}");
+                    return Err(error);
+                }
 
-        if can_use_libp2p(&cfd) {
-            // Using send here is fine because we dispatch to a task internally
-            match self
-                .libp2p_rollover
-                .send(daemon::rollover::maker::Accept {
-                    order_id,
-                    tx_fee_rate: current_offers.tx_fee_rate,
-                    long_funding_rate: current_offers.funding_rate_long,
-                    short_funding_rate: current_offers.funding_rate_short,
-                })
-                .await
-            {
-                // Return early if dispatch to libp2p rollover worked
-                Ok(Ok(())) => return Ok(()),
-                Ok(Err(error)) => {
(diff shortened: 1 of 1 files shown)


klochowicz (commented):
One drawback of this approach is that if for whatever reason accepting/rejecting won't work with a new taker (that uses libp2p), we'll try to use the old protocol with him, leading to more errors, as it cannot handle it. :/  (fortunately, it will only produce errors on the taker side, not a panic!)

I'm tempted to keep the `can_use_libp2p()` so that we know when we should *not* fallback (if we have `PeerId` stored).

The other alternative worth exploring would be keeping the old rollovers roll over using old protocol, that's perfectly reasonable too, as we will be supporting it for the foreseeable future.

da-kami:
I adapted the solution to log an `error` and `exit` in case we failed to use libp2p rollover and know the `peer_id`.
That we we make sure to catch these errors.

daemon/src/libp2p_utils.rs
```diff
@@ -48,11 +48,3 @@ pub fn create_listen_tcp_multiaddr(ip: &IpAddr, port: u16) -> Result<Multiaddr>
 pub fn libp2p_socket_from_legacy_networking(legacy_addr: &SocketAddr) -> SocketAddr {
     SocketAddr::new(legacy_addr.ip(), legacy_addr.port() + 1)
 }
-
-/// Determine whether to use libp2p or fallback to a legacy protocol
-pub fn can_use_libp2p(cfd: &model::Cfd) -> bool {
```
klochowicz: I'd prefer if you kept the function and used it to do the early exit

maker/src/cfd.rs
```diff
@@ -529,31 +529,38 @@ impl<O, T, W> Actor<O, T, W> {
...
+                if can_use_libp2p(&cfd) {
+                    tracing::error!("Failed to accept rollover for libp2p: {error:#}");
+                    return Err(error);
                 }
+
+                tracing::debug!("Try fallback to legacy rollover because unable to handle accept via libp2p: {error:#}");
+            }
+            Err(error) => {
+                // we should never see this given that the libp2p rollover actor is always running
+                tracing::error!("Try fallback to legacy rollover because unable to dispatch accept to libp2p rollover actor: {error:#}");
```
klochowicz, line 565: I would just return an error here. We should not be trying to use a different protocol because maker has errors; this will just incur errors on the taker because of maker's misbehavior.

klochowicz (approved):
I guess it's good for now. It will simplify itself when we get rid of the legacy connection anyway...

bonomat (commented):
Could we fix this differently if we auto-accept the rollover request if a request comes in via libp2p and keep the old mode of accepting using the legacy connection?

maker/src/cfd.rs
da-kami: Given that it might be a legacy scenario we have to fallback here.
We cannot just not rollover legacy requests because libp2p is broken.

But with the additional check for `peer_id` this should be fine.

da-kami:
> Could we fix this differently if we auto-accept the rollover request if a request comes in via libp2p and keep the old mode of accepting using the legacy connection?

Potentially, but what would be the benefit?
I know this code looks somewhat messy, but the decision logic is still quite straight forward and it should only be temporarily. I would not invest time before we have more stats on this working properly 🤷‍♂️

bonomat (approved):

bonomat:
> > Could we fix this differently if we auto-accept the rollover request if a request comes in via libp2p and keep the old mode of accepting using the legacy connection?
> 
> Potentially, but what would be the benefit? I know this code looks somewhat messy, but the decision logic is still quite straight forward and it should only be temporarily. I would not invest time before we have more stats on this working properly 🤷‍♂️

Fair enough. Happy to roll with this.

PR #2125, opened 2022-05-27
Maker wallet address cache size configurable

da-kami:
This allows us to experiment with different cache sizes and overcome problems if a hardcoded address cache is not sufficient anymore or causes problems with public nodes.
Default is set to `500`. This is a change to the previously hardcoded cache of `1000` addresses. We use `1000` to be "on the safe side" for longer, but given that it is configurable now we don't have to be this exhaustive.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -89,6 +89,7 @@ pub struct Actor<B> {
     used_utxos: LockedUtxos,
     tasks: Tasks,
     sender: watch::Sender<Option<WalletInfo>>,
+    address_cache_size: u32,
 }
 
 #[derive(thiserror::Error, Debug, Clone, Copy)]
@@ -99,6 +100,7 @@ impl Actor<ElectrumBlockchain> {
     pub fn new(
         electrum_rpc_url: &str,
         ext_priv_key: ExtendedPrivKey,
+        address_cache_size: u32,
     ) -> Result<(Self, watch::Receiver<Option<WalletInfo>>)> {
         let client = bdk::electrum_client::Client::new(electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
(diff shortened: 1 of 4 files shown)


klochowicz (approved):
this is truly a parameter that should be tweakable. 👍 

Could you please change the PR & commit description to remove the "Maker"? Perhaps you wanted to write "Make"? (the PR adds the capability on both sides, so it was slightly confusing...)

bonomat (commented):
I'm not convinced that this is a configurable parameter that is needed. 
Without knowing the internals of our code base it's impossible to know what it really does. 

Also, I would argue that I've some experience with wallet management and I've never changed the default setup a wallet provider gave me, in the worst case, it takes some time until the money is there but usually wallets are smart enough to keep scanning until the first unused address has been found. 

Given the recent learning from @da-kami and the creation of https://github.com/bitcoindevkit/rust-electrum-client/issues/67 I don't think we need this.

luckysori:
> Given the recent learning from @da-kami and the creation of https://github.com/bitcoindevkit/rust-electrum-client/issues/67 I don't think we need this. 

Could/should we disable batch processing for the time being?

da-kami:
>> Given the recent learning from @da-kami and the creation of https://github.com/bitcoindevkit/rust-electrum-client/issues/67 I don't think we need this.

> Could/should we disable batch processing for the time being? 

We could try it, but we should be very careful that monitoring still behaves as expected and is performant.

da-kami:
I'm closing this for now, because our local tests have shown that reducing the cache size has not helped with overcoming the warnings. Will be solved with updating the setup rather than changing ItchySats.

klochowicz:
@da-kami are we intending to merge this? I've noticed it has been reopened some time ago, but I'm not sure whether it's needed.

da-kami:
Closing again because we move forward with #2174

da-kami:
> @da-kami are we intending to merge this? I've noticed it has been reopened some time ago, but I'm not sure whether it's needed.

Sorry, I should have probably added some context when re-opening.

PR #2129, opened 2022-05-30
Taker checks balance before taking an offer

da-kami:
I think it is more resilient to check this *before* we start. 

At the moment even if we know that we can't open the CFD we will create it in the database and then hope to add a failed event and eventually clean it up and move it to failed. Quite a lot of work if we already know it won't work out in the first place. 
This might impact usability, because we get feedback on the wallet having a problem early - which might be annoying at times, but I think in the long run this is better because we can react to the problem rather than pushing this out by "failing the setup".

The taker just checks before sending a request to the maker.
The maker checks upon receiving a request and rejects right away if the balance is insufficient.

### DIFF
--- a/daemon-tests/src/mocks/mod.rs
+++ b/daemon-tests/src/mocks/mod.rs
@@ -3,6 +3,7 @@ use crate::mocks::monitor::MockMonitor;
 use crate::mocks::oracle::MockOracle;
 use crate::mocks::price_feed::MockPriceFeed;
 use crate::mocks::wallet::MockWallet;
+use crate::Amount;
 use model::olivia;
 use std::sync::Arc;
 use tokio::sync::Mutex;
@@ -58,6 +59,17 @@ impl Mocks {
             .await
             .expect_sign()
             .returning(|sign_msg| Ok(sign_msg.psbt));
+    }
+
+    pub async fn mock_balance(&mut self) {
+        self.mock_balance_with(Amount::from_sat(100000000)).await;
+    }
+
+    pub async fn mock_balance_with(&mut self, balance: Amount) {
+        self.wallet()
+            .await
+            .expect_get_balance()
+            .returning(move |_| Ok(balance));
     }
 
     pub async fn mock_oracle_announcement(&mut self) {
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -1,6 +1,7 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 
 use crate::bitcoin::Txid;
+use anyhow::bail;
 use anyhow::Context as _;
 use anyhow::Result;
 use bdk::bitcoin;
@@ -18,6 +19,7 @@ use model::Order;
 use model::OrderId;
 use model::Price;
 use model::Role;
+use model::TxFeeRate;
 use model::Usd;
 use parse_display::Display;
 use seed::Identities;
@@ -99,6 +101,7 @@ where
     W: Handler<wallet::BuildPartyParams>
         + Handler<wallet::Sign>
         + Handler<wallet::Withdraw>
+        + Handler<wallet::GetBalance>
         + Actor<Stop = ()>,
     P: Handler<xtra_bitmex_price_feed::LatestQuote> + Actor<Stop = xtra_bitmex_price_feed::Error>,
 {
@@ -360,4 +363,39 @@ impl Environment {
             _ => Environment::Unknown,
         }
     }
+}
+
+pub async fn balance_check(
+    margin: Amount,
+    tx_fee_rate: TxFeeRate,
+    build_party_params: &(impl MessageChannel<wallet::BuildPartyParams> + 'static),
+    get_balance: &(impl MessageChannel<wallet::GetBalance> + 'static),
+) -> Result<()> {
+    let dummy_pk = "032e58afe51f9ed8ad3cc7897f634d881fdbe49a81564629ded8156bebd2ffd1af"
+        .parse()
+        .expect("static pk to parse");
+    let build_params = build_party_params
+        .send(wallet::BuildPartyParams {
+            amount: margin,
+            identity_pk: dummy_pk,
+            fee_rate: tx_fee_rate,
+        })
+        .await
+        .context("Wallet actor unavailable")??;
+
+    let lock_amount = build_params.lock_amount;
+    let balance = get_balance
+        .send(wallet::GetBalance)
+        .await
+        .context("Wallet actor unavailable")??;
+
+    if balance < lock_amount {
+        bail!(
+            "Wallet balance {} insufficient to open cfd with lock amount {}",
+            balance,
+            lock_amount
+        )
+    }
+
+    Ok(())
 }
--- a/daemon/src/taker_cfd.rs
+++ b/daemon/src/taker_cfd.rs
@@ -1,3 +1,4 @@
+use crate::balance_check;
 use crate::collab_settlement_taker;
 use crate::connection;
 use crate::oracle;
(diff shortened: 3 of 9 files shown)


da-kami:
Please let me know what you think considering adding a check on the maker as well!

daemon-tests/src/mocks/mod.rs
```diff
@@ -60,6 +61,17 @@ impl Mocks {
...
+    pub async fn mock_balance(&mut self) {
+        self.mock_balance_with(Amount::from_sat(100000000)).await;
+    }
+
+    pub async fn mock_balance_with(&mut self, balance: Amount) {
+        self.wallet()
+            .await
+            .expect_get_balance()
+            .returning(move |_| Ok(balance));
+    }
```
luckysori, line 73: ❓ Make it private if it's not used elsewhere?

daemon/src/taker_cfd.rs
```diff
@@ -191,6 +193,31 @@ where
...
 
+        // Construct the CFD so we can retrieve the margin
+        let cfd = Cfd::from_order(
+            &order_to_take,
+            quantity,
+            self.maker_identity,
+            Some(self.maker_peer_id),
+            Role::Taker,
+            leverage,
+        );
```
luckysori: 🔧 I would remove the comment since we are constructing the CFD to retrieve the margin _and_ to insert it into the database like we used to do.

daemon/src/taker_cfd.rs
```diff
@@ -191,6 +193,31 @@ where
...
+
+        // Make sure we have enough balance to open this position
```
luckysori: :upside_down_face: Nitpicky, but this comment is telling me _what_ we are doing which is sufficiently clear to me from the condition checked and the error message emitted.

I would remove it.

daemon/src/taker_cfd.rs
```diff
@@ -191,6 +193,31 @@ where
...
+        let balance = self
+            .wallet
+            .send(wallet::GetBalance)
+            .await
+            .context("Wallet actor unavailable")??;
+        let my_margin = cfd.margin();
```
luckysori: Probably impossible to action, but it's interesting to note that the margin doesn't take into account the transaction fee, which is calculated later on.

This means that we could sometimes succeed on the following check and still not have enough money to pay for the lock transaction.

luckysori (approved):
No concerns about this for taker or maker.

One thing to consider is the frequency of false positives i.e. times when we wrongfully report that we don't have enough money to cover the margin. I suppose it doesn't matter because this can also happen when we try to construct the lock transaction later on.

daemon-tests/src/mocks/mod.rs
da-kami: that's just for completion on the API, could be removed.

daemon/src/taker_cfd.rs
da-kami: It is actually possible by using `BuildPartyParams` - I don't think building `lock` is expensive so I gave it another iteration :)

daemon/src/lib.rs
```diff
@@ -361,3 +365,36 @@ impl Environment {
...
     }
 }
+
+pub async fn balance_check(
+    margin: Amount,
+    tx_fee_rate: TxFeeRate,
+    build_party_params: &(impl MessageChannel<wallet::BuildPartyParams> + 'static),
+    get_balance: &(impl MessageChannel<wallet::GetBalance> + 'static),
+) -> Result<()> {
+    let (_, pk) = keypair::new(&mut rand::thread_rng());
```
luckysori: 🔧 Could we use a static dummy value? It would be cheaper and more obvious that it's irrelevant.

daemon/src/lib.rs
```diff
@@ -361,3 +365,36 @@ impl Environment {
...
+    let try_build_params = build_party_params
```
luckysori: ❓I guess this is what you meant?

```suggestion
    let build_params = build_party_params
```

maker/src/cfd.rs
```diff
@@ -367,6 +370,25 @@ where
...
 
+        if let Err(e) = balance_check(
+            cfd.margin(),
+            order_to_take.tx_fee_rate,
+            &self.wallet,
+            &self.wallet,
+        )
+        .await
+        {
+            tracing::warn!("{e}");
```
luckysori: 🔧 I don't think this would be redundant.

```suggestion
            tracing::warn!("Rejecting taker's order: {e}");
```

daemon/src/wallet.rs
```diff
@@ -607,4 +618,26 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn given_insufficient_funds_then_error_when_building_party_params() {
+        let mut tasks = Tasks::default();
+
+        let actor = Actor::new_offline(Amount::from_btc(0.1).unwrap(), 1, Duration::from_secs(120))
+            .unwrap()
+            .create(None)
+            .spawn(&mut tasks);
```
luckysori: 🔧 This is a bit more convenient via our custom `TokioGlobalSpawnExt` trait.

```suggestion
            .spawn_global();
```

daemon/src/wallet.rs
```diff
@@ -607,4 +618,26 @@ mod tests {
             .unwrap()
             .expect("single UTXO to be available after unlocking it");
     }
+
+    #[tokio::test]
+    async fn given_insufficient_funds_then_error_when_building_party_params() {
+        let mut tasks = Tasks::default();
+
+        let actor = Actor::new_offline(Amount::from_btc(0.1).unwrap(), 1, Duration::from_secs(120))
```
luckysori: 🙃 With the values being tested next to each other in a dedicated section it would be a little bit easier to read this test. Something like:

```rust
let balance = Amount::from_btc(0.1).unwrap();
let margin = Amount::from_btc(0.2).unwrap();
```

daemon/src/wallet.rs
```diff
@@ -607,4 +618,26 @@ mod tests {
             .unwrap()
             .expect("single UTXO to be available after unlocking it");
     }
+
+    #[tokio::test]
+    async fn given_insufficient_funds_then_error_when_building_party_params() {
```
luckysori: :+1: Thank you for adding a test! Should we add another one for the situation where margin and lock amount are equal (i.e. can't pay for the fee)?

luckysori (approved):
Thanks for investing more time into this and nice idea to make it even more user-friendly.

daemon/src/wallet.rs
```diff
@@ -607,4 +619,51 @@ mod tests {
...
+        actor
+            .send(BuildPartyParams {
+                amount: margin,
+                identity_pk,
+                fee_rate: TxFeeRate::default(),
+            })
+            .await
+            .unwrap()
+            .expect_err("insufficient funds");
+    }
```
bonomat: Am I missing something or is this the same test as above?

daemon/src/wallet.rs
```diff
@@ -607,4 +619,51 @@ mod tests {
             .unwrap()
             .expect("single UTXO to be available after unlocking it");
     }
+
+    #[tokio::test]
+    async fn given_margin_less_balance_then_insufficient_funds_error_when_building_party_params() {
```
bonomat, line 624: You only test the negative case here. Shouldn't we add a positive one as well?

bonomat (approved):
Looks good. Either I missed something or you have one test twice :)

daemon/src/wallet.rs
da-kami: The positive one is implicitly tested in the already existing tests.

da-kami:
Let's see what testnet says :)

Issue #1997, opened 2022-05-10
Cleanup CFDs without events when handling failed CFDs


da-kami:
This is a solution for https://github.com/itchysats/itchysats/issues/1987

Loading and moving CFDs into `failed_cfds` should additionally handle CFDs that that fullfil:

* CFD has no events AND CFD is older than 1h

This is, unfortunately not trivial to fix because of the time constraint, because we don't save a creation timestamp in the `cfds` table but rely on `events` only.

Here are some ideas to tackle the issue of not knowing when a CFD (without events) was created:

1. Add a `created` timestamp to `cfds` - we can easily do that. We can also write a migration that fills in e.g. the first event's timestamp for existing CFDs - or if no events are found just default to some timestamp in the past.
     * This has the advantage that we cleanup "no event" CFDs as described above because we will have a timestamp that we can rely on.
2. Add a new `Created` event that is put into the database when `intert`ing a CFD (in a db-transaction, so we make sure we have it...)
    * This does not solve moving no-event CFDs periodically - but we could still check on startup and later rely on this event.

I'd be for finally adding a timestamp to the `cfds` table - it has been annoying *not* to have this timestamp numerous times. Do we have any good reason not to do so besides "not needing it"?

luckysori:
We could just hide it in projection if a CFD has no events?

da-kami:
Closing until https://github.com/itchysats/itchysats/discussions/2128 has a conclusion. 
@luckysori I am not sure if you already invested time into a solution, if so it would be great if you could add to the discussion!

PR #2134, opened 2022-05-30
Delete trivial CFDs on `maker`/`taker` startup

luckysori:
This work was started before #2128 superseded #1997.

### DIFF
--- a/sqlite-db/src/delete_trivial_cfds.rs
+++ b/sqlite-db/src/delete_trivial_cfds.rs
@@ -1,0 +1,107 @@
+use crate::Connection;
+use anyhow::Result;
+
+impl Connection {
+    /// Delete all CFDs which have no events associated with them.
+    ///
+    /// It is recommended to call this method right after calling
+    /// `sqlite_db::connect`, before new CFDs are added to the
+    /// database. This is to ensure that no relevant data is deleted.
+    pub async fn delete_trivial_cfds(&self) -> Result<()> {
+        let mut conn = self.inner.acquire().await?;
+
+        // FIXME: I cannot run `cargo sqlx prepare` locally :(, so I
+        // can't add/modify any of the `sqlx` macros.
+        let affected_rows = sqlx::query(
+            r#"
+            DELETE FROM
+                cfds
+            WHERE
+                id
+            NOT IN
+            (
+                SELECT
+                    events.cfd_id
+                FROM
+                    events
+            )
+            "#,
+        )
+        .execute(&mut *conn)
+        .await?
+        .rows_affected();
+
+        if affected_rows > 0 {
+            tracing::info!(n = %affected_rows, "Deleted trivial CFDs");
+        } else {
+            tracing::info!("No trivial CFDs to delete");
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use model::CfdEvent;
+    use model::EventKind;
+
+    use crate::memory;
+    use crate::tests::dummy_cfd;
+
+    #[tokio::test]
+    async fn given_trivial_cfds_then_they_are_deleted_from_cfds() {
+        let db = memory().await.unwrap();
+
+        let cfd_0 = dummy_cfd();
+        db.insert_cfd(&cfd_0).await.unwrap();
+
+        let cfd_1 = dummy_cfd();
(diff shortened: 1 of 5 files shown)


sqlite-db/src/delete_trivial_cfds.rs
```diff
@@ -0,0 +1,107 @@
...
+impl Connection {
+    /// Delete all CFDs which have no events associated with them.
+    ///
+    /// It is recommended to call this method right after calling
+    /// `sqlite_db::connect`, before new CFDs are added to the
+    /// database. This is to ensure that no relevant data is deleted.
+    pub async fn delete_trivial_cfds(&self) -> Result<()> {
+        let mut conn = self.inner.acquire().await?;
+
+        // FIXME: I cannot run `cargo sqlx prepare` locally :(, so I
```
da-kami: Is this still relevant? 
Given that you don't have any compile errors I think this is outdated.
Note: I am able to run `prepare_db.sh` without problems.

da-kami (approved):
I am not sure deleting these CFDs is what we want. Do we want to keep track of failed positions? 
If we don't care about tracking failed ones we should delete them - but probably preventing them from being inserted in the first place would be better.
I just find it hard (from a users perspective) to understand the difference between something that failed and something that just vanishes. But I don't know if that's relevant. Happy to merge it and see how it feels.

sqlite-db/src/delete_trivial_cfds.rs
luckysori: Thanks for noticing this!

You might notice that we're using `query` instead of `query!` here. `query` gives you no compile-time guarantees. I think we always want to use the macro versions of these APIs.

The tests verify that it works, but I think it would still be good to use `query!` here. Could you or someone else modify the code and run `prepare_db.sh` afterwards?

luckysori:
> I am not sure deleting these CFDs is what we want. Do we want to keep track of failed positions? If we don't care about tracking failed ones we should delete them 

I hadn't considered that we would want to keep them. I would have thought that logs would be sufficient for these cases. There's not much interesting info to learn about these CFDs.

> but probably preventing them from being inserted in the first place would be better. 

I like the way you're thinking. I considered modifying the code to ensure that we don't insert a CFD if the started event is not being inserted too. It could even be a single database transaction.

But in practice it is harder than I thought, because I think our UI responsiveness relies on immediately inserting CFDs into the database. Apparently the current difference between inserting the CFD and inserting `ContractSetupStarted` is only ~200ms, but it's not nothing. 

Furthermore, we would me modifying soon-to-be legacy code (around the `setup_taker::Actor`), so I'd rather wait for those changes to come before we work in this direction. 

> I just find it hard (from a users perspective) to understand the difference between something that failed and something that just vanishes.

This PR should produce very little unpredictability in the UI. IIRC, we cache the CFDs in the `projection` layer which means that moving or deleting things from `cfds` and `events` doesn't actually result in the UI changing until the application is restarted (or we send a `CfdChanged` message, which is not always the case). Also keep in mind that deleting trivial CFDs would only run on startup.

I think it is weird that we keep CFDs that we failed to construct in the same UI area where we keep relevant CFDs. So moving them out of there, whether periodically or on restart, cannot be harmful for me.

da-kami:
> I think it is weird that we keep CFDs that we failed to construct in the same UI area where we keep relevant CFDs. So moving them out of there, whether periodically or on restart, cannot be harmful for me.

I agree with this, but given our discussion in https://github.com/itchysats/itchysats/discussions/2128 I am not sure the current solution is necessary because removing these CFDs only on startup does not help an (Umbrel) user much, because they usually don't restart the daemon. I would try to avoid a solution that is only triggered upon restart. We can still merge this for now, but I would see it as something temporary.
I think what we need is a proper cleanup task. In addition I would try to invest into more prevention - that's what motivated me to do https://github.com/itchysats/itchysats/pull/2129.

sqlite-db/src/delete_trivial_cfds.rs
bonomat: Pushed a commit for you ;)

luckysori: Thanks! I'm gonna close this for now, because we don't seem to agree on it.

luckysori:
Closed because we don't seem to want this. See https://github.com/itchysats/itchysats/discussions/2128.

PR #2135, opened 2022-05-31
Introduce collab settlement via libp2p

### LINKED ISSUE #320: Enable the taker to publish the collaborative settlement tx

### LINKED ISSUE #2112: Reintroduce collab settlement unit test failing due to `maia` close transaction txfee bug
A long time ago we added a test that ensures that if we get liquidated by "too many rollovers" the party that got liquidated ends up with a `0` amount:

https://github.com/itchysats/itchysats/blob/6699ca22011fc63f6a23549314d4f7aa5778bb45/model/src/cfd.rs#L3045-L3074

I just tried to run the test but it is still failing. This means that the payout calc is not complete yet, we should be able to handle this edge case.

klochowicz:
Fixes #320.
Fixes #2112.

Note: We shouldn't release until we finish the following tickets (I've added them to the current iteration, as they can be done in parallel):

TODO:
- #2155 
- #2156
- #2157


daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+    PeerId,
+);
+
+/// Permanent actor to handle incoming substreams for the `/itchysats/close-position/1.0.0`
+/// protocol.
+///
+/// There is only one instance of this actor for all connections, meaning we must always spawn a
+/// task whenever we interact with a substream to not block the execution of other connections.
+pub struct Actor {
+    tasks: Tasks,
```
luckysori: :smile: Unbounded growth incoming!

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+                        propose,
+                        framed,
+                        peer,
+                    })
+                    .await?;
+
+                anyhow::Ok(())
+            },
+            |e| async move { tracing::warn!("Failed to handle incoming close position: {e:#}") },
+        );
```
luckysori: 🤔 I would strongly consider passing the stream to an ephemeral actor/task. We could then forward the (unfortunately required) `Accept/Reject` message from the HTTP API through the `close_position::maker::Actor` to the corresponding ephemeral actor/task.

The advantages I see:

1. Not blocking all other protocols when we are handling `ProposeReceived`, `Accept` or `Reject`. Each protocol per CFD is completely independent of all other CFDs. It's true that this is already mitigated by spawning tasks after handling different messages, but I think this could be a more straightforward way of avoiding delays.
2. The pending_protocols `HashMap` can be simplified to just hold the address/channel-sender of the ephemeral rollover actors/tasks.
3. The ephemeral rollover actor/task design is very simple:
    - Wait for taker proposal.
    - Start protocol.
    - Wait for decision either via handler or dedicated channel.
    - Complete the rest of the protocol.
    
I think this could look a lot like a linear task, with the unfortunate awkwardness of having to wait for user interaction in the middle. Thinking about it I would favour a function like `setup_contract::new` with a channel, over an ephemeral actor.

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+            propose,
+            framed,
+            peer,
+        } = msg;
+        let order_id = propose.id;
+
+        let result = self
+            .executor
+            .execute(order_id, |cfd| {
+                // TODO: Validate that the correct peer is invoking this?
```
luckysori: 🔧 Let's just do it or remove it, please. The CFD holds the `counterparty_peer_id`.

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+                        .finalize()
+                        .context("Failed to finalize transaction")?;
+
+                    framed
+                        .send(ListenerMessage::Msg3(Msg3 { listener_signature }))
+                        .await
+                        .map_err(|source| Failed::AfterReceiving {
+                            source: anyhow!(source),
+                            settlement: settlement.clone(),
+                        })?;
```
luckysori: :question: This does not follow the [diagram](https://github.com/itchysats/itchysats/issues/320#issuecomment-1088344881) posted in the original issue and is therefore not symmetric. Is that intentional at this stage?

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+use model::ClosePositionTransaction;
+use model::CollaborativeSettlement;
+use model::OrderId;
+use model::SettlementProposal;
+use std::collections::HashMap;
+use tokio_tasks::Tasks;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_libp2p::Substream;
+use xtra_productivity::xtra_productivity;
+
```
luckysori: 🔧 We normally omit newlines among `use` statements.

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+        source: Error,
+    },
+}
+
+// Allows for easy use of `?`.
+impl From<Error> for Failed {
+    fn from(source: Error) -> Self {
+        Failed::BeforeReceiving { source }
+    }
+}
```
luckysori: :thinking: Not a huge fan of this because it is really not obvious from reading the code that most of the errors above result in a particular path whereas the very last one diverges. I would prefer to be explicit with mapping to avoid surprises at the cost of a few lines of code.

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+            },
+            {
+                let executor = self.executor.clone();
+                move |failed| async move {
+                    match failed {
+                        Failed::BeforeReceiving { source } => {
+                            emit_failed(order_id, source, &executor).await;
+                        }
+                        Failed::AfterReceiving { source, settlement } => {
+                            tracing::trace!(?settlement, "Received settlement");
```
luckysori: 🔧 I would move this to the happy path. It's a bit strange to only log this if we fail.

Also I think it would be clearer if you said "received close transaction signature from taker" or something like that. "Settlement" doesn't mean much to me.

daemon/src/close_position/protocol.rs
```diff
@@ -0,0 +1,226 @@
...
+}
+
+pub(crate) async fn emit_rejected(order_id: OrderId, executor: &command::Executor) {
+    if let Err(e) = executor
+        .execute(order_id, |cfd| {
+            Ok(cfd.reject_collaborative_settlement(anyhow!("maker decision")))
+        })
+        .await
+    {
+        tracing::error!(%order_id, "Failed to execute rollaborative settlement rejected: {e:#}")
```
luckysori: 🔧 Message is wrong.

luckysori: ❓ Also, do we want to be consistent and make this one a `warn` too?

daemon/src/close_position/taker.rs
```diff
@@ -0,0 +1,106 @@
...
+            maker_peer_id,
+        } = msg;
+
+        let (close_position_tx, _) = self
+            .executor
+            .execute(order_id, |cfd| {
+                cfd.start_close_position_taker(price, self.n_payouts)
+            })
+            .await
+            .context("could not start closing position")?;
```
luckysori: ❓ Why not do this inside the following task?

daemon/src/close_position/protocol.rs
```diff
@@ -0,0 +1,226 @@
...
+    BeforeSendingSignature {
+        source: anyhow::Error,
+    },
+}
+
+impl From<anyhow::Error> for DialerFailed {
+    fn from(source: anyhow::Error) -> Self {
+        Self::BeforeSendingSignature { source }
+    }
+}
```
luckysori: 🔧 Similarly, very unintuitive to me.

daemon/src/connection.rs
```diff
@@ -460,14 +418,8 @@ impl Actor {
...
-            wire::MakerToTaker::Settlement { order_id, msg } => {
-                if let Err(NotConnected(_)) = self
-                    .collab_settlement_actors
-                    .send_async(&order_id, msg)
-                    .await
-                {
-                    tracing::warn!(%order_id, "No active collaborative settlement")
-                }
+            wire::MakerToTaker::Settlement { .. } => {
+                tracing::error!("legacy handler - use libp2p instead");
```
luckysori, line 422: ❓ When should we remove this? Why not now?

daemon/src/lib.rs
```diff
@@ -156,6 +156,13 @@ where
             &oracle_addr,
         )));
 
+        let (endpoint_addr, endpoint_context) = Context::new(None);
+
+        let libp2p_collab_settlement_addr =
+            close_position::taker::Actor::new(endpoint_addr.clone(), executor.clone(), n_payouts)
```
luckysori: :question: Why do we use different terminology for the same protocol? Not sure what the situation was like before the PR, but it would be good for our sanity to decide on close or settlement.

daemon/src/taker_cfd.rs
```diff
@@ -139,25 +141,18 @@ impl<O, W> Actor<O, W> {
...
-            self.n_payouts,
-            self.conn_actor.clone(),
-            self.process_manager_actor.clone(),
-            self.db.clone(),
-        )
-        .create(None)
-        .spawn(&mut self.tasks);
-
-        disconnected.insert(addr);
+        // Sending msg sync to check for invariant
```
luckysori: 🙃 I don't think "sync" is the correct term, because this is `async` code. `xtra`'s documentation says that we are sending a message to the actor and "asynchronously wait[ing] for a response".

maker/src/actor_system.rs
```diff
@@ -116,6 +117,11 @@ where
             &oracle_addr,
         )));
 
+        let libp2p_collab_settlement_addr =
+            close_position::maker::Actor::new(executor.clone(), n_payouts)
+                .create(None)
+                .spawn(&mut tasks);
```
luckysori: ❓ Shall we follow the new pattern of supervising permanent actors?

model/src/cfd.rs
```diff
@@ -1022,18 +1023,66 @@ impl Cfd {
...
+                && self.can_settle_collaboratively(),
+            "Failed to start collaborative settlement"
+        );
+
+        let (close_position_tx, proposal) = self.make_proposal(current_price, n_payouts)?;
+
+        anyhow::ensure!(
+            close_position_tx.unsigned_transaction == proposed_close_transaction,
+            "Proposed close transaction does not equal locally created one"
+        );
```
luckysori: :upside_down_face: It would be good to see the difference between the two transactions here.

model/src/cfd.rs
```diff
@@ -1854,8 +1888,75 @@ pub struct Dlc {
...
+    own_signature: Signature,
+
+    counterparty_pk: PublicKey,
+    counterparty_signature: Option<Signature>,
+}
+
+impl ClosePositionTransaction {
+    pub fn unsigned_transaction(&self) -> Transaction {
+        self.unsigned_transaction.clone()
+    }
```
luckysori: 🔧 I think it's better to return a `&Transaction` here. I think the return value is only ever used to compare, which can always be done with references.

The method `start_close_position_maker` will have to take a `&Transaction` for this to work.

model/src/cfd.rs
```diff
@@ -1854,8 +1888,75 @@ pub struct Dlc {
...
+        Ok(Self {
+            counterparty_signature: Some(counterparty_signature),
+            ..self
+        })
+    }
+
+    pub fn finalize(self) -> Result<CollaborativeSettlement> {
+        let counterparty_signature = self
+            .counterparty_signature
+            .context("Missing `their_signature`")?;
```
luckysori: 🙃 Probably should be "Missing counterparty signature".

model/src/lib.rs
```diff
@@ -26,7 +26,7 @@ mod sqlx_ext; // Must come first because it is a macro.
 
 mod cfd;
 mod contract_setup;
-mod hex_transaction;
+pub mod hex_transaction;
```
luckysori, line 29: ❓ Is this needed?

model/src/cfd.rs
```diff
@@ -2005,8 +2005,7 @@ impl Dlc {
             (outpoint, amount)
```
luckysori, line 2023: 🔧 Commit message is a bit misleading: the payout curve calculations have not changed!

What has changed is how we calculate and split fees when building a close transaction.

luckysori (commented):
Nice work!

I think I would like the maker side to look more like the taker side. Other than that, a few recommendations along the way.

---

Also: the squashed commit message of https://github.com/itchysats/itchysats/pull/2135/commits/f3caf11948688cdf4a0c1402e38ff909b0c3331f is not very useful. I would appreciate either a little summary or nothing at all 😛

daemon/src/close_position/maker.rs
da-kami: @luckysori I am strongly against ephemeral actors because these ephemeral actors added a lot of complexity and I don't think we gain much.

This implementation is similar to the new rollover actor. Each message handler dispatches to a task as soon as possible - we do not process long running work in the handler but always dispatch to a task. 
Could you elaborate where you see a problem of blocking the actor here? Is your concern related to the fact that we dispatch a message to ourselves here?

> Thinking about it I would favor a function like setup_contract::new with a channel, over an ephemeral actor.

We can consider doing this. For rollover we dispatch to a task upon `accept` - i.e. when the actual protocol is triggered. In the long run we could consider splitting these protocols in two, one for the decision and one for the execution. 
Are you proposing to pass in a channel for handling `accept`/`reject` inside a task? i.e.:

1. We create an accept/reject channel that handles all CFDs per protocol
2. Upon receiving the protocol substream (`NewInboundSubstream`) we spawn a task and pass the accept/reject channel in there.
3. The task waits until accept/reject for the respective CFD is received...?

Would you create one channel for accept/reject per protocol or create one per CFD? If you do it per CFD then the endpoint cannot just send into it I think, we have to go through the actor again to retrieve the channel association.
I feel the current design might be simpler than that.

daemon/src/close_position/maker.rs
da-kami: And while we are at it: Please either remove it or fix this comment in rollover as well. I don't think the TODO adds any value :)

daemon/src/close_position/maker.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    async fn handle(&mut self, msg: Accept) -> Result<()> {
+        let Accept { order_id } = msg;
+
+        let (mut framed, transaction, proposal, _peer) =
+            self.pending_protocols
+                .remove(&order_id)
+                .with_context(|| format!("No active protocol for order {order_id}"))?;
+
+        self.tasks.add_fallible(
```
da-kami: I think this part should be the same for both! That's the point of making it symmetric - the actual execution of the protocol (after accept) is the same for both. 
I imagine this to be similar to what we have in `rollover` where both dispatch to the same function. I think that's a good pattern to ensure symmetry.

daemon/src/close_position/taker.rs
da-kami: Probably the idea is that this error should be reported back to the caller. I am not sure I agree though - unless the caller actually does something sensible with this I would move it into the task. reporting back information only makes sense if the caller does something with it. If the caller is not interested we should design the API accordingly.

maker/src/actor_system.rs
da-kami: I think we should - and also consider this for the new `rollover` actor.

maker/src/cfd.rs
```diff
@@ -482,41 +485,60 @@ impl<O, T, W> Actor<O, T, W> {
...
-            Ok(_) => Ok(()),
-            Err(NotConnected(e)) => {
-                self.executor
-                    .execute(order_id, |cfd| {
-                        Ok(cfd.fail_collaborative_settlement(anyhow!(e)))
-                    })
-                    .await?;
-
-                bail!("Accept failed: No settlement in progress for order {order_id}")
+        // TODO: Handle case when taker upgraded after opening and *can* use libp2p
```
da-kami, line 488: I am sure you are aware of this, but I still state it: Since the old code for the `taker` was removed this `TODO` has to be actioned before merging, otherwise the taker is unable to collab close previously opened positions after an upgrade!

da-kami (commented):
Thanks for all the work! I think my biggest concern is that the protocol symmetry is not obvious and it would be great to extract the execution part of the protocol in a function that is called by both sides (similar to rollover).

I'll give this another review once the current feedback is addressed!

da-kami:
> Thanks for all the work! I think my biggest concern is that the protocol symmetry is not obvious and it would be great to extract the execution part of the protocol in a function that is called by both sides (similar to rollover).
> 
> I'll give this another review once the current feedback is addressed!

After reading through https://github.com/itchysats/itchysats/issues/320#issuecomment-1088344881 again I wonder if not extracting the "symmetric" protocol part is on purpose. If I understand this correctly what is proposed there explicitly states that the dialer sends the signature before the listener sends it. But does that make sense? What is the problem with both parties always sending first before waiting to receiving? 

The only difference I see is, that both parties have to deal with the problem where they sent but do not receive the other party's signature. Is that what we are trying to avoid here?
We can, but then we are not actually making the protocol symmetric - we only enable the taker to publish the transaction as well. Given the original ticket that is probably the intention.

daemon/src/close_position/maker.rs
da-kami: Not sure we are actually going for a symmetric protocol, please see https://github.com/itchysats/itchysats/pull/2135#issuecomment-1142868508

da-kami:
> > Thanks for all the work! I think my biggest concern is that the protocol symmetry is not obvious and it would be great to extract the execution part of the protocol in a function that is called by both sides (similar to rollover).
> > I'll give this another review once the current feedback is addressed!
> 
> After reading through [#320 (comment)](https://github.com/itchysats/itchysats/issues/320#issuecomment-1088344881) again I wonder if not extracting the "symmetric" protocol part is on purpose. If I understand this correctly what is proposed there explicitly states that the dialer sends the signature before the listener sends it. But does that make sense? What is the problem with both parties always sending first before waiting to receiving?
> 
> The only difference I see is, that both parties have to deal with the problem where they sent but do not receive the other party's signature. Is that what we are trying to avoid here? We can, but then we are not actually making the protocol symmetric - we only enable the taker to publish the transaction as well. Given the original ticket that is probably the intention.

After another discussion with @klochowicz I think that the current version is good. The goal was to give more feedback to the taker (and enable the taker to publish the transaction), but not to make the protocol symmetric. 
The maker and taker are (to some extent) in a service agreement when it comes to collaborative close. I don't think we can overcome that. The maker should not have an incentive to not publish the transaction, however, the taker might have. The taker should not be able to hold the maker hostage, so I I think

1. Taker sends sig
2. Maker receives sig
3. Maker sends sig
4. Taker receives sig

is correct. That is what is implemented.

daemon/src/close_position/maker.rs
da-kami: To elaborate a bit more on how we could evolve this: 

1. Split the protocol(s) into a decision protocol and execution protocol
2. The decision protocol ends with saving the decision in the database (on both sides)
3. The execution protocol starts with reading the decision from the database and then only proceeds if the decision is as expected.

That way there is not "ephemeral" state and decision in between protocols anymore.

da-kami: One additional thing to note:

The `executor` combines decision making and execution. The decision making is what we want to report back to the caller (i.e. we are able to run the protocol or not), but the execution part (i.e. save event in database)  is actually irrelevant. I know that the `executor` is convenient, but this combination of decision making and execution might not be helpful here. We could consider separating this.
We could also consider handing the saving of the event in the db (as part of the `process_manager` part of the `executor`) in a task. This might have weird side effects if we expect the information to be present after calling `execute` though.

klochowicz: @luckysori I'm happy to chat about it, but I don't think we need another layer of abstraction. It's true that we need to keep in mind to spawn from the handler into a task, but this makes things simple (for example, we don't need to garbage-collect them, we don't need to dispatch to it etc.

I'm reluctant to venture into ephemeral actors again just yet, after we had all the problems last time.

That said, I'd be happy to see a POC and have a look if you're keen to try this out.

klochowicz:
> Nice work!
> 

> Also: the squashed commit message of [f3caf11](https://github.com/itchysats/itchysats/commit/f3caf11948688cdf4a0c1402e38ff909b0c3331f) is not very useful. I would appreciate either a little summary or nothing at all 😛

missed that one 🙈 removed the cruft!

daemon/src/close_position/maker.rs
klochowicz: it's not much of a concern here, as we're not likely to do heaps of these (at most 1 per order?) ! I wrapped it the same as with rollover for now (don't want to roll out Spawner actor just yet, can be refactored later).

daemon/src/close_position/maker.rs
klochowicz: fixed in both

daemon/src/close_position/maker.rs
klochowicz: from what I can tell, this is how Thomas interpreted his own diagram at the time. I'll mention this is in the commit description that this protocol is not symmetric to be explicit.

daemon/src/close_position/maker.rs
klochowicz: that's the rust-analyzer again :/

daemon/src/close_position/protocol.rs
klochowicz: fixed the message and converted all 3 to `error` (consistent with rollover protocol)

daemon/src/connection.rs
klochowicz: I'd refrain to do so until we fully drop our legacy networking; then the whole enum will go away.

daemon/src/lib.rs
klochowicz: I think the old discussion is still open: https://github.com/itchysats/itchysats/discussions/741

honestly, I prefer keeping the old name for now- I inherited the patch from Thomas, who started using the other name.

If we decide to change it later on, we will do so in one sweep.

maker/src/actor_system.rs
klochowicz: After having a second thought, I'll add the supervision to both collab settlement and rollover actors in a follow-up PR - this one is already pretty big :)
added a ticket: https://github.com/itchysats/itchysats/issues/2154

daemon/src/taker_cfd.rs
klochowicz: that is correct. I've amended the comment :)

model/src/cfd.rs
klochowicz: Great suggestion. If anyone will need to ever clone it, they can do it on the outside. actioned!

model/src/lib.rs
klochowicz: yes, it's needed, otherwise I can't serialise the message containing `Transaction`...

maker/src/cfd.rs
klochowicz: it has to be actioned before *releasing* - and I fully intend to do it, will spin-off a couple of the follow-up tickets to deal with it. 
Don't want to make this PR even bigger than it is :)

daemon/src/close_position/maker.rs
klochowicz: I hear what you're saying, however failure `AfterReceiving` is an edge case - `BeforeReceiving` is the typical error, hence the convenience.

daemon/src/close_position/protocol.rs
klochowicz: again, failure before signature is the typical error path, the other one is an edge case

daemon/src/close_position/taker.rs
klochowicz: we do this because we're checking for invariants (ie. whether we are able to perform a collab close). If we check here, we actually show this is the taker's UI, which gives good user experience.

It's preserving current behaviour, and I'm reluctant to change within this PR.
@da-kami I believe it's the caller's decision whether to care or not and we can't tell from this point.

model/src/cfd.rs
klochowicz: thank you for pointing this out, amended the description.

klochowicz:
@da-kami @luckysori thank you very much for the feedback, I have tried to address as many comments as possible (if anything slipped through and you would prefer me to fix it during this PR, let me know). 
I have added a bunch of follow up issues stemming from the TODOs left over, that should be finished before we can release again after merging this PR (they're all added to the iteration).

klochowicz:
> Nice work!
> 
> I think I would like the maker side to look more like the taker side. Other than that, a few recommendations along the way.

Oh, sorry, missed this one! Can try to align them more in a follow-up!

daemon/src/close_position/maker.rs
luckysori: You are right, I must have misinterpreted the diagram. All good!

luckysori:
> > Nice work!
> > I think I would like the maker side to look more like the taker side. Other than that, a few recommendations along the way.
> 
> Oh, sorry, missed this one! Can try to align them more in a follow-up!

In case this wasn't clear: I was referring to extracting the protocol execution into a function (as opposed to sending messages to ourselves).

daemon/src/close_position/maker.rs
luckysori: Nothing to action here. My point is that it would be better to spawn a single `protocol::listener` task, like what we do with the `close_position::taker::Actor`.

I can try to add this in a follow-up PR. Then it would be easier to test the protocol in isolation.

luckysori:
I've pushed a small fixup commit that should make the error handling a little simpler.

luckysori (approved):
Still approved!

daemon/src/collab_settlement/protocol.rs
```diff
@@ -0,0 +1,229 @@
...
+    framed
+        .send(DialerMessage::Propose(Propose {
+            id: order_id,
+            price: collab_settlement_tx.price(),
+            unsigned_tx: unsigned_tx.clone(),
+        }))
+        .await
+        .context("Failed to send Msg0")?;
+
+    // TODO: We will need to apply a timeout to these. Perhaps we can put a timeout generally into
```
da-kami, line 49: 🛠 Please add the timeout. For rollover we also have a timeout on the `taker` (i.e.dialer) side. I think `30` seconds is reasonable to expect and `accept` or `reject` message.

daemon/src/collab_settlement/protocol.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+#[derive(Clone, Copy, Serialize, Deserialize)]
+pub enum Decision {
+    Accept,
+    Reject,
+}
+
+#[derive(Clone, Copy, Serialize, Deserialize)]
+pub struct Msg2 {
```
da-kami: I find it a bit odd to have `Msg2` and `Msg3` but no `Msg1` because it is the decision.

Maybe we should just rename `Msg2` and `Msg3` to what they are (i.e. `DialerSignature` and `ListenerSignature`).

daemon/src/collab_settlement/taker.rs
```diff
@@ -0,0 +1,106 @@
...
+#[xtra_productivity]
+impl Actor {
+    pub async fn handle(&mut self, msg: Settle) -> Result<()> {
+        let Settle {
+            order_id,
+            price,
+            maker_peer_id,
+        } = msg;
+
+        let (collab_settlement_tx, _) = self
```
da-kami, line 55: We could consider moving the executor into the task for the taker as well. Not super important here, but I am not sure if the feedback is actually necessary in the caller and we avoid blocking then. It's unlikely that blocking will cause issues on the taker, but still.

daemon/src/collab_settlement/maker.rs
```diff
@@ -0,0 +1,264 @@
...
+                            cfd.accept_collaborative_settlement_proposal(&proposal)
+                        })
+                        .await?;
+
+                    framed
+                        .send(ListenerMessage::Msg1(Decision::Accept))
+                        .await
+                        .context("Failed to send Msg1::Accept")?;
+
+                    // TODO: We will need to apply a timeout to these. Perhaps we can put a timeout
```
da-kami, line 155: I'd just add the timeout and remove the Todo.

daemon/src/collab_settlement/maker.rs
```diff
@@ -0,0 +1,264 @@
...
+            },
+            {
+                let executor = self.executor.clone();
+                move |failed| async move {
+                    match failed {
+                        Failed::BeforeReceiving { source } => {
+                            emit_failed(order_id, source, &executor).await;
+                        }
+                        Failed::AfterReceiving { source, settlement } => {
+                            // TODO: proceed with the transaction when taker will be able to handle
```
da-kami, line 197: ❓I don't understand this TODO, could you elaborate what you mean here? Is this still relevant?

daemon/src/collab_settlement/taker.rs
```diff
@@ -0,0 +1,106 @@
...
+            },
+            {
+                let executor = self.executor.clone();
+                move |e| async move {
+                    match e {
+                        DialerFailed::AfterSendingSignature {
+                            unsigned_tx: _tx,
+                            error,
+                        } => {
+                            // TODO: We should start monitoring whether other party published the
```
da-kami, line 88: ❓ This is about handling scenarios where we send the signature but then we don't receive it from the other party?
Not sure this TODO helps much, I'd say we create a ticket describing the problem and potentialy solution(s) and remove the TODO.

maker/src/cfd.rs
```diff
@@ -504,41 +507,60 @@ impl<O, T, W> Actor<O, T, W> {
...
-            Ok(_) => Ok(()),
-            Err(NotConnected(e)) => {
-                self.executor
-                    .execute(order_id, |cfd| {
-                        Ok(cfd.fail_collaborative_settlement(anyhow!(e)))
-                    })
-                    .await?;
-
-                bail!("Reject failed: No settlement in progress for order {order_id}")
+        // TODO: Handle case when taker upgraded after opening and *can* use libp2p
```
da-kami, line 518: I think this still has to be handled before we can merge, otherwise takers cannot collab close previously opened positions.

maker/src/cfd.rs
```diff
@@ -504,41 +507,60 @@ impl<O, T, W> Actor<O, T, W> {
...
+        let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
+        if can_use_libp2p(&cfd) {
+            self.libp2p_collab_settlement
+                .send(daemon::collab_settlement::maker::Reject { order_id })
+                .await
+                .map_err(|e| anyhow!(e))?
+            // TODO: handle failure case
```
da-kami: Is this still relevant?

da-kami (commented):
Please go over all the TODOs and evaluate if they are still needed.
I think we should *not* merge this if the maker is not able to handle previously opened positions. Can't we just try libp2p and fallback to legacy if we can't dispatch?

maker/src/cfd.rs
klochowicz: yes.

daemon/src/collab_settlement/protocol.rs
klochowicz: There is no easy way of adding a timeout. 
There is a separate tracking issue to deal with this, as such it spans more than this protocol. Please have a look at PR description.

daemon/src/collab_settlement/taker.rs
klochowicz: I wrote multiple times in this and other tickets, that this spans 

> Please go over all the TODOs and evaluate if they are still needed.
> I think we should _not_ merge this if the maker is not able to handle previously opened positions. Can't we just try libp2p and fallback to legacy if we can't dispatch?

Thanks for double-checking: they are needed, I would not have kept them if they weren't. Please read the PR description and the comment above.

Your concern is addressed in a separate ticket added to this iteration.

daemon/src/collab_settlement/maker.rs
klochowicz: I'd prefer to merge this body of work and then deal with it in #2156 due to potential of splitting remaining work.

da-kami (approved):
Not super happy about merging something into master that may block a release, but if you feel strong about merging go for it :)

daemon/src/collab_settlement/maker.rs
klochowicz: it's still relevant and tracked by a separate ticket: #2157

daemon/src/collab_settlement/taker.rs
klochowicz: the TODO is a pointer to the problem, the ticket is tracked #2157

maker/src/cfd.rs
klochowicz: it's somewhat a duplicate of the above TODO, removed this one.

maker/src/cfd.rs
klochowicz: Yes, that's described in the PR description. There is a ticket for it in this iteration: #2156

klochowicz:
> Not super happy about merging something into master that may block a release, but if you feel strong about merging go for it :)

I understand your point. I feel that remaining issues can be tackled in parallel (they're orthogonal). Making more targeted PRs will also make it easier for the reviewers to give feedback. I feel confident that they can be tackled in this iteration if we split the remaining work.
Note that this won't block a maker-only release (it only would be a problem if we rolled it out into the takers).

I will personally ensure that we only release after the problems are addressed.

klochowicz:
> I've pushed a small fixup commit that should make the error handling a little simpler.

thank you 🙏

daemon/src/collab_settlement/protocol.rs
klochowicz: Great point. the numeration was indeed off. I have added a small commit that cleans this up (including comments and variable names).

Issue #2139, opened 2022-06-01
Move sql specific types out of model into sqlite-db module


bonomat:
Our `model` module is a bit of a mess as it mixes business logic with db logic. This should be cleaned up. 
All sql specific types should live inside `sqlite-db` module and converter functions should be provided.

PR #2140, opened 2022-06-01
Don't store RolloverCompleted event as json data

### LINKED ISSUE #2120: Plan of attack: Normalize `RolloverCompleted`Event
Based on the discussion https://github.com/itchysats/itchysats/discussions/2084 we agree that we want to get rid of the event data JSON blob and have a normalized DB for the event data. 

Below is my proposal on how to tackle this: 

## 1. Introduce new tables

_Note: the tables have some errors on the fields and some are duplicated, so don't follow naively, at the end, we want to store our current RollOverCompelted event data_ 

image

_Note: the fields might be not complete, please check in the code_

<details>
<summary>Table descriptions</summary>

```
Table cfds {
  id int [pk, increment] // auto-increment
  order_id varchar
 }

Table events {
  id int [pk, increment] // auto-increment
  cfd_id int [ref: > cfds.id]
  name varchar
  data varchar
  created_at timestamp
}


Table RolloverEventData as rollover{
  id int [pk, increment] // auto-increment
  cfd_id int [pk, ref: > cfds.id]
  event_id int [ref: > events.id]
  funding_fee varchar
  rate varchar
  rest varchar
  commit_tx varchar
  idendity varchar
  lock varchar
  maker_address varchar
  maker_lock_amount varchar
  publish varchar
  publish_pk_counterparty varchar
  refund varchar
  refund_timelock varchar
  revocation varchar
  revocation_pk_counterparty varchar
  settlement_event_id varchar
  taker_address varchar
  taker_lock_amount varchar
}

Table RevokedCommitTransactions {
  id int [pk, increment] // auto-increment
  rollover_id int [ref: > rollover.id]
  encsig_ours varchar
  publication_pk_theirs varchar
  revocation_sk_theirs varchar
  script_pubkey varchar
  txid varchar
}

Table AdaptorSignatures as adaptor {
  id int [pk, increment] // auto-increment
  rollover int [ref: > rollover.id]
  adaptor_sig varchar
  maker_amount varchar
  taker_amount varchar
  n_bits varchar
  range_end varchar
  range_start varchar
  txid varchar
}

```

</details>

## 2. Manual migrate data

It is hard to build an automated migration script, so instead, I would opt for migrating existing rollover events first thing when starting up. Similar to `cull_old_events` we should block the application and only continue once the is done. 
For the first migration this could take some time but ever subsequent restart will be fast because there should be no more fields to migrate. 

### 3.a Insert 

Given we have multiple tables for a single rollover event, inserting into the db might be a bit more complex. 
We've done these things in the past, but the important thing is that the whole action is atomic while following fk-constraints. 
Given the table structure above, we still insert `RolloverCompleted` events into the `Events` table which will give us a nice audit trail. 

### 3.b Update

On a subsequent `RolloverEvent` we need to update the tables. 
In my proposal I have a `pk` on `cfd_id` in `RolloverCompleted` table, which means, we can only have one row there. 
I propose to do a simple drop from this table. This will force us to drop the corresponding rows from `AdaptorSignatures` and `RevokedCommitTransactions` due to the `fk` relationship. 

## 4. Loading

Loading in a single db query as before might be really hard. Instead, I would do two queries,
1. load all legacy events using `Events.data` but ignore `RolloverCompleted`
2. load `RolloverCompleted` by reading from the new tables

bonomat:
This took longer than expected... mostly because of our mix between model types and database types. 
I probably should have gone for #2139 before this PR but then I was halfway through. I would be in favor of not changing this with this PR but allowing this somewhat of a mess and then going straight to #2139. 

When implementing I noticed one downside of this solution: 
- when rolling over the first time, we do not clean the `ContractSetupCompleted` row.  I propose to not clean this up but rather start migrating `ContractSetupCompleted` into its own tables as well (or reuse the one from rollover completed).

Resolves #2120


sqlite-db/migrations/20220525225235_rename_cets_table.sql
```diff
@@ -0,0 +1,3 @@
+-- Add migration script here
+ALTER TABLE
+    cets RENAME TO closed_cets;
```
da-kami: 🛠 I feel this is a quite random change that should not be part of this PR.
I am not against renaming this, but if we do so we should also rename the other tables related to `closed` CFDs, i.e. `commit_txs`, `refund_txs`.

sqlite-db/migrations/20220526225235_rollover-completed-table.sql
```diff
@@ -0,0 +1,51 @@
...
+    revocation_secret text NOT NULL,
+    revocation_pk_counterparty text NOT NULL,
+    lock_tx text NOT NULL,
+    lock_tx_descriptor text NOT NULL,
+    commit_tx text NOT NULL,
+    commit_adaptor_signature text NOT NULL,
+    commit_descriptor text NOT NULL,
+    refund_tx text NOT NULL,
+    refund_signature text NOT NULL,
+    FOREIGN KEY (cfd_id) REFERENCES cfds (id),
```
da-kami: ❓ Is there any good reason why this table references `cfds`?
I don't think we need the relationship to `cfds`, because that is already given by `EVENTS` being associated to `cfds`, i.e. `cfds -> events -> rollover_completed_event_data`.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,416 @@
...
+    event_id: i64,
+    dlc: &Dlc,
+    funding_fee: FundingFee,
+    offer_id: OrderId,
+) -> Result<()> {
+    let (lock_tx, lock_tx_descriptor) = dlc.lock.clone();
+    let (commit_tx, commit_adaptor_signature, commit_descriptor) = dlc.commit.clone();
+    let (refund_tx, refund_signature) = dlc.refund.clone();
+
+    // casting because u64 is not implemented for sqlx: https://github.com/launchbadge/sqlx/pull/919#discussion_r557256333
```
da-kami: 😅

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,416 @@
...
+    let event_kind = event.event;
+    match event_kind {
+        EventKind::RolloverCompleted {
+            dlc: Some(dlc),
+            funding_fee,
+        } => {
+            // we begin a transaction so that it rollsback in case of a failure further down
+            let mut inner_transaction = connection.begin().await?;
+
+            delete_rollover_completed_event_data(&mut inner_transaction, event.id).await?;
```
da-kami: 🛠 This is less resilient than having a separate task (i.e. `cull_old_dlcs`) that deals with the removal. 
If a rollover was completed it should be inserted no matter if deleting the old data worked or not. Otherwise we might fail on inserting an already finished rollover on deleting old data.

I think we can remove `cull_old_dlcs` as separate task, but I think we should still separate the deletion from inserting! i.e. the caller should first finish `insert_rollover_completed_event` and then call `delete_rollover_completed_event_data` (and if the delete errors we should just log the error but not report a failure).

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    // casting because u64 is not implemented for sqlx: https://github.com/launchbadge/sqlx/pull/919#discussion_r557256333
+    let funding_fee_as_sat = funding_fee.fee.as_sat() as i64;
+    // TODO: these seem to be redundant and should be in `cfds` table only
+    let maker_lock_amount = dlc.maker_lock_amount.as_sat() as i64;
+    let taker_lock_amount = dlc.taker_lock_amount.as_sat() as i64;
+
+    let query_result = sqlx::query(
+        r#"
+            insert into rollover_completed_event_data (
```
da-kami: 🙃 I find the name of this table weird - I think it is a result of what currently exists and does not reflect what it actually represent. Maybe at some point we can rename it to `current_contract_data` or similar. 
Does not have to be actioned as part of this PR.

sqlite-db/src/load_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,369 @@
...
+    offer_id: OrderId,
+    event_row_id: i64,
+) -> Result<Option<(Dlc, FundingFee)>> {
+    let revoked_commit = load_revoked_commit_transactions(transaction, offer_id).await?;
+    let cets = load_cets(transaction, offer_id).await?;
+
+    let row = sqlx::query!(
+            r#"
+            SELECT
+                main_data.settlement_event_id as "settlement_event_id: model::olivia::BitMexPriceEventId",
```
da-kami: 🙃 I would find it more intuitive to use the table name here, i.e. `rollover_completed_event_data` instead of giving it an alias `main_data` that is not expressive.

da-kami (commented):
Thanks for all the work! Concept ack, a few comments.

klochowicz (commented):
concept ACK.
I was a bit terrified of the amount of code added just for this event, but I was really pleased that from the outside layer is just as easy to use.

sqlite-db/migrations/20220526225235_rollover-completed-table.sql
bonomat: We might get around without it but it ensures that this `cfd_id` actually exists. The idea is that there is only 1 active rollover completed entry in the table, hence the `unique` criteria. 

It might also simplify look ups for rollover data because if you only care about ther rollover details, you can ignore `event_id` and join it only against `cfds`.

What do you think?

sqlite-db/migrations/20220525225235_rename_cets_table.sql
bonomat: Not really random: I'm introducing a new table which are the _actually_ open cets. I'm happy to rename `commit_txs` and `refund_txs` as well

model/src/cfd.rs
```diff
@@ -1913,6 +1917,44 @@ impl From<AdaptorSignature> for secp256k1_zkp::EcdsaAdaptorSignature {
 
 impl_sqlx_type_display_from_str!(AdaptorSignature);
 
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub struct Transaction(bitcoin::Transaction);
```
klochowicz, line 1977: if anything, the old way of  serialising  was not that hard, it just needed the 
`        #[serde(with = "hex_transaction")] Transaction` 🙃

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: I disagree with this, the table is designed that there can only be **one** rollover completed event in the db.  Which means, we need to delete the old entry before being able to insert a new one. 

Having an extra actor which does this on a regular interval makes this unnecessary complex. 

I'm happy to make this infailible and just print an error if you prefer?

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: Why? It only includes `data` for the event `RolloverCompleted`, i.e. it moves `data` from `events` into this table.

sqlite-db/src/load_rollover_completed_event_data.rs
bonomat: Will do!

model/src/cfd.rs
bonomat: Really? 🙈

sqlite-db/migrations/20220525225235_rename_cets_table.sql
bonomat: [`1438b56` (#2140)](https://github.com/itchysats/itchysats/pull/2140/commits/1438b56ebaf1f5697c7e3cb0e0a0746f0030bd86)

sqlite-db/src/insert_rollover_completed_event_data.rs
da-kami: To clarify: I don't think we need to have the `cull` actor, but I think `insert` should not do the delete. 
I think the delete should be a separate call from the outside. I don't like making it infallible on the inside - I don't think it's a good API. What's the problem of having two calls on the outsides? Then the caller can decide what to do in case there was a problem.

Concerning 

> the table is designed that there can only be one rollover completed event in the db.

Does that mean you *have* to delete before you can insert?
I am not against it, but I feel that goes beyond the scope of this PR. This changes the constraint that we only remove data if we were able to save the new one. Doing this in a transaction mitigates that, but the problem that you might fail the insert on deleting remains. How would you make deleting infallible?

sqlite-db/src/insert_rollover_completed_event_data.rs
da-kami: Ah right, that's what you meant in the PR description. Fair enough then!

sqlite-db/src/load_rollover_completed_event_data.rs
bonomat: [`84afd77` (#2140)](https://github.com/itchysats/itchysats/pull/2140/commits/84afd77b10e004d0aa0f9f8afb7c68f4907102b9)

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: I looked into it, imho this can't fail unless the db connection fails. Since this is all in a `transaction`, the whole `insert` should fail.

sqlite-db/migrations/20220526225235_rollover-completed-table.sql
da-kami: hm... it might help us with disassociating it from the event eventually if we ever want to do so. I am fine with keeping it like this, but we have to be aware that this is a pretty hard restriction. This literally means that we can only have one rollover per cfd in the database. 

Given our discussion below (https://github.com/itchysats/itchysats/pull/2140#discussion_r887440179) wouldn't it make more sense to just do an `update` on the table instead of a `delete` and then `insert`?

bonomat: I wanted to have an `update` but could not find an easy way of doing this in a single query because we need to differentiate between `insert` (if no rollover exists so far) and `update` (if this is the second rollover).

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: How would it behave differently if you have two API functions outside?

bonomat: > How would you make deleting infallible?

The sql query is infailible, meaning, if there is nothing to delete, then nothing will be deleted :) This query should only fail if the connection to the db fails. 
Which, imho, is a severe error and we should try again? 
But this is something we would need to deal from the outside.

sqlite-db/migrations/20220526225235_rollover-completed-table.sql
bonomat: > hm... it might help us with disassociating it from the event eventually if we ever want to do so. I am fine with keeping it like this, but we have to be aware that this is a pretty hard restriction. This literally means that we can only have one rollover per cfd in the database.

We need to because of our st** event sourcing and aggregates where we load all events and apply them one after each other. Because of this, we need to know which RolloverCompleted event has the actual data attached.  We could do this with a simple `max` or `latest` on a list, but this feels like a hack to me. With a fk relation this is guaranteed that these two rows belong together. 

If we get rid of event sourcing, we can get rid of this fk easily :).

model/src/cfd.rs
bonomat: ah, but then I'd need to use serde for serializing it which will give me a json right?

bonomat: Boy, how did I survive all these years without knowing this ...

bonomat: downside is that it serializes it to a json string, i.e. including `""`. I'll leave it as is if this is fine?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Migrate away from JSON blobs in the DB to a more normalized database
```
luckysori: :wrench: I guess we should specify that this is only for `RolloverCompleted` at the moment.

sqlite-db/migrations/20220526225235_rollover-completed-table.sql
```diff
@@ -0,0 +1,51 @@
...
+);
+CREATE TABLE IF NOT EXISTS open_cets (
+    id integer PRIMARY KEY autoincrement,
+    cfd_id integer NOT NULL,
+    adaptor_sig TEXT NOT NULL,
+    maker_amount TEXT NOT NULL,
+    taker_amount TEXT NOT NULL,
+    n_bits TEXT NOT NULL,
+    range_end TEXT NOT NULL,
+    range_start TEXT NOT NULL,
```
luckysori: :question: Shouldn't these be integers?

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+    event: CfdEvent,
+) -> Result<()> {
+    let event_kind = event.event;
+    match event_kind {
+        EventKind::RolloverCompleted {
+            dlc: Some(dlc),
+            funding_fee,
+        } => {
+            // we begin a transaction so that it rollsback in case of a failure further down
+            let mut inner_transaction = connection.begin().await?;
```
luckysori: :upside_down_face: I would argue these comments are redundant.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+use model::RevokedCommit;
+use sqlx::pool::PoolConnection;
+use sqlx::Connection as SqlxConnection;
+use sqlx::Sqlite;
+use sqlx::Transaction;
+
+pub async fn insert_rollover_completed_event(
+    connection: &mut PoolConnection<Sqlite>,
+    event_id: i64,
+    event: CfdEvent,
```
luckysori: :upside_down_face: It would be good to define a dedicated type so that we don't need to handle all other events inside this function.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+            for revoked in dlc.revoked_commit {
+                insert_revoked_commit_transaction(&mut inner_transaction, event.id, revoked)
+                    .await?;
+            }
+
+            for (_event_id, cets) in dlc.cets {
+                // ignoring event_id for now because we only have one event
+                for cet in cets {
+                    insert_cet(&mut inner_transaction, event.id, cet).await?;
+                }
```
luckysori: :thinking: If this is gonna affect the table design I would prefer if we already handled the possibility of multiple events, because we're gonna be using them soon.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+    .bind(dlc.publish_pk_counterparty) //publish_pk_counterparty
+    .bind(dlc.revocation) //revocation_secret
+    .bind(dlc.revocation_pk_counterparty) //revocation_pk_counterparty
+    .bind(lock_tx) //lock_tx
+    .bind(lock_tx_descriptor.to_string()) //lock_tx_descriptor
+    .bind(commit_tx) //commit_tx
+    .bind(commit_adaptor_signature) //commit_adaptor_signature
+    .bind(commit_descriptor.to_string()) //commit_descriptor
+    .bind(refund_tx) //refund_tx
+    .bind(refund_signature.to_string()) //refund_tx
```
luckysori: :wrench: I guess we don't need these inline comments.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+    funding_fee: FundingFee,
+    offer_id: OrderId,
+) -> Result<()> {
+    let (lock_tx, lock_tx_descriptor) = dlc.lock.clone();
+    let (commit_tx, commit_adaptor_signature, commit_descriptor) = dlc.commit.clone();
+    let (refund_tx, refund_signature) = dlc.refund.clone();
+
+    // casting because u64 is not implemented for sqlx: https://github.com/launchbadge/sqlx/pull/919#discussion_r557256333
+    let funding_fee_as_sat = funding_fee.fee.as_sat() as i64;
+    // TODO: these seem to be redundant and should be in `cfds` table only
```
luckysori: :thinking: I see what you mean, but I'm not sure it's gonna be convenient to separate the `Dlc` from the `lock_amount`s. I don't think I would bother with this TODO.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+    let maker_lock_amount = dlc.maker_lock_amount.as_sat() as i64;
+    let taker_lock_amount = dlc.taker_lock_amount.as_sat() as i64;
+
+    let query_result = sqlx::query(
```
luckysori: :wrench: I would use the `query!` macro all across the board. It gives us compile-time guarantees, as opposed to `query`.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,304 @@
...
+}
+
+async fn insert_cet(
+    db_transaction: &mut Transaction<'_, Sqlite>,
+    offer_id: OrderId,
+    cet: Cet,
+) -> Result<()> {
+    let query_result = sqlx::query(
+        r#"
+                insert into open_cets (
```
luckysori: :question: Rename this to `cets`?

model/src/lib.rs
```diff
@@ -683,8 +683,8 @@ impl Default for OpeningFee {
 #[derive(Copy, Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub struct FundingFee {
     #[serde(with = "bdk::bitcoin::util::amount::serde::as_sat")]
-    fee: Amount,
-    rate: FundingRate,
+    pub fee: Amount,
+    pub rate: FundingRate,
```
luckysori, line 687: :upside_down_face: Add accessors in a separate commit?

CHANGELOG.md
luckysori: Also, this commit (2f9c9970a77511f25682b3c0fa1698041b4dbb7e) only adds the API. I think modifying the changelog should go later, and the commit message is inaccurate.

sqlite-db/src/lib.rs
```diff
@@ -42,6 +42,7 @@ pub mod event_log;
 pub mod failed;
 mod impls;
 mod insert_rollover_completed_event_data;
+mod load_rollover_completed_event_data;
```
luckysori: :question: What do you think about having a submodule `rollover`, with two further submodules `load` and `insert`?

sqlite-db/src/load_rollover_completed_event_data.rs
```diff
@@ -0,0 +1,369 @@
...
+use model::RevokedCommit;
+use sqlx::Transaction;
+use std::collections::HashMap;
+use std::ops::RangeInclusive;
+use std::str::FromStr;
+
+/// Load RolloverCompleted event data
+///
+/// Returns Ok(Some(..) if one was found or Ok(None) if none was found.
+/// In error case, it returns Err(..)
```
luckysori: :wrench: Should be `Ok(Some(..))` i.e. missing bracket.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -62,6 +65,40 @@ pub async fn insert_rollover_completed_event(
     Ok(())
 }
 
+/// Inserts RolloverCompleted data and returns the resulting rowid
+async fn delete_rollover_completed_event_data(
```
luckysori: :wrench: Could move it to a separate submodule `rollover/delete.rs`.

sqlite-db/src/insert_rollover_completed_event_data.rs
```diff
@@ -62,6 +65,40 @@ pub async fn insert_rollover_completed_event(
     Ok(())
 }
 
+/// Inserts RolloverCompleted data and returns the resulting rowid
```
luckysori: :wrench: Copy-paste error with the docstring.

luckysori (approved):
Nice work! Pretty nifty solution.

It's obviously a lot more complicated than dealing with raw JSON, so I hope it's worth it!

The thing I like the least about all this is how much of the `model` crate needed to change to make this happen. I'm not 100% sure we're being consistent when using the new types.

I would really like to see https://github.com/itchysats/itchysats/issues/2139 in the next iteration.

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: How do you imagine this?

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: true, this was just for me to know which bind belongs to which field.

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: Why do we need this information in the rolloverevent?

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: good catch. I think we use a weird mix of `query` and `query!` all over the place... I'll change it here.

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: I prefer `open_cets` because we also have `closed_cets`

sqlite-db/src/lib.rs
bonomat: can definitely do that

bonomat: I'm afraid doing a clean rebase is not worth the effort. I'll keep it in a separate commit. But this is a good idea, should have done this in the first place.

[`1e0b87e` (#2140)](https://github.com/itchysats/itchysats/pull/2140/commits/1e0b87ea56c30cf792a8d0046259bad2a261ebe8)

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: done

model/src/lib.rs
bonomat: I'll leave this and hope this will be fixed with #2150

CHANGELOG.md
bonomat: Resolved

sqlite-db/src/insert_rollover_completed_event_data.rs
da-kami: Yeah, fair enough - this is an extension of our other discussion of update vs. delete + insert.
I think the update can be achieved by checking if a record exists. If not insert, if yes then update. But at this point I don't mind much. 
The only thing to note is, that the difference is that we will have a new ID in `rollover_completed_data` because we always insert. Otherwise we would literally have only one associated record where all the fields (besides the `id` and the `cfd_id`) would be updated. 
I don't think it makes much difference at the moment, but might make a difference later, depending on how this evolves. Happy to roll with the current implementation.

da-kami (approved):
Happy to test it on testnet!

> I would really like to see https://github.com/itchysats/itchysats/issues/2139 in the next iteration.

I added it to the next iteration so we don't forget ;)

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: It did affect it slightly so I have changed the design. We now support multiple event ids

sqlite-db/src/insert_rollover_completed_event_data.rs
bonomat: I'll leave this is as a follow-up once I know how you imagine this :)

model/src/cfd.rs
da-kami: I think you can overcome that with using `Hyphenated`

sqlite-db/src/insert_rollover_completed_event_data.rs
luckysori: It's a fair question because rollover happens on top of lock. But it's just convenient to build the `Dlc` based on one event. Otherwise I think aggregates will get more complicated.

bonomat: Until we get rid of the aggregates (😈) we could do a sql join and load this information from the other table. That way we don't store reduntant information and would not have to change the model.

bonomat:
bors retry

PR #2141, opened 2022-06-01
Connection monitor

da-kami:
This started out as a quest for defining a trigger for the `identify` - we should trigger it upon establishing a new connection.

@klochowicz and me agreed that the current abstraction of each actor having to poll the `endpoint` for connection information is problematic, because already with ping we have quite a load of messages on the `endpoint` - and with adding more and more of actors that require this information this is inefficient and cumbersome.

So, we decided to add a `connection_monitor` that acts as a proxy. It monitors connections being established and dropped, as well as listening addresses. This new `connection_monitor` actor is the only actor that polls the endpoint for connection information.
Additionally we came to the conclusion that other actors don't want to poll to get connection information but want to get notified on changes, so we let actors that are interested in changes subscribe to the `connection_monitor` that notifies them about the relevant changes. (i.e. observer pattern)

--- 

With this work in place adding `identify` becomes easy.


xtra-libp2p/src/endpoint_monitor.rs
```diff
@@ -0,0 +1,268 @@
...
+/// All subscribers to the connection monitor
+pub struct Subscribers {
+    connection_established: Vec<Box<dyn MessageChannel<ConnectionsEstablished>>>,
+    connection_dropped: Vec<Box<dyn MessageChannel<ConnectionsDropped>>>,
+    listen_address_added: Vec<Box<dyn MessageChannel<ListenAddressAdded>>>,
+    listen_address_removed: Vec<Box<dyn MessageChannel<ListenAddressRemoved>>>,
+}
+
+/// Convenience factory function to create Subscribers struct
+pub fn subscribers(
```
da-kami: 🛠 I don't see the point of using a factory function here. Why not 

```
impl Subscribers {
 pub fn new(...) -> Self {
    ...
  }
}
```

Or even just making the fields of `Subscribers` `pub` since we always have to set all of them?

xtra-libp2p/src/endpoint_monitor.rs
```diff
@@ -0,0 +1,268 @@
...
+
+pub struct ConnectionsDropped {
+    pub dropped: HashSet<PeerId>,
+}
+
+impl xtra::Message for ConnectionsDropped {
+    type Result = ();
+}
+
+pub struct ListenAddressAdded {
```
da-kami:
```suggestion
pub struct ListenAddressesAdded {
```

it is multiple, same as for the connections

xtra-libp2p/src/endpoint_monitor.rs
```diff
@@ -0,0 +1,268 @@
...
+    pub added: HashSet<Multiaddr>,
+}
+
+impl xtra::Message for ListenAddressAdded {
+    type Result = ();
+}
+
+pub struct ListenAddressRemoved {
```
da-kami:
```suggestion
pub struct ListenAddressesRemoved {
```

da-kami (commented):
@klochowicz I added some comments for the last commit.

xtra-libp2p/src/endpoint_monitor.rs
klochowicz: I would make a constructor / factory function as this makes it easier to produce - otherwise, the likely mismatch in names will make this look really awful in endpoint_monitor constructor :( 

Happy to convert it into a `new()` function!

klochowicz:
> @klochowicz I added some comments for the last commit.

thanks for having a look! actioned and pushed a commit.

xtra-libp2p-ping/src/lib.rs
```diff
@@ -95,10 +104,39 @@ mod tests {
...
         );
 
+        let (supervisor, _connection) = supervisor::Actor::with_policy(
+            {
+                let ping_address = ping_address.clone();
+                let endpoint_address = endpoint_address.clone();
+                move || {
+                    endpoint_monitor::Actor::new(
+                        endpoint_address.clone(),
+                        endpoint_monitor::Subscribers::new(
```
da-kami, line 114: 🙃 nit: For readability outside of an IDE it would actually be better to have just the `struct` initialization without constructor because then you see which field is set to what.

i.e.:

```
                          endpoint_monitor::Subscribers {
                            connection_established: vec![xtra::message_channel::MessageChannel::clone_channel(
                                &ping_address,
                            )],
                            connection_dropped: vec![xtra::message_channel::MessageChannel::clone_channel(
                                &ping_address,
                            )],
                            listen_address_added: vec![],
                            listen_address_removed: vec![],
                        },
```

klochowicz: My preference is to optimise for the IDE :p We still have type safety, so they can't be miswired  (except when not plugging it at all, which can always be the case).
I don't mind the other way too much either though, so if you feel strongly about it, please go ahead 🙃

maker/src/actor_system.rs
```diff
@@ -164,10 +167,33 @@ where
...
+                    endpoint_monitor::Actor::new(
+                        endpoint_addr.clone(),
+                        endpoint_monitor::Subscribers::new(
+                            vec![xtra::message_channel::MessageChannel::clone_channel(
+                                &ping_address,
+                            )],
+                            vec![xtra::message_channel::MessageChannel::clone_channel(
+                                &ping_address,
+                            )],
+                            vec![],
```
da-kami: ⚠️ @klochowicz isn't the listener actor missing here?

klochowicz: thanks for spotting, preparing a fixup right now!

klochowicz (approved):
LGTM! 
I'm excited to test it.

I also plan to add some unit tests to prove the behaviour, but that can be done in a follow-up PR.

xtra-libp2p/src/connection_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+            self.stop_reason = Some(e);
+            ctx.stop();
+        }
+    }
+}
+
+#[derive(thiserror::Error, Copy, Clone, Debug)]
+pub enum Error {
+    #[error("Endpoint actor is disconnected")]
+    NoEndpoint,
```
luckysori: :wrench: This can hold the `xtra::Error`. Then we can simply use a `?` instead of having to map the error.

luckysori: It also provides more context in case we use our `timeout` feature on `xtra`: the endpoint could either be `Disconnected` or `TimedOut`.

xtra-libp2p/src/connection_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+        for subscriber in &self.connection_dropped_subscribers {
+            if let Err(e) = subscriber
+                .send(ConnectionsDropped {
+                    peers: connections_dropped.clone(),
+                })
+                .await
+            {
+                tracing::error!(
+                    "Unable to reach a subscriber that is supposed to be supervised: {e:#}"
+                );
```
luckysori: :wrench: I would drop the "is supposed to be supervised". Not sure how this component knows that.

daemon/src/lib.rs
```diff
@@ -224,15 +227,36 @@ where
...
+                        &dialer_actor,
+                    )],
+                    vec![xtra::message_channel::MessageChannel::clone_channel(
+                        &dialer_actor,
+                    )],
+                )
+            }
+        };
+
+        let (supervisor, _connection) = supervisor::Actor::with_policy(
```
luckysori, line 258: :upside_down_face: If we are using a trivial restart policy I think we can use `supervisor::Actor::new` like we do here:

https://github.com/itchysats/itchysats/blob/bbdf369c3cfe75ce03776ea85c80a309d37fe485/maker/src/actor_system.rs#L168-L170

xtra-libp2p/src/connection_monitor.rs
```diff
@@ -0,0 +1,178 @@
...
+use tokio_tasks::Tasks;
+use xtra::message_channel::MessageChannel;
+use xtra::Address;
+use xtra_productivity::xtra_productivity;
+use xtras::SendInterval;
+
+/// How often we check for connection changes in the endpoint
+pub const CHECK_INTERVAL: Duration = Duration::from_secs(5);
+
+/// xtra actor that takes care of dialing (connecting) to an Endpoint.
```
luckysori: :question: Isn't this actor only involved with monitoring new connections/disconnections and relaying that information to subscribers? I don't think this docstring is accurate.

xtra-libp2p/src/connection_monitor.rs
```diff
@@ -0,0 +1,178 @@
+use crate::ConnectionStats;
```
luckysori: :upside_down_face: Would be good to add some unit tests to this actor.

xtra-libp2p/src/connection_monitor.rs
luckysori: Actually this docstring is just copy-pasted from the `dialer::Actor` :laughing:

xtra-libp2p/src/dialer.rs
```diff
@@ -1,107 +1,124 @@
...
+            .connect_address
+            .clone()
+            .extract_peer_id()
+            .ok_or(Error::InvalidPeerId)
+        {
+            Ok(peer_id) => self.peer_id = Some(peer_id),
+            Err(e) => {
+                self.stop_with_error(e, ctx);
+            }
+        }
```
luckysori, line 71: :question: Why can't we make `new` fallible instead? There shouldn't be any rules around that.

xtra-libp2p/src/dialer.rs
```diff
@@ -1,107 +1,124 @@
...
+        // tokio::time::sleep(CONNECTION_TIMEOUT).await;
+        //
+        // if !self.connected {
+        //     self.stop_with_error(
+        //         Error::Failed {
+        //             source: anyhow!("Connection timeout lapsed"),
+        //         },
+        //         ctx,
+        //     );
+        // }
```
luckysori: :wrench: Please remove commented out code from the patch.

xtra-libp2p/src/dialer.rs
```diff
@@ -1,107 +1,124 @@
...
 }
 
 impl Actor {
     pub fn new(endpoint: Address<Endpoint>, connect_address: Multiaddr) -> Self {
         Self {
-            tasks: Tasks::default(),
             endpoint,
             connect_address,
+            connected: false,
+            peer_id: None,
```
luckysori, line 37: :wrench: It's not so clear whose peer ID this is. Rename to `listener_peer_id` perhaps?

xtra-libp2p/src/dialer.rs
```diff
@@ -1,107 +1,124 @@
...
         self.stop_reason.unwrap_or(Error::Unspecified)
     }
 }
 
+impl Actor {
+    fn peer_id(&self) -> PeerId {
+        self.peer_id
+            .expect("to always have peer id if successfully started")
+    }
+}
```
luckysori, line 96: :question: Perhaps not needed?

xtra-libp2p/src/dialer.rs
```diff
@@ -1,107 +1,124 @@
...
 
-    async fn handle(&mut self, _msg: CheckConnection, ctx: &mut xtra::Context<Self>) {
-        if let Err(e) = self.check_connection_active_in_endpoint().await {
-            self.stop_reason = Some(e);
-            ctx.stop();
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, msg: connection_monitor::ConnectionsEstablished) {
+        if msg.peers.contains(&self.peer_id()) {
+            self.connected = true;
```
luckysori, line 120: :question: What's the logic around `connected`?

I think we can remove it. We seem to only want to know if we have disconnected in order to trigger a restart via the `supervisor::Actor`.

xtra-libp2p-ping/src/lib.rs
```diff
@@ -79,7 +82,13 @@ mod tests {
         assert!(!bob_to_alice_latency.is_zero());
     }
 
-    fn create_endpoint_with_ping() -> (PeerId, Address<ping::Actor>, Address<Endpoint>) {
+    #[allow(clippy::type_complexity)]
+    fn create_endpoint_with_ping() -> (
+        PeerId,
+        Address<ping::Actor>,
+        Address<Endpoint>,
+        Address<supervisor::Actor<connection_monitor::Actor, connection_monitor::Error>>,
```
luckysori: :question: Is the supervisor needed for this test?

xtra-libp2p-ping/src/ping.rs
```diff
@@ -150,15 +142,24 @@ impl Actor {
...
-    async fn handle(&mut self, _: UpdateConnectedPeers) {
-        self.connected_peers = match self.endpoint.send(GetConnectionStats).await {
-            Ok(connection_stats) => connection_stats.connected_peers,
-            Err(e) => {
-                tracing::warn!("Cannot ping peers: {e}");
-                return;
-            }
-        };
+        self.connected_peers
+            .retain(|peer_id| !msg.peers.contains(peer_id))
```
luckysori: :upside_down_face: Introducing a variable `dropped_peers` or similar would make the predicate easier to understand.

xtra-libp2p/src/dialer.rs
```diff
@@ -7,6 +7,7 @@ use async_trait::async_trait;
 use libp2p_core::Multiaddr;
```
luckysori, line 7: :clown_face: Lol, please don't use commit messages like that :(

xtra-libp2p/src/dialer.rs
```diff
@@ -71,17 +74,14 @@ impl xtra::Actor for Actor {
...
-        //         ctx,
-        //     );
-        // }
+        let this = ctx.address().expect("self to be alive");
+
+        self.tasks.add(async move {
+            tokio::time::sleep(CONNECTION_TIMEOUT).await;
+            this.send(StopIfNotConnected)
+                .await
+                .expect("to deliver stop message");
```
luckysori, line 82: :upside_down_face: It's not a "stop" message. We could use the generic "self to be alive" or something like that.

xtra-libp2p/src/dialer.rs
```diff
@@ -98,6 +98,17 @@ impl Actor {
...
+    async fn handle(&mut self, _msg: StopIfNotConnected, ctx: &mut xtra::Context<Self>) {
+        if !self.connected {
+            self.stop_with_error(
+                Error::Failed {
+                    source: anyhow!("Did not connect in time"),
+                },
+                ctx,
+            )
+        }
+    }
```
luckysori, line 109: It's unfortunate that we have to do this, but I can't come up with a better alternative right now. This explains why `connected` was introduced in the earlier commit (although it wasn't needed until now).

I guess it's the nature of pushing as opposed to polling. Perhaps a less invasive solution could be to poll the endpoint in `started` directly until `ConnectionStats` holds the peer ID we expect. That should allow us to remove `connected` and this handler.

daemon/src/lib.rs
```diff
@@ -27,8 +27,8 @@ use tokio::sync::watch;
 use tokio_tasks::Tasks;
 use xtra::prelude::*;
 use xtra_bitmex_price_feed::QUOTE_INTERVAL_MINUTES;
-use xtra_libp2p::connection_monitor;
 use xtra_libp2p::dialer;
+use xtra_libp2p::endpoint_monitor;
```
luckysori, line 31: :wrench: Better if we could rename in the original commit that introduces this.

xtra-libp2p/src/endpoint_monitor.rs
```diff
@@ -14,32 +15,51 @@ use xtras::SendInterval;
...
+        }
+    }
+}
+
+/// xtra actor that takes care of monitoring the Endpoint for changes to listen addresses and
+/// connection.
 ///
-/// Periodically polls Endpoint to check whether connection is still active.
+/// Periodically polls Endpoint and dispatches the changes to subscribers.
 /// Should be used in conjunction with supervisor maintaining resilient connection.
```
luckysori, line 46: :wrench: Still irrelevant I think.

xtra-libp2p/src/endpoint_monitor.rs
```diff
@@ -14,32 +15,51 @@ use xtras::SendInterval;
...
 pub struct Actor {
     tasks: Tasks,
     endpoint: Address<Endpoint>,
-    connected_peers: HashSet<PeerId>,
+    last_connection_stats: ConnectionStats,
```
luckysori, line 50: :upside_down_face: I think it should be "latest".

luckysori (approved):
I think it's a very good idea! I'm also excited to see how it works in practice :D

I always review patch by patch and some things were fixed in later patches. I think https://github.com/itchysats/itchysats/pull/2141/commits/65c6b02b3c88a40b2e090ef0ee7d080afb62eaff should be squashed with https://github.com/itchysats/itchysats/pull/2141/commits/427bbf95343b3b957d81d1bb7933abe1183b6309.

xtra-libp2p/src/dialer.rs
da-kami: Oh my, how did that end up in here 😅

xtra-libp2p/src/dialer.rs
klochowicz: I don't think it's possible to find out which commit you were referring to using github UI - please quote the commit or hash in multi-commit PRs, otherwise it's hard to respond!

xtra-libp2p/src/dialer.rs
klochowicz: it's `StopIfNotConnected` message.... I wanted it to be different on purpose, because here we can't really expect "self to be alive" as it's spawned into a long-running task.

xtra-libp2p/src/dialer.rs
klochowicz: It's impossible to make `new` fallible with the current supervisor. I'd be very open to support such construct, as it would reduce some awkwardness in these constructors, but until we support that there's not much I can do 😭

xtra-libp2p/src/connection_monitor.rs
klochowicz: Def keen for that, I actually mentioned that in the daily yesterday, that I'll be adding them in a separate PR shortly (I don't want to block everything else, like the identify and offer, on this PR going in)
Note that this code is tested by all our actor tests, so there's some confidence with it :)

xtra-libp2p/src/connection_monitor.rs
klochowicz: I would prefer to avoid having `xtra::Error` - to me they need to be translated, because otherwise they have no meaning to an outside observer.

daemon/src/lib.rs
klochowicz: it was done by different people at different times - prefer to keep the history.

da-kami:
Sorry to come up with this now, but after thinking about the `endpoint_monitor` overnight I came to the conclusion that this solution is flawed.

The current `endpoint_monitor` will not detect re-connects properly. If the user disconnects and re-connects within the polling interval of the `endpoint_monitor` we will not detect that a new connection was established. (Because the delta will be the same, we never notify about disconnect / connect again).
This can lead to not sending offers to the user, because the plan is to use the notifications in the maker to open the substream to the user to send offers. If the maker does not detect the re-connect it will not open the substream. 
I think this might already be a problem in the current implementation where we poll the connections!

I think the proper solution is to build the notifications about `ConnectionEstablished` and `ConnectionDropped` into the endpoint itself, once it actually happens. Only then can we properly detect re-connects. 

I don't think that building this into the endpoint has downsides besides the endpoint getting a bit more complex (I don't think this is a big deal at this point, it's not especially hard code).
If we build the subscription and notification mechanism into the endpoint we will also be more efficient - we won't have to send around all the `ConnectionStats` anymore; `GetConnectionStats` could just go away.

klochowicz:
> Sorry to come up with this now, but after thinking about the `endpoint_monitor` overnight I came to the conclusion that this solution is flawed.
> 
> The current `endpoint_monitor` will not detect re-connects properly. I the user disconnects and re-connects within the polling interval of the `endpoint_monitor` we will not detect that a new connection was established. (Because the delta will be the same, we never notify about disconnect / connect again). This can lead to not sending offers to the user, because the plan is to use the notifications in the maker to open the substream to the user to send offers. If the maker does not detect the re-connect it will open the substream. I think this might already be a problem in the current implementation where we poll the connections!
> 
> I think the proper solution is to build the notifications about `ConnectionEstablished` and `ConnectionDropped` into the endpoint itself, once it actually happens. Only then can we properly detect re-connects.
> 
> I don't think that building this into the endpoint has downsides besides the endpoint getting a bit more complex (I don't think this is a big deal at this point, it's not especially hard code). If we build the subscription and notification mechanism into the endpoint we will also be more efficient - we won't have to send around all the `ConnectionStats` anymore; `GetConnectionStats` could just go away.

Interesting thoughts, initial response so far:
- I am open to try to cram this into the endpoint itself, as this will reduce the periodic message load a bit further, however: I think your point showcases a flaw in the `offer' protocol more than this - we should be as resilient as possible. in this case, the taker who connects again is in a better position to open a substream again to retrieve the latest state.
- I wouldn't go as far as removing the polling API, because it can be useful (metrics, ephemeral actors etc). 

That said, I think it's worth a shot. Happy to do a POC in the meantime to see how much complex will it get (I bet 'not much', if we pass in a single Subscribers struct).

I'll stop trying to address the review feedback until we settle on this.

da-kami:
> > Sorry to come up with this now, but after thinking about the `endpoint_monitor` overnight I came to the conclusion that this solution is flawed.
> > The current `endpoint_monitor` will not detect re-connects properly. I the user disconnects and re-connects within the polling interval of the `endpoint_monitor` we will not detect that a new connection was established. (Because the delta will be the same, we never notify about disconnect / connect again). This can lead to not sending offers to the user, because the plan is to use the notifications in the maker to open the substream to the user to send offers. If the maker does not detect the re-connect it will open the substream. I think this might already be a problem in the current implementation where we poll the connections!
> > I think the proper solution is to build the notifications about `ConnectionEstablished` and `ConnectionDropped` into the endpoint itself, once it actually happens. Only then can we properly detect re-connects.
> > I don't think that building this into the endpoint has downsides besides the endpoint getting a bit more complex (I don't think this is a big deal at this point, it's not especially hard code). If we build the subscription and notification mechanism into the endpoint we will also be more efficient - we won't have to send around all the `ConnectionStats` anymore; `GetConnectionStats` could just go away.
> 
> Interesting thoughts, initial response so far:
> 
>     * I am open to try to cram this into the endpoint itself, as this will reduce the periodic message load a bit further, however: I think your point showcases a flaw in the `offer' protocol more than this - we should be as resilient as possible. in this case, the taker who connects again is in a better position to open a substream again to retrieve the latest state.
> 
>     * I wouldn't go as far as removing the polling API, because it can be useful (metrics, ephemeral actors etc).
> 
> 
> That said, I think it's worth a shot. Happy to do a POC in the meantime to see how much complex will it get (I bet 'not much', if we pass in a single Subscribers struct).
> 
> I'll stop trying to address the review feedback until we settle on this.

The problem is not only the `offers` protocol, but also `identify`. For `identify` each peer is supposed to poll the identity message upon connection established. But both sides might not notice that a connection was re-established. 
Note that a taker re-connecting might not notice neither because the delta does not change. (Unless I'm missing something)

I think the problem is not the protocols, but that this solution cannot handle re-connects properly. Even if we re-design the protocols to be resilient against that, this is a flaw!

klochowicz:
> The problem is not only the `offers` protocol, but also `identify`. For `identify` each peer is supposed to poll the identity message upon connection established. But both sides might not notice that a connection was re-established. Note that a taker re-connecting might not notice neither because the delta does not change. (Unless I'm missing something)
> 
> I think the problem is not the protocols, but that this solution cannot handle re-connects properly. Even if we re-design the protocols to be resilient against that, this is a flaw!

Identify would not be problematic - it's unlikely that someone would swap to a different version within the polling time.

klochowicz:
Closing this PR in favour of #2159 based on the discussion with @da-kami above - it turned out to be simpler to just add the new capability of subscribing directly into the Endpoint.

xtra-libp2p/src/dialer.rs
klochowicz: I suspect it would be the same level of invasiveness and don't see much problem in tracking the state here based on the nofications?

xtra-libp2p/src/connection_monitor.rs
klochowicz: addressed this in the new PR :)

xtra-libp2p/src/dialer.rs
klochowicz: I'd prefer to have it, because allows us to state the invariant in a single place, and then be used (we know that if we started the actor, it can never be `None`

xtra-libp2p-ping/src/ping.rs
klochowicz: I think we actually have done that.

xtra-libp2p-ping/src/lib.rs
klochowicz: removed in another PR.

xtra-libp2p/src/dialer.rs
klochowicz: the thinking behind was as such:
we should know whether the initial connection has succeeded within the specified timeout - otherwise we might be hanging with a connection attempt.... (this is to cover some edge cases where we could be left hanging, and just trigger a reconnect instead)

After having a second thought, whilst this might be an overkill, it should not make the matter worse in terms of resiliency.

klochowicz:
@luckysori thank you for your valuable feedback! due to creation of the new PR some of the issues became obsolete so I marked them as 'resolved',  I have also addressed some of them in the new PR, and then I have left open the issues I pushed back on. Happy to discuss them here to avoid doubling on work in the new PR.

xtra-libp2p/src/connection_monitor.rs
luckysori: By wrapping it you just get extra context. And the mapping is more natural with just `?`.

xtra-libp2p/src/dialer.rs
luckysori: Good point. Should be easy to support it on the supervisor. I'll have a go.

xtra-libp2p/src/dialer.rs
luckysori: Oops, I might the wrapping in...

...between commit title and message.

xtra-libp2p/src/connection_monitor.rs
klochowicz: aaaaa, I see! thanks for explaining! this file went away in the meantime, so I can't really address it though...

xtra-libp2p/src/dialer.rs
luckysori: To me it seems more natural, predictable and easier to understand to block the startup of the actor until we get feedback from the endpoint that we have a connection.

It would just be one `await` with a timeout.

But not a big deal.

Issue #1266, opened 2022-01-30
Add parcel size as one of the parameters of the offer


klochowicz:
We're currently relying on the implicit assumption that parcel size is `100`:
https://github.com/itchysats/itchysats/blob/master/daemon/src/projection.rs#L788

Exposing the parcel size as one of the parameters sent as a part of the offer is ultimately needed - however, in v0.4 we could probably still live with it being hard-coded to a 100, as long as we check against it.

I was quite surprised that we don't even validate against this assumption currently, as the values chosen in the actor tests don't respect it.

luckysori:
Duplicate of #2660.

PR #2150, opened 2022-06-01
Use `/itchysats/offer` protocol in itchysats applications 

### LINKED ISSUE #1768: Design proposal for `/itchysats/offer` protocol
Tracking issue for libp2p-based network layer: https://github.com/itchysats/itchysats/issues/1595

With the capabilities of multiplexing, we have several choices in how we can send offers from the maker to a taker:

- Push vs pull (i.e. maker pushes by opening a substream vs taker pulls by opening a substream)
- one-shot substream vs streaming multiple offers (one substream per protocol vs reusing the same substream to send another message)

All four combinations are technically possible. The design we are proposing here is the _one-shot push_ combination. In other words: The maker opens a substream to the taker _per offer_ they want to send. We prefer this combination because

- The maker knows when the offer changes, pushing the latest offer is more efficient.
- Opening a new substream per offer is more resilient because there is less state management. If we fail to open a substream to a taker, we can mostly ignore that problem. On the next offer update, we will naturally try again.

At the moment, we send an (optional) short and an (optional) long offer to the taker. We can keep these data structures as is for this migration.

## Optional scope extension

There is an optional scope extension that we may want to contemplate. We have been talking about re-modelling our usage of orders and offers for a while. In particular, we could leave the data structure almost as is but only remove the `OrderId` from it, thus not invalidating an offer just because it is taken by a taker. With https://github.com/itchysats/itchysats/pull/1455, we are re-sending another offer once the current one is taking. This suggests that an offer should not actually get invalidated, the offered price is still valid for the next taker!

Many parts of our system currently rely on the concept of a unique `OrderId` per offer/order. We can still maintain that by designing a compatible `/itchysats/order` protocol that allows peers to place an order with another peer. An order to be placed would come with a unique ID and thus, the invariant of one unique ID per offer/order is still maintained.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/1768.

We run the legacy protocol side by side on the maker for the time being.

---

I decided against tackling #1266 with this PR because it's not strictly necessary and we can always add it in a follow-up PR without needing to create a new version of the protocol (as long as we don't release).


xtra-libp2p-offer/src/maker.rs
```diff
@@ -0,0 +1,112 @@
...
+                .spawner
+                .send_async_safe(SpawnFallible::new(task, err_handler))
+                .await
+            {
+                tracing::warn!("Failed to spawn task to send offers: {e:#}");
+            };
+        }
+    }
+
+    async fn handle(&mut self, _: UpdateConnectedPeers) {
```
da-kami: Since we are keeping the pattern of pushing the offers from maker to taker https://github.com/itchysats/itchysats/pull/2141 is actually relevant for this PR.
With https://github.com/itchysats/itchysats/pull/2141 we will be able to trigger an `offer` update upon a new connection (and not just when upon `NewOffers` message). This will make the UX slightly better, because  `NewOffers` is not guaranteed to happen in a fixed (short) interval. It might be worth rebasing this work on top of https://github.com/itchysats/itchysats/pull/2141

xtra-libp2p-offer/src/taker.rs
```diff
@@ -0,0 +1,71 @@
...
+
+        Self {
+            maker_offers: maker_offers.clone_channel(),
+            spawner,
+        }
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
```
da-kami, line 30: No need to action this - some ideas:

Somehow in my head I had turned the pattern around and would have let the taker open a substream and the maker sending the current offers upon incoming substream. i.e. poll, not push.

This is different to what we have now, but it has the advantage that we could more easily implement a feature where inactive takers "go to sleep" and the maker does not have to bother sending offers to them.
When it comes to resilience and stability I feel it might be more resilient, because each taker tries to get an offer and reconnects and the maker does not have to care about a peer being unavailable at this point. 

Not sure if it's worth thinking about this now though. I'm happy to roll this out using this pattern and potentially change it later.

daemon/src/lib.rs
```diff
@@ -196,12 +196,15 @@ where
             .create(None)
             .spawn(&mut tasks);
 
+        let libp2p_offer_addr = xtra_libp2p_offer::taker::Actor::new(&cfd_actor_addr)
```
da-kami: ❓ Why not supervise this actor?

daemon/src/taker_cfd.rs
```diff
@@ -119,15 +119,20 @@ impl<O, W> Actor<O, W> {
...
+            .projection_actor
             .send(projection::Update(
                 takers_perspective_of_maker_offers.clone(),
             ))
-            .await?;
-
-        Ok(())
+            .await
+        {
+            tracing::warn!("Failed to send current offers to projection actor: {e:#}");
```
da-kami, line 130: I guess it makes sense to change this to not report back to the caller - the protocol actor won't know what to do with this problem.

da-kami (approved):
Nice work! 
I think it would be better to rebase this on top of https://github.com/itchysats/itchysats/pull/2141 before merging (see my more detailed comment below)

xtra-libp2p-offer/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+    use xtra_libp2p::libp2p::identity::Keypair;
+    use xtra_libp2p::libp2p::multiaddr::Protocol;
+    use xtra_libp2p::libp2p::transport::MemoryTransport;
+    use xtra_libp2p::libp2p::Multiaddr;
+    use xtra_libp2p::libp2p::PeerId;
+    use xtra_libp2p::Connect;
+    use xtra_libp2p::Endpoint;
+    use xtra_libp2p::ListenOn;
+    use xtra_productivity::xtra_productivity;
+
```
klochowicz, line 44: Great work with adding tests for the protocol. 🥇

xtra-libp2p-offer/src/maker.rs
klochowicz: exactly, it would be great to use `EndpointMonitor` to stop polling the Endpoint manually (less chance for message slowdown)

xtra-libp2p-offer/src/taker.rs
```diff
@@ -0,0 +1,71 @@
...
+use xtras::spawner::SpawnFallible;
+use xtras::SendAsyncSafe;
+
+pub struct Actor {
+    maker_offers: Box<dyn MessageChannel<LatestMakerOffers>>,
+    spawner: xtra::Address<spawner::Actor>,
+}
+
+impl Actor {
+    pub fn new(maker_offers: &(impl MessageChannel<LatestMakerOffers> + 'static)) -> Self {
```
klochowicz, line 19: nit: why not just clone the channel on the outside and move the `Box`?

Cargo.toml
```diff
@@ -16,6 +16,7 @@ members = [
   "bitmex-stream",
   "xtra-libp2p",
   "xtra-libp2p-ping",
+  "xtra-libp2p-offer",
```
klochowicz, line 19: Given the amount of dependencies, I'm not sure whether extracting this protocol into a separate crate was a good move (contrary to xtr-libp2p-ping for example, which is standalone and readily reusable in different projects).

I guess we might try it out and see how we go, there's no harm in folding it later (or extract others) if we decide too.

It seems like the `itchysats/model` dependency is the troublesome one (it pulls in all sorts of stuff, even 2 versions of `maia`!), perhaps if we decided to keep it standalone, we could create `MakerOffers` locally and a conversion functions in the daemon?

klochowicz (approved):
Looks really cool - I feel inspired to write more tests in other protocols after seeing the amount of tests you've written 👏

xtra-libp2p-offer/src/maker.rs
luckysori: I agree! I just wanted to get something working to open the PR.

PR #2141 is pretty much necessary to merge this, otherwise this introduces the regression that you're hinting at: a newly connected taker may not hear about the latest offers for a while.

This can be seen in the actor tests failing, since they must have been written with the assumption that a new taker will know the latest offers as soon as they connect.

In conclusion, I will wait for #2141 to merge before fixing CI on this PR.

xtra-libp2p-offer/src/taker.rs
luckysori: I followed the design in the proposal as much as possible. It states:

> The maker knows when the offer changes, pushing the latest offer is more efficient.

You bring up some interesting points, but I would say introducing a new design for this protocol is out of the scope of this PR.

xtra-libp2p-offer/src/taker.rs
luckysori: Good question!

I think either approach is fine in this case. We started doing this so I stuck with it as much as possible. From my understanding there are tradeoffs between the two approaches. I learnt about this topic reading this [post](https://www.ncameron.org/blog/dyn-trait-and-impl-trait-in-rust/).

Cargo.toml
luckysori: It requires a bit of boilerplate to extract a crate, but I don't see the downside as long as they are standalone. This new crate is independent of all the code in `daemon`, so it seems unnecessary to burden it with all that code when iterating on it.

> It seems like the itchysats/model dependency is the troublesome one (it pulls in all sorts of stuff, even 2 versions of maia!), perhaps if we decided to keep it standalone, we could create MakerOffers locally and a conversion functions in the daemon?

Thank you for caring about the dependency tree! I think that could be a great idea. I'm happy to look into this before merging.

luckysori: Another note on the motivation to extract this into a crate:

AFAIK, if, for example, we extract two libp2p actors into crates, and the graph looks like this

```
        +- offer -+
       /            \
model-+              +-> daemon
       \            /
        +- order -+
```

not only are we speeding up the development cycle when working on `offer` and `order` in isolation, but we are also tapping into the compiler's ability to parallelise the compilation of `offer` and `order` when working on `daemon`!

Given that, I would recommend to extract `rollover` into a separate crate.

da-kami:
Note: we moved the functionality of `endpoint_monitor` into the endpoint (reason see comments in #2141 ) with PR #2159

Cargo.toml
klochowicz: I would love to see some benchmarks to prove this - I suspect we might actually be slower due to linking time (and this one will hit us more often than the compilation from scratch - due to incremental compilation it's not really an issue). case in point: when I applied your branch, it took actually longer than if it wasn't a separate crate, because it had to link all the dependencies again....

I'm also not too happy about the flat structure of the general and really specific crates being mixed on the top level of our workspace :/ I'm not sure how to mitigate this though.

overall, I don't think the theoretical upsides are worth the time invested. 

> Given that, I would recommend to extract rollover into a separate crate. 

I feel that we should have reached a consensus before extracting this - I'm a bit worried about such needless inconsistency in our code. We can extract the crates later if we see fit and then benchmark whether this was actually beneficial and settle this once for all.

klochowicz: > Another note on the motivation to extract this into a crate:
> 
> AFAIK, if, for example, we extract two libp2p actors into crates, and the graph looks like this
> 
> ```
>         +- offer -+
>        /            \
> model-+              +-> daemon
>        \            /
>         +- order -+
> ```
> 
> not only are we speeding up the development cycle when working on `offer` and `order` in isolation, but we are also tapping into the compiler's ability to parallelise the compilation of `offer` and `order` when working on `daemon`!
> 
> Given that, I would recommend to extract `rollover` into a separate crate.

It's actually more like this 😐

```
        +- offer -+
       /            \
model-+  ----------- +-> daemon
       \            /
        +- order -+

luckysori: Yes, removing the dependency between the actors and `model` would allow us to parallelise the build of `model`, `offer` and `order`.

That doesn't take away from the fact that building `model`, then `offer` and `order` in parallel, then `daemon`, should be faster than `model` -> `daemon`.

klochowicz: > Yes, removing the dependency between the actors and `model` would allow us to parallelise the build of `model`, `offer` and `order`.
> 
> That doesn't take away from the fact that building `model`, then `offer` and `order` in parallel, then `daemon`, should be faster than `model` -> `daemon`.

again, I'd love to see some results. given we're having incremental compilation, I don't think this is worth the time investment.
I feel that it's a premature optimisation, and I'm not too keen on it due to all the boilerplate needed, lack of consistency in pattern usage (unless all of them are migrated), lack of reusability of a crate (for me this actually constitutes a crate, and pulling in `model` makes it not reusable :/), and polluting top-level workspace directory (maybe we could work around this one somehow by putting in subdirs).
All in all, this can be moved *after* we have libp2p in. If you're keen, you could show the speedups on rollover example.

luckysori: I think it's enough of a gain that one can iterate on the standalone crate without having to build the huge `daemon` crate.
My other argument is just icing on top (and a generally good approach to reduce workspace compile times overall).

> I feel that it's a premature optimisation

If we don't exercise good habits, we're gonna pay the price down the line. We don't want to have to prioritise reducing compilation times for one of our future iterations in a few months. We have been appending things to `daemon` by default because it is very easy to do. I think extracting this crate is just a small step in the right direction.

>  all the boilerplate needed

A `Cargo.toml` file? Even the exercise of extracting this into a crate allows us to think about our dependencies and start questioning them. For example, you came up with considering if we even need to depend on `model`. Had I defaulted to adding all this to `daemon`, we may have never had that thought.

> lack of consistency in pattern usage (unless all of them are migrated)

I am not gonna be convinced to keep adding things to `daemon` just because it is "what we do". We have to start somewhere.

> lack of reusability of a crate

We could reuse these actors differently if we ever decide to rewrite `itchysats`.

> polluting top-level workspace directory

You could argue the same thing about the top-level of the `daemon` crate.

klochowicz: Sorry, but I think what you perceive as "good habits" is quite arbitrary. I am not for this change precisely because I think it might not be a change in a good direction in terms of structure. 
There are more variables than compilation speed.

Can we please not conflate adding new features with arguing about workplace structure? It feels that it mandates a separate discussions / proposals / meeting and *benchmarks*.

This discussion is as such out of scope of this ticket and IMHO consistency is most important for maintainability sake. I'm happy to move it elsewhere and then work together on the best structure moving forward.

luckysori: > Sorry, but I think what you perceive as "good habits" is quite arbitrary. I am not for this change precisely because I think it's not a change in a good direction in terms of structure.
>
> Can we please not conflate adding new features with arguing about workplace structure? It feels that it mandates a separate discussions / proposals / meeting and benchmarks.

We can take this discussion elsewhere. I have simply been replying to your comment on the PR. I thought it was important to address it because, well, I like to address all feedback. And because I'm pretty sure this is a good change.

> One last argument for this: The recent moves of half of the files from daemon caused so much wasted time for me personally due to the new lack of consistency introduced that I can guarantee that I wasted more that I'd get back with a clean compilation.

I would like to hear more about what exactly was the problem there, because I can't relate.

klochowicz: > I would like to hear more about what exactly was the problem there, because I can't relate.

one actor system etc being *inside* the daemon, and one outside. inconsistency that trips me every time!

luckysori: It's true that we haven't done the same thing for the `taker` yet, so it's not ideal. Would it help if I just did it so that we can have consistency again?

The split is not arbitrary, those are actors and code that only run on the `maker` binary.

klochowicz: > It's true that we haven't done the same thing for the `taker` yet, so it's not ideal. Would it help if I just did it so that we can have consistency again?
> 
> The split is not arbitrary, those are actors and code that only run on the `maker` binary.

I never said that split was arbitrary, I believe it has merits. I only stated that because of a half-done change we are lacking consistency and it is objectively harder to find stuff, because the same actors/structs for different binaries are in completely different crates. I find this situation unfortunate as I often spend much more time trying to find stuff ever since it went out-of-sync.

Please keep in mind that I am not *against* refactors such like that, I would like to just minimise the state of flux if possible, because otherwise we run the risk of a lingering half-done refactor that makes the codebase less maintainable and will be a nightmare for newcomers to understand
I believe  separating larger refactors from feature work (this is everything I'm trying to argue for in this ticket) is one way to do it. Then we can at least create a ticket,  discuss and implement them, potentially parallelising the work :-)
it would be also easier to benchmark by applying/reverting such a commit.

luckysori: I can accept that the change you are referring to wasn't ideal. With it I was hoping to move the codebase in a better direction (subjective or not this is what I believed at the time and still believe). I prefer to work on refactors in patches before or after introducing features because otherwise it is very hard to find time to do them. 

What I take away from this discussion is that the particular refactor we are talking about was controversial and should have been done at a different time, ideally in full (including the taker side).

I honestly cannot relate to the struggles you are voicing. But I can understand that others might have a different experience to mine. And it's objectively inconsistent when it could easily be consistent.

Now, the change proposed here has little to do with all that. It's not a refactor and it doesn't move any files around that could confuse anyone. I have explained why I think it's a good choice, but at this stage I no longer care.

If anyone is keen to take this over they are welcome to cram it into the `daemon` crate. I'm unassigning myself because I won't be able to push this forward for the next iteration.

da-kami:
Rebased on latest `master`, using `ConnectionEstablished` and `ConnectionDropped` messages now. Will see to address the reviews tomorrow.

Cargo.toml
klochowicz: Let's not throw away the baby with the bathwater here: 
-  I have never said that this other example was in bad direction, I just said it left things in inconsistent state for ages (potentially forever, as there are no TODOs for that, nor issues on github). That example was only pointed out briefly to showcase the downsides of inconsistency, nothing else - my apologies if it derailed the conversation and made it longer than it could have been. I agree with that direction and would be happy to share the load in rolling it out completely.
- doing refactors before or after features is great, I believe this is the right way to go in order not to be overwhelmed by technical debt.

What I had a problem with was doing structural changes amongst the same PR if they could be done in separate steps, especially when extracting it required more work. (and now backtracking to daemon would require even more, so it's not that obvious which path to take!). Perhaps we just have different definitions of what a refactor is, or a threshold for confusion is different (e.g. for me, 1 of the 4 itchysats-specific protocols being outside the daemon is confusing, really). 

Anyway, the work done by you here is awesome - both the protocol implementation along with the comprehensive test suite. I am saddened that we dragged this discussion past the point of caring - I take it that I failed conveying my reasons, and should have backtracked earlier.

luckysori: I think we have actually reached an agreement on the extraction of part of `daemon` into `maker` being subpar (not the idea necessarily, but certainly the execution). Moving forward I will be more careful to not introduce major inconsistencies because it's absolutely true that they can be a bit confusing, particularly to new devs.

I shouldn't have said that I no longer care. It was an emotional response that I should have kept to myself. Sorry about that. The problem is that it's gonna be much harder to extract other protocols and I don't want to keep pushing for it if it's gonna result in annoying inconsistencies and we don't have consensus.

For that reason I would actually prefer it if we put this work back into `daemon` at this stage.

xtra-libp2p-offer/src/taker.rs
klochowicz: thanks for the link, I'll have a look into it today - I was just curious :)

klochowicz (approved):
cool, still looks good! thanks for rebasing using the subscribers @da-kami 👍 
I'm happy for it go in as is (separate crate), it can be factored back later on if we decide to do so - but at least we can start testing this earlier!

klochowicz (approved):
thanks for making the actor supervised! :) still happy with it.

da-kami:
bors retry

PR #2152, opened 2022-06-02
Revert "Taker checks balance before taking an offer"

da-kami:
Reverts itchysats/itchysats#2129

---

Unfortunately this can have very weird side effects on the `taker` side.
I am not totally sure why we run into the problems, but it seems that the way `GetBalance` is implemented does not pick up the balance from mempool. (This is my guess based on the behaviour we saw...)
This can result to a balance being displayed in the UI, but when opening a position we get an `Insufficient funds` error in the UI. 
This can likely be solved, but for now it's easier to revert this to have more stable master.


bonomat (approved):

luckysori (approved):
Ah, what a shame! At least we tried.

da-kami:
Things we may want to look into:



> Ah, what a shame! At least we tried.

I'm planning to give this another go. I think our wallet is not acting sane. But that requires some more investigation

PR #2159, opened 2022-06-03
Add notification capabilities to Endpoint

klochowicz:
Furthermore, we can now push connection information (such as connected peer or
listen addresses) to interested subscribers instead of polling this information.

Roll out the usage in the existing actors, so they don't need to poll Endpoint continuously.

This helps to overcome the problem that all actors interested in connection
information bombard the endpoint with `GetConnectionStats`.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -28,6 +28,7 @@ use tokio_tasks::Tasks;
 use xtra::prelude::*;
 use xtra_bitmex_price_feed::QUOTE_INTERVAL_MINUTES;
 use xtra_libp2p::dialer;
+use xtra_libp2p::endpoint;
 use xtra_libp2p::multiaddress_ext::MultiaddrExt;
 use xtra_libp2p::Endpoint;
 use xtra_libp2p_ping::pong;
@@ -216,6 +217,13 @@ where
 
         tasks.add(oracle_ctx.run(oracle_constructor(executor.clone())));
 
+        let dialer_constructor =
+            { move || dialer::Actor::new(endpoint_addr.clone(), maker_multiaddr.clone()) };
+        let (supervisor, dialer_actor) = supervisor::Actor::with_policy(
+            dialer_constructor,
+            |_: &dialer::Error| true, // always restart dialer actor
+        );
+
         let pong_address = pong::Actor::default().create(None).spawn(&mut tasks);
         let endpoint = Endpoint::new(
             TokioTcpConfig::new(),
@@ -225,17 +233,19 @@ where
                 xtra_libp2p_ping::PROTOCOL_NAME,
                 xtra::message_channel::StrongMessageChannel::clone_channel(&pong_address),
             )],
+            endpoint::Subscribers::new(
+                vec![xtra::message_channel::MessageChannel::clone_channel(
+                    &dialer_actor,
+                )],
+                vec![xtra::message_channel::MessageChannel::clone_channel(
+                    &dialer_actor,
+                )],
+                vec![],
+                vec![],
+            ),
         );
 
         tasks.add(endpoint_context.run(endpoint));
-
-        let dialer_constructor =
-            { move || dialer::Actor::new(endpoint_addr.clone(), maker_multiaddr.clone()) };
-
-        let (supervisor, _dialer_actor) = supervisor::Actor::with_policy(
-            dialer_constructor,
-            |_: &dialer::Error| true, // always restart dialer actor
-        );
         let dialer_supervisor = supervisor.create(None).spawn(&mut tasks);
 
         let (supervisor, price_feed_actor) = supervisor::Actor::with_policy(
--- a/maker/src/actor_system.rs
+++ b/maker/src/actor_system.rs
@@ -35,6 +35,7 @@ use xtra::Actor;
 use xtra::Address;
 use xtra::Context;
 use xtra::Handler;
+use xtra_libp2p::endpoint;
 use xtra_libp2p::libp2p::Multiaddr;
 use xtra_libp2p::listener;
 use xtra_libp2p::Endpoint;
@@ -146,6 +147,16 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
+        let (ping_supervisor, ping_address) = supervisor::Actor::new({
+            let endpoint_addr = endpoint_addr.clone();
+            move || ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
+        });
+
+        let (listener_supervisor, listener_actor) = supervisor::Actor::with_policy(
+            move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone()),
+            |_: &listener::Error| true, // always restart listener actor
+        );
+
         let endpoint = Endpoint::new(
             TokioTcpConfig::new(),
             identity.libp2p,
@@ -164,23 +175,18 @@ where
                     ),
                 ),
             ],
+            endpoint::Subscribers::new(
+                vec![ping_address.clone_channel()],
+                vec![ping_address.clone_channel()],
+                vec![listener_actor.clone_channel()],
+                vec![listener_actor.clone_channel()],
+            ),
         );
 
         tasks.add(endpoint_context.run(endpoint));
 
-        let (supervisor, _listener_actor) = supervisor::Actor::with_policy(
-            {
-                let endpoint_addr = endpoint_addr.clone();
-                move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone())
-            },
-            |_: &listener::Error| true, // always restart listener actor
-        );
-        let listener_supervisor = supervisor.create(None).spawn(&mut tasks);
-
-        let (supervisor, _ping_address) = supervisor::Actor::new({
-            move || ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
-        });
-        let _ping_supervisor = supervisor.create(None).spawn(&mut tasks);
+        let listener_supervisor = listener_supervisor.create(None).spawn(&mut tasks);
+        let ping_supervisor = ping_supervisor.create(None).spawn(&mut tasks);
 
         tasks.add(
             inc_conn_ctx
@@ -219,7 +225,7 @@ where
             executor,
             _tasks: tasks,
             _listener_supervisor: listener_supervisor,
-            _ping_supervisor,
+            _ping_supervisor: ping_supervisor,
             _position_metrics_actor: position_metrics_actor,
         })
     }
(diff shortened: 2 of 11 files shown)


xtra-libp2p/src/endpoint.rs
```diff
@@ -155,15 +157,42 @@ pub enum Error {
...
-    BadConnection(#[from] yamux::ConnectionError), // TODO(public-api): Consider removing this.
+    BadConnection(#[from] yamux::ConnectionError),
+    // TODO(public-api): Consider removing this.
     #[error("Address {0} does not end with a peer ID")]
     NoPeerIdInAddress(Multiaddr),
     #[error("Either currently connecting or already connected to peer {0}")]
     AlreadyConnected(PeerId),
 }
 
+/// All subscribers to the connection monitor
```
da-kami:
```suggestion
/// Subscribers that get notified on connection changes
///
/// This allows other actors to get notified on connection changes such as a new connection being established or dropped as well as listening addresses being added or removed.
```

xtra-libp2p/src/endpoint.rs
```diff
@@ -492,7 +527,55 @@ impl Endpoint {
...
+
+    async fn notify_connection_dropped(&mut self, peer: PeerId) {
+        for subscriber in &self.subscribers.connection_dropped {
+            if let Err(e) = subscriber.send_async_safe(ConnectionDropped { peer }).await {
+                tracing::error!("Unable to reach subscriber: {e:#}");
+            }
+        }
+    }
+
+    async fn notify_listen_addresses_added(&mut self, added: Multiaddr) {
```
da-kami:
```suggestion
    async fn notify_listen_address_added(&mut self, added: Multiaddr) {
```

xtra-libp2p/src/endpoint.rs
```diff
@@ -492,7 +527,55 @@ impl Endpoint {
...
+                    address: added.clone(),
+                })
+                .await
+            {
+                tracing::error!("Unable to reach subscriber: {e:#}");
+            }
+        }
+    }
+
+    async fn notify_listen_addresses_removed(&mut self, removed: Multiaddr) {
```
da-kami:
```suggestion
    async fn notify_listen_address_removed(&mut self, removed: Multiaddr) {
```

xtra-libp2p/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use libp2p_core::Negotiated;
 use libp2p_core::PeerId;
 
 pub mod dialer;
-mod endpoint;
+pub mod endpoint;
```
da-kami, line 22: Is that actually necessary? How did we avoid that before our changes? :D

da-kami (approved):
Just a few cleanup comments! Looking forward to @luckysori 's opinion on this :)

xtra-libp2p/src/lib.rs
klochowicz: yes. it selectively exported `Endpoint` struct before.

daemon/src/lib.rs
```diff
@@ -216,6 +217,13 @@ where
 
         tasks.add(oracle_ctx.run(oracle_constructor(executor.clone())));
 
+        let dialer_constructor =
+            { move || dialer::Actor::new(endpoint_addr.clone(), maker_multiaddr.clone()) };
+        let (supervisor, dialer_actor) = supervisor::Actor::with_policy(
```
luckysori, line 222: :wrench: Can use `new` constructor given trivial restart policy.

maker/src/actor_system.rs
```diff
@@ -164,23 +175,18 @@ where
...
-            |_: &listener::Error| true, // always restart listener actor
-        );
-        let listener_supervisor = supervisor.create(None).spawn(&mut tasks);
-
-        let (supervisor, _ping_address) = supervisor::Actor::new({
-            move || ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
-        });
-        let _ping_supervisor = supervisor.create(None).spawn(&mut tasks);
+        let listener_supervisor = listener_supervisor.create(None).spawn(&mut tasks);
+        let ping_supervisor = ping_supervisor.create(None).spawn(&mut tasks);
```
luckysori, line 189: :upside_down_face: 

```suggestion
        let _listener_supervisor = listener_supervisor.create(None).spawn(&mut tasks);
        let _ping_supervisor = ping_supervisor.create(None).spawn(&mut tasks);
```

xtra-libp2p/src/endpoint.rs
```diff
@@ -155,15 +157,44 @@ pub enum Error {
     #[error("Timeout in protocol negotiation")]
     NegotiationTimeoutReached,
     #[error("Failed to negotiate protocol")]
-    NegotiationFailed(#[from] NegotiationError), // TODO(public-api): Consider breaking this up.
+    NegotiationFailed(#[from] NegotiationError),
+    // TODO(public-api): Consider breaking this up.
```
luckysori: :wrench: I think this and below were not on purpose.

xtra-libp2p/src/endpoint.rs
```diff
@@ -155,15 +157,44 @@ pub enum Error {
...
 
+/// Subscribers that get notified on connection changes
+///
+/// This allows other actors to get notified on connection changes such as a new connection being established or dropped as well as listening addresses being added or removed.
+#[derive(Default)]
+pub struct Subscribers {
+    connection_established: Vec<Box<dyn MessageChannel<ConnectionEstablished>>>,
+    connection_dropped: Vec<Box<dyn MessageChannel<ConnectionDropped>>>,
+    listen_address_added: Vec<Box<dyn MessageChannel<ListenAddressAdded>>>,
+    listen_address_removed: Vec<Box<dyn MessageChannel<ListenAddressRemoved>>>,
```
luckysori, line 178: :wrench: We should probably use the `fan_out::Actor` pattern here.

xtra-libp2p/src/endpoint.rs
```diff
@@ -210,15 +242,18 @@ impl Endpoint {
...
     }
 
-    fn drop_connection(&mut self, peer: &PeerId) {
+    async fn drop_connection(&mut self, peer: &PeerId) {
         let (mut control, tasks) = match self.controls.remove(peer) {
             None => return,
             Some(control) => control,
         };
 
+        self.notify_connection_dropped(*peer).await;
```
luckysori: :question: Are we sure we don't want to call this inside the task below, after calling `control.close().await`?

xtra-libp2p/src/endpoint.rs
```diff
@@ -492,7 +529,55 @@ impl Endpoint {
...
+}
+
+impl Endpoint {
+    async fn notify_connection_established(&mut self, peer: PeerId) {
+        for subscriber in &self.subscribers.connection_established {
+            if let Err(e) = subscriber
+                .send_async_safe(ConnectionEstablished { peer })
+                .await
+            {
+                tracing::error!("Unable to reach subscriber: {e:#}");
```
luckysori: :thinking: Error seems excessive to me. We don't necessarily know what it means for a subscriber to disappear. A warning should suffice.

luckysori: Same for all others.

xtra-libp2p/src/endpoint.rs
luckysori: I'm not sure how making this async could affect the overall performance of the `Endpoint`.

luckysori (approved):
LGTM. Happy to give this a go.

Only concern is that the extra work `Endpoint` does could be a problem when handling many connections and messages at the same time.

Perhaps we could stress test this using the ping protocol branch I created a while back: https://github.com/itchysats/itchysats/commit/5882fd5851d3a36b886f651edd4a17b58154c44a.

xtra-libp2p/src/endpoint.rs
klochowicz: can you elaborate? these message channels are orthogonal, different entities can subscribe there.

also: `fan_out::Actor` does not exist in the codebase anymore.

daemon/src/lib.rs
klochowicz: There is an open ticket, we actually need to amend this restart policy to fail in some cases. Given this knowledge, I'll leave it as-is to reduce the churn.
My apologies for not writing a TODO there 🙃

maker/src/actor_system.rs
klochowicz: ?? they are used, so I don't think we should mark them as unused?

xtra-libp2p/src/endpoint.rs
klochowicz: fair enough!

xtra-libp2p/src/endpoint.rs
klochowicz: > I'm not sure how making this async could affect the overall performance of the `Endpoint`.

good question. thanks for pointing out that we could do it in the task below - I opted for reverting this function to not be async, and notifying after the drop (this is actually more correct, because don't advertise something as dropped anymore before it actually got dropped).
Thanks!

klochowicz: edit: turns out that we can't notify it from the task, because it needs access to `&self` :/ I opted for moving it *after* the drop for now. if it will become problematic, we'll revisit the pattern.

xtra-libp2p/src/endpoint.rs
luckysori: On second thought, I'm not sure it would add value in this case. We would achieve exactly the same functionality.

I guess the point would be that the `Endpoint` could be oblivious of the fact that it was communicating with multiple actors. And we wouldn't have to iterate through the subscribers here. But it's very natural in this case.

maker/src/actor_system.rs
luckysori: Ah, I thought you just passed them to the `ActorSystem` to be remembered.

da-kami:
> LGTM. Happy to give this a go.
> 
> Only concern is that the extra work `Endpoint` does could be a problem when handling many connections and messages at the same time.
> 
> Perhaps we could stress test this using the ping protocol branch I created a while back: [5882fd5](https://github.com/itchysats/itchysats/commit/5882fd5851d3a36b886f651edd4a17b58154c44a).

I like the idea to stress test this! My expectation: We async dispatch all the notification messages to subscribers, this should not become a problem even with large number of connections.

xtra-libp2p/src/endpoint.rs
da-kami: True in the context of this actor, but knowing the architecture this should not happen. We deliberately expect the subscribers always to be available and don't actively remove them, because it's expected that they are supervised. I would keep this as error because this should not happen.

xtra-libp2p/src/endpoint.rs
da-kami: The point of an async dispatch is that it is non-blocking, i.e. you dispatch and don't wait for any response. If this does not work as expected we generally have a problem. I think think we should be able to trust the framework to behave as expected 😅

PR #2160, opened 2022-06-03
Fallback to legacy settlement if libp2p-based settlement fails

### LINKED ISSUE #2156: Use libp2p protocol for collab settlement if taker upgraded with an open CFD
Similar to rollover, we need to fallback to the older protocol.

klochowicz:
Unlike for rollover, we cannot utilise the presence of peer_id to check whether
it should work - we fallback at all times.

Fixes #2156


da-kami (approved):
I think this is good enough for now. Might see some `debug` spam until the takers upgrade but I think that is acceptable. Collab settlement is not periodic so that's fine.

luckysori (approved):

klochowicz:
bors rebase

PR #2161, opened 2022-06-03
Identify protocol

### LINKED ISSUE #1769: Design proposal for `/itchysats/identify` protocol

da-kami:
Fixes #1769.

Adds the identify protocol. 

---

I did not yet take decisions based on the information received from the other peer. 
Given that we had *some* decision making for the old connection we could consider introducing this for identify as well, but 

I feel this could be done in a follow-up, but we might consider it important before we release identify. Here are some rules that we could implement from the top of my head:

- Dialer could close the connection if listener does not support all protocols as expected. This requires:
    - The dialer to know own supported protocols (we can just pass them in as we do for the listener)
    - Upon receiving the identity information from the listener we can decide to close the connection => Implication?!
    - We would then have to notify the user *somehow* that his daemon version is incompatible (we can likely re-use the existing mechanism; might need some adaption).

---

Done:

- [x] Distinguish maker and taker in the identify message (-> what should the environment of a maker be? should it even be set?). 
    - Note: I just set the `environment` to `Maker` at the moment. This might not be optimal for the future if we see this more in a context of equal peers in the network - but that vision is too blurry to take a proper decision now so I did not invest further time.
- [x] Plug the identify actors into the maker and taker system. Both require the listening and the dialing side because both request the identity from each other.
- [x] add protocol tests similar to @luckysori 's work in https://github.com/itchysats/itchysats/pull/2150
- [x] Figure out how to serialize / deserialize the pubkey (serde with on encode/decode enough?)
- [x] Cleanup other TODOs

### DIFF
--- a/daemon/src/identify/dialer.rs
+++ b/daemon/src/identify/dialer.rs
@@ -1,0 +1,162 @@
+use crate::identify::protocol;
+use crate::identify::PROTOCOL;
+use crate::Environment;
+use async_trait::async_trait;
+use std::collections::HashMap;
+use tokio_extras::spawn_fallible;
+use xtra::Address;
+use xtra::Context;
+use xtra_libp2p::endpoint;
+use xtra_libp2p::libp2p::PeerId;
+use xtra_libp2p::Endpoint;
+use xtra_libp2p::OpenSubstream;
+use xtra_productivity::xtra_productivity;
+
+// TODO: Move NUM_CONNECTIONS_GAUGE to a shared crate and then use it here and
+
+pub struct Actor {
+    endpoint: Address<Endpoint>,
+    peer_infos: HashMap<PeerId, PeerInfo>,
+}
+
+impl Actor {
+    pub fn new(endpoint: Address<Endpoint>) -> Self {
+        Self {
+            endpoint,
+            peer_infos: HashMap::default(),
+        }
+    }
+}
+
+#[async_trait]
+impl xtra::Actor for Actor {
+    type Stop = ();
+
+    async fn stopped(self) -> Self::Stop {}
+}
+
+pub(crate) struct GetPeerInfo(pub PeerId);
+
+pub(crate) struct IdentifyMsgReceived {
+    peer_id: PeerId,
+    identify_msg: protocol::IdentifyMsg,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct PeerInfo {
+    pub wire_version: String,
+    pub daemon_version: String,
+    pub environment: Environment,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("Conversion to peer info failed: {error}")]
+pub struct ConversionError {
+    #[source]
+    error: anyhow::Error,
+}
+
+impl TryFrom<protocol::IdentifyMsg> for PeerInfo {
+    type Error = ConversionError;
+
+    fn try_from(identify_msg: protocol::IdentifyMsg) -> Result<Self, Self::Error> {
+        let peer_info = PeerInfo {
+            wire_version: identify_msg.wire_version(),
+            daemon_version: identify_msg
+                .daemon_version()
+                .map_err(|error| ConversionError { error })?,
+            environment: identify_msg.environment().into(),
+        };
+
+        Ok(peer_info)
+    }
+}
+
+#[xtra_productivity]
+impl Actor {
+    async fn handle(&mut self, msg: GetPeerInfo) -> Option<PeerInfo> {
+        let peer_id = msg.0;
+        self.peer_infos.get(&peer_id).cloned()
+    }
+
+    async fn handle(&mut self, msg: IdentifyMsgReceived) {
+        let peer_id = msg.peer_id;
+        let peer_info = match PeerInfo::try_from(msg.identify_msg.clone()) {
+            Ok(peer_info) => peer_info,
+            Err(e) => {
+                tracing::error!("Peer info discarded {:?}: {e:#}", msg.identify_msg);
+                return;
+            }
+        };
+
+        let wire_version = peer_info.wire_version.clone();
+        let daemon_version = peer_info.daemon_version.clone();
+        let environment = peer_info.environment;
+
+        NUM_LIBP2P_CONNECTIONS_GAUGE
+            .with(&HashMap::from([
+                (WIRE_VERSION_LABEL, wire_version.as_str()),
+                (DAEMON_VERSION_LABEL, daemon_version.as_str()),
+                (ENVIRONMENT_LABEL, environment.to_string().as_str()),
+            ]))
+            .inc();
+
+        tracing::info!(%peer_id, %daemon_version, %environment, %wire_version, "New identify message received");
+        self.peer_infos.insert(peer_id, peer_info);
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle_connections_established(
+        &mut self,
+        msg: endpoint::ConnectionEstablished,
+        ctx: &mut Context<Self>,
+    ) {
+        let peer_id = msg.peer;
+        let endpoint = self.endpoint.clone();
+        let this = ctx.address().expect("we are alive");
+
+        let request_identify_msg_fut = {
+            let this = this.clone();
+            async move {
+                let stream = endpoint
+                    .send(OpenSubstream::single_protocol(peer_id, PROTOCOL))
+                    .await??;
+
+                let identify_msg = protocol::recv(stream).await?;
+
+                this.send(IdentifyMsgReceived {
+                    peer_id,
+                    identify_msg,
+                })
+                .await?;
+
+                anyhow::Ok(())
+            }
+        };
+
+        let err_handler = move |e| async move {
+            tracing::debug!(%peer_id, "Identify protocol failed upon request: {e:#}")
+        };
+
+        spawn_fallible(&this, request_identify_msg_fut, err_handler);
+    }
+
+    async fn handle_connections_dropped(&mut self, msg: endpoint::ConnectionDropped) {
+        let peer_id = msg.peer;
+        tracing::trace!(%peer_id, "Remove peer-info because connection dropped");
+        if let Some(peer_info) = self.peer_infos.remove(&peer_id) {
+            NUM_LIBP2P_CONNECTIONS_GAUGE
+                .with(&HashMap::from([
+                    (WIRE_VERSION_LABEL, peer_info.wire_version.as_str()),
+                    (DAEMON_VERSION_LABEL, peer_info.daemon_version.as_str()),
+                    (
+                        ENVIRONMENT_LABEL,
+                        peer_info.environment.to_string().as_str(),
+                    ),
+                ]))
+                .dec();
+        }
+    }
+}
--- a/daemon/src/identify/dialer.rs
+++ b/daemon/src/identify/dialer.rs
@@ -12,8 +12,6 @@ use xtra_libp2p::Endpoint;
 use xtra_libp2p::OpenSubstream;
 use xtra_productivity::xtra_productivity;
 
-// TODO: Move NUM_CONNECTIONS_GAUGE to a shared crate and then use it here and
-
 pub struct Actor {
     endpoint: Address<Endpoint>,
     peer_infos: HashMap<PeerId, PeerInfo>,
@@ -21,6 +19,8 @@ pub struct Actor {
 
 impl Actor {
     pub fn new(endpoint: Address<Endpoint>) -> Self {
+        NUM_LIBP2P_CONNECTIONS_GAUGE.reset();
+
         Self {
             endpoint,
             peer_infos: HashMap::default(),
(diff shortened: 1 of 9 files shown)


daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,192 @@
...
+                .dec();
+        }
+    }
+}
+
+const WIRE_VERSION_LABEL: &str = "wire_version";
+const DAEMON_VERSION_LABEL: &str = "daemon_version";
+const ENVIRONMENT_LABEL: &str = "environment";
+
+static NUM_LIBP2P_CONNECTIONS_GAUGE: conquer_once::Lazy<prometheus::IntGaugeVec> =
```
da-kami, line 166: Note: we could try to re-use the old gauge (i.e. `NUM_P2P_CONNECTIONS_GAUGE`) in the following way:

1. The maker has to still collect information for old and new connections but uses the same gauge.
2. We remove the old connection setup from the taker as part of this PR.

klochowicz:
@da-kami is this ready for review? it's a bit confusing to see `drop!` commits in there?

daemon/src/identify/dialer.rs
klochowicz: I think this is a good idea. otherwise, it gets unnecessarily complicated on grafana side.

da-kami:
> @da-kami is this ready for review? it's a bit confusing to see `drop!` commits in there?

I started debugging the CI related problems on this branch and added these commits to see if it makes a difference (i.e. if the tests would pass, because before they did and I could not reproduce it locally). I will remove the commits, they are irrelevant. Will rebase and then the tests should pass.

da-kami:
@klochowicz cleaned up and ready for review now. I will do https://github.com/itchysats/itchysats/issues/2307 (compatibility feedback to the UI) in a separate PR on top.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+use std::collections::HashSet;
+use std::time::Duration;
+use tokio_extras::FutureExt;
+
+// Start libp2p based protocols from 0.3.0 since the last wire version was 0.2.1
+const PROTOCOL_VERSION: &str = "0.3.0";
+
+const TIMEOUT: Duration = Duration::from_secs(5);
+
+// TODO: Byte serialization for compliance with identify protocol spec
```
luckysori: 🔧 I don't think we need this TODO. We've already added a field which makes us non-compliant and I'm not sure we gain anything from being compliant anyway.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+use futures::StreamExt;
+use libp2p_core::identity::PublicKey;
+use serde::Deserialize;
+use serde::Serialize;
+use std::collections::HashSet;
+use std::time::Duration;
+use tokio_extras::FutureExt;
+
+// Start libp2p based protocols from 0.3.0 since the last wire version was 0.2.1
+const PROTOCOL_VERSION: &str = "0.3.0";
```
luckysori, line 19: ❓ Should we really try to base this on the previous network layer? I would start from scratch.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    pub fn daemon_version(&self) -> Result<String> {
+        let splitted = self.agent_version.split('/').collect::<Vec<_>>();
+        splitted
+            .get(1)
+            .map(|str| str.to_string())
+            .context("Unable to extract daemon version")
+    }
```
luckysori, line 64: 🔧 Should we use https://docs.rs/semver/latest/semver/struct.Version.html#method.parse here? We already use that dependency elsewhere.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    let peer_identity = framed
+        .next()
+        .timeout(TIMEOUT, |parent| {
+            tracing::debug_span!(parent: parent, "Received peer identity")
+        })
+        .await
+        .context("Waiting for peer identity timed out")?
+        .context("Receive identify message failed")?
+        .context("Failed to decode peer identity")?;
```
luckysori: 🔧 We have a mix of `identity` and `identify` here. I think it should always be identify message.

daemon/src/identify/protocol.rs
luckysori: I'm not entirely sure how we're supposed to use the `protocol_version` field that seems to be inspired by the [spec](https://github.com/libp2p/specs/tree/master/identify#protocolversion).

It seems confusing and not very useful to me. The spec even says it's optional.

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,176 @@
...
+    type Stop = ();
+
+    async fn stopped(self) -> Self::Stop {}
+}
+
+pub(crate) struct GetIdentifyInfo(pub PeerId);
+
+pub(crate) struct PeerIdentityReceived {
+    peer_id: PeerId,
+    peer_identity: protocol::IdentifyMsg,
```
luckysori: 🔧 Otherwise it reads like we have the same field twice.

```suggestion
    identify_msg: protocol::IdentifyMsg,
```

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl TryFrom<protocol::IdentifyMsg> for IdentityInfo {
+    type Error = ConversionError;
+
+    fn try_from(identity_msg: protocol::IdentifyMsg) -> Result<Self, Self::Error> {
+        let identity_info = IdentityInfo {
+            wire_version: identity_msg.wire_version(),
+            daemon_version: identity_msg
+                .daemon_version()
+                .map_err(|error| ConversionError { error })?,
```
luckysori: No need to explicitly map if you annotate line 55 with `from` instead of `source`.

```suggestion
               ?,
```

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,176 @@
...
+                    peer_identity,
+                })
+                .await?;
+
+                anyhow::Ok(())
+            }
+        };
+
+        let err_handler = move |e| async move {
+            tracing::debug!(%peer_id, "Identity protocol failed upon request: {e:#}")
```
luckysori:
```suggestion
            tracing::debug!(%peer_id, "Identify protocol failed upon request: {e:#}")
```

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+const WIRE_VERSION_LABEL: &str = "wire_version";
+const DAEMON_VERSION_LABEL: &str = "daemon_version";
+const ENVIRONMENT_LABEL: &str = "environment";
+
+static NUM_LIBP2P_CONNECTIONS_GAUGE: conquer_once::Lazy<prometheus::IntGaugeVec> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_int_gauge_vec!(
+            "libp2p_connections_total",
```
luckysori, line 169: ❓ Is this the only place where we are tracking the number of connections via libp2p?

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,176 @@
...
+        Ok(identity_info)
+    }
+}
+
+#[xtra_productivity]
+impl Actor {
+    async fn handle(&mut self, msg: GetIdentifyInfo) -> Option<IdentityInfo> {
+        let peer_id = msg.0;
+        self.identities.get(&peer_id).cloned()
+    }
```
luckysori: 🔧 This is never used. Remove the handler and message?

daemon/src/identify/listener.rs
```diff
@@ -0,0 +1,70 @@
...
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, message: NewInboundSubstream, ctx: &mut Context<Self>) {
+        let NewInboundSubstream { stream, peer } = message;
+
+        // TODO: Set observed address according to the address we observed when establishing the
+        //  connection
```
luckysori, line 44: 🔧 So we should fetch it from the `Endpoint` here?

daemon/src/lib.rs
```diff
@@ -414,15 +484,31 @@ pub enum Environment {
     Test,
     Legacy,
     Unknown,
+    Maker,
```
luckysori: ❓ What is a `Maker` environment?

maker/src/actor_system.rs
```diff
@@ -177,30 +231,55 @@ where
...
-            move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone()),
+            {
+                let listen_multiaddr = listen_multiaddr.clone();
+                let endpoint_addr = endpoint_addr.clone();
+                move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone())
+            },
             always_restart_after(RESTART_INTERVAL),
         );
 
+        // TODO: Shouldn't this actor also be supervised?
```
luckysori, line 192: 👍 Yes. And also for the taker.

maker/src/actor_system.rs
```diff
@@ -177,30 +231,55 @@ where
...
         let pong_address = pong::Actor.create(None).spawn(&mut tasks);
 
+        let listen_protocols = ListenProtocols::default();
+        let (identify_listener_supervisor, identify_listener_actor) = Supervisor::new({
+            let identity = identity.libp2p.clone();
+            move || {
+                identify::listener::Actor::new(
+                    daemon::version::version().to_string(),
+                    Environment::Maker,
```
luckysori: 🤔 Just say `Environment::Unknown` if we don't want to bother with makers sharing this information? Adding a dedicated variant for makers seems very strange.

maker/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+extern crate core;
```
luckysori: ❓ ❗ ❓

luckysori (approved):
Some small suggestions.

I would consider extracting the `identify` protocol into a separate crate, but that's just me :)

daemon/src/identify/dialer.rs
da-kami: I am not convonced of this anymore, unless we manage to finish the `order` protocol first and only then change this PR. Otherwise there are too many dependencies to the old connection to be able to remove this cleanly. 
Generally the time invest to re-use the old gauge might not be worth it, might be simpler to adapt the metric in Grafana to the new one and then just remove the complete old gauge.

daemon/src/identify/listener.rs
da-kami: We would need the complete multiaddress (i.e. IP + port + p2p, ...) to satisfy this.
I dug into that for a while, but could not figure out how to actually retrieve it from the network stack. It is currently not exposed in our implementation. We would have to find the point in building up the connection where it is available, I was unable to find it. Given that we are not using this information at the moment (and are not planning to in the near future) I would ignore this for now.

daemon/src/lib.rs
da-kami: The more high level question is: What should the maker specify? 
To some degree the maker's environment is irrelevant and should just not be sent. But this is specific to our current setup. This might be useful for something once there are multiple makers. Not sure. If you have a better idea let me know; I wanted to overcome the decision for now and just introduced this variant.

maker/src/lib.rs
da-kami: I have not idea how that happened, must be an accidental auto-complete of the IDE, thanks for spotting it!

daemon/src/identify/protocol.rs
da-kami: Then I'll remove it. We are not using it at the moment anyway! I included it because it is "recommended", spec says:

> The field is optional but recommended for debugging and statistic purposes.

daemon/src/identify/protocol.rs
da-kami: We could but I don't see the advantage given that we don't do anything with this information besides logging it.

daemon/src/identify/dialer.rs
da-kami: It is used in the tests - I actually removed it in the follow up when I noticed that a different API makes more sense, see #2383

daemon/src/identify/protocol.rs
da-kami: @klochowicz what do you think, remove `protocol_version`?

daemon/src/identify/protocol.rs
da-kami: I tried to clean this up and removed the `identity` names alltogether because it was confusing

daemon/src/identify/dialer.rs
da-kami: Maybe the name is misleading, I tried to mimic the previous functionality, so we are able to track takers and versions. This is actually to some extent independent to the connection as such.

maker/src/actor_system.rs
da-kami: Fair enough, I don't really care much for now 😅
If we ever want to use this for the maker it will require some more design.

maker/src/lib.rs
da-kami: We actually had this in `sqlite` `lib` as well, I removed it from there, or was that on purpose??

da-kami:
closing in favor of https://github.com/itchysats/itchysats/pull/2383 - I think it's actually easier to merge this together

PR #2163, opened 2022-06-03
Remove `timestamp` field from `SettlementProposal`

### LINKED ISSUE #2162: Flaky test on master
Some of the rollover + collab settlement tests in `model::Cfd` are flaky after merging the new collab settlement and the changes to rollover.

What we know so far:

CI keeps failing here:

```
 error: test failed, to rerun pass '-p model --lib'
---- cfd::tests::given_one_rollover_with_positive_rate_then_long_pays_rollover_fee stdout ----
thread 'cfd::tests::given_one_rollover_with_positive_rate_then_long_pays_rollover_fee' panicked at 'called `Result::unwrap()` on an `Err` value: Condition failed: `self.role == Role::Maker &&
    self.settlement_proposal.as_ref() == Some(proposal)`', model/src/cfd.rs:4201:18

---- cfd::tests::given_ongoing_rollover_then_can_start_collaborative_settlement stdout ----
thread 'cfd::tests::given_ongoing_rollover_then_can_start_collaborative_settlement' panicked at 'called `Result::unwrap()` on an `Err` value: Condition failed: `self.role == Role::Maker &&
    self.settlement_proposal.as_ref() == Some(proposal)`', model/src/cfd.rs:4201:18
```

We fail on this unwrap:

https://github.com/itchysats/itchysats/blob/7407915fcab8f663873a1c1276164771674c4998/model/src/cfd.rs#L4198-L4201

This happens when the maker accepts the collab settlement proposal. 
Likely the proposal that the maker creates is somehow different than the one of the taker - I expect the condition that fails is `self.settlement_proposal.as_ref() == Some(proposal)` (the proposal is somehow different).

The maker creates the proposal here (in the test setup this is called right before accept):

https://github.com/itchysats/itchysats/blob/7407915fcab8f663873a1c1276164771674c4998/model/src/cfd.rs#L1053-L1067

It is somehow different than the proposal of of the taker. 
Note: We pass in the `taker` proposal in the test (!):

https://github.com/itchysats/itchysats/blob/7407915fcab8f663873a1c1276164771674c4998/model/src/cfd.rs#L4200

But I think this was on purpose to ensure that the maker has the same values. 

Note: I am unable to reproduce this locally. Since the test only sometimes fails it must be related to something that is not deterministic, i.e. time. I have not figured out what could cause this.

luckysori:
Fixes #2162.

The motivation for this is that we don't want to compare the timestamps generated by taker and maker when deciding if the maker
should accept a settlement proposal.

We could just modify the `PartialEq` implementation of `SettlementProposal`, but it turns out that this field is never used.
We therefore remove it altogether.

We do no modify `wire::taker_to_maker::Settlement::Propose` because this code is on the way out anyway. We simply ignore the `timestamp` field if we receive it over the wire.

---

Because we don't know what caused #2162 and we can't easily reproduce it, we're only guessing that this fixes the problem.

### DIFF
--- a/maker/src/cfd.rs
+++ b/maker/src/cfd.rs
@@ -776,7 +776,7 @@ where
                 order_id,
                 msg:
                     wire::taker_to_maker::Settlement::Propose {
-                        timestamp,
+                        timestamp: _,
                         taker,
                         maker,
                         price,
@@ -787,7 +787,6 @@ where
                         taker_id,
                         SettlementProposal {
                             order_id,
-                            timestamp,
                             taker,
                             maker,
                             price,
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1031,12 +1031,9 @@ impl Cfd {
         current_price: Price,
         n_payouts: usize,
     ) -> Result<(CfdEvent, SettlementTransaction, SettlementProposal)> {
-        anyhow::ensure!(
-            !self.is_in_collaborative_settlement()
-                && self.role == Role::Taker
-                && self.can_settle_collaboratively(),
-            "Failed to propose collaborative settlement"
-        );
+        anyhow::ensure!(!self.is_in_collaborative_settlement());
+        anyhow::ensure!(self.role == Role::Taker);
+        anyhow::ensure!(self.can_settle_collaboratively());
 
         let (collab_settlement_tx, proposal) = self.make_proposal(current_price, n_payouts)?;
 
@@ -1057,12 +1054,9 @@ impl Cfd {
         n_payouts: usize,
         proposed_settlement_transaction: &bitcoin::Transaction,
     ) -> Result<(CfdEvent, SettlementTransaction, SettlementProposal)> {
-        anyhow::ensure!(
-            !self.is_in_collaborative_settlement()
-                && self.role == Role::Maker
-                && self.can_settle_collaboratively(),
-            "Failed to start collaborative settlement"
-        );
+        anyhow::ensure!(!self.is_in_collaborative_settlement());
+        anyhow::ensure!(self.role == Role::Maker);
+        anyhow::ensure!(self.can_settle_collaboratively());
 
         let (settlement_tx, proposal) = self.make_proposal(current_price, n_payouts)?;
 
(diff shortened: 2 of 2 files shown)


maker/src/cfd.rs
```diff
@@ -776,7 +776,7 @@ where
                 order_id,
                 msg:
                     wire::taker_to_maker::Settlement::Propose {
-                        timestamp,
+                        timestamp: _,
```
da-kami, line 779: Wow, we actually don't use this timestamp at all 😅

da-kami (commented):
I don't really understand how this should help with the test being less flaky. 
The test failing is a test in `model::cfd` - it has nothing to do with sending messages over the wire or am I missing something?

I think you might be right, the timestamp might be the problem, but I this PR will not make the test less flaky.
Is this PR just to learn more about the problem if it happens again? 
I don't think we can close the ticket with this yet :)

luckysori:
> I don't really understand how this should help with the test being less flaky. The test failing is a test in `model::cfd` - it has nothing to do with sending messages over the wire or am I missing something?
> 
> I think you might be right, the timestamp might be the problem, but I this PR will not make the test less flaky. Is this PR just to learn more about the problem if it happens again? I don't think we can close the ticket with this yet :)

With https://github.com/itchysats/itchysats/pull/2163/commits/27366a55c7122cf7bbc8456aeabfd809ad8f877b I've removed the `timestamp` from the `SettlementProposal`. This means we can no longer fail if the timestamps differ when we call `accept_collaborative_settlement_proposal` and compare settlement proposals on the maker side (because they don't exist anymore).

model/src/cfd.rs
```diff
@@ -1031,12 +1031,9 @@ impl Cfd {
         current_price: Price,
         n_payouts: usize,
     ) -> Result<(CfdEvent, SettlementTransaction, SettlementProposal)> {
-        anyhow::ensure!(
-            !self.is_in_collaborative_settlement()
-                && self.role == Role::Taker
-                && self.can_settle_collaboratively(),
-            "Failed to propose collaborative settlement"
-        );
+        anyhow::ensure!(!self.is_in_collaborative_settlement());
```
bonomat, line 1033: That bugged me for a while :D Thanks.

bonomat (approved):
Thanks for the cleanup on the `ensure!`, definitely better. 

I’ll bors it so that we can see if this helps. 

bors r+

Issue #1770, opened 2022-04-05
Design proposal for `/itchysats/order`


thomaseizinger:
This proposal is dependent on the scope extension mentioned in https://github.com/itchysats/itchysats/issues/1768.

Placing an order is a core abstraction in the financial industry. Placing an order in a system can result in it being filled at which point the positions of both contract partners are adjusted accordingly. Orders can also remain unfilled. In a marketplace, unfilled orders are gathered in an orderbook. They are the foundation for how the mid-market price is defined for example.

We can move one step closer to such a model through the following design:

- Peers can open a substream with the `/itchysats/order` protocol
- The dialing peer sends an `Order` structure to the listening peer. An `Order` struct contains (at least) the peer's intention (buy/sell), the asset, price, quantity and leverage.
- Because we are placing an order directly with a peer and for improved user experience, the listening peer SHOULD decide straight away whether the are willing to fill this order or not and communicate their response back to the dialer through an `Accept` or `Reject` message.

```mermaid
sequenceDiagram
    Taker->>Maker: Open substream `/itchysats/order`, send `Order` struct
    Maker->>Taker: Accept
```

## Interim solution

As an interm solution, we propose to directly attach the creation of a new position to the `/itchysats/order` protocol. We can change these semantics in a later version of the protocol.

```mermaid
sequenceDiagram
    Taker->>Maker: Open substream `/itchysats/order`, send `Order` struct
    Maker->>Taker: Accept
	Taker->>Taker: Transition to contract-setup
	Maker->>Maker: Transition to contract-setup
	Taker->>Maker: Msg1
	Maker->>Taker: Msg1
	Note over Taker,Maker: More messages ...
```

## Appendix: With combined position

At the moment, itchysats does not have a "combined position". Eventually, it would make sense to track a peer's position and adjust it based on the orders they are placing. Doing so would allow the `/itchysats/order` protocol to _end_ at this point and be followed by a separate protocol that - based on the new balance of each peer's account - aligns the blockchain state with the account state. This look something like this:

```mermaid
sequenceDiagram
	Note over Taker,Maker: Assume the taker is 500 contracts long
    Taker->>Maker: Open substream `/itchysats/order`, buy 200 contracts @ $54000 with x3 leverage
    Maker->>Taker: Accept
	Taker->>Taker: Adjust local position: +200 contracts
	Maker->>Maker: Adjust local position: +200 contracts
	Note over Taker,Maker: +200 contracts, requires adding more margin for maker and taker
[...]

PR #2167, opened 2022-06-05
Introduce `/itchysats/order` protocol

luckysori:
- The key test `taker_takes_order_and_maker_accepts_and_contract_setup` passes. Adapting the rest of the actor tests should be trivial.

- I am not sold at all on using an internal channel over handling the decision at the actor level on the maker. The current design was motivated by the possibility of testing the protocol in isolation, but it's basically impossible given all the coupling with actors in
`daemon`.

- The file `taker_cfd.rs` is not very clean. Conversely, `maker_cfd.rs` gets more complicated because we have to support the
legacy connection for a while longer. The logic there is to first try to accept/reject using the new protocol and if it fails attempt it
over the legacy connection. This is very stupid, but I would rather do something stupid rather than trying to differentiate between different types of CFDs.

- You can see that the new design follows the proposal in https://github.com/itchysats/itchysats/issues/1770. This means that
the taker now generates an `OrderId` when _placing_ the order. We currently reuse the same type for the offer (legacy) and the taker's order, which is a bit confusing (but hard to avoid, because our model wants an `OrderId`). With the new protocol the taker ignores the `OrderId` associated with the offer, but we have to keep it around for legacy connections.

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -369,7 +369,6 @@ impl Taker {
             config.heartbeat_interval,
             Duration::from_secs(10),
             projection_actor,
-            maker_identity,
             maker_multiaddr.clone(),
             Environment::Test,
         )
@@ -459,8 +458,6 @@ macro_rules! wait_next_state {
             taker_cfd.order_id, maker_cfd.order_id,
             "order id mismatch between maker and taker"
         );
-        assert_eq!(taker_cfd.order_id, $id, "unexpected order id in the taker");
-        assert_eq!(maker_cfd.order_id, $id, "unexpected order id in the maker");
     };
     ($id:expr, $maker:expr, $taker:expr, $state:expr) => {
         wait_next_state!($id, $maker, $taker, $state, $state)
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -23,12 +23,11 @@ use daemon_tests::MakerConfig;
 use daemon_tests::Taker;
 use daemon_tests::TakerConfig;
 use model::olivia;
+use model::Contracts;
 use model::Identity;
 use model::Leverage;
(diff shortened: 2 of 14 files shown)


klochowicz:
I assumed we were trying to first migrate our network stack, and only then change the protocols if possible (as we agreed). If we change two things at the same time, we risk much longer debugging time and misattributing the problems to the network stack, not the protocol changes - we have seen that so far in other protocols that were migrated and changed along the way.

As it is fundamentally easier to iterate on a protocol with libp2p, I'll see to rework this changeset to just move the network stack.

luckysori:
> I assumed we were trying to first migrate our network stack, and only then change the protocols if possible (as we agreed). If we change two things at the same time, we risk much longer debugging time and misattributing the problems to the network stack, not the protocol changes - we have seen that so far in other protocols that were migrated and changed along the way.
> 
> 
> 
> As it is fundamentally easier to iterate on a protocol with libp2p, I'll see to rework this changeset to just move the network stack.

I tried to only do contract setup over the new network layer, but I could not come up with anything acceptable. That's why I went down this route. Feel free to give it a go though!

klochowicz:
> > I assumed we were trying to first migrate our network stack, and only then change the protocols if possible (as we agreed). If we change two things at the same time, we risk much longer debugging time and misattributing the problems to the network stack, not the protocol changes - we have seen that so far in other protocols that were migrated and changed along the way.
> > As it is fundamentally easier to iterate on a protocol with libp2p, I'll see to rework this changeset to just move the network stack.
> 
> I tried to only do contract setup over the new network layer, but I could not come up with anything acceptable. That's why I went down this route. Feel free to give it a go though!

fair enough! I suspected there might be some blocker :) I'll timebox it, in the worst case, I'll just move forward with this one as a fallback!

klochowicz:
Closing in favour of #2292 (which is heavily based on this PR).

PR #2170, opened 2022-06-06
xtras: Panic if send_interval fails

klochowicz:
We rely on SendInterval trait to send periodic messages for the whole lifetime
of an actor. If for whatever reason we stop sending periodic messages, we should
panic to trigger spawning a new instance. Otherwise, we risk at continuing
running the actor that *appears* to work on the surface level, but in reality
some of its functionality is crippled.

Note for the reviewers: This will only be merged when I update all the actors to be supervised
(submitting a draft early to get feedback)

### DIFF
--- a/xtras/src/send_interval.rs
+++ b/xtras/src/send_interval.rs
@@ -37,8 +37,6 @@ where
         }
         let type_name = std::any::type_name::<M>();
 
-        tracing::warn!(
-            "Task for periodically sending message {type_name} stopped because actor shut down"
-        );
+        panic!("Task for periodically sending message {type_name} stopped because actor shut down");
     }
 }


klochowicz:
A bit more context: in the last 2 weeks, we have stopped sending out heartbeats because of this 87 times: https://itchysats.grafana.net/d/RvjotM97z/container-logs?orgId=1&var-setup=mainnet&var-level=WARN&var-filter=Task&var-container=maker&from=now-14d&to=now

bonomat (commented):
Are you sure that this is needed? 

If `send` to a local actor is not possible anymore it means that the actor is gone (`Stopped`) which means, the supervisor should be able to realize this and can restart.

luckysori (commented):
I'm also skeptical that this is the problem because of what @bonomat said.

Also it would be unfortunate to need to add a panic :(

klochowicz:
> I'm also skeptical that this is the problem because of what @bonomat said.
> 
> Also it would be unfortunate to need to add a panic :(

Why it would be unfortunate? we're breaking invariants, we should, under no circumstances, continue to operate the actor if this happens

 edit: I guess it depends on the particular actor. in the case of `maker::connection`, the heartbeats are being sent *independently* per taker! somehow this escaped my attention. In this case panic is perhaps not warranted, but IMHO for this case we should rather use use a loop and `add_fallible` construct (will post a separate PR for that)

klochowicz:
> Are you sure that this is needed?
> 
> If `send` to a local actor is not possible anymore it means that the actor is gone (`Stopped`) which means, the supervisor should be able to realize this and can restart.



> Are you sure that this is needed?
> 
> If `send` to a local actor is not possible anymore it means that the actor is gone (`Stopped`) which means, the supervisor should be able to realize this and can restart.

this is not always the case in our recent fork of xtra, now it can happen because of `TimedOut` error, which happens here.
I guess checking for `is_ok()` instead explicit match tripped us :/

you can see it here:
https://itchysats.grafana.net/d/RvjotM97z/container-logs?orgId=1&var-setup=mainnet&var-level=WARN&var-filter=&var-container=maker&from=1653308798000&to=1653308858000

note that this is why we have multiple connections from the same taker without us noticing that it disconnected.

klochowicz:
closing this PR as no consensus reached.

PR #2171, opened 2022-06-06
More libp2p connection logs

da-kami:
We need more information about the connection. These logs will help. 
We could consider downgrading the connection established/dropped to `debug`.

### DIFF
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -347,6 +347,7 @@ impl Endpoint {
                 let _ = this.send(ExistingConnectionFailed { peer, error }).await;
             },
         );
+
         self.controls.insert(peer, (control, tasks));
         self.notify_connection_established(peer).await;
     }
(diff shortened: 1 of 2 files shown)


klochowicz (approved):
LGTM!

xtra-libp2p/src/endpoint.rs
```diff
@@ -534,6 +535,8 @@ impl Endpoint {
 
 impl Endpoint {
     async fn notify_connection_established(&mut self, peer: PeerId) {
+        tracing::info!(%peer, "New connection established");
```
klochowicz: nit: it will be easier grep for these if they look similar when establishing/dropping. 

```suggestion
        tracing::info!(%peer, "Connection established");
```

klochowicz (approved):

bonomat (commented):
This will result in some spam in our logs. I propose to put this behind a target e.g. `target : "connection"` and on `trace` so that it can be easily disabled/enabled.

PR #2173, opened 2022-06-06
Add a timeout to dialing

da-kami:
We don't expect to ever hit this timeout, but we should be resilient against the dialer never reaching the peer and dialing forever.
In case such an (unlikely) scenario is possible we should stop and retry, otherwise it could happen that we are stuck dialing forever.


da-kami:
Sorry for the force push, I noticed that it will be sufficient to just use the `tokio::time::timeout` function instead of pulling in `future_ext`.

bonomat (approved):
Should not do any harm :)

klochowicz (approved):

PR #2174, opened 2022-06-06
Use file-based db for wallet

### LINKED ISSUE #2158: Optimize wallet and/or monitoring
I'm not sure which part of our code base is causing this, but electrs is printing the following warnings: 

```bash
[2022-06-02T07:35:12.036Z INFO  electrs::electrum] your wallet uses less efficient method of querying electrs, consider contacting the developer of your wallet. Reason: blockchain.scripthash.get_history called for unsubscribed scripthash: 2b3ce0a1810e2255eb556532a94273496bdeb705323adcfbd30e7fa673d8ba84
[2022-06-02T07:35:12.044Z INFO  electrs::electrum] your wallet uses less efficient method of querying electrs, consider contacting the developer of your wallet. Reason: blockchain.scripthash.get_history called for unsubscribed scripthash: 1787b9cfe70daf31e0cbebbf684d1565ef866e350e6031ee4051e5094ef99275
[2022-06-02T07:35:12.053Z INFO  electrs::electrum] your wallet uses less efficient method of querying electrs, consider contacting the developer of your wallet. Reason: blockchain.scripthash.get_history called for unsubscribed scripthash: 89be2a62b1ae5b1f74396bcbeb131c888222ec74b305169f81a5402fb0b8f9b7
[2022-06-02T07:35:12.061Z INFO  electrs::electrum] your wallet uses less efficient method of querying electrs, consider contacting the developer of your wallet. Reason: blockchain.scripthash.get_history called for unsubscribed scripthash: 1772e81bbfbe92d289a175b4e366a4312e267c003bf381f662716bf693a846ea
[2022-06-02T07:35:12.069Z INFO  electrs::electrum] your wallet uses less efficient method of querying electrs, consider contacting the developer of your wallet. Reason: blockchain.scripthash.get_history called for unsubscribed scripthash: 0b45a072c427744ffb7bb5df0557c3c40eb4d6ef1c8d6f4fb085fb2823195810
...
```

It's a never ending flood of these logs and electrs needs 100% cpu

bonomat:
This will improve performance because
- we don't have to generate the addresses on every restart, we can just read them from the db
- we don't have to generate 1000 addresses after are restart to miss unspent addresses: if the wallet db grows naturally, i.e. all addresses are cached, we don't have to pre-cache addresses
- we use less memory as we don't have to store the whole db in memory: particularly when having 1000s spent addresses this is a significant amount of data
- sync is faster because we don't have re-fetch the whole raw tx which is stored in the db

Note: if you have used the daemon for some time in the past the first restart can be scary because it will not immediately fetch all addresses. It will take some time until the wallet has caught up because we only sync for 100 addresses at a time.

Fixes #2158 

Note 2: this depends on a bdk fork "https://github.com/coblox/bdk", branch = "sqlite-version-bumps"  because of incompatible dependency versions. I'll see to contribute this upstream

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -9,6 +9,7 @@ use daemon::projection;
 use daemon::seed::RandomSeed;
 use daemon::seed::Seed;
 use daemon::wallet;
+use daemon::wallet::MAKER_WALLET_ID;
 use daemon::HEARTBEAT_INTERVAL;
 use daemon::N_PAYOUTS;
 use maker::routes;
@@ -65,7 +66,15 @@ async fn main() -> Result<()> {
 
     let mut tasks = Tasks::default();
 
-    let (wallet, wallet_feed_receiver) = wallet::Actor::new(opts.network.electrum(), ext_priv_key)?;
+    let mut wallet_dir = data_dir.clone();
+
+    wallet_dir.push(MAKER_WALLET_ID);
+    let (wallet, wallet_feed_receiver) = wallet::Actor::new(
+        opts.network.electrum(),
+        ext_priv_key,
+        wallet_dir,
+        MAKER_WALLET_ID.to_string(),
+    )?;
 
     let wallet = wallet.create(None).spawn(&mut tasks);
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Migrate away from JSON blobs in the DB to a more normalized database for RolloverCompleted events
+- Use sled database for wallet. The wallet file is stored in your data-dir as either `maker-wallet` for the maker and `taker-wallet` for the taker respectively
 
 ## [0.4.20] - 2022-05-26
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,6 +286,7 @@ dependencies = [
  "bdk",
  "rand 0.6.5",
  "secp256k1",
+ "tempfile",
 ]
 
 [[package]]
@@ -1090,6 +1091,15 @@ name = "event-listener"
 version = "2.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77f3309417938f28bf8228fcff79a4a37103981e3e186d2ccd19c74b38f4eb71"
+
+[[package]]
+name = "fastrand"
+version = "1.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3fcf0cee53519c866c09b5de1f6c56ff9d647101f81c1964fa632e148896cdf"
+dependencies = [
+ "instant",
+]
 
 [[package]]
 name = "figment"
@@ -4327,13 +4337,13 @@ dependencies = [
 
 [[package]]
 name = "tempfile"
-version = "3.2.0"
+version = "3.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dac1c663cfc93810f88aed9b8941d48cabf856a1b111c29a40439018d870eb22"
+checksum = "5cdb1ef4eaeeaddc8fbd371e5017057064af0911902ef36b39801f67cc6d79e4"
 dependencies = [
  "cfg-if 1.0.0",
+ "fastrand",
  "libc",
- "rand 0.8.4",
  "redox_syscall",
  "remove_dir_all",
  "winapi 0.3.9",
--- a/bdk-ext/Cargo.toml
+++ b/bdk-ext/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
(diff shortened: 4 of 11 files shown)


Cargo.toml
```diff
@@ -26,3 +26,4 @@ maia = { git = "https://github.com/comit-network/maia", rev = "a963084189e1598ca
 maia-core = { git = "https://github.com/comit-network/maia", rev = "e419cb2cf19bed6ec53eaa3672408a6205a9b705", package = "maia-core" } # Unreleased version, pinned before maia's breaking change
 xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity" } # Unreleased
 electrum-client = { git = "https://github.com/da-kami/rust-electrum-client/", branch = "fix/ignore-empty-lines" }
+bdk = { git = "https://github.com/coblox/bdk", branch = "sqlite-version-bumps" }
```
da-kami: Why not a fork on itchysats?

da-kami (approved):
Can we test it before merging, i.e. take the testnet seed start for first time expect the balance to be `0`. Trigger sync until the balance is picked up. Then restart and the balance should be picked up without triggering another sync (because the addresses are cached in the db).

Cargo.toml
bonomat: No particular reason and we had a fork on coblox already.

bonomat:
> Can we test it before merging, i.e. take the testnet seed start for first time expect the balance to be `0`. Trigger sync until the balance is picked up. Then restart and the balance should be picked up without triggering another sync (because the addresses are cached in the db).

It takes a long time. I'd rather go for: https://github.com/itchysats/itchysats/pull/2184 and start from scratch.

da-kami:
> > Can we test it before merging, i.e. take the testnet seed start for first time expect the balance to be `0`. Trigger sync until the balance is picked up. Then restart and the balance should be picked up without triggering another sync (because the addresses are cached in the db).
> 
> It takes a long time. I'd rather go for: #2184 and start from scratch.

I would have done it to make sure the behavior is as expected. But if you are confident enough to just go with the logs that's fine with me :)

bonomat:
> > > Can we test it before merging, i.e. take the testnet seed start for first time expect the balance to be `0`. Trigger sync until the balance is picked up. Then restart and the balance should be picked up without triggering another sync (because the addresses are cached in the db).
> > 
> > 
> > It takes a long time. I'd rather go for: #2184 and start from scratch.
> 
> I would have done it to make sure the behavior is as expected. But if you are confident enough to just go with the logs that's fine with me :)

I'm actually losing confident right now 🙈

I've been running it over night and the funds were there but now I get errors and can't recover from it: 
```
2022-06-07 00:49:39  WARN daemon::wallet: Syncing failed: Failed to sync wallet: Rusqlite(SqliteFailure(Error { code: DatabaseBusy, extended_code: 5 }, Some("database is locked")))
```

Even after restart the db is `locked`. This is not good

bonomat:
@da-kami / @klochowicz : What do you think of using sled instead? 

- I did not get any errors as with sqlite
- we don't need to read that db 
- we don't get into maintaining a fork for a long time because my version bump was only scheduled for 0.20. 


See https://github.com/itchysats/itchysats/pull/2174/commits/c03babfc6636372c9967297b00c530daf9223a98

da-kami:
If we consider dropping this and staying with the in-memory wallet for now we have to think about how to deal with testnet. Triggering the wallet-sync manually after every restart is not an option there! 
I am running into this problem at the moment and just can't do anything because the balance is `0` :/

da-kami:
Maybe we can re-consider the configurable wallet address cache again?
That would allow us to configure the addresses cache in different environments. More of an in-between solution though.

da-kami:
> @da-kami / @klochowicz : What do you think of using sled instead?
> 
>     * I did not get any errors as with sqlite
> 
>     * we don't need to read that db
> 
>     * we don't get into maintaining a fork for a long time because my version bump was only scheduled for 0.20.
> 
> 
> See [c03babf](https://github.com/itchysats/itchysats/commit/c03babfc6636372c9967297b00c530daf9223a98)

How sure are we that we won't run into similar issues? Maybe try running it longer and see. We can consider going for this but we would have to test it over a longer period on testnet.

bonomat:
@da-kami : cleaned-up the git history and made the test compile. 

The last commit believe it will give us a better test performance. 

Let's give this a try?

klochowicz (approved):
looks good, I'm keen to try it out! 👍

maker/src/main.rs
```diff
@@ -65,7 +65,14 @@ async fn main() -> Result<()> {
 
     let mut tasks = Tasks::default();
 
-    let (wallet, wallet_feed_receiver) = wallet::Actor::new(opts.network.electrum(), ext_priv_key)?;
+    let mut wallet_dir = data_dir.clone();
+    wallet_dir.push("maker-wallet");
```
da-kami: 🙃 nit: We only use it twice but maybe we still want to extract it into a variable to avoid errors in the future.
Same for taker main

da-kami (approved):
Happy to give it a try. Worst case we should be able to go back to the in-memory code without any migration problems.

PR #2184, opened 2022-06-07
Allow maker to provide wallet

### LINKED ISSUE #2183: Decouple wallet seed from identity seed
Using the same seed for deriving identity and wallet seed was done due to simplicity reasons but is a problem because you cannot change your wallet seed without changing your identity. 

We should decouple these and provide a second seed for the wallet. 

I'd only focus on the maker for now.

bonomat:
By allowing the maker to provide a private key as arg we split the identity generation from the wallet

I opted for passing in an xprv key directly instead of a seed. This should allow us to reuse the key in other wallets. 

If you want to try it out, use `wagyu`: https://github.com/AleoHQ/wagyu

e.g. `wagyu bitcoin hd -d "m/84'/0'/0'" -n testnet`

fix #2183

### DIFF
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use clap::StructOpt;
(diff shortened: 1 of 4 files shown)


maker/src/main.rs
```diff
@@ -50,7 +50,17 @@ async fn main() -> Result<()> {
     let seed = RandomSeed::initialize(&data_dir.join("maker_seed")).await?;
 
     let bitcoin_network = opts.network.bitcoin_network();
-    let ext_priv_key = seed.derive_extended_priv_key(bitcoin_network)?;
+
+    let ext_priv_key = match opts.wallet_xprv {
+        Some(wallet_xprv) => {
+            if wallet_xprv.network != bitcoin_network {
+                let is_network = wallet_xprv.network;
```
da-kami:
```suggestion
                let network = wallet_xprv.network;
```

🙃 Reads better in the message below.

da-kami (approved):
Nice!
For understanding purpose: Is the expectation that we start with a fresh wallet (i.e. a fresh key) or that the maker passes the extended private key and derivation path according to the existing wallet?

bonomat:
> For understanding purpose: Is the expectation that we start with a fresh wallet (i.e. a fresh key) or that the maker passes the extended private key and derivation path according to the existing wallet?

I'm not sure I understand your question: 
- The expectation is to start the maker with a fresh wallet key (whenever you want to), e.g. now because we have use the same key for a long time and would need to sync so many addresses to find money
- We use the Bip84 standard internally which tells the user that he can use the same key in a different wallet which supports this standard to find his money (e.g. electrum)

bonomat:
bors retry

PR #2190, opened 2022-06-07
Smoke test separately

bonomat:
By smoke testing the two binaries separately we should be able to see better which of them is failing.

Tested here: https://github.com/bonomat/hermes/runs/6767574562?check_suite_focus=true

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,15 +108,18 @@ jobs:
         run: rustup show
       - uses: Swatinem/rust-cache@v1.4.0
       - run: cargo build --bin maker --bin taker
-      - name: Smoke test ${{ matrix.os }} binary
+      - name: Smoke testing maker for ${{ matrix.os }} binary
         shell: bash
         run: |
           cargo dev-maker &
+          sleep 10s # Wait for binaries to start
+          curl --fail http://localhost:8001/api/alive
+      - name: Smoke testing taker for ${{ matrix.os }} binary
+        shell: bash
+        run: |
           cargo dev-taker &
           sleep 10s # Wait for binaries to start
-
           curl --fail http://localhost:8000/api/alive
-          curl --fail http://localhost:8001/api/alive
 
   daemons_arm_build:
     runs-on: ubuntu-latest


klochowicz (commented):
Would it still make sense? Smoke test used to test whether taker can connect to the maker. 

If we split them, perhaps the taker could try connecting to our testnet setup, ie.`cargo run --bin taker -- testnet` ?

bonomat:
> Would it still make sense? Smoke test used to test whether taker can connect to the maker.
> 
This test does not test that the maker can connect to the taker but only that the binaries are starting up and the http endpoints are available.

> If we split them, perhaps the taker could try connecting to our testnet setup, ie.`cargo run --bin taker -- testnet` ?

We can, given we don't test that they can connect 😅 
If we want to test that the taker can connect, than I'd be against this because we randomly restart our testnet setup which could lead to a flaky CI.

klochowicz:
> > Would it still make sense? Smoke test used to test whether taker can connect to the maker.
> 
> This test does not test that the maker can connect to the taker but only that the binaries are starting up and the http endpoints are available.
> 
> > If we split them, perhaps the taker could try connecting to our testnet setup, ie.`cargo run --bin taker -- testnet` ?
> 
> We can, given we don't test that they can connect 😅 If we want to test that the taker can connect, than I'd be against this because we randomly restart our testnet setup which could lead to a flaky CI.

now I remember: we used to test that they connect in this test - because  we used to bail if the maker was not available. Since then there's no reason for starting them together. My comment above is irrelevant as well :)

.github/workflows/ci.yml
```diff
@@ -108,15 +108,18 @@ jobs:
         run: rustup show
       - uses: Swatinem/rust-cache@v1.4.0
       - run: cargo build --bin maker --bin taker
-      - name: Smoke test ${{ matrix.os }} binary
+      - name: Smoke testing maker for ${{ matrix.os }} binary
         shell: bash
         run: |
           cargo dev-maker &
+          sleep 10s # Wait for binaries to start
```
klochowicz, line 115:
```suggestion
          sleep 10s # Wait for binary to start
```

.github/workflows/ci.yml
```diff
@@ -108,15 +108,18 @@ jobs:
...
+          curl --fail http://localhost:8001/api/alive
+      - name: Smoke testing taker for ${{ matrix.os }} binary
+        shell: bash
+        run: |
           cargo dev-taker &
           sleep 10s # Wait for binaries to start
```
klochowicz, line 121:
```suggestion
          sleep 10s # Wait for binary to start
```

klochowicz (approved):
makes sense, there's no reason to couple them anymore (as per comment)

da-kami (approved):

PR #2197, opened 2022-06-07
Don't sort events by timestamp

### LINKED ISSUE #2196: Don't sort events by timestamp
Note: The problem where we find events with the same timestamp should not happen in production because there is more latency involved. Nevertheless we should remove sorting by timestamp because with recent changes it should be unnecessary.

**Problem**

We sort events before we apply them:

https://github.com/itchysats/itchysats/blob/96e11d6ccf9f8c35568db4c534d2d1807791a69c/sqlite-db/src/lib.rs#L646

But within the testnet setup it can happen that in the database we have collab close entries with the same timestamp.
From testnet `taker_master` db, order-id `17a4515f-905f-4489-a213-0e097f95810a`, cfd_id `144`:

image

This can result in a position being displayed as `Pending Payout` even though the position was actually closed. More generally this can result in other unwanted behaviour because the application is just not prepared for events being processed in the "wrong" order.

We added the sorting step because use process events using parallel iterator (i.e. `into_par_iter`). However, this optimization might not be necessary anymore, because we only needed this before we culled DLCs.

**Potential solution**

- Remove using parallel iterators
- Ensure that the SQL statement orders correctly (better be explicit)
- Remove the sorting step alltogether

da-kami:
fixes https://github.com/itchysats/itchysats/issues/2196

It appears that it can happen that two events in the database have the same timestamp.
This was observed for collab started and completed as well as rollover started and accepted events.
If they have the same timestamp, and the event history is sorted by that timestamp it can happen that the order is incorrect.

Iterating over the list of events should not change any ordering.
Thus, the database query can ensure the order and we don't need an additional sorting step.

---

I remember we discussed that we want to be totally sure that the order is guaranteed that's this sorting step was added. But when looking at the code again I don't understand why we opted for this. We only use iterators - why would an iterator change the order?

### DIFF
--- a/sqlite-db/src/lib.rs
+++ b/sqlite-db/src/lib.rs
@@ -607,6 +607,7 @@ async fn load_cfd_events(
             cfds c on c.id = events.cfd_id
         where
             uuid = $1
+        order by events.id asc
         limit $2,-1
             "#,
         id,
@@ -638,12 +639,10 @@ async fn load_cfd_events(
         }
     }
 
-    let mut events = events
+    let events = events
         .into_iter()
         .map(|(_, event)| event)
         .collect::<Vec<CfdEvent>>();
-
-    events.sort_unstable_by(CfdEvent::chronologically);
 
     Ok(events)
 }
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -53,7 +53,6 @@ use rust_decimal_macros::dec;
 use serde::de::Error as _;
 use serde::Deserialize;
 use serde::Serialize;
-use std::cmp::Ordering;
 use std::collections::HashMap;
 use std::fmt;
 use std::ops::RangeInclusive;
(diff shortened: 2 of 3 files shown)


da-kami:
Even if this problem is likely only relevant for the testnet setup, I think this sorting step is dangerous and we should remove it.

sqlite-db/src/lib.rs
```diff
@@ -638,13 +639,11 @@ async fn load_cfd_events(
         }
     }
 
-    let mut events = events
+    let events = events
         .into_iter()
         .map(|(_, event)| event)
         .collect::<Vec<CfdEvent>>();
 
-    events.sort_unstable_by(CfdEvent::chronologically);
```
bonomat, line 646: The reason why this was introduced was because of the introduction of `into_par_iter`. We were not sure if `into_par_iter` would keep the ordering and hence ordered it again. 

Maybe it's better to also remove `into_par_iter` as well and take into account the non-parallel iter.

da-kami: Thanks for that, I actually had forgotten about his :/ 
Given the recent database changes I think we can actually consider removing the parallel iterator. Converting this to draft until then.

bonomat: I think so too: let's remove the `into_par_iter` and keep the sql sort you added in this commit. That should guarantee us the ordering.

klochowicz:
@da-kami are you planning to finish this PR? It's been in draft stage for a while, and I'm not sure whether I should review it or wait until it matures or fades away 😅

da-kami:
Closing this for now, I think it can be picked up separately.
Corresponding ticket #2197 updated.

PR #2200, opened 2022-06-08
Ensure message order at the end of rollover

### LINKED ISSUE #2199: Collab close and rollover problems on testnet
We ran into problem with CFD `aabb011f-9a22-40bb-b6cc-45c2c2a81fd7`.

## `taker_master` logs

```
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.01898658Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"model::cfd::Cfd","cfd_version":"7","num_events":"0"},"target":"sqlite_db"}
	
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.026144209Z","level":"INFO","fields":{"message":"Appended event to database","event":"CollaborativeSettlementStarted","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"sqlite_db"}
	
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.026448835Z","level":"DEBUG","fields":{"message":"Handled request","method":"POST","path":"/api/cfd/aabb011f-9a22-40bb-b6cc-45c2c2a81fd7/settle","status":"200 OK"},"target":"http"}
	
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.027357261Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"daemon::projection::Cfd","cfd_version":"7","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.035812097Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"daemon::position_metrics::Cfd","cfd_version":"7","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 12:20:53	
{"timestamp":"2022-06-08T02:20:53.140958891Z","level":"ERROR","fields":{"message":"Rollover failed: End of stream while receiving Decision","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"daemon::collab_settlement::protocol"}
	
2022-06-08 12:20:53	
{"timestamp":"2022-06-08T02:20:53.142207492Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"model::cfd::Cfd","cfd_version":"7","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 12:20:53	
{"timestamp":"2022-06-08T02:20:53.142764198Z","level":"WARN","fields":{"message":"Collaborative settlement failed: End of stream while receiving Decision","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"model::cfd"}
	
2022-06-08 12:20:53	
{"timestamp":"2022-06-08T02:20:53.151720153Z","level":"INFO","fields":{"message":"Appended event to database","event":"CollaborativeSettlementFailed","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"sqlite_db"}
```

And later rollover ran into trouble with this CFD:

```
2022-06-08 13:02:05	
{"timestamp":"2022-06-08T03:02:05.174389754Z","level":"ERROR","fields":{"message":"Rollover failed: End of stream while receiving rollover decision from maker","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"daemon::rollover::protocol"}
	
2022-06-08 13:02:05	
{"timestamp":"2022-06-08T03:02:05.175288564Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"model::cfd::Cfd","cfd_version":"9","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 13:02:05	
{"timestamp":"2022-06-08T03:02:05.175717612Z","level":"WARN","fields":{"message":"Rollover failed: End of stream while receiving rollover decision from maker","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"model::cfd"}Show context
	
2022-06-08 13:02:05	
{"timestamp":"2022-06-08T03:02:05.179112123Z","level":"INFO","fields":{"message":"Appended event to database","event":"RolloverFailed","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"sqlite_db"}
```

and then:

```
2022-06-08 13:07:05	
{"timestamp":"2022-06-08T03:07:05.527961805Z","level":"INFO","fields":{"message":"Rollover proposal got accepted","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"daemon::rollover::taker"}Show context
	
2022-06-08 13:07:05	
{"timestamp":"2022-06-08T03:07:05.546760813Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"daemon::position_metrics::Cfd","cfd_version":"12","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 13:07:14	
{"timestamp":"2022-06-08T03:07:14.883172161Z","level":"ERROR","fields":{"message":"Rollover failed: Counterparty sent invalid revocation sk","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"daemon::rollover::protocol"}
	
2022-06-08 13:07:14	
{"timestamp":"2022-06-08T03:07:14.884128743Z","level":"DEBUG","fields":{"message":"Applying new events to CFD","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7","aggregate":"model::cfd::Cfd","cfd_version":"12","num_events":"1"},"target":"sqlite_db"}
	
2022-06-08 13:07:14	
{"timestamp":"2022-06-08T03:07:14.885425984Z","level":"WARN","fields":{"message":"Rollover failed: Counterparty sent invalid revocation sk","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"model::cfd"}
```

It kept failing with this problem on both sides and will be force closed.

## `maker` logs

Collab settlement failed to start error:

```
2022-06-08 12:20:50	
{"timestamp":"2022-06-08T02:20:50.130892495Z","level":"ERROR","fields":{"message":"Rollover failed: Failed to start collab settlement protocol: Failed to execute command on CFD: Proposed collab settlement does not equal locally created one. Local: Transaction { version: 2, lock_time: 0, input: [TxIn { previous_output: OutPoint { txid: f0ebf41aff66609349e80593fe514908477283e1f413904f0e5732c4c93b3feb, vout: 0 }, script_sig: Script(), sequence: 4294967295, witness: [] }], output: [TxOut { value: 328457, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 ba6b63c217bf0ec2938b54ba04739a56d9bfeea0) }, TxOut { value: 158637, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 416777184a21af22328634360c2a2dcfa8f6c3aa) }] }, proposed: Transaction { version: 2, lock_time: 0, input: [TxIn { previous_output: OutPoint { txid: f0ebf41aff66609349e80593fe514908477283e1f413904f0e5732c4c93b3feb, vout: 0 }, script_sig: Script(), sequence: 4294967295, witness: [] }], output: [TxOut { value: 328455, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 ba6b63c217bf0ec2938b54ba04739a56d9bfeea0) }, TxOut { value: 158639, script_pubkey: Script(OP_0 OP_PUSHBYTES_20 416777184a21af22328634360c2a2dcfa8f6c3aa) }] }","order_id":"aabb011f-9a22-40bb-b6cc-45c2c2a81fd7"},"target":"daemon::collab_settlement::protocol"}
```

da-kami:
fixes: https://github.com/itchysats/itchysats/issues/2199

This should mitigate a problem related to libp2p-rollover, where the taker closed a substream too early.
If both parties send and then wait for the other party in parallel it is unclear who will receive first.
It can happen that the taker sent and received and closed the stream, while the maker is still reading the message that was sent by the taker. This can lead to a panic because the substream is already close.
By ensuring that the taker first sends and then waits to receive, and the maker first waits to receive and then sends should solve this problem.

This is backwards compatible, because old takers will do the same as before (first send then wait to receive).

---

Note: This is not the most clean solution. We should consider re-designing contract setup and rollover to generally have ensured message order; but then we would have to split the function into maker and taker. 
Distinguishing by role is somewhat ugly given that we could just distinguish in the maker and taker actors. But given that legacy and libp2p use the same code here this was the easier fix for now.

---

Note: We can decide for changing this for `contract setup` as well, but I have not so far. Once we change it into a libp2p protocol it will have to be changed.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -665,16 +665,35 @@ pub async fn roll_over(
 
     // TODO: Remove send- and receiving ACK messages once we are able to handle incomplete DLC
     // monitoring
-    sink.send(RolloverMsg::Msg3(RolloverMsg3))
-        .await
-        .context("Failed to send Msg3")?;
-    let _ = stream
-        .next()
-        .timeout(ROLLOVER_MSG_TIMEOUT)
-        .await
-        .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
-        .context("Empty stream instead of Msg3")?
-        .try_into_msg3()?;
+    match our_role {
+        Role::Taker => {
+            // Taker sends first and then waits for maker to send so we don't accidentally close the
+            // substream too early
+            sink.send(RolloverMsg::Msg3(RolloverMsg3))
+                .await
+                .context("Failed to send Msg3")?;
+            let _ = stream
+                .next()
+                .timeout(ROLLOVER_MSG_TIMEOUT)
+                .await
+                .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
+                .context("Empty stream instead of Msg3")?
+                .try_into_msg3()?;
+        }
+        Role::Maker => {
+            // Maker first waits for taker to send and then sends.
+            let _ = stream
+                .next()
+                .timeout(ROLLOVER_MSG_TIMEOUT)
+                .await
+                .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
+                .context("Empty stream instead of Msg3")?
+                .try_into_msg3()?;
+            sink.send(RolloverMsg::Msg3(RolloverMsg3))
+                .await
+                .context("Failed to send Msg3")?;
+        }
+    }
 
     Ok(Dlc {
         identity: sk,


bonomat (approved):
Why does closing a substream on one side result in a panic on the otherside? This sounds like something we should fix because you can't guarantee that a substream stays open. 

Even though it looks a bit ugly to have this match in here, I think the solution is OK. Next step I'd like to see though is splitting this protocol into two files, one for the maker and for taker. The order of sending and receiving messages does matter in our protocols and I think that's the only way of how to design it properly.

klochowicz:
I am not sure whether this is the issue (see comment on the issue)

klochowicz (changes requested):
I feel like we should stop using `select_next_some()` instead (see the issue). We could gracefully handle this on the new protocol level

da-kami:
After more analysis of https://github.com/itchysats/itchysats/issues/2199 we came to the conclusion that it is unclear if the `panic` is related to the substream being closed by the other party. 

Likely this problem is related to introduction of the `FailureCounter` that had a severe bug. We are reverting the failure counter here: https://github.com/itchysats/itchysats/pull/2207

I'm changing this PR to draft, will close after further monitoring.

da-kami:
> Why does closing a substream on one side result in a panic on the otherside? This sounds like something we should fix because you can't guarantee that a substream stays 

It is likely because of using `select_next_some`. We should not do that, see https://github.com/itchysats/itchysats/issues/2199#issuecomment-1153561917

> Even though it looks a bit ugly to have this match in here, I think the solution is OK. Next step I'd like to see though is splitting this protocol into two files, one for the maker and for taker. The order of sending and receiving messages does matter in our protocols and I think that's the only way of how to design it properly.

Yes! I think we cannot just send and receive in parallel - this was never a good design in the first place. It is error prone and I think at least for the last message we have to ensure the message order, so I think this PR is needed. 
I agree that we should properly model the protocol and split it into two files as follow up!

daemon/src/setup_contract.rs
```diff
@@ -658,15 +658,33 @@ pub async fn roll_over(
...
-        .select_next_some()
-        .timeout(ROLLOVER_MSG_TIMEOUT)
-        .await
-        .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
-        .try_into_msg3()?;
+    match our_role {
+        Role::Taker => {
+            // Taker sends first and then waits for maker to send so we don't accidentally close the
+            // substream too early
+            sink.send(RolloverMsg::Msg3(RolloverMsg3))
```
bonomat, line 672: Should we consider a sending timeout?

bonomat (approved):

luckysori (approved):
I think this patch is harmless and I do like the predictability of knowing who sends what message when, but I don't understand what problem this is fixing. You're saying that we need to actively wait for the other party to finish the protocol before we can drop the substream on our side? That sounds incredibly annoying and brittle.

But I would be keen to see `contract_setup_maker` and `contract_setup_taker` functions, which is where this patch is ultimately taking us.

da-kami:
> I think this patch is harmless and I do like the predictability of knowing who sends what message when, but I don't understand what problem this is fixing. You're saying that we need to actively wait for the other party to finish the protocol before we can drop the substream on our side? That sounds incredibly annoying and brittle.
> 
> But I would be keen to see `contract_setup_maker` and `contract_setup_taker` functions, which is where this patch is ultimately taking us.

I do see the argument that closing the substream should not lead to message loss.
I have been thinking about this a lot, and I cannot explain 100% why this is happening. 
The behavior I see (i.e. whenever the taker closes the substream the maker runs into a problem where the stream returns `Ready(None)`) lead me to create this fix. The fact that this currently causes a panic (due to `select_next_some`) is irrelevant. The stream will still report `Ready(None)`, which will cause the maker to exit with an error while the taker succeeded.

Sending is `async`, but when using `send` this should include a `flush`. As far as I understand this should guarantee that the message is sent. But, I don't know what happens if you send, flush and then wait to receive - while the other party is doing exactly the same. Once you are done receiving you close the stream. But how is your sent message guaranteed to arrive at the other side? 

It might be worth creating a test where we try to reproduce this behaviour in isolation to understand better what is happening.

I do believe that this change will fix that problem, because it is clear who is sending and who is receiving. I don't think we have this uncertainty when implementing it this way.

luckysori:
> > I think this patch is harmless and I do like the predictability of knowing who sends what message when, but I don't understand what problem this is fixing. You're saying that we need to actively wait for the other party to finish the protocol before we can drop the substream on our side? That sounds incredibly annoying and brittle.
> > But I would be keen to see `contract_setup_maker` and `contract_setup_taker` functions, which is where this patch is ultimately taking us.
> 
> I do see the argument that closing the substream should not lead to message loss. I have been thinking about this a lot, and I cannot explain 100% why this is happening. The behavior I see (i.e. whenever the taker closes the substream the maker runs into a problem where the stream returns `Ready(None)`) lead me to create this fix. The fact that this currently causes a panic (due to `select_next_some`) is irrelevant. The stream will still report `Ready(None)`, which will cause the maker to exit with an error while the taker succeeded.
> 
> Sending is `async`, but when using `send` this should include a `flush`. As far as I understand this should guarantee that the message is sent. But, I don't know what happens if you send, flush and then wait to receive - while the other party is doing exactly the same. Once you are done receiving you close the stream. But how is your sent message guaranteed to arrive at the other side?
> 
> It might be worth creating a test where we try to reproduce this behaviour in isolation to understand better what is happening.
> 
> I do believe that this change will fix that problem, because it is clear who is sending and who is receiving. I don't think we have this uncertainty when implementing it this way.

Thanks for the explanation. I completely agree with trying to reproduce the behaviour in a test, because it's not something that I would have expected to happen. But feel free to merge this in the meantime!

da-kami:
I tried to reproduce this in a test but was unable to. I will see to push something later. It might be related to the test setup and dropping the stream (I am not setting the test up over the wire with two binaries...).

Additionally, I also had another look at the `select_next_some` implementation and I think the problem could be related to the first assert where we run into the panic:

```
impl<St: ?Sized + FusedStream + Unpin> Future for SelectNextSome<'_, St> {
    type Output = St::Item;

    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
        assert!(!self.stream.is_terminated(), "SelectNextSome polled after terminated");

        if let Some(item) = ready!(self.stream.poll_next_unpin(cx)) {
            Poll::Ready(item)
        } else {
            debug_assert!(self.stream.is_terminated());
            cx.waker().wake_by_ref();
            Poll::Pending
        }
    }
}
```

The line `assert!(!self.stream.is_terminated(), "SelectNextSome polled after terminated");` asserts prior to actually checking `ready`. This could be error prone - We would have to evaluate when `is_terminated` would result to true and if it is possible for the stream to still yield `ready!` even if `terminated` is set. I was unable to dig that deep yet.
It would be great to get to the very bottom of this to ensure that the stream behaves as expected and we don't run into problems where one side finishes and the other ones tries to read from the terminated stream.

If this is where the problem comes from then it would be contained to using `select_next_some` and we don't have to ensure the message order.
We can still opt for ensuring message order because, at least to my mind, this is better protocol design than what we currently do.

klochowicz (approved):

da-kami:
Update:

The problem is not just related to `select_next_some`, we do see `Empty stream` error messages in the logs after merging the `select_next_some` fix:

```
2022-06-15 16:05:28	maker	
{"timestamp":"2022-06-15T06:05:28.527828138Z","level":"WARN","fields":{"message":"Rollover failed: Empty stream instead of Msg3","order_id":"83bd8ef8-86be-4773-98d0-3d72404e488e"},"target":"model::cfd"}
	
2022-06-15 16:05:28	maker	
{"timestamp":"2022-06-15T06:05:28.525060542Z","level":"ERROR","fields":{"message":"Rollover failed: Empty stream instead of Msg3","order_id":"83bd8ef8-86be-4773-98d0-3d72404e488e"},"target":"daemon::rollover::protocol"}
	
2022-06-15 16:05:28	(no unique labels)	
{"timestamp":"2022-06-15T06:05:28.527828138Z","level":"WARN","fields":{"message":"Rollover failed: Empty stream instead of Msg3","order_id":"83bd8ef8-86be-4773-98d0-3d72404e488e"},"target":"model::cfd"}
	
2022-06-15 16:05:28	(no unique labels)	
{"timestamp":"2022-06-15T06:05:28.525060542Z","level":"ERROR","fields":{"message":"Rollover failed: Empty stream instead of Msg3","order_id":"83bd8ef8-86be-4773-98d0-3d72404e488e"},"target":"daemon::rollover::protocol"}
```

Let's hope that merging this PR will fix this for now. I am working on generally changing the rollover protocol to be more deterministic and let the taker always send first and the maker always wait for a message first.

This lead to exactly the same behavior described in https://github.com/itchysats/itchysats/issues/2199 - just without the panic, i.e. the maker appended rollover failed, but all consecutive rollovers failed. 
Let's hope the message order helps for now. I think splitting and passing around the stream into the `rollover` function might also be part of the problem - but this is just a hunch so far. I think by cleaning this up it might just be more resilient.

PR #2175, opened 2022-06-06
Add a failure counter for opening substreams

da-kami:
fixes https://github.com/itchysats/itchysats/issues/2166

If we fail `FAILURE_LIMIT` times the connection will be dropped.
The failure counter is added independent of the protocols, but we log the protocols for which we had a failure.

@klochowicz I wanted to keep the solution simple so I did not try to implement the counter as part of the substream / connection components. It was not super straight forward.

PR #2207, opened 2022-06-09
Revert "Add a failure counter for opening substreams"

da-kami:
Reverts itchysats/itchysats#2175

This was not well thought through and there was a severe bug that caused random peer connection closure on the failure because we don't count per peer.
Additionally, we were closing connections for peers that just don't speak a certain protocol, which can happen with legacy takers (expected, should not error...).

I kept the fix for the failure counter [here](https://github.com/itchysats/itchysats/tree/fix-failure-counter), but I don't think the whole idea of this counter was good.


klochowicz:
we suspect that this caused most (if not all) issues seen on testnet recently: #2199

PR #2211, opened 2022-06-09
Re-enable libp2p ping from the taker to maker

klochowicz:
This should make our connection more resilient.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -31,6 +31,7 @@ use xtra_libp2p::dialer;
 use xtra_libp2p::endpoint;
 use xtra_libp2p::multiaddress_ext::MultiaddrExt;
 use xtra_libp2p::Endpoint;
+use xtra_libp2p_ping::ping;
 use xtra_libp2p_ping::pong;
 use xtras::supervisor;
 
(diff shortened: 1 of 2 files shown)


daemon/src/lib.rs
```diff
@@ -310,6 +319,7 @@ where
             _archive_failed_cfds_actor: archive_failed_cfds_actor,
             _tasks: tasks,
             maker_online_status_feed_receiver,
+            _pong_actor: pong_address,
```
da-kami, line 324: ❓How did this ever work, if we dropped `pong_address`? - probably it was fine because endpoint never stopped 😅

daemon/src/lib.rs
```diff
@@ -275,6 +279,11 @@ where
 
         tasks.add(endpoint_context.run(endpoint));
 
+        let (supervisor, _ping_address) = supervisor::Actor::new({
+            move || ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL)
+        });
+        let _ping_supervisor = supervisor.create(None).spawn(&mut tasks);
```
da-kami: ❓ Why do we not add this one to `Self`? 

🛠 I think we should keep this one around as well.

da-kami (approved):
See questions before merging please :)

daemon/src/lib.rs
klochowicz: yeah, the endpoint never stopped, and it contains a `StrongMessageChannel`, that increases the refcount.

PR #2214, opened 2022-06-10
Activate debug logs for `xtra_libp2p`

da-kami:
This helps understand what is going on in the endpoint. I think we need this inforamation for the time being. We can alternatively decide to set

### DIFF
--- a/shared-bin/src/logger.rs
+++ b/shared-bin/src/logger.rs
@@ -29,6 +29,7 @@ pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
         .add_directive("multistream_select=warn".parse()?)
         .add_directive("libp2p_noise=warn".parse()?)
         .add_directive("xtra_libp2p_offer=debug".parse()?)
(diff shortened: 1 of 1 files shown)


bonomat (changes requested):
It sounds like this is only for temporary use. If so, I'm against merging this PR and please use the environment variable to enable it:

```
RUST_LOG="xtra_libp2p=debug" 
```

Otherwise we have to push another release to disable it again. 

I would only put log filters here which force the log level higher as they would produce too much spam on debug otherwise.

da-kami:
I can re-phrase is, I think we actually do always want to see this information. Otherwise we don't know enough about he connection.

bonomat:
> I can re-phrase is, I think we actually do always want to see this information. Otherwise we don't know enough about he connection.

Wouldn't it be better to just export it as an environment variable?

da-kami:
https://github.com/itchysats/cloud-conf/pull/83

PR #2216, opened 2022-06-10
Rename error variant for clarity

da-kami:
We only use this error variant for `inflight` connections, i.e. when we are *trying* to connect.
Seeing the error message in the logs is misleading because this error is only relevant when we are trying to connect, not when already connected.

### DIFF
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -162,8 +162,8 @@ pub enum Error {
     BadConnection(#[from] yamux::ConnectionError), // TODO(public-api): Consider removing this.
     #[error("Address {0} does not end with a peer ID")]
     NoPeerIdInAddress(Multiaddr),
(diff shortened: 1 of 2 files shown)


xtra-libp2p/src/endpoint.rs
```diff
@@ -162,8 +162,8 @@ pub enum Error {
     BadConnection(#[from] yamux::ConnectionError), // TODO(public-api): Consider removing this.
     #[error("Address {0} does not end with a peer ID")]
     NoPeerIdInAddress(Multiaddr),
-    #[error("Either currently connecting or already connected to peer {0}")]
-    AlreadyConnected(PeerId),
+    #[error("Either currently connecting to peer {0}")]
```
bonomat: `Either` ...`or` ? 

Sentence is not complete. Did you mean? 

```suggestion
    #[error("Already trying to connect to peer {0}")]
```

PR #2217, opened 2022-06-10
Rework dialer and listener to not assume anything at startup

klochowicz:
In short, supervised actors can be restarted at any time. If they subscribe to changes in the Endpoint, they first need to figure out the current state at startup. 
Establishing this assumption allowed to rework and simplify the code of these actors. 

### DIFF
--- a/xtra-libp2p/src/dialer.rs
+++ b/xtra-libp2p/src/dialer.rs
@@ -47,6 +47,7 @@ impl Actor {
     }
 
     fn stop_with_error(&mut self, e: Error, ctx: &mut xtra::Context<Self>) {
+        tracing::debug!("Stopping dialer with an error: {e:#}");
         self.stop_reason = Some(e);
         ctx.stop();
     }
--- a/xtra-libp2p/src/dialer.rs
+++ b/xtra-libp2p/src/dialer.rs
@@ -2,14 +2,16 @@ use crate::endpoint;
 use crate::multiaddress_ext::MultiaddrExt;
 use crate::Connect;
 use crate::Endpoint;
+use crate::GetConnectionStats;
 use anyhow::anyhow;
+use anyhow::Result;
 use async_trait::async_trait;
 use libp2p_core::Multiaddr;
 use libp2p_core::PeerId;
 use std::time::Duration;
-use tokio_tasks::Tasks;
 use xtra::Address;
 use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncSafe;
 
 /// If we're not connected by this time, stop the actor.
 pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
@@ -19,10 +21,8 @@ pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
 /// Periodically polls Endpoint to check whether connection is still active.
 /// Should be used in conjunction with supervisor maintaining resilient connection.
 pub struct Actor {
-    tasks: Tasks,
     endpoint: Address<Endpoint>,
     connect_address: Multiaddr,
-    connected: bool,
     listener_peer_id: Option<PeerId>,
     stop_reason: Option<Error>,
 }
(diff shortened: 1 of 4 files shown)


xtra-libp2p/src/dialer.rs
```diff
@@ -96,18 +84,42 @@ impl Actor {
...
+        );
+
+        if let Err(e) = self.connect().await {
+            tracing::warn!("Failed to request connection from endpoint: {e:#}");
+        }
+
+        // Only check the connection again after it had enough time to be established
+        tokio::time::sleep(CONNECTION_TIMEOUT).await;
+
+        anyhow::ensure!(
```
da-kami, line 111: Should this really be an ensure? This changes the constraints. I would have seen it as a `warn` log if we could not establish the connection and I would keep a faster retry from the dialer (i.e. 5 secs)

xtra-libp2p/src/dialer.rs
```diff
@@ -2,38 +2,37 @@ use crate::endpoint;
...
 use libp2p_core::PeerId;
 use std::time::Duration;
-use tokio_tasks::Tasks;
 use xtra::Address;
 use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncSafe;
 
 /// If we're not connected by this time, stop the actor.
-pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
+pub const CONNECTION_TIMEOUT: Duration = Duration::from_secs(20);
```
da-kami: I don't think this is necessary. We should potentially try more often then every `20` seconds. 

If we trigger this and the endpoint is still waiting the endpoint will just return an error hat we are still trying to connect.
But it's fine for the dialer to check again after 5 secs.

Aligning this could mean that retrying to connect takes 20 seconds.

Consider this:

1. The dialer triggers a `Connect` in the endpoint.
2. The endpoint fails to connect after 6 seconds. 
3. The dialer restarts only after 20... 

If the dialer would restart with the 5 secs it would catch such a scenario faster, and it should not do harm.

da-kami (commented):
My comments are not blocking, but I do think we should not align the timeouts between dialer and endpoint. This was different for a reason.

The concept of just using `GetConnectionStats` and `ConnectionDropped`: ack
The concept of using `send_async_safe` instead of tasks: ack

xtra-libp2p/src/dialer.rs
klochowicz: what's the changed constraint here? 

- the original behaviour: we wait, then we check, if we're not connected we stop the actor
- subscriber, v1: we spawn a task in the background that checks after the same timeout, and if we're not connected by then, we stop the actor.
- this PR: we wait, check, and if we're not connected we stop the actor (same thing really).

What changed a bit is how the actor is stopped, and that might have caused the confusion:
- original version: we dispatch an internal message to stop the actor
- subscribers, v1: we have a bunch of `if let Err(e)` that calls the `stop_with_error()` method
- this PR: I extracted an internal method, so I could use `if let Err(e)` once for the resulting anyhow error.

IMHO the second ensure is the crucial one, because this is the one that covers trying again. I don't see need for faster retry? (if you want, we can drop the alignment of the timeouts, so it always just tries every 5s, I can be convinced to that one :) )

xtra-libp2p/src/dialer.rs
klochowicz: I will revert the last commit, as 5s was catching the scenario faster; as such, it probably should have not been included in this PR.

we can add the more elaborate reconnection scenario in a separate PR.

xtra-libp2p/src/dialer.rs
```diff
@@ -96,18 +84,42 @@ impl Actor {
...
-        if !self.connected {
-            self.stop_with_error(
-                Error::Failed {
-                    source: anyhow!("Did not connect in time"),
-                },
-                ctx,
-            )
+    async fn handle(&mut self, _msg: Dial, ctx: &mut xtra::Context<Self>) {
+        if let Err(e) = self.dial().await {
+            self.stop_with_error(Error::Failed { source: e }, ctx);
```
da-kami, line 123: Maybe add a log with the error here so we know what failed :)
Or let the supervisor log it.

da-kami (approved):

xtra-libp2p/src/dialer.rs
klochowicz: added inside `stop_with_error` function, both on dialer and listener.

da-kami (approved):

PR #2220, opened 2022-06-10
Ensure restart of protocol actors knows connections

da-kami:
follow up of https://github.com/itchysats/itchysats/pull/2217

### DIFF
--- a/xtra-libp2p-ping/src/ping.rs
+++ b/xtra-libp2p-ping/src/ping.rs
@@ -13,6 +13,7 @@ use xtra::Context;
 use xtra_libp2p::endpoint;
 use xtra_libp2p::libp2p::PeerId;
 use xtra_libp2p::Endpoint;
+use xtra_libp2p::GetConnectionStats;
 use xtra_libp2p::OpenSubstream;
 use xtra_productivity::xtra_productivity;
 use xtras::spawner;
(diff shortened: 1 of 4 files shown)


xtra-libp2p-ping/src/ping.rs
```diff
@@ -66,6 +67,18 @@ impl xtra::Actor for Actor {
...
 
+        match self.endpoint.send(GetConnectionStats).await {
+            Ok(connection_stats) => self
+                .connected_peers
+                .extend(connection_stats.connected_peers),
+            Err(e) => {
+                tracing::error!(
+                    "Unable to receive connection stats from the endpoint upon startup: {e:#}"
+                );
+                ctx.stop();
```
bonomat, line 82: Will the actor be immediately restarted? 

If so, and for some reason `GetConnectionStats` keeps erroring, we will bombard the `endpoint` with `GetConnectionStats` requests until our CPU starts to glow 😅

da-kami: We had this consideration when discussing the implementation. The `endpoint` is designed to be always available. It is supervised. we should *always* be able to query `GetConnectionStats` from the `endpoint. 

I am happy to add a `sleep` there to adhere to defensive coding. I don't think we can do much else unless we want to change the design more radically. The protocol actors are designed to be "always available" as well, i.e. they are also supervised.
Would a sleep before restarting the actor be OK for you? 

We could actually make the sleep part of the restart policy.

da-kami: > We could actually make the sleep part of the restart policy.

I take it back - we might not want that. This would mean that we would `sleep` in all cases before restarting the actor. I will introduce a specific `sleep` for this specific case.

bonomat: It was more a question but I think the sleep is ok.

bonomat (approved):

xtra-libp2p-ping/src/ping.rs
klochowicz: > > We could actually make the sleep part of the restart policy.
> 
> I take it back - we might not want that. This would mean that we would `sleep` in all cases before restarting the actor. I will introduce a specific `sleep` for this specific case.

it might still be part of the restart policy (e.g. if an actor got stopped with error `A`, start it again straight away, but if with the error `B`, then sleep).
That would require stopping defining a custom `Error` type, just like in e.g. `xtra_libp2p::Dialer` though...(something to think about for the future)

klochowicz (approved):

xtra-libp2p-ping/src/ping.rs
da-kami: Good point with the  custom error, I think that would be a bit more explicit. Given that we have to implement it for both actors I don't see much advantage there though, so I'm merging. If we have other error cases for these actors that require a sleep we can extract it into the policy.

PR #2221, opened 2022-06-13
Distinguish network for bitmex price feed

da-kami:
For the `testnet` and `signet` setup we should use query the `testnet` price and not `mainnet`.

### DIFF
--- a/bitmex-stream/src/lib.rs
+++ b/bitmex-stream/src/lib.rs
@@ -14,12 +14,14 @@ pub use tokio_tungstenite::tungstenite::Error;
 /// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
 pub fn subscribe<const N: usize>(
     topics: [String; N],
+    network: Network,
 ) -> impl Stream<Item = Result<String, Error>> + Unpin {
     let stream = stream! {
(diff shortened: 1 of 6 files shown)


bonomat (approved):
I think this makes sense.

bitmex-stream/src/lib.rs
```diff
@@ -63,3 +65,18 @@ pub fn subscribe<const N: usize>(
...
+pub enum Network {
+    Mainnet,
+    Testnet,
+}
+
+impl Network {
+    pub fn to_url(&self) -> String {
+        match self {
+            Network::Mainnet => "www.bitmex.com".to_string(),
+            Network::Testnet => "testnet.bitmex.com".to_string(),
```
klochowicz, line 79: nit: `www.` prefix only in one type of network? it seems like it's probably not needed in mainnet either?

klochowicz (approved):

bitmex-stream/src/lib.rs
da-kami: I tried without and it is needed. I don't have an easy answer why.

bonomat: The subdomain of `www` is  not voluntarily, it's the official url they are running their API on :D 

See https://testnet.bitmex.com/api/explorer/ vs https://www.bitmex.com/api/explorer/

PR #2231, opened 2022-06-14
Activate info logs for `xtras` crate

da-kami:
We actually don't see when supervised actors are restarted in the logs at the moment.
I do think we want to see that.

---

Note: Recently we started splitting the system into multiple crates. That's OK, but I feel we did not properly configure logging for all these crates.


bonomat (approved):

klochowicz (approved):

luckysori (approved):
I think it would be nice if we could configure everything to be on `Info` by default, but not sure if it's possible.

da-kami:
> I think it would be nice if we could configure everything to be on `Info` by default, but not sure if it's possible.

I like the idea, could you open a ticket for that? I think it makes sense to have a look at the logging configuration overall and improve it. I remember that we tried different ways for defining these rules in comit-rs and xmr-btc before but I don't remember if we ever had a "golden path".

PR #2232, opened 2022-06-14
Replace `select_next_some` with `next` to avoid `panics`

da-kami:
`select_next_some` will panic if we read from a stream that is empty already unless used in a `FusedFuture` aware primitive.
See: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.select_next_some

We saw panics on the maker side after a taker had already closed the substream.

This can be avoided by using `next`.
Note that we could remove the `.fuse()` when creating the stream, but for more resilience we keep it.
The `.fuse()` prevents all consecutive reads from the stream to return `Ready(None)` instead of panicking.
See: https://docs.rs/futures/latest/futures/stream/trait.Stream.html#panics

---

Note: the file `setup_contract_deprecated.rs` was not adapted to incorporate these changes. We could opt for applying these changes there as well, but given that it is only used by the old connection we don't have to.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -22,9 +22,9 @@ use bdk::bitcoin::Transaction;
 use bdk::descriptor::Descriptor;
 use bdk::miniscript::DescriptorTrait;
 use bdk_ext::keypair;
-use futures::stream::FusedStream;
 use futures::Sink;
 use futures::SinkExt;
+use futures::Stream;
 use futures::StreamExt;
 use maia::commit_descriptor;
 use maia::compute_adaptor_pk;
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
looks good! 
Great job with detailed commit message too 🥇

luckysori (approved):
Nice find and thanks for educating us with the commit message and PR description.

I would opt to remove the `.fuse()` from the call site because we are not taking advantage of the `FusedStream` and it makes the code a tiny bit more confusing. I know I will end up removing it if I notice that it is unused :stuck_out_tongue:

da-kami:
> Nice find and thanks for educating us with the commit message and PR description.
> 
> I would opt to remove the `.fuse()` from the call site because we are not taking advantage of the `FusedStream` and it makes the code a tiny bit more confusing. I know I will end up removing it if I notice that it is unused 😛

I would opt on keeping it for now for resilience. Unless I missed something calling `fuse` will prevent the stream from panicking if called after it returned `Ready(None)`. We are not using it in the sense that we return an error in all cases where we receive `Ready(None)`, but I still think it is more resilient to call `fuse`. I am actually surprised that the implementation of `Stream` would panic in this way.

PR #2239, opened 2022-06-14
Bump libp2p stack to 0.33

klochowicz:
Endpoint API had to be adjusted, as Transport is not cloneable anymore. Instead,
we box a closure that can create and upgrade a transport when needed.


bonomat:
supersedes #2146  https://github.com/itchysats/itchysats/pull/2143 https://github.com/itchysats/itchysats/pull/2144

bonomat (approved):
LGTM.

Side note: that's another reason why I dislike libp2p, continuously getting minor, but breaking, updates.

PR #2246, opened 2022-06-16
Improve rollover tests

da-kami:
This should help get more assurance on rollovers from the tests.

Note that the double rollover is not significantly more expensive - the rollover itself is fast, the setup is what takes most time. I still opted for keeping the single rollover tests in parallel. We could drop those but then it is a bit harder to know which one failed. We would have to look at the logs then.

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -414,20 +414,42 @@ async fn rollover_an_open_cfd(maker_position: Position) {
         .set_offer_params(dummy_offer_params(maker_position))
         .await;
 
-    taker.trigger_rollover(order_id).await;
+    let mut maker_fees_before_rollover = maker.first_cfd().accumulated_fees;
+    let mut taker_fees_before_rollover = taker.first_cfd().accumulated_fees;
 
-    wait_next_state!(
-        order_id,
-        maker,
-        taker,
-        CfdState::IncomingRolloverProposal,
-        CfdState::OutgoingRolloverProposal
-    );
+    for _ in 0..nr_rollovers {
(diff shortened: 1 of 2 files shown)


daemon-tests/tests/happy_path.rs
```diff
@@ -395,16 +395,32 @@ async fn force_close_open_cfd(maker_position: Position) {
...
+    rollover_an_open_cfd(Position::Short, 2).await;
+}
+
+#[tokio::test]
+async fn double_rollover_an_open_cfd_maker_going_long() {
+    // double rollover ensures that both parties properly succeeded can do another rollover
+
+    let _guard = init_tracing();
+    rollover_an_open_cfd(Position::Long, 2).await;
+}
```
luckysori: :question: Why do we need to test this both for long and short? I think we introduced certain taker-goes-short tests to mark the milestone at the time, but I'm not sure there is any major difference at the actor level between long and short. I suspect we don't need to encumber ourselves with duplicated tests in this layer.

luckysori (approved):
I'm happy to test double rollover at this level, but I'm not sure we need two very similar instances of the same test.

daemon-tests/tests/happy_path.rs
da-kami: I added them for having more complete tests. The advantage is marginal - happy to remove one of them.

da-kami: I removed one of the tests, adapted test name and comment.

klochowicz (approved):
thanks, it is certainly better than before :) I suspect we'll need to soon assert on fees as well (esp. with retries), but thas can be done on top!

PR #2247, opened 2022-06-16
Remove react-refreshed as it's unused

bonomat:
supersede https://github.com/itchysats/itchysats/pull/2241


da-kami (approved):
😅 now I wonder what we used it for... not important though :)

luckysori (approved):

PR #2248, opened 2022-06-16
Move rollover into respective rollover protocol actors

da-kami:
This allows to ensure the message order, i.e. taker always sends message first then waits to receive, maker always waits to receive first then sends.
This also makes sense in the context of the maker starting to wait to receive the first message after accepting and if the taker never comes back nothing no message is ever sent.

I recommend reviewing by commit but please keep the following in mind:
I am planning to `drop` the commits that change the code in `setup_contract.rs` - we could actually keep it, but given that this code will be removed I would drop the commits and keep the changeset to the resulted code in the `rollover` module.

The commits are structured to ensure that the code is the same and works. I always refactored a part and then ran the tests to ensure that it still works as expected.
I first extracted all code to be called from both parties into function in `setup_contract.rs`, then moved the code into the `rollover` module and the respective `maker` and `taker` actor.
For the `maker` I flipped the order to first wait to receive messages and then send.

### DIFF
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -1,13 +1,19 @@
+use crate::command;
+use crate::future_ext::FutureExt;
+use crate::oracle;
+use crate::rollover::protocol::*;
+use crate::shared_protocol::format_expect_msg_within;
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
 use asynchronous_codec::Framed;
 use asynchronous_codec::JsonCodec;
-use futures::future;
+use bdk_ext::keypair;
 use futures::SinkExt;
 use futures::StreamExt;
 use libp2p_core::PeerId;
 use maia_core::secp256k1_zkp::schnorrsig;
+use model::Dlc;
 use model::FundingRate;
 use model::OrderId;
 use model::Position;
@@ -20,11 +26,6 @@ use xtra::message_channel::MessageChannel;
 use xtra_libp2p::NewInboundSubstream;
 use xtra_libp2p::Substream;
 use xtra_productivity::xtra_productivity;
-
-use crate::command;
-use crate::oracle;
-use crate::rollover::protocol::*;
-use crate::setup_contract;
 
 use super::protocol;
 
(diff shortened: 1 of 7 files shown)


daemon/src/setup_contract.rs
```diff
@@ -496,75 +497,13 @@ async fn build_own_cfd_transactions(
...
-        .with_context(|| format_expect_msg_within("Msg1", ROLLOVER_MSG_TIMEOUT))?
-        .context("Empty stream instead of Msg1")?
-        .try_into_msg1()?;
-
-    let lock_amount = dlc.taker_lock_amount + dlc.maker_lock_amount;
-
-    let ((maker_identity, maker_punish_params), (taker_identity, taker_punish_params)) =
-        (punish_maker, punish_taker);
-    let commit_desc = commit_descriptor(
+fn build_commit_descriptor(
```
bonomat: This function seems to be unnecessary as all you do is passing the parameters down. Happy to leave it here as it improves readability though.

daemon/src/rollover/protocol.rs
```diff
@@ -141,3 +190,402 @@ pub(crate) async fn emit_failed(order_id: OrderId, e: anyhow::Error, executor: &
         tracing::error!(%order_id, "Failed to execute rollover failed: {e:#}")
     }
 }
+
+pub(crate) fn build_punish_params(
+    our_role: Role,
+    sk: SecretKey,
+    counterparty_pk: PublicKey,
+    other_msg0: RolloverMsg0,
```
bonomat: can we unify the naming: e.g. 
```suggestion
    counterparty_msg0: RolloverMsg0,
```

or rename `counterparty_pk` to `other_pk`.

daemon/src/rollover/protocol.rs
```diff
@@ -141,3 +190,402 @@ pub(crate) async fn emit_failed(order_id: OrderId, e: anyhow::Error, executor: &
...
+    rev_pk: bdk::bitcoin::PublicKey,
+    publish_pk: bdk::bitcoin::PublicKey,
+) -> (
+    (PublicKey, PunishParams),
```
bonomat: It might be worth introducing different types at some point, 
e.g. 

```suggestion
    MakerPunishParams,
    TakerPunishParams, ...
```

bonomat (commented):
I've only some comments for now. I think it makes sense what you do but I could not tell yet if some vars have been swapped by accident. I guess not, as otherwise the tests wouldn't run 😅 

Anyway, I'd like to have another look at it tomorrow morning if you don't mind.

daemon/src/setup_contract.rs
da-kami: Readability is the only reason why it exists - initially I had it inlined but found the unnecessarily bloated :)

daemon/src/setup_contract.rs
```diff
@@ -488,6 +493,61 @@ pub async fn roll_over(
...
+    let (punish_maker, punish_taker, other_punish_params) = create_punish_params(
+        our_role,
+        dlc.identity,
+        dlc.identity_counterparty,
+        msg0,
+        rev_pk,
+        publish_pk,
+    );
+
+    let own_cfd_txs = create_own_cfd_transactions(
```
bonomat: Where is `create_own_cfd_transactions` coming from?

daemon/src/setup_contract.rs
```diff
@@ -488,6 +493,61 @@ pub async fn roll_over(
...
+
+    let msg0 = stream
+        .next()
+        .timeout(ROLLOVER_MSG_TIMEOUT)
+        .await
+        .with_context(|| format_expect_msg_within("Msg0", ROLLOVER_MSG_TIMEOUT))?
+        .context("Empty stream instead of Msg0")?
+        .try_into_msg0()?;
+
+    let (punish_maker, punish_taker, other_punish_params) = create_punish_params(
```
bonomat: Where does `create_punish_params` come from?

daemon/src/rollover/protocol.rs
```diff
@@ -141,3 +190,402 @@ pub(crate) async fn emit_failed(order_id: OrderId, e: anyhow::Error, executor: &
...
+    )
+    .context("Commit adaptor signature does not verify")?;
+
+    let other_address = match our_role {
+        Role::Maker => dlc.taker_address.clone(),
+        Role::Taker => dlc.maker_address.clone(),
+    };
+
+    for own_grouped_cets in own_cets.clone() {
+        let other_cets = msg1
```
bonomat: Is this clone needed?

Sorry wrong line, I mean the clone a line up.

daemon/src/rollover/protocol.rs
```diff
@@ -141,3 +190,402 @@ pub(crate) async fn emit_failed(order_id: OrderId, e: anyhow::Error, executor: &
...
+        let derived_rev_pk = bdk::bitcoin::PublicKey::new(
+            secp256k1_zkp::PublicKey::from_secret_key(SECP256K1, &revocation_sk_theirs),
+        );
+
+        if derived_rev_pk != bdk::bitcoin::PublicKey::from(dlc.revocation_pk_counterparty) {
+            anyhow::bail!("Counterparty sent invalid revocation sk");
+        }
+    }
+
+    // TODO: Are the clones in the two lines below expensive? (those are extra now)
```
bonomat: I think they are ok.

daemon/src/rollover/maker.rs
```diff
@@ -203,30 +212,152 @@ impl Actor {
...
+                        .next()
+                        .timeout(ROLLOVER_MSG_TIMEOUT)
+                        .await
+                        .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
+                        .context("Empty stream instead of Msg3")?
+                        .context("Unable to decode dialer Msg3")?
+                        .into_rollover_msg()?
+                        .try_into_msg3()?;
+
+                    // TODO: If the maker fails so send Msg3 we should probably still treat the
```
bonomat: I agree, this msg is unnecessary as discussed.

bonomat (approved):
I think the first three commits are a bit of a mess and the first one particular does not build as it makes use of functions which do not exist. 

Ideally we would keep the extracts function from commit 1-3 and reuse them in the rollover actor (or vice versa). This will prevent us from having to maintain the same protocol in two places until we drop the old network stack.

da-kami:
> I think the first three commits are a bit of a mess and the first one particular does not build as it makes use of functions which do not exist.

Sorry if the commits are not clean, I might squash them now that the review is done :)

> Ideally we would keep the extracts function from commit 1-3 and reuse them in the rollover actor (or vice versa). This will prevent us from having to maintain the same protocol in two places until we drop the old network stack.

I am happy to achieve this. I would then opt for keeping them in `rollover::protocol` (because the new code should own them) and use them in the old code. Is that what you had in mind as well?

bonomat:
> > 
> 
> I am happy to achieve this. I would then opt for keeping them in `rollover::protocol` (because the new code should own them) and use them in the old code. Is that what you had in mind as well?

yes, I think this is a good idea.

klochowicz:
I tried dropping the "drop" commits and rebasing and the task was a bit too much for me without deep knowledge of the way you wanted to change it, I have since abandoned the task. Any chance we could split this into smaller/more managable PRs? like just the pure refactor first, and then all the changes to the protocol itself? If it's too late for this and splitting would cause more pain then don't worry about it, but in hindsight I suspect it would have been much easier....

daemon/src/rollover/protocol.rs
da-kami: I opted for `counterparty` instead of `other` because it is more explicit and furhermore we have used it in other places in the code (e.g. the `Dlc`).

See https://github.com/itchysats/itchysats/pull/2248/commits/f7342e12be9941f33eaf735eb8af325d7cf5eaa3

daemon/src/rollover/protocol.rs
da-kami: Happy to see this in a follow-up

daemon/src/rollover/protocol.rs
da-kami: I tried to make this more explicit, see https://github.com/itchysats/itchysats/pull/2248/commits/690e63b02f4b5b3f3cd2a3088f196af6391f8a2f

daemon/src/rollover/protocol.rs
```diff
@@ -141,3 +301,302 @@ pub(crate) async fn emit_failed(order_id: OrderId, e: anyhow::Error, executor: &
...
+
+    Ok(own_cfd_txs)
+}
+
+pub(crate) fn build_commit_descriptor(
+    ((maker_identity, maker_punish_params), (taker_identity, taker_punish_params)): (
+        (PublicKey, PunishParams),
+        (PublicKey, PunishParams),
+    ),
+) -> Descriptor<PublicKey> {
```
bonomat: This doesn't have to be a tuple of tuples. You can get rid of at least one level

bonomat (approved):
It's really hard to see if any variables have been swapped by accident. I guess we have to rely on the tests and wait to see if things still work as expected after the merge

daemon/src/rollover/protocol.rs
da-kami: In an effort to have more assurance that we did not accidentally flip something I added https://github.com/itchysats/itchysats/pull/2248/commits/b644793afccb61552e2061f9121669af9381d313

Note: The `double_rollover_test` added recently is a good safety net - during the refactoring I did run into a flip and the test failed on the second run with the `Counterparty sent invalid revocation sk` ;)

PR #2252, opened 2022-06-16
Allow waiting before restarting a supervised actor

klochowicz:
In order to allow waiting, first I refactored the restart policy to allow passing in an async closure. Second commit adds the new capability, and adds a test to cover it.
Last but not least, dialer and listener use the new capability.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -34,6 +34,7 @@ use xtra_libp2p::Endpoint;
 use xtra_libp2p_ping::ping;
 use xtra_libp2p_ping::pong;
 use xtras::supervisor;
+use xtras::supervisor::always_restart;
 
 pub use bdk;
 pub use maia;
@@ -241,10 +242,8 @@ where
             let endpoint_addr = endpoint_addr.clone();
             move || dialer::Actor::new(endpoint_addr.clone(), maker_multiaddr.clone())
         };
-        let (dialer_supervisor, dialer_actor) = supervisor::Actor::with_policy(
-            dialer_constructor,
-            |_: &dialer::Error| true, // always restart dialer actor
-        );
+        let (dialer_supervisor, dialer_actor) =
+            supervisor::Actor::with_policy(dialer_constructor, always_restart());
 
         let (offers_supervisor, libp2p_offer_addr) = supervisor::Actor::new({
             let cfd_actor_addr = cfd_actor_addr.clone();
@@ -286,10 +285,8 @@ where
         let dialer_supervisor = dialer_supervisor.create(None).spawn(&mut tasks);
         let offers_supervisor = offers_supervisor.create(None).spawn(&mut tasks);
 
-        let (supervisor, price_feed_actor) = supervisor::Actor::with_policy(
-            price_feed_constructor,
-            |_| true, // always restart price feed actor
-        );
+        let (supervisor, price_feed_actor) =
+            supervisor::Actor::with_policy(price_feed_constructor, always_restart());
 
         let price_feed_supervisor = supervisor.create(None).spawn(&mut tasks);
 
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -35,6 +35,7 @@ use xtra_libp2p_ping::ping;
 use xtra_libp2p_ping::pong;
 use xtras::supervisor;
 use xtras::supervisor::always_restart;
+use xtras::supervisor::always_restart_after;
 
 pub use bdk;
 pub use maia;
@@ -69,6 +70,10 @@ pub mod wire;
 /// Duration between the heartbeats sent by the maker, used by the taker to
 /// determine whether the maker is online.
 pub const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
+
+/// Duration between the restart attempts after a supervised actor has quit with
+/// a failure.
+pub const RESTART_INTERVAL: Duration = Duration::from_secs(5);
 
 pub const ENDPOINT_CONNECTION_TIMEOUT: Duration = Duration::from_secs(20);
 pub const PING_INTERVAL: Duration = Duration::from_secs(5);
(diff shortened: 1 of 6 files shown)


xtras/src/supervisor.rs
```diff
@@ -17,11 +19,25 @@ pub struct Actor<T, R> {
...
+        + Sync,
+>;
+
+/// Closure that configures the supervisor to restart on every kind of error
+pub fn always_restart<E>() -> AsyncClosure<E>
+where
+    E: Error + Send + Sync + 'static,
+{
+    Box::new(|_: &E| Box::pin(async move { true }))
+}
```
luckysori, line 40: :question: Could we make this private and use `new` where necessary? I think there was a problem with that approach, but I can't figure it out when comparing the signatures of `new` and `with_policy`.

xtras/src/supervisor.rs
```diff
@@ -249,8 +267,11 @@ struct GetMetrics;
 
 #[cfg(test)]
 mod tests {
+    use crate::SendAsyncSafe;
+
```
luckysori: :wrench: Unwanted newline.

xtras/src/supervisor.rs
```diff
@@ -280,6 +301,48 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn supervisor_can_delay_respawn() {
```
luckysori, line 304: :+1: Very clear test! Thanks.

The second assertion makes me a bit nervous because it relies on relatively tricky timing, but I see no alternative!

daemon/src/lib.rs
```diff
@@ -70,6 +71,10 @@ pub mod wire;
 /// determine whether the maker is online.
 pub const HEARTBEAT_INTERVAL: Duration = Duration::from_secs(5);
 
+/// Duration between the restart attempts after a supervised actor has quit with
+/// a failure.
+pub const RESTART_INTERVAL: Duration = Duration::from_secs(5);
```
luckysori, line 76: :question: Does this apply to all supervised actors in this file? The name of this constant and the commit message are not completely aligned imo.

luckysori (approved):
Good work!

Only thing missing for me is the motivation for the last patch. I don't know why we need to delay the restart for those particular supervised actors. Could you explain it in the body of the commit description?

da-kami (commented):
Why is this feature needed?
Before we go forward this I would like to understand the motivation for this - what problem does it solve?

When I had another look at the dialer I came to the conclusion that we don't need a feature to wait in the supervisor. The `CONNECTION_TIMEOUT` prevents that we restart immediately.

There is only one code path that could lead to restarting in an endless loop:

https://github.com/itchysats/itchysats/blob/b77ea81958aff8e10e43223aff0f94c3ca83c79c/xtra-libp2p/src/dialer.rs#L99-L102

If the endpoint has a connection, but the dialer for whatever reason tries to restart this might loop forever. But I don't think that waiting for restart in the supervisor would solve this problem.
We might want to re-evaluate if this `ensure` is actually needed. I think in case the dialer restarts and notices that we are in fact connected it should just keep running (without connecting again).

klochowicz:
> Why is this feature needed? Before we go forward this I would like to understand the motivation for this - what problem does it solve?
> 
> When I had another look at the dialer I came to the conclusion that we don't need a feature to wait in the supervisor. The `CONNECTION_TIMEOUT` prevents that we restart immediately.
> 
> There is only one code path that could lead to restarting in an endless loop:
> 
> https://github.com/itchysats/itchysats/blob/b77ea81958aff8e10e43223aff0f94c3ca83c79c/xtra-libp2p/src/dialer.rs#L99-L102
> 
> If the endpoint has a connection, but the dialer for whatever reason tries to restart this might loop forever. But I don't think that waiting for restart in the supervisor would solve this problem. We might want to re-evaluate if this `ensure` is actually needed. I think in case the dialer restarts and notices that we are in fact connected it should just keep running (without connecting again).

It is needed to avoid hacks with sleep before stopping the actor. We even briefly talked about this in your PR:

I find it much more clean to do specify this on the supervisor level than as a hack somewhere inside the actor - see where I suggested such a solution: https://github.com/itchysats/itchysats/pull/2220#discussion_r896827764

 I'd be even open to have some form of this on by default, but we can talk about it in a github discussion / ADR.

xtras/src/supervisor.rs
klochowicz: ah, that rust-analyzer again! I think it tries to establish the convention (not used here) that crate elements should come in its own block at the top. fixed :)

xtras/src/supervisor.rs
klochowicz: I prefer explicit assignment of policies where the error is defined - at the end of the day, it's easy to swap one for another (end the `new()` makes a bigger diff then). I would leave this public for maximum flexibility for downstream crates.

xtras/src/supervisor.rs
klochowicz: thank you! 
I agree -it's not too pretty, but does the job in the test. I've tried a few alternatives, but could not find anything easier :/

daemon/src/lib.rs
klochowicz: it applies to all actors that use the constant; any suggestions to improve it?

klochowicz:
> Why is this feature needed? Before we go forward this I would like to understand the motivation for this - what problem does it solve?
> 
> When I had another look at the dialer I came to the conclusion that we don't need a feature to wait in the supervisor. The `CONNECTION_TIMEOUT` prevents that we restart immediately.
> 
> There is only one code path that could lead to restarting in an endless loop:
> 
> https://github.com/itchysats/itchysats/blob/b77ea81958aff8e10e43223aff0f94c3ca83c79c/xtra-libp2p/src/dialer.rs#L99-L102
> 
> If the endpoint has a connection, but the dialer for whatever reason tries to restart this might loop forever. But I don't think that waiting for restart in the supervisor would solve this problem. We might want to re-evaluate if this `ensure` is actually needed. I think in case the dialer restarts and notices that we are in fact connected it should just keep running (without connecting again).

As for dialer: It should never keep running without connecting again, this is the whole point of that entity. If you'd like to rework that, please feel free to submit a PR to change that.

As such, I think the new capability is tremendously useful for a supervisor (allow async code in restart policy). If we decide to reimplement dialer to something different, this feature will still be useful for other things.

da-kami (approved):
I still think that this functionality is not super important but I don't have a problem with merging it.

PR #2253, opened 2022-06-17
WIP - start replacing `asynchronous_codec` with `tokio-util::codec`

da-kami:
Including own JSON codec impl (mostly copied from old crate found online; might need adaption).


klochowicz:
closing the PR, as it's unlikely we'll need to go down this path

PR #2254, opened 2022-06-17
Move sql types into sql module

### LINKED ISSUE #2139: Move sql specific types out of model into sqlite-db module

bonomat:
This PR should only move things back and forth between different modules. 
The goal of this exercise was to have a completely independent sqlite-module which only makes use of the module crate and no-one uses types from the sqlite-module.

That way, we can avoid vendor lock-in and move to a different db easily.

We are not done yet. But I'm afraid I will die in merge conflicts with this already so I'm pushing this PR. 

I also don't really know how to proceed with the remaining types 😅

- sqlite_db::Cfd
- sqlite_db::CfdAggregate
- sqlite_db::ClosedCfdAggregate
- sqlite_db::FailedCfdAggregate

resolves #2139


sqlite-db/src/test_events/rollover_completed.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "RolloverCompleted",
   "data": {
-    "funding_fee": { "fee": 100, "rate": 1.5 },
```
klochowicz, line 4: ❓
Hm, have we changed how we store the funding rate? (are we backwards compatible - ie. won't parsing old events result in wrong values being used?

sqlite-db/src/rollover/load.rs
```diff
@@ -70,55 +71,54 @@ pub async fn load(
...
-
     cfd_row_id: i64,
 ) -> Result<Vec<RevokedCommit>> {
     let revoked_commit = sqlx::query!(
         r#"
             SELECT
-                encsig_ours as "encsig_ours: model::AdaptorSignature",
-                publication_pk_theirs as "publication_pk_theirs: model::PublicKey",
-                revocation_sk_theirs as "revocation_sk_theirs: model::SecretKey",
+                encsig_ours as "encsig_ours: models::AdaptorSignature",
```
klochowicz, line 117: I see what you did here - creating a new module named  `models` might reduce the confusion a bit with our "model" crate, but perhaps we could contemplate renaming the "model" crate to something better?

(not for this PR though, and I think "models" is just a good choice :-) )

model/src/cfd.rs
```diff
@@ -60,14 +61,12 @@ use std::ops::RangeInclusive;
 use std::str;
 use time::Duration;
 use time::OffsetDateTime;
-use uuid::adapter::Hyphenated;
 use uuid::Uuid;
 
 pub const CET_TIMELOCK: u32 = 12;
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, sqlx::Type)]
-#[sqlx(transparent)]
```
klochowicz, line 69: won't this make it harder to debug things manually in the DB, if we only add hyphens in `Display` trait?

sqlite-db/src/closed.rs
```diff
@@ -88,6 +88,7 @@ pub struct Lock {
 /// fields so that only sane combinations of transactions can be
 /// loaded from the database.
 #[derive(Debug, Clone, Copy, PartialEq)]
+// TODO: this type should not live here
```
da-kami: ❓ Can we action this Todo?

sqlite-db/src/models.rs
```diff
@@ -338,3 +338,39 @@ impl From<Position> for model::Position {
         }
     }
 }
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash)]
+pub struct Identity(x25519_dalek::PublicKey);
+
+impl fmt::Display for Identity {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let hex = hex::encode(self.0.as_bytes());
```
da-kami, line 356: ⚠️ Isn't that a breaking change because we would now write something hex encoded into the db that was previously not hex encoded?

sqlite-db/src/test_events/rollover_completed.json
```diff
@@ -1,7 +1,7 @@
...
+    "funding_fee": { "fee": 100, "rate": 0.015 },
```
da-kami, line 4: ❓How come the tests data changes but no test changed that uses this data? I think the test changes should be in the same commit... or were there none? (then this data is probably not used and we can just set the value to 0 to indicate it is not important?)

da-kami (approved):
Thanks for all the cleanup!
Only one potential dangerous change, the one where we now encode/decode hex - I don't think we previously did this, so it might be a breaking change. Please evaluate before merging! (If I missed something and it's not breaking feel free to merge)

klochowicz (approved):
thank you for cleaning this up, impressive size of the diff itself 👍

sqlite-db/src/closed.rs
bonomat: I've actioned it in a later commit.

sqlite-db/src/models.rs
bonomat: It shouldn't be a breaking change but let me double check this.

sqlite-db/src/test_events/rollover_completed.json
bonomat: This was actually a "bug" in the old tests but we have never seen it because we did not need it. Now, I parse the rate using `new` which will fail if the rate is above 1.

sqlite-db/src/test_events/rollover_completed.json
bonomat: See other comment for Daniel: this was a "bug" but only in the test data. I checked this with the production db and we always store it in the format 0-1.

sqlite-db/src/models.rs
bonomat: It's not a breaking change. I just copied the existing display implementation from here: 
https://github.com/itchysats/itchysats/blob/8bb7d6dc524213f16a99617d16571f249ac4673d/model/src/lib.rs#L507-L513

Tested with mainnet DB locally as well and works.

da-kami (approved):

bonomat:
bors retry

PR #2270, opened 2022-06-20
Retire heartbeats

### LINKED ISSUE #2036: Connection actor can panic on broken clock invariant
We can panic  when clock drifts, e.g. when macOS system goes to sleep and awakes itself, syncing with NTP server after coming back online and going "backwards in time".

log:
```
thread 'rocket-worker-thread' panicked at 'clock is monotonic: SystemTimeError(1.454591091s)', daemon/src/connection.rs:102:22
thread 'rocket-worker-thread' panicked at 'watch channel should outlive the future: RecvError(())', /home/runner/work/itchysats/itchysats/daemon/src/connection.rs:625:14
```

We should either not panic or supervise the connection actor.

### LINKED ISSUE #2234: use libp2p ping to derive `connectedToMaker` state - retire heartbeats

klochowicz:
Details in the commits.

Mostly removed unused code and added a simple actor that adapts connection information onto the watch channel, which was used before.

Fixes #2234

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -3,7 +3,6 @@ use daemon::bdk::bitcoin::Amount;
 use daemon::bdk::bitcoin::SignedAmount;
 use daemon::bdk::bitcoin::Txid;
 use daemon::connection::ConnectionStatus;
-use daemon::connection::MAX_RECONNECT_INTERVAL_SECONDS;
 use daemon::projection::CfdOrder;
 use daemon::projection::CfdState;
 use daemon::projection::MakerOffers;
@@ -852,16 +851,14 @@ async fn open_cfd_is_refunded() {
 
 #[tokio::test]
 async fn taker_notices_lack_of_maker() {
-    let short_interval = Duration::from_secs(1);
-
     let _guard = init_tracing();
 
     let maker_config = MakerConfig::default()
-        .with_heartbeat_interval(short_interval)
-        .with_dedicated_port(35123); // set a fixed port so the taker can reconnect
+        .with_dedicated_port(35123)
+        .with_dedicated_libp2p_port(35124); // set fixed ports so the taker can reconnect
     let maker = Maker::start(&maker_config).await;
 
-    let taker_config = TakerConfig::default().with_heartbeat_interval(short_interval);
+    let taker_config = TakerConfig::default();
     let mut taker = Taker::start(
         &taker_config,
         maker.listen_addr,
@@ -870,6 +867,8 @@ async fn taker_notices_lack_of_maker() {
     )
     .await;
 
+    sleep(Duration::from_secs(5)).await; // wait a bit until taker notices change
+
     assert_eq!(
         ConnectionStatus::Online,
         next(taker.maker_status_feed()).await.unwrap()
@@ -877,7 +876,7 @@ async fn taker_notices_lack_of_maker() {
 
     drop(maker);
 
-    sleep(taker_config.heartbeat_interval).await;
+    sleep(Duration::from_secs(5)).await; // wait a bit until taker notices change
 
     assert_eq!(
         ConnectionStatus::Offline { reason: None },
@@ -886,8 +885,7 @@ async fn taker_notices_lack_of_maker() {
 
     let _maker = Maker::start(&maker_config).await;
 
-    sleep(Duration::from_secs(MAX_RECONNECT_INTERVAL_SECONDS) + taker_config.heartbeat_interval)
-        .await;
+    sleep(Duration::from_secs(5)).await; // wait a bit until taker notices change
 
     assert_eq!(
         ConnectionStatus::Online,
--- a/daemon/src/connection.rs
+++ b/daemon/src/connection.rs
@@ -22,8 +22,6 @@ use rand::thread_rng;
 use rand::Rng;
 use std::net::SocketAddr;
 use std::time::Duration;
-use std::time::SystemTime;
-use time::OffsetDateTime;
 use tokio::net::TcpStream;
 use tokio::sync::watch;
 use tokio_tasks::Tasks;
@@ -32,7 +30,6 @@ use xtra::KeepRunning;
 use xtra_productivity::xtra_productivity;
 use xtras::address_map::NotConnected;
 use xtras::AddressMap;
-use xtras::SendInterval;
 
 /// Time between reconnection attempts
 pub const MAX_RECONNECT_INTERVAL_SECONDS: u64 = 60;
@@ -43,10 +40,6 @@ const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 #[allow(clippy::large_enum_variant)]
 enum State {
     Connected {
-        last_heartbeat: SystemTime,
-        /// Last pulse measurement time. Used for checking whether measuring
-        /// task is not lagging too much.
-        last_pulse: SystemTime,
         write: wire::Write<wire::MakerToTaker, wire::TakerToMaker>,
         _tasks: Tasks,
     },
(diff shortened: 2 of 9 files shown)


daemon-tests/tests/happy_path.rs
```diff
@@ -616,14 +616,16 @@ async fn taker_notices_lack_of_maker() {
     )
     .await;
 
+    sleep(Duration::from_secs(5)).await;
```
bonomat: What's this sleep needed for?

daemon-tests/tests/happy_path.rs
```diff
@@ -616,14 +616,16 @@ async fn taker_notices_lack_of_maker() {
...
+
     assert_eq!(
         ConnectionStatus::Online,
         next(taker.maker_status_feed()).await.unwrap()
     );
 
     drop(maker);
 
-    sleep(taker_config.heartbeat_interval).await;
+    sleep(Duration::from_secs(5)).await;
```
bonomat: Same here, I think a short comment could help understand why we need to sleep exactly 5 secs.

daemon/src/connection.rs
```diff
@@ -78,83 +71,11 @@ impl State {
...
-            State::Connected { last_heartbeat, .. } => SystemTime::now()
-                .duration_since(*last_heartbeat)
-                .expect("clock is monotonic"),
-            State::Disconnected => return false,
-        };
-
-        if duration_since_last_heartbeat < timeout {
-            return false;
-        }
-
```
bonomat, line 122: I'm almost sad to see this go 😢

Nah.. actually I'm not 🥳

daemon/src/online_status_actor.rs
```diff
@@ -0,0 +1,105 @@
...
+            msg.peer
+        );
+        if msg.peer == self.watched_peer {
+            self.sender
+                .send(ConnectionStatus::Online)
+                .expect("Receiver to outlive this actor");
+        }
+    }
+
+    async fn handle_connection_dropped(&mut self, msg: endpoint::ConnectionDropped) {
```
bonomat: If the connection was dropped, is there a reason why this actor should exist? Might be better to shut it down?

maker/src/connection.rs
```diff
@@ -481,7 +490,18 @@ impl Actor {
                 }
             },
         );
-        tasks.add(this.send_interval(self.heartbeat_interval, move || SendHeartbeat(identity)));
+
+        let daemon_semver = parse_vergen_version(&daemon_version);
+        let no_need_for_heartbeats =
+            semver::VersionReq::parse(">= 0.4.20").expect("to parse VersionReq");
```
bonomat, line 496: I think this should be `0.4.21`, i.e. only from the next release.

bonomat (commented):
Added a few comments for now.

daemon-tests/tests/happy_path.rs
klochowicz: it's a magic value, because detecting loss and subsequent reconnect takes a few seconds. 5s is a safe bet (note that this value is used in many actor tests).

added a short comment in all places.

daemon/src/online_status_actor.rs
klochowicz: This actor does not perform any connection management; it only translates the subscribed notifications into a watch channel; as such, it needs to be present all the time to:
a) not miss the subsequent notifications when the connection comes back online
b) own the sender (and thus keep the channel open)

Perhaps it's the naming of the actor that's confusing?

maker/src/connection.rs
klochowicz: indeed, I will change this to `0.4.21` and downgrade the tracing warning on the taker side to trace (as this change will invariably still send heartbeats to the taker right until the next release time).

daemon-tests/tests/happy_path.rs
klochowicz: it's needed to detect the loss. We were able to avoid it before due to a "special" config, where we had a really short heartbeat interval for the tests. Now, we don't use heartbeats, so that mechanism doesn't work, and I didn't want to add a special code for tests into production.

daemon/src/connection.rs
klochowicz: note that this fixes the bug with the panic in production ;D

daemon/src/online_status_actor.rs
bonomat: I meant: 
There will be one actor per connected peer, even if the peer only came online once and never ever again. Would it not be better to just drop the actor as well if the peer disappears and just recreate it when it appears again.

I don't think it's a big issue as the memory footprint of this actor is quite small.

daemon-tests/tests/happy_path.rs
```diff
@@ -628,16 +627,14 @@ async fn open_cfd_is_refunded() {
...
 #[tokio::test]
 async fn taker_notices_lack_of_maker() {
-    let short_interval = Duration::from_secs(1);
-
     let _guard = init_tracing();
 
     let maker_config = MakerConfig::default()
-        .with_heartbeat_interval(short_interval)
-        .with_dedicated_port(35123); // set a fixed port so the taker can reconnect
+        .with_dedicated_port(35123)
```
bonomat, line 857: Not needed for this PR: 

I think it would be nice to use random ports here to avoid port-in-use conflicts.

bonomat (approved):

daemon/src/online_status_actor.rs
klochowicz: ah, I see what you're saying! this actor actually is taker-only, so there only over will be one of them; the peer id is configured at startup. (for some reason, taker-specific code still lives in the `daemon` crate, we should clean this up :/)

bonomat: ah, got it. Thanks for the clarification :)

klochowicz:
I'll wait with merging this until after rollover release is in.

klochowicz:
@da-kami @luckysori are you ok with me merging this? it has been for a while!

daemon-tests/tests/happy_path.rs
da-kami: I agree, would be good to use random ports so we avoid flaky tests!

daemon-tests/tests/happy_path.rs
```diff
@@ -646,14 +643,16 @@ async fn taker_notices_lack_of_maker() {
     )
     .await;
 
+    sleep(Duration::from_secs(5)).await; // wait a bit until taker notices change
```
da-kami, line 870: I think it would be better to do a `retry_until` logic where we only sleep a short period of time and re-check frequently if the value already changed.

Something similar to this: https://github.com/itchysats/itchysats/blob/f270f0353d84f164a3701952f4118886fa5d2fec/xtra-libp2p-ping/src/lib.rs#L116-L127

That way we don't waste test time - this test was already super long running.

daemon/src/connection.rs
```diff
@@ -310,15 +220,6 @@ impl Actor {
             Some(wire::MakerToTaker::Hello(actual_version)) => {
                 tracing::info!(%maker_identity, %actual_version, "Received Hello message from maker");
                 if proposed_version != actual_version {
```
da-kami, line 222: How is removing this code related to the heartbeats?
Why was this removed?

daemon/src/connection.rs
```diff
@@ -491,6 +352,11 @@ pub async fn connect(
     maker_addresses: Vec<SocketAddr>,
 ) {
     loop {
+        maker_online_status_feed_receiver
```
da-kami, line 355: Why did you move this one to the top of the loop - doesn't this change the behavior?
Are we expected to see a status change initially?

daemon/src/online_status_actor.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Actor that transmits updates of ConnectionStatus of a specified PeerId based on
+/// information transmitted by the Endpoint via a watch channel.
+pub struct Actor {
+    endpoint: Address<Endpoint>,
+    watched_peer: PeerId,
+    sender: watch::Sender<ConnectionStatus>,
+}
+
+impl Actor {
```
da-kami: Oh, you actually created an actor for that, interesting!

I would have seen this functionality as part of identify because this is where you learn if maker and taker are compatible. But could keep *compatibility* separate from the pure connection (i.e. am I connected or not) if we want, but this might just make things more complicated than they have to be.

daemon/src/online_status_actor.rs
```diff
@@ -0,0 +1,105 @@
+use std::time::Duration;
```
da-kami: 🛠 I think the file should be called `online_status` - we never use `actor` as part of the module name, the `Actor` is inside.

da-kami (approved):
I don't think any of my comments is blocking, but I do think we have to discuss the concerns of this `online_status::Actor` and the `identify::Actor`. There might be some overlap.

daemon/src/connection.rs
klochowicz: This PR removes heartbeats, which were the mechanism that was in charge of  the `ConnectionStatus`, used here. With the change to new networking, we don't change `ConnectionStatus` from here anymore. This code was a bit hacky, it was reusing the same channel to communicate that the version negotiation failed. Technically, we could try to remodel this behaviour in a different place, but there's little point, as using `identify` protocol will supercede this in a few days and we don't plan to do breaking wire changes on legacy connection anyway.

daemon/src/online_status_actor.rs
klochowicz: I would lean towards separation, as they have clear and distinct responsibilities, but am open to talk about reworking this when `identify` gets merged. 
I figured out creating a simple, purpose-built actor was easier than cramming it inside ping, which was my original idea :)

daemon/src/online_status_actor.rs
klochowicz: indeed! renamed the file.

daemon/src/connection.rs
klochowicz: this change preserves the same behaviour - it was needed as the implementation changed and we indeed see a status change initially as the connection status is managed by lib2p.  (this was caught by tests)

daemon-tests/tests/happy_path.rs
klochowicz: I've created a follow-up ticket from this: #2309  - there's plenty of such calls in that file.

daemon-tests/tests/happy_path.rs
klochowicz: I've extracted #2308

PR #2271, opened 2022-06-20
Show info box if a new version is available on GitHub

### LINKED ISSUE #2233: Poll latest version from github and display in UI
I was thinking of how can we tell a user that a new version is ready to be used: 

1. the maker it is connected to can tell 
2. the daemon could fetch it from GitHub

With our network protocol we implicitely tell the taker to upgrade if their version is not compatible with the maker. 
This works well assuming we do network version bumps. (Which we haven't for a while 😅)

This won't work anymore if we have multiple makers, because then we might have different version out there and a taker might never see a new version with yummy new features.

Instead, we could go for (2) and let the daemon poll github regularly to check if a new version is available.

bonomat:
Resolves #2233

Note: our latest version 0.4.20.1 is not a valid semver version hence, I added a small logic which uses `coerce` to parse the version string. The resulting version of `0.4.20.1` will be `0.4.20`.  

image


taker-frontend/src/App.tsx
```diff
@@ -68,10 +79,47 @@ export const BG_LIGHT = "gray.50";
...
+const fetchGithubVersion = async (setGithubVersion: (version: SemVer | null) => void) => {
+    let response = await fetch("https://api.github.com/repos/itchysats/itchysats/releases/latest", {
+        headers: { accept: "application/json" },
+    });
+
+    let githubVersion = await response.json() as GithubVersion;
+    if (githubVersion?.tag_name) {
+        let version = semver.parse(githubVersion?.tag_name);
+        if (!version) {
+            // needed because 0.20.1 is not a valid semver version
```
klochowicz:
```suggestion
            // needed because 0.4.20.1 is not a valid semver version
```

I believe you would have the same problems with the usual vergen-created version that happens when we're running from master? e.g. `0.4.20-a243bc` (we append git tag to distinguish the versions)
note: the latter is a valid version, but it would fail at comparison, as semver treats such versions as *prereleases* 😕 (edit: the screenshot shows this issue, where it tries to tell the user to downgrade back to `0.4.20`)

taker-frontend/src/App.tsx
```diff
@@ -68,10 +79,47 @@ export const BG_LIGHT = "gray.50";
 export const BG_DARK = "gray.800";
 export const FAQ_URL = "http://faq.itchysats.network";
 
+const fetchGithubVersion = async (setGithubVersion: (version: SemVer | null) => void) => {
```
klochowicz: perhaps we could put `fetchGithubVersion` and `fetchDaemonVersion` into a different file? they clutter the top-level file a bit...

klochowicz (approved):
great addition!

taker-frontend/src/App.tsx
bonomat: If you look at the screenshot, it was able to parse the tag correctly as `0.4.20-217-g4d...`

The screenshot was a hack to get it up, i.e. I turned the condition temporarily around 😬 Otherwise it would not show it.

taker-frontend/src/App.tsx
bonomat: can do

taker-frontend/src/App.tsx
klochowicz: > If you look at the screenshot, it was able to parse the tag correctly as `0.4.20-217-g4d...`
> 
> The screenshot was a hack to get it up, i.e. I turned the condition temporarily around 😬 Otherwise it would not show it.

really? that's good! in the rust `semver` crate the comparison would certainly show it (it treats the tag as prerelease, which should happen before the official release).

Issue #1775, opened 2022-04-06
Design proposal for `/itchysats/rollover` protocol


thomaseizinger:
## Context

Here is what we know about the current rollover protocol:

1. We have an instance of the [two-generals problem](https://en.wikipedia.org/wiki/Two_Generals%27_Problem). Either party should only move to the new state if the other party completed the protocol successfully. We tried to naively solve this with ACKs but literature tells us that this doesn't do anything.
2. During rollover, the parties share revocation keys. Once a revocation key is shared, we have to assume it can be used and thus we MUST NOT go back to a previous commit transaction.
3. The current protocol does not explicitly state, which commit transaction we are rolling over over from. This creates problems in case we run into a 2-general situation where one party completes successfully and they other party does not.

## New protocol

The design proposal for a new protocol is as follows:

```mermaid
sequenceDiagram
	Dialer->>Listener: Open substream `/itchysats/rollover`
	Note right of Listener: Contains CfdId, current commit transaction and `BitMexPriceEventId`
	Listener->>Dialer: Send Accept
	Note right of Listener: Reject on unknown CFD, commit transaction or bad event ID
	Dialer->>Listener: Send PubKeys
	Listener->>Dialer: Send PubKeys
	Dialer->>Listener: Send Signatures
	Listener->>Dialer: Send Signatures
	Dialer->>Listener: Send SecretKey
	Listener->>Dialer: Send SecretKey
	Dialer->>Listener: Send Done
```

```rust
struct Accept {
	funding_rate: f64
}

struct PubKeys {
	revocation: Bytes,
	publish: Bytes,
}

struct Signatures {
    commit: EcdsaAdaptorSignature,
    cets: HashMap<String, Vec<(RangeInclusive<u64>, EcdsaAdaptorSignature)>>,
    refund: Signature,
}

struct SecretKey {
	revocation: Bytes
}
```

The pillars of the new design are:

1. Sending the current commit transaction and event ID allows us to detect state disparity
2. Send messages one after the other to get implicit ACKs
3. Send final ACK to know whether dialer succeeded
4. Listener is idempotent IF previous rollover failed (no ACK received)

## Detailed explanation

The two-generals problem tells us that it is not possible for both parties to know whether they other party successfully completed the protocol. However, what is possible is for the listener to know, whether the dialer succeeded if we add a final `Done` message. If the listener receives this, they know that the dialer completed successfully.

[...]

PR #2272, opened 2022-06-21
Rollover 2.0

### LINKED ISSUE #1622: Make `ProposeRollover` more explicit

da-kami:
This implements a more resilient rollover protocol where the taker proposes a `commit_txid` from which to roll over from to the maker. Protocol details can be found here: https://github.com/itchysats/itchysats/issues/1775

TODO:

- [x] Assert on the exact fees at the end of the rollover tests, make sure that the fees did not change compared to the code on master.
- [x] Add actor tests where we rollover twice and on the second rollover the taker uses the same `event_id`. (This might be a bit tricky)

Resolves [#1622](https://github.com/itchysats/itchysats/issues/1622), https://github.com/itchysats/itchysats/issues/1775

--- 

Implications when merging this PR and rolling it into production:

Previous ItchySats version that included the `libp2p` `rollover` protocol will not be compatible because the protocol contains breaking changes. Since we did *not* roll any version that included `libp2p` `rollover` out on Umbrel there should not be any users affected. 
The legacy rollover should not be affected by the breaking changes. The feature was implemented in a way that is backwards compatible (to the legacy version). This has to be tested on `testnet`.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -426,12 +426,22 @@ impl Cfd {
             OfferRejected => {
                 self.aggregated.state = CfdState::Rejected;
             }
-            RolloverCompleted { dlc, funding_fee } => {
+            RolloverCompleted {
+                dlc,
+                funding_fee,
+                complete_fee,
+            } => {
                 self.aggregated.rollover_state = None;
                 self.expiry_timestamp = dlc.as_ref().map(|dlc| dlc.settlement_event_id.timestamp());
                 self.aggregated.latest_dlc = dlc;
-                self.aggregated.fee_account =
-                    self.aggregated.fee_account.add_funding_fee(funding_fee);
+
+                self.aggregated.fee_account = match complete_fee {
+                    None => self.aggregated.fee_account.add_funding_fee(funding_fee),
+                    Some(complete_fee) => {
+                        self.aggregated.fee_account.from_complete_fee(complete_fee)
+                    }
+                };
+
                 self.accumulated_fees = self.aggregated.fee_account.balance();
 
                 self.aggregated.state = CfdState::Open;
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -303,8 +303,12 @@ impl Actor {
                         .await
                         .context("Failed to send Msg2")?;
 
-                    let revoked_commit =
-                        finalize_revoked_commits(&dlc, own_cfd_txs.commit.1, msg2)?;
+                    let revoked_commit = finalize_revoked_commits(
+                        &dlc,
+                        own_cfd_txs.commit.1,
+                        msg2,
+                        rollover_params.complete_fee_before_rollover(),
+                    )?;
 
                     let dlc = Dlc {
                         identity: dlc.identity,
@@ -328,7 +332,7 @@ impl Actor {
                         refund_timelock: rollover_params.refund_timelock,
                     };
 
-                    emit_completed(order_id, dlc, funding_fee, &executor).await;
+                    emit_completed(order_id, dlc, funding_fee, complete_fee, &executor).await;
 
                     Ok(())
                 }
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -306,23 +306,6 @@ impl Actor {
                     let revoked_commit =
                         finalize_revoked_commits(&dlc, own_cfd_txs.commit.1, msg2)?;
 
-                    let _msg3 = framed
-                        .next()
-                        .timeout(ROLLOVER_MSG_TIMEOUT)
-                        .await
-                        .with_context(|| format_expect_msg_within("Msg3", ROLLOVER_MSG_TIMEOUT))?
-                        .context("Empty stream instead of Msg3")?
-                        .context("Unable to decode dialer Msg3")?
-                        .into_rollover_msg()?
-                        .try_into_msg3()?;
-
-                    framed
-                        .send(ListenerMessage::RolloverMsg(Box::new(RolloverMsg::Msg3(
-                            RolloverMsg3,
-                        ))))
-                        .await
-                        .context("Failed to send Msg3")?;
-
                     let dlc = Dlc {
                         identity: dlc.identity,
                         identity_counterparty: dlc.identity_counterparty,
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -13,6 +13,7 @@ use futures::SinkExt;
 use futures::StreamExt;
 use libp2p_core::PeerId;
 use maia_core::secp256k1_zkp::schnorrsig;
+use model::olivia::BitMexPriceEventId;
 use model::Dlc;
 use model::FundingRate;
 use model::OrderId;
@@ -32,6 +33,7 @@ use super::protocol;
 type ListenerConnection = (
     Framed<Substream, JsonCodec<ListenerMessage, DialerMessage>>,
     PeerId,
+    (BitMexPriceEventId, model::CompleteFee),
 );
 
 /// Permanent actor to handle incoming substreams for the `/itchysats/rollover/1.0.0`
@@ -120,28 +122,32 @@ impl Actor {
         } = msg;
         let order_id = propose.order_id;
 
-        if let Err(e) = self
+        let (from_event_id, from_complete_fee) = match self
             .executor
             .execute(order_id, |cfd| {
                 cfd.verify_counterparty_peer_id(&peer.into())?;
-                cfd.start_rollover()
+                cfd.start_rollover_maker(propose.from_commit_txid)
             })
             .await
         {
-            tracing::warn!(%order_id, "Rollover failed after handling taker proposal: {e:#}");
+            Ok(event_id) => event_id,
+            Err(e) => {
+                tracing::warn!(%order_id, "Rollover failed after handling taker proposal: {e:#}");
 
-            // We have to append failed to ensure that we can rollover in the future
-            // The cfd logic might otherwise prevent us from starting a rollover if there is still
-            // one ongoing that was not properly ended.
-            emit_failed(order_id, e, &self.executor).await;
+                // We have to append failed to ensure that we can rollover in the future
+                // The cfd logic might otherwise prevent us from starting a rollover if there is
+                // still one ongoing that was not properly ended.
+                emit_failed(order_id, e, &self.executor).await;
 
-            return;
-        }
+                return;
+            }
+        };
 
         // In case we fail to accept/reject some proposals might never get cleaned up. Given that
         // the taker will retry this should not do much harm because we will replace the proposal
         // with a new one. This is acceptable for now.
-        self.pending_protocols.insert(order_id, (framed, peer));
+        self.pending_protocols
+            .insert(order_id, (framed, peer, (from_event_id, from_complete_fee)));
     }
 
     async fn handle(&mut self, msg: Accept) -> Result<()> {
@@ -152,9 +158,10 @@ impl Actor {
             short_funding_rate,
         } = msg;
 
-        let (mut framed, _) = self.pending_protocols.remove(&order_id).with_context(|| {
-            format!("No active protocol for {order_id} when accepting rollover")
-        })?;
+        let (mut framed, _, from_params) =
+            self.pending_protocols.remove(&order_id).with_context(|| {
+                format!("No active protocol for {order_id} when accepting rollover")
+            })?;
 
         let mut tasks = Tasks::default();
         tasks.add_fallible(
@@ -175,6 +182,7 @@ impl Actor {
                                 .accept_rollover_proposal(
                                     tx_fee_rate,
                                     funding_rate,
+                                    Some(from_params),
                                     RolloverVersion::V3,
                                 )?;
 
--- a/daemon/src/rollover/maker.rs
+++ b/daemon/src/rollover/maker.rs
@@ -302,14 +302,15 @@ impl Actor {
                         .try_into_msg2()?;
 
                     // reveal revocation secrets to the counterparty
-                    framed
+                    if let Err(e) = framed
                         .send(ListenerMessage::RolloverMsg(Box::new(RolloverMsg::Msg2(
                             RolloverMsg2 {
                                 revocation_sk: dlc.revocation,
                             },
                         ))))
-                        .await
-                        .context("Failed to send Msg2")?;
+                        .await {
+                        tracing::warn!(%order_id, "Failed to last rollover message to taker, this rollover will likely be retried by the taker: {e:#}");
+                    }
 
                     let revoked_commit = finalize_revoked_commits(
                         &dlc,
--- a/daemon/src/rollover/protocol.rs
+++ b/daemon/src/rollover/protocol.rs
@@ -29,7 +29,6 @@ use model::olivia;
 use model::olivia::BitMexPriceEventId;
 use model::Cet;
 use model::Dlc;
-use model::FeeFlow;
 use model::FundingFee;
 use model::FundingRate;
 use model::OrderId;
@@ -232,22 +231,22 @@ pub enum CompleteFee {
     Nein,
 }
 
-impl From<FeeFlow> for CompleteFee {
-    fn from(fee_flow: FeeFlow) -> Self {
-        match fee_flow {
-            FeeFlow::LongPaysShort(a) => CompleteFee::LongPaysShort(a),
-            FeeFlow::ShortPaysLong(a) => CompleteFee::ShortPaysLong(a),
-            FeeFlow::Nein => CompleteFee::Nein,
+impl From<model::CompleteFee> for CompleteFee {
+    fn from(complete_fee: model::CompleteFee) -> Self {
+        match complete_fee {
+            model::CompleteFee::LongPaysShort(a) => CompleteFee::LongPaysShort(a),
+            model::CompleteFee::ShortPaysLong(a) => CompleteFee::ShortPaysLong(a),
+            model::CompleteFee::None => CompleteFee::Nein,
         }
     }
 }
 
-impl From<CompleteFee> for FeeFlow {
-    fn from(fee_flow: CompleteFee) -> Self {
-        match fee_flow {
-            CompleteFee::LongPaysShort(a) => FeeFlow::LongPaysShort(a),
-            CompleteFee::ShortPaysLong(a) => FeeFlow::ShortPaysLong(a),
-            CompleteFee::Nein => FeeFlow::Nein,
+impl From<CompleteFee> for model::CompleteFee {
+    fn from(complete_fee: CompleteFee) -> Self {
+        match complete_fee {
+            CompleteFee::LongPaysShort(a) => model::CompleteFee::LongPaysShort(a),
+            CompleteFee::ShortPaysLong(a) => model::CompleteFee::ShortPaysLong(a),
+            CompleteFee::Nein => model::CompleteFee::None,
         }
     }
 }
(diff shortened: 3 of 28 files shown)


daemon/src/projection.rs
```diff
@@ -418,12 +418,22 @@ impl Cfd {
...
-                self.aggregated.fee_account =
-                    self.aggregated.fee_account.add_funding_fee(funding_fee);
+
+                if let Some(complete_fee) = complete_fee {
+                    self.aggregated.fee_account =
+                        self.aggregated.fee_account.from_complete_fee(complete_fee);
+                } else {
+                    self.aggregated.fee_account =
+                        self.aggregated.fee_account.add_funding_fee(funding_fee);
+                }
```
bonomat: Is clippy not complaining about this? I think it would be nicer to use a `match` statement.

daemon/src/rollover/taker.rs
```diff
@@ -270,8 +272,13 @@ impl Actor {
                                 .into_rollover_msg()?
                                 .try_into_msg2()?;
 
-                            let revoked_commit =
-                                finalize_revoked_commits(&dlc, own_cfd_txs.commit.1, msg2)?;
+                            // TODO: Shouldn't the adaptor sig be the one of the dlc.commit_tx?
```
bonomat: You might be right: we want to store the _revoked_ commit signatures, i.e. the one from the last commit transaction which should be in `dlc.commit`. @luckysori : what do you think ? 
Was this is a bug?

da-kami:
Note concerning this comment https://github.com/itchysats/itchysats/issues/1775#issuecomment-1162610819:

The current tests are actually not affected because it's implement in a way that if the current active `event_id` from the `dlc` is picked the maker does not apply the check if the `event_id` is already expired. This is the same as the previous behavior.

Once we add more tests where the taker pricks a previous `commit_txid` for the rollover we would run into this problem though.

sqlite-db/migrations/20220622055305_add_complete_fee_to_rollover_tables.sql
```diff
@@ -0,0 +1,19 @@
+ALTER TABLE
+    rollover_completed_event_data
+ADD
+    -- The complete fee including this rollover (contrary to the funding_fee this field contains the accumulated fees until now).
+    COLUMN complete_fee text NULL;
```
bonomat: I think this should be `INTEGER`

sqlite-db/migrations/20220622055305_add_complete_fee_to_rollover_tables.sql
```diff
@@ -0,0 +1,19 @@
...
+ALTER TABLE
+    rollover_completed_event_data
```
bonomat, line 7: You could also combine these two alter statements: 

```
ALTER TABLE
    rollover_completed_event_data
ADD
    COLUMN complete_fee text NULL,
    COLUMN complete_fee_flow text NULL;
```

sqlite-db/src/closed.rs
```diff
@@ -297,8 +297,16 @@ impl ClosedCfdInputAggregate {
...
             RolloverAccepted => {}
             RolloverRejected => {}
-            RolloverCompleted { dlc, funding_fee } => {
-                self.fee_account = self.fee_account.add_funding_fee(funding_fee);
+            RolloverCompleted {
+                dlc,
+                funding_fee,
+                complete_fee,
+            } => {
+                if let Some(complete_fee) = complete_fee {
```
bonomat: same as earlier, I think it would be nicer to have match statements

sqlite-db/src/models.rs
```diff
@@ -901,3 +901,138 @@ impl From<Settlement> for model::Settlement {
...
 }
+
+#[derive(Copy, Clone, Debug, PartialEq, sqlx::Type)]
+pub enum FeeFlow {
+    LongPaysShort,
+    ShortPaysLong,
+    None,
+}
+
+pub fn into_complete_fee(
```
bonomat, line 912: Maybe this can be changed to a `impl From` block?

```
impl From<(Option<FeeFlow>, Option<i64>) for Option<model::CompleteFee> {

}
```

daemon/src/rollover/protocol.rs
```diff
@@ -232,22 +231,22 @@ pub enum CompleteFee {
...
-impl From<FeeFlow> for CompleteFee {
-    fn from(fee_flow: FeeFlow) -> Self {
-        match fee_flow {
-            FeeFlow::LongPaysShort(a) => CompleteFee::LongPaysShort(a),
-            FeeFlow::ShortPaysLong(a) => CompleteFee::ShortPaysLong(a),
-            FeeFlow::Nein => CompleteFee::Nein,
+impl From<model::CompleteFee> for CompleteFee {
+    fn from(complete_fee: model::CompleteFee) -> Self {
+        match complete_fee {
+            model::CompleteFee::LongPaysShort(a) => CompleteFee::LongPaysShort(a),
```
bonomat, line 239: I'm dreaming of a world where we get rid of `LongPaysShort` and `ShortPaysLong` terminology and just express everything with positive and negative numbers...

model/src/cfd.rs
```diff
@@ -898,10 +898,64 @@ impl Cfd {
...
+        } else {
+            match dlc
+                .revoked_commit
+                .iter()
+                .find(|revoke_commit| revoke_commit.txid == from_tx_id)
+            {
+                Some(revoke_commit) => revoke_commit.settlement_event_id.ok_or_else(|| anyhow!(
+                    "Proposed commit-txid {} not eligible for rollover because no event-id attached", from_tx_id
+                ))?,
+                None => bail!("Unknown commit-txid {} proposed by taker", from_tx_id),
```
bonomat: This error would mean the taker has a commit-txid the maker doesn't know about right? Meaning, the taker succeeded in a former rollover but the maker failed?

model/src/cfd.rs
```diff
@@ -913,14 +967,25 @@ impl Cfd {
...
         let candidate_event_id = olivia::next_announcement_after(now + self.settlement_interval);
+
+        let from_event_id = if let Some(from_event_id) = from_event_id {
+            from_event_id
+        } else {
+            self.dlc
+                .as_ref()
+                .context("Cannot roll over without DLC")?
+                .settlement_event_id
+        };
```
bonomat: I think this would benefit from a short comment why we are taking `self.dlc.settlement_event_id`.

model/src/cfd.rs
```diff
@@ -913,14 +967,25 @@ impl Cfd {
...
+
         let hours_to_charge = match version {
             rollover::Version::V1 => 1,
             rollover::Version::V2 => self.hours_to_extend_in_rollover(now)?,
-            rollover::Version::V3 => {
-                self.hours_to_extend_in_rollover_based_on_event(candidate_event_id, now)?
-            }
+            rollover::Version::V3 => self.hours_to_extend_in_rollover_based_on_event(
+                candidate_event_id,
```
bonomat: Below you rename the arg to `to_event_id`. I think it might be worth also renaming it here.

model/src/cfd.rs
```diff
@@ -913,14 +967,25 @@ impl Cfd {
         }
 
         let now = OffsetDateTime::now_utc();
-
         let candidate_event_id = olivia::next_announcement_after(now + self.settlement_interval);
+
+        let from_event_id = if let Some(from_event_id) = from_event_id {
+            from_event_id
+        } else {
```
bonomat: What do you think of adding a log message saying that this is a retry? 
That way we can track how often a rollover is being retried.

bonomat (approved):
Good job. A few minor comments but overall concept should be sound

sqlite-db/migrations/20220622055305_add_complete_fee_to_rollover_tables.sql
da-kami: I tried, but could not get it working. I think this it unfortunately not possible in `sqlite`, see: 

https://stackoverflow.com/questions/6172815/sqlite-alter-table-add-multiple-columns-in-a-single-statement
https://www.sqlite.org/lang_altertable.html

daemon/src/rollover/protocol.rs
da-kami: I am not sure that will make the code easier to read and understand, but feel free to refactor this at some point :)

model/src/cfd.rs
da-kami: Yes. Given that the maker now fires and forgets the last message this should never happen, unless the `maker` fails on the "last mile", e.g. when saving `RolloverCompleted` into the database.
I don't think we can overcome that problem without the maker messaging the taker about a failure. 
The taker could then roll-over from the previous event, but I am not sure about the security assumptions for such a scenario. Definitely out of scope for now.

daemon/src/rollover/taker.rs
da-kami: I added this TODO on the way when I noticed this, thanks for reacting! I will move this into a separate PR so we can discuss it in context!

sqlite-db/src/models.rs
da-kami: After giving this a quick try: Not possible because `Option` is not defined within the crate so we run into:

```
Only traits defined in the current crate can be implemented for arbitrary types [E0117]
```

I don't think there is a way around this at the moment - if you know one let me know :)

model/src/cfd.rs
da-kami: This is what was previously picked (inside the function `hours_to_extend_in_rollover_based_on_event`) but I'm happy to add a comment!

sqlite-db/migrations/20220622055305_add_complete_fee_to_rollover_tables.sql
bonomat: oh, learned something new today. Thanks for links.

sqlite-db/src/models.rs
bonomat: oh well, you could probably simplify it to

```
impl From<(FeeFlow, i64) for model::CompleteFee {

}
```

but feel free to leave as is.

model/src/cfd.rs
da-kami: Note: I added the statement when we `start` the rollover because I think it makes more sense to log the information there. 
At this point in time we cannot say for sure if it is a retry, because the `start_rollover_maker` logic sets the `from_event_id` also when it is the current one. I implemented it this way because I think it makes the feature more complete. The fact that we have to evaluate if `from_event_id` is `Some` in `accept_rollover_proposal` is just due to being backwards compatible. 

In the future we can opt for making the `settlement_event_id` always set in `RevokeCommit`, i.e. remove the `Option` and then this check for `Some` in `accept_rollover_proposal` would go away.

sqlite-db/src/models.rs
da-kami: I'd have to deal with the `Options` multiple times upon conversion then, because this code is re-used by two tables.
The `Options` are actually quite annoying - and only needed to be backwards compatible. I think I'll leave it for now and focus on the tests

daemon/src/rollover/taker.rs
da-kami: see https://github.com/itchysats/itchysats/pull/2280

daemon/src/rollover/protocol.rs
klochowicz: We had that in the beginning, didn't we? then we decided that it will be easier and less confusing to use this terminology. Happy to move back to positive and negative numbers, as long as we have a clear doc where we state explicitly what is what.

da-kami:
Totally worth adding the rollover retry test (https://github.com/itchysats/itchysats/pull/2272/commits/615fc6955f8eb85791b28241d89fad2a5255238e)! I found several bugs and missing parts on the way! 

Retry from a previous `commit_txid` is tested now, we can opt for adding even more extensive tests to also test what happens if we go back multiple rollovers - but the internal logic to pick the `from_event_id` and `from_complete_fee` is the same, so I expect it to behave the same!

daemon/src/rollover/maker.rs
```diff
@@ -120,28 +122,32 @@ impl Actor {
...
                 cfd.verify_counterparty_peer_id(&peer.into())?;
-                cfd.start_rollover()
+                cfd.start_rollover_maker(propose.from_commit_txid)
             })
             .await
         {
-            tracing::warn!(%order_id, "Rollover failed after handling taker proposal: {e:#}");
+            Ok(event_id) => event_id,
+            Err(e) => {
+                tracing::warn!(%order_id, "Rollover failed after handling taker proposal: {e:#}");
```
da-kami, line 135: Note: I think we should change this into a `reject` scenario. The maker can send a `reject` back to the taker at this point. 

I did not change this yet because I am not sure about side-effects (I think they should be minimal, but still).

daemon/src/rollover/protocol.rs
da-kami: > We had that in the beginning

IMHO this is only half of the truth - what we had initially was incomplete. To make it complete we introduced `FeeAccount` and other structs on the way. I acknowledge that changing the logic of`+` and `-` instead to an `enum` might mean more code, but I would be careful stating that it can easily be made "better" by using `+` and `-`. We re-wrote this code three times already. 

 I am happy to see another refactoring to this part as long as it has at least the same test coverage that we have now :)

daemon/src/rollover/maker.rs
da-kami: see https://github.com/itchysats/itchysats/issues/2294

daemon/src/rollover/taker.rs
```diff
@@ -51,6 +53,8 @@ impl xtra::Actor for Actor {
 pub struct ProposeRollover {
     pub order_id: OrderId,
     pub maker_peer_id: PeerId,
+    pub from_commit_txid: Txid,
+    pub from_settlement_event_id: BitMexPriceEventId,
 }
```
klochowicz, line 58: why not be explicit and also state `until_settlement_event_id` here as well? Then it would be unambiguous until when we could extend the rollover?
I believed that what #1622 originally imagined too.

model/src/cfd.rs
```diff
@@ -780,7 +784,10 @@ impl Cfd {
         self.commit_tx.is_some()
     }
 
-    pub fn can_auto_rollover_taker(&self, now: OffsetDateTime) -> Result<(), NoRolloverReason> {
+    pub fn can_auto_rollover_taker(
+        &self,
+        now: OffsetDateTime,
+    ) -> Result<(Txid, BitMexPriceEventId), NoRolloverReason> {
```
klochowicz, line 790: it would be great to introduce some type that discribes this tuple, because it's pretty random in this context.

for example:
```rust
struct RolloverStartParams {
from_commit_tx: Txid,
from_settlement_event_id: BitMexPriceEventId
}
```

model/src/cfd.rs
```diff
@@ -894,10 +903,74 @@ impl Cfd {
...
+        let event = CfdEvent::new(self.id, EventKind::RolloverStarted);
+
+        Ok((event, from_event_id, from_complete_fee))
+    }
+
     pub fn accept_rollover_proposal(
         self,
         tx_fee_rate: TxFeeRate,
         funding_rate: FundingRate,
+        from_params: Option<(BitMexPriceEventId, CompleteFee)>,
```
klochowicz, line 973: again, it's unfortunate to send `None`. Rust can be really descriptive, it would be nicer to send an enum that clarifies this better :)

model/src/cfd.rs
```diff
@@ -1538,15 +1643,15 @@ impl Cfd {
...
+        let to_settlement_time = to_event_id.timestamp();
+        let time_to_extend = to_settlement_time - from_settlement_time;
 
         let hours_to_extend = time_to_extend.whole_hours();
 
         if !hours_to_extend.is_positive() {
             bail!(
-                "Cannot rollover if candidate event ID is no later than current event ID:
-                 {candidate_settlement_time} <= {current_settlement_time}",
+                "Cannot rollover if to event ID is not later than from event ID:
```
klochowicz, line 1653: nit: "to" and "from" don't read too well. perhaps "until" instead of "to" would make it easier? 

or maybe "start" and "end" as a pair?

model/src/cfd.rs
```diff
@@ -2216,6 +2331,10 @@ pub struct RevokedCommit {
     // To monitor revoked commit transaction
     pub txid: Txid,
     pub script_pubkey: Script,
+    // To enable the taker to rollover from a previous commit-tx
+    // The maker requires these parameters to reason about the previous commit
+    pub settlement_event_id: Option<BitMexPriceEventId>,
+    pub complete_fee: Option<CompleteFee>,
```
klochowicz: this comment has a high likelihood to become stale (maybe we could extract a struct where we could put these 2 fields, so that it will never get out of place?)

model/src/cfd.rs
```diff
@@ -3345,15 +3465,15 @@ mod tests {
     #[test]
     fn given_current_settlement_in_12_hours_and_candidate_in_19_then_7_hour_extension() {
         for now in common_time_boundaries() {
-            let current_event_id = BitMexPriceEventId::with_20_digits(now + 12.hours());
-            let candidate_event_id = BitMexPriceEventId::with_20_digits(now + 19.hours());
```
klochowicz, line 3349: actually, "current" and "candidate" naming sounds pretty good too. I wouldn't mind having this one throughout.

klochowicz (approved):
I have added some comments, but nothing blocking; happy to fix most of the nitpicks myself in a follow-up. 

concept ACK.

daemon/src/rollover/taker.rs
da-kami: The maker already tells the taker what event to use. I am not sure what value it would have to add it to the taker's proposal.

model/src/cfd.rs
da-kami: fair enough, I changed it into more descriptive doc comments

da-kami:
I am happy to see the rest of the changes in follow-ups. Will see to squash and bors this.

model/src/cfd.rs
da-kami: Note that all these are only `Option`s because of backwards compatibility for the legacy protocol versions. 

I would hope to clean that up once we are a few versions ahead. If you feel strong about it feel free to do a small refactoring round as a follow up!

model/src/cfd.rs
da-kami: Yeah, I am not super happy with `to` and `from` neither, but I found `candidate` super confusing, because for me it indicates that it is only a candidate but might not be used (i.e. something else would be used instead). That's not what we want to express here. It's literally a to-from scenario (from a cost perspective, as in the hours from and to you get charged for). Happy to see a follow up on that that changes it throughout (but please not candidate :)

PR #2276, opened 2022-06-22
Bump time to 0.3.11

klochowicz:


bonomat (approved):
Thanks. I think the revert commit was not necessary though:)

klochowicz:
> Thanks. I think the revert commit was not necessary though:)

indeed - I thought a pinned crate could have behaved differently when trying to run `cargo upgrade --workspace time@0.3.11`, so I wanted to play it safe - went back and verified that I could have done it without it; will remember for the future.

Issue #2279, opened 2022-06-23
Save the `protocol` version in the rollover data


da-kami:
For rollover it can be essential to know with what protocol version the rollover was done so breaking changes do not affect already existing open cfds (i.e. open cfds that were opened prior to the breaking change).

In order to achieve this we will have to save the protocol version with the rollover data and open the rollover substream according to the version.

klochowicz:
Note that this would require us to run either multiple rollover actors on the taker side in order to support this.

bonomat:
I'm not sure I understand why this is needed: 

- for the taker it shouldn't matter, he always only has 1 possible rollover protocol to chose from. Which means, if we make a breaking change to the protocol we need to provide an upgrade path for the taker or accept his position will be force closed.
- the maker on the other hand will have potentially 2+ rollover protocols. Since the taker is requesting the rollover, the maker _knows_ which rollover protocol he is expecting i.e. it is the libp2p-rollover-protocol version. Unless we break changes on the db layer, I don't think the maker needs additional information.

da-kami:
This concerns scenarios where you *upgrade* a taker. We might have scenarios where a taker also has two versions, to ensure compatibility for open positions.

bonomat:
Do you have a scenario in mind where it would make sense to deal with 2 rollover protocols on the taker side? 

If we really can't migrate a rollover protocol version from one to the other (i.e. have to run 2 on the taker side) I would accept force closing the position. I think the maintenance hell out of this would cost us too much. 

We can revisit this if we ever need it but for now I don't see the need. 

Closing this for now. Feel free to re-open if you have more input @da-kami :)

PR #2280, opened 2022-06-23
Potential revoke commit fix

da-kami:
When working on #2272 I noticed that using the adaptor sig of the newly created commit tx here might not be correct.

Note on the second commit https://github.com/itchysats/itchysats/commit/e251fdddceb437d3517ffda62c1731ef753fc7de - This might not be backwards compatible so we might drop it.

Note that I think this might also be a breaking change for existing positions once we rollout rollover over libp2p! This might mean that positions that were opened prior to a release might get force closed. 

Note: I did *not* adapt this code in `setup_contract_deprecated`.

---

Note on breaking protocol changes:

We are currently not capable to distinguish different rollover versions per `Cfd`. We might want to add this feature by saving what rollover protocol version was used upon saving a rollover. Otherwise we accept that upon breaking change in the protocol *existing* positions that were opened prior to the breaking change might get force closed. I documented this in https://github.com/itchysats/itchysats/issues/2279

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -1,10 +1,4 @@
 use crate::future_ext::FutureExt;
-use crate::rollover;
-use crate::rollover::protocol::build_and_verify_cets_and_refund;
-use crate::rollover::protocol::build_commit_descriptor;
-use crate::rollover::protocol::build_own_cfd_transactions;
-use crate::rollover::protocol::build_punish_params;
-use crate::rollover::protocol::finalize_revoked_commits;
 use crate::shared_protocol::format_expect_msg_within;
 use crate::shared_protocol::verify_adaptor_signature;
 use crate::shared_protocol::verify_cets;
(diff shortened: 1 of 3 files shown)


bonomat (approved):
I don't think this is a breaking change since we don't make use of this yet anyway 😅. 
@luckysori : what do you think about this? Is this correct now?

klochowicz (commented):
makes sense, based on my limited understanding of this part of the code.

daemon/src/setup_contract.rs
```diff
@@ -459,7 +459,7 @@ pub async fn roll_over(
         .context("Empty stream instead of Msg2")?
         .try_into_msg2()?;
 
-    let revoked_commit = finalize_revoked_commits(&dlc, own_cfd_txs.commit.1, msg2.into())?;
+    let revoked_commit = finalize_revoked_commits(&dlc, dlc.commit.1, msg2.into())?;
```
luckysori: :wrench: This `roll_over` function is not used anymore. Just remove it? 

Maybe you meant to modify the deprecated version? I wouldn't bother because it would be a breaking change for a feature that is not implemented fully. And the code should disappear soon.

luckysori (approved):
Amazing find! An 8-month old bug, a bit older than my daughter :baby:

I guess it would have been easier to find if we actually used and tested the revocation/punishment feature!

daemon/src/setup_contract.rs
da-kami: Oh my, my effort to re-use code in that function was totally useless because it is unused 😅

I added a separate commit to remove the function. 
We might actually still adapt `contract_setup_deprecated` in that case - I thought that this was an even older version, but it is the code path used by the currently released takers. Given that I think we should update that code in a separate commit because it might be around for some more time until the takers upgraded.

PR #2281, opened 2022-06-23
Fetch announcements SETTLEMENT_INTERVAL + 2

### LINKED ISSUE #2264: `Failed to get announcement` happens too often
We see `Failed to get announcement` errors quite frequently in the logs. This should however only be an error of we fail to fetch announcements repeatedly. If this happens once for an announcement we should log on `debug`.

Ideas for solution:

There are different ways to achieve this. I think the cleanest way would be to let the `oracle` actor track failures for specific announcements. If we fail to fetch the same announcement x times then we report the error.

bonomat:
For a rollover to happen successfully we need to know the oracle announcement details.
Our actor is checking if a new announcement can be fetched every SYNC_ANNOUNCEMENTS_INTERVAL and
ANNOUNCEMENT_LOOKAHEAD hours into the future. Given we rollover every for
hour model::SETTLEMENT_INTERVAL into the future, we want to have at least model::SETTLEMENT_INTERVAL
announcements ready. Due to sync interval coincidence, it might happen that we do not have
synced for a specific announcement yet. Hence, we need to fetch more announcements.
We fetch model::SETTLEMENT_INTERVAL + 2 announcement into the future because of this example:

Assume the last fetch was at 01.01.2022 00:59:55, i.e. 5 seconds before midnight and we would
have synced for model::SETTLEMENT_INTERVAL+1 we would have synced announcements until 02.01.2022
01:00:00. A rollover request happening exactly at 01.01.2022 01:00:00 would ask for the
announcement at 02.01.2022 02:00:00 because of how olivia::next_announcement_after works. Note:
even if the underlying logic of olivia::next_announcement_after changes, fetching model::
SETTLEMENT_INTERVAL + 2 won't hurt.

resolves #2264

### DIFF
--- a/daemon/src/oracle.rs
+++ b/daemon/src/oracle.rs
@@ -36,11 +36,29 @@ pub struct Actor {
     announcements: HashMap<BitMexPriceEventId, (OffsetDateTime, Vec<schnorrsig::PublicKey>)>,
     pending_attestations: HashSet<BitMexPriceEventId>,
     executor: command::Executor,
-    announcement_lookahead: Duration,
     tasks: Tasks,
     db: sqlite_db::Connection,
     client: reqwest::Client,
 }
+
+/// We want to fetch at least this much announcements into the future
+///
+/// For a rollover to happen successfully we need to know the oracle announcement details.
+/// Our actor is checking if a new announcement can be fetched every SYNC_ANNOUNCEMENTS_INTERVAL and
+/// ANNOUNCEMENT_LOOKAHEAD hours into the future. Given we rollover every for
+/// hour model::SETTLEMENT_INTERVAL into the future, we want to have at least
+/// model::SETTLEMENT_INTERVAL announcements ready. Due to sync interval coincidence, it might
+/// happen that we do not have synced for a specific announcement yet. Hence, we need to fetch more
+/// announcements. We fetch model::SETTLEMENT_INTERVAL + 2 announcement into the future because of
+/// this example:
+///
+/// Assume the last fetch was at 01.01.2022 00:59:55, i.e. 5 seconds before midnight and we would
+/// have synced for model::SETTLEMENT_INTERVAL+1 we would have synced announcements until 02.01.2022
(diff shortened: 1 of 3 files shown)


daemon/src/oracle.rs
```diff
@@ -36,12 +36,14 @@ pub struct Actor {
...
     pending_attestations: HashSet<BitMexPriceEventId>,
     executor: command::Executor,
-    announcement_lookahead: Duration,
     tasks: Tasks,
     db: sqlite_db::Connection,
     client: reqwest::Client,
 }
 
+/// We want to fetch at least this much announcements into the future
+const ANNOUNCEMENT_LOOKAHEAD: Duration = Duration::hours(26);
```
da-kami, line 61: 🙃 I liked it better when we saw the dependency on `SETTLEMENT_INTERVAL` directly here, i.e.:

```
const ANNOUNCEMENT_LOOKAHEAD: Duration = SETTLEMENT_INTERVAL + Duration::hours(2);
```

The test ensures that it is, but when just looking at `ANNOUNCEMENT_LOOKAHEAD` it is not clear why it is set to `26`.

da-kami (approved):
nice find!

daemon/src/oracle.rs
klochowicz: agreed with the suggestion

daemon/src/oracle.rs
```diff
@@ -388,3 +380,15 @@ impl Attestation {
...
 }
+
+#[cfg(test)]
+pub mod tests {
+    #[test]
+    fn ensure_lookahead_constant() {
+        use time::Duration;
+        assert_eq!(
+            crate::oracle::ANNOUNCEMENT_LOOKAHEAD,
+            model::SETTLEMENT_INTERVAL + Duration::hours(2)
```
klochowicz, line 407: maybe we could write down the reasoning for +2 here or at the constant definition?

klochowicz (approved):
👍

daemon/src/oracle.rs
bonomat: I had that at the beginning but rust does not allow it.

PR #2284, opened 2022-06-23
Is this test supposed to pass?

da-kami:
If we calculate the funding fee for `24h` shouldn't that equal an accumulated funding fee of 24 x `1h`?
The test fails with:

Left:  SignedAmount(0.00000100 BTC)
Right: SignedAmount(0.00000120 BTC)

so the accumulated fee apparently is more.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -3745,6 +3745,43 @@ mod tests {
         assert!(!sane, "an oracle event id that is outdated got accepted")
     }
 
+    #[test]
+    fn calculated_funding_fee_of_24h_should_equal_24_hourly_fees() {
+        let price = Price::new(dec!(10000)).unwrap();
+        let quantity = Usd::new(dec!(10));
+        let long_leverage = Leverage::TWO;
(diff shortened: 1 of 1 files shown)


bonomat:
The answer is no because of our rounding strategy: 

if the funding rate is `100` sats a day, then per hour it is `100/24=4.16666666667` given we round to `MidpointAwayFromZero` this ends up being `5` sats/hour or `5*24=120` per day.

klochowicz:
> The answer is no because of rounding errors:
> 
> if the funding rate is `100` sats a day, then per hour it is `100/24=4.16666666667` given we round to `MidpointAwayFromZero` this ends up being `5` sats/hour or `5*24=120` per day.

correct. If anything, we should assert that the funding fee for hourly rollovers for 24h is at least the amount changed for the day. 

this is explicitly asserted in this proptest (as in, that it is OK to charge more):
https://github.com/itchysats/itchysats/blob/master/model/src/cfd.rs#L3612

klochowicz:
closing as the question got resolved

Issue #2282, opened 2022-06-23
Restricting rollover to `event_id` within `24h`


da-kami:
This is a simple feature with a rat-tail of changes for the test setup.
First described in https://github.com/itchysats/itchysats/issues/1775#issuecomment-1162610819

## Feature:

The maker should not allow rollovers of `dlc`s where the `settlement_event_id` has already expired.

In the context of the taker proposing `commit_tx` - "Rollover 2.0":
The maker should not allow rollovers if the `from_event_id` associated to the `RevokeCommit`, that is picked by the `commit_txid` sent by the taker in the proposal has already expired (i.e. is more than `24h` in the past given the current `24h` settlement interval constraint).

This would mean that we add a check to the `maker`'s `start_rollover_maker` logic that validates that the `from_event-id` is still valid, i.e. `!has_likely_occurred`. 
 
Additionally we should `bail!` in `hours_to_extend_in_rollover_based_on_event` if `hours_left` is negative and *not* charge `24h` as fallback. This is for cases where `start_rollover` still works, but then the `event_id` is expired upon handling the accept. It is fair to say that this rollover should fail early, i.e. fail in `hours_to_extend_in_rollover_based_on_event` because the `cfd` is just not eligible anymore.

Without adding this feature the maker's fees over time are not ensured properly. 

## On controlling time in the actor tests

Currently our actor tests are set up with fixed (old) event-ids from the past. If we want the maker to check this, we will have a dependency on `now` (i.e. the current time) in the tests.
This could be fixed by "smartly picking" an event-id in the test, but this is super error prone if we run the test close to the hour because we might stumble into an edge case where the test assertions do not match the dynamic input.

I am advising against trying to "pick" an event ID based on the current time in the tests, but keep using fixed events in the past. In addition we would have to...

To *properly* fix this we will need a way to "control time" in the test setup. We have discussed in the past to be able to control the time in the tests as a feature. Unfortunately this is not that easy to achieve. Since there are quite a lot of different solution for this possible, I opened a separate discussion for this problem: https://github.com/itchysats/itchysats/discussions/2283

Until we discussed how to achieve controlling time here is a hacky solution to "pick" an event-id:

## Hacky solution to dynamically pick `event-id` `timestamps` based on current time

[...]

PR #2285, opened 2022-06-23
Rollover asserts exact fees

da-kami:
Note that initially I tried to accumulate the rollover fees which lead me to open https://github.com/itchysats/itchysats/pull/2284

I had to use `FundingFee::calculate` to make the test pass. I find that a bit odd, but it could be expected behavior.

---

I am planning to now add tests using different event IDs to create a failing test on master that should then pass on #2272  . This will be a bit tricky, because the fees are always `24h` at the moment given that we use events with outdated timestamps. I might touch on problems described in #2282

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -1,5 +1,6 @@
 use anyhow::Context;
 use daemon::bdk::bitcoin::Amount;
+use daemon::bdk::bitcoin::SignedAmount;
 use daemon::connection::ConnectionStatus;
 use daemon::connection::MAX_RECONNECT_INTERVAL_SECONDS;
 use daemon::projection::CfdOrder;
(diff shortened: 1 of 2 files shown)


model/src/cfd.rs
```diff
@@ -1535,6 +1535,8 @@ impl Cfd {
         if !hours_left.is_positive() {
             tracing::warn!("Rolling over a contract that can be settled non-collaboratively");
 
+            // TODO: This is an incorrect fallback. Should be changed to the actual hours we
+            // have  to charge or we have to disallow using events older than 24h
```
klochowicz: can we raise an issue so we won't forget this? (if it's not yet tracked)

klochowicz (approved):
cool! I like that this test actually shows how to use this tremendously complicated `FeeAccount`!

daemon-tests/tests/happy_path.rs
```diff
@@ -412,20 +419,24 @@ async fn double_rollover_an_open_cfd() {
     rollover_an_open_cfd(Position::Short, 2).await;
 }
 
-async fn rollover_an_open_cfd(maker_position: Position, nr_rollovers: u8) {
+async fn rollover_an_open_cfd(maker_position: Position, nr_rollovers: i64) {
```
klochowicz: nit: IMHO `u8` chosen by you earlier was better due type safety guarantees, e.g.that no-one could try out e. `-1` value. Perhaps consider casting to `i64` internally? (it cannot fail from such a small number)

klochowicz (approved):

daemon-tests/tests/happy_path.rs
da-kami: I only changed it because I needed a an `i64` further down in the code. I don't think it matter much but I'm happy to cast instead.

klochowicz: I figured :) it's probably cleaner to cast it when you use it instead of changing the API (not a biggie though, it's just a test...).

model/src/cfd.rs
da-kami: I adapted https://github.com/itchysats/itchysats/issues/2282 and added this change there. Removed `TODO`.

PR #2292, opened 2022-06-23
Introduce /itchysats/order protocol over libp2p

### LINKED ISSUE #1770: Design proposal for `/itchysats/order`

klochowicz:
Fix https://github.com/itchysats/itchysats/issues/1770.

The new design follows the proposal in #1770. This means that
the taker now generates an `OrderId` when _placing_ the order. We
currently reuse the same type for the offer (legacy) and the taker's
order, which is a bit confusing (but hard to avoid, because our model
wants an `OrderId`). With the new protocol the taker ignores the
`OrderId` associated with the offer, but we have to keep it around for
legacy connections.

One actor test got removed - with libp2p we're not changing the offers after
one was picked; the OrderId is generated on the taker side.

Co-authored-by: Lucas Soriano del Pino <l.soriano.del.pino@gmail.com>

### DIFF
--- a/daemon-tests/tests/offer.rs
+++ b/daemon-tests/tests/offer.rs
@@ -1,25 +1,16 @@
 use daemon::bdk::bitcoin::Amount;
-use daemon::projection::CfdOrder;
-use daemon::projection::CfdState;
+use daemon::projection::CfdOffer;
 use daemon::projection::MakerOffers;
-use daemon_tests::confirm;
 use daemon_tests::dummy_offer_params;
 use daemon_tests::flow::is_next_offers_none;
 use daemon_tests::flow::next_maker_offers;
-use daemon_tests::flow::next_with;
-use daemon_tests::flow::one_cfd_with_state;
 use daemon_tests::start_both;
-use daemon_tests::wait_next_state;
 use model::Leverage;
 use model::Position;
-use model::Usd;
 use otel_tests::otel_test;
-use rust_decimal_macros::dec;
-use std::time::Duration;
-use tokio_extras::time::sleep;
 
 #[otel_test]
-async fn taker_receives_order_from_maker_on_publication() {
+async fn taker_receives_offer_from_maker_on_publication() {
     let (mut maker, mut taker) = start_both().await;
 
     assert!(is_next_offers_none(taker.offers_feed()).await.unwrap());
@@ -35,146 +26,23 @@ async fn taker_receives_order_from_maker_on_publication() {
     assert_eq_offers(published, received);
 }
 
-#[otel_test]
-async fn taker_takes_order_and_maker_rejects() {
-    let (mut maker, mut taker) = start_both().await;
-
-    is_next_offers_none(taker.offers_feed()).await.unwrap();
-
-    maker
-        .set_offer_params(dummy_offer_params(Position::Short))
-        .await;
-
-    let (_, received) = next_maker_offers(maker.offers_feed(), taker.offers_feed())
-        .await
-        .unwrap();
-
-    let order_id = received.short.unwrap().id;
-
-    taker.mocks.mock_oracle_announcement().await;
-    maker.mocks.mock_oracle_announcement().await;
-    taker
-        .system
-        .take_offer(order_id, Usd::new(dec!(100)), Leverage::TWO)
-        .await
-        .unwrap();
-
-    wait_next_state!(order_id, maker, taker, CfdState::PendingSetup);
-
-    maker.system.reject_order(order_id).await.unwrap();
-
-    wait_next_state!(order_id, maker, taker, CfdState::Rejected);
-}
-
-#[otel_test]
-async fn another_offer_is_automatically_created_after_taker_takes_order() {
-    let (mut maker, mut taker) = start_both().await;
-
-    is_next_offers_none(taker.offers_feed()).await.unwrap();
-
-    maker
-        .set_offer_params(dummy_offer_params(Position::Short))
-        .await;
-
-    let (_, received) = next_maker_offers(maker.offers_feed(), taker.offers_feed())
-        .await
-        .unwrap();
-
-    let order_id_take = received.short.unwrap().id;
-
-    taker.mocks.mock_oracle_announcement().await;
-    maker.mocks.mock_oracle_announcement().await;
-    taker
-        .system
-        .take_offer(order_id_take, Usd::new(dec!(10)), Leverage::TWO)
-        .await
-        .unwrap();
-
-    // For the taker the offer is reset after taking it, so we can't take the same one twice
-    is_next_offers_none(taker.offers_feed()).await.unwrap();
-
-    let (maker_update, taker_update) = next_maker_offers(maker.offers_feed(), taker.offers_feed())
-        .await
-        .unwrap();
-
-    let new_order_id_maker = maker_update.short.unwrap().id;
-    let new_order_id_taker = taker_update.short.unwrap().id;
-
-    assert_ne!(
-        new_order_id_taker, order_id_take,
-        "Another offer should be available, and it should have a different id than first one"
-    );
-
-    assert_ne!(
-        new_order_id_maker, order_id_take,
-        "Another offer should be available, and it should have a different id than first one"
-    );
-
-    assert_eq!(
-        new_order_id_taker, new_order_id_maker,
-        "Both parties have the same new id"
-    )
-}
-
-#[otel_test]
-async fn taker_takes_order_and_maker_accepts_and_contract_setup() {
-    let (mut maker, mut taker) = start_both().await;
-
-    is_next_offers_none(taker.offers_feed()).await.unwrap();
-
-    maker
-        .set_offer_params(dummy_offer_params(Position::Short))
-        .await;
-
-    let (_, received) = next_maker_offers(maker.offers_feed(), taker.offers_feed())
-        .await
-        .unwrap();
-
-    let order_id = received.short.unwrap().id;
-
-    taker.mocks.mock_oracle_announcement().await;
-    maker.mocks.mock_oracle_announcement().await;
-
-    taker
-        .system
-        .take_offer(order_id, Usd::new(dec!(100)), Leverage::TWO)
-        .await
-        .unwrap();
-    wait_next_state!(order_id, maker, taker, CfdState::PendingSetup);
-
-    maker.mocks.mock_party_params().await;
-    taker.mocks.mock_party_params().await;
-
-    maker.mocks.mock_wallet_sign_and_broadcast().await;
-    taker.mocks.mock_wallet_sign_and_broadcast().await;
-
-    maker.system.accept_order(order_id).await.unwrap();
-    wait_next_state!(order_id, maker, taker, CfdState::ContractSetup);
-
-    sleep(Duration::from_secs(5)).await; // need to wait a bit until both transition
-    wait_next_state!(order_id, maker, taker, CfdState::PendingOpen);
-
-    confirm!(lock transaction, order_id, maker, taker);
-    wait_next_state!(order_id, maker, taker, CfdState::Open);
-}
-
 fn assert_eq_offers(published: MakerOffers, received: MakerOffers) {
     match (published.long, received.long) {
         (None, None) => (),
-        (Some(published), Some(received)) => assert_eq_orders(published, received),
-        (None, Some(_)) => panic!("Long orders did not match: Published None, received Some "),
-        (Some(_), None) => panic!("Long orders did not match: Published Some, received None "),
+        (Some(published), Some(received)) => assert_eq_offer(published, received),
+        (None, Some(_)) => panic!("Long offers did not match: Published None, received Some "),
+        (Some(_), None) => panic!("Long offers did not match: Published Some, received None "),
     }
 
     match (published.short, received.short) {
         (None, None) => (),
-        (Some(published), Some(received)) => assert_eq_orders(published, received),
-        (None, Some(_)) => panic!("Short orders did not match: Published None, received Some "),
-        (Some(_), None) => panic!("Short orders did not match: Published Some, received None "),
+        (Some(published), Some(received)) => assert_eq_offer(published, received),
+        (None, Some(_)) => panic!("Short offers did not match: Published None, received Some "),
+        (Some(_), None) => panic!("Short offers did not match: Published Some, received None "),
     }
 }
 
-fn assert_eq_orders(mut published: CfdOrder, received: CfdOrder) {
+fn assert_eq_offer(mut published: CfdOffer, received: CfdOffer) {
     // we fix the leverage to TWO for our test
     let fixed_leverage = Leverage::TWO;
 
@@ -197,12 +65,12 @@ fn assert_eq_orders(mut published: CfdOrder, received: CfdOrder) {
             .initial_funding_fee_per_lot
             * -1
     );
-    // align leverage details so we can assert on the order
+    // align leverage details so we can assert on the offer
     published.leverage_details = received.leverage_details.clone();
 
     assert_eq!(published, received);
 
-    // Hard-coded to match the dummy_new_order()
+    // Hard-coded to match the dummy_new_offer()
     assert_eq!(received.opening_fee.unwrap(), Amount::from_sat(2));
     assert_eq!(received.funding_rate_hourly_percent, "0.00100");
 }
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -16,7 +16,8 @@ use model::libp2p::PeerId;
 use model::olivia;
 use model::Identity;
 use model::Leverage;
-use model::Order;
+use model::Offer;
+use model::OfferId;
 use model::OrderId;
 use model::Price;
 use model::Role;
@@ -56,6 +57,7 @@ pub mod monitor;
 pub mod noise;
 mod online_status;
 pub mod oracle;
+pub mod order;
 pub mod position_metrics;
 pub mod process_manager;
 pub mod projection;
@@ -82,9 +84,10 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
 pub const N_PAYOUTS: usize = 200;
 
 pub struct TakerActorSystem<O, W, P> {
-    pub cfd_actor: Address<taker_cfd::Actor<O, W>>,
+    pub cfd_actor: Address<taker_cfd::Actor>,
     pub connection_actor: Address<connection::Actor>,
     wallet_actor: Address<W>,
+    _oracle_actor: Address<O>,
     pub auto_rollover_actor: Address<auto_rollover::Actor>,
     pub price_feed_actor: Address<P>,
     executor: command::Executor,
@@ -176,6 +179,27 @@ where
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
+        let (order_supervisor, order) = Supervisor::new({
+            let oracle = oracle_addr.clone();
+            let db = db.clone();
+            let process_manager = process_manager_addr;
+            let wallet = wallet_actor_addr.clone();
+            let projection = projection_actor.clone();
+            let endpoint = endpoint_addr.clone();
+            move || {
+                order::taker::Actor::new(
+                    n_payouts,
+                    oracle_pk,
+                    oracle.clone().into(),
+                    (db.clone(), process_manager.clone()),
+                    (wallet.clone().into(), wallet.clone().into()),
+                    projection.clone(),
+                    endpoint.clone(),
+                )
+            }
+        });
+        tasks.add(order_supervisor.run_log_summary());
+
         let (collab_settlement_supervisor, libp2p_collab_settlement_addr) = Supervisor::new({
             let endpoint_addr = endpoint_addr.clone();
             let executor = executor.clone();
@@ -192,14 +216,9 @@ where
         let (connection_actor_addr, connection_actor_ctx) = Context::new(None);
         let cfd_actor_addr = taker_cfd::Actor::new(
             db.clone(),
-            wallet_actor_addr.clone(),
-            oracle_pk,
             projection_actor,
-            process_manager_addr,
-            connection_actor_addr.clone(),
-            oracle_addr.clone(),
             libp2p_collab_settlement_addr,
-            n_payouts,
+            order,
             maker_identity,
             PeerId::from(
                 maker_multiaddr
@@ -214,6 +233,7 @@ where
         let (rollover_supervisor, libp2p_rollover_addr) = Supervisor::new({
             let endpoint_addr = endpoint_addr.clone();
             let executor = executor.clone();
+            let oracle_addr = oracle_addr.clone();
             move || {
                 rollover::v_2_0_0::taker::Actor::new(
                     endpoint_addr.clone(),
@@ -324,6 +344,7 @@ where
             cfd_actor: cfd_actor_addr,
             connection_actor: connection_actor_addr,
             wallet_actor: wallet_actor_addr,
+            _oracle_actor: oracle_addr,
             auto_rollover_actor: auto_rollover_addr,
             price_feed_actor,
             executor,
@@ -337,20 +358,22 @@ where
     }
 
     #[instrument(skip(self), err)]
-    pub async fn take_offer(
+    pub async fn place_order(
         &self,
-        order_id: OrderId,
+        offer_id: OfferId,
         quantity: Usd,
         leverage: Leverage,
-    ) -> Result<()> {
-        self.cfd_actor
-            .send(taker_cfd::TakeOffer {
-                order_id,
+    ) -> Result<OrderId> {
+        let order_id = self
+            .cfd_actor
+            .send(taker_cfd::PlaceOrder {
+                offer_id,
                 quantity,
                 leverage,
             })
             .await??;
-        Ok(())
+
+        Ok(order_id)
     }
 
     #[instrument(skip(self), err)]
--- a/daemon/src/order.rs
+++ b/daemon/src/order.rs
@@ -1,0 +1,5 @@
+pub const PROTOCOL_NAME: &str = "/itchysats/order/1.0.0";
+
+pub mod maker;
+mod protocol;
+pub mod taker;
--- a/daemon/src/order/maker.rs
+++ b/daemon/src/order/maker.rs
@@ -1,0 +1,372 @@
+use crate::command;
+use crate::oracle;
+use crate::oracle::NoAnnouncement;
+use crate::order::protocol;
+use crate::order::protocol::MakerMessage;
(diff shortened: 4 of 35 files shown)


klochowicz:
I have slightly modified @luckysori 's PR:
- removed `Contracts` changes, as they are orthogonal to the network migration, and the discussion on github is not resolved yet - moreover, they caused grief during rebasing, when types got migrated). 
- addressed some leftover TODOs
- made the tests pass 
- removed one actor level test (we no longer remove the offer after an order got taken)

I didn't end up going back to recreating the old protocol in order to save time; this solution was close to finish, and it fixes some awkwardness in our protocol (need to recreate offers). I would be keen on trying to let it run on testnet over the weekend.

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,332 @@
...
+        }
+    }
+
+    pub async fn handle_new_inbound_stream(&mut self, msg: NewInboundSubstream) -> Result<()> {
+        let NewInboundSubstream { peer, stream } = msg;
+
+        let mut framed = Framed::new(stream, JsonCodec::<MakerMessage, TakerMessage>::new());
+
+        let (
+            id,
```
da-kami: With the protocol change of letting the `taker` define the `OrderId` (independent from the `OfferId`), the maker has to make sure that *all* parameters provided by the taker are sane! This is not what is currently implemented.

I am happy to go that way, but I think this will require changes when handling the decision to accept/reject. If there is not ticket for that yet please create one.

daemon/src/order/taker.rs
```diff
@@ -0,0 +1,306 @@
...
+            let n_payouts = self.n_payouts;
+            let projection = self.projection.clone();
+            async move {
+                tracing::info!(order = ?msg, "Placing order");
+
+                let PlaceOrder {
+                    id,
+                    quantity,
+                    leverage,
+                    position,
```
da-kami:
```suggestion
                    maker_position,
```

da-kami: Otherwise the `position.counterposition()` below when creating the `Cfd` comes out of nowhere.

daemon/src/order/taker.rs
```diff
@@ -0,0 +1,306 @@
...
+                    .execute(id, |cfd| cfd.start_contract_setup())
+                    .await?;
+
+                let (sink, stream) = framed.split();
+
+                let announcement = get_announcement
+                    .send(oracle::GetAnnouncement(oracle_event_id))
+                    .await??;
+
+                let dlc = setup_contract::new(
```
da-kami, line 177: I think it would be a good idea to split up contract setup into reusable chunks as well and then inline the messaging here. Can be done in a follow up.

da-kami (commented):
I assume that you are aware that the protocol changes will require logic changes to the maker's decision to evaluate if an `order` is acceptable.

We should evaluate implementing at least some of the acceptance criteria before we can roll this out. Unless there is already a ticket for this (I did not find one) please create a ticket for that (automation). We have to make sure to do this before releasing.

daemon/src/order/taker.rs
klochowicz: I would only do it if it's necessary; if it works well without this, I see no point in doing so; I doubt that such a refactor would even be more maintainable...


> I assume that you are aware that the protocol changes will require logic changes to the maker's decision to evaluate if an `order` is acceptable.
> 
> We should evaluate implementing at least some of the acceptance criteria before we can roll this out. Unless you are already working on it please create a ticket for that. We have to make sure to do this before releasing.

great point. I will add it an additional commit here, because without this it's not a complete PR :)

da-kami:
I like separating offer and order more clearly. 
We might also want to add tests that show that one taker (or different takers for that matter) can take the same `offer_id` twice now (resulting in different `order_ids`.

daemon/src/order/maker.rs
klochowicz: good point. will add a check :)

klochowicz:
I have rebased the PR and finally made the tests pass again; it's not ready yet for another review though as I need to add:

TODO: 
- [x] re-add identity (we're using it in grafana now)
- [x] validate the offer by tying order and offer id together (as discussed with @da-kami )

Marking as draft until then.

da-kami:
I rebased on latest master for now, will see to work in the changes of the `taker` sending the offer id when creating an order instead of all the parameters. 

@klochowicz what is the second TODO about? - i.e.:

>  re-add identity (we're using it in grafana now)

klochowicz:
> I rebased on latest master for now, will see to work in the changes of the `taker` sending the offer id when creating an order instead of all the parameters.
> 
> @klochowicz what is the second TODO about? - i.e.:
> 
> > re-add identity (we're using it in grafana now)

the TODO was about re-adding the `identity` - @luckysori has hardcoded a random "identity" field as it wasn't used, but we're now using it for some reporting.

da-kami:
> > I rebased on latest master for now, will see to work in the changes of the `taker` sending the offer id when creating an order instead of all the parameters.
> > @klochowicz what is the second TODO about? - i.e.:
> > > re-add identity (we're using it in grafana now)
> 
> the TODO was about re-adding the `identity` - @luckysori has hardcoded a random "identity" field as it wasn't used, but we're now using it for some reporting.

Sorry, but could you provide a bit more context? What "reporting" are we talking about? Is this important for the future as well? 
This identity is only relevant for the old connection as far as I understand, how would we do the "reporting" once we remove it?

klochowicz:
> > > I rebased on latest master for now, will see to work in the changes of the `taker` sending the offer id when creating an order instead of all the parameters.
> > > @klochowicz what is the second TODO about? - i.e.:
> > > > re-add identity (we're using it in grafana now)
> > 
> > 
> > the TODO was about re-adding the `identity` - @luckysori has hardcoded a random "identity" field as it wasn't used, but we're now using it for some reporting.
> 
> Sorry, but could you provide a bit more context? What "reporting" are we talking about? Is this important for the future as well? This identity is only relevant for the old connection as far as I understand, how would we do the "reporting" once we remove it?

I've linked you the reports on matrix, and explained them in greater detail.

daemon/src/lib.rs
```diff
@@ -85,9 +86,10 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
 pub const N_PAYOUTS: usize = 200;
 
 pub struct TakerActorSystem<O, W, P> {
-    pub cfd_actor: Address<taker_cfd::Actor<O, W>>,
+    pub cfd_actor: Address<taker_cfd::Actor>,
     pub connection_actor: Address<connection::Actor>,
     wallet_actor: Address<W>,
+    _oracle_actor: Address<O>, // NOTE: Hard to supervise because of our tests
```
klochowicz: why is it hard to supervise this because of the tests? oracle actor is mocked in the tests, but we're using DI here, so it should not matter?

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+            .context("Unable to decode order")?;
+
+        Ok(order)
+    }
+
+    async fn pick_offer(&self, offer_id: OfferId) -> Result<Offer> {
+        // TODO: Evaluate if this can be a problem during sending new offers to all connected peers
+        // -> Should be fine because the offer is already outdated by then and we should actually
+        // wait until finished?
+        // -> Cannot use send_async_next here - is that expected?
```
klochowicz: yes, it's expected that we can't use `send_async_next` here; we are awaiting the completion of the handler as we need the result back :)

note: we could use `send(msg).priority(NUMBER)` if we wanted to utilise higher priority to ensure that this protocol does go smoothly (at the risk of actually getting an outdated offer, so I'm not sure whether it's worth it)

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+            .timeout(ORDER_TIMEOUT, || tracing::debug_span!("receive order"))
+            .await
+            .context("Timout when waiting for order")?
+            .context("Stream terminated")?
+            .context("Unable to decode order")?;
+
+        Ok(order)
+    }
+
+    async fn pick_offer(&self, offer_id: OfferId) -> Result<Offer> {
```
klochowicz, line 110: in order to address the TODO below it would be good to instrument this function:

```suggestion
    #[instrument(skip(self), err)]
    async fn pick_offer(&self, offer_id: OfferId) -> Result<Offer> {
```

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+
+    async fn await_order(
+        &mut self,
+        framed: &mut Framed<Substream, JsonCodec<MakerMessage, TakerMessage>>,
+    ) -> Result<TakerMessage> {
+        let order = framed
+            .next()
+            .timeout(ORDER_TIMEOUT, || tracing::debug_span!("receive order"))
+            .await
+            .context("Timout when waiting for order")?
```
klochowicz:
```suggestion
            .context("Timeout when waiting for order")?
```

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+            sign,
+            projection,
+            n_payouts,
+            decision_senders: HashMap::default(),
+            db,
+            latest_offers,
+        }
+    }
+
+    async fn await_order(
```
klochowicz:
```suggestion
    #[instrument(skip(self), err)]
    async fn await_order(
```

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+        };
+
+        match offer {
+            None => bail!("Offer with id {offer_id} not found in current offers"),
+            Some(offer) => Ok(offer),
+        }
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
```
klochowicz: `message_impl` that used to implement message handlers is not needed anymore. :)

```suggestion
#[xtra_productivity]
```

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        let oracle_event_id = offer.oracle_event_id;
+
+        let cfd = Cfd::from_order(
+            id,
+            &offer,
+            quantity,
+            Identity::new(x25519_dalek::PublicKey::from(
+                *b"hello world, oh what a beautiful",
+            )),
```
klochowicz, line 190: oops, I was meant to reinstate this one! I only did it on the other side. we'll need to pass it in the message unfortunately, there's no other way. will add one more commit :)

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,385 @@
...
+            Decision::Reject(_) => protocol::Decision::Reject,
+        }
+    }
+}
+
+impl fmt::Display for Decision {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let s = match self {
+            Decision::Accept(_) => "accept",
+            Decision::Reject(_) => "reject",
```
klochowicz: nit: why lowercase?

daemon/src/order/taker.rs
```diff
@@ -0,0 +1,269 @@
...
+                    Some(maker_peer_id.into()),
+                    Role::Taker,
+                    leverage,
+                );
+
+                // If this fails we shouldn't try to append
+                // `ContractSetupFailed` to the nonexistent CFD
+                if let Err(e) = db.insert_cfd(&cfd).await {
+                    tracing::error!("Inserting new cfd failed: {e:#}");
+                    return anyhow::Ok(());
```
klochowicz, line 124: do we need to return `Ok()` here? this obscures the error in the traces....

daemon/src/taker_cfd.rs
```diff
@@ -27,16 +19,15 @@ use model::Role;
...
+use xtras::SendAsyncSafe;
 
 #[derive(Clone)]
 pub struct CurrentMakerOffers(pub Option<MakerOffers>);
 
 #[derive(Clone, Copy)]
-pub struct TakeOffer {
-    pub order_id: OrderId,
+pub struct PlaceOrder {
+    pub offer_id: OrderId,
```
klochowicz: maybe at least add a TODO to rollout `OfferId` in daemon later, when we decide to drop legacy protocols?

daemon/src/taker_cfd.rs
```diff
@@ -164,108 +135,50 @@ impl<O, W> Actor<O, W> {
...
-        if !order_to_take.is_safe_to_take(OffsetDateTime::now_utc()) {
-            bail!("The maker's offer appears to be outdated, refusing to take offer",);
+            .take_order(offer_id)
+            .0
+            .context("Offer to take could not be found in current maker offers, you might have an outdated offer")?;
+
+        // TODO: Consider removing this. Why can't we just place the
+        // order we want, if it's the want we selected in the UI? The
+        // maker should simply reject our request to place an order if
+        // it's outdated
```
klochowicz: 👍 I'm for removing this. maker should deal with outdated orders by just rejecting them. at least we have visibility how often it happens, otherwise we don't know that takers run into this frustrating problem.

maker/src/contract_setup.rs
```diff
@@ -19,15 +19,15 @@ use maia_core::PartyParams;
...
+use model::Offer;
 use model::Role;
 use model::Usd;
 use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 use xtras::SendAsyncSafe;
 
 pub struct Actor {
-    order: Order,
+    order: Offer,
```
klochowicz: should this be `offer`? 

```suggestion
   offer: Offer,
```

sqlite-db/src/rollover/insert.rs
```diff
@@ -80,7 +80,7 @@ async fn insert_rollover_completed_event_data(
     dlc: &Dlc,
     funding_fee: FundingFee,
     complete_fee: Option<CompleteFee>,
-    offer_id: models::OrderId,
+    order_id: models::OrderId,
```
klochowicz, line 84: 🤔should this be order_id or offer_id? how can we tell which one of the two we need here?

xtra-libp2p-offer/src/lib.rs
```diff
@@ -207,8 +207,8 @@ mod tests {
         })
     }
 
-    fn dummy_order(position: Position) -> Order {
-        Order::new(
+    fn dummy_order(position: Position) -> Offer {
```
klochowicz: 🙃

```suggestion
    fn dummy_offer(position: Position) -> Offer {
```

klochowicz (commented):
great piece of work, thanks for linking OrderId to OfferId!

It seems I can't approve this PR, as I was the author of it? 😆

from my side ACK (after squashing before merging)

klochowicz:
@da-kami @luckysori I've added in total 2 commits into this PR  that bring `Identity` (taker_id and maker_id) into Order protocol. please have a look at these too :)

daemon/src/lib.rs
luckysori: Unfortunately I don't remember the specifics because it has been so long since I wrote this comment. But I just couldn't make it work nicely. Feel free to give it a go!

daemon-tests/tests/offer.rs
```diff
@@ -101,12 +101,12 @@ async fn another_offer_is_automatically_created_after_taker_takes_order() {
...
-        new_order_id_taker, order_id_take,
+        new_order_id_taker, offer_id,
         "Another offer should be available, and it should have a different id than first one"
     );
 
     assert_ne!(
-        new_order_id_maker, order_id_take,
+        new_order_id_maker, offer_id,
         "Another offer should be available, and it should have a different id than first one"
     );
```
luckysori: 🔧 Rename to `new_offer_id_{taker,maker}`.

daemon/src/order.rs
```diff
@@ -0,0 +1,5 @@
+pub const PROTOCOL_NAME: &str = "/ipfs/order/0.1.0";
```
luckysori: 🔧 I would expect this identifier to be:

- `/itchysats/order`; and
- version `1.0.0`, to be consistent with the first version of our other protocols.

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,383 @@
...
+        let latest_offers = match self
+            .latest_offers
+            .send(xtra_libp2p_offer::maker::GetLatestOffers)
+            .await
+        {
+            Ok(latest_offers) => latest_offers,
+            Err(e) => {
+                bail!("Failed to retrieve latest offer from offers actor: {e:#}")
+            }
+        };
```
luckysori: 🔧 Using `anyhow::Context` is good here.

```suggestion
        let latest_offers = match self
            .latest_offers
            .send(xtra_libp2p_offer::maker::GetLatestOffers)
            .await
            .context("Failed to retrieve latest offer from offers actor")?;
```

And the same goes for the two `match`es below.

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, msg: NewInboundSubstream, ctx: &mut xtra::Context<Self>) {
+        let NewInboundSubstream { peer, stream } = msg;
+
+        let mut framed = Framed::new(stream, JsonCodec::<MakerMessage, TakerMessage>::new());
+
+        let order = match self.await_order(&mut framed).await {
```
luckysori: 🙃 Too many awaits in one line.

```suggestion
        let order = match self.receive_order(&mut framed).await {
```

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,383 @@
...
+                    &ctx.address().expect("self to be alive"),
+                    future,
+                    move |e| async move {
+                        tracing::debug!(%peer, "Failed to send reject order message: {e}");
+                    },
+                );
+
+                return;
+            }
+        };
```
luckysori, line 180: 🤔 Do we actually need this? If someone comes and tries to place an order, why does the `maker` daemon care about which offer they were referencing? Shouldn't the daemon just let the automation (or whoever is operating the application) decide, since they can?

I think this has been discussed before, but I keep forgetting the arguments because intuitively it doesn't make sense to me.

daemon/src/order/maker.rs
luckysori: This is definitely an area where we can have a problem. I can already imagine the offer actor, overwhelmed with it's usual job of relaying offers, taking ages to reply to this message.

daemon/src/order/taker.rs
luckysori: I don't see the alternative: the task is fallible and we are early returning here.

sqlite-db/src/rollover/insert.rs
da-kami: This one was super confusing for myself, because when this was introduced we treated the offer as order (i.e. we always spoke of the offer as order and wanted to rename it to offer), hence @bonomat opted for naming this `offer_id`. However, with the introduction of clear boundaries for offer and order as part of my commit this actually is related to the order, and this *is* an `OrderId` here. From a lifecylce point of view when this is created the order is filled alrady.
Currently, the `offer_id` in the `Cfd` is just to identify the offer related to the `Cfd`, but not the `Cfd` itself. The `OrderId` identifies the `Cfd`, because if an order gets filled the taker creates the `Cfd` with a unique `OrderId`.


Note: The lifecycle of the `Cfd` can still be improved, we could introduce a dedicated `Order` and only once said `Order` is filled we create the actual `Cfd`. But I felt we can do that in another iteration on top of this work if we want to. So for now I kept the `cfds` table as a flattened out data conglomerate that includes all the information and the lifecylce start with creating the `Order` on the taker side. This means that, as before, we can have a `Cfd` in the database that was never actually filled by the maker - which is a bit weird. 
I think eventually a model of `Offer -> Order -> Cfd` would be correct, where the `offer_id` is linked to the `order_id` and the `order_id` *may be* the same as the `cfd_id`. One `offer_id` can be linked to multiple `order_id`s, one `order_id` should be linked to exactly one `cfd_id` (or `cfd_id` = `order_id`)

luckysori (commented):
Haven't reviewed all of it, but I have to stop now until a bit later.

daemon/src/lib.rs
klochowicz: all good! I'll try after this gets in, was just curious :)

daemon/src/order/maker.rs
da-kami: 😅 I wonder what my brain was thinking when I wrote this

daemon/src/order/maker.rs
da-kami: I agree. This was discussed to some extent, but it would be a change that requires adapting the automation. I would like to keep this changeset contained and just keep the logic as it was before, and then change this in another iteration.

This is a change that purely concerns the maker, so we can change this easily on top, I will see to create a ticket.

da-kami: Note: I am happy to action this straight away, so we also solve the problem of not needing to know the latest offer in this actor. 
But, we have to evaluate how the automation then keeps the state of the current offer. Currently it just does not keep that state as far as I know, so I don't think this is such a simple change and would not see it in scope of this PR.

luckysori: Thanks for your reply! I think it's fine to not tackle it immediately, particularly if it's not a simple change. If you create a ticket and we tackle it soon I will be more than happy.

daemon/src/order/maker.rs
da-kami: I think we can "solve" this by moving this decision into the automation, but we have to be aware that we will have to manage the state there then. I will see to write down a ticket for this and action it later, because it is not just copying code ;)
For now I think this solution is good enough, and I don't expect to run into too much trouble, because sending to each peer is in a separate task, so the offer actor should not be blocked for too long.

daemon/src/order/maker.rs
da-kami: Created :) see linked tickets

sqlite-db/src/rollover/insert.rs
da-kami: I created https://github.com/itchysats/itchysats/issues/2555 to shed some more light on this and outline how I envision to clean this up.

daemon/src/order/maker.rs
da-kami: I'll change it to the same name as the enum. @luckysori if this was important please let me know :)

daemon/src/taker_cfd.rs
da-kami: This should actually be `OfferId` here :)

daemon/src/taker_cfd.rs
da-kami: I disagree. The taker UI does not tell you that an offer is outdated, and I don't want to introduce this as part of this PR. If we want to remove this feature we can, but it should not be part if this changeset because it has notthing to do with it.
We introduces this to prevent the taker from taking outdated offers created by the maker (!) after we introduces replicating the offer in the daemon. This can still happen and this is a safety net for the taker.

maker/src/contract_setup.rs
da-kami: Yes and no 😅 
This is legacy code and I think no matter how much I try to change it, it will always be odd because the boundaries are just not right. I agree that there is quite a lot of weirdness in this file because of renaming `Order` to `Offer`. I am not sure if it's a good idea to try to clean this up if we remove it soon.

The problem is: If I change it here then outside it will be super weird, because we are passing the `order_to_take` in, and that cannot really be changed, because it is a construct that is offer and order at the same time 😅

da-kami: Given that this is the only instance where we literally have a direct conflict I'll change that one :)

da-kami:
Interesting: installing an using a new `sqlx-cli` to prepare the db can indeed lead to the changes in the `sqlx-data.json` file. I just had such an instance, where the CI complained and I could only reproduce it after installing the new version.

daemon/src/order/maker.rs
```diff
@@ -0,0 +1,372 @@
...
+            Identity::new(x25519_dalek::PublicKey::from(
+                *b"hello world, oh what a beautiful",
+            )),
+            Some(peer.into()),
+            Role::Maker,
+            leverage,
+        );
+
+        // If this fails we shouldn't try to append
+        // `ContractSetupFailed` to the nonexistent CFD
```
da-kami, line 197: Note: We will be able to overcome this with #2555

da-kami (approved):

PR #2293, opened 2022-06-24
Re-structure rollovers test setup for configurability

da-kami:
Split into `prepare_rollover` and `rollover`.
`prepare_rollover` prepares `maker`, `taker` and `fee_structure` to be used for the rollover; it prepares the open `cfd` and asserts on the initial fees.
`rollover` performs one rollover and can be parametrized.

--- 

One thing that is a bit unfortunate is that we now need access to the previously `private` `aggregated` in the `projection::Cfd` - but I think this is an acceptable change given that we can write better tests!
Potentially this could be configured with only being public in `cfg(test)` but I did not invest time into this.

---

This is for preparing more elaborate tests for #2272 
I thought it's better to keep the changeset smaller so I opened a separate PR with these improvements against master :)

### DIFF
--- a/daemon-tests/tests/happy_path.rs
+++ b/daemon-tests/tests/happy_path.rs
@@ -532,11 +532,21 @@ async fn rollover(
     // Ensure that the event ID of the latest dlc is the event ID used for rollover
     assert_eq!(
         oracle_data.announcement().id,
-        maker_cfd.aggregated.latest_dlc.unwrap().settlement_event_id
+        maker_cfd
+            .aggregated()
+            .latest_dlc()
+            .as_ref()
+            .unwrap()
+            .settlement_event_id
     );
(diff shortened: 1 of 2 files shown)


klochowicz (approved):
looks good! 
I have added a tiny cleanup commit at the end that exposes methods returning immutable borrowed fields instead of making the fields public (so it's clearer we're never modifying them). hope you don't mind.

klochowicz:
a bit more about the motivation for the direction of the cleanup commit:

- `cfg(test)` wouldn't work, as it's only useful for unit tests (inside the same file)
- `pub(crate)` would work for integration tests, but only within a crate; `daemon-tests` are in a separate crate unfortunately :( 

we could consider using doc(hidden) for these functions, but I'm not sure about that.

PR #2298, opened 2022-06-27
Fix to store the correct commit adaptor-sig in deprecated code

da-kami:
We don't know how long it will take to remove this code, so we better apply this fix.
It should not cause problems.


da-kami:
I also evaluated the re-use the rollover code there (so we can share code - but given the different `maia` versions it is hard to achieve this cleanly. I would not invest the time.

klochowicz (approved):

PR #2306, opened 2022-06-27
Use xtra upstream

Restioson:
This PR switches itchysats to an upstream branch of xtra. It also removes `IPromiseIamReturningStopAllFromStopping`, since the bug in xtra this compensated for was fixed in upstream.

# To do
- [x] ~~Switch to master after https://github.com/Restioson/xtra/pull/105 is merged, or find another solution to reimplement `SendAsyncSafe`~~ Another solution was found!
- [ ] Wait for https://github.com/Restioson/xtra/pull/113 to be merged and update the supervisor to use the `Controller` type, or find another way to do this.
- [x] Reimplement handler timeouts
    - [x] Itchysats side (extension trait & custom event loop)
    - [ ] Switch to master after https://github.com/Restioson/xtra/pull/112/ is merged (technically optional - this would just result in a dependency on a branch other than master)
- [x] Fix outstanding test failures
    - [x] `supervisor/restarted_actor_is_usable`
    - [x] `supervisor_can_delay_respawn`
    - [x] `supervisor_tracks_spawn_metrics`


Restioson:
Oops, this was meant to be a draft PR but I clicked the wrong button :sweat_smile: it's not quite ready yet (see the todo section)

klochowicz:
> Oops, this was meant to be a draft PR but I clicked the wrong button 😅 it's not quite ready yet (see the todo section)

I think #105 should be a stable link to the xtra repo, because I don't think you were meant to refer to #105 in itchysats? :)

klochowicz:
> Oops, this was meant to be a draft PR but I clicked the wrong button 😅 it's not quite ready yet (see the todo section)

converted to draft for you :)

Restioson:
> I think https://github.com/itchysats/itchysats/pull/105 should be a stable link to the xtra repo, because I don't think you were meant to refer to https://github.com/itchysats/itchysats/pull/105 in itchysats? :)

Yes, I've fixed this now :)

> converted to draft for you :)

Thanks!

Restioson:
To elaborate on the stop_all issue: currently, `stop_all` drains all messages and masks the channel as disconnected. So, if supervised actors `stop_all`, the supervisor cannot restart it properly, as `attach` will work fine but any attempt to send to the actor (even from itself) will fail. This should be addressed in xtra upstream, but does not technically block this PR.

Restioson:
# Impact of the open xtra PRs that affect this PR:

## https://github.com/Restioson/xtra/pull/112

This is required for handler timeouts to be instrumented nicely. This PR can be merged without it, but it would be depending on a branch of xtra rather than master. Then, when https://github.com/Restioson/xtra/pull/112 is merged, the dependency would need to be updated to point at master rather than the branch.

## https://github.com/Restioson/xtra/pull/113

This changes the following:
1. It fixes weird behaviour in `stop_all` to make it more supervisor friendly. It used to immediately disconnect the channel, resulting in any subsequent restarts failing. Now, it will just tell all the actors to stop, allowing them to be restarted later.
2. Adds the `Controller` type which is intended to be kept by supervisors instead of `Context`. There is currently an issue in xtra master which means that if a supervisor keeps a context, it will prevent the broadcast tail from advancing. This means that only `bound` broadcast messages can be sent in until the next sender will wait for the messages to be received. However, the `Context` kept by the supervisor is also counted amongst the receivers, and it will never receive the message, so therefore no messages beyond `bound` can ever be sent in. 
3. Fixes a bug where the bound of the broadcast channel can be enforced inconsistency, leading to edge cases of it having technically unbounded capacity. This was actually quite hard to write a test to observe, though it can always be backported to master, since it was something like a one line order-of-operations fix.

However, there is no actual impact on itchysats as of present from any of these changes, as itchysats does not use the broadcast features or `stop_all` (after [IPromiseIamReturningStopAllFromStopping](https://github.com/itchysats/itchysats/pull/2306/commits/edb53ffb64c2a5ef28c4bff71148e8aca48cc481)). Therefore, this PR could be merged without it. If so, when https://github.com/Restioson/xtra/pull/113 or something like it is merged, the supervisor would need to be updated to use `Controller` instead of `Context`, but I anticipate this to really just be a chore and not very difficult, since it isn't any less flexible than at present, given that you can easily convert between `Context`s and `Controller`s.

## Conclusion

This PR could be merged before both of the two PRs mentioned, with a little bit of work required in itchysats to update to xtra master after this. It could be worth it to get earlier access to trying out instrumentation and tracing metrics, but it's really down to what you all think is best :slightly_smiling_face:

Restioson:
Closed in favour of #2333.

PR #2320, opened 2022-06-29
Change log level

bonomat:
Debugs are normally not shown, meaning we would not see this log level. Hence, I propose to change it to warn.

### DIFF
--- a/xtra-libp2p-ping/src/ping.rs
+++ b/xtra-libp2p-ping/src/ping.rs
@@ -128,7 +128,7 @@ impl Actor {
             };
 
             let err_handler = move |e| async move {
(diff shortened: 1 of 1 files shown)


xtra-libp2p-ping/src/ping.rs
```diff
@@ -128,7 +128,7 @@ impl Actor {
             };
 
             let err_handler = move |e| async move {
```
klochowicz, line 130: nit: I believe you have a typo in the commit message - s/wanr/warn :)

klochowicz (approved):
warn seems reasonable on the taker side, on the maker it might end up too spammy - it will happen whenever a taker disappears (I guess we can always filter them out for stats/alerts).

PR #2321, opened 2022-06-29
Workaround to prevent overwriting RUST_LOG

bonomat:
So far we overwrote RUST_LOG log settings using . This is annoying because hardcoded log levels can't be changed later on. With this workaround we can have smart defaults but allow the user to overwrite the log level using RUST_LOG

// edit// unfortunately tracing does not allow us to do this differently. There is an open ticket which inspired this solution: https://github.com/tokio-rs/tracing/issues/1466


klochowicz (approved):
thanks for finding a workaround for this issue and linking the issue in the lib!

PR #2326, opened 2022-06-29
Use macos 12 in github actions

klochowicz:
For some reason, macos-latest still points to Big Sur (11).

Check whether the CI situation (a lot of failures on macos jobs) will improve
with the latest macos version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,7 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


klochowicz:
I'll merge this, hoping that this will help with the CI issues that we're experiencing (macos-12 build succeeded much faster than macos-latest used to, and even faster than ubuntu-latest).

bors r+

bonomat:
Good call. Thanks for looking into it

PR #2328, opened 2022-06-30
Add configurable load tests

da-kami:
This allows us to simulate multiple bobs (i.e. takers) for one alice (i.e. maker) to see if the endpoint / protocol handler can handle it.

---

Summary:

On my machine if I go above `150` bobs I see flakyness, above `200` I have consistent failure because something "hangs" - somehow the test never finishes. I cannot really explain what causes the hanging yet. Note that this is independent of how many times I trigger the protocol for each bob - with `200` bobs and `1` time I see consistent failure on my machine.

Note: On my machine it does not really make a difference if I trigger with one protocol and handler or multiple.

Note: If we go below `800-900` millis wait time for the taker to establish substreams we see connection failures. This scales (to some extent) with the number of bobs and protocols, if here is less substreams one can allow less waiting time. Establishing the connection is independent of that, only creating the substram is influenced by that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4616,6 +4616,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "tracing-test"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


da-kami:
closed in favor of https://github.com/itchysats/itchysats/pull/2403

PR #2329, opened 2022-06-30
Add metric for unique takers

bonomat:
I'd like to push this intro production asap so that we start collecting metrics. 

This metric shows us the unique takers with open/closed/.. positions.
We need the additional `all` position because some takers might have long AND short positions. If this is the case we would count them twice, i.e. `long` + `short` != `all`

It looks like this:

```
# HELP counterparty_number_total Total number of counterparties we had a position with on ItchySats.
# TYPE counterparty_number_total gauge
counterparty_number_total{position="all",status="closed"} 4
counterparty_number_total{position="all",status="failed"} 4
counterparty_number_total{position="all",status="new"} 6
counterparty_number_total{position="all",status="open"} 0
counterparty_number_total{position="all",status="refunded"} 3
counterparty_number_total{position="all",status="rejected"} 0
counterparty_number_total{position="long",status="closed"} 2
counterparty_number_total{position="long",status="failed"} 2
counterparty_number_total{position="long",status="new"} 2
counterparty_number_total{position="long",status="open"} 0
counterparty_number_total{position="long",status="refunded"} 2
counterparty_number_total{position="long",status="rejected"} 0
counterparty_number_total{position="short",status="closed"} 2
counterparty_number_total{position="short",status="failed"} 2
counterparty_number_total{position="short",status="new"} 3
counterparty_number_total{position="short",status="open"} 0
counterparty_number_total{position="short",status="refunded"} 2
counterparty_number_total{position="short",status="rejected"} 0
# HELP positions_number_total Total number of positions on ItchySats.
```

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -1,15 +1,19 @@
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
+use bdk::bitcoin::Amount;
 use futures::StreamExt;
+use model::calculate_margin;
 use model::CfdEvent;
 use model::ClosedCfd;
 use model::EventKind;
 use model::FailedCfd;
 use model::FailedKind;
 use model::Identity;
+use model::Leverage;
 use model::OrderId;
 use model::Position;
+use model::Role;
 use model::Settlement;
 use model::Usd;
 use rust_decimal::Decimal;
@@ -121,6 +125,8 @@ pub struct Cfd {
     id: OrderId,
     position: Position,
     quantity_usd: Usd,
+    margin: Amount,
+    margin_counterparty: Amount,
 
     state: AggregatedState,
     counterparty_network_identity: Identity,
@@ -143,10 +149,21 @@ impl sqlite_db::CfdAggregate for Cfd {
     type CtorArgs = ();
 
     fn new(_: Self::CtorArgs, cfd: sqlite_db::Cfd) -> Self {
+        let (our_leverage, counterparty_leverage) = match cfd.role {
+            Role::Maker => (Leverage::ONE, cfd.taker_leverage),
+            Role::Taker => (cfd.taker_leverage, Leverage::ONE),
+        };
+
+        let margin = calculate_margin(cfd.initial_price, cfd.quantity_usd, our_leverage);
+        let margin_counterparty =
+            calculate_margin(cfd.initial_price, cfd.quantity_usd, counterparty_leverage);
+
         Self {
             id: cfd.id,
             position: cfd.position,
             quantity_usd: cfd.quantity_usd,
+            margin,
+            margin_counterparty,
             state: AggregatedState::New,
             counterparty_network_identity: cfd.counterparty_network_identity,
             version: 0,
@@ -260,6 +277,9 @@ impl sqlite_db::ClosedCfdAggregate for Cfd {
             n_contracts,
             settlement,
             counterparty_network_identity,
+            role,
+            taker_leverage,
+            initial_price,
             ..
         } = closed_cfd;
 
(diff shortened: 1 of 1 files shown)


bonomat:
Something looks odd, will re-open later.

bonomat:
> Something looks odd, will re-open later.

fixed :D

daemon/src/position_metrics.rs
```diff
@@ -271,6 +275,7 @@ impl sqlite_db::ClosedCfdAggregate for Cfd {
             position,
             quantity_usd,
             state,
+            counterparty_network_identity,
```
luckysori, line 309: 🤡 I think this will need to be changed to the `PeerId` at some point. But we probably can't yet because not all positions have an associated `PeerId` (since it's an `Option` in the `Cfd`).

daemon/src/position_metrics.rs
```diff
@@ -382,6 +401,18 @@ mod metrics {
...
+        let short_takers = short.iter().map(|cfd| cfd.counterparty_network_identity);
+
+        let counterparties = long_takers.chain(short_takers).unique().count();
+
+        COUNTERPARTY_NUMBER_GAUGE
+            .with(&HashMap::from([
+                (POSITION_LABEL, POSITION_ALL_LABEL),
+                (STATUS_LABEL, status),
+            ]))
+            .set(counterparties as i64);
```
luckysori: 🔧 With this snippet you're trying to capture the takers that have both long and short positions, right?

I think for that we need to do [this](https://programming-idioms.org/idiom/218/list-intersection/3834/rust).

luckysori (commented):
I found one thing that could be wrong (or at least I didn't understand it).

daemon/src/position_metrics.rs
bonomat: Yes, once every peer has a peer_id we can change this. For now, every peer has a network identity .

daemon/src/position_metrics.rs
bonomat: Yes, this should capture all takers which have either a long or a short position but don't count them double if they have both. 

The problem I was facing is that on line 405 we will get duplicated entries, hence, I have to call `unique`. 
So that I don't have to call `unique` twice, I merge the two vecs first (line 408) and then call `unique`.

luckysori: Hmmm, I'm not sure I understand it yet.

When we call `set_metrics_for` on lines 402 and 403 we're counting all takers with at least long and short positions respectively.

We then _also_ want to capture takers which have both long _and_ short positions. But there's gonna be an overlap between `long`, `short` and `all` the way this is arranged, right? Is that wanted?

bonomat: Yes, the thing is that `long+short!=all`, for example

`long=[taker1, taker2]=2`
`short=[taker1, taker3]=2`
then we want all to be: 
`all=[taker1,taker2,taker3]=3`

luckysori: Okay, I misunderstood the meaning of `all` in this context.

It is not `all` in the sense of takers with all kinds of positions.

It is `all` in the sense of all the takers with any kind of position.

Thanks for clarifying and sorry for the confusion!

luckysori (approved):
I was wrong!

daemon/src/position_metrics.rs
bonomat: You are welcome. Is there a better name for `all` which would make this easier to understand?

luckysori: Perhaps `any`? 

This value is for the `POSITION_LABEL` which I think indicates the type of position associated with a CFD. The values `long` and `short` are self-explanatory, but if we want to express that it could be either one `any` might be clearer.

It's almost like a wildcard or the absence of the label itself.

PR #2330, opened 2022-06-30
Add margin metric

bonomat:
This metric shows total margin locked up by party and counterparty for open positions only.

Looks like this

```
# HELP margin_counterparty_quantities Total position margin of our counterparties on ItchySats.
# TYPE margin_counterparty_quantities gauge
margin_counterparty_quantities 246994
# HELP margin_quantities Total position margin on ItchySats.
# TYPE margin_quantities gauge
margin_quantities 493987
```

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -1,15 +1,19 @@
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
+use bdk::bitcoin::Amount;
 use futures::StreamExt;
+use model::calculate_margin;
 use model::CfdEvent;
 use model::ClosedCfd;
 use model::EventKind;
 use model::FailedCfd;
 use model::FailedKind;
 use model::Identity;
+use model::Leverage;
 use model::OrderId;
 use model::Position;
+use model::Role;
 use model::Settlement;
 use model::Usd;
 use rust_decimal::Decimal;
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
should do the job!

It's unfortunate that we have to calculate margin in all of these places, and copy-paste that code around, but there's not much we can do about it until we store the margin in the db (which you've created a ticket for already).

daemon/src/position_metrics.rs
```diff
@@ -343,6 +389,24 @@ mod metrics {
             .unwrap()
         });
 
+    static MARGIN_QUANTITY_GAUGE: conquer_once::Lazy<prometheus::IntGauge> =
+        conquer_once::Lazy::new(|| {
+            prometheus::register_int_gauge!(
+                "margin_quantities",
```
luckysori: 🙃 I don't remember the naming conventions around metrics, but what about `margin` and `margin_counterparties` (or `margin_quantity` and `margin_counterparties_quantity` if the suffix is needed).

luckysori (approved):
LGTM!

daemon/src/position_metrics.rs
bonomat: Thanks for pointint this out. The convention is that the last word should indicate the type, e.e. in this case it should be `margin_satoshis`. 

I will fix this.

luckysori: Ah cool, that makes sense!

PR #2333, opened 2022-06-30
Update to xtra upstream 

Restioson:
Continuation of #2306, but rebased for cleaner history.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update `xtra` to [upstream](https://github.com/Restioson/xtra). This involved re-implementing some of the features
+  in [comit-network's fork](https://github.com/comit-network/xtra) internally. Xtra message handler metrics were also
+  removed in favour of the new `instrumentation` feature combined with
+  [Grafana Tempo's span metrics](https://grafana.com/docs/tempo/latest/server_side_metrics/span_metrics/).
+
 ## [0.4.21] - 2022-06-27
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,10 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
+xtra = { git = "https://github.com/Restioson/xtra", rev = "afff02dd0fc8b92ae264db8a5457773f8af95487" } # Unreleased
 maia = { git = "https://github.com/comit-network/maia", rev = "fc6b78b98407b10b55f8cfd152062ad77f98cd9f" } # Unreleased
 maia-core = { git = "https://github.com/comit-network/maia", tag = "0.1.1", package = "maia-core" } # Pinned to support maia 0.1 and 0.2
-xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity" } # Unreleased
+xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity", rev = "bebe45425ae44980186df7b96b41f70cad58a4bb" } # Unreleased
 
 [profile.dev.package.sqlx-macros]
 opt-level = 3
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -50,7 +50,7 @@ tokio-util = { version = "0.7", features = ["codec"] }
 tracing = { version = "0.1" }
 uuid = { version = "1.1", features = ["serde", "v4"] }
 x25519-dalek = { version = "1.1" }
-xtra = { version = "0.6", features = ["metrics", "timeout", "tracing"] }
+xtra = { version = "0.6", features = ["instrumentation"] }
 xtra-bitmex-price-feed = { path = "../xtra-bitmex-price-feed" }
 xtra-libp2p = { path = "../xtra-libp2p" }
 xtra-libp2p-offer = { path = "../xtra-libp2p-offer" }
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -1,5 +1,6 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 
+use crate::bitcoin::util::psbt::PartiallySignedTransaction;
 use crate::bitcoin::Txid;
 use anyhow::Context as _;
 use anyhow::Result;
@@ -36,6 +37,7 @@ use xtra_libp2p_ping::pong;
 use xtras::supervisor;
 use xtras::supervisor::always_restart;
 use xtras::supervisor::always_restart_after;
+use xtras::HandlerTimeoutExt;
 
 pub use bdk;
 pub use maia;
@@ -112,13 +114,18 @@ pub struct TakerActorSystem<O, W, P> {
 
 impl<O, W, P> TakerActorSystem<O, W, P>
 where
-    O: Handler<oracle::MonitorAttestation> + Handler<oracle::GetAnnouncement> + Actor<Stop = ()>,
-    W: Handler<wallet::BuildPartyParams>
-        + Handler<wallet::Sign>
-        + Handler<wallet::Withdraw>
-        + Handler<wallet::Sync>
+    O: Handler<oracle::MonitorAttestation, Return = ()>
+        + Handler<
+            oracle::GetAnnouncement,
+            Return = Result<olivia::Announcement, oracle::NoAnnouncement>,
+        > + Actor<Stop = ()>,
+    W: Handler<wallet::BuildPartyParams, Return = Result<maia_core::PartyParams>>
+        + Handler<wallet::Sign, Return = Result<PartiallySignedTransaction>>
+        + Handler<wallet::Withdraw, Return = Result<Txid>>
+        + Handler<wallet::Sync, Return = ()>
         + Actor<Stop = ()>,
-    P: Handler<xtra_bitmex_price_feed::LatestQuote> + Actor<Stop = xtra_bitmex_price_feed::Error>,
+    P: Handler<xtra_bitmex_price_feed::LatestQuote, Return = Option<xtra_bitmex_price_feed::Quote>>
+        + Actor<Stop = xtra_bitmex_price_feed::Error>,
 {
     #[allow(clippy::too_many_arguments)]
     pub fn new<M>(
@@ -137,11 +144,11 @@ where
         environment: Environment,
     ) -> Result<Self>
     where
-        M: Handler<monitor::StartMonitoring>
-            + Handler<monitor::Sync>
-            + Handler<monitor::MonitorCollaborativeSettlement>
-            + Handler<monitor::MonitorCetFinality>
-            + Handler<monitor::TryBroadcastTransaction>
+        M: Handler<monitor::StartMonitoring, Return = ()>
+            + Handler<monitor::Sync, Return = ()>
+            + Handler<monitor::MonitorCollaborativeSettlement, Return = ()>
+            + Handler<monitor::MonitorCetFinality, Return = Result<()>>
+            + Handler<monitor::TryBroadcastTransaction, Return = Result<()>>
             + Actor<Stop = ()>,
     {
         let (maker_online_status_feed_sender, maker_online_status_feed_receiver) =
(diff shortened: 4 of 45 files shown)


klochowicz (commented):
would you mind adding a bit more info in `Update xtra to upstream`, ie. list notable changes in xtra into which we're adapting?

Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
+xtra = { git = "https://github.com/Restioson/xtra", rev = "fc3dca2b14653095659ed8aff067150854aed1c2" } # Unreleased
```
klochowicz: are you planning to merge this xtra branch to `master` before merging this PR? I am ok with pinning rev inside the `master` branch, but not that keen on depending on some side branch.

(if yes, please add a TODO about that)

daemon/Cargo.toml
```diff
@@ -50,7 +50,7 @@ tokio-util = { version = "0.7", features = ["codec"] }
 tracing = { version = "0.1" }
 uuid = { version = "1.1", features = ["serde", "v4"] }
 x25519-dalek = { version = "1.1" }
-xtra = { version = "0.6", features = ["metrics", "timeout", "tracing"] }
```
klochowicz, line 53: I guess it's really important to mention in the PR description and in the commit message that:
- we're removing xtra metrics (and that the new tracing instrumentation framework will be ready before the new release)
- timeouts are supported inside `xtra` out-of-the-box (new implementation)
- `tracing` got renamed to `instrumentation` 

otherwise, one might think we're losing a lot of features :)

daemon/src/lib.rs
```diff
@@ -26,6 +26,7 @@ use time::ext::NumericalDuration;
 use tokio::sync::watch;
 use tokio_tasks::Tasks;
 use xtra::prelude::*;
+use xtra_bitmex_price_feed::Quote;
```
klochowicz: nit: we typically qualify this one at the callsite, just like other structs from this module

daemon/src/projection.rs
```diff
@@ -75,7 +75,8 @@ pub struct Actor {
     db: sqlite_db::Connection,
     tx: Tx,
     state: State,
-    price_feed: Box<dyn MessageChannel<xtra_bitmex_price_feed::LatestQuote>>,
+    price_feed:
+        MessageChannel<xtra_bitmex_price_feed::LatestQuote, Option<xtra_bitmex_price_feed::Quote>>,
```
klochowicz, line 79: as a user: this is a bit more verbose, but I actually don't mind it as I see what's being returned from that handler!

Restioson:
Sure!

xtra-libp2p/src/dialer.rs
```diff
@@ -47,9 +47,9 @@ impl Actor {
     }
 
     fn stop_with_error(&mut self, e: Error, ctx: &mut xtra::Context<Self>) {
-        tracing::debug!("Stopping dialer with an error: {e:#}");
+        tracing::debug!("Stopping dialer with an error: {e:?}");
```
klochowicz: did you mean to change this line? it probably doesn't belong to this changeset :)

xtra-libp2p-offer/Cargo.toml
```diff
@@ -14,7 +14,7 @@ thiserror = "1"
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "net"] }
 tokio-tasks = { path = "../tokio-tasks" }
 tracing = "0.1"
-xtra = { version = "0.6", features = ["metrics", "timeout", "tracing"] }
+xtra = { version = "0.6" }
```
klochowicz, line 17: no need to add "instrumentation" to this one?

xtras/src/handler_timeout.rs
```diff
@@ -0,0 +1,92 @@
+use std::ops::ControlFlow;
+use std::time::Duration;
+use tokio::time::timeout;
+use xtra::Actor;
+use xtra::Context;
+
+pub trait HandlerTimeoutExt<A> {
```
klochowicz, line 9: would you mind adding a bit of docs to this trait and/or method?

maker/src/connection.rs
```diff
@@ -94,16 +94,18 @@ pub struct TakerMessage {
     pub msg: wire::MakerToTaker,
 }
 
+pub struct TakerMessageIgnoreErr(pub TakerMessage);
```
klochowicz: Please add a comment describing the reason for wrapping this type

klochowicz (commented):
looks pretty good, I've added some comments, mostly to aid in readability :)

daemon/src/projection.rs
Restioson: It is a bit more verbose in these kind of places, yea. That is the downside. For historians, the PR that removed the Message trait is https://github.com/Restioson/xtra/pull/62

xtra-libp2p/src/dialer.rs
Restioson: Thanks, this was by accident

xtra-libp2p-offer/Cargo.toml
Restioson: `instrumentation` doesn't affect the public API, so only binaries really care about it being enabled I think.

maker/src/connection.rs
Restioson: Ah, I remember that this is something that we discussed - the alternative to having this type and having its own handler would be to replace the sink somewhere else with just an address send which intentionally swallows the error. The reason for this now is that MessageSink only exists for `Handler<Return = ()>`, to avoid swallowing errors or return values accidentally.

klochowicz: if this is preserving the old behaviour, and just forcing us to be explicit about dropping the errors, this is also fine!

klochowicz:
I'm afraid you might need to rebase on master (there's lots of conflicts, probably after I merged a huge dependency upgrade recently)....

klochowicz:
as for the supervisor refactor - I recommend leaving that for a separate PR :) then supervisor and adding instrumentation could be looked at independently

Restioson:
> I'm afraid you might need to rebase on master (there's lots of conflicts, probably after I merged a huge dependency upgrade recently)....

Will do!

> as for the supervisor refactor - I recommend leaving that for a separate PR :) then supervisor and adding instrumentation could be looked at independently

I've pushed it to the `supervisor-refactor` on my fork for now, so it's there for when the time comes

Cargo.toml
Restioson: This could get merged https://github.com/Restioson/xtra/pull/112 does, if I remove instrumentation from the timeout handler - the impact of this would be that the "Handler execution timed out" log wouldn't be attached to the handler span. Otherwise, it should wait for that to be merged and I can add a TODO. Maybe this is a small price to pay for getting this merged earlier though? We can always change it back to properly attach the log to the span when #112 does get merged - it should only be a few lines change. Let me know which is best!

maker/src/connection.rs
Restioson: The other change with sinks was about backpressure - now, it will wait for the messages it sends in to be handed before it sends the next.

Restioson: I'll reimplement this with `sink::unfold` to be equivalent to the old version - throwing away errors and no backpressure. It should be pretty easy to change to logging errors or stopping the sink on errors with an `unfold`-based sink

Restioson:
Could someone enable CI workflows please? :slightly_smiling_face:

daemon/Cargo.toml
bonomat: Can you update changelog.md as well please?

Restioson: Will do!

Restioson: Should I mention just the fact that xtra was updated, or also the internal work such as replacing handler timeouts with a solution in `xtras`?

Restioson: Pushed a minimal update, let me know if I should add more detail to it! My plan is probably to squash the commit into "Update xtra to upstream" via interactive rebase before this is merged.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Update `xtra` to [upstream](https://github.com/Restioson/xtra). This involved re-implementing some of the features
```
bonomat, line 12: Maybe add that with this patch the metrics from xtra are being removed.

bonomat:
> Could someone enable CI workflows please? 🙂

I've added you as a contributor. Hopefully this will help.

Restioson:
Thank you!

Cargo.toml
klochowicz: I would prefer merging this earlier so that it can run on our testnet over the weekend - as you correctly noticed, that span addition can always be added later :)

Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/comit-network/xtra" } # We need to use unreleased patches.
+xtra = { git = "https://github.com/Restioson/xtra", rev = "8b3f5e7c3307a5b97a84401a3189247017362d37" } # Unreleased
```
klochowicz: as talked earlier, this not being from `master` is the only blocker for merging this PR.

maker/src/contract_setup.rs
```diff
@@ -91,19 +92,26 @@ impl Actor {
...
-                })
+            sink::unfold((), move |_, msg| {
+                let taker = taker.clone();
+                async move {
+                    let msg = connection::TakerMessage {
+                        taker_id,
+                        msg: wire::MakerToTaker::Protocol { order_id, msg },
+                    };
+
+                    let _ = taker.send(msg).split_receiver().await;
```
klochowicz, line 106: if we can easily log the error here, I'd opt for that (as it is an improvement :) ). Perhaps something like:

```suggestion
                    if let Err(e) = taker.send(msg).split_receiver().await {
                        tracing::error!("failed to send contract_setup message to the taker: {e:#});
                    } 
```

(in all instances where you introduced unfolding)

maker/src/rollover.rs
```diff
@@ -216,13 +217,20 @@ impl Actor {
...
-                })
+            sink::unfold((), move |_, msg| {
+                let send_to_taker_actor = send_to_taker_actor.clone();
+                async move {
+                    let msg = connection::TakerMessage {
+                        taker_id,
+                        msg: wire::MakerToTaker::RolloverProtocol { order_id, msg },
+                    };
+
+                    let _ = send_to_taker_actor.send(msg).split_receiver().await;
```
klochowicz, line 231: same suggestion as above, instead of dropping the error, we can at least log it.

xtras/src/handler_timeout.rs
```diff
@@ -0,0 +1,99 @@
...
+    impl Handler<()> for MyActor {
+        type Return = ();
+
+        async fn handle(&mut self, _: (), _ctx: &mut Context<Self>) {
+            assert!(self.started);
+        }
+    }
+
+    #[tokio::test]
+    async fn started_is_called() {
```
klochowicz, line 87: nice to see this covered by a test 😀

klochowicz (commented):
Thanks for updates, I really like usage of `unfold` instead of a separate message channel! As stated above, the span for handler timeout can be done as a follow-up when `xtra` has these changes in master.

maker/src/contract_setup.rs
Restioson: Can do!

xtras/src/handler_timeout.rs
Restioson: Haha, I felt it was needed after that!

maker/src/contract_setup.rs
Restioson: Ah, we cannot actually do this currently without https://github.com/Restioson/xtra/pull/105, unless we first wait for the message to be handled and then print the result. This has a different backpressure behaviour. On the other hand, it could still log an error if the taker is disconnected right after sending the message, or right before.

Restioson: Additionally, should the sink stop accepting messages if an error occurs? I think this can be done by returning `Err` from `unfold`

klochowicz: interesting idea, although I would probably not go that far yet, as this will indeed change the behaviour. I would like to first get some stats on how often we see errors here.

klochowicz: > Ah, we cannot actually do this currently without [Restioson/xtra#105](https://github.com/Restioson/xtra/pull/105), unless we first wait for the message to be handled and then print the result. This has a different backpressure behaviour. On the other hand, it could still log an error if the taker is disconnected right after sending the message, or right before.

Ok, no worries! we can also print the messages later. happy for it to go without printing these logs :)

klochowicz (approved):
looks good!
please squash the small fixup commits in the first one, and it's ready to merge :)

klochowicz:
bros r+

klochowicz:
@Restioson could you please rebase on top of latest master and force-push? (I tried to do a merge in github, which fixed the error, but formatting failed). 

Amending the merge commit would work too (by adding the newline).

klochowicz:
@Restioson could you please rebase on top of latest master and force-push? (I tried to do a merge in github, which fixed the error, but formatting failed). 

Amending the merge commit would work too (by adding the newline).

PR #2343, opened 2022-07-01
Basic instrumentation and export traces via OTLP

Restioson:
This PR exports traces to jaeger, as well as adding a few spans and `#[instrument]` macro invocations.

## TODO
- [x] Instrument a few functions
- [x] Convert to use OTEL exporter
  - [x] Should jaeger support be retained in some fashion for local debugging?
- [x] Command-line flag to disable instrumentation
- [x] ~~Instrument tests (?)~~ moved to followup


klochowicz:
@Restioson I guess it can be using the recent updates in xtra-productivity (with renamed feature name)!

Restioson:
@klochowicz and I were observing memory leaks on this branch, but after rebasing to the latest master, it seems like this fixed the issue. This could be due to the high CPU usage bug being fixed, as when the memory usage was climbing, the CPU usage was pinned to 8% relative to a normal 0-2%. I left it running for many hours with no issues (previously, a max of ~30min was needed to observe the issue), so I think it's fixed now :)

For now, just the instrumentation should be reviewed, I think - this needs to be converted to use the OTEL exporter, which I am not totally sure how to test. It might also be useful to retain jaeger functionality for local testing, perhaps behind a feature flag?

tokio-tasks/src/lib.rs
```diff
@@ -1,6 +1,72 @@
...
+pub use future_ext::*;
 pub use task_map::*;
 pub use tasks::*;
 
 mod future_ext;
 mod task_map;
 mod tasks;
+
+#[tracing::instrument(name = "Sleep")]
+pub async fn sleep(duration: Duration) {
```
Restioson: I was not really sure which crate or module these functions should go in

klochowicz: I guess we can keep them here. I will probably rename the crate to tokio-helpers or something like that. I will do this as a follow-up so we can run the name change past everyone.

Restioson: Sure. For now should I move them into a separate module, though? E.g: tokio_tasks::time

tokio-tasks/src/lib.rs
```diff
@@ -1,6 +1,72 @@
...
+where
+    F: Future,
+{
+    #[allow(clippy::disallowed_methods)]
+    Timeout {
+        fut: tokio::time::timeout(duration, fut),
+        span: tracing::debug_span!(
+            "Future with timeout",
+            timeout_secs = duration.as_secs(),
+            timed_out = field::Empty
```
klochowicz: this flag is really cool, I like it! we can quickly gather stats on timed out handlers 🦈

klochowicz:
as for the remaining TODOs:
> Instrument more functions (which?)
I believe this is a great start to showcase the functionality. I guess we'll want to add the HTTP endpoints via a Rocket fairing, but this can be done in parallel :) 

>  Should jaeger support be retained in some fashion for local debugging?

I would keep a side branch with jaeger just in case we won't manage to make it work with OTEL, but from some websites I've read today it seems that we should be able to use OpenTelemetry collector with jaeger directly; it might not be as convenient as jaeger-all-in-one, but a simple docker-compose file for debugging should do the trick!

> Instrument tests (?)
I would keep this as a separate, follow-up task. There are a few open questions, e.g;
- how can we connect traces using our protocols? (some wrappers for protocols that send TraceIds?) or just add that field to the messages? a POC showing feasability and trade-offs would be nice then. 
- can we upload CI results to our grafana? (including defining some upper limit of storage, e.g. we probably would like to remove traces from more than 24h ago)

tokio-tasks/src/lib.rs
klochowicz: > Sure. For now should I move them into a separate module, though? E.g: tokio_tasks::time

good idea!

klochowicz (commented):
great start! I'm keen to see another look when we convert it to OpenTelemetry export instead of jaeger (jaeger should be able to ingest it as well if needed)

other TODOs can wait after it gets merged, and can be done in parallel :)

shared-bin/src/logger.rs
```diff
@@ -31,23 +36,37 @@ pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
...
+        .install_batch(opentelemetry::runtime::Tokio)
+        .unwrap();
+
+    let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+
+    Registry::default()
+        .with(telemetry)
+        .with(fmt_layer)
+        .try_init()
+        .map_err(|e| anyhow!("Failed to init logger: {e}"))?;
```
luckysori: 🔧 `anyhow` should allow you to add context more ergonomically:

```rust
use anyhow::Context;

// [...]
.try_init()
.context("Failed to init logger")?;
```

Restioson:
On the more functions side, I haven't really looked at the maker at. Is there anything worth instrumenting there? I'll reply to the other things tomorrow since they need a bit more research first

shared-bin/src/logger.rs
```diff
@@ -31,23 +36,37 @@ pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
...
                 "[year]-[month]-[day] [hour]:[minute]:[second]"
             )))
-            .try_init()
+            .boxed()
     };
 
-    result.map_err(|e| anyhow!("Failed to init logger: {e}"))?;
+    let fmt_layer = fmt_layer.with_filter(filter);
+
+    opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
```
luckysori: ❓ With this configuration do we emit logs to stdout normally too?

taker/src/routes.rs
```diff
@@ -54,6 +55,7 @@ pub struct IdentityInfo {
 }
 
 #[rocket::get("/feed")]
+#[instrument(name = "GET /feed", skip(rx))]
```
luckysori, line 58: ❓ Just to help me understand: with the changes to `logger::init` and _without_ adding any of these `instrument` attributes, what would we see on `jaeger` (and stdout).

And what would happen if we add the `instrument` attributes and do not change the way we configure `tracing`?

More generally, I'm wondering if it would make sense for these changes to be in separate commits. It's not a big deal at all, but I'd like to understand it.

tokio-tasks/src/lib.rs
luckysori: Could we rename this crate to something like `tokio-util` and modify the description then? Currently it reads:

```
description = "Control the lifetime of async tasks by leveraging Drop"
```

With some of these changes that is no longer true.

luckysori (approved):
Looks great!

I would recommend to squash all commits if we can't find any way to split this changeset into independent steps. At the moment each commit appears to be iterating towards the same goal.

klochowicz:
> Looks great!
> 
> I would recommend to squash all commits if we can't find any way to split this changeset into independent steps. At the moment each commit appears to be iterating towards the same goal.

I would probably keep the tokio instrumentation separate, because it's quite an independed addition and the bulk of the changeset (will be easier to inspect the main tracing instrumentation).

shared-bin/src/logger.rs
```diff
@@ -1,15 +1,20 @@
...
 
 pub use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
 
 const RUST_LOG_ENV: &str = "RUST_LOG";
 
 #[allow(clippy::print_stdout)] // because the logger is only initialized at the end of this function but we want to print a warning
-pub fn init(level: LevelFilter, json_format: bool) -> Result<()> {
+pub fn init(level: LevelFilter, json_format: bool, name: &str) -> Result<()> {
```
bonomat: From the api perspective it is not clear what `name` is used for. If I understand it correctly, than it's the `jaeger service name`. Might be good to rename it.

bonomat (approved):
Do we know that this does not add too much overhead? If not, I would prefer if we could add a feature toggle, either by flag or even by commandline arg so that we can enable/disable only when needed.

Restioson:
> Do we know that this does not add too much overhead? If not, I would prefer if we could add a feature toggle, either by flag or even by commandline arg so that we can enable/disable only when needed.

I've noticed a couple mb of memory overhead - I'm not really sure about CPU overhead. I think an `Opts` flag could disable this by an if-statement, maybe.

Edit: done :)

shared-bin/src/logger.rs
Restioson: Yes, we still do through the `fmt_layer`

taker/src/routes.rs
Restioson: > :question: Just to help me understand: with the changes to logger::init and without adding any of these instrument attributes, what would we see on jaeger (and stdout).

We would see xtra spans, and spans from the automatically-instrumented `#[xtra_productivity]` handlers. They look like this, when combined:
image

Here is an example of an xtra handler which invokes another xtra handler:
image

Here is how the first handler looks with some spans sprinkled in:
image

So, we get to see which messages are sent, how long they wait, and how long it takes to handle them, as well as any logs that are attached, but not the subtasks executed inside of the handlers.

tokio-tasks/src/lib.rs
Restioson: `tokio-util` is an already existing crate - could `tokio-extras` or similar be used instead?

shared-bin/src/logger.rs
Restioson: Good point

tokio-tasks/src/lib.rs
klochowicz: `tokio-extras` sounds good!

taker/src/routes.rs
klochowicz: for completeness, so `stdout` remains exactly as it was?

shared-bin/src/logger.rs
klochowicz: `tracing_service_name` perhaps?

taker/src/routes.rs
Restioson: Actually, not quite. Logs will now print with their span in front of the message. Spans also log returns and errors when `ret` and `err` are passed to `#[instrument]`

Example of span in front of log:

```
2022-07-05 09:15:53 DEBUG Handle NewConnection:Handle endpoint :: ConnectionEstablished: daemon::online_status: Adding newly established connection to online_status: PeerId("12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1")
```

Example of return being logged:
```
2022-07-05 09:17:19  INFO Handle Dial:dial:is_connection_established: xtra_libp2p::dialer: return=true self=Actor { endpoint: Address(Sender<xtra_libp2p::endpoint::Endpoint> { rx_count: 2, tx_count: 9, rc: TxStrong(()) }), connect_address: "/ip4/35.189.57.114/tcp/10000/p2p/12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1", listener_peer_id: Some(PeerId("12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1")), stop_reason: None } self=Actor { endpoint: Address(Sender<xtra_libp2p::endpoint::Endpoint> { rx_count: 2, tx_count: 10, rc: TxStrong(()) }), connect_address: "/ip4/35.189.57.114/tcp/10000/p2p/12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1", listener_peer_id: Some(PeerId("12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1")), stop_reason: None }
```

This is quite a verbose one! It could be trimmed down by changing the debug implementation for `dialer::Actor`. We could also probably change the log level here if desired.

Here is how a similar example (but with the return of `is_connection_established` being `false`) is rendered in Jaeger:

image

So, these tags do clutter the logs themselves a bit, but they are more designed to be viewed in a table format like in Jaeger rather than all on one line. I think that this format can be customised, though - see [`tracing_subscriber::fmt::Pretty`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/format/struct.Pretty.html)

Restioson:
I've rebased this on master, as well as squashed this into two commits: one to add OLTP exporting, and one to add instrumentation basics.

jaeger.sh
```diff
@@ -0,0 +1,6 @@
+docker run \
```
klochowicz, line 2: maybe add a comment what's the purpose of this script on the first line? 
like `# run jaeger for viewing traces when debugging locally` (or something like that?)

taker/src/routes.rs
klochowicz: OK, thanks for the info. We can adjust the logs later if it becomes a nuisance, no need to hold that PR.

rocket-basicauth/src/lib.rs
```diff
@@ -176,6 +176,7 @@ mod tests {
     }
 
     #[rocket::get("/protected")]
+    #[tracing::instrument(name = "GET /protected")]
```
klochowicz: I would probably skip this one rather than adding a dependency on tracing (this one makes the CI fail :/)

klochowicz (commented):
concept ACK

looks good to me, just some last nits and making the CI pass left :) 

thanks for adding the `jaeger.sh` example for local debugging!

klochowicz (approved):
LGTM!

taker/src/routes.rs
luckysori: Awesome, thanks for the detailed responses :D

PR #2344, opened 2022-07-03
Use patched electrum-client to fix a bug with high CPU load

klochowicz:
BDK 0.19 release includes the change that ignores empty lines from electrum,
which was known to cause issues on testnet. Revert this in order to reduce CPU
load until BDK merges a fix.


bonomat (approved):
Thanks for the fix.

PR #2345, opened 2022-07-04
Add db-failsafe mode

bonomat:
This will prevent the maker from starting up if the database migration was unsuccessful

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   removed in favour of the new `instrumentation` feature combined with
   [Grafana Tempo's span metrics](https://grafana.com/docs/tempo/latest/server_side_metrics/span_metrics/).
 
+### Added
+
+- Add new argument to the maker: `ignore-migration-errors`. If enabled, the maker will start if an error occurred when opening the database, if not, it will fail fast. This can come handy to prevent accidentally creating a new empty database in case database migration was unsuccessful.
+
 ## [0.4.21] - 2022-06-27
 
 ### Added
--- a/maker/src/lib.rs
+++ b/maker/src/lib.rs
@@ -38,6 +38,10 @@ pub struct Opts {
     #[clap(short, long)]
     pub json: bool,
 
+    /// If enabled the application will not fail if an error occurred during db migration.
+    #[clap(short, long)]
+    pub ignore_migration_errors: bool,
+
     /// If provided will be used for internal wallet instead of a random key. The keys will be
     /// derived according to Bip84
     #[clap(short, long)]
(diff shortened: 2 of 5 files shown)


maker/src/lib.rs
```diff
@@ -38,6 +38,11 @@ pub struct Opts {
     #[clap(short, long)]
     pub json: bool,
 
+    /// If enabled the application will fail fast and terminate if an error occurred during db
+    /// migration.
+    #[clap(short, long)]
+    pub db_failsafe: bool,
```
luckysori: 🔧 I think we should default to a value of `true` so that we can always safely omit this argument.

luckysori (approved):

maker/src/lib.rs
bonomat: I thought about it but then we would need to call it `no-db-failsafe` as you can't set a default for booleans.

luckysori: Fair. What about the flag `--ignore-migration-errors`? It sounds pretty natural.

bonomat: can do :)

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   removed in favour of the new `instrumentation` feature combined with
   [Grafana Tempo's span metrics](https://grafana.com/docs/tempo/latest/server_side_metrics/span_metrics/).
 
+### Added
+
+- Add new argument to the maker: `db-failsafe`. If enabled, the maker will not start if an error occurred when opening the database. This can come handy to prevent accidentally creating a new empty database in case database migration was unsuccessful.
```
klochowicz: Alternative suggestion for the changelog entry (name got changed, and IMHO it's important to state that this behaviour is the default on the maker now. 

```suggestion
- Do not start the maker an error occurred when opening the database. This can come handy to prevent accidentally creating a new empty database in case database migration was unsuccessful. If needed, this behaviour can be bypassed by passing an additional argument to the maker: `ignore-migration-errors`.
```

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   removed in favour of the new `instrumentation` feature combined with
   [Grafana Tempo's span metrics](https://grafana.com/docs/tempo/latest/server_side_metrics/span_metrics/).
 
+### Added
```
klochowicz, line 17: I guess the commit message should also be updated (can paste the suggested changelog entry there maybe :) )

klochowicz (approved):
cool! just some nits that should get updated after changing the flag to `ignore_migration_errors`

klochowicz (approved):

PR #2353, opened 2022-07-04
Reduce ping interval to 30s

klochowicz:
With the endpoint timeout interval set to 20s, initiating ping protocol
unconditionally every 5s can lead to messages in the endpoint piling up in case
endpoint encounters problems.

Note: Ideally, we should not schedule a new ping if the previous one was not
processed yet.

It might be worthwhile to backport this change on top of 0.4.21 release on the maker and release a small patch release hotfix, in hope that this will alleviate the issue with not sending offers.


Restioson:
> Note: Ideally, we should not schedule a new ping if the previous one was not processed yet.

The way to do this in xtra would be `send(Ping).await`, without using `do_send_async` or `split_receiver`, as this waits for the message to be handled before it returns, though I am not sure if this would be as simple of a change as this in this context (maybe something else relies on this returning early)

bonomat (approved):
Let's give it a try

bonomat:
bors r+

Merging this to test effects on testnet

Issue #2355, opened 2022-07-05
IntroJs is broken


bonomat:
On latest master (58ed6e9d1f59e55478c252dbd1dbb2e5c5c50eac) the intro js is not shown. Since it is not shown, it cannot be cancelled and the app cannot be used as all interaction with the page are disabled. 

Firefox 102.0 (64-bit):
<img width="1504" alt="image" src="https://user-images.githubusercontent.com/224613/177235872-9a8b196b-0a79-4161-9c8b-d10082be878f.png">

Chrome Version 103.0.5060.53 (Official Build) (x86_64): 
<img width="1504" alt="image" src="https://user-images.githubusercontent.com/224613/177235905-3a9a0cee-cc6b-432d-af0c-fa74a66439f7.png">

To reproduce: 
1. checkout latest master, 
3. run `yarn dev` or `yarn build` in `taker-ui`
4. start taker
5. make sure to clear browser storage/cookies

bonomat:
resolved with https://github.com/itchysats/itchysats/pull/2356

PR #2356, opened 2022-07-05
Remove introjs

bonomat:
If the intro is not shown, there is no way to use the application as there is an overlay which disables all interaction. 

I spend half a day debugging #2355 and can't figure it out.  I see no other option than removing it.


klochowicz (approved):
makes sense, at least we can continue moving forward until we fix the underlying issue.

Let's hope bors will pass as the taker-frontend job got stuck on CI...

bonomat:
bors cancel

bonomat:
Fixed a redundant import. Tests should pass now.

bonomat:
looks good. 

bors r+

PR #2358, opened 2022-07-05
Allow starting in regtest mode

### LINKED ISSUE #2313: Add `regtest` support
Right now, we do not support this network but it would be nice if we could start itchysats in `regtest` mode. 

https://github.com/itchysats/itchysats/blob/8c8d12624e80d0d2041b6ed3531c6ed4a46c5f73/taker/src/main.rs#L164-L192

bonomat:
Resolves #2313

Next I'd make the oracle configurable. Thatway we can have an offline-dev setup :) 

Tested using https://github.com/vulpemventures/nigiri

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -118,7 +118,9 @@ impl Opts {
             None => match network {
                 Network::Mainnet { .. } => MAINNET_MAKER.to_string(),
                 Network::Testnet { .. } => TESTNET_MAKER.to_string(),
-                Network::Signet { .. } => bail!("No maker default URL configured for signet"),
+                Network::Signet { .. } | Network::Regtest { .. } => {
+                    bail!("No maker default URL configured for {network:?}")
+                }
             },
         };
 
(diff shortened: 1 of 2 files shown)


taker/src/main.rs
```diff
@@ -118,7 +118,9 @@ impl Opts {
             None => match network {
                 Network::Mainnet { .. } => MAINNET_MAKER.to_string(),
                 Network::Testnet { .. } => TESTNET_MAKER.to_string(),
-                Network::Signet { .. } => bail!("No maker default URL configured for signet"),
+                Network::Signet { .. } | Network::Regtest { .. } => {
+                    bail!("No maker default URL configured for signet")
+                }
```
luckysori: 🔧

```suggestion
                Network::Signet { .. } | Network::Regtest { .. } => {
                    bail!("No maker default URL configured for {network:?}")
                }
```

And for the other cases too.

luckysori (approved):

PR #2360, opened 2022-07-05
Remove xtra_productivity `message_impl = false` argument

klochowicz:
We don't add message impls anymore, so there's no need to specify this as
`false`.


Restioson:
This was on my to-do list, but I forgot about it, thanks!

Restioson (approved):

bonomat (approved):

luckysori (approved):
Yes!

PR #2361, opened 2022-07-05
Use xtra::scoped where possible

Restioson:
This removes a substantial number of Tasks structs in favour of `xtra::scoped`, which is a bit cleaner for when the tasks are tied to the actor's lifecycle. 

cc @klochowicz, this might be a little less ambitious than what you had in your branch, which is maybe why it works? ~~There are some actors (Maker and Taker) which still have actor-lifetime tasks structs, which I haven't touched yet since they are involved in `new`, when the context is not yet available. Could some of these be moved to `Actor::started`, perhaps?~~ edit: looks like I just assumed these are actors, but they aren't :sweat_smile:

### DIFF
--- a/tokio-extras/src/tasks/actor_scoped.rs
+++ b/tokio-extras/src/tasks/actor_scoped.rs
@@ -1,0 +1,35 @@
+use futures::FutureExt;
+use std::future::Future;
+use xtra::refcount::RefCounter;
+use xtra::Address;
+
+/// Spawn a task that is scoped to the lifetime of the given address.
+pub fn spawn<A, Rc, F>(addr: &Address<A, Rc>, fut: F)
+where
+    Rc: RefCounter,
+    F: Future + Send + 'static,
+{
+    #[allow(clippy::disallowed_methods)]
+    tokio::spawn(xtra::scoped(addr, fut.map(|_| ())));
+}
+
+/// Spawn a fallible task that is scoped to the lifetime of the given address.
+pub fn spawn_fallible<A, Rc, Task, Ok, Err, Fn, FnFut>(
+    addr: &Address<A, Rc>,
+    fut: Task,
+    handle_err: Fn,
+) where
+    Rc: RefCounter,
+    Task: Future<Output = Result<Ok, Err>> + Send + 'static,
+    Fn: FnOnce(Err) -> FnFut + Send + 'static,
+    FnFut: Future + Send,
+    Ok: Send,
+    Err: Send,
+{
+    #[allow(clippy::disallowed_methods)]
+    tokio::spawn(xtra::scoped(addr, async {
+        if let Err(e) = fut.await {
+            handle_err(e).await;
+        }
+    }));
+}
--- a/xtra-libp2p-ping/src/pong.rs
+++ b/xtra-libp2p-ping/src/pong.rs
@@ -1,24 +1,26 @@
 use crate::protocol;
 use async_trait::async_trait;
-use tokio_extras::Tasks;
+use xtra::Context;
 use xtra_libp2p::NewInboundSubstream;
 use xtra_productivity::xtra_productivity;
 
-#[derive(Default)]
-pub struct Actor {
-    tasks: Tasks,
-}
+#[derive(Copy, Clone)]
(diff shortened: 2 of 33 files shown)


Cargo.lock
```diff
@@ -805,7 +805,7 @@ dependencies = [
  "time",
```
klochowicz: did you mean to downgrade all the dependencies from this file? will it still work if  you revert `Cargo.lock` changes done here?

tokio-extras/src/tasks/actor_scoped.rs
```diff
@@ -0,0 +1,35 @@
+use futures::FutureExt;
```
klochowicz, line 1: maybe `xtras` would be a more natural place for this module - what do you think? (then we can still keep xtra an optional dependency here).

Restioson:
I'll rebase to remove these fixup commits when this has been reviewed, if that's okay?

klochowicz:
> I'll rebase to remove these fixup commits when this has been reviewed, if that's okay?

perfect; adding `fixup!` was actually great, as I am in the middle of the review :)

Restioson:
Haha, it finally clicked for me that this is a better approach than just rebasing locally every time (especially after I noticed that it can make github comment threads disconnect)

xtra-libp2p-ping/src/pong.rs
```diff
@@ -1,24 +1,26 @@
...
-use tokio_extras::Tasks;
+use xtra::Context;
 use xtra_libp2p::NewInboundSubstream;
 use xtra_productivity::xtra_productivity;
 
-#[derive(Default)]
-pub struct Actor {
-    tasks: Tasks,
-}
+#[derive(Default, Copy, Clone)]
```
klochowicz: why does this one need `Copy` and `Clone`? not sure if it even needs `Default` anymore, after removing Tasks struct!

xtra-libp2p/examples/hello_world_listener.rs
```diff
@@ -77,20 +76,21 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
-#[derive(Default)]
-pub struct HelloWorld {
-    tasks: Tasks,
-}
+#[derive(Default, Copy, Clone)]
```
klochowicz: same here, maybe we can get rid of all derives?

klochowicz (commented):
concept ACK, just some pending questions.

xtra-libp2p-ping/src/pong.rs
Restioson: `-Wmissing-copy-implementations` is enabled, so that is why I added it. I could remove it and add an `#[allow]` though

tokio-extras/src/tasks/actor_scoped.rs
Restioson: Perhaps, although this is specifically working with `tokio` (that's where it spawns), so that's why I went with `tokio_extras` initially. I kept xtra as an optional dependency in a followup fixup

Cargo.lock
Restioson: Oops, this was a git stash mistake, thanks!

xtra-libp2p-ping/src/pong.rs
Restioson: I've done this now

klochowicz: ok then, let's reinstate the `Copy,Clone` then - might as well! sorry for the trouble!

xtra-libp2p/examples/hello_world_listener.rs
klochowicz: same here - let's actually keep this without the Default (if it's mandated by clippy).

klochowicz (approved):
LGTM!

xtra-libp2p/examples/hello_world_listener.rs
Restioson: Sure

Restioson:
Rebased & force-pushed

klochowicz:
bors retry

PR #2362, opened 2022-07-05
Enable instrumentation in dev-taker and dev-maker

### LINKED ISSUE #2346: Allow exporting OpenTelemetry compatible traces

klochowicz:
In order to use instrumentation, it has to be explicitly enabled.
For convenience, enable instrumentation by default inside both
`cargo dev-maker` and `cargo dev-taker` as it's intented to be used by devs.


klochowicz:
Merging this, as it's not contentious (and verified locally that it works).

PR #2363, opened 2022-07-05
Reduce log spam from h2 and tower

klochowicz:
After adding instrumentation, h2 and tower debug logs were spamming the logs a
bit, as they're used by Rocket and libp2p respectively.
Setting them to INFO should reduce that spam.


klochowicz:
merging as it's a trivial change.

bors r+

Restioson:
While this does work for logs, it is important to note that this still results in traces which are somewhat irrelevant, since `base_directives` is only applied to `fmt_layer`, and not `tracer` or the `Registry` as a whole. I would suggest a followup PR to apply a different set of base directives to the `tracer`, since more detail is wanted than in logs, given that traces are inherently more readable (IMO) since they can be collapsed, but still not every h2 poll or `send_data`.

bonomat:
`a bit` 😅 We produced 100gig of logs in 12h

PR #2364, opened 2022-07-05
Use xtra::scoped instead of Spawner actor

klochowicz:
Spawner actor abstraction is not necessary anymore, now that xtra supports
scoping tasks within an actor


Restioson (approved):
This is a nice simplification!

bonomat (approved):

luckysori (approved):
Nice!

PR #2366, opened 2022-07-06
Do not log all CFDs every time `GET /cfds` is called

luckysori:
This may be the reason why the size of our logs has increased so much since d89e3dac31031062a944e3ca0b337a0c765e5210.

### DIFF
--- a/maker/src/routes.rs
+++ b/maker/src/routes.rs
@@ -232,7 +232,7 @@ pub async fn put_sync_wallet(
 }
 
 #[rocket::get("/cfds")]
-#[instrument(name = "GET /cfds", skip(rx), ret, err)]
+#[instrument(name = "GET /cfds", skip(rx), err)]
 pub async fn get_cfds<'r>(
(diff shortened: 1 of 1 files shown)


bonomat (commented):
We might also want to exclude `#[rocket::get("/<_paths..>", format = "text/html")]` and  `#[rocket::get("/assets/<file..>")]`. If I understand it correctly than we would instrument the whole file content

Restioson:
> We might also want to exclude #[rocket::get("/<_paths..>", format = "text/html")] and #[rocket::get("/assets/<file..>")]. If I understand it correctly than we would instrument the whole file content

There is no `ret` here in the instrumentation so if I am not mistaken it would just log the path that it was getting and not the whole file

Restioson (approved):

klochowicz (approved):
I found logs of logs spamming testnet containers with CFD contents, so I'm pretty confident that this will fix the issue.

PR #2368, opened 2022-07-06
Reimplement supervisor

Restioson:
The old supervisor kept a `Context` around that was never received from, which meant that broadcasts would pile up until they reach the bound. This new supervisor does not do that, since the introduction of `Context::next_message` allows us to write our own manager loop and restart the actor that way.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -34,9 +34,9 @@ use xtra_libp2p::multiaddress_ext::MultiaddrExt;
 use xtra_libp2p::Endpoint;
 use xtra_libp2p_ping::ping;
 use xtra_libp2p_ping::pong;
-use xtras::supervisor;
 use xtras::supervisor::always_restart;
 use xtras::supervisor::always_restart_after;
+use xtras::supervisor::Supervisor;
 use xtras::HandlerTimeoutExt;
 
 pub use bdk;
@@ -90,17 +90,6 @@ pub struct TakerActorSystem<O, W, P> {
     pub auto_rollover_actor: Address<auto_rollover::Actor>,
     pub price_feed_actor: Address<P>,
     executor: command::Executor,
-    /// Keep this one around to avoid the supervisor being dropped due to ref-count changes on the
-    /// address.
-    _price_feed_supervisor: Address<supervisor::Actor<P, xtra_bitmex_price_feed::Error>>,
-    _collab_settlement_supervisor:
-        Address<supervisor::Actor<collab_settlement::taker::Actor, supervisor::UnitReason>>,
-    _rollover_supervisor:
-        Address<supervisor::Actor<rollover::taker::Actor, supervisor::UnitReason>>,
-    _dialer_supervisor: Address<supervisor::Actor<dialer::Actor, dialer::Error>>,
-    _offers_supervisor:
-        Address<supervisor::Actor<xtra_libp2p_offer::taker::Actor, supervisor::UnitReason>>,
-    _ping_supervisor: Address<supervisor::Actor<ping::Actor, supervisor::UnitReason>>,
     _close_cfds_actor: Address<archive_closed_cfds::Actor>,
     _archive_failed_cfds_actor: Address<archive_failed_cfds::Actor>,
     _pong_actor: Address<pong::Actor>,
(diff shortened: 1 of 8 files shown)


daemon/src/lib.rs
```diff
@@ -90,17 +91,6 @@ pub struct TakerActorSystem<O, W, P> {
     pub auto_rollover_actor: Address<auto_rollover::Actor>,
     pub price_feed_actor: Address<P>,
     executor: command::Executor,
-    /// Keep this one around to avoid the supervisor being dropped due to ref-count changes on the
-    /// address.
```
klochowicz, line 94: so the ref-count issue got fixed too? awesome! :)

maker/src/actor_system.rs
```diff
@@ -154,15 +145,15 @@ where
...
 
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
-        let (supervisor, maker_offer_address) = supervisor::Actor::new({
+        let (supervisor, maker_offer_address) = supervisor::Supervisor::new({
             let endpoint_addr = endpoint_addr.clone();
             move || xtra_libp2p_offer::maker::Actor::new(endpoint_addr.clone())
         });
-        let _maker_offer_supervisor = supervisor.create(None).spawn(&mut tasks);
+        tasks.add(supervisor.run().map(|_| ()));
```
klochowicz: It might be beneficial to log summary after supervisor exits instead of silencing these stats (here and elsewhere)

xtras/src/supervisor.rs
```diff
@@ -258,75 +206,69 @@ struct GetMetrics;
...
-        tokio::spawn(task);
+        let (supervisor, addr) =
+            Supervisor::with_policy(|| RemoteShutdown, always_restart::<io::Error>());
+        let task = supervisor.run();
 
-        let metrics = supervisor.send(GetMetrics).await.unwrap();
+        drop(addr);
         assert_eq!(
-            metrics.num_spawns, 1,
+            task.await.1.num_spawns,
```
klochowicz: nit: perhaps it would be nicer to destructure this above the assert for readability?
`let (_, metrics) = task.await`

so that the assert can pretty much stay the same?

xtras/src/supervisor.rs
```diff
@@ -258,75 +206,69 @@ struct GetMetrics;
...
+        tasks.add(async move {
+            let _ = addr.send(Shutdown).await;
+            drop(addr);
+        });
 
-        let metrics = supervisor.send(GetMetrics).await.unwrap();
         assert_eq!(
-            metrics.num_spawns, 2,
-            "after shutdown, should have 2 spawns"
+            task.await.1.num_spawns,
```
klochowicz: same here, IMHO with destructuring the tuple it would read easier!

xtras/src/supervisor.rs
```diff
@@ -356,15 +298,13 @@ mod tests {
...
         let (supervisor, address) =
-            Actor::with_policy(|| PanickingActor, always_restart::<io::Error>());
-        let (supervisor, task) = supervisor.create(None).run();
+            Supervisor::with_policy(|| PanickingActor, always_restart::<io::Error>());
+        let task = supervisor.run();
 
-        #[allow(clippy::disallowed_methods)]
-        tokio::spawn(task);
-
-        address.send(Panic).await.unwrap_err(); // Actor will be dead by the end of the function call because it panicked.
```
klochowicz, line 365: did you mean to remove comment? it still has value IMHO ;)

xtras/src/supervisor.rs
```diff
@@ -124,128 +120,80 @@ where
...
-                        let _ = this
-                            .send(Stopped {
-                                reason: reason.into(),
-                            })
-                            .await;
-                    }
-                    Err(error) => {
-                        let _ = this.send(Panicked { error }).await;
-                    }
+    pub async fn run(mut self) -> (anyhow::Error, Metrics) {
```
klochowicz, line 142: interesing way of returning the metrics, instead of relying on sending a message!

if you don't want to rely on the destructuring tuples, an alternative could be just returning a`Summary` struct (otherwise it's a bit odd to return `anyhow::Error` in a "positive" sense)
```rust
struct Summary {
     actor_stop_reason: anyhow::Error, 
     metrics: Metrics,
}
```

daemon/src/lib.rs
```diff
@@ -180,7 +170,7 @@ where
         let (endpoint_addr, endpoint_context) = Context::new(None);
 
         let (collab_settlement_supervisor, libp2p_collab_settlement_addr) =
-            supervisor::Actor::new({
+            supervisor::Supervisor::new({
```
klochowicz: I guess you might as well import `Supervisor` directly now :)

klochowicz (approved):
thanks for this cleanup!

daemon/src/lib.rs
Restioson: I am not sure exactly which issue this was referring to in the first case, to be honest, but with this refactor, a supervised actor should have the same lifecycle behaviour when all addresses are dropped as an unsupervised actor.

xtras/src/supervisor.rs
Restioson: It is no longer true that the actor will be dead by the end of the `send` call, since in this case the `send` is split off, as I rewrote the test to not have a parallel-running supervisor, but run it on the same thread.

xtras/src/supervisor.rs
Restioson: I guess the main thing with the `anyhow::Error` is what you call it - would it be okay to just destructure it as `exit` instead of `err` when taking output from `run`?

klochowicz: with the new method that hides it anyway, yes, it's OK 👍

daemon/src/lib.rs
klochowicz: fair enough; definitely a cleaner approach.

klochowicz (approved):
looks good!

Restioson:
Squashed!

klochowicz (approved):

PR #2376, opened 2022-07-06
Log timeouts inside of handler span and set interrupted

Restioson:
This sets interrupted=timed_out on handler spans that time out, which could be helpful for filtering spans by which time out, instead of having to rely on just looking for spans that take longer than others. Furthermore, the "Handler execution timed out" log message will be attached to the handler span. Xtra was updated to facilitate this.

This PR is based off of https://github.com/Restioson/xtra/pull/112/, which is yet to be merged, so for now this is a proof of concept.

### DIFF
--- a/xtras/src/handler_timeout.rs
+++ b/xtras/src/handler_timeout.rs
@@ -36,12 +36,15 @@ where
         }
 
         loop {
-            let msg = self.ctx.next_message().await;
+            let mut fut = self.ctx.tick(self.ctx.next_message().await, &mut actor);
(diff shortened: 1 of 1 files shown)


Restioson:
#112 was merged, so I just need to rebase this on master and update to xtra master, and it should be ready for review I think.

klochowicz (approved):
👍

Restioson:
@klochowicz any more reviews required or is this ready? I think this is probably a non controversial change, though

PR #2378, opened 2022-07-06
Use hourly liquidation events when constructing DLCs

### LINKED ISSUE #2299: Add liquidation events to the DLC
We want to be able to liquidate the counterparty if the price of Bitcoin falls or rises sufficiently before the settlement time. For this we need to make use of liquidation events (as opposed to the settlement event which marks the end of the CFD).

The oracle `olivia` currently attests to the price of Bitcoin every hour, so our liquidation events are currently limited by that.

On contract setup and rollover, both parties need to produce a handful of extra CETs per hour which allow to close the CFD if the oracle attests to a price near the boundaries of the payout curve.

The library `maia` already supports multiple events per DLC. Even `itchysats` already takes this into account, but we currently only make use of the settlement event.

luckysori:
Fix #2299.

The diff is misleadingly massive because when introducing the new feature in https://github.com/itchysats/itchysats/pull/2378/commits/35c4e4e658f8d72f23522a48fa6587cf12222ac5 I chose to simply duplicate the rollover actor. The old version is used to stay compatible with old takers. In my opinion, the correct approach would be to release different versions of `xtra_libp2p_rollover` as we introduce breaking changes. But this is blocked by being able to release some of the crates in our workspace (and `maia`): #2393.


model/src/shared_protocol.rs
```diff
@@ -13,45 +13,40 @@ use maia_core::secp256k1_zkp::XOnlyPublicKey;
 use maia_core::PartyParams;
 use std::ops::RangeInclusive;
 
-pub(crate) async fn verify_cets(
+pub async fn verify_cets(
```
bonomat: Looks like it doesn't need async anymore

xtra-libp2p-rollover/src/maker.rs
```diff
@@ -137,50 +149,61 @@ where
...
         let mut tasks = Tasks::default();
         tasks.add_fallible(
             {
                 let executor = self.executor.clone();
                 let oracle = self.oracle.clone();
+                let rates =  self.rates.clone();
                 let oracle_pk = self.oracle_pk;
                 let n_payouts = self.n_payouts;
                 async move {
+                    let Rates {
```
bonomat: I think we will need to add a `ttl` or similar to the rates. 
This is needed in case the automation died/is not functioning. Otherwise we might be rolling over with outdated rates.

model/src/cfd.rs
```diff
@@ -2318,21 +2318,26 @@ impl Dlc {
     pub fn signed_cet(
         &self,
         attestation: &olivia::Attestation,
-    ) -> Result<Result<Transaction, IrrelevantAttestation>> {
-        let cets = match self.cets.get(&attestation.id) {
+    ) -> Result<Result<Result<Transaction, IrrelevantAttestation>, PriceOutOfRange>> {
```
bonomat, line 2324: I think I don't see it: why do we need another layer of results?

model/src/olivia.rs
```diff
@@ -137,6 +139,32 @@ impl From<Announcement> for maia_core::Announcement {
     }
 }
 
+pub fn hourly_events(
```
bonomat, line 150: What do you think of adding a doc-comment that the resulting list is sorted and adding a test for it?

model/src/cfd.rs
```diff
@@ -2376,6 +2405,23 @@ impl Dlc {
...
+    /// All the oracle event IDs associated with the DLC.
+    ///
+    /// This includes:
+    /// - the settlement event; and
+    /// - all the liquidation events that precede it.
+    pub fn event_ids(&self) -> Vec<BitMexPriceEventId> {
+        self.cets.keys().copied().collect_vec()
+    }
+
+    pub fn liquidation_event_ids(&self) -> Vec<BitMexPriceEventId> {
```
bonomat, line 2392: Took me a while, is my understanding correct that a `liquidation_event_id` is every id which is not the `settlement_id`, i.e. the last one? 
It might be worth adding a comment.

model/src/olivia.rs
bonomat: All good, just saw that there are tests for `spaced_events`.

bonomat (commented):
I might have missed it: what's the expectation of how the contract is liquidated? Is the `monitoring` taking care of it?

luckysori:
> I might have missed it: what's the expectation of how the contract is liquidated? Is the `monitoring` taking care of it?

Yes, I think so.

My understanding is that before this PR we monitored all of `olivia`'s attestations that we could care about. Any time we found a new relevant attestation we would attempt to decrypt a CET adaptor signature using it. If successful, we would publish the commit transaction and the CET.

With these changes we have added more attestations to monitor. But the process that occurs after finding an attestation that decrypts a CET is the same: ensure that the CFD is closed with the given CET.

xtra-libp2p-rollover/src/maker.rs
luckysori: I suppose you're implying that because we removed the intermediate confirmation step on the maker for rolling over, we don't check that the `automaker` is up before rolling over. And if it is up we understand that the offer must be up-to-date.

I would argue that the current state of affairs of assuming that the offer is up-to-date simply because the `automaker` is up is not much safer than not checking anything at all. Nevertheless, I do think we should introduce a safety mechanism. But I would recommend doing in a separate PR, since this one is quite big already and the situation is not significantly worse with this work.

model/src/cfd.rs
luckysori: It's not strictly necessary to model it as separate results. It could be modelled as an enum too.

The `PriceOutOfRange` error needs to be handled separately because when we introduce liquidation events we need to ensure that we don't emit an error if the attested price is out of range for a liquididation event (see https://github.com/itchysats/itchysats/commit/151972547b99336f2709b6803cc3345cc211f6f2#diff-f078e5601c18fcb29f773d04d16902354c8f66a57b41aedaea122e11a6f8fd3dR1397-R1402).

model/src/olivia.rs
luckysori: I've added some documentation anyway!

model/src/cfd.rs
luckysori: Correct. I'll add a comment!

luckysori:
I've addressed your comments, @bonomat. Thanks for your review! I've force-pushed, but you can use the `Compare` button next to the force-push event on the timeline to see what I've changed compared to what you reviewed. Here's the link: https://github.com/itchysats/itchysats/compare/5cb86ff5182dab07d9cabcfe17bef2076042272d..187cb6f9fa983adfb1b7050c1c207e05014535ec.

I'm only mentioning it because I only recently discovered this. It makes force-pushing a lot safer for reviewers.

xtra-libp2p-rollover/src/maker.rs
bonomat: > I suppose you're implying that because we removed the intermediate confirmation step on the maker for rolling over, we don't check that the automaker is up before rolling over. And if it is up we understand that the offer must be up-to-date.

yes, this was what I thought :) 

I think a follow-up PR is fine given the size of this PR. Can you make sure to create a ticket for that once this is merged please?

model/src/cfd.rs
bonomat: Ah, now I get it. Thanks.

bonomat:
> > I might have missed it: what's the expectation of how the contract is liquidated? Is the `monitoring` taking care of it?
> 
> Yes, I think so.
> 
> My understanding is that before this PR we monitored all of `olivia`'s attestations that we could care about. Any time we found a new relevant attestation we would attempt to decrypt a CET adaptor signature using it. If successful, we would publish the commit transaction and the CET.
> 
> With these changes we have added more attestations to monitor. But the process that occurs after finding an attestation that decrypts a CET is the same: ensure that the CFD is closed with the given CET.

This makes sense. Do you think it would be worth adding a different _monitoring_ agent which monitors the price movement instead? Particularly for _collaboratively_ liquidation this could be useful.

model/src/cfd/rollover_v_1_0_0.rs
```diff
@@ -0,0 +1,156 @@
...
+use crate::Position;
+use crate::Role;
+use crate::RolloverParams;
+use crate::TxFeeRate;
+use anyhow::bail;
+use anyhow::Context;
+use anyhow::Result;
+use time::OffsetDateTime;
+
+impl Cfd {
```
bonomat, line 21: Is anyone running `rollover_v1.0`? 
Afaik we only introduced it with 0.4.19 which is a buggy version and no one is running it. That being said, we could decide to not support `rollover_v1.0` as we don't expect people to use it?

bonomat (approved):
Great work and very clean commits. It's a shame that we can't use it yet because `setup_contract::new` is not used. Do you want to merge this PR anyway or hold back until it is actually being used?

model/src/cfd/rollover_v_1_0_0.rs
luckysori: Yes, that's a valid point. I considered it, but I didn't know if we knew for certain that no one was running it. And I wanted to take the opportunity to see if the approach with multiple "versions" of the same protocol crate would work.

I don't mind if we decide to delete  the `v_1_0_0` module now, but I don't want to make the call because it's more work for me now haha.

luckysori:
> It's a shame that we can't use it yet because `setup_contract::new` is not used. Do you want to merge this PR anyway or hold back until it is actually being used?

In case it's not clear, we _can_ use dynamic liquidation, but only after the first rollover. I think we're okay to merge it, since it should be tested regularly ASAP, and it's useful by itself. 

Also the longer we wait the more tricky rebases we'll have to do 😁

bonomat:
> In case it's not clear, we can use dynamic liquidation, but only after the first rollover. I think we're okay to merge it, since it should be tested regularly ASAP, and it's useful by itself.

Thanks for the clarification. I think we should get this tested soon and see what implication this has in regards to storage size and performance.

model/src/shared_protocol.rs
```diff
@@ -13,45 +13,40 @@ use maia_core::secp256k1_zkp::XOnlyPublicKey;
...
 ) -> Result<()> {
-    tokio::task::spawn_blocking(move || {
-        for (tx, _, digits) in own_cets.iter() {
-            let counterparty_encsig = counterparty_cets
-                .iter()
-                .find_map(|(range, encsig)| (range == &digits.range()).then(|| encsig))
-                .with_context(|| {
-                    let range = digits.range();
+    for (tx, _, digits) in own_cets.iter() {
+        let counterparty_encsig = counterparty_cets
```
da-kami: I suppose finding the counterparty encsig within the CETs is significantly less expensive than the verification, so it is fine to move `spwan_blocking` inside. Might be interesting to get some benchmarks at some point.

xtra-libp2p-rollover/src/maker.rs
```diff
@@ -137,50 +149,61 @@ where
...
+                                Position::Long => funding_rate_long,
+                                Position::Short => funding_rate_short,
                             };
 
                             let (event, params, dlc, position, oracle_event_id) = cfd
                                 .accept_rollover_proposal(
                                     tx_fee_rate,
                                     funding_rate,
-                                    Some(from_params),
+                                    Some((from_event_id, from_complete_fee)),
```
da-kami: Something in this commit seems odd, `from_event_id` and `complete_fee` are removed, but here we are still using it - where are they coming from? I assume the changes are not cleanly applied after a rebase.

maker/src/routes.rs
```diff
@@ -276,6 +276,30 @@ pub async fn get_metrics<'r>(_auth: Authenticated) -> Result<String, HttpApiProb
...
 }
 
+#[derive(Debug, Clone, Copy, Deserialize)]
+pub struct RolloverConfig {
+    is_accepting_rollovers: bool,
+}
+
+#[rocket::post("/rollover-config", data = "<config>")]
+#[instrument(name = "POST /rollover-config", skip(maker), err)]
+pub async fn update_rollover_configuration(
```
da-kami: What is the motivation of this change? Because we remove the `accept` step we need another way to just stop accepting rollovers? What's the use case for this?

model/src/cfd.rs
da-kami: Maybe a solution involving an `enum` would actually be nicer. A tripple `Result` is not particularly beautiful. Not a blocker though.

maker/src/cfd.rs
```diff
@@ -589,6 +589,10 @@ impl<O, T, W> Actor<O, T, W> {
         }
     }
 
+    // TODO: Because of this we need to have two rollover actor
```
da-kami: I think we could actually refactor the old protocol to remove this step as well.
For me the decision if we want to handle `accept` in a different way is independent of the liquidation feature and could be in a separate PR.

model/src/cfd.rs
```diff
@@ -979,9 +989,11 @@ impl Cfd {
         let hours_to_charge = match version {
             rollover::Version::V1 => 1,
             rollover::Version::V2 => self.hours_to_extend_in_rollover(now)?,
-            rollover::Version::V3 => {
-                self.hours_to_extend_in_rollover_based_on_event(to_event_id, now, from_event_id)?
-            }
+            rollover::Version::V3 => self.hours_to_extend_in_rollover_based_on_event(
+                *settlement_event_id,
+                now,
+                from_event_id,
```
da-kami, line 969: There is some kind of mixup here, the variable `from_event_id` does not exist.

model/src/cfd/rollover_v_1_0_0.rs
da-kami: Given that we released `0.5.0` with rollover v1 I suppose this comment is irrelevant because we will need both.

model/src/payouts.rs
```diff
@@ -79,3 +117,35 @@ impl Payouts {
...
+    /// All the intermediate oracle announcements between the start
+    /// and end of an epoch for a CFD.
+    liquidation: Vec<Announcement>,
+}
+
+impl Announcements {
+    fn new(announcements: Vec<olivia::Announcement>) -> Result<Self> {
+        let announcements = announcements
+            .into_iter()
+            .sorted_by(|a, b| a.id.cmp(&b.id))
```
da-kami, line 135: Does it make sense to sort using the complete `BitMexPriceEventId` or should we rather sort by the `timestamp` within `BitMexPriceEventId`?

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
```diff
@@ -0,0 +1,395 @@
...
+use model::Role;
+use model::RolloverVersion;
+use std::collections::HashMap;
+use tokio_extras::FutureExt;
+use tokio_extras::Tasks;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_libp2p::Substream;
+use xtra_productivity::xtra_productivity;
+
+/// Permanent actor to handle incoming substreams for the `/itchysats/rollover/1.0.0`
```
da-kami:
```suggestion
/// Permanent actor to handle incoming substreams for the `/itchysats/rollover/2.0.0`
```

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
```diff
@@ -0,0 +1,395 @@
...
+                // The cfd logic might otherwise prevent us from starting a rollover if there is
+                // still one ongoing that was not properly ended.
+                emit_failed(order_id, e, &self.executor).await;
+
+                return;
+            }
+        };
+
+        if !self.is_accepting_rollovers {
+            emit_rejected(order_id, &self.executor).await;
```
da-kami, line 153: So, there is no `accept` anymore, but there is a `reject`?
Does that mean that the taker could either receive a `reject` or the first rollover message in the protocol flow?

da-kami (commented):
This is great work, thanks for structuring the code into commits.
I think the scope of this PR exploded a bit. 

I feel the decision to remove the rollover confirmation could have done separately to keep this contained. There are so many changes that it becomes hard to evaluate if the complete solution is still sane and no bugs sneaked in. Furthermore, rebasing this work onto the recent rollover changes will not be fun. Maybe it would be worth creating smaller PRs that build on each other next time, so some work can alrady make it into master and we don't have such a big changeset?

I will give this another review after the rebase is done - if you can easily split this work up into multiple PRs that focus on refactor / feature to change `accept` / actual new `rollover_v2` that would help, but I am not sure it's worth the effort at this stage.

xtra-libp2p-rollover/src/maker.rs
da-kami: My bad

maker/src/routes.rs
da-kami: @bonomat maybe you can explain why this feature is needed. I feel we never really turned rollovers off, so I wonder if this feature is actually justified.

model/src/cfd.rs
da-kami: Sorry, my bad

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
da-kami: Sorry, my bad, there is still an accept

maker/src/cfd.rs
luckysori: Note to self: think about why this TODO is still here AND attempt to split this PR into:

- One that removes the accept/reject automation step.
- One that extracts the new crate and introduces the liquidation feature.

xtra-libp2p-rollover/src/maker.rs
luckysori: Done: https://github.com/itchysats/itchysats/issues/2531.

luckysori:
> Do you think it would be worth adding a different _monitoring_ agent which monitors the price movement instead? Particularly for _collaboratively_ liquidation this could be useful.

I don't fully see what you have in mind. By monitoring `olivia`'s attestations we _are_ effectively monitoring price movement: each `olivia` attestation includes a price and we use it to select CETs that should have been "activated".

But I suppose if we add the feature of collaborative liquidation we will need to look at a regular price feed (the same one that `olivia` uses!).

model/src/shared_protocol.rs
luckysori: Maybe it's not clear: I've moved it _outside_. The motivation was for the `model` library not to mandate that this computation has to be called from within `spawn_blocking`. And this is only relevant now because we just moved these methods from `daemon` to `model` in https://github.com/itchysats/itchysats/pull/2533/commits/3717fa29f1a78f1f9bc1da6fa95197bbc43362db.

maker/src/cfd.rs
luckysori: As promised:

1. https://github.com/itchysats/itchysats/pull/2533 removes the maker's rollover confirmation step.
2. This PR introduces hourly liquidation events.

model/src/payouts.rs
luckysori: Great comment! 

Currently we have derived `PartialOrd` on `BitMexPriceEventId` and the `timestamp` field happens to be the first field, so we are effectively sorting by timestamp. Unless the timestamps are equal, in which case we would compare the `digits`. But those are always `20`, given that we always construct them using `BitMexPriceEventId::with_20_digits`.

But this is a bad approach, since someone could nonchalantly flip the order of the fields and assume that everything is fine. So I will add a patch fixing this.

luckysori: See https://github.com/itchysats/itchysats/pull/2378/commits/46e0890b2c8a105ef63d70c8a2dff2410e903559.

luckysori:
I want to test the liquidation feature somehow, but this can be (re-)reviewed already.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add dynamic liquidation to the DLC, enabling the CFD to be unilaterally closed every hour by either party if the oracle attests to a price close to the ends of the payout curve's domain.
+  This is currently only active _after the first rollover_.
```
da-kami, line 13: Just to make sure my understanding is correct: This means activation with any rollover after rolling this out. If I have a position already open that already had rollovers then it will become active with the first rollover after upgrading.

model/src/cfd.rs
```diff
@@ -2350,6 +2379,25 @@ impl Dlc {
 
         Ok(Ok(Ok(signed_cet)))
     }
+
+    /// All the oracle event IDs associated with the DLC.
+    ///
+    /// This includes:
+    /// - the settlement event; and
+    /// - all the liquidation events that precede it.
```
da-kami, line 2387: ❓ Would it make sense to separate the settlement event from the liquidation ones? 
Why did you opt for modelling it like this?

da-kami (approved):
Just to make sure my understanding is correct: 
With this change we will be able to liquidate hourly (as maker and taker) because we create the necessary liquidation transactions. 
This PR does not include executing the liquidation. I suppose this will be done on top.

da-kami:
Thanks for splitting it up! The reduced scope helped me quite a bit! :)

CHANGELOG.md
luckysori: Good question. If you upgrade to version 5.0.0, you will be running `/itchysats/rollover/2.0.0`, so yes.

model/src/cfd.rs
luckysori: The main reason for keeping them together is that I don't want to make changes to the `Dlc` because we then have to think about how it affects the database.

When we generate the `Dlc` we, for a very long time, have been collecting the CETs into a `HashMap<BitMexPriceEventId, Vec<Cet>>`. We can rely on the latest `BitMexPriceEventId` (the key of the map) being the settlement event ID. And we already store the settlement event ID separately in the DLC anyway.

In conclusion: we could explicitly model this information in the DLC and I think it would be an improvement. But I don't want to think about breaking changes.

maker/src/routes.rs
luckysori: @bonomat I'm merging this as is. But we can remove this feature later if it's deemed unnecessary.

bonomat: Sorry, I missed this comment. We might indeed not need this feature.

luckysori:
bors r+

I'll add the tests in a separate PR. Let's not block this any longer :D

luckysori:
> Merge conflict.

bors r+

I'm begging you.

luckysori:
bors retry

PR #2379, opened 2022-07-06
Revert addition of instrumentation argument to cargo aliases

luckysori:
Without an OTEL collector available locally we just emit errors nonstop.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,8 +2,8 @@
 linker = "aarch64-linux-gnu-gcc"
 
 [alias]
-dev-maker = "run --bin maker -- --instrumentation testnet"
-dev-taker = "run --bin taker -- --instrumentation --maker localhost:9999 --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e --maker-peer-id 12D3KooWDjzHna3pNi1Bt1DoRfrpsBREykJKXDRDxXvhJNAdDZEk testnet" # Maker ID matches seed found in `testnet/maker_seed`
+dev-maker = "run --bin maker -- testnet"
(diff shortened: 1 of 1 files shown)


klochowicz (changes requested):
but we have OTEL collector available locally? It's enough to  run `jaeger.sh`.

if anything, we can hide tonic debug logs, if you don't want to run `jaeger.sh`.

bonomat (approved):
I'm in favor of this and rather add a second command to run with instrumentation on, e.g. 

```
dev-maker-instrumented = "run --bin maker -- --instrumentation testnet"
...
```

klochowicz:
> I'm in favor of this and rather add a second command to run with instrumentation on, e.g.
> 
> ```
> dev-maker-instrumented = "run --bin maker -- --instrumentation testnet"
> ...
> ```

I guess this is a reasonable middle ground; I'll add these 2 new aliases as a second commit to this PR.

klochowicz:
After merging https://github.com/itchysats/itchysats/pull/2381 (which disables all OTEL logs by default), would there any upside of having 2 sets of aliases? I would prefer keeping it enabled as this will be more representative of our production/testnet env.

Restioson:
Related: maybe `start_all` should start jaeger too, or have an option to do so?

klochowicz:
> Related: maybe `start_all` should start jaeger too, or have an option to do so?

this would add Docker dependency to it, but  I don't mind it much as it's mostly for dev too.
 Also, but as @bonomat pointed out, a short guide about instrumentation in README might also do the trick!

luckysori:
With https://github.com/itchysats/itchysats/pull/2381 and letting `start_all.sh` start `jaeger` too I think we don't need to merge this PR anymore.

PR #2381, opened 2022-07-07
Log OpenTelemetry errors using tracing

### LINKED ISSUE #2380: Opentelemetry goes berserker mode
I had the taker and maker running locally locally over night. Now I get spammed with these errors. 

```
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full
OpenTelemetry trace error occurred. cannot send span to the batch span processor because the channel is full

```
2 questions
- What does this mean?
- Can we use our tracing library to log those errors? Since they are not json we won't be able to parse our logs to json anymore.

klochowicz:
- Create 'opentelemetry' log target to be able to filter them out easily
- disable logging by default to prevent spamming, use `RUST_LOG=opentelemetry` if needed (e.g. in cloud-conf or when collector does not work locally)

### DIFF
--- a/shared-bin/src/logger.rs
+++ b/shared-bin/src/logger.rs
@@ -68,6 +68,12 @@ pub fn init(
 
     let telemetry = if instrumentation {
         opentelemetry::global::set_text_map_propagator(TraceContextPropagator::new());
+
+        opentelemetry::global::set_error_handler(|error| {
+            ::tracing::error!(target: "opentelemetry", "OpenTelemetry error occurred: {:#}", anyhow::anyhow!(error));
+        })
+        .expect("to be able to set error handler");
+
         let cfg = trace::Config::default()
             .with_resource(Resource::new([KeyValue::new("service.name", service_name)]));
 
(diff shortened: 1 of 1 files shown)


klochowicz:
tested locally; it works well even if the OTEL collector is not running at all.
`RUST_LOG=opentelemetry` adds the logs correctly`.

Restioson (approved):

PR #2383, opened 2022-07-07
Identify Protocol

### LINKED ISSUE #1769: Design proposal for `/itchysats/identify` protocol

### LINKED ISSUE #2307: Provide feedback about protocol between maker/taker incompatibility based on `identify` protocol

---

**Example:**

---

---

Out of scope:

da-kami:
Resolves #1769 
Resolves https://github.com/itchysats/itchysats/issues/2307

In the legacy connection we had bound compatibility to the connection status, not we use the `identify` protocol to ensure compatibility.

Notes: 

- Takers don't disconnect if the maker does not speak the protocol, but we display a warning in the UI. This is more graceful, because if e.g. contract setup is not compatible but rollover is existing positions can still roll over.
- Makers currently don't care about connection information. Eventually we should change this. The maker should validate the *listening* protocols of the taker as well, otherwise a maker may repeatedly fail to e.g. end offers (if offers protocol is incompatible).

More generally speaking: We can invest more time into handling if the *expected* protocols of one side are not supported on the other side.


luckysori:
Somehow we see the commits from the [previous PR](https://github.com/itchysats/itchysats/pull/2161) here :(

daemon/src/lib.rs
```diff
@@ -118,9 +168,13 @@ impl TakerListenProtocols {
             (self.offers, offers.into()),
         ]
     }
+
+    pub fn compatible_maker_protocols() -> HashSet<String> {
+        MakerProtocols::default().supported_protocols()
+    }
```
luckysori: 📣 As soon as the maker starts supporting more than one version of a `libp2p`-based protocol for backwards compatibility (for example, after this https://github.com/itchysats/itchysats/pull/2378 gets merged), this won't work.


For example:

- Taker supports `/foo/2.0.0`.
- Build-time maker supports `/foo/1.0.0` and `/foo/2.0.0`, which is the value we get here.
- Run-time maker no longer supports `/foo/1.0.0` for whatever reason.
- Taker connects with run-time maker. We display an error even though the taker doesn't care about (cannot even perform) `/foo/1.0.0`.

da-kami: > For example:
> 
>     * Taker supports `/foo/2.0.0`.
> 
>     * Build-time maker supports `/foo/1.0.0` and `/foo/2.0.0`, which is the value we get here.
> 
>     * Run-time maker no longer supports `/foo/1.0.0` for whatever reason.
> 
>     * Taker connects with run-time maker. We display an error even though the taker doesn't care about (cannot even perform) `/foo/1.0.0`.

I think you are right, we have to be explicit about what the taker expects on the other side. This sucks a bit though, because we have to manually ensure that it is up to date :/

da-kami: I changed #2161 to introduce a `ProtocolFactory` and explicitly use the latest maker protocols for what the taker expects. I think this should solve the problem.

daemon/src/identify.rs
```diff
@@ -0,0 +1,182 @@
...
+            ),
+        );
+
+        #[allow(clippy::disallowed_methods)]
+        tokio::spawn(endpoint_context.run(endpoint));
+
+        (id.public().to_peer_id(), endpoint_address, receiver)
+    }
+
+    async fn retry_until_some<F, T>(mut f: F) -> T
```
klochowicz, line 166: maybe this could also live inside `tokio_extras` ? we seem to be copy-pasting this around for no good reason.

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub fn new(
+        endpoint: Address<Endpoint>,
+        identified_peer_channel: Option<watch::Sender<Option<PeerInfo>>>,
+    ) -> Self {
+        NUM_LIBP2P_CONNECTIONS_GAUGE.reset();
+
+        Self {
+            endpoint,
+            peer_infos: HashMap::default(),
+            peer_info_channel: identified_peer_channel,
```
klochowicz, line 32: could you settle on one of the 2 names and use it both inside and in the parameter?

daemon/src/identify/listener.rs
```diff
@@ -0,0 +1,70 @@
...
+use async_trait::async_trait;
+use libp2p_core::Multiaddr;
+use libp2p_core::PublicKey;
+use std::collections::HashSet;
+use tokio_extras::spawn_fallible;
+use xtra::Context;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor {
```
klochowicz, line 12: could you add a comment stating the purpose of this actor? (same for dialer)

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+use futures::StreamExt;
+use libp2p_core::identity::PublicKey;
+use serde::Deserialize;
+use serde::Serialize;
+use std::collections::HashSet;
+use std::time::Duration;
+use tokio_extras::FutureExt;
+
+// Start libp2p based protocols from 0.3.0 since the last wire version was 0.2.1
+const PROTOCOL_VERSION: &str = "0.3.0";
```
klochowicz, line 19:
```suggestion
const WIRE_PROTOCOL_VERSION: &str = "0.3.0";
```

without the wire_ prefix it took me a while to undestand the line and the comment, as we're in `identify` module.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,170 @@
...
+
+const TIMEOUT: Duration = Duration::from_secs(5);
```
klochowicz, line 21:
```suggestion
const MSG_TIMEOUT: Duration = Duration::from_secs(5);
```

daemon/src/lib.rs
```diff
@@ -83,6 +87,89 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
 
 pub const N_PAYOUTS: usize = 200;
 
+#[derive(Clone, Copy)]
+pub struct ProtocolFactory;
+
+impl ProtocolFactory {
```
klochowicz, line 93: I'm not a fan of this solution. It seems really fragile and hard to maintain :(

can we at least put these additions into `identify` module?

daemon/src/lib.rs
```diff
@@ -83,6 +87,89 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
...
+    const PONG_V1: &'static str = xtra_libp2p_ping::PROTOCOL_NAME;
+    const IDENTIFY_V1: &'static str = identify::PROTOCOL;
+    const OFFERS_V1: &'static str = xtra_libp2p_offer::PROTOCOL_NAME;
+    const ROLLOVER_V1: &'static str = rollover::PROTOCOL;
+    const COLLAB_SETTLEMENT_V1: &'static str = collab_settlement::PROTOCOL;
```
klochowicz: no need for these variables, you might as well just use these values below.

daemon/src/lib.rs
```diff
@@ -83,6 +87,89 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
...
+
+    const LATEST_PONG: &'static str = Self::PONG_V1;
+    const LATEST_IDENTIFY: &'static str = Self::IDENTIFY_V1;
+    const LATEST_OFFERS: &'static str = Self::OFFERS_V1;
+    const LATEST_ROLLOVER: &'static str = Self::ROLLOVER_V1;
+    const LATEST_COLLAB_SETTLEMENT: &'static str = Self::COLLAB_SETTLEMENT_V1;
+
+    const MAKER_LISTEN: usize = 4;
+    const TAKER_LISTEN: usize = 3;
+    const TAKER_EXPECTS_FROM_MAKER: usize = 4;
```
klochowicz: these variables also don't add anything (the current name actually obscures the meaning).  consider inlining, at least you can read then that it's just the array size - it will be easier to update.

maker/src/actor_system.rs
```diff
@@ -177,30 +181,54 @@ where
...
-            move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone()),
+            {
+                let listen_multiaddr = listen_multiaddr.clone();
+                let endpoint_addr = endpoint_addr.clone();
+                move || listener::Actor::new(endpoint_addr.clone(), listen_multiaddr.clone())
+            },
             always_restart_after(RESTART_INTERVAL),
         );
 
+        // TODO: Shouldn't this actor also be supervised?
```
klochowicz, line 245: yes, it would be good if it was!

shared-bin/src/to_sse_event.rs
```diff
@@ -90,3 +73,28 @@ impl ToSseEvent for Option<Quote> {
         Event::json(self).event("quote")
     }
 }
+
+#[derive(Debug, Clone, Serialize)]
+pub struct MakerCompatibility {
```
klochowicz, line 71: it's taker-specific, so it's a bit unfortunate to keep it here ;/

Perhaps move it to `taker/src/routes.rs`? or at least leave a `TODO` that it should be somewhere taker-specific.

klochowicz, line 71: nit: perhaps `MakerIncompatibility`? you seem to only send unsupported protocols :p

taker-frontend/src/App.tsx
```diff
@@ -119,6 +120,13 @@ export const App = () => {
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
     const connectedToMakerOrUndefined = useLatestEvent<ConnectionStatus>(source, "maker_status");
+    const makerCompatibilityOrUndefined = useLatestEvent<MakerCompatibility>(source, "maker_identity");
+
+    let incompatible = false;
+    if (makerCompatibilityOrUndefined) {
+        incompatible = makerCompatibilityOrUndefined.unsupported_protocols != undefined;
```
klochowicz, line 128:
```suggestion
        incompatible = makerCompatibilityOrUndefined.unsupported_protocols !== undefined;
```

taker-frontend/src/App.tsx
```diff
@@ -119,6 +120,13 @@ export const App = () => {
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
     const connectedToMakerOrUndefined = useLatestEvent<ConnectionStatus>(source, "maker_status");
+    const makerCompatibilityOrUndefined = useLatestEvent<MakerCompatibility>(source, "maker_identity");
```
klochowicz, line 124: 🚨does this work at all? isn't the event yielded from the backend `maker_compatibility`? 

`maker_identity` yields the peer info, right?

taker-frontend/src/components/IncompatibleWarning.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export default function IncompatibleWarning({ onClose }: IncompatibleWarningProps) {
+    return (
+        <Alert status="warning">
+            <AlertIcon />
+            <AlertTitle>Your daemon is incompatible with the connected maker!</AlertTitle>
+            <AlertDescription>
+                Running an incompatible version can result in unexpected behaviour. Ensure compatibility by updating to
+                the latest ItchySats version!
```
klochowicz, line 15:
```suggestion
                the latest ItchySats version
```

no need to shout all the time 😝

klochowicz (commented):
Lots of great stuff here. 

I got some ideas of improving `ProtocolFactory` pattern, but it can wait after this gets in.

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,160 @@
...
+use tokio_extras::spawn_fallible;
+use xtra::Address;
+use xtra::Context;
+use xtra_libp2p::endpoint;
+use xtra_libp2p::libp2p::PeerId;
+use xtra_libp2p::Endpoint;
+use xtra_libp2p::OpenSubstream;
+use xtra_productivity::xtra_productivity;
+
+// TODO: Move NUM_CONNECTIONS_GAUGE to a shared crate and then use it here and
```
bonomat: Stale TODO?

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,160 @@
...
+        let daemon_version = peer_info.daemon_version.clone();
+        let environment = peer_info.environment;
+
+        NUM_LIBP2P_CONNECTIONS_GAUGE
+            .with(&HashMap::from([
+                (WIRE_VERSION_LABEL, wire_version.as_str()),
+                (DAEMON_VERSION_LABEL, daemon_version.as_str()),
+                (ENVIRONMENT_LABEL, environment.to_string().as_str()),
+            ]))
+            .inc();
```
bonomat, line 78: I think this is misplaced or wrongly named: the metric says: `number_of_connections` but here you are incrementing it whenever a new `IdentifyMsgReceived` was received.

bonomat: Second question: where is `NUM_LIBP2P_CONNECTIONS_GAUGE` defined?

daemon/src/identify/listener.rs
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
+impl Actor {
+    async fn handle(&mut self, message: NewInboundSubstream, ctx: &mut Context<Self>) {
+        let NewInboundSubstream { stream, peer } = message;
+
+        // TODO: Set observed address according to the address we observed when establishing the
```
bonomat, line 43: This TODO does not seem to be connected with what it is saying. Either move it into the `new(..`) function or extract the `Multiaddr::empty` into a var. Otherwise this todo might run into the danger of becoming stale.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,166 @@
...
+pub struct IdentifyMsg {
+    protocol_version: String,
+    agent_version: String,
+    public_key: Vec<u8>,
+    listen_addrs: HashSet<Multiaddr>,
+    observed_addr: Multiaddr,
+    protocols: HashSet<String>,
+
+    /// Optional environment field that is not part of the identify spec
+    environment: Option<Environment>,
```
bonomat, line 33: Does not seem to need to be optional: 
The only way to construct `IdentifyMsg` seems to be via `IdentifyMsg::new()` which takes an `environment`.

daemon/src/identify/protocol.rs
```diff
@@ -0,0 +1,166 @@
...
+impl IdentifyMsg {
+    pub fn new(
+        daemon_version: String,
+        environment: Environment,
+        public_key: PublicKey,
+        listen_addrs: HashSet<Multiaddr>,
+        observed_addr: Multiaddr,
+        protocols: HashSet<String>,
+    ) -> Self {
+        let agent_version = format!("itchysats/{}", daemon_version);
```
bonomat, line 45: What is the additional `itchysats/` needed for?

daemon/src/identify/dialer.rs
```diff
@@ -158,3 +158,17 @@ impl Actor {
         }
     }
 }
+
+const WIRE_VERSION_LABEL: &str = "wire_version";
+const DAEMON_VERSION_LABEL: &str = "daemon_version";
+const ENVIRONMENT_LABEL: &str = "environment";
+
+static NUM_LIBP2P_CONNECTIONS_GAUGE: conquer_once::Lazy<prometheus::IntGaugeVec> =
```
bonomat, line 151: Ah, here it is. I guess the first commit does not build? 

This metric says `number of active libp2p connections` but it looks like you are tracking the number of received `IdentifyMsgReceived` messages.

bonomat (commented):
Unfortunately I can't say much about the concept so I've only added some comments.

daemon/src/identify.rs
```diff
@@ -0,0 +1,182 @@
...
+            environment: identity_msg.environment().into(),
+            protocols: identity_msg.protocols(),
+        };
+
+        Ok(identity_info)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("Conversion to identity info failed: {error}")]
```
luckysori, line 36: 🙃 
```suggestion
#[error("Conversion to peer info failed: {error}")]
```

daemon/src/identify.rs
```diff
@@ -0,0 +1,182 @@
...
+pub struct ConversionError {
+    #[source]
```
luckysori, line 38: 🙃 And then you don't have to map explicitly on line 26.

```suggestion
    #[from]
```

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,158 @@
...
+use xtra_libp2p::endpoint;
+use xtra_libp2p::libp2p::PeerId;
+use xtra_libp2p::Endpoint;
+use xtra_libp2p::OpenSubstream;
+use xtra_productivity::xtra_productivity;
+
+pub struct Actor {
+    endpoint: Address<Endpoint>,
+    peer_infos: HashMap<PeerId, PeerInfo>,
+    peer_info_channel: Option<watch::Sender<Option<PeerInfo>>>,
```
luckysori, line 19: 🤔 It's weird to me that this needs to be an `Option`.

daemon/src/identify/listener.rs
luckysori: Perhaps we could just remove the `observed_address` from the protocol since we're clearly not using it atm.

daemon/src/identify/protocol.rs
luckysori: I would remove the constant and field in `IdentifyMsg` altogether, as discussed [here](https://github.com/itchysats/itchysats/pull/2161#discussion_r916478075).

daemon/src/identify/protocol.rs
luckysori: I think this is trying to follow the spec: https://github.com/libp2p/specs/tree/master/identify#agentversion.

daemon/src/lib.rs
```diff
@@ -61,6 +64,7 @@ pub mod rollover;
 pub mod seed;
 pub mod setup_contract;
 // TODO: Remove setup_contract_deprecated module after phasing out legacy networking
+pub mod identify;
```
luckysori, line 70: 🔧 
```suggestion
pub mod identify;
// TODO: Remove setup_contract_deprecated module after phasing out legacy networking
```

taker/src/routes.rs
```diff
@@ -60,6 +61,7 @@ pub async fn feed(
     rx: &State<Feeds>,
     rx_wallet: &State<watch::Receiver<Option<WalletInfo>>>,
     rx_maker_status: &State<watch::Receiver<ConnectionStatus>>,
+    rx_maker_identity: &State<watch::Receiver<Option<identify::PeerInfo>>>,
```
luckysori, line 65: 🙃 I get confused with the mixed terminology between peer info, identity and identify. I think the chosen name `PeerInfo` is good enough, so it'd be good to be consistent.

```suggestion
    rx_maker_peer_info: &State<watch::Receiver<Option<identify::PeerInfo>>>,
```

shared-bin/src/to_sse_event.rs
luckysori: I would definitely move it to the taker if possible. The `ToSseEvent` trait is public, so it should be doable.

shared-bin/src/to_sse_event.rs
```diff
@@ -90,3 +73,28 @@ impl ToSseEvent for Option<Quote> {
...
+        let unsupported_protocols = self.as_ref().map(|identified_peer| {
+            maker_protocols_expected_by_taker
+                .difference(&identified_peer.protocols)
+                .cloned()
+                .collect()
+        });
+
+        let maker_compatibility = MakerCompatibility {
+            unsupported_protocols,
+        };
```
luckysori, line 89: 🔧 This bit belongs in a `MakerCompatibility::new` method.

shared-bin/src/to_sse_event.rs
luckysori: Something like `MakerMissingProtocols(HashSet<_>)` (without a named field) would be clearer for me.

shared-bin/src/to_sse_event.rs
```diff
@@ -90,3 +73,28 @@ impl ToSseEvent for Option<Quote> {
...
+    unsupported_protocols: Option<HashSet<String>>,
+}
+
+impl ToSseEvent for Option<identify::PeerInfo> {
+    fn to_sse_event(&self) -> Event {
+        let maker_protocols_expected_by_taker = ProtocolFactory::taker_expects_from_maker();
+
+        let unsupported_protocols = self.as_ref().map(|identified_peer| {
+            maker_protocols_expected_by_taker
+                .difference(&identified_peer.protocols)
```
luckysori, line 82: 🤔 Comparing strings is not enough as soon as the maker runs a _different_ version of a protocol that is actually _compatible_ with ours e.g. maker runs `rollover:2.1.0` and taker runs `rollover:2.0.0`.

maker/src/actor_system.rs
luckysori: Either do it in a separate commit or create an issue. But delete the TODO imo.

daemon/src/lib.rs
```diff
@@ -261,6 +352,24 @@ where
...
+                    environment,
+                    identity.public(),
+                    HashSet::new(),
+                    ProtocolFactory::taker_listen_protocols(),
+                )
+            }
+        });
+
+        let identify_dialer_actor =
+            identify::dialer::Actor::new(endpoint_addr.clone(), Some(identify_info_feed_sender))
```
luckysori, line 408: 🔧 Should be supervised like the one for the maker.

daemon/src/lib.rs
luckysori: I think this logic belongs in `daemon` because it's application-specific (the set of protocols that maker and taker run). But I would move it to a sub-module. Although `identify` seems like the wrong place to me.

daemon/src/lib.rs
```diff
@@ -83,6 +87,89 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
 
 pub const N_PAYOUTS: usize = 200;
 
+#[derive(Clone, Copy)]
+pub struct ProtocolFactory;
+
+impl ProtocolFactory {
+    const PONG_V1: &'static str = xtra_libp2p_ping::PROTOCOL_NAME;
```
luckysori, line 94: 🙃 The protocol is called `/ipfs/ping`.

daemon/src/lib.rs
```diff
@@ -83,6 +87,89 @@ pub const PING_INTERVAL: Duration = Duration::from_secs(30);
...
+    const IDENTIFY_V1: &'static str = identify::PROTOCOL;
+    const OFFERS_V1: &'static str = xtra_libp2p_offer::PROTOCOL_NAME;
+    const ROLLOVER_V1: &'static str = rollover::PROTOCOL;
+    const COLLAB_SETTLEMENT_V1: &'static str = collab_settlement::PROTOCOL;
+
+    const LATEST_PONG: &'static str = Self::PONG_V1;
+    const LATEST_IDENTIFY: &'static str = Self::IDENTIFY_V1;
+    const LATEST_OFFERS: &'static str = Self::OFFERS_V1;
+    const LATEST_ROLLOVER: &'static str = Self::ROLLOVER_V1;
+    const LATEST_COLLAB_SETTLEMENT: &'static str = Self::COLLAB_SETTLEMENT_V1;
```
luckysori: 🤔 I don't think these constants give you any guarantees that you are using the correct protocol identifiers in the end. And they kind of obfuscate what's actually going on. I would try to use the actual identifiers directly.

I like the idea of tying the set of listen protocols we say we support with the actual listen protocols we register with the `Endpoint`. It would be better to ask the `Endpoint` about this information, but we have to register the `identify::listener::Actor` too, so there would seem to be a circular dependency there.

I'm gonna try to make sense of what we're doing here. I think this `struct` is trying to do 3 things:

1. Ensuring that taker and maker are compatible at compile-time.
2. Ensuring that the listen protocols we say we support match the ones we run.
3. Allowing us to decide if the maker supports all the listen protocols the taker expects.

I think (1) is somewhat covered by the test `ensure_expected_protocols_are_supported`. We do need to define somewhere what the maker supports and what the taker expects, because they can be different: the maker might support a few more protocols than the taker cares about. But we should be using semantic versioning here, otherwise we will make mistakes in the future.

This makes me think that we should model these protocols a bit like this:

```rust
struct MakerListenProtocols {
    identify: Vec<semver::Version>,
    rollover: Vec<semver::Version>,
    // etc.
}

struct RequiredMakerProtocols {
    identify: semver::VersionReq,
    rollover: semver::VersionReq,
    // etc.
}

fn does_maker_fulfill_taker_needs(maker: MakerListenProtocols, req: RequiredMakerProtocols) -> bool {
    maker.identify.iter().any(|identify| req.identify.matches(&identify)) &&
    maker.rollover.iter().any(|rollover| req.rollover.matches(&rollover)) && 
    // etc.
}
```

We could use this function in the test to ensure compile-time compatibility. And we could also use it for (3) when actually checking compatibility during runtime: `RequiredMakerProtocols` would be set at compile time for the taker and we would have to deserialise the `MakerListenProtocols` during runtime.

Finally (2) could be achieved by adding a method to `MakerListenProtocols` (and a `TakerListenProtocols` that is not specified here but can be imagined) that works similarly to the `{maker,taker}_protocol_handlers` methods in this PR.

In conclusion, I think the intentions and ideas are good, but I would like to see it modelled differently. I'm happy to give it a go myself if you think my proposal is sound.

luckysori (commented):
I think we're getting closer!

I had some extended thoughts on how to model the `ProtocolFactory` differently. Let me know what you think.

daemon/src/identify/dialer.rs
luckysori: No longer that weird because I see that the maker doesn't care and passes a `None`. 

Perhaps it would be cleaner to have two constructors. Having to provide `None` is not very ergonomic.

daemon/src/lib.rs
da-kami: I am not convinced we should thrive for compatibility anymore. It might bring more problems than it helps at the moment. Notably:

1. we might get "spammed" by other nodes speaking it.
2. we should be fully compatible, which we are not. (i.e. serialization, ...)

So I think we might as well roll our own. If we ever see an advantage in being compatible we can roll that out as separate protocol I think.

daemon/src/lib.rs
da-kami: Thanks for the analysis!

> [snip]
> I like the idea of tying the set of listen protocols we say we support with the actual listen protocols we register with the Endpoint. It would be better to ask the Endpoint about this information, but we have to register the identify::listener::Actor too, so there would seem to be a circular dependency there.

I wanted to avoid this also because of the points that you list below. With using the protocol factory we *should* be able to ensure compatibility at compile time. Also, we avoid sending around unnecessary messages in the system - this information is static and I don't think going to the endpoint is necessary.

> [snip]
> But we should be using semantic versioning here, otherwise we will make mistakes in the future.

I don't see the advantage of using semver for the protocol comparison. The taker expects the maker to support the exact same protocol string, otherwise it will not work. As far as I understand protocol negotiation is not semantically versioned but the protocol string has to match exactly - correct me if I'm wrong! That's why I tied this to the actual protocol strings. 
Currently the protocols are modeled as simple `str`s, e.g. `/itchysats/rollover/1.0.0`. Would you then use `semver` to come up with the actual protocol string as well?

---

I am happy if you iterate on this one more time to see improvements!
I agree that it is not perfect yet, because the factory might return different protocols for different function calls. We could try to cover this by more tests, but I agree that it is a bit cumbersome.

daemon/src/lib.rs
luckysori: All valid points, but calling this `PONG` was jarring.

shared-bin/src/to_sse_event.rs
luckysori: It's not so easy to move it to a taker specific place (outside of this crate), because you run into not being able to implement a foreign trait on a foreign type.

daemon/src/lib.rs
luckysori: > I don't see the advantage of using semver for the protocol comparison. The taker expects the maker to support the exact same protocol string, otherwise it will not work. As far as I understand protocol negotiation is not semantically versioned but the protocol string has to match exactly - correct me if I'm wrong! That's why I tied this to the actual protocol strings.

I fully expected `multistream-select` to support protocol negotiations using semver but I was wrong! Nothing to action here unfortunately.

da-kami:
Since the rebase was somewhat of a pain I opted for quashing the `identify` related commits. I moved the `ConnectionStatus` cleanup to the end`.

The `ProtocolFactory` is still somewhat odd - @luckysori I thought you wanted to change that one, but in the commits on this PR it was not changed before I squashed. Was there something missing?

da-kami:
closed in favor of https://github.com/itchysats/itchysats/pull/2569

PR #2386, opened 2022-07-08
Don't sort events by timestamp

### LINKED ISSUE #2196: Don't sort events by timestamp

da-kami:
- Remove the parallel iterator because it might change the order of events when processing them.
- Remove ordering by timestamp, because duplicated timestamps can happen; e.g. for collab settlement events this can have problematic side effects.
- Ensure order of events by `order by events.id`, because `events.id` is incremental, other than the timestamp.

Resolves #2196

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -658,17 +658,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-deque"
-version = "0.8.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6455c0ca19f0d2fbf751b908d5c55c1f5cbc65e03c4225427254b46890bdde1e"
-dependencies = [
- "cfg-if",
- "crossbeam-epoch",
- "crossbeam-utils",
-]
-
-[[package]]
 name = "crossbeam-epoch"
 version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -786,7 +775,6 @@ dependencies = [
  "pretty_assertions",
  "prometheus",
  "rand 0.6.5",
- "rayon",
  "reqwest",
  "rust_decimal",
  "rust_decimal_macros",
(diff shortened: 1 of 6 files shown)


da-kami:
Note: I don't think this should have significant impact on performance, but we should see to test it with the mainnet db.

Note: I tried to write a test to reproduce the previous behaviour where the order of events gets bodged, but was unable to reproduce it locally using an in-memory db. I was writing 4 different events a thousand times and loading them and then checked the order and it was always ensured. But from the ticket we see that the order can be messed up sometimes (I was able to "proof" that on testnet, given the behavior we saw). I stopped investing time into the test because there are too many possibilities concerning the test setup (e.g. is it only when there is an event with DLC attached? Is it only if not in-memory db? ...). I think removing the parallel iterator is a good thing, because we need ensured order, and we just can't be sure what happens. Removing this complexity will help us to be able to understand what is actually happening.

klochowicz (approved):
let's give this a go over the weekend!

PR #2388, opened 2022-07-08
Ensure to monitor for finality after `LockConfirmedAfterFinality`

### LINKED ISSUE #2387: Collab settlement never moved to `closed` if CFD closed before lock confirmation
This is a follow up of #2196 

When looking at the event history of these CFDs:

image

image

we can see that we get a `LockConfirmedAfterfinality` but we never receive a `CollaborativeSettlementConfirmed` event.

I decoded the collab settlement transaction of CFD `257` and the transaction did indeed make it on chain, see:
https://mempool.space/tx/c8fdd148d1dd6dad47672cedb4114eb4749ebb46bfc6900fa33f854899c97d29

However, monitoring never reported the confirmation of the collab settlement transaction.
This can have various reasons and will need some more analysis.

As a side effect these two CFDs are never moved to the `closed_cfds` table, because the necessary criteria would be `CollaborativeSettlementConfirmed`.

da-kami:
Resolves #2387 

We should not treat `LockConfirmedAfterFinality` as final state in the monitor, because this would mean we just stop monitoring for other things.
This can lead to scenarios where we never confirm collab settlement or CET in cases where `LockConfirmedAfterFinality` is recorded before the finality confirmation is picked up.

Treating `LockConfirmedAfterFinality` like `LockConfirmed` is correct and should not have side effects in the monitor.
`LockConfirmedAfterFinality` should not have any influence on other monitoring decisions besides the lock transaction.

---

Note: I went through the commit history of this code section to make sure that we didn't specify `LockConfirmedAfterFinality` on purpose at this point. There was no specific change concerning this. I think this was wrong all along but we never noticed because it only becomes apparent in edge cases where a user closes a position before lock was confirmed (which happens rarely).

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -242,7 +242,7 @@ impl Cfd {
                 commit_tx: None,
                 ..self
             },
-            LockConfirmed => Self {
+            LockConfirmed | LockConfirmedAfterFinality => Self {
                 monitor_lock_finality: false,
                 lock_tx: None,
                 ..self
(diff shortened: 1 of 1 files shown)


klochowicz (approved):
seems reasonable.
thanks for doing the code archeology re: history behind the change 👍

klochowicz:
bors retry

PR #2391, opened 2022-07-08
Add more spans

Restioson:
This adds more spans, which makes the jaeger span search a bit more readable - more semantic names, and less `xtra_actor_request` at the top level. I also set the level filter for traces to DEBUG, and added a new `sleep_silent` which does not emit a span.

~~This is a draft, as it requires some changes in xtra to be merged first.~~ I will switch this PR over to xtra master on Monday, rebase, and it should be ready for review.

## Before

This is with the backported DEBUG filter for a fair comparison.

view of jaeger before this change

This is kinda hard to read without opening all the spans manually to see what message they are.

## After

More readable at a glance, with semantically named spans representing the *purpose* of the operations being done in the code (e.g receive a new quote), rather than the *mechanism* by which they are carried out (i.e an actor request).

view of jaeger after this change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/Restioson/xtra", rev = "7e9fccc1e02dae3df4ba99126e6f1a0059673d7e" } # Unreleased
+xtra = { git = "https://github.com/comit-network/xtra", branch = "span-late-create" } # TODO Switch back to xtra/master after https://github.com/Restioson/xtra/pull/120 gets merged
 maia = { git = "https://github.com/comit-network/maia", rev = "fc6b78b98407b10b55f8cfd152062ad77f98cd9f" } # Unreleased
 maia-core = { git = "https://github.com/comit-network/maia", tag = "0.1.1", package = "maia-core" } # Pinned to support maia 0.1 and 0.2
 xtra_productivity = { git = "https://github.com/comit-network/xtra-productivity", rev = "0bfd589b42a63149221dec7e95aca932875374dd" } # Unreleased
--- a/maker/src/cfd.rs
+++ b/maker/src/cfd.rs
@@ -239,6 +239,7 @@ impl<O, T, W> Actor<O, T, W> {
         }
     }
 
+    #[tracing::instrument(skip(self))]
     async fn update_connected_takers(&mut self) -> Result<()> {
         self.projection
             .send_async_safe(projection::Update(
--- a/xtra-libp2p-offer/src/taker.rs
+++ b/xtra-libp2p-offer/src/taker.rs
@@ -1,6 +1,7 @@
 use crate::protocol;
 use async_trait::async_trait;
 use model::MakerOffers;
+use tracing::Instrument;
 use xtra::prelude::MessageChannel;
 use xtra_libp2p::NewInboundSubstream;
 use xtra_productivity::xtra_productivity;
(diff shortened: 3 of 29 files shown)


Restioson:
I've just realised that this still requires one outstanding change in xtra: https://github.com/Restioson/xtra/pull/120

xtra-libp2p-offer/src/taker.rs
```diff
@@ -25,10 +26,11 @@ impl Actor {
 
         let task = async move {
             let offers = protocol::recv(stream).await?;
-
-            tracing::debug!(%peer, ?offers, "Received offers");
-
-            maker_offers.send(LatestMakerOffers(offers)).await?;
+            let span = tracing::debug_span!("Received offers", %peer, ?offers);
```
klochowicz: nit: perhaps more readable name: 

```suggestion
            let span = tracing::debug_span!("Received new offers from maker", %peer, ?offers);
```

Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/Restioson/xtra", rev = "7e9fccc1e02dae3df4ba99126e6f1a0059673d7e" } # Unreleased
+xtra = { git = "https://github.com/Restioson/xtra", rev = "afeffb7155f02ca81f84d5d74e096fca906dae69" } # TODO Unreleased
```
klochowicz: this adds dependency on a side branch again? is there a chance to avoid it? if it's absolutely necessary, I'd prefer to be explicit about it (ie. `branch = span-late-create`), but only if there's no reasonable chance to merge that soon.

klochowicz (commented):
meaningful names make a huge difference! I'm happy for it to go ahead, but slightly worried about the side branch situation (e.g. if you create a PR from that branch, the branch will get automatically deleted when merged, resulting in itchysats not building).

if needed, we could create one more branch in our fork with that temporarily, so at least it's a stable ref. I guess it all depends on the timeline in xtra.

Cargo.toml
Restioson: In an ideal world, we could get https://github.com/Restioson/xtra/pull/120 merged sooner rather than later and then get this PR in after, pinning xtra again a commit on master which includes the changes from that PR. It depends how soon we want to get this PR in though.

Restioson: Side note: this will technically compile without https://github.com/Restioson/xtra/pull/120, but `send(..).instrument(span)` will not work right.

Cargo.toml
```diff
@@ -22,7 +22,7 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-xtra = { git = "https://github.com/Restioson/xtra", rev = "7e9fccc1e02dae3df4ba99126e6f1a0059673d7e" } # Unreleased
+xtra = { git = "https://github.com/Restioson/xtra", branch = "span-late-create" } # TODO Unreleased
```
klochowicz: I have created a stable reference until xtra#120 gets in (and updated the comment - TODO comments that ain't actionable are super easy too address later)

```suggestion
xtra = { git = "https://github.com/comit-network/xtra", branch = "span-late-create" } # TODO Switch back to xtra/master after https://github.com/Restioson/xtra/pull/120 gets merged
```

klochowicz (approved):
thanks! I'm happy as long as you point to the branch in comit-network (we'll switch back when xtra#120 gets merged).

maker/src/cfd.rs
```diff
@@ -239,6 +239,7 @@ impl<O, T, W> Actor<O, T, W> {
         }
     }
 
+    #[instrument]
```
klochowicz: does this compile? I tried to compile from scratch and it seems like it needs `tracing::instrument`. we might want to also `skip(self)` here as otherwise we'll get a lot of info that's pretty useless for us (maker actor does not change that much)

Restioson: Ah yes, my mistake. I'll add it now.

We can add skip, yes, but if it isn't really worth debug printing, maybe we could also skim down the debug implementation itself so that we don't need to remember to skip it all the time? This might be detrimental if we do actually want to print it. The other option is to remove the debug derive if it's unused except here.

klochowicz: I don't think we have `Debug` implemented for this actor at all? I don't mind really basic debug impl, but in this case, I don't see any point in printing this info. we can add it later on.

Restioson: Addressed now

Restioson:
Force pushed this with all the changes

PR #2399, opened 2022-07-11
Add support for tokio console

klochowicz:
Tokio console support can be enabled by passing `--tokio-console` flag to maker
or taker binary.

When enabled, it is possible to see spawned tasks via `tokio-console` CLI util.

See more: https://docs.rs/tokio-console/


klochowicz:
screenshots of tokio-console running:

image

image

image

da-kami (approved):

Restioson (approved):

bonomat:
Do we really need this merged in master? I think this tool is more useful locally and can be added on demand if needed. I'm afraid that with additional dependencies and features we might be adding more problems.

klochowicz:
> Do we really need this merged in master? I think this tool is more useful locally and can be added on demand if needed. I'm afraid that with additional dependencies and features we might be adding more problems.

this is meant to be used locally, it is not enabled by default and there is no intention of enabling it in production - this just adds one more tool into our toolbox for debugging issues locally. 

Otherwise, if we e.g. had it on a feature branch somewhere, it would require us to recompile the whole workspace from scratch every time a person wants to use it (and then back again).

bonomat:
If it is really just for locally I propose to hide it behind a feature flag and enable that feature flag by default for our aliases `dev-maker` and `dev-taker`.

PR #2400, opened 2022-07-12
Fix getting daemon version

### LINKED ISSUE #2318: Frontend shows `Daemon version:  unknown`
Version 0.4.21
Platforms: 
- raspiblitz
- umbrel

Both show daemon version unknown in UI

image

bonomat:
I cannot really explain why fetch does not work here. It throws an error when fetching an unauthorized api url even though authorization is not necessary. I've fixed it with a different library 'axios' which works like charm

Fixes #2318

### DIFF
--- a/taker-frontend/package.json
+++ b/taker-frontend/package.json
@@ -27,6 +27,7 @@
     "@types/semver": "^7.3.10",
     "@vitejs/plugin-react": "^1.3.2",
     "@vitejs/plugin-react-refresh": "^1.3.1",
+    "axios": "^0.27.2",
     "dayjs": "^1.11.2",
     "eslint": "^8.19.0",
     "eslint-config-react-app": "^7.0.1",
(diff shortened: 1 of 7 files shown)


taker-frontend/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/semver": "^7.3.10",
     "@vitejs/plugin-react": "^1.3.2",
     "@vitejs/plugin-react-refresh": "^1.3.1",
+    "axios": "^0.27.2",
```
luckysori, line 30: 👍 For some reason this reminds of [this](https://github.com/comit-network/comit-i/blob/8a8496931747fd283632ee18963ca75f082b195d/package.json#L17).

luckysori (approved):

taker-frontend/package.json
bonomat: Mhmm.. comit-i ... good memories 😅

PR #2402, opened 2022-07-12
Remove spammy return values from instrumented GET requests

klochowicz:
GET /metrics would print all the metric values, which whole purpose is to be
used in prometheus metrics, not in loki/tempo.

taker identities are also already collected in a different place.


Restioson (approved):
Sorry, I missed this one :sweat_smile:

PR #2403, opened 2022-07-12
Configurable load test

da-kami:
Note: @Restioson I used branch `test-instrumentation-load-test` to play with the instrumented tests (that branch was on top of your instrumentation code from yesterday).

This PR just contains the load test part; since I removed the `traced_test` library we were initially using one cannot see trace output at the moment. Rebasing this on @Restioson's work wil fix this so I did not bother too much.

I changed the load tests to be more configurable per test and simplified a few things.
The test runner is still somewhat complex, because one can run multiple bobs, multiple protocols and multiple protocol runs in a configurable way. I think this is acceptable, the tests not proof that multiple different scenarios pass under load.

I am not sure how the tests will behave on CI. 
I am not sure if the tests should always be run on CI given that some of them are quite expensive. 
But even if we don't run (all of) them I think we should definitely keep them but potentially ignore some/all. 

Observations: 

- After cleaning up the test setup to be independent of initial connection time the tests pass with high numbers on debug for me. My conclusion so far: Our design should be fine with handling `1000+` takers. 
- The initial connection time changes with the amount of bobs, i.e. the more bobs the more time we need to connect. This is (to some extent) independent of having multiple protocol handlers.

### DIFF
--- a/xtra-libp2p/tests/load.rs
+++ b/xtra-libp2p/tests/load.rs
@@ -30,6 +30,7 @@ use xtras::SendAsyncSafe;
 mod util;
 
 #[tokio::test]
+#[ignore]
 async fn given_single_protocol_when_one_protocol_run_then_alice_handles_1000_bobs() {
     const BOBS: usize = 1000;
     const PROTOCOLS: usize = 1;
@@ -40,6 +41,7 @@ async fn given_single_protocol_when_one_protocol_run_then_alice_handles_1000_bob
 }
 
 #[tokio::test]
+#[ignore]
 async fn given_single_protocol_when_multiple_protocol_run_then_alice_handles_100_bobs_with_100_protocol_runs(
 ) {
     // Note: It can actually handle 1000 bobs x 1000 runs, but that takes significant time
@@ -53,6 +55,7 @@ async fn given_single_protocol_when_multiple_protocol_run_then_alice_handles_100
 }
 
 #[tokio::test]
+#[ignore]
 async fn given_multiple_protocol_when_one_protocol_run_then_alice_handles_1000_bobs() {
     const BOBS: usize = 1000;
     const PROTOCOLS: usize = BOBS; // One distinct protocol per Bob and distinct protocol handlers for each protocol on Alice
--- a/xtra-libp2p/tests/load.rs
+++ b/xtra-libp2p/tests/load.rs
@@ -1,0 +1,394 @@
+use crate::util::make_node;
+use crate::util::EndpointSubscriberStats;
+use anyhow::Context as _;
+use anyhow::Result;
+use async_trait::async_trait;
+use asynchronous_codec::Bytes;
+use futures::SinkExt;
+use futures::StreamExt;
+use libp2p_core::identity::Keypair;
+use libp2p_core::transport::MemoryTransport;
+use libp2p_core::Multiaddr;
+use std::time::Duration;
+use tracing::subscriber::DefaultGuard;
+use tracing_subscriber::util::SubscriberInitExt;
+use xtra::message_channel::MessageChannel;
+use xtra::spawn::TokioGlobalSpawnExt;
+use xtra::Actor;
+use xtra::Address;
+use xtra::Context;
+use xtra_libp2p::endpoint::ConnectionEstablished;
+use xtra_libp2p::endpoint::Subscribers;
+use xtra_libp2p::Connect;
+use xtra_libp2p::Endpoint;
+use xtra_libp2p::ListenOn;
+use xtra_libp2p::NewInboundSubstream;
+use xtra_libp2p::OpenSubstream;
+use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncSafe;
+
+mod util;
+
+#[tokio::test]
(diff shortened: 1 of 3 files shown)


xtra-libp2p/tests/load.rs
```diff
@@ -0,0 +1,392 @@
...
+    let name = String::from_utf8(bytes.to_vec())?;
+
+    stream.send(Bytes::from("Listener")).await?;
+
+    assert_eq!(name, "Dialer");
+
+    Ok(())
+}
+
+pub fn init_tracing() -> DefaultGuard {
```
luckysori, line 388: 🙃 Could move this to `util`.

xtra-libp2p/tests/load.rs
```diff
@@ -0,0 +1,392 @@
...
+    tracing_subscriber::fmt()
+        .with_env_filter("WARN")
+        .with_env_filter("load=debug")
+        .with_env_filter("xtra=debug")
+        .with_env_filter("xtra_libp2p=debug")
+        .with_env_filter("xtras=debug")
+        .with_test_writer()
+        .set_default()
```
luckysori, line 396: 🔧 In order to configure `jaeger` and for the logs to also appear on `stdout` I recommend using something like what I did in the `simple-load-test` branch (which is inspired by what we do in `shared_bin::logger`):

https://github.com/itchysats/itchysats/blob/470d6d7f7111384d3e8c1792d750e771a2c87b83/xtra-libp2p/tests/load.rs#L228-L288

The bit that allows us to see the logs is calling `try_init` on the `Registry`. It's different to what we were doing when we were working on these tests together.

bonomat (approved):
Good job. Do we intent to merge this and run it as part of the CI? If so, aren't we running into the danger of flaky tests if the CI is under heavy load?

luckysori:
> Do we intent to merge this and run it as part of the CI? If so, aren't we running into the danger of flaky tests if the CI is under heavy load?

I think we should merge it after it gets some more reviews, but I would not run any of these tests on CI. I think they're a useful tool to investigate problems with the `Endpoint`, but they're too expensive on CI, particularly for all those changes which don't touch this part of the codebase.

I think we should just run them locally when we iterate on `Endpoint`.

luckysori (commented):
Just a couple things I've noticed. Haven't reviewed in full yet.

xtra-libp2p/tests/load.rs
da-kami: I purposely created this PR without the Jaeger instrumentation because Cael was still working on it on his fork which makes it a bit hard to depend on without creating conflicts.
But I think these tests are actually independent of the other instrumentation, i.e. these tests have a separate instrumentation setup, so I can bring that commit back into this PR :)
Sorry, did not think that through enough, I will see to add the commit that adds the Jaeger instrumentation.

da-kami:
> > Do we intent to merge this and run it as part of the CI? If so, aren't we running into the danger of flaky tests if the CI is under heavy load?
> 
> I think we should merge it after it gets some more reviews, but I would not run any of these tests on CI. I think they're a useful tool to investigate problems with the `Endpoint`, but they're too expensive on CI, particularly for all those changes which don't touch this part of the codebase.
> 
> I think we should just run them locally when we iterate on `Endpoint`.



> > Do we intent to merge this and run it as part of the CI? If so, aren't we running into the danger of flaky tests if the CI is under heavy load?
> 
> I think we should merge it after it gets some more reviews, but I would not run any of these tests on CI. I think they're a useful tool to investigate problems with the `Endpoint`, but they're too expensive on CI, particularly for all those changes which don't touch this part of the codebase.
> 
> I think we should just run them locally when we iterate on `Endpoint`.



> > Do we intent to merge this and run it as part of the CI? If so, aren't we running into the danger of flaky tests if the CI is under heavy load?
> 
> I think we should merge it after it gets some more reviews, but I would not run any of these tests on CI. I think they're a useful tool to investigate problems with the `Endpoint`, but they're too expensive on CI, particularly for all those changes which don't touch this part of the codebase.
> 
> I think we should just run them locally when we iterate on `Endpoint`.

If you consider that the longest load test run is configured for allowing `120` seconds they should not take longer than 2 mins given a run in parallel
But I think it's reasonable to ignore them on CI, I will mark them with `ignore`.

xtra-libp2p/tests/load.rs
da-kami: After talking to @Restioson I came to the conclusion that which ever PR comes first, this one or https://github.com/itchysats/itchysats/pull/2404 will require some rebase in the other. 

> But I think these tests are actually independent of the other instrumentation

Counter to my comment above we should actually make use of the macro that is added in #2404 so I don't intend to add the instrumentation in this PR unless #2404 merges first :)

da-kami:
Since we are not sure about merging these tests I am closing this PR for now. If you feel we should keep them as ignored tests in the codebase please re-open.

klochowicz:
@da-kami I feel it's good to include them if they add value (a lot of time and effort went into creating load tests, it would be a shame to need to start from scratch!), even if they are ignored by default by CI.
After this gets merged, I can add a CI cron job that runs them once a day or so.

PR #2404, opened 2022-07-12
Instrument tests with custom macro & export over OTLP

Restioson:
This PR uses a custom `#[otel_test]` macro to instrument `daemon-tests` and export the spans over OTLP to Jaeger, when the `ITCHYSATS_TEST_INSTRUMENTATION` environment variable is set to `1`.

This currently uses the simple exporter. I tried the batch exporter, but this strangely caused some tests to hang.

## TODOs
- [x] Disable instrumentation option for CI
- [x] Get gRPC to build on ARM CI, or swap it for tonic (which doesn't currently work)
- [x] How can we configure the service name and log levels for crates other than daemon-tests?

### DIFF
--- a/daemon-tests/Cargo.toml
+++ b/daemon-tests/Cargo.toml
@@ -13,6 +13,7 @@ maker = { path = "../maker" }
 mockall = "0.11"
 mockall_derive = "0.11"
 model = { path = "../model" }
+otel-tests = { path = "../otel-tests" }
 rand = "0.6"
 rust_decimal = "1.25"
 rust_decimal_macros = "1.25"
@@ -21,7 +22,6 @@ time = "0.3.11"
 tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "net", "tracing"] }
 tokio-extras = { path = "../tokio-extras", features = ["xtra"] }
 tracing = { version = "0.1" }
-tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt", "ansi", "env-filter", "local-time", "tracing-log", "json"] }
 xtra = { version = "0.6", features = ["instrumentation"] }
 xtra-bitmex-price-feed = { path = "../xtra-bitmex-price-feed" }
 xtra-libp2p = { path = "../xtra-libp2p" }
--- a/daemon/src/seed.rs
+++ b/daemon/src/seed.rs
@@ -9,6 +9,8 @@ use model::libp2p::PeerId;
 use rand::Rng;
 use sha2::Sha256;
 use std::convert::TryInto;
+use std::fmt;
+use std::fmt::Debug;
 use std::path::Path;
 
 /// Struct containing keys for both legacy and libp2p connections.
@@ -94,6 +96,12 @@ pub struct RandomSeed([u8; 256]);
 impl Seed for RandomSeed {
     fn seed(&self) -> Vec<u8> {
         self.0.to_vec()
+    }
+}
+
+impl Debug for RandomSeed {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("RandomSeed").field(&"...").finish()
     }
 }
 
--- a/otel-tests-macro/src/lib.rs
+++ b/otel-tests-macro/src/lib.rs
@@ -1,0 +1,48 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::ItemFn;
+
+#[proc_macro_attribute]
+pub fn otel_test(_attribute: TokenStream, item: TokenStream) -> TokenStream {
+    let fn_item = syn::parse::<ItemFn>(item).unwrap();
+
+    let sig = fn_item.sig;
+    let block = fn_item.block;
+    let name = &sig.ident;
+    let attrs = fn_item.attrs;
+
+    let test = if attrs.is_empty() {
+        quote!(#[otel_tests::__reexport::tokio::test])
+    } else {
+        quote!()
+    };
+
+    let tokens = quote! {
+        #test
+        #(#attrs)*
+        #sig {
+            ::otel_tests::init_tracing(module_path!());
+
+            let caught = {
+                ::otel_tests::__reexport::futures::FutureExt::catch_unwind(std::panic::AssertUnwindSafe(tracing::Instrument::instrument(
+                    async #block, tracing::info_span!(stringify!(#name))
+                ))).await
+            };
+
+            // Give the otel thread time to receive the spans before flush
+            #[allow(clippy::disallowed_methods)]
+            ::otel_tests::__reexport::tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+
+            // If this is the last test that's running, the main thread might exit. Then, the
+            // opentelemetry exporter thread might not have exported all of its spans yet, leading
+            // to some dropped spans. This ensures in most cases that it happens.
+            ::otel_tests::__reexport::opentelemetry::global::force_flush_tracer_provider();
+
+            if let Err(e) = caught {
+                panic!("{:#?}", e);
+            }
+        }
+    };
+
+    tokens.into()
+}
--- a/otel-tests/src/lib.rs
+++ b/otel-tests/src/lib.rs
@@ -1,0 +1,71 @@
+use opentelemetry::sdk::trace;
+use opentelemetry::sdk::Resource;
+use opentelemetry::KeyValue;
+use opentelemetry_otlp::WithExportConfig;
+use std::sync::Once;
+use tracing_subscriber::filter::LevelFilter;
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::Layer;
+use tracing_subscriber::Registry;
+
+#[doc(hidden)]
+pub mod __reexport {
+    pub use futures;
+    pub use opentelemetry;
+    pub use tokio;
+}
+
+pub use otel_tests_macro::otel_test;
+
+static INIT_OTLP_EXPORTER: Once = Once::new();
+
+pub fn init_tracing(module: &'static str) {
+    INIT_OTLP_EXPORTER.call_once(|| {
+        let env = std::env::var("ITCHYSATS_TEST_INSTRUMENTATION").unwrap_or_default();
+        let telemetry = if env == "1" {
+            let cfg = trace::Config::default()
+                .with_resource(Resource::new([KeyValue::new("service.name", module)]));
+
+            let tracer = opentelemetry_otlp::new_pipeline()
+                .tracing()
+                .with_trace_config(cfg)
+                .with_exporter(
+                    opentelemetry_otlp::new_exporter()
+                        .grpcio()
+                        .with_endpoint("localhost:4317"),
+                )
+                .install_simple()
+                .unwrap();
+
+            Some(
+                tracing_opentelemetry::layer()
+                    .with_tracer(tracer)
+                    .with_filter(LevelFilter::DEBUG),
+            )
+        } else {
+            None
+        };
+
+        let filter = EnvFilter::from_default_env()
+            // apply warning level globally
+            .add_directive(LevelFilter::WARN.into())
+            // log traces from test itself
+            .add_directive(format!("{module}=debug").parse().unwrap())
+            .add_directive("wire=trace".parse().unwrap())
+            .add_directive("taker=debug".parse().unwrap())
+            .add_directive("maker=debug".parse().unwrap())
+            .add_directive("daemon=debug".parse().unwrap())
+            .add_directive("xtra_libp2p=debug".parse().unwrap())
+            .add_directive("xtra_libp2p_offer=debug".parse().unwrap())
+            .add_directive("xtra_libp2p_ping=debug".parse().unwrap())
+            .add_directive("rocket=warn".parse().unwrap());
+
+        let fmt_layer = tracing_subscriber::fmt::layer()
+            .with_test_writer()
+            .with_filter(filter);
+
+        Registry::default().with(telemetry).with(fmt_layer).init();
+    })
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -133,4 +133,5 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install gcc-aarch64-linux-gnu
+          sudo apt-get install g++-aarch64-linux-gnu
       - run: cargo build --target=aarch64-unknown-linux-gnu --bins
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,6 +335,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "383d29d513d8764dcdc42ea295d979eb99c3c9f00607b3692cf68a431f7dca72"
 
 [[package]]
+name = "bindgen"
+version = "0.57.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd4865004a46a0aafb2a0a5eb19d3c9fc46ee5f063a6cfc605c69ac9ecf5263d"
+dependencies = [
+ "bitflags",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "peeking_take_while",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+]
+
+[[package]]
 name = "bit-vec"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -410,6 +429,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "boringssl-src"
+version = "0.3.0+688fc5c"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f901accdf830d2ea2f4e27f923a5e1125cd8b1a39ab578b9db1a42d578a6922b"
+dependencies = [
+ "cmake",
+]
+
+[[package]]
 name = "bs58"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -479,6 +507,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
+name = "cexpr"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4aedb84272dbe89af497cf81375129abda4fc0a9e7c5d317498c15cc30c0d27"
+dependencies = [
+ "nom 5.1.2",
+]
+
+[[package]]
 name = "cfg-if"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -527,6 +564,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7ee52072ec15386f770805afd189a01c8841be8696bed250fa2f13c4c0d6dfb7"
 dependencies = [
  "generic-array",
+]
+
+[[package]]
+name = "clang-sys"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a050e2153c5be08febd6734e29298e844fdb0fa21aeddd63b4eb7baa106c69b"
+dependencies = [
+ "glob",
+ "libc",
+ "libloading",
 ]
 
 [[package]]
@@ -788,9 +836,9 @@ dependencies = [
 
 [[package]]
 name = "ctor"
-version = "0.1.21"
+version = "0.1.22"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccc0a48a9b826acdf4028595adc9db92caea352f7af011a3034acd172a52a0aa"
+checksum = "f877be4f7c9f246b183111634f75baa039715e3f46ce860677d3b19a69fb229c"
 dependencies = [
  "quote",
  "syn",
@@ -905,6 +953,7 @@ dependencies = [
  "mockall",
  "mockall_derive",
  "model",
+ "otel-tests",
  "rand 0.6.5",
  "rust_decimal",
  "rust_decimal_macros",
@@ -913,7 +962,6 @@ dependencies = [
  "tokio",
  "tokio-extras",
  "tracing",
- "tracing-subscriber",
  "xtra",
  "xtra-bitmex-price-feed",
  "xtra-libp2p",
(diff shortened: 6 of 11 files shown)


Restioson:
Should I squash [Disable instrumentation by default for CI](https://github.com/itchysats/itchysats/pull/2404/commits/4c59ab56bc7c8222108562c1cbe5701904302579) into the first commit?

klochowicz:
> Should I squash [Disable instrumentation by default for CI](https://github.com/itchysats/itchysats/pull/2404/commits/4c59ab56bc7c8222108562c1cbe5701904302579) into the first commit?

I'm happy for it to be in a separate commit! if anything, I'd slightly adjust the commit message to say "Disable **test** instrumentation by default on CI"

Restioson:
> if anything, I'd slightly adjust the commit message to say "Disable test instrumentation by default on CI"

Cool, I will make sure to rename it when I rebase!

Another thing: it seems like the ARM build fails here due to use of gRPC-sys. I did try tonic, but it didn't actually work, so if we can get gprc working that'd be fantastic. The error is:

```
  CMake Error at CMakeLists.txt:9 (project):
    The CMAKE_CXX_COMPILER:
      aarch64-linux-gnu-g++
    is not a full path and was not found in the PATH.
    Tell CMake where to find the compiler by setting either the environment
    variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
    to the compiler, or to the compiler name if it is in the PATH.
```

Maybe g++ needs to be added to the path for the ARM run? I am not entirely familiar with this CI setup, though, so I'm not sure how to go about this

klochowicz:
> > if anything, I'd slightly adjust the commit message to say "Disable test instrumentation by default on CI"
> 
> Cool, I will make sure to rename it when I rebase!
> 
> Another thing: it seems like the ARM build fails here due to use of gRPC-sys. I did try tonic, but it didn't actually work, so if we can get gprc working that'd be fantastic. The error is:
> 
> ```
>   CMake Error at CMakeLists.txt:9 (project):
>     The CMAKE_CXX_COMPILER:
>       aarch64-linux-gnu-g++
>     is not a full path and was not found in the PATH.
>     Tell CMake where to find the compiler by setting either the environment
>     variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
>     to the compiler, or to the compiler name if it is in the PATH.
> ```
> 
> Maybe g++ needs to be added to the path for the ARM run? I am not entirely familiar with this CI setup, though, so I'm not sure how to go about this

I'll have a look into this!

klochowicz:
> > > if anything, I'd slightly adjust the commit message to say "Disable test instrumentation by default on CI"
> > 
> > 
> > Cool, I will make sure to rename it when I rebase!
> > Another thing: it seems like the ARM build fails here due to use of gRPC-sys. I did try tonic, but it didn't actually work, so if we can get gprc working that'd be fantastic. The error is:
> > ```
> >   CMake Error at CMakeLists.txt:9 (project):
> >     The CMAKE_CXX_COMPILER:
> >       aarch64-linux-gnu-g++
> >     is not a full path and was not found in the PATH.
> >     Tell CMake where to find the compiler by setting either the environment
> >     variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
> >     to the compiler, or to the compiler name if it is in the PATH.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe g++ needs to be added to the path for the ARM run? I am not entirely familiar with this CI setup, though, so I'm not sure how to go about this
> 
> I'll have a look into this!

Try rebasing on this: https://github.com/itchysats/itchysats/pull/2410
(you can incorporate your PR in yours, or I can merge mine first if this fixes your problem)

I just added C++ compiler; if it's installed, cmake should be enough to make it work. 
Note: it would be good to doublecheck with @bonomat whether we have access to C++ compiler on Raspiblitz, where we compile ItchySats too.

daemon-tests/Cargo.toml
```diff
@@ -9,10 +9,14 @@ anyhow = "1"
 async-trait = "0.1"
 bdk-ext = { path = "../bdk-ext" }
 daemon = { path = "../daemon" }
+futures = "0.3"
 maker = { path = "../maker" }
 mockall = "0.11"
 mockall_derive = "0.11"
 model = { path = "../model" }
+opentelemetry = { version = "0.17.0", features = ["rt-tokio"] }
+opentelemetry-otlp = { version = "0.10.0", features = ["grpc-sys", "trace"] }
```
klochowicz: it feels like one (or both), should really be inside `otel-tests` (you can't use `otel-tests` without it)

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+    let attrs = fn_item.attrs;
+
+    let test = if attrs.is_empty() {
+        quote!(#[tokio::test])
+    } else {
+        quote!()
+    };
+
+    let init_tracing = if attribute.is_empty() {
+        quote!(::daemon_tests::init_tracing)
```
klochowicz: this crate should not depend on `daemon-tests` (especially in such a sneaky way, when it compiles and then fails on usage!)

IMHO we should move `init_tracing()` inside this crate, and also consider making it public (if someone doesn't want / need the macro)

Alternatively, we could move the proc macro into daemon-tests; that said, we might want to sometimes instrument tests e.g. like recently, the load test from xtra-libp2p - so separate crate makes sense.

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+            #init_tracing();
+
+            let caught = {
+                futures::FutureExt::catch_unwind(std::panic::AssertUnwindSafe(tracing::Instrument::instrument(
+                    async #block, tracing::info_span!(stringify!(#name))
+                ))).await
+            };
+
+            #[allow(clippy::disallowed_methods)]
+            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
klochowicz: why do we need a sleep here?

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+            opentelemetry::global::force_flush_tracer_provider();
```
klochowicz: this seems to be the biggest value added from having the macro (besides putting test name into the span, which is also cool ☺️ ) - could you add a comment why is it needed?

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+
+            if let Err(e) = caught {
+                panic!("{:?}", e);
```
klochowicz: nit: I believe you can use short-hand notation:
```suggestion
                panic!("{e:?}");
```

also: wouldn't `e:#` be better here?

Restioson:
Thanks! I'll take a look in a sec.

This is a dev dependency so hopefully it wouldn't affect deployment on raspiblitz.

klochowicz (commented):
We might have the opportunity to make this crate more usable for other crates in the workspace if needed (we wanted to use this in xtra-libp2p recently already)

otel-tests/src/lib.rs
Restioson: The main issue is that we would need a way to pass the service name.

> Alternatively, we could move the proc macro into daemon-tests

This is not possible since last I checked you can't define a proc macro in a normal library crate - it must be in its own crate.

daemon-tests/Cargo.toml
Restioson: I can do this, but it'd require two crates:
- otel_tests: exports `otel_test` and `init_tracing(service_name: &str)`
- otel_tests_macro: defines the macro itself

This is because proc macros, last I checked, couldn't be defined in regular library crates. So, the macro itself needs a separate crate.

otel-tests/src/lib.rs
Restioson: If this is the last test that's running, the main thread might exit. Then, the opentelemetry exporter thread might not have exported all of its spans yet, leading to some dropped spans. This ensures in most cases that it happens.

Once I did see this not work, which was strange :/ but it's much harder to reproduce with this.

klochowicz: that makes sense, just add this paragraph as a comment, so it will be obvious for everyone later down the track :)

otel-tests/src/lib.rs
Restioson: I tried `panic!("{e:#?}");`, but it gave this error:
```
    = note: did you intend to capture a variable `e` from the surrounding scope?
    = note: to avoid ambiguity, `format_args!` cannot capture variables when the format string is expanded from a macro
    = note: this error originates in the attribute macro `otel_test` (in Nightly builds, run with -Z macro-backtrace for more info)
```

otel-tests/src/lib.rs
Restioson: Sure!

otel-tests/src/lib.rs
Restioson: I'm not entirely sure :( It makes it all work, and I theorise that it lets the otel thread receive the spans, and then when we flush they are there to be flushed. If it's removed, the flush doesn't as consistently reduce spans being dropped.

Restioson:
I've rebased on top of #2410

klochowicz:
> I've rebased on top of #2410

oh wow, we fixed the previous error, but there still seems to be some onerous error when cross-compiling to ARM. 
I'll spin this up locally and see whether this can be solved. you sure `tonic` can't be used? 😅

Restioson:
> you sure tonic can't be used? 

Tonic had panics and dropped spans last I checked, so I think so, unfortunately.

How important is instrumentation in ARM? If not so important, and the tests are just for CI, maybe the dep could be disabled? I'm not sure how this would work in a convenient fashion, though.

daemon-tests/Cargo.toml
Restioson: I've refactored into this model in the latest commit

otel-tests/src/lib.rs
klochowicz: oh wow, I've learned something new! to be fair, I've never tried this is a macro before ;)

daemon/src/seed.rs
```diff
@@ -97,6 +99,12 @@ impl Seed for RandomSeed {
     }
 }
 
+impl Debug for RandomSeed {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("RandomSeed").field(&"...").finish()
+    }
+}
```
luckysori, line 106: ❓ Should we just skip logging the seed? It's almost certainly not useful information and at the same time logging it can be a security issue.

bonomat:
> > > > if anything, I'd slightly adjust the commit message to say "Disable test instrumentation by default on CI"
> > > 
> > > 
> > > Cool, I will make sure to rename it when I rebase!
> > > Another thing: it seems like the ARM build fails here due to use of gRPC-sys. I did try tonic, but it didn't actually work, so if we can get gprc working that'd be fantastic. The error is:
> > > ```
> > >   CMake Error at CMakeLists.txt:9 (project):
> > >     The CMAKE_CXX_COMPILER:
> > >       aarch64-linux-gnu-g++
> > >     is not a full path and was not found in the PATH.
> > >     Tell CMake where to find the compiler by setting either the environment
> > >     variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
> > >     to the compiler, or to the compiler name if it is in the PATH.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Maybe g++ needs to be added to the path for the ARM run? I am not entirely familiar with this CI setup, though, so I'm not sure how to go about this
> > 
> > 
> > I'll have a look into this!
> 
> Try rebasing on this: #2410 (you can incorporate your PR in yours, or I can merge mine first if this fixes your problem)
> 
> I just added C++ compiler; if it's installed, cmake should be enough to make it work. Note: it would be good to doublecheck with @bonomat whether we have access to C++ compiler on Raspiblitz, where we compile ItchySats too.

seems like it was already installed :)

```
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'g++' instead of 'g++-aarch64-linux-gnu'
g++ is already the newest version (4:10.2.1-1).
g++ set to manually installed.
0 upgraded, 0 newly installed, 0 to remove and 88 not upgraded.
```

otel-tests-macro/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+    let attrs = fn_item.attrs;
+
+    let test = if attrs.is_empty() {
+        quote!(#[otel_tests::tokio::test])
+    } else {
+        quote!()
+    };
+
+    let init_tracing = if attribute.is_empty() {
+        quote!(::otel_tests::init_tracing)
```
luckysori: ❓ Is this a bit hacky or a standard way of defining proc macros? We seem to both depend on `otel-tests` here, but we are also a dependency of the `otel-tests` crate.

luckysori: I guess this is just a way to get around the fact that you need a separate crate for proc macros.

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+use tracing_subscriber::layer::SubscriberExt;
+use tracing_subscriber::util::SubscriberInitExt;
+use tracing_subscriber::EnvFilter;
+use tracing_subscriber::Layer;
+use tracing_subscriber::Registry;
+
+pub use futures;
+pub use opentelemetry;
+pub use otel_tests_macro::otel_test;
+pub use tokio;
```
luckysori: ❓ Might it be preferable to move these re-exported dependencies to the `otel-tests-macro` crate instead of re-exporting them? From the perspective of any crate that depends on `otel-tests` it doesn't make sense that it should re-export anything other than the `otel_test` macro.

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+            Some(tracing_opentelemetry::layer().with_tracer(tracer))
+        } else {
+            None
+        };
+
+        let filter = EnvFilter::from_default_env()
+            // apply warning level globally
+            .add_directive(LevelFilter::WARN.into())
+            // log traces from test itself
+            .add_directive("happy_path=debug".parse().unwrap())
```
luckysori: 🤔 I guess this needs to be configurable if we wanna see test logs for anything other than `happy_path` tests.

otel-tests/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+                "daemon-tests",
+            )]));
+
+            let tracer = opentelemetry_otlp::new_pipeline()
+                .tracing()
+                .with_trace_config(cfg)
+                .with_exporter(
+                    opentelemetry_otlp::new_exporter()
+                        .grpcio()
+                        .with_endpoint("localhost:4317"),
```
luckysori, line 37: ❓ Should/can this be configurable? Or is it okay to just hard-code it?

luckysori (commented):
Looks like a very convenient macro! Thanks for adding this.

daemon/src/seed.rs
Restioson: I think I added it to be able to derive Debug for something else, without having to reimplement everything in that debug impl. This debug doesn't actually print the seed - it always prints `RandomSeed(...)`

otel-tests/src/lib.rs
Restioson: IIRC depending on a crate transitively doesn't bring it into scope?

otel-tests/src/lib.rs
Restioson: Yea, this (and the service name) still needs to be addressed. I'm open to suggestions on how to achieve configuration of these 2 aspects.

otel-tests/src/lib.rs
Restioson: This is okay to hard code, I think, given it will be run locally with jaeger.sh

otel-tests-macro/src/lib.rs
Restioson: Indeed, that was the aim here

daemon/src/seed.rs
luckysori: Jesus, I'm blind. Sorry.

Restioson:
The current compile error is from `boringssl`, so if we can get rid of that dependency we can make this work I think? gRPC lets us use openSSL rather, but I wonder if opentelemetry-otlp lets us use this...

klochowicz:
> The current compile error is from `boringssl`, so if we can get rid of that dependency we can make this work I think? gRPC lets us use openSSL rather, but I wonder if opentelemetry-otlp lets us use this...

we can give this a shot, from the perspective of opentelemetry it shouldn't make a difference.
I found some instructions online how we could use openssl instead:
https://stackoverflow.com/questions/54775076/how-to-cross-compile-a-rust-project-with-openssl


Feel free to give it a shot, because it would be nice to have this dependency! (and it would be good to keep running daemon-tests on arm, as it's one of our target platforms (main one as for now!)

klochowicz:
> seems like it was already installed :)
> 
> ```
> Reading package lists... Done
> Building dependency tree... Done
> Reading state information... Done
> Note, selecting 'g++' instead of 'g++-aarch64-linux-gnu'
> g++ is already the newest version (4:10.2.1-1).
> g++ set to manually installed.
> 0 upgraded, 0 newly installed, 0 to remove and 88 not upgraded.
> ```


> > > > > if anything, I'd slightly adjust the commit message to say "Disable test instrumentation by default on CI"
> > > > 
> > > > 
> > > > Cool, I will make sure to rename it when I rebase!
> > > > Another thing: it seems like the ARM build fails here due to use of gRPC-sys. I did try tonic, but it didn't actually work, so if we can get gprc working that'd be fantastic. The error is:
> > > > ```
> > > >   CMake Error at CMakeLists.txt:9 (project):
> > > >     The CMAKE_CXX_COMPILER:
> > > >       aarch64-linux-gnu-g++
> > > >     is not a full path and was not found in the PATH.
> > > >     Tell CMake where to find the compiler by setting either the environment
> > > >     variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
> > > >     to the compiler, or to the compiler name if it is in the PATH.
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > Maybe g++ needs to be added to the path for the ARM run? I am not entirely familiar with this CI setup, though, so I'm not sure how to go about this
> > > 
> > > 
> > > I'll have a look into this!
> > 
> > 
> > Try rebasing on this: #2410 (you can incorporate your PR in yours, or I can merge mine first if this fixes your problem)
> > I just added C++ compiler; if it's installed, cmake should be enough to make it work. Note: it would be good to doublecheck with @bonomat whether we have access to C++ compiler on Raspiblitz, where we compile ItchySats too.
> 
> seems like it was already installed :)
> 
> ```
> Reading package lists... Done
> Building dependency tree... Done
> Reading state information... Done
> Note, selecting 'g++' instead of 'g++-aarch64-linux-gnu'
> g++ is already the newest version (4:10.2.1-1).
> g++ set to manually installed.
> 0 upgraded, 0 newly installed, 0 to remove and 88 not upgraded.
> ```

interesting, as the CI job wouldn't pass before we added that commit 😳 maybe it cached it after first run or updated paths? no se.

otel-tests/src/lib.rs
klochowicz: as for the test log, we could agree that we use a "test" (or any other name) log target  by convention in all crates using `otel-tests`.

as for service name, can we somehow retrieve module name? (macro noob here 😅 )

Restioson: Macros transform input tokens to output tokens - they are not more special than that. So, if we can get the module name in normal rust code, then yes. I think this might be possible through panic info and location APIs?

bonomat:
> interesting, as the CI job wouldn't pass before we added that commit 😳 maybe it cached it after first run or updated paths? no se.

sorry, I was not clear: the library is already installed on raspiblitz

otel-tests/src/lib.rs
klochowicz: `module_path!()` might be a good enough approximation

otel-tests/src/lib.rs
luckysori: It's definitely not a big deal. But I think what I'm expecting is `rust-analyzer` to, for example, offer me an `otel-tests::futures::Future` import when I try to auto-import `Future`. Just a minor annoyance if I'm understanding how this works.

otel-tests/src/lib.rs
luckysori: Well, in this case `happy_path` is the module name. If we can get the module name somehow we should be able to use it here to configure its log level.

otel-tests/src/lib.rs
Restioson: That's true. On the other hand, I don't want to force transitive dependencies to declare a whole lot of extra hidden dependencies.

Will RA still show up for `#[doc(hidden)]` stuff? Maybe we could just `#[doc(hidden)]` all of them...

Restioson:
Success in CI on ARM!! Woo!

otel-tests/src/lib.rs
luckysori: Haha, apparently we can hide re-exports that way because of this bug: https://github.com/rust-lang/rust/issues/59368.

But would that also break the macro?

Restioson: I think this just affects rustdoc, but I think RA might ignore doc hidden stuff too

luckysori: It does say

> Both of the re-exports above work just fine

in the issue I linked. I should read what I link to lol.

>  but I think RA might ignore doc hidden stuff too

Interesting! Can't hurt to try.

Restioson: That didn't seem to work for intellij - but it doesn't come first at least
image

otel-tests/src/lib.rs
Restioson: Done in latest commit

Restioson:
Last issue is just around the IDE experience of re-exporting stuff. I'm not sure if this can be helped, though

otel-tests/src/lib.rs
```diff
@@ -46,7 +52,7 @@ pub fn init_tracing() {
             // apply warning level globally
             .add_directive(LevelFilter::WARN.into())
             // log traces from test itself
-            .add_directive("happy_path=debug".parse().unwrap())
+            .add_directive(dbg!(format!("{module}=debug")).parse().unwrap())
```
klochowicz: I guess you only used `dbg!` for debugging :) (clippy would complain anyway before merging)

```suggestion
            .add_directive(format!("{module}=debug").parse().unwrap())
```

otel-tests/src/lib.rs
klochowicz: > That didn't seem to work for intellij - but it doesn't come first at least
> image

I think this is good enough!

klochowicz (approved):
LGTM! 

(commit history needs to be cleaned up a bit to satisfy the CI)

klochowicz:
> Last issue is just around the IDE experience of re-exporting stuff. I'm not sure if this can be helped, though

It's a net positive change even with that small awkwardness. IMHO as long as it doesn't come up first, and happens only in tests, it's not too bad.

otel-tests/src/lib.rs
Restioson: Good point! I missed this

Restioson:
I've squished this down into two commits - one adding g++ to CI, and one adding instrumentation.

luckysori (approved):
Thanks!

Restioson:
~~ARM build is failing again 😓~~ Edit: I messed it up during rebase 😅

PR #2405, opened 2022-07-12
Bump xtra to try to fix panic when using grafana agent

klochowicz:
When using grafana agent, we started encountering multiple panics in the logs.
This might be either due to a bug in a 3rd-party crate (tracing-subscriber)or us
misusing that crate.

changes in xtra:
   -  add more asserts
   -  removes redundant `parent` argument in debug_span! invocation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5371,7 +5371,7 @@ dependencies = [
 [[package]]
 name = "xtra"
 version = "0.6.0"
-source = "git+https://github.com/comit-network/xtra?branch=span-late-create#764260a53bab43a22c2c6b81ffd7e8e53aeb2d44"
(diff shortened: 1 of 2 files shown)


klochowicz:
@Restioson I opened the PR, but I guess we'll have to leave this for tomorrow, as this xtra bump made clippy unhappy...

The plan is to merge this and see whether panics on testnet still occur.

Restioson:
I concur, looks like I messed up something small in the xtra branch, but it doesn't look too hard to fix. I'll fix it tomorrow and we can bump again then, and retry on testnet.

The theory for why this might fix the panic is that it removes the explicit Span::current call, which I theorized to be a possible source of the error. I've also had issues passing parent spans by value instead of by reference. Removing the redundant parent arg solves both of these potential issues.

Restioson (approved):

Restioson:
Hopefully this fixes the panic...

Issue #907, opened 2021-12-16
Permit publication of punish transaction if counterparty publishes revoked commit transaction


thomaseizinger:
Now that we are confident that our rollover protocol is stable[1], we are ready to introduce the feature of punishing the counterparty if they publish a revoked (old) commit transaction. This is already supported by `maia`, so it's just a matter of adding some logic to `itchysats`.

The reason why we were hesitant to introduce this (very important) functionality was because we did not want to run into situations where the punished party didn't "mean" to publish the revoked commit transaction. The punishing party has no way of knowing what the intentions of their counterparty are, nor do they have a practical alternative to punishing a revoked commit transaction published by their counterparty. From this I conclude that **publishing a punish transaction should be handled automatically by both parties**.

In order to reduce the number of punishments, we can ensure that a party that knows that they are behind in the state of the CFD compared to their counterparty does not nonchalantly publish a commit transaction that would inevitably see them getting punished. Only takers can fall behind in rollovers according to our rollover protocol[2], so I propose to **prevent takers from force-closing[3] a CFD through the UI if their CFD is lagging behind**. This is not a perfect solution because:

- We would have to explain what's going on with a tooltip, making the UI more complex.
- The maker could theoretically hold the taker hostage. This would be fixed with https://github.com/itchysats/itchysats/issues/3019.

The taker could still be able to manually trigger such a transaction if they hit a particular endpoint. At that stage it would be on them to face the likely consequences (loss of funds).

How would a taker know that they are behind in the state of the CFD? We should model it explicitly. This is equivalent to what is outlined in https://github.com/itchysats/itchysats/issues/3019, so it would make sense to tackle that issue first before we proceed with this one.

[1]: We rarely run into a situation where one party falls behind in the state of the CFD because of a partially failed rollover. And when we do, we are able to recover by letting parties roll over from a specific commit TXID.
[2]: The maker always finishes first, so it's only the taker that can fail to receive the last bit of information that would allow them to finish the rollover protocol.
[3]: Force-closing is only available via the UI if the taker is not connected to the maker.

thomaseizinger:
Descoped because it is not too critical right now. Would require forking the software.

thomaseizinger:
We now have a real-world case where this feature would be needed: https://github.com/itchysats/itchysats/issues/1617

bonomat:
This project is unmaintained

PR #2408, opened 2022-07-13
Remove duplicated 'Revoked logic not implemented' error message

klochowicz:
This issue is tracked in the issue tracker:
https://github.com/itchysats/itchysats/issues/907

We became blind to this error message (even turned it off on our mainnet maker).
Meanwhile, testnet is spammed with hundreds of these error messages whenever we restart.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -296,7 +296,7 @@ impl Cfd {
             | CollaborativeSettlementFailed
             | CollaborativeSettlementProposalAccepted => self,
             RevokeConfirmed => {
-                tracing::error!("Revoked logic not implemented");
+                // TODO: Implement revoked logic
                 self
             }
         }
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -547,7 +547,7 @@ impl Cfd {
                 self.aggregated.state = CfdState::PendingCommit;
             }
             RevokeConfirmed => {
-                tracing::error!(order_id = %self.order_id, "Revoked logic not implemented");
+                // TODO: Implement revoked logic
                 self.aggregated.state = CfdState::OpenCommitted;
             }
             RolloverStarted { .. } => {


da-kami:
In production *new instances* of this problem are actually very important, because that means that our rollovers have a problem that we likely did not detect. Old instances should be removed, but it's hard to avoid them given how the system currently works.
I don't think we should remove this completely.

Ideas to overcome this for now:

1. We can log it in a different place in a smarter way, i.e. when we detect it in the monitor! Not sure that would help because we would re-trigger monitoring on startup...
2. We could add a way to remove specific CFDs from the database (at least on the maker side).
3. We could also treat these instances as a form of `closed_cfds` (because essentially they are....) and move them to the closed table for now and potentially mark them somehow. That way we would not run into the logs anymore. 

I tend to lean towards (3.) because I think these CFDs are actually closed at the moment, and should just not be treated as open. We could opt for creating a separate table for them as well if we feel that is cleaner.

klochowicz:
> In production _new instances_ of this problem are actually very important, because that means that our rollovers have a problem that we likely did not detect. Old instances should be removed, but it's hard to avoid them given how the system currently works. I don't think we should remove this completely, but maybe we can log it in a different place in a smarter way, i.e. when we detect it in the monitor! Not sure that would help because we would re-trigger monitoring on startup...
> 
> Alternatively we could add a way to remove specific CFDs from the database (at least on the maker side). We could also treat these instances as a form of `closed_cfds` (because essentially they are....) and move them to the closed table for now and potentially mark them somehow. That way we would not run into the logs anymore.

are you aware that we're not monitoring for them in production (we filter them out from 'errors' as they're too spammy)?

da-kami:
> are you aware that we're not monitoring for them in production (we filter them out from 'errors' as they're too spammy)?

Yes, but I don't think it is good ;)

klochowicz:
@da-kami I guess I can make this PR less radical by removing this spam from 2 places, and keeping it in cfd.rs - otherwise the same log happens 3 times for each CFD (as you rehydrate in multiple places).

Ideally I'd love to see us either treating this error seriously, or prioritising the underlying missing feature.

da-kami:
> @da-kami I guess I can make this PR less radical by removing this spam from 2 places, and keeping it in cfd.rs - otherwise the same log happens 3 times for each CFD (as you rehydrate in multiple places).
> 
> Ideally I'd love to see us either treating this error seriously, or prioritising the underlying missing feature.

Removing redundant logs is definitely a good thing independent of other solutions!

luckysori (commented):
I'm okay with:

1. Removing duplicate logs.
2. Pushing to actually implement this logic.

I suppose the problem with (2) is that we are not confident enough in rollover to do it, since we don't want to "wrongly" punish someone because of a faulty rollover.

bonomat:
> I suppose the problem with (2) is that we are not confident enough in rollover to do it, since we don't want to "wrongly" punish someone because of a faulty rollover.

yes, this is the main concern. We could nevertheless implement it and just don't follow up but print a warning saying: `would have punished xyz`...

klochowicz:
> > I suppose the problem with (2) is that we are not confident enough in rollover to do it, since we don't want to "wrongly" punish someone because of a faulty rollover.
> 
> yes, this is the main concern. We could nevertheless implement it and just don't follow up but print a warning saying: `would have punished xyz`...

I like the idea - this would be a good first step to gain some confidence in the solution by letting it run on testnet and mainnet first!

klochowicz:
@da-kami @luckysori @bonomat I've removed only the duplicates; happy to keep the last one, and potentially create a ticket to action at least initial implementation that just prints the suggestion instead of actually punishing other party.

bonomat (approved):

luckysori (approved):

PR #2411, opened 2022-07-13
Add test spans

Restioson:
On testnet, some spans in handle connection ready last for 3 minutes... this is strange, since there are no children lasting this long, and the handler prints that it's done!

This adds some extra spans just to try close in on this suspected trace time leak.

CC @klochowicz

### DIFF
--- a/maker/src/connection.rs
+++ b/maker/src/connection.rs
@@ -453,76 +453,81 @@ impl Actor {
             .send_async_safe(cfd::TakerConnected { id: identity })
             .await;
 
-        let mut tasks = Tasks::default();
-        tasks.add_fallible(
-            {
-                let this = this.clone();
+        tracing::debug_span!("Test span container").in_scope(|| {
+            let mut tasks = Tasks::default();
 
-                async move {
-                    loop {
-                        let msg = read
-                            .try_next()
-                            .await
-                            .context("Failed to read from socket")?
-                            .context("End of stream")?;
+            tracing::debug_span!("Add fallible").in_scope(|| {
+                tasks.add_fallible(
+                    {
+                        let this = this.clone();
 
-                        this.send(cfd::FromTaker {
-                            peer_id,
-                            taker_id: identity,
-                            msg,
-                        })
-                        .await
-                        .context(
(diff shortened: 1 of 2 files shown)


xtras/src/send_async_safe.rs
```diff
@@ -47,11 +47,11 @@ where
             return Err(xtra::Error::Disconnected);
         }
 
-        let send_fut = self.send(msg);
+        let rx = self.send(msg).split_receiver().await;
```
klochowicz, line 50: IMHO this is a very good change - we will send faster, and then only await the result for log purposes in a spawned task.

klochowicz (approved):
LGTM!

xtras/src/send_async_safe.rs
Restioson: I think the send won't necessarily be faster, but the reason I made the change in here was so that we can see the rest of the spans that are caused by it, thus bringing it in line with the other send_async_safe impls.

Of course, whether the handlers should be children of the span at all is its own discussion.

luckysori (approved):
LGTM. Could you move the changes to `send_async_safe` to a separate commit?

Restioson:
> LGTM. Could you move the changes to send_async_safe to a separate commit?

Done!

klochowicz (approved):

maker/src/connection.rs
```diff
@@ -453,76 +453,81 @@ impl Actor {
...
-                    let _ = this
-                        .send(ReadFail {
-                            taker_id: identity,
-                            error,
-                        })
-                        .await;
-                }
-            },
-        );
+        tracing::debug_span!("Test span container").in_scope(|| {
```
luckysori, line 456: @Restioson Could the reason why we see these long-lasting spans be that this task is an endless loop? I would expect it to never resolve (unless we encounter an error).

Restioson: Yes, I think so. However, the main issue at hand is that these tasks shouldn't extend the duration of the calling span at all. For some reason, they get attached as children and this is the issue.

PR #2412, opened 2022-07-14
Log taker connection info on startup

da-kami:
At the moment we don't log this, so the only way to retrieve the taker peer-id is the UI which can be annoying.

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -401,9 +401,14 @@ async fn main() -> Result<()> {
     let maker_libp2p_address = libp2p_socket_from_legacy_networking(first_maker_address);
     let maker_multiaddr = create_connect_tcp_multiaddr(&maker_libp2p_address, maker_peer_id)?;
 
+    let hex_pk = hex::encode(identities.identity_pk.to_bytes());
+    let peer_id = identities.libp2p.public().to_peer_id().to_string();
+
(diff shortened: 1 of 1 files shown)


bonomat (approved):

klochowicz (approved):

taker/src/main.rs
```diff
@@ -401,9 +401,14 @@ async fn main() -> Result<()> {
     let maker_libp2p_address = libp2p_socket_from_legacy_networking(first_maker_address);
     let maker_multiaddr = create_connect_tcp_multiaddr(&maker_libp2p_address, maker_peer_id)?;
 
+    let hex_pk = hex::encode(identities.identity_pk.to_bytes());
+    let peer_id = identities.libp2p.public().to_peer_id().to_string();
+
+    tracing::info!("Connection details: maker_id='{hex_pk}', peer_id='{peer_id}'");
```
luckysori: :wrench: Right?

```suggestion
    tracing::info!("Connection details: taker_id='{hex_pk}', peer_id='{peer_id}'");
```

luckysori (commented):
Not sure why I didn't submit my review yesterday. Isn't that field name wrong?

taker/src/main.rs
klochowicz: wow, indeed, good catch! 😌 sorry for almost merging without your review!

da-kami:
bors retry

PR #2415, opened 2022-07-14
Do not block `Endpoint` when opening substreams

### LINKED ISSUE #2414: Opening a substream may block the endpoint for significant time
This is an overall problem, but becomes severe on the maker when the maker opens substreams with takers that are not availble (for whatever reason). The endpoint handler for `OpenSubstream` then only times out after `20s`, blocking the complete endpoint while waiting.

Since the maker loops over all "connected" takers when opening substreams for `ping` and `offers`, and it can run into this problem with *some* takers, the blockage then might be so long that other takers lose connections because ping might not be handled.

luckysori:
Fixes #2414 (I think).

It is sufficient to just look for the corresponding `yamux::Control` for a particular peer inside `Endpoint` and then release the lock (by returning from the handler).

The actual opening of the substream using `yamux` can be delegated to a future which is returned to the caller, who must `await` it.

### DIFF
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -24,6 +24,7 @@ use multistream_select::Version;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::marker::PhantomData;
+use std::pin::Pin;
 use std::time::Duration;
 use thiserror::Error;
 use tokio_extras::Tasks;
@@ -265,28 +266,24 @@ impl Endpoint {
         self.notify_connection_dropped(*peer).await;
     }
 
-    #[instrument(skip(self), ret, err)]
+    #[instrument(skip(control), ret, err)]
     async fn open_substream(
-        &mut self,
+        mut control: yamux::Control,
         peer: PeerId,
         protocols: Vec<&'static str>,
+        connection_timeout: Duration,
     ) -> Result<(&'static str, Substream), Error> {
-        let (control, _) = self
-            .controls
-            .get_mut(&peer)
-            .ok_or(Error::NoConnection(peer))?;
-
         let stream = control
             .open_stream()
             .instrument(tracing::debug_span!("open yamux stream"))
             .await?;
 
         let (protocol, stream) = tokio_extras::time::timeout(
-            self.connection_timeout,
+            connection_timeout,
             multistream_select::dialer_select_proto(stream, protocols, Version::V1),
             |parent| tracing::debug_span!(parent: parent, "dialer_select_proto", version = ?Version::V1)
         )
-        .instrument(tracing::debug_span!("timeout dialing select proto", timeout_secs = self.connection_timeout.as_secs()))
+        .instrument(tracing::debug_span!("timeout dialing select proto", timeout_secs = connection_timeout.as_secs()))
         .await
         .map_err(|_timeout| Error::NegotiationTimeoutReached)?
         .map_err(Error::NegotiationFailed)?;
@@ -517,7 +514,13 @@ impl Endpoint {
         );
     }
 
-    async fn handle(&mut self, msg: OpenSubstream<Single>) -> Result<Substream, Error> {
+    #[must_use]
+    async fn handle(
+        &mut self,
+        msg: OpenSubstream<Single>,
+        _: &mut Context<Self>,
+    ) -> Result<Pin<Box<dyn futures::Future<Output = Result<Substream, Error>> + Send>>, Error>
+    {
         let peer = msg.peer;
         let protocols = msg.protocols;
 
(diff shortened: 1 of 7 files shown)


xtra-libp2p/src/endpoint.rs
```diff
@@ -526,26 +529,53 @@ impl Endpoint {
...
+            async move {
+                let (protocol, stream) =
+                    Self::open_substream(control, peer, protocols, connection_timeout).await?;
+
+                Ok((protocol, stream))
+            }
+        };
 
-        Ok((protocol, stream))
+        Ok(Box::pin(fut))
```
klochowicz, line 578: what's the upside of returning a `Box::pin(fut)` instead of oneshot receiver here?

Restioson, line 578: An alternate solution to returning the future would be to use `Context::join` (used to be `handle_while`), e.g:

```rust
ctx.join(self, fut).await // Context::select might be more suited here actually
```

This would let the endpoint handle other messages whilst it is waiting for the future to evaluate, so we could avoid blocking it whilst still returning the result itself rather than the future, which flattens the await chain.

I am not sure of the implications in terms of how this would look in the trace (how do the spans show up?), overflow behaviour (I think it should not overflow, as these are boxed?) in the case of nested `join`s, or overhead, but it might be a little cleaner. It's something that would need to be tested.

Restioson: I think it's slightly more direct. Returning a oneshot would look like

```rust
let (tx, rx) = oneshot();
tokio::spawn(fut.map(move |r| tx.send(r)));
Ok(rx)
```

This way has one less task spawned and the same await behaviour for the caller.

Restioson: If it nests too deeply, I wonder if `poll` might overflow the stack?

luckysori: I like the suggestion, but if we `await` the `join_future` _within_ the handler, isn't it equivalent to just awaiting the future itself? The way I see it, until we exit the handler (because the `join_future` has completed) we cannot process any other messages in the actor.

Restioson: `join` essentially handles other messages whilst inside of a handler itself and waiting for a given future to complete. In a way, it explicitly breaks the invariant that the actor handles one message at a time - you opt into this by using `join`. 

This invariant being broken here is probably okay, since it's right at the end of the function, so no intermediary state is ever exposed.

It might not be suited here, as it was intended to solve more limited, one-time use cases such as circular initialisation messages: https://github.com/Restioson/xtra/issues/13. The way it is used here would cause it to nest as many times as there are incoming protocol messages in the worst case of a 20s timeout being reached, possibly causing poll to overflow its stack. Therefore, it might be a bad fit here.

luckysori: Not sure. Just what first came to mind when we identified the problem.

It seems more direct to just return a future as the result of the handler, rather than having to spawn a task, pass in a sender and return a receiver. But I'm not sure what's better in terms of performance.

Restioson: I would expect only microseconds difference here, if any, and since this is a network operation it probably doesn't matter (?) as that will usually take milliseconds and more. On testnet, we've seen this operation take 900ms in the happy path

luckysori: Ahhh, thanks for the explanation. It's really cool that `join` was introduced nevertheless! Somehow I didn't understand from the documentation that it was meant to break that invariant. I think I got it confused with `scoped`.

Restioson: I should probably explicitly word it like that in the docs too!

klochowicz: it would be good to see a performance comparison, might run them both with the load test when they're here to see. 

this is likely going to be our main bottleneck as we scale (it is already :P) , so picking up faster is likely to make a difference. However, I have no idea which one will be faster without looking at some results.

Restioson: The only performance gain I could see is that if you spawn the task, you start it before the receiving task is woken, which could save a few microseconds, and then you get the finishing the future -> receive anyway to pay that spawn back. So, I don't think it should be faster, but a benchmark could be nice. I'm not sure it'd be measurable in release, though.

klochowicz: fair enough! so I guess `oneshot` would only make sense if the future would not be `Send`?

klochowicz: also: how does the memory footprint change? is pinning only passing already allocated memory on the heap, or are we doing new allocations?

Restioson: > fair enough! so I guess oneshot would only make sense if the future would not be Send?

I think it would need to be Send to spawn anyway

> also: how does the memory footprint change? is pinning only passing already allocated memory on the heap, or are we doing new allocations?

`Box::pin` does allocate new memory, yes, but I think that tokio might do this under the hood with `spawn` anyway? I'm not sure - maybe it's more optimised...

klochowicz (commented):
Thanks for implementing this, after looking at this closer, this would be easier to use than my similar solution with `oneshot` - oneshot requires one more `?` due to possibly failing to receive - this one is safe from this. 

When the dust settles and the load tests are in, I'd still like to do a performance comparison to satisfy my curiosity; but as mentioned in the comments, this is unlikely to make any difference - meanwhile, this solves a huge bottleneck.

klochowicz (approved):
👍

xtra-libp2p/src/endpoint.rs
```diff
@@ -526,26 +529,53 @@ impl Endpoint {
...
         );
 
-        let (protocol, stream) = self.open_substream(peer, protocols.clone()).await?;
+        let (control, _) = self.controls.get(&peer).ok_or(Error::NoConnection(peer))?;
 
-        debug_assert!(
-            protocol == protocols[0],
-            "If negotiation is successful, must have selected the only protocol we sent."
-        );
+        let fut = {
```
da-kami, line 534: Nice! Returning the future to the caller is a good one. And since we spin off tasks when looping over all connected peers this should solve this problem completely. 

Note: If we ever integrate a new network stack again or refactor on a bigger scale 😬😬 I would still try to thrive for a pattern where we have isolation per taker and not handle multiple takers in one actor (besides the low level connection handling in the endpoint).

da-kami (approved):

PR #2423, opened 2022-07-14
Identify compatibility check v2

### LINKED ISSUE #1769: Design proposal for `/itchysats/identify` protocol

### LINKED ISSUE #2307: Provide feedback about protocol between maker/taker incompatibility based on `identify` protocol

luckysori:
Fixes #1769.
Fixes #2307.

Hopefully the same stuff as in #2383, but with the listener protocol compatibility modelled differently.

I also addressed a couple of my own comments from the original PR.


daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,172 @@
...
+pub(crate) struct GetPeerInfo(pub PeerId);
+
+pub(crate) struct IdentifyMsgReceived {
+    peer_id: PeerId,
+    identify_msg: protocol::IdentifyMsg,
+}
+
+#[xtra_productivity]
+impl Actor {
+    async fn handle(&mut self, msg: GetPeerInfo) -> Option<PeerInfo> {
```
da-kami: We should be able to remove this - if I forgot to remove it on #2383 then that was by accident.
After the channel was introduced this is not used.

da-kami (commented):
I am happy to see a refactoring, but I am confused by this PR.

Are we expecting to close #2383 in favor of this? - it might be cleaner to create a PR on top of #2383 that showcases the changes. Like this it is super hard to know what changed.
Could you maybe change it so it is a PR on top of #2383 rather than replacing it? Otherwise I'm happy to talk this through together tomorrow.

luckysori:
This work has been force-pushed onto https://github.com/itchysats/itchysats/pull/2383 so that it can be reviewed.

klochowicz:
@luckysori it's super hard to compare the 2 approaches when it got force-pushed to the other PR - any chance it could have been added as a PR against @da-kami 's PR?

PR #2424, opened 2022-07-14
Cleanup timeout instrumentation

Restioson:
This ensures that the span is only created at first poll, as well as removing the redundant `parent` argument from the child span constructor function by just running it in the scope of the parent


klochowicz (approved):
thanks for the cleanup! 

I think we just need to rebase it on master before merging.

Restioson:
I'll do so now

Issue #2413, opened 2022-07-14
Exchange signed lock transaction took 30 seconds


bonomat:
I thought I create a ticket for this. Just experienced on master (d7c332d8ca1231dae5b4904151aae727366c3390) against testnet maker: 

between `verified all signatures` and `exchanged signed lock transaction` 30 seconds have passed. Imho that's too long to ignore.

```
2022-07-14 08:09:11  INFO daemon::setup_contract_deprecated: Exchanged setup parameters
2022-07-14 08:09:13  INFO daemon::setup_contract_deprecated: Created CFD transactions
2022-07-14 08:09:14  INFO daemon::setup_contract_deprecated: Exchanged CFD transactions
2022-07-14 08:09:17  INFO daemon::setup_contract_deprecated: Verified all signatures
2022-07-14 08:09:26  INFO Handle Ping:Handle OpenSubstream < Single >:open_substream: xtra_libp2p::endpoint: return=("/ipfs/ping/1.0.0", Substream { inner: Negotiated { state: Completed { io: Stream { id: 7, connection: d578cfda } } } }) peer=PeerId("12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1") protocols=["/ipfs/ping/1.0.0"]
2022-07-14 08:09:56  INFO Handle Ping:Handle OpenSubstream < Single >:open_substream: xtra_libp2p::endpoint: return=("/ipfs/ping/1.0.0", Substream { inner: Negotiated { state: Completed { io: Stream { id: 9, connection: d578cfda } } } }) peer=PeerId("12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1") protocols=["/ipfs/ping/1.0.0"]
2022-07-14 08:09:59  INFO daemon::setup_contract_deprecated: Exchanged signed lock transaction
2022-07-14 08:10:00 DEBUG Handle SetupSucceeded: sqlite_db: Applying new events to CFD order_id=a5646688-436a-4dd0-afee-a966466704f5 aggregate=model::cfd::Cfd cfd_version=0 num_events=1
```

Your Taker ID: 477ea119e067298765f47e078873f44581a49f4fac0f7b6fde39cf96bbc49329
Your Peer ID: 12D3KooWBgvtt61jkGDYeN4fwzP6cLUbLA7BiZzK7iuRMf9E9Zrb

Restioson:
[...]

PR #2425, opened 2022-07-14
Instrument contract setup

Restioson:
Instruments the contract setup - related to #2413.

## TODO
- [x] Better name for "Do work on CETs"
- [ ] ~~Replace some log entries with semantic spans~~

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -36,6 +36,8 @@ use std::collections::HashMap;
 use std::iter::FromIterator;
 use std::time::Duration;
 use tokio_extras::FutureExt;
+use tracing::instrument;
+use tracing::Instrument;
 use xtra::prelude::MessageChannel;
 
 /// How long contract setup protocol waits for the next message before giving up
@@ -48,6 +50,12 @@ const CONTRACT_SETUP_MSG_TIMEOUT: Duration = Duration::from_secs(120);
 /// Given an initial set of parameters, sets up the CFD contract with
 /// the counterparty.
 #[allow(clippy::too_many_arguments)]
+#[instrument(
+    name = "Setup contract"
+    skip_all,
+    err
+)]
+// TODO(restioson) semantic spans lining up with the logs here
 pub async fn new(
     mut sink: impl Sink<SetupMsg, Error = anyhow::Error> + Unpin,
     mut stream: impl Stream<Item = SetupMsg> + Unpin,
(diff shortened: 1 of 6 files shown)


Restioson:
Example with some spans collapsed:

image

Restioson:
@bonomat I'd be interested to see if you can reproduce #2413 under this PR and see what comes up? For me it mostly takes 10s

daemon/src/setup_contract.rs
```diff
@@ -315,11 +341,13 @@ pub async fn new(
             })
             .collect::<Result<HashMap<_, _>>>()
     }})
+    .instrument(tracing::debug_span!("Do some work on CETs")) // TODO(restioson) what is this doing?
```
bonomat: Congratulations, you found the potentially ugliest part of our code-base 😬

In here we are extracting the counterparty's adaptor signature from their message to us.

daemon/src/shared_protocol.rs
```diff
@@ -13,7 +13,10 @@ use maia_core::secp256k1_zkp::XOnlyPublicKey;
 use maia_core::PartyParams;
 use std::ops::RangeInclusive;
 use std::time::Duration;
+use tracing::instrument;
+use tracing::Instrument;
 
+#[instrument]
```
bonomat: If we only do `#[instrument]` it will include all the arguments right? I'm not sure if we need this information but at the very least we should exclude `own_cets` and `counterparty_cets`. That will be a significant amount of data.

bonomat:
First attempt :) I ran it in `release` mode to see if my slow execution came from this.

image

It took 27 seconds to receive the setup message.

bonomat:
second run even worse: 35 seconds for the setup message

image

bonomat:
@luckysori : I think we should get this in to see what's happening on the maker side, any thoughts?

daemon/src/setup_contract.rs
luckysori: 🔧 Perhaps

```suggestion
    .instrument(tracing::debug_span!("Add counterparty adaptor signatures to our CETs"))
```

luckysori (approved):
I like the idea of instrumenting this!

daemon/src/shared_protocol.rs
Restioson: Thanks, good point! I'll circlr back to the others too and make sure it isn't logging too much.

Restioson:
I think we can leave the last TODO for a followup. From my side, this is ready

bonomat:
@Restioson  : I rebased for you and will go ahead and merge this. 

bors r+

Restioson:
@bonomat thanks!

PR #2426, opened 2022-07-14
Add xtras::SendNext trait to schedule prioritised messages

klochowicz:
`xtras::SendAsyncSafe` by default does not take priority into account, so the
message dispatched by an actor will get scheduled on the back of the queue.

send_next() will dispatch messages with fixed priority
INTERNAL_MSG_PRIORITY (currently set to 100). This trait should be preferred
over send_async_safe() in cases where state changes as a result of the message
    handler.

Examples where send_next() was used in this changeset:
- Subscribers want to know about changes in the Endpoint before they handle a
    new external requests (because e.g. a connection got dropped in the meantime)
- Endpoint wants to process newly created/dropped connection as soon as it can,
    instead of scheduling it on the queue.

### DIFF
--- a/daemon/src/auto_rollover.rs
+++ b/daemon/src/auto_rollover.rs
@@ -12,7 +12,7 @@ use std::time::Duration;
 use time::OffsetDateTime;
 use xtra::Address;
 use xtra_productivity::xtra_productivity;
-use xtras::SendAsyncSafe;
+use xtras::SendAsyncNext;
 use xtras::SendInterval;
 
 pub struct Actor {
@@ -96,15 +96,13 @@ impl Actor {
 
             match cfd.can_auto_rollover_taker(OffsetDateTime::now_utc()) {
                 Ok((from_commit_txid, from_settlement_event_id)) => {
-                    // If we disconnect, we don't care.
-                    let _ = this
-                        .send_async_safe(Rollover {
-                            order_id: id,
-                            maker_peer_id,
-                            from_commit_txid,
-                            from_settlement_event_id,
-                        })
-                        .await;
+                    this.send_async_next(Rollover {
+                        order_id: id,
+                        maker_peer_id,
+                        from_commit_txid,
+                        from_settlement_event_id,
+                    })
+                    .await;
                 }
                 Err(reason) => {
                     tracing::trace!(order_id = %id, %reason, "CFD is not eligible for auto-rollover");
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -20,7 +20,7 @@ use rust_decimal::Decimal;
 use sqlite_db;
 use std::collections::HashMap;
 use xtra_productivity::xtra_productivity;
-use xtras::SendAsyncSafe;
+use xtras::SendAsyncNext;
 
 pub struct Actor {
     db: sqlite_db::Connection,
@@ -47,9 +47,7 @@ impl xtra::Actor for Actor {
 
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("we are alive");
-        this.send_async_safe(Initialize)
-            .await
-            .expect("we just started");
+        this.send_async_next(Initialize).await;
     }
 
     async fn stopped(self) -> Self::Stop {}
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -58,7 +58,7 @@ use tracing::info_span;
 use tracing::Instrument;
 use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
-use xtras::SendAsyncSafe;
+use xtras::SendAsyncNext;
 
 /// Store the latest state of `T` for display purposes
 /// (replaces previously stored values)
@@ -1115,7 +1115,7 @@ impl State {
 
 #[xtra_productivity]
 impl Actor {
-    async fn handle(&mut self, _: Initialize) -> Result<()> {
+    async fn handle(&mut self, _: Initialize) {
         let mut stream = self.db.load_all_cfds::<Cfd>(self.state.network);
 
         let mut cfds = HashMap::new();
@@ -1141,8 +1141,6 @@ impl Actor {
                 .expect("we initialized the state above; qed"),
             self.state.quote,
         );
-
-        Ok(())
     }
 
     async fn handle(&mut self, msg: CfdChanged) {
@@ -1189,9 +1187,7 @@ impl xtra::Actor for Actor {
     type Stop = ();
     async fn started(&mut self, ctx: &mut xtra::Context<Self>) {
         let this = ctx.address().expect("we just started");
-        this.send_async_safe(Initialize)
-            .await
-            .expect("we just started");
+        this.send_async_next(Initialize).await;
 
         tokio_extras::spawn(&this.clone(), {
             let price_feed = self.price_feed.clone();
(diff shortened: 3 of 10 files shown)


klochowicz:
a few notes for the reviewers:
- I opted for a new trait (unsure of the name though) to be able to quickly grep for the usages of prioritised messages, without needing to insert a magic number everywhere (could potentially use a public const in xtras though as well)
- I've left some cases of send_async_safe without priority as I wasn't sure whether to prioritise these messages,
- I opted to just log a warning in rare case of actor not being connected internally (most callsites were discarding the error anyway)
- I feel like this might need an ADR to codify this, but given the size of changeset, we might as well show a PoC and discuss it here.

xtras/src/send_next.rs
```diff
@@ -0,0 +1,51 @@
...
+use xtra::refcount::RefCounter;
+
+/// Arbitrarily hard number allowing flexibility of prioritising messages in actors
+const INTERNAL_MSG_PRIORITY: u32 = 100;
+
+#[async_trait]
+pub trait SendNext<M: Send + 'static, R> {
+    /// Dispatch a prioritised message to an actor without waiting for it to handle it.
+    // If the actor is not connected, it logs a warning.
+    async fn send_next(&self, msg: M);
```
bonomat: I don't really want to participate in a naming discussion but since you asked for it :D 
How about calling it `send_now`?

xtras/src/send_next.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[async_trait]
+impl<A, M> SendNext<M, ()> for xtra::Address<A>
+where
+    A: xtra::Handler<M, Return = ()>,
+    M: Send + 'static,
+{
+    #[instrument(skip_all)]
+    async fn send_next(&self, msg: M) {
```
bonomat: Since it's meant to be used as a prioritized message, why don't you return an error and let the caller decide?

xtras/src/send_next.rs
```diff
@@ -0,0 +1,51 @@
+use async_trait::async_trait;
+use tracing::instrument;
+use xtra::message_channel::MessageChannel;
+use xtra::refcount::RefCounter;
+
+/// Arbitrarily hard number allowing flexibility of prioritising messages in actors
```
luckysori: ❓ Did you mean high?

```suggestion
/// Arbitrarily high number allowing flexibility of prioritising messages in actors
```

xtras/src/send_next.rs
da-kami: I would probably call it `send_prioritized` - in case we opt for letting the caller decide the priority it could be `send_with_priority`. Don't have a strong opinion about this.

da-kami (approved):
Is there a (simple) way to test that this is working as expected? 
The actor tests pass, but they passed before as well. I feel it would be great to crate a load test that shows that this actually helps with avoiding congestion. (Not part of the scope of this PR, but wanted to mention it)

xtras/src/send_next.rs
luckysori: None of these names (including the original) tell me that we won't wait for the result. Which I think `send_async_safe` tried to do with the word `async`.

luckysori (commented):
I'm not against this change, but I think I would be more in favour if we (in order of importance):

1. Introduced each change in priority by commit, explaining why.
2. Used an internal constant per module. A single constant for all internal messages ever is almost certainly fine too. 

The second point means I'm not over the moon about the trait. I like that it's very descriptive if we say something like `send_now`, but the name quickly loses its meaning as soon as we decide that any message should have a higher priority and we have to set it manually. I would inline the trait, but it's not a deal breaker.

xtras/src/send_next.rs
klochowicz: based on the feedback above, I'll go with `send_async_next()` (because this encapsulates that we won't wait for result).

klochowicz:
> I'm not against this change, but I think I would be more in favour if we (in order of importance):
> 
> 1. Introduced each change in priority by commit, explaining why.
> 2. Used an internal constant per module. A single constant for all internal messages ever is almost certainly fine too.
> 
> The second point means I'm not over the moon about the trait. I like that it's very descriptive if we say something like `send_now`, but the name quickly loses its meaning as soon as we decide that any message should have a higher priority and we have to set it manually. I would inline the trait, but it's not a deal breaker.

I purposefully didn't want to allow arbitrary priority numbers, as this can of worms is hard to tame when it gets loose. Instead, I have chosen a high number, so we can manually opt-in to go below it, so it doesn't lose it's meaning.

As for 1), I tried to describe all the cases in the commit message. There are broadly two kinds:
- internal message that is scheduled right after so we don't block the handler, but it's clear that we should process it before next external request comes in,
- subscriber notifications. 

I will seek to split them out into distinct commits, so it's clearer. 

The idea is somewhat taken from UI frameworks, where events are often sorted by priority, where:
- you have fixed priorities for normal and priority messages (e.g. window update)
- you can create your events with priorities lower/higher than that if you really wish so, but the consts are global so you can calculate your custom priorities in relation to these constants.

I'll open an ADR where we can comment and see how to progress with it (for now I think at least the trait will make all the occurrences stand out of the crowd, making it easy to see/discuss them).

klochowicz:
> Is there a (simple) way to test that this is working as expected?
> The actor tests pass, but they passed before as well. I feel it would be great to crate a load test that shows that this actually helps with avoiding congestion. (Not part of the scope of this PR, but wanted to mention it)

if there was, then I'd already have added it. 
this changeset can be reasoned about, and I strongly believe that this is what we actually wanted (and the tests pass, so there's no regression).

luckysori:
> I purposefully didn't want to allow arbitrary priority numbers, as this can of worms is hard to tame when it gets loose. Instead, I have chosen a high number, so we can manually opt-in to go below it, so it doesn't lose it's meaning.

I think you mean that you're offering an API with smart defaults, hiding the complexity of the specific number used from the caller. From that perspective the trait is a good idea.

> I will seek to split them out into distinct commits, so it's clearer.

Thanks! I think that helps a lot if later on we need to understand why we did it.

> The idea is somewhat taken from UI frameworks, where events are often sorted by priority, where:
> 
>     * you have fixed priorities for normal and priority messages (e.g. window update)
> 
>     * you can create your events with priorities lower/higher than that if you really wish so, but the consts are global so you can calculate your custom priorities in relation to these constants.

I think it's good to use constants to set the priority. Even a global constant for all internal messages makes sense, as you can just go above and below if you ever need to. But the trait does hide this and without looking at the internals I might expect my message to always be processed... next? Haha.

> I'll open an ADR where we can comment and see how to progress with it (for now I think at least the trait will make all the occurrences stand out of the crowd, making it easy to see/discuss them).

We can certainly merge this as is. But it would be good to use an ADR to make a "final" decision, so that sounds good.

luckysori (approved):
Approved because I think we're on the same page.

xtras/src/send_next.rs
Restioson: That sounds like a good name. I was going to write my own comment about this but @luckysori expressed it perfectly

xtras/src/send_next.rs
klochowicz: IMHO there is no error strategy / decision that can really be made in such context - we never did anything other than logging it (if we remembered to do so :/ ).
also, in this case, the API with an error could falsely suggest that the message got handled if it returns `Ok()`, but it's not the case (it just means that the actor got it delivered on the inbox.

klochowicz:
@luckysori I split the commit, and it looks nicer. thanks for the push :)

PR #2427, opened 2022-07-14
Load tests that shows starving bobs

da-kami:
Adds a load test that simulates a similar setup as `offers` protocol, i.e. alice sends out a message to "all connected bobs".
We loop over all the bobs, send a message to all and then disconnect the current bob looped over.
This should result in the bobs getting disconnected one by one each receiving one more message.
For ten problems this is no problem, but note that we do see `ERROR` logs even with 10 bobs where Alice tries to an already disconnected bob.

With load the test panics because earlier bobs some bobs don't receive the message within 10 seconds because they get starved.

---

Note: This test does *not* reproduce the problem where the endpoint runs into a timeout when opening substreams as we see on `testnet` as described in https://github.com/itchysats/itchysats/issues/2414 - it might be worth writing another test that tries to reproduce this, but I was unable to achieve this so far. 
With the test proposed in this PR we should hopefully be able to validate that https://github.com/itchysats/itchysats/pull/2415 and https://github.com/itchysats/itchysats/pull/2426 help us overcome problems.

Note: This load test shows problems with our protocols where the maker opens a substream with all takers. This is potentially just not a good pattern. This explains, to some extent, why we only recently run into trouble, because we now have multiple protocols that follow this pattern, namely `ping`, `offers` and `identify`.
When running the test with `20s` timeout for Bob to receive the message it passes but takes an insane amount of time. Alice sends `300 * 300 - 300` (`300 + 299 + 298, ...`) = `89700` messages (and consequently opens as many substreams), which is a big amount. I let it run for a while and it felt like it was slowing down over time. I stopped it after about `30` minutes. With `10s` timeout it fails quite consistently within `40s`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,6 +335,25 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "383d29d513d8764dcdc42ea295d979eb99c3c9f00607b3692cf68a431f7dca72"
 
 [[package]]
+name = "bindgen"
+version = "0.57.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd4865004a46a0aafb2a0a5eb19d3c9fc46ee5f063a6cfc605c69ac9ecf5263d"
+dependencies = [
+ "bitflags",
+ "cexpr",
+ "clang-sys",
+ "lazy_static",
+ "lazycell",
+ "peeking_take_while",
+ "proc-macro2",
+ "quote",
+ "regex",
+ "rustc-hash",
+ "shlex",
+]
+
+[[package]]
 name = "bit-vec"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


da-kami:
hahaha, the tests actually pass on CI, that is not expected

Restioson:
Uh oh...

da-kami:
I thought about this test setup overnight and I am not that convinced of the value anymore :/ 
We put a lot of load, and we can proof that messages are piling up (even though it is not deterministic, because sometimes it passes), but e.g. prioritizing messages won't help to make this test pass more consistently. 
Delivering the `ConnectionDropped` message with high priority is important, and we should see less errors in the test logs, but the overall speed of the test will not be influenced by this. 
In order to properly proof that the fixes work we would need a test setup that reproduces https://github.com/itchysats/itchysats/issues/2414 - if you have an idea how to achieve this let me know. With a simple test with three Bobs where we first message two, then disconnect them and then handle a third one, I was unable to reproduce this. I guess the problem is related to load, or another issue that we don't know yet.

da-kami:
Closing for now. I think these test have some value, but more evaluation and adaption is needed so the tests are of use. At the moment the tests show *some* behavior, but we cannot draw a clear conclusion from it, so it is not super helpful.

@luckysori maybe we can combine our findings so far in a ticket to see if we want to some up with load tests that we keep in the codebase.

PR #2431, opened 2022-07-15
Prove that `Endpoint` is harder to clog up 

luckysori:
Unfortunately I could not write a mergeable test, but perhaps someone else has a better idea and can build on this.


luckysori:
No feedback and could not make it mergeable. Try again later?

PR #2432, opened 2022-07-15
Instrument spawned tasks as trace level follows_from

Restioson:
This should hopefully give better traceability on `level=trace`, but it is mostly to stop the spawned task from being included in the current one's time of execution.

This is a follow-up based on the information gained from #2411 - see the [following trace][1]:
image

The fact that the "Add fallible" span goes for so long tells me that the task is being included in it. By explicitly spawning the task as instrumented with a new span with no parent, this should fix the issue I think.

[1]: https://itchysats.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-itchysats-traces%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22queryType%22:%22traceId%22,%22minDuration%22:%225s%22,%22spanName%22:%22test%20span%20container%22,%22search%22:%22%22,%22query%22:%228a99ce72331d14128e98e9b3fef7a7b4%22%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D

### DIFF
--- a/tokio-extras/src/future_ext.rs
+++ b/tokio-extras/src/future_ext.rs
@@ -5,6 +5,7 @@ use futures::FutureExt as _;
 use std::any::Any;
 use std::any::TypeId;
 use std::time::Duration;
+use tracing::Instrument;
 use tracing::Span;
 
 pub trait FutureExt: Future + Sized {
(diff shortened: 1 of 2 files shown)


klochowicz (approved):
LGTM!

Restioson:
Can this be borsed or does it need another review? If so, who should I request?

klochowicz:
> Can this be borsed or does it need another review? If so, who should I request?

can be borsed, will do it now!

Restioson:
Thanks!

PR #2433, opened 2022-07-18
Log foreign peer ID when failing to upgrade connection

luckysori:
This helps verify if particular peers are systematically failing to connect to a listening `Endpoint`.


luckysori:
bors r+

It's just adding context.

PR #2434, opened 2022-07-18
Allow taker to pass in extend private key for wallet

bonomat:
This key will be used to derive the internal wallet according to (Bip84)[https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki].

This will allow us to easier start a taker with a wallet with funds.

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -1,3 +1,4 @@
+use crate::bitcoin::util::bip32::ExtendedPrivKey;
 use crate::routes::IdentityInfo;
 use anyhow::bail;
 use anyhow::Context;
@@ -115,6 +116,11 @@ struct Opts {
 
     #[clap(short, long, parse(try_from_str = parse_umbrel_seed))]
     umbrel_seed: Option<[u8; 32]>,
+
+    /// If provided will be used for internal wallet instead of a random key or umbrel_seed. The
+    /// keys will be derived according to Bip84.
+    #[clap(short, long)]
+    pub wallet_xprv: Option<ExtendedPrivKey>,
 }
 
 impl Opts {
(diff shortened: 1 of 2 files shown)


bonomat:
you can easily generate a xprv key with [wagyu](https://github.com/AleoHQ/wagyu): 

```
wagyu bitcoin hd -d "m/84'/0'/0'" -n testnet
```

taker/src/main.rs
```diff
@@ -346,6 +352,11 @@ async fn main() -> Result<()> {
         }
     };
 
+    let ext_priv_key = match opts.wallet_xprv {
+        Some(wallet_xprv) => wallet_xprv,
+        None => ext_priv_key,
+    };
```
luckysori, line 358: 🔧 I think it would be clearer if we had a

```rust
match (opts.umbrel_seed, opts.wallet_xprv) {
// handle combinations
}
```

handling all the cases explicitly.

And I would expect it to be mutually exclusive to the Umbrel seed. But I'm not sure how to express that in the CLI.

luckysori (approved):

taker/src/main.rs
bonomat: The way it is is the least invasive in regards to new lines of code 😬

I thought about the xor between `umbrel_seed` and `wallet_xprv`. The thing is, that the password is derived from the `umbrel_seed` unless provided.

da-kami (approved):

taker/src/main.rs
bonomat: Actually this is not an issue because we control umbrel setup 😃

bonomat:
I think this is fine as is and will go ahead and merge it. 

bors r+

PR #2437, opened 2022-07-18
Instrument wallet sync

Restioson:
Also add a span for monitor sync to disambiguate.

Example span from this PR: 
image

Unfortunately, it is not a whole lot better than not instrumenting this at all, since the main block of time is uninstrumented inside of bdk.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -415,6 +415,7 @@ impl Actor {
         }
     }
 
+    #[tracing::instrument("Sync monitor", skip_all, err)]
     async fn sync(&mut self) -> Result<()> {
         // Fetch the latest block for storing the height.
         // We do not act on this subscription after this call, as we cannot rely on
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -142,23 +142,29 @@ impl<DB> Actor<ElectrumBlockchain, DB>
 where
     DB: BatchDatabase,
 {
+    #[tracing::instrument(name = "Sync wallet", skip_all, err)]
     fn sync_internal(&mut self) -> Result<WalletInfo> {
         let now = Instant::now();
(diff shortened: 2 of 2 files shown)


da-kami (approved):
Note: I find it a bit hard to follow the code, given that different code parts are wrapped by spans. I don't think we can action this if we want to use spans, because I think that's the point of a span, i.e. instrument a part of code and that requires wrapping it.

Question for understanding: Why do we use three different spans here and not just one?

Restioson:
> Note: I find it a bit hard to follow the code, given that different code parts are wrapped by spans. I don't think we can action this if we want to use spans, because I think that's the point of a span, i.e. instrument a part of code and that requires wrapping it.

Ideally if we could change `bdk` we would add `#[instrument]` macros to the functions themselves, but otherwise we either need to
```rust
span.in_scope(|| /* some code */)
```
or
```rust
{
    let _g = span.entered();
    /* some code */
}
```
or, in the case of async
```rust
fut.instrument(span)
```

> Question for understanding: Why do we use three different spans here and not just one?

If we used one, we don't get the breakdown by operation - all of the time would count to the same span. Here, we have some time spent in syncing wallet with db, some in getting wallet balance, and some in collecting utxo values.

That being said, it seems like the overwhelming majority of the time is spent in syncing the wallet db - maybe we can get rid of the other two spans and just let them be unaccounted time in the parent span? This means we can't pick out exactly how much time each took, but if we expect them always to be fast it could be ok. I'm not sure if they will always be fast since I am not super familiar with the code here.

klochowicz (approved):
LGTM!

indeed, it is a bit harder to follow code when using a 3rd-party lib that doesn't support instrumentation; thankfully we don't have many places where we reach out to do it and there's no other way unless we contribute there :)

bonomat (approved):

PR #2439, opened 2022-07-18
Append network type to service.name in traces

klochowicz:
Allows easier differentiation between mainnet and testnet configs.

note: I've resisted the urge to refactor `Network` enum in this commit, but I'm doing it as a follow-up :)

### DIFF
--- a/maker/src/lib.rs
+++ b/maker/src/lib.rs
@@ -171,4 +171,14 @@ impl Network {
             Network::Regtest { withdraw, .. } => withdraw,
         }
     }
+
+    /// Stringified network kind
+    pub fn kind(&self) -> &str {
+        match self {
+            Network::Mainnet { .. } => "mainnet",
+            Network::Testnet { .. } => "testnet",
+            Network::Signet { .. } => "signet",
+            Network::Regtest { .. } => "regtest",
+        }
+    }
 }
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -30,13 +30,14 @@ use xtras::supervisor::Supervisor;
 #[rocket::main]
 async fn main() -> Result<()> {
     let opts = Opts::parse();
+    let service_name = "maker_".to_string() + opts.network.kind();
 
     logger::init(
         opts.log_level,
         opts.json,
         opts.instrumentation,
         opts.tokio_console,
-        "maker",
+        &service_name,
         &opts.collector_endpoint,
     )
     .context("initialize logger")?;
--- a/shared-bin/src/logger.rs
+++ b/shared-bin/src/logger.rs
@@ -26,7 +26,7 @@ pub fn init(
     json_format: bool,
     instrumentation: bool,
     use_tokio_console: bool,
-    service_name: &'static str,
+    service_name: &str,
     collector_endpoint: &str,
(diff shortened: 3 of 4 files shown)


klochowicz:
The alternative that I've considered (and which we might need to use anyway, in order to support both `taker_release` from `taker_master`) would be just adding an optional argument `--service-name`

This could make this PR redundant if we decided to specify service name in our configs fully, e.g. `--service-name=maker_release`. Thoughts?

luckysori (approved):
LGTM.

Can you remind me why we need to use different service names depending on the network? Is it just for Grafana to be able to tell them apart?

klochowicz:
> LGTM.
> 
> Can you remind me why we need to use different service names depending on the network? Is it just for Grafana to be able to tell them apart?

yes, exactly; other wise they'd get aggregated into the same one.

Restioson:
Do you think it's worth appending the commit hash for testnet and version number for mainnet too, or not really? If not to the service name, maybe another resource?

klochowicz:
> Do you think it's worth appending the commit hash for testnet and version number for mainnet too, or not really? If not to the service name, maybe another resource?

I like the continuity of the service name, so I wouldn't do it there. There's nothing stopping us for adding it as another key/value though, and that sounds like a great idea! we should have it accessible already, it's in the `daemon-version` (it appends git hash to the last released version of itchysats)

Restioson:
Cool! This should prevent the confusion that I had earlier about whether it was running on the updated version already or not :sweat_smile:

klochowicz:
> Cool! This should prevent the confusion that I had earlier about whether it was running on the updated version already or not 😅

I'll have a look into this as soon as I'll fix the ARM github action :)

klochowicz:
> Do you think it's worth appending the commit hash for testnet and version number for mainnet too, or not really? If not to the service name, maybe another resource?

`service.version` seems the most natural place to put that - would that work for you @Restioson ?

Restioson:
> service.version seems the most natural place to put that - would that work for you @Restioson ?

Is that a reserved name or made up for this purpose?

klochowicz:
> > service.version seems the most natural place to put that - would that work for you @Restioson ?
> 
> Is that a reserved name or made up for this purpose?

I found it mentioned somewhere in the docs as a valid example of naming: https://opentelemetry.io/docs/reference/specification/common/attribute-naming/

bonomat:
> The alternative that I've considered (and which we might need to use anyway, in order to support both `taker_release` from `taker_master`) would be just adding an optional argument `--service-name`
> 
> This could make this PR redundant if we decided to specify service name in our configs fully, e.g. `--service-name=maker_release`. Thoughts?

I think this would have been the better solution to our problem.

klochowicz:
> > The alternative that I've considered (and which we might need to use anyway, in order to support both `taker_release` from `taker_master`) would be just adding an optional argument `--service-name`
> > This could make this PR redundant if we decided to specify service name in our configs fully, e.g. `--service-name=maker_release`. Thoughts?
> 
> I think this would have been the better solution to our problem.

I'll proceed with this one then instead. I merged the above PR only so this wouldn't be a release blocker.

PR #2446, opened 2022-07-18
Document the #[otel_test] macro

Restioson:
This adds some documentation to `#[otel_test]`, explaining what it is for and about the environment variable that needs to be enabled.

### DIFF
--- a/otel-tests-macro/src/lib.rs
+++ b/otel-tests-macro/src/lib.rs
@@ -2,6 +2,11 @@ use proc_macro::TokenStream;
 use quote::quote;
 use syn::ItemFn;
 
+/// Like `#[tokio::test]`, but instrument the test with a span matching the test name, and export
+/// the span over OTLP to Jaeger for local debugging. To enable the span exporting, the
+/// `ITCHYSATS_TEST_INSTRUMENTATION` env var should be set to `1`. It is disabled by default for CI,
+/// as these tests may fail if OTLP exporting is enabled but Jaeger is not active, since this will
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thanks for the thorough explanation!

Completely orthogonal to this PR: could we allow using `tokio::test` attributes via `otel::test`? I tried `otel::test(flavor = "multi_thread")` but it didn't work.

Restioson:
> Completely orthogonal to this PR: could we allow using tokio::test attributes via otel::test? I tried otel::test(flavor = "multi_thread") but it didn't work.

This should actually work already now :) It's spelled differently though:
```rust
#[otel_test]
#[tokio::test(flavour = "multi_thread")]
```

Basically, if it sees there is already a `tokio::test` macro it will not emit its own. `#[otel_test]` should come above `tokio::test`, I think. This is also worth adding to the docs.

I have just realised it's actually untested, so let me quickly give it a shot...

luckysori:
> Basically, if it sees there is already a `tokio::test` macro it will not emit its own. `#[otel_test]` should come above `tokio::test`, I think.

Oh, cool! That's a good way of composing these macros.

Restioson:
It works but it also has an issue: see #2449

klochowicz:
> It works but it also has an issue: see #2449

have you confirmed that it works properly now? in other wors, is it safe to bors it? :)

Restioson:
This PR just adds docs, and is unrelated to the issue fixed in #2449. So yes, it's safe to bors :)

PR #2448, opened 2022-07-18
Add C++ compiler for ARM build in build release binary workflow

klochowicz:
We updated this on the CI, but not for here.

### DIFF
--- a/.github/actions/build-release-binary/action.yml
+++ b/.github/actions/build-release-binary/action.yml
@@ -28,6 +28,7 @@ runs:
       run: |
         sudo apt-get update
         sudo apt-get install 'gcc-aarch64-linux-gnu'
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thanks!

Restioson (commented):
I'm not 100% sure how the workflows interact and compose here, but this looks good I think :sweat_smile:

klochowicz:
> I'm not 100% sure how the workflows interact and compose here, but this looks good I think 😅

it's all described in these yaml files - they essentially setup a build env from scratch from `ubuntu:latest`, and for every place where we build arm binary we should set it up. technically we should be able to export that common think into shared action for both CI and release workflow, but this is updated like once a year 😁

PR #2450, opened 2022-07-18
Refactor contract setup into helper functions

Restioson:
This helps make the main body a bit more readable, as well as showing clearer sub-tasks in the trace. This replaces many "done {subtask}"-style logs with instrumented functions, which create tasks simply named {subtask}.

I've requested everyone's review here in case I accidentally broke something or named something poorly whilst refactoring, and it also touches a lot more code than my previous instrumentation-oriented refactors.

This only refactors the new contract setup, which is currently unused. If I am not mistaken, this would therefore affect #2292, so maybe it should be a PR into that branch instead? I could port the changes to the deprecated version too.

### DIFF
--- a/daemon/src/setup_contract.rs
+++ b/daemon/src/setup_contract.rs
@@ -1,3 +1,5 @@
+use crate::bitcoin::secp256k1::SecretKey;
+use crate::bitcoin::PublicKey;
 use crate::wallet;
 use crate::wire::Msg0;
 use crate::wire::Msg1;
@@ -7,20 +9,26 @@ use crate::wire::SetupMsg;
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
+use bdk::bitcoin::secp256k1::ecdsa::Signature;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::Amount;
+use bdk::bitcoin::Transaction;
+use bdk::miniscript::Descriptor;
 use bdk_ext::keypair;
 use futures::Sink;
 use futures::SinkExt;
 use futures::Stream;
 use futures::StreamExt;
 use maia::commit_descriptor;
-use maia::create_cfd_transactions;
 use maia::lock_descriptor;
+use maia_core::secp256k1_zkp::EcdsaAdaptorSignature;
 use maia_core::secp256k1_zkp::XOnlyPublicKey;
+use maia_core::Cets;
+use maia_core::CfdTransactions;
 use maia_core::PartyParams;
 use maia_core::PunishParams;
 use model::olivia;
+use model::olivia::BitMexPriceEventId;
 use model::shared_protocol::verify_adaptor_signature;
 use model::shared_protocol::verify_cets;
 use model::shared_protocol::verify_signature;
@@ -34,6 +42,7 @@ use model::SetupParams;
 use model::TransactionExt;
 use model::CET_TIMELOCK;
 use std::collections::HashMap;
+use std::ops::RangeInclusive;
 use std::time::Duration;
 use tokio_extras::FutureExt;
 use tracing::instrument;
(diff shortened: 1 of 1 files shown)


luckysori (commented):
Looks good! I think we should keep this open until we merge #2292, when this module will start being used. Then we can verify that this code works by looking a the actor tests.

da-kami (commented):
I am not sure this refactoring will make sense if we opt to change the protocol to be more deterministic in `sending` and `receiving` as we did for rollover. I think ensuring the message order is a good protocol re-design.

If one party first sends and the other first receives I would recommend refactoring the function into usable chunks to prepare the messages, but I would not include the message sending itself into the function.

See: https://github.com/itchysats/itchysats/pull/2248/files

We could still do this as in-between step but I'm not sure about the value then.

Restioson:
> If one party first sends and the other first receives I would recommend refactoring the function into usable chunks to prepare the messages, but I would not include the message sending itself into the function.

Should I give extracting the message send and receive from the helper functions a go?

Restioson:
@da-kami I've given this a shot in https://github.com/itchysats/itchysats/pull/2450/commits/e6757286a392eb666f6ae49f77c85b635473de1e; if this is the right solution it can always be squashed into the previous commit later.

~~I accidentally made it a `fixup!` but had "amend commit" checked; rewriting the message didn't seem to reflect remotely. Oh well.~~ EDIT: fixed

da-kami:
> @da-kami I've given this a shot in [e675728](https://github.com/itchysats/itchysats/commit/e6757286a392eb666f6ae49f77c85b635473de1e); if this is the right solution it can always be squashed into the previous commit later.
> 
> ~I accidentally made it a `fixup!` but had "amend commit" checked; rewriting the message didn't seem to reflect remotely. Oh well.~ EDIT: fixed

I think this is in line with the plan to make the message order more deterministic, thanks!
@luckysori what do you think about merging this work as preparation for https://github.com/itchysats/itchysats/issues/2563 ?

luckysori:
> @luckysori what do you think about merging this work as preparation for #2563 ?

I'm happy to merge this now that the code is being used and tested! But there are some conflicts to be resolved.

luckysori (approved):

Restioson:
I'll take a look at the conflicts tomorrow, thanks for letting me know!

Restioson:
I've rebased it on master now

daemon/src/setup_contract.rs
```diff
@@ -113,18 +99,204 @@ pub async fn new(
...
+#[instrument(name = "Generate own params", skip_all, err)]
+async fn own_setup_params(
+    build_party_params_channel: MessageChannel<wallet::BuildPartyParams, Result<PartyParams>>,
+    setup_params: SetupParams,
+) -> Result<(PartyParams, PunishParams, KeyPairs)> {
+    let key_pairs = KeyPairs {
+        identity: keypair::new(&mut rand::thread_rng()).into(),
+        revoke: keypair::new(&mut rand::thread_rng()).into(),
+        publish: keypair::new(&mut rand::thread_rng()).into(),
+    };
```
klochowicz, line 267: nit: this could be `KeyPairs` new()  or default() function instead? otherwise it's pretty hard to see the purpose of the abstraction.

daemon/src/setup_contract.rs
```diff
@@ -218,39 +388,42 @@ pub async fn new(
...
         lock_amount,
-        &msg1.commit,
+        commit_sig,
         &params.own_punish.publish_pk,
         &params.counterparty.identity_pk,
     )
     .context("Commit adaptor signature does not verify")?;
 
-    for own_grouped_cets in own_cets.clone() {
+    {
```
klochowicz, line 397: unneeded set of braces?

klochowicz (approved):
LGTM! (a couple of nits only)

daemon/src/setup_contract.rs
Restioson: This is to make sure that `verify_own` is dropped after the for loop finishes its last iteration

daemon/src/setup_contract.rs
Restioson: That's true. Currently it's only here to group everything together to avoid having to pass 3 args to anything that needs keys

daemon/src/setup_contract.rs
klochowicz: maybe we could add a comment to that extent, so it's not that confusing?

PR #2451, opened 2022-07-18
Share Network enum in both binaries

klochowicz:
Remove unnecessary duplication to ease maintenance

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -4,10 +4,7 @@ use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use clap::Parser;
-use clap::Subcommand;
 use daemon::bdk::bitcoin;
-use daemon::bdk::bitcoin::Address;
-use daemon::bdk::bitcoin::Amount;
 use daemon::bdk::FeeRate;
 use daemon::connection::connect;
 use daemon::libp2p_utils::create_connect_tcp_multiaddr;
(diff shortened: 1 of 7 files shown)


bonomat (approved):

shared-bin/src/network.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn default() -> Self {
+        Network::Mainnet {
+            electrum: MAINNET_ELECTRUM.to_string(),
+            withdraw: None,
+        }
+    }
+}
+
+#[derive(Subcommand, Clone)]
+pub enum Withdraw {
```
luckysori: 🔧 I think the withdraw subcommand doesn't belong in the `network` module. Perhaps we should create a `cli` module and put it there.

taker/src/main.rs
```diff
@@ -427,7 +306,7 @@ async fn main() -> Result<()> {
         Err(_) => Environment::Binary,
     };
 
-    let price_feed_network = network.price_feed_network();
+    let bitmex_network = network.bitmex_network();
```
luckysori, line 310: 🙃 Not worth changing now, but it's easier to review this PR if renaming this method happens in a separate commit.

luckysori (commented):
LGTM, but I would  move `Withdraw` to a different module.

taker/src/main.rs
klochowicz: will keep in mind next time!

shared-bin/src/network.rs
klochowicz: just to be clear: would renaming `network` to `cli` be acceptable?

luckysori: I think `Network` and `Withdraw` don't belong together.

klochowicz: well, `Withdraw` is *part* of `Network` right now, isn't it? 🤔

I will move `Withdraw` to a separate `withdraw` module for now as a first step towards separating them.

luckysori: Nvm, your proposal is better: renaming `network` to `cli`. I failed to notice that `Network` was actually a CLI subcommand. Sorry for the confusion!

PR #2454, opened 2022-07-19
Don't trace offers when sending to taker

bonomat:
This should remove quite some trace spam


luckysori (approved):
Maybe we can add a `trace_span` which adds the extra context for local debugging. But this is good to try now.

bonomat:
Let's see how much it reduces.

bors r+

PR #2455, opened 2022-07-19
Add spans with `OrderId` when broadcasting TXs

luckysori:
Otherwise it is very hard to associate transactions with CFDs.


bonomat (approved):
Looks good, thanks for the cleanups

PR #2456, opened 2022-07-19
Reduce traces a bit more

bonomat:


luckysori (approved):
I think it's good to try this.

We can always re-enable things later if we think they could be useful.

bonomat:
last force push was only a rebase

bors+

xtra-libp2p-ping/src/ping.rs
```diff
@@ -128,11 +127,7 @@ impl Actor {
                 tracing::warn!(%peer, "Outbound ping protocol failed: {e:#}")
             };
 
-            spawn_fallible(
-                &this,
-                ping_fut.instrument(tracing::debug_span!("Ping peer", %peer)),
```
Restioson, line 133: I think this will probably still result in spans being produced:
- for the second to the endpoint
- for the send_async_next

They just wouldn't be grouped together anymore. 

If the goal is to make them not be children of the current span, you could replace the `debug_span!` with `debug_span!(parent: Span::none(), "Ping peer", %peer)`, I think

klochowicz: if it doesn't remove the span, I'd consider removing this part of the changeset (the goal here was to remove some traces).

klochowicz (approved):
happy for it to go in (bors didn't notice it was meant to be borsed) - I would only action the comment made by @Restioson

xtra-libp2p-ping/src/ping.rs
bonomat: it did remove the additional `peer_id` though.

xtra-libp2p/src/endpoint.rs
```diff
@@ -267,13 +280,13 @@ impl Endpoint {
         self.notify_connection_dropped(*peer).await;
     }
 
-    #[instrument(skip(control), ret, err)]
+    #[instrument(skip(control, connection_timeout), ret(Display), err)]
```
Restioson: #2467 removes the ret - we can probably remove it here too

bonomat: Ok, I'll drop this commit and remove the ret

Restioson: Shall I close #2467 in favour of this?

Restioson (approved):

Issue #2453, opened 2022-07-19
Wallet might block contract setup


bonomat:
I have the suspicion that our wallet actor might be the source of blocked/delayed contract setup: 

The wallet actor syncs synchronously, meaning, it is blocked for the time it is syncing. This can be a significant amount of time (>40 seconds). In the meantime, the actor is not able to handle other messages. 
However, in the contract setup we need to ask this actor for information
- [BuildPartyParams](https://github.com/itchysats/itchysats/blob/5d87284fe3324ee82ca00ab7b30c9567f511230b/daemon/src/setup_contract.rs#L78)
- and [signing](https://github.com/itchysats/itchysats/blob/5d87284fe3324ee82ca00ab7b30c9567f511230b/daemon/src/setup_contract.rs#L255)

bonomat:
This is related: https://github.com/itchysats/itchysats/issues/947

bonomat:
I can confirm this: 
- manually hit wallet sync on maker
- open position on taker

see that nothing is happening until the maker sync is done.

Restioson:
We are blocking the tokio from the Wallet actor, which I think is probably one of the issues

bonomat:
let's hope that https://github.com/itchysats/itchysats/pull/2469 made it better

PR #2469, opened 2022-07-19
Spawn wallet actor as a blocking task

Restioson:
This is solution 2 to #2453. It spawns the wallet actor as a blocking task. I am not 100% sure if this will correctly make the thread park when nothing is being sent to the wallet actor.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5550,7 +5550,7 @@ dependencies = [
 [[package]]
 name = "xtra"
 version = "0.6.0"
-source = "git+https://github.com/Restioson/xtra?rev=a0c1ad6b6141e28417dadeaaf9595b405153f84e#a0c1ad6b6141e28417dadeaaf9595b405153f84e"
+source = "git+https://github.com/Restioson/xtra?rev=ff782ad379abad33c715173d21e2221189e252c0#ff782ad379abad33c715173d21e2221189e252c0"
 dependencies = [
  "async-trait",
  "catty",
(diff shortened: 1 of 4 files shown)


luckysori (approved):
Happy to try this!

Restioson:
I am currently preferring this solution over https://github.com/comit-network/xtra-productivity/pull/7 since it's easier to verify that everything works correctly. https://github.com/comit-network/xtra-productivity/pull/7 might be helpful in the future if there are actors which have a mix of sync and async actors, but the wallet only has sync handlers, so there the point is moot.

klochowicz (approved):
would you mind adding a bit more info into the commit description? it's a significant find, some context would be great :) thanks!

da-kami:
I just noticed that the `macos` tests were running for `15` minutes before finishing. That is quite some CI time 😅
I doubt that this is related to the PR.

Restioson:
> would you mind adding a bit more info into the commit description? it's a significant find, some context would be great :) thanks!

Sure!

Restioson:
I've realised that this has the side effect of no longer being added to the `Tasks` struct in the taker and maker main functions - this should be okay, though, since when the main thread exits it should kill the other thread too.

klochowicz:
@Restioson it seems clippy complains about some unused import now

Restioson:
Fixed too :)

klochowicz:
> I've realised that this has the side effect of no longer being added to the `Tasks` struct in the taker and maker main functions - this should be okay, though, since when the main thread exits it should kill the other thread too.

we never documented the purpose of `Tasks` struct in main.rs anywhere, but as far as I remember this was slightly different concern at play as clearly the tasks/threads will exit when the main thread does.

What we had to deal with here is to ensure we drop all the references to some resources (notably the DB) so that we could close it without errors. In wallet case, we don't use the DB, so it should not matter anyway.

PR #2470, opened 2022-07-19
Share env filter with log & trace

Restioson:
Adjusted some log levels for xtra crates too to make it work better for instrumentation.


klochowicz (approved):
let's give this a shot and compare how traces look vs how much effect it has on ingestion rate.

PR #2476, opened 2022-07-19
Disable span list from json logs by default

Restioson:
Added a new `--json-span-list` option to re-enable them. This should help reduce log size on disk.


klochowicz (approved):

klochowicz:
@Restioson I think you have to run `dprint fmt` locally before merging

Restioson:
Yea, I just forgot :sweat_smile: Force pushed now with formatting fixed

PR #2477, opened 2022-07-19
Fix panic when parsing vergen version

klochowicz:
Vergen version was used to determine need for heartbeats.
Whilst at there, move the version without the heartbeats to past 0.5

### DIFF
--- a/maker/src/connection.rs
+++ b/maker/src/connection.rs
@@ -39,13 +39,18 @@ use xtras::SendInterval;
 
 const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 
-fn parse_vergen_version(vergen_string: &str) -> semver::Version {
-    let vergen = semver::Version::parse(vergen_string).expect("daemon version to parse");
+fn parse_daemon_version(vergen_string: &str) -> Result<semver::Version> {
+    let vergen = semver::Version::parse(vergen_string)
+        .with_context(|| format!("Could not parse daemon_version: {vergen_string}"))?;
 
     // vergen appends git tag at the end, which is wrongfully parsed by
     // semver as a pre-release (it's in fact the opposite, the tag only
     // comes *after* release). Here we strip it out.
-    semver::Version::new(vergen.major, vergen.minor, vergen.patch)
+    Ok(semver::Version::new(
+        vergen.major,
+        vergen.minor,
+        vergen.patch,
+    ))
 }
 
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


maker/src/connection.rs
```diff
@@ -39,13 +39,18 @@ use xtras::SendInterval;
 
 const TCP_TIMEOUT: Duration = Duration::from_secs(10);
 
-fn parse_vergen_version(vergen_string: &str) -> semver::Version {
-    let vergen = semver::Version::parse(vergen_string).expect("daemon version to parse");
+fn parse_vergen_version(vergen_string: &str) -> Result<semver::Version> {
+    let vergen = semver::Version::parse(vergen_string)
+        .context("Could not parse daemon_version: {vergen_string}")?;
```
bonomat: Do you know what the version was when it paniced?

bonomat (approved):

maker/src/connection.rs
luckysori: 🔧 Ran into this myself recently, it's easy to miss:

```suggestion
        .with_context(|| format!("Could not parse daemon_version: {vergen_string}"))?;
```

Otherwise you will print the static string `"{vergen_string}"`.

luckysori: 🙃 

It's weird that the context talks about `daemon` here, when this is just a helper that maps string to `semver::Version`.

bonomat: haha, good find 😅

klochowicz: omg, that is indeed easy to miss!

klochowicz: maybe I'll rename the helper to `parse_daemon_version`? I could create a newtype, but I'm not sure whether this would help us much in this case!

klochowicz: unfortunately, it printed only the offending character, which was, if I remember correctly `<` (None that this would not have happened in production, as we only make weird tags where we append git hashes when building from `master`.

klochowicz: I have reworked the PR slightly to actually log the error when we encounter offending string, so we have more information now.

luckysori: It's very nitpicky (hence the 🙃), but it would make more sense to me to:

- Keep the helper as a general helper called `parse_vergen_version`.
- Remove the `context` and make the helper return a `Result<semver::Version, semver::Error>`.
- Add the `context` to line 503, where the helper is called (or alternatively to the trace on line 510).

luckysori (approved):

maker/src/connection.rs
klochowicz: I'm sorry, I completely missed this comment, only noticed now when going through my notifications! I'll put it on my list of small fixes.

luckysori: I think what you did is fine too! Just a tiny style thing almost.

PR #2478, opened 2022-07-19
Order `revoked_commit_transactions` by table id

da-kami:
Otherwise, the chronological order might not be insured.
We don't depend on the order at the moment, but it is more correct to ensure it.

Noticed this when driving by and thought I'll add it.

### DIFF
--- a/sqlite-db/sqlx-data.json
+++ b/sqlite-db/sqlx-data.json
@@ -1,16 +1,83 @@
 {
   "db": "SQLite",
   "060f5ccae5e675d3118ae357eb225c7fa37d54802a24e328db1dc84e547d4d9d": {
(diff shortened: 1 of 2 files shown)


bonomat (approved):

luckysori (approved):

bonomat:
~I guess this PR needs the second commit of this: https://github.com/itchysats/itchysats/pull/2503 ~

Wrong, you need to run it yourself on top of your commit :)

da-kami:
> ~I guess this PR needs the second commit of this: #2503 ~
> 
> Wrong, you need to run it yourself on top of your commit :)

I just forgot to run the db prepare on this one, that's why it was failing.

PR #2479, opened 2022-07-19
Add two more test spans to try isolate long add fallible span

Restioson:
I'm very stuck with this issue since it I can't reproduce locally at all, so I thought I'd add some more debug spans so we can see what it's doing better.

### DIFF
--- a/maker/src/connection.rs
+++ b/maker/src/connection.rs
@@ -31,6 +31,7 @@ use tokio_extras::FutureExt;
 use tokio_extras::Tasks;
 use tokio_util::codec::Framed;
 use tracing::Instrument;
+use tracing::Span;
 use xtra::message_channel::MessageChannel;
 use xtra_productivity::xtra_productivity;
 use xtras::address_map::NotConnected;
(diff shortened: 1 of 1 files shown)


maker/src/connection.rs
```diff
@@ -478,7 +480,7 @@ impl Actor {
                                         "we are not connected to ourselves, this should really not happen",
                                     )?;
                             }
-                        }
+                        }.instrument(tracing::debug_span!(parent: Span::none(), "Test long span"))
```
klochowicz:
```suggestion
                        }.instrument(tracing::debug_span!(parent: Span::none(), "Fallible task handler span"))
```

maker/src/connection.rs
```diff
@@ -490,7 +492,7 @@ impl Actor {
                                     error,
                                 })
                                 .await;
-                        }
+                        }.instrument(tracing::debug_span!(parent: Span::none(), "Test error span"))
```
klochowicz:
```suggestion
                        }.instrument(tracing::debug_span!(parent: Span::none(), "Fallible task error span"))
```

klochowicz (approved):
thanks! I've only suggested more permanent span names that should be unambiguous.

Restioson:
I've incorporated the suggestions and rebased this on top of master

PR #2482, opened 2022-07-20
Show order id in maker-ui and taker-ui

bonomat:
This can come handy for faster debugging

image

image

### DIFF
--- a/maker-frontend/src/components/cfdtables/CfdTable.tsx
+++ b/maker-frontend/src/components/cfdtables/CfdTable.tsx
@@ -95,7 +95,7 @@ export function CfdTable(
             },
             {
                 Header: "Details",
-                accessor: ({ details, expiry_timestamp }) => {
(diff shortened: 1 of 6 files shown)


luckysori (approved):
Maybe we could do the same for the taker UI when we're not building on release.

bonomat:
I think it can be specifically handy on a release build (our testnet is running on release builds). 

Hence, I've added a simple toggle: things are a bit jumpy but I think this is fine for this :)

image


image

klochowicz (approved):
I love how discrete and non-intrusive this additional info is on the taker side, even when enabled.

useful and tasteful change 🥇

bonomat:
> I love how discrete and non-intrusive this additional info is on the taker side, even when enabled.
> 
> useful and tasteful change 🥇

I take that ☺️

bors r+

PR #2486, opened 2022-07-20
Disable more of the otel_test macro according to env var

### LINKED ISSUE #2473: CI test run times can be really long
We had a CI `macos` test run that took `15` minutes to complete! 
I don't think it is expected that CI takes *that* long.

See: https://github.com/itchysats/itchysats/runs/7407394638?check_suite_focus=true on https://github.com/itchysats/itchysats/pull/2469

Restioson:
Maybe fixes #2473

### DIFF
--- a/otel-tests-macro/src/lib.rs
+++ b/otel-tests-macro/src/lib.rs
@@ -31,23 +31,29 @@ pub fn otel_test(_attribute: TokenStream, item: TokenStream) -> TokenStream {
         #sig {
             ::otel_tests::init_tracing(module_path!());
 
-            let caught = {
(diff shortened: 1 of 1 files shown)


bonomat (approved):
That second definitely added up. Thanks :)

bonomat:
As you can see from here: https://github.com/itchysats/itchysats/runs/7424358882?check_suite_focus=true

it's a bit faster: ~13 minutes

Restioson:
Hmm, that's still quite a bit slower than before...

klochowicz (approved):
thanks, at least it makes it a bit more bearable :)

klochowicz:
I'll go ahead and merge this, as it shaves a few minutes of each CI run.

Issue #2484, opened 2022-07-20
Log/Span load reduction


da-kami:
Evaluate what causes a lot of load; e.g. `verify encsig` creates a lot of load on the taker, and see to disable the instrumentation to specific code parts.

[This](https://itchysats.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-itchysats-traces%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22queryType%22:%22nativeSearch%22,%22limit%22:1000,%22search%22:%22%22,%22serviceName%22:%22maker_testnet%22%7D%5D,%22range%22:%7B%22from%22:%221658297351000%22,%22to%22:%221658297668000%22%7D%7D) query may help to evaluate what is creating load.

klochowicz:
I'd be for removing (or adding a target that's disabled by default) everything from this file:

https://github.com/itchysats/itchysats/blob/master/daemon/src/shared_protocol.rs#L20

Restioson:
Three initial targets:
- ~~verify_cet_encsig~~ EDIT: functions from shared_protocol.rs such as `verify_cet_encsig`
- successful pings
- successful taker offers

klochowicz:
> Three initial targets:
> 
> * verify_cet_encsig
> * successful pings
> * successful taker offers

as per your first point, see my comment above - all of the instrumented futures from there are not of interest for us (can be either disabled by target or completely :) )

Restioson:
I think this might be fixed with #2519, but it's worth keeping an eye on in Grafana

Restioson:
I would tentatively say that this is fixed, as long as we disable `--verbose-spans`. I would propose we test this on mainnet next update/restart, as the current projected usage will be 110GB, but earlier it was 80GB only. I suspect a lot of this is the ping peer spans which would be disabled with `--verbose-spans` off.

PR #2487, opened 2022-07-20
Remove rets & set some spammy functions to trace level

Restioson:
WIP for now until I've addressed all of the three current targets of #2484

### DIFF
--- a/daemon/src/shared_protocol.rs
+++ b/daemon/src/shared_protocol.rs
@@ -14,9 +14,9 @@ use maia_core::PartyParams;
 use std::ops::RangeInclusive;
 use std::time::Duration;
 use tracing::instrument;
-use tracing::Instrument;
+use tracing::Span;
 
-#[instrument(skip_all)]
+#[instrument(target = "verify_crypto", skip_all)]
 pub(crate) async fn verify_cets(
     (oracle_pk, nonce_pks): (XOnlyPublicKey, Vec<XOnlyPublicKey>),
     counterparty: PartyParams,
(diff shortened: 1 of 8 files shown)


klochowicz:
@Restioson could we merge this in stages? all 3 points in the issue linked are orthogonal, and it actually is easier to measure the effect if they're merged separately?

daemon/src/shared_protocol.rs
```diff
@@ -50,13 +52,12 @@ pub(crate) async fn verify_cets(
...
         anyhow::Ok(())
     })
-    .in_current_span()
     .await??;
 
     Ok(())
 }
 
-#[instrument(skip_all)]
+#[instrument(target = "verify_crypto", level = "trace", skip_all)]
```
klochowicz, line 60: IMHO I would probably call them like module names, `shared_protocol` here, and `contract_setup` in the other case (no need to append the `_deprecated` there.

I'm not too fussed though, I'm happy as long as they're not printed out by default :)

klochowicz (approved):

daemon/src/shared_protocol.rs
Restioson: > contract_setup in the other case (no need to append the _deprecated there.

How come? They do the same thing, don't they, so why different names?

klochowicz: fair enough. I was mostly pointing out that `shared_protocol` would not work for both, and this seemed to be the name liked by @luckysori most. As said above, I'm not too fussed; `verify_crypto` is a decent name too 👍

luckysori: I think `verify_crypto` is also fine!

luckysori (approved):

PR #2492, opened 2022-07-20
Define transparent Debug implementation for PeerId newtype

### LINKED ISSUE #2485: Use consistent format for `peer_id` tag
Currently, we use two different formats for the peer id tag: `PeerId("<id>")` and just `<id>`. We should correct them to all use `<id>`

klochowicz:
This makes it appear consistently in tracing, as sometimes we return the
underlying libp2p type, and sometimes the newtype.

### DIFF
--- a/model/src/libp2p.rs
+++ b/model/src/libp2p.rs
@@ -3,8 +3,14 @@ use serde::Serialize;
 use std::fmt;
 use std::str::FromStr;
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


bonomat (approved):

Restioson (approved):

Restioson:
Note that there are still some places where we log the peer ID as part of a format string and not as a tag; those should be corrected to tags in a later PR.

klochowicz:
> Note that there are still some places where we log the peer ID as part of a format string and not as a tag; those should be corrected to tags in a later PR.

thanks for reminder, will hunt for them too!

PR #2493, opened 2022-07-20
Ignore frequent spans

Restioson:
Using a Directive in an EnvFilter didn't seem to do anything at all so I landed on this. This was very painful as reordering anything at all in the registry init lead to completely broken output.

 It is still not perfect as it will ignore every span, even if it has an error. I am opening this for now since I am stumped on how to solve this large issue and I've been trying to solve it for a few hours to no avail. This means this change is a bit far in the *other* direction - even erroring send offers, ping, and send heartbeat spans will be ignored. This will still reduce the load and log the error, but the error will not be attached to any span anymore.

As part of the implementation, I added a new `quiet: bool` field to `send_interval`, which is currently only `true` for `SendHeartbeat`. An alternative is a new `send_interval_quiet`, but I thought that it could be good to force the caller to consider whether to be quiet or noisy. For most situation, noisy will be fine, I think.

### DIFF
--- a/shared-bin/src/logger.rs
+++ b/shared-bin/src/logger.rs
@@ -6,10 +6,12 @@ use opentelemetry::sdk::Resource;
 use opentelemetry::KeyValue;
 use opentelemetry_otlp::WithExportConfig;
 use time::macros::format_description;
+use tracing::Level;
+use tracing_subscriber::filter::Directive;
+use tracing_subscriber::filter::FilterExt;
 use tracing_subscriber::fmt::time::UtcTime;
 use tracing_subscriber::EnvFilter;
 use tracing_subscriber::Layer;
-use tracing_subscriber::Registry;
 
 pub use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::layer::SubscriberExt;
@@ -20,13 +22,20 @@ pub const LOCAL_COLLECTOR_ENDPOINT: &str = "http://localhost:4317";
 
 const RUST_LOG_ENV: &str = "RUST_LOG";
 
-#[allow(clippy::print_stdout)] // because the logger is only initialized at the end of this function but we want to print a warning
+/// Names for specific spans which may be disabled as they are generally too spammy
+pub mod span_names {
+    pub const OLD_BROADCAST_OFFERS: &str = "Broadcast offers to taker";
+}
+
+// because the logger is only initialized at the end of this function but we want to print a warning
+#[allow(clippy::print_stdout, clippy::too_many_arguments)]
 pub fn init(
     level: LevelFilter,
     json_format: bool,
     json_span_list: bool,
     instrumentation: bool,
     use_tokio_console: bool,
+    verbose_spans: bool,
     service_name: &str,
     collector_endpoint: &str,
 ) -> Result<()> {
@@ -36,27 +45,32 @@ pub fn init(
 
     let is_terminal = atty::is(atty::Stream::Stderr);
 
-    let filter = match std::env::var_os(RUST_LOG_ENV).map(|s| s.into_string()) {
-        Some(Ok(env)) => {
-            let mut filter = log_base_directives(EnvFilter::new(""))?;
-            for directive in env.split(',') {
-                match directive.parse() {
-                    Ok(d) => filter = filter.add_directive(d),
-                    Err(e) => println!("WARN ignoring log directive: `{directive}`: {e}"),
-                };
+    let filter = || -> Result<EnvFilter> {
+        let filter = match std::env::var_os(RUST_LOG_ENV).map(|s| s.into_string()) {
+            Some(Ok(env)) => {
+                let mut filter = log_base_directives(EnvFilter::new(""))?;
+                for directive in env.split(',') {
+                    match directive.parse() {
+                        Ok(d) => filter = filter.add_directive(d),
+                        Err(e) => println!("WARN ignoring log directive: `{directive}`: {e}"),
+                    };
+                }
+                filter
             }
-            filter
-        }
-        _ => log_base_directives(EnvFilter::from_env(RUST_LOG_ENV))?,
-    };
-    let filter = filter.add_directive(format!("{level}").parse()?);
+            _ => log_base_directives(EnvFilter::from_env(RUST_LOG_ENV))?,
+        };
 
-    let filter = if use_tokio_console {
-        filter
-            .add_directive("tokio=trace".parse()?)
-            .add_directive("runtime=trace".parse()?)
-    } else {
-        filter
+        let filter = filter.add_directive(format!("{level}").parse()?);
+
+        let filter = if use_tokio_console {
+            filter
+                .add_directive("tokio=trace".parse()?)
+                .add_directive("runtime=trace".parse()?)
+        } else {
+            filter
+        };
+
+        Ok(filter)
     };
 
     let fmt_layer = tracing_subscriber::fmt::layer()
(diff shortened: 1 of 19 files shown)


xtras/src/send_interval.rs
```diff
@@ -14,7 +14,9 @@ where
     /// handler returns before scheduling a new one, thus preventing them from
     /// piling up.
     /// As a bonus, this function is non-fallible.
-    async fn send_interval<F>(self, duration: Duration, constructor: F)
+    ///
+    /// If quiet is set to `true`, the span and logs will only be shown if an error occurs.
+    async fn send_interval<F>(self, duration: Duration, constructor: F, quiet: bool)
```
klochowicz: instead of bool, consider using an enum for readability, e.g.

```rust
enum IncludeSpan {
Always, OnError`
}
```

shared-bin/src/logger.rs
```diff
@@ -108,9 +115,21 @@ pub fn init(
...
     };
 
-    Registry::default()
-        .with(Layer::and_then(telemetry, fmt_layer).with_filter(filter))
+    let disable_spans = tracing_subscriber::filter::filter_fn(|meta| {
+        const ERROR_ONLY: &[&str] = &[
+            "Ping peer",
+            "Broadcast offers to taker",
+            "Broadcast offers to taker (libp2p)",
+            "Send message every interval (quiet)",
```
klochowicz: I'd consider using `const`s to avoid issues when anyone decides to update this string - it's extremely fragile. As a side effect, we'll be able to jump to the reference of the span and see what's actually disabled.

klochowicz:
is there any reason to still be verbose when sending the heartbeats? 

I like the explicit policy choice (but with an enum that states the policy). 

zooming out: I'm not sure whether we should disable spans on offers and pings at this stage (maybe when we stabilize libp2p? or allow increasing that verbosity from the command line, so we don't need to redeploy a version if there are some issues lurking there and we'd like to debug it further? 🤔

Restioson:
> is there any reason to still be verbose when sending the heartbeats?

The argument is the opposite: `true` means that it will be quiet, not verbose.

> but with an enum that states the policy

I considered this and am happy to change it to that for clarity!

> zooming out: I'm not sure whether we should disable spans on offers and pings at this stage (maybe when we stabilize libp2p? or allow increasing that verbosity from the command line, so we don't need to redeploy a version if there are some issues lurking there and we'd like to debug it further? thinking

We can leave them enabled, but maybe have a way to disable them on mainnet with a flag? I'd reckon that these represent the overwhelming bulk of traces in the app currently.

Restioson:
Addressed the two concerns you had @klochowicz

shared-bin/src/logger.rs
```diff
@@ -121,6 +144,8 @@ pub fn init(
 
 fn log_base_directives(env: EnvFilter) -> Result<EnvFilter> {
     let filter = env
+        .add_directive(Directive::from(LevelFilter::INFO))
+        // .add_directive("[Send message every interval]=off".parse()?)
```
klochowicz: was this meant to be commented out?

xtras/src/send_interval.rs
```diff
@@ -14,27 +14,42 @@ where
...
         A: xtra::Handler<M, Return = ()>;
 }
 
+/// How verbose a given trace will be. If it is set to quiet, it will be disabled, alongside all
+/// of its children.
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum TraceVerbosity {
+    Quiet,
+    Normal,
+}
```
klochowicz: we're getting there, but maybe we could iterate on this name a bit more.

What do you think of the following?
````
pub enum VerboseTraces {
Enabled,
Disabled
}
```

xtras/src/send_interval.rs
```diff
@@ -14,27 +14,42 @@ where
...
+    async fn send_interval<F>(self, duration: Duration, constructor: F, verbosity: TraceVerbosity)
     where
         F: Send + Sync + Fn() -> M,
     {
         let span = || {
-            tracing::debug_span!(
-                "Send message every interval",
-                interval_secs = %duration.as_secs()
-            )
+            if verbosity == TraceVerbosity::Quiet {
```
klochowicz: prefer `match` rather than if on enums. this would wreak havoc if we add one more option ;D

taker/src/main.rs
```diff
@@ -92,6 +92,11 @@ struct Opts {
     #[clap(long)]
     pub tokio_console: bool,
 
+    /// If enabled, libp2p ping and offer broadcast spans will be included in the traces exported
+    /// by the application.
+    #[clap(long)]
+    pub noisy_libp2p_spans: bool,
```
klochowicz: I would keep the name a bit more general, so that we can use this toggle with other verbose traces if needed

```suggestion
    pub verbose_spans: bool,
```

shared-bin/src/logger.rs
```diff
@@ -113,9 +122,23 @@ pub fn init(
         None
     };
 
-    Registry::default()
-        .with(Layer::and_then(telemetry, fmt_layer).with_filter(filter))
+    let mut error_only = vec![
+        "Broadcast offers to taker",
```
da-kami: ❓ Does this mean if we change the log message this will not be filtered anymore?

If so we might want to put the log message in a constant and use it in the filter here and when we log to tie it together. I don't especially like that, but it would at least tie the message to the filter.

klochowicz: these strings are still really fragile - if you want to depend on particular string to disable spans, export it as `pub const` from the other crate. (applies to all span names you want to use here)

shared-bin/src/logger.rs
```diff
@@ -126,6 +149,8 @@ pub fn init(
 
 fn log_base_directives(env: EnvFilter) -> Result<EnvFilter> {
     let filter = env
+        .add_directive(Directive::from(LevelFilter::INFO))
+        // .add_directive("[Send message every interval]=off".parse()?)
```
luckysori: 🔧 Can remove commented out line.

xtras/src/send_interval.rs
```diff
@@ -14,27 +14,42 @@ where
...
+                tracing::debug_span!(
+                    "Send message every interval (quiet)",
```
da-kami: I don't really understand this span description. What does sending quietly mean? I thought it just won't be sent.

klochowicz (commented):
getting there, if we need to release I'm happy if comments are actioned afterwards except the one with the flag name, as this will be de-facto public API :)

da-kami (approved):

shared-bin/src/logger.rs
```diff
@@ -113,9 +122,23 @@ pub fn init(
...
-    Registry::default()
-        .with(Layer::and_then(telemetry, fmt_layer).with_filter(filter))
+    let mut error_only = vec![
+        "Broadcast offers to taker",
+        "Send message every interval (quiet)",
+    ];
+
+    if !noisy_libp2p_spans {
+        error_only.extend(["Ping peer", "Broadcast offers to taker (libp2p)"]);
+    }
```
luckysori: ❓ Do these strings need to match the events/spans they correspond to for this to work? I guess so. This is a bit brittle as soon as we make any changes to the originals :(

luckysori (approved):
I can't say I know exactly what's going on in `logger.rs`, but I'm happy to give it a go.

shared-bin/src/logger.rs
klochowicz: exactly, I suggested the same in my review - it ain't pretty ,but at least we're explicit that we're depending on that information from a different crate.

shared-bin/src/logger.rs
Restioson: I'm not sure if spans can be named with expressions (I think they might need to be literals) but I can see if it will work

shared-bin/src/logger.rs
Restioson: Ah, it was meant to be removed entirely. Good catch!

shared-bin/src/logger.rs
Restioson: If we change the span name, yes. This does not apply to log messages

shared-bin/src/logger.rs
klochowicz: I was aiming at something like the following (and then using it in both places). I believe the following is the same for the compiler:

`pub const PING_PEER_SPAN_NAME: &'static str = "Ping peer"

xtras/src/send_interval.rs
Restioson: This means that it and all of its children won't be exported over OTEL

shared-bin/src/logger.rs
Restioson: Yea, this seems to work. I thought that it had to be a literal in the macro argument, though

xtras/src/send_interval.rs
Restioson: I've gone to the `IncludeSpan` name you suggested earlier

shared-bin/src/logger.rs
Restioson: Done now, but please review where I kept the span names - not sure of it

klochowicz (approved):
thank you for addressing the comments!

Restioson:
Force pushed now - should work

PR #2494, opened 2022-07-20
Maker decides revoke commit parameters based on proposed `commit-txid`

### LINKED ISSUE #2475: Counterparty sent invalid revocation sk
Observed in the `testnet` `maker` logs

Relevant log: `Counterparty sent invalid revocation sk`

Relevant CFDs where this happened:

 - `97a4920c-4848-4f63-a58a-1f2918ff46dc` - `taker_id` = `25d1efa9fd521d5b8f1908ed105dec7fe158f3d913204bc91e7fff276a53b221` - `peer_id` = `12D3KooWFZYotR8yFLdP28QX3MqD117bbvgJ4uC6m5iJ258Pq3Zx`
  - `a07e8148-80df-400b-aea9-696603818432` - `taker_id` = `a82cfcf840677c43a955127a4579f14c52056c0ec1b917f5e415a9fbbd5b4034` - `peer_id` = `12D3KooWFkkdjJY5aijC9QJUeyHocekErd45YuhWavoU9MoDYLvk`
  - `d8ceb1b8-5566-44b4-a519-2533a616d562` = `taker_id` = `7d3a0d444d63ce4f160ab542c41baeecdd1f54ff2e40c44346bc481d96cb4f38` - `peer_id` = `12D3KooWGy1XddxM5JsnkNma9zfNRedqUBcrnWUK62MPi2QCwATC`
  - `e0a45941-3a66-42fa-9f9d-936de64c14fe` = `taker_id` = `7d3a0d444d63ce4f160ab542c41baeecdd1f54ff2e40c44346bc481d96cb4f38` - `peer_id` = `12D3KooWGy1XddxM5JsnkNma9zfNRedqUBcrnWUK62MPi2QCwATC`

Behavior:

Analysis for `97a4920c-4848-4f63-a58a-1f2918ff46dc`: 

- Last successful rollover was at `2022-07-19 00:01:29`.
- After that the `taker` tried `Starting rollover from previous commit-txid`, namely commit-txid `8519f8ff8acb71fbd4922a2b654478d8bc84076cf39818745a261ebadf5a7f57`. 
- It appears that the maker and taker somehow disagree on the state for rolling over from `commit-txid``8519f8ff8acb71fbd4922a2b654478d8bc84076cf39818745a261ebadf5a7f57` because we always run into `Counterparty sent invalid revocation sk` on the `maker`.
- This likely points to a problem when rolling over from a previous `commit-txid`, because the parties disagree on the state

da-kami:
Fixes https://github.com/itchysats/itchysats/issues/2475

We reproduce the behavior in a test:

On `testnet`, when the maker is ahead with rollovers we run into error `Counterparty sent invalid revocation sk` when rolling over from a previous commit.
This was not detected by the `retry_rollover_an_open_cfd` test because, even though we trigger rolling over from a previous commit-txid, the taker's DLC actually matched the one of the maker. The test setup did no properly simulate that the maker is in fact one rollover ahead of the taker.

As it is hard to remove events, the easiest way to simulate the previous rollover on the taker side, is to add another `RolloverCompleted` event with the old data. This should be sufficient to simulate that the taker has in fact the old state and the maker is ahead.

---

For the fix the maker decides the revocation parameters based on the `commit-txid`, taking them either from the current DLC or from the revoked commits.


daemon-tests/src/lib.rs
```diff
@@ -462,6 +485,30 @@ impl Taker {
             .await
             .unwrap();
     }
+
+    /// Appends an event that overwrites the current DLC
+    ///
+    /// Note that the projection does not get updated, this change only manipulates the database!
+    /// When triggering another rollover this data will be loaded and used.
+    pub async fn simulate_previous_rollover(
```
klochowicz, line 493: `simulate_one_rollover_ahead` might be a better name?

klochowicz (commented):
kudos for reproducing the problem in a test!

da-kami:
Note to self: Squash before merging :)

bonomat (approved):
Good find 👏👏

daemon-tests/src/lib.rs
```diff
@@ -462,6 +485,30 @@ impl Taker {
...
+        dlc: Dlc,
+        complete_fee: CompleteFee,
+    ) {
+        self.db
+            .append_event(CfdEvent::new(
+                id,
+                EventKind::RolloverCompleted {
+                    dlc: Some(dlc),
+                    // Funding fee irrelevant because only CompleteFee is used
+                    funding_fee: dummy_funding_fee(),
```
luckysori, line 507: ❓ Interesting, but why do we have a `funding_fee` field if it's irrelevant? Are we planning to remove this from this event?

daemon-tests/src/lib.rs
luckysori: I'm not sure about `simulate_one_rollover_ahead` because that depends on the `Dlc` that you pass in as an argument. The current name might suggest that we're overwriting the previous rollover, but we're actually just adding a new one. An alternative could be the straightforward `append_rollover`.

It's a nifty trick that we can just _append_ a rollover completed event and make it seem like the taker is actually _behind_. But we really want to eventually be able to run a test like this based on a state that is truly behind i.e. rollover failed instead of rollover completed.

daemon-tests/tests/happy_path.rs
```diff
@@ -523,6 +525,16 @@ async fn retry_rollover_an_open_cfd() {
         "The maker and taker should have the same commit_txid after the first rollover"
```
luckysori: 🔧 I would remove all these asserts on the taker's first rollover success because:

- They're irrelevant for the test. We actually want to immediately "forget" this rollover and pretend it never worked.
- They make the test harder to understand.

luckysori (approved):
Great job! I'll make some minor tweaks and get this merged so that we can try to release sooner.

daemon-tests/tests/happy_path.rs
da-kami: 👍

daemon-tests/src/lib.rs
da-kami: Yeah, I'm not proud of the nifty trick, I just wanted to make sure to reproduce it and see the test turn green. 
In the long run I think it'd be better to test this by removing data rather than adding, but that's currently not easily possible.

daemon-tests/tests/happy_path.rs
```diff
@@ -506,36 +508,29 @@ async fn retry_rollover_an_open_cfd() {
...
+        .await;
 
-    // We mock the initial oracle event-id for the second rollover again
+    // We mock the initial oracle event-id for the rollover retry again
     mock_oracle_announcements(
         &mut maker,
         &mut taker,
         contract_setup_oracle_data_announcement.clone(),
     )
     .await;
```
luckysori, line 527: I would like to leave a better comment here as to why this is necessary. I feel like it's just a quirk of the test, but I might be missing something! cc @da-kami

model/src/rollover.rs
```diff
@@ -84,3 +99,167 @@ impl RolloverParams {
...
+            .clone();
+
+        let settlement_event_id = settlement_event_id.context("Missing settlement event id")?;
+        let complete_fee = complete_fee.context("Missing complete fee")?;
+        let revocation_sk_ours = revocation_sk_ours.context("Missing own revocation sk")?;
+
+        let revoked_commits = self
+            .revoked_commit
+            .iter()
+            .take_while(|commit| commit.txid != revoked_commit_txid)
```
bonomat, line 205: 🔧
Is it guaranteed that `revoked_commit` is sorted?
It's implied with `take_while` but we could get side effects if this ever changes.

daemon-tests/src/lib.rs
klochowicz: `append_rollover` sounds good too!

model/src/rollover.rs
klochowicz: indeed, perhaps a simple `filter` might be better here as it doesn't need such assumption.

klochowicz (approved):

model/src/rollover.rs
da-kami: Once https://github.com/itchysats/itchysats/pull/2478 merges it should be guaranteed. I will see to fix and merge that one.

daemon-tests/src/lib.rs
luckysori: I think it's great that you found a clever workaround. Otherwise we wouldn't have a test!

daemon-tests/tests/happy_path.rs
da-kami: To some extent it is a quirk, because we only have two events that we currently use in the test, i.e. we use `exmaple_0` for contract setup, then `example_1` for the first rollover, and `example_0` for the second one. I originally opted for this so I can assert that the value changes, i.e. the mocked event is correctly picked when rolling over from a previous commit. But re-using `example_0` is indeed confusing and should be changed.

model/src/rollover.rs
luckysori: It's a good point, but I don't see any reason why it shouldn't be sorted. We just append to it during rollover.

It would be nice if could just call `sort_by` on the iterator, but we need the `settlement_id`, which is optional for backwards compatibility.

PR #2495, opened 2022-07-20
Open browser with maker UI automatically

klochowicz:
There's no reason why we should constrain this behaviour to the taker only,
especially when it's useful when debugging/testing.


bonomat:
I honestly get annoyed with this behavior and would be in favor of removing it all together 😬

My setup for the taker is: 
```
cargo run --bin taker -- --password=test
```

that way I always know the taker's password. I usually have one tab open for the taker and for the maker and this random opening browsers is just bothering me.

klochowicz:
> I honestly get annoyed with this behavior and would be in favor of removing it all together 😬
> 
> My setup for the taker is:
> 
> ```
> cargo run --bin taker -- --password=test
> ```
> 
> that way I always know the taker's password. I usually have one tab open for the taker and for the maker and this random opening browsers is just bothering me.

thanks for sharing your opinion - maybe we could add a flag that can disable this behaviour on both sides? (AFAICT there were other reasons why we wanted it on the taker by default - so that's easier for the users).

luckysori (approved):
I'm okay with this!

da-kami (commented):
I would rarely need this, because I rarely start the maker locally, and when running remotely this won't help.

I agree with @bonomat that this behavior can be annoying and I think a flag to disable automatically opening the UI would be useful for test setups for maker and taker.
I don't really have a strong opinion on merging this, feel free to go ahead :)

klochowicz:
I'll go ahead with this one then, and add a possibility to start daemon headless on both sides in a follow-up.

Issue #2464, opened 2022-07-19
Update internal metrics to deal with new contract symbols


bonomat:
- [ ] - collect counterparty's margin per trading pair and direction: `position_margin_counterparty_satoshis`
- [ ] - ~cummulative volume should be cummulative counterparty margin~  redundant to the first metrics
- [ ] - open positions per trading pair: `positions_number_total`
- [ ] - closed positions per trading pair `positions_number_total`

bonomat:
The metrics exists. We only have to check if they are dynamically generated over the `contract_symbols`

PR #2501, opened 2022-07-21
Add trading symbol to position metrics

bonomat:
it's prepared to dea with multiple trading pairs. The idea is to sort the map of CFDs by trading pair first and then pass down to `update_position_metrics`.

prepares https://github.com/itchysats/itchysats/issues/2464

### DIFF
--- a/daemon/src/position_metrics.rs
+++ b/daemon/src/position_metrics.rs
@@ -1,3 +1,4 @@
+use crate::position_metrics::metrics::BTCUSD_LABEL;
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
@@ -73,6 +74,8 @@ impl Actor {
         self.state.cfds = Some(cfds);
         metrics::update_position_metrics(
             self.state.cfds.clone().expect("We've initialized it above"),
+            // TODO: once we have multiple trading pairs, this needs to come from the positions
+            BTCUSD_LABEL,
         );
     }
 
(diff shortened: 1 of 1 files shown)


da-kami (approved):

daemon/src/position_metrics.rs
```diff
@@ -377,30 +382,35 @@ mod metrics {
     const STATUS_REJECTED_LABEL: &str = "rejected";
     const STATUS_REFUNDED_LABEL: &str = "refunded";
 
+    const SYMBOL_LABEL: &str = "symbol";
+    pub const BTCUSD_LABEL: &str = "BTCUSD";
```
klochowicz, line 386: maybe for consistency with broader financial system we could use `XBTUSD`?

klochowicz:
I'm not that familiar with metrics, but I'm not sure whether this is the best practice.
wouldn't this add a lot of data as this string is now appended to all of these passed metrics? how would this affect our metrics ingestion rate? 

I guess the alternative would be to expose a metric per trading pair? (maybe in a hashmap?)

bonomat:
Long answer: 

It is best practice but it will indeed increase the number of metrics, i.e. we will have a metric for each trading pair. 
The number of `symbols` should be small (2 to be precise), so it does not grow too much. 

Short answer: 
There is no alternative :D 

Fun fact: 
```
metric_1{symbol='btcusd'}
metric_1{symbol='ethusd'}
```
has the same count as: 
```
metric_1_btc_usd
metric_1_eth_usd
```

The former one is the one to prefer as it would allow you to easily do aggregates over all `symbol`s.

daemon/src/position_metrics.rs
bonomat: I think it's pretty random if exchange use `xbtusd` or `btcusd`

PR #2503, opened 2022-07-21
Make tmp db file absolute

bonomat:
It seems that relative path-based execution leads to errors from time to time where the db file can't be found. Using an absolute path solves this problem. The script needs to be executed from within sqlite-db module anyway to produce the sqlx-data.json file. Hence, I think this is a justified solution.

This ticket is related: https://github.com/launchbadge/sqlx/issues/1399

### DIFF
--- a/sqlite-db/prepare_db.sh
+++ b/sqlite-db/prepare_db.sh
@@ -6,12 +6,12 @@ TEMPDB=${PWD}/tempdb
 set -e
 
 # create temporary DB
-DATABASE_URL=sqlite:$TEMPDB cargo sqlx database create
+DATABASE_URL=sqlite:${TEMPDB} cargo sqlx database create
 # make sure we remove the tempdb when exiting even if one of the following commands fails
-trap 'rm -f $TEMPDB' EXIT
+trap 'rm -f ${TEMPDB}' EXIT
 
 # run the migration scripts to create the tables
-DATABASE_URL=sqlite:$TEMPDB cargo sqlx migrate run
+DATABASE_URL=sqlite:${TEMPDB} cargo sqlx migrate run
 
 # prepare the sqlx-data.json rust mappings
-DATABASE_URL=sqlite:$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare -- --tests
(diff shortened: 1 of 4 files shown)


da-kami (commented):
Is it expected that the `sqlite-db/sqlx-data.json` file changes when only changing the path to the database? I would have not expected that 😅

bonomat:
> Is it expected that the `sqlite-db/sqlx-data.json` file changes when only changing the path to the database? I would have not expected that 😅

No, it means someone else did not check it in before me.

luckysori (commented):
Could the second commit be because of a different version of `sqlx`?

bonomat:
> Could the second commit be because of a different version of `sqlx`?

Not sure, but I'm preparing a ci target to ensure that the file is always up2date: It's currently being tested here: https://github.com/bonomat/hermes/runs/7444408411?check_suite_focus=true

sqlite-db/prepare_db.sh
```diff
@@ -15,4 +14,4 @@ trap 'rm -f $TEMPDB' EXIT
 DATABASE_URL=sqlite:$TEMPDB cargo sqlx migrate run
 
 # prepare the sqlx-data.json rust mappings
-DATABASE_URL=sqlite:./$CRATE_DIR/$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare -- --tests
+DATABASE_URL=sqlite:$TEMPDB SQLX_OFFLINE=true cargo sqlx prepare -- --tests
```
klochowicz: classic bash nit 🙃

```suggestion
DATABASE_URL=sqlite:${TEMPDB} SQLX_OFFLINE=true cargo sqlx prepare -- --tests
```

klochowicz (approved):
LGTM!

bonomat:
CI check works 😅 
https://github.com/itchysats/itchysats/runs/7460197517?check_suite_focus=true

luckysori (approved):

bonomat:
> Build failed:
> 
>     * [formatting](https://github.com/itchysats/itchysats/runs/7460558400?check_suite_focus=true)

? Works locally and the other ci job was also fine?

bonomat:
bors retry

PR #2504, opened 2022-07-21
Release version 0.5.0

1010Tom:
Hi @da-kami!
This PR was created in response to a manual trigger of the release workflow here: https://github.com/itchysats/itchysats/actions/runs/2710515212.
I've bumped the versions in the manifest files in this commit: 2f371f959c21df1333333859d6427a76db7d25a5.
Merging this PR will create a GitHub release!

PR #2505, opened 2022-07-21
Structure tests by use case instead of one `happy_path` file

da-kami:
Moves tons of code around. Should be pure refactor.


daemon-tests/tests/none-collaborative_settlement.rs
```diff
@@ -0,0 +1,55 @@
+use daemon::projection::CfdState;
```
klochowicz: nit: `non-collaborative-settlement` would be a much better name :P (I would also remove mix of hyphens and underscores, but I understand if you feel different).

klochowicz (approved):
a much needed TLC! :) thanks!

daemon-tests/tests/none-collaborative_settlement.rs
luckysori: I think files (modules) have to respect the [naming rules](https://rust-lang.github.io/api-guidelines/naming.html), so snake_case it is.

da-kami:
I just had a look at the test runtime for `macos` for different PRs.

For https://github.com/itchysats/itchysats/pull/2504 (the release) it took `18` minutes
For this PR it took `21` minutes (!)
For https://github.com/itchysats/itchysats/pull/2378 it took only `8` minutes (but that was already noted in https://github.com/itchysats/itchysats/issues/2473)

I wonder if splitting the tests into multiple files is not making them run longer, because test might not get triggered in parallel as flexible? Is there a correlation between tests and files, or are all tests from all files run in parallel?
Would this be relevant for https://github.com/itchysats/itchysats/issues/2473?

da-kami:
> I just had a look at the test runtime for `macos` for different PRs.
> 
> For #2504 (the release) it took `18` minutes For this PR it took `21` minutes (!) For #2378 it took only `8` minutes (but that was already noted in #2473)
> 
> I wonder if splitting the tests into multiple files is not making them run longer, because test might not get triggered in parallel as flexible? Is there a correlation between tests and files, or are all tests from all files run in parallel? Would this be relevant for #2473?

If my understanding is correct, then *all* the tests should be run in parallel when we trigger `cargo test --workspace` on CI. So, putting the tests into separate files should not have any effect on test execution times. Please correct me if I'm wrong.

klochowicz:
> > I just had a look at the test runtime for `macos` for different PRs.
> > For #2504 (the release) it took `18` minutes For this PR it took `21` minutes (!) For #2378 it took only `8` minutes (but that was already noted in #2473)
> > I wonder if splitting the tests into multiple files is not making them run longer, because test might not get triggered in parallel as flexible? Is there a correlation between tests and files, or are all tests from all files run in parallel? Would this be relevant for #2473?
> 
> If my understanding is correct, then _all_ the tests should be run in parallel when we trigger `cargo test --workspace` on CI. So, putting the tests into separate files should not have any effect on test execution times. Please correct me if I'm wrong.

it should not make much of a difference; I think this is definitely a step in the right direction. I will look into using https://nexte.st in our project to see the difference, it seems that it might help us a bit!

da-kami:
CI ran into a failure where the `taker` dialer is unable to establish a connection, see: https://github.com/itchysats/itchysats/runs/7450380260?check_suite_focus=true

I was unable to reproduce this locally, will re-trigger the tests. This might point to some connection management problem.

klochowicz:
> CI ran into a failure where the `taker` dialer is unable to establish a connection, see: https://github.com/itchysats/itchysats/runs/7450380260?check_suite_focus=true
> 
> I was unable to reproduce this locally, will re-trigger the tests. This might point to some connection management problem.

It's an issue that would fail the CI even with the legacy connection, although I have reported it as an issue as indeed it can be viewed as a regression that came with libp2p - in the old connection we'd exit this test soon after startup, instead of trying to pointlessly reconnect.
See the reported issue:
https://github.com/itchysats/itchysats/issues/2514

as this is an OS error, stating that we can't bind to the TCP socket, it should be reproducible to start maker config with dedicated libp2p port that is already used by something (even by another maker). I will write such test in the morning as a part of dealing with the bug noticed by you. I've added it to the iteration.

daemon-tests/tests/collaborative_settlement.rs
```diff
@@ -0,0 +1,102 @@
...
+async fn collaboratively_close_an_open_cfd_maker_going_short() {
+    collaboratively_close_an_open_cfd(Position::Short).await;
+}
+
+#[otel_test]
+async fn collaboratively_close_an_open_cfd_maker_going_long() {
+    collaboratively_close_an_open_cfd(Position::Long).await;
+}
+
+async fn collaboratively_close_an_open_cfd(maker_position: Position) {
```
luckysori: 🙃 I think it's better in test files when all the tests are at the top, and all the helpers at the bottom.

luckysori (approved):
Nice, thanks!

PR #2510, opened 2022-07-21
Separate test execution from compilation on the CI

klochowicz:
This should give us more insight into test execution time regressions vs
compilation issues.


klochowicz:
From the image below, we can see that running the tests is actually faster on macos, but for some reason compilation takes way longer (cc @da-kami )

image

bonomat (approved):

luckysori (approved):

bonomat:
let's get this in.

bors r+

PR #2512, opened 2022-07-21
More rollover tests

### LINKED ISSUE #2513: Investigate failures on extended rollover retry tests
See: https://github.com/itchysats/itchysats/pull/2512

da-kami:
Fixes https://github.com/itchysats/itchysats/issues/2513


da-kami:
Note: These tests increase our CI test run even more. I am not sure they are worth running on every CI run. (unless we touch the rollover logic). 
Shall we `ignore` these tests? I could also put them in a separate file (e.g. `rollover_expensive_ignored.rs` or similar) to raise awareness that these tests are ignored.

klochowicz:
> Note: These tests increase our CI test run even more. I am not sure they are worth running on every CI run. (unless we touch the rollover logic).
> Shall we `ignore` these tests? I could also put them in a separate file (e.g. `rollover_expensive_ignored.rs` or similar) to raise awareness that these tests are ignored.

We could run such expensive tests on a schedule (e.g. once a day, not within a PR). I will seek to create a PoC today.
Looking at the CI length, most of it is the test compile times ;/

daemon-tests/tests/rollover.rs
```diff
@@ -190,42 +182,98 @@ async fn maker_accepts_rollover_after_commit_finality() {
...
+        None,
+        predict_fees.calculate_for_hours(24),
+    )
+    .await;
+
+    // 2. Retry the rollover
+    // We simulate the taker being one rollover behind by setting the
+    // latest DLC to the one generated by contract setup
+    taker
+        .simulate_previous_rollover(
```
klochowicz: still not a fan of `simulate_previous_rollover`... 😟 maybe `undo_rollover_event` or something like that?

daemon-tests/tests/rollover.rs
```diff
@@ -235,51 +283,108 @@ async fn retry_rollover_an_open_cfd() {
         )
         .await;
 
-    // We mock the initial oracle event-id for the rollover retry again
-    mock_oracle_announcements(
+    // The expected fee is less within the test setup, because we now charge only one full turn,
+    // i.e. 24h
```
klochowicz: I have a hard time parsing this comment - would you mind rephrasing it?

klochowicz (approved):
Thanks for reproducing the problem in the tests! LGTM!

daemon-tests/tests/rollover.rs
```diff
@@ -190,42 +182,110 @@ async fn maker_accepts_rollover_after_commit_finality() {
 /// The first rollover is done with `example_1`.
 /// The second rollover is done with `example_0` (we re-use it)
```
luckysori: ❓ Could we retry the rollover using the `OracleData` from the first attempt? It is more natural, since when you are retrying in production you are almost certainly going to use the same announcement. It's confusing to go back in time and use the data from `example_0`.

daemon-tests/tests/rollover.rs
```diff
@@ -190,42 +182,110 @@ async fn maker_accepts_rollover_after_commit_finality() {
...
 #[otel_test]
-async fn retry_rollover_an_open_cfd() {
-    let contract_setup_oracle_data = OliviaData::example_0();
-    let contract_setup_oracle_data_announcement = contract_setup_oracle_data.announcement();
+async fn retry_rollover_an_open_cfd_from_contract_setup() {
```
luckysori: 🙃 What about naming this `given_contract_setup_completed_when_taker_fails_first_rollover_can_retry`?

daemon-tests/tests/rollover.rs
```diff
@@ -273,6 +403,17 @@ async fn retry_rollover_an_open_cfd() {
...
+        &mut taker,
+        order_id,
+        OliviaData::example_1(),
+        None,
+        fee_structure.predict_fees(72),
+    )
+    .await;
 }
 
 async fn prepare_rollover(
```
luckysori, line 388: 🙃 It would be good to have some documentation for this helper.

daemon-tests/tests/rollover.rs
```diff
@@ -190,42 +182,110 @@ async fn maker_accepts_rollover_after_commit_finality() {
...
+    );
+
+    assert_eq!(
+        taker.latest_commit_txid(),
+        maker.latest_commit_txid(),
+        "The maker and taker should have the same commit_txid after the rollover retry"
+    );
+
+    // 3. Ensure that we can do another rollover after the retry
+    // After another rollover we expect to be charged for 48h
```
luckysori: 🙃 Perhaps put this kind of comment next to the fee-related argument.

bonomat:
> We could run such expensive tests on a schedule (e.g. once a day, not within a PR). I will seek to create a PoC today.

I like that idea. It's pretty easy to create a github action which runs [once a day or so](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).

daemon-tests/src/flow.rs
```diff
@@ -7,7 +7,7 @@ use std::time::Duration;
 use tokio::sync::watch;
 
 /// Waiting time for the time on the watch channel before returning error
-const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) { 90 } else { 30 });
+const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) { 120 } else { 30 });
```
luckysori: 🤔 This change doesn't make sense to me based on the commit message.

The number of rollovers in any given test should not affect how long we wait for any individual rollover to complete, which is what `NEXT_WAIT_TIME` controls.

daemon-tests/src/lib.rs
```diff
@@ -280,10 +280,10 @@ pub async fn start_from_open_cfd_state(
     confirm!(lock transaction, order_to_take.id, maker, taker);
     wait_next_state!(order_to_take.id, maker, taker, CfdState::Open);
 
-    (maker, taker, order_to_take.id, fee_structure)
+    (maker, taker, order_to_take.id, predict_fees)
 }
 
-pub struct FeeStructure {
+pub struct PredictFees {
```
luckysori: 🙃 What about `FeeCalculator`? Constructing it with `new` is no longer misleading. And calling `calculate_for_hours` on it sounds more natural.

daemon-tests/tests/rollover.rs
```diff
@@ -515,4 +482,22 @@ async fn rollover(
...
+        commit_tx_id_before_rollover_maker,
+        maker.latest_commit_txid(),
+        "The commit_txid of the taker after the rollover retry should have changed"
+    );
+
+    assert_ne!(
+        commit_tx_id_before_rollover_taker,
+        taker.latest_commit_txid(),
+        "The commit_txid of the maker after the rollover retry should have changed"
+    );
```
luckysori: 👍 I'm happy that these were moved here.

🔧 But you have to remove the `retry` terminology, because this function is called in other contexts too.

luckysori (approved):
Good stuff, thank you!

daemon-tests/tests/rollover.rs
```diff
@@ -190,42 +182,98 @@ async fn maker_accepts_rollover_after_commit_finality() {
...
-        expected_maker_fee,
-        expected_taker_fee,
+        predict_fees.calculate_for_hours(48),
     )
     .await;
+}
 
-    // We simulate the taker being one rollover behind by setting the
+#[otel_test]
+async fn retry_rollover_an_open_cfd_from_contract_setup_with_rollover_in_between() {
```
luckysori: 🙃 And what about naming this `given_maker_two_rollovers_ahead_when_retry_then_taker_catches_up`?

daemon-tests/tests/rollover.rs
```diff
@@ -235,51 +283,108 @@ async fn retry_rollover_an_open_cfd() {
...
+        order_id,
+        OliviaData::example_1(),
+        None,
+        predict_fees.calculate_for_hours(48),
+    )
+    .await;
+}
+
+#[otel_test]
+async fn retry_rollover_an_open_cfd_from_previous_rollover() {
```
luckysori: 🙃 And this one could be `given_maker_one_rollover_ahead_when_retry_then_taker_catches_up`.

luckysori: :wrench: We could order the tests by complexity of the scenario:

1. Taker behind by one rollover.
2. Taker behind by two rollovers.
3. Taker failed the very first rollover.

luckysori (commented):
I have a suggestion about verifying that we can keep rolling over after a retry: could we just write a separate test for that? Or at least just verify it in _one_ of the tests? It seems a bit overkill, making the tests more complicated.

daemon-tests/tests/rollover.rs
da-kami: Initially this was on purpose because we want to make sure that the new `event-id` is used. Given that we now assert on multiple fields changing (i.e. `commit-txid`, `fees`) I think the `event-id` does not matter that much anymore and I can change it. The test still does not reflect the real world as good as it could.

daemon-tests/src/flow.rs
da-kami: Good point! It looks like this change should not have any effect on the test runtime. Could be that I just ran into a longer run while my machine was under heavy load and then prematurely changed this without thinking about it in depth.

da-kami:
> I have a suggestion about verifying that we can keep rolling over after a retry: could we just write a separate test for that? Or at least just verify it in _one_ of the tests? It seems a bit overkill, making the tests more complicated.

I understand the motivation but given the complexity of the code I am not sure this is a good idea. I'd like to ensure for *all* the scenarios that we can add another rollover so we don't run into scenarios where an edge cases works, but then consecutive rollovers fail.
I thought about duplicating the setup, so we run tests that don't add an additional rollover all the time, and the one with additional rollover only e.g. once per day, but given that the test that includes the additional rollover includes the other test I am not sure this is worth it.

da-kami:
> > We could run such expensive tests on a schedule (e.g. once a day, not within a PR). I will seek to create a PoC today.
> 
> I like that idea. It's pretty easy to create a github action which runs [once a day or so](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule).

After thinking about this a bit more I am not sure this is the way to go. This will require developers to be mindful to run rollover tests when changing code that affects rollovers. This can be tricky, because sometimes we have side effects of code pathes influencing rollover. This could mean that a PR merges that breaks the tests and we only notice this once the `cron` job triggers. 

I think the additional time added by these tests is actually not significant; when I wrote my comment I was not aware that most of the current CI time is compilation. 
For now I would just merge the tests and optimize how to run them separately If that's OK.

PR #2516, opened 2022-07-22
Allow authenticated ws to BitMEX

bonomat:
Some APIs of BitMEX expect authentication such as retrieving updated for `Order`, `Executions` or `Position`.
By passing in `credentials` the library will try to send an authentication message.

I've also changed the way how we subscribe to topics: instead of passing them into the url, a explicit subscription method was added.

Note: if authentication fails, an error message as string will be returned and if a topic was chosen which requires authentication the stream will be closed.

### DIFF
--- a/bitmex-stream/examples/authenticated_api.rs
+++ b/bitmex-stream/examples/authenticated_api.rs
@@ -1,0 +1,26 @@
+use anyhow::Result;
+use bitmex_stream::Credentials;
+use bitmex_stream::Network;
+use futures::TryStreamExt;
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    tracing_subscriber::fmt()
+        .with_env_filter("info,bitmex_stream=trace")
+        .init();
+
+    let mut stream = bitmex_stream::subscribe_with_credentials(
+        ["execution".to_owned()],
+        Network::Testnet,
+        Credentials {
+            api_key: "some_api_key".to_string(),
+            secret: "some_secret".to_string(),
+        },
+    );
+
+    while let Some(result) = stream.try_next().await? {
+        tracing::info!("{result}");
+    }
+
+    Ok(())
+}
--- a/bitmex-stream/src/lib.rs
+++ b/bitmex-stream/src/lib.rs
@@ -1,30 +1,85 @@
+use crate::tungstenite::http::Method;
+use anyhow::Context;
 use async_stream::stream;
 use futures::SinkExt;
 use futures::Stream;
 use futures::StreamExt;
+use serde::ser::SerializeTuple;
+use serde::Serialize;
+use serde_json::to_string;
+use std::ops::Add;
 use std::time::Duration;
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
 use tokio_tungstenite::tungstenite;
 use tracing::Instrument;
+use url::Url;
 
-pub use tokio_tungstenite::tungstenite::Error;
+pub use anyhow::Error;
 
-/// Connects to the BitMex websocket API, subscribes to the specified topics (comma-separated) and
-/// yields all messages.
+/// Connects to the BitMex websocket API
 ///
-/// To keep the connection alive, a websocket `Ping` will be sent every 5 seconds in case no other
-/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
+/// It subscribes to the specified topics (comma-separated) and yields all messages.
+/// If the topics need authentication please use `subscribe_with_credentials` instead.
 pub fn subscribe<const N: usize>(
     topics: [String; N],
     network: Network,
 ) -> impl Stream<Item = Result<String, Error>> + Unpin {
+    subscribe_impl(topics, network, None)
+}
+
+/// Connects to the BitMex websocket API with authentication
+///
+/// It subscribes to the specified topics (comma-separated) and yields all messages.
+/// If invalid credentials have been provided but a topic was provided which needs authentication
+/// the stream will be closed.
+pub fn subscribe_with_credentials<const N: usize>(
+    topics: [String; N],
+    network: Network,
+    credentials: Credentials,
+) -> impl Stream<Item = Result<String, Error>> + Unpin {
+    subscribe_impl(topics, network, Some(credentials))
+}
+
+/// Connects to the BitMex websocket API, subscribes to the specified topics (comma-separated) and
+/// yields all messages.
+///
+/// To keep the connection alive, a websocket `Ping` is sent every 5 seconds in case no other
+/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
+fn subscribe_impl<const N: usize>(
+    topics: [String; N],
+    network: Network,
+    credentials: Option<Credentials>,
+) -> impl Stream<Item = Result<String, Error>> + Unpin {
+    let url = network.to_url();
+    let url = format!("wss://{url}/realtime");
+
     let stream = stream! {
         tracing::debug!("Connecting to BitMex realtime API");
 
-        let subscription = topics.join(",");
-        let url = network.to_url();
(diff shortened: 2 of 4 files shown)


bitmex-stream/src/lib.rs
```diff
@@ -16,16 +25,41 @@ pub use tokio_tungstenite::tungstenite::Error;
 pub fn subscribe<const N: usize>(
```
klochowicz, line 24: in order not to break the public API (especially for users not interested in passing in credentials), it would be better to create an additional public function `subscribe_with_credentials` and then both forward to this implementation.

I know you want this sooner rather than later in production, so I'll quickly create a PR that shifts that around so that you can have a look if it's needed over the weekend.

It's probably not worth creating a builder pattern here as it's unlikely to grow more than that, but it's also an option 👯‍♂️

bitmex-stream/src/lib.rs
```diff
@@ -16,16 +25,41 @@ pub use tokio_tungstenite::tungstenite::Error;
...
+            let expires = start
+                .duration_since(UNIX_EPOCH)
+                .expect("Time went backwards")
+                .add(Duration::from_secs(5))
+                .as_secs();
+            let signature = credentials.sign(Method::GET, expires, &Url::parse(url.as_str())?, "");
+            messages.push(tungstenite::Message::try_from(signature.into_command())?);
+        }
+
+        messages.push(tungstenite::Message::try_from(Command::Subscribe(
```
da-kami: What's the advantage of sending multiple subscribe messages? 
Did you make sure that subscribing twice does not overwrite the previous subscription?

bitmex-stream/src/lib.rs
klochowicz: feel free to cherry-pick and squash the commit: https://github.com/itchysats/itchysats/pull/2529/files

bitmex-stream/src/lib.rs
```diff
@@ -16,16 +25,41 @@ pub use tokio_tungstenite::tungstenite::Error;
...
+            .await.context("Could not connect to websocket")?;
 
         tracing::info!("Connected to BitMex realtime API");
 
+        let mut messages = vec![];
+        if let Some(credentials) = credentials {
+            let start = SystemTime::now();
+            let expires = start
+                .duration_since(UNIX_EPOCH)
+                .expect("Time went backwards")
```
klochowicz: no need for expect here, as we can use `?`

klochowicz:
IMHO it would be nice if eventually bad credentials returned something different than anyhow::Error, so that the user could fail hard (or try different credentials) instead of plain retry. 

maybe overall you could create a custom error enum (for `subscribe_with_credentials) that returns either anyhow::Error or `BadCredentials` error.

klochowicz (approved):
concept ACK :)

if you don't mind, try cherry-picking and squashing my commit in the linked PR - I find `Option` and bool arguments in API a code smell that's hard to read later on. 

other suggestions are not critical, so if this needs to get in for the weekend, I'm leaving an 👍

bitmex-stream/src/lib.rs
bonomat: > What's the advantage of sending multiple subscribe messages?

We have the option to subscribe to multiple topics. 

>  Did you make sure that subscribing twice does not overwrite the previous subscription?

That's the way how it works :) Yes, I've tried.

bitmex-stream/src/lib.rs
bonomat: Thanks. Next time feel free to push on top of this PR or create a PR on top of this instead of creating a new one against master.

bonomat:
> IMHO it would be nice if eventually bad credentials returned something different than anyhow::Error, so that the user could fail hard (or try different credentials) instead of plain retry.
> 
> maybe overall you could create a custom error enum (for `subscribe_with_credentials) that returns either anyhow::Error or `BadCredentials` error.

I agree with this statement :) 
It's really hard to do error handling on the messages. @luckysori made some efforts to parse the response messages which come as _table_-typed json. The plan is to integrate that in the future.

bonomat:
failed at rebasing :/  I'll fix it in a sec

bonomat:
> failed at rebasing :/ I'll fix it in a sec

fixed

bitmex-stream/examples/authenticated_api.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::Result;
```
luckysori, line 1: 👍 Thanks for adding an example.

bitmex-stream/src/lib.rs
```diff
@@ -1,31 +1,84 @@
...
     network: Network,
 ) -> impl Stream<Item = Result<String, Error>> + Unpin {
+    subscribe_impl(topics, network, None)
+}
+
+/// Connects with authentication to the BitMex websocket API, subscribes to the
+/// specified topics (comma-separated) and yields all messages.
+///
+/// To keep the connection alive, a websocket `Ping` is sent every 5 seconds in case no other
+/// message was received in-between. This is according to BitMex's API documentation: https://www.bitmex.com/app/wsAPI#Heartbeats
```
luckysori: 🔧 This section is duplicated from `subscribe`. Can we move to `subscribe_impl`?

bitmex-stream/src/lib.rs
```diff
@@ -1,31 +1,84 @@
...
+            let start = SystemTime::now();
+            let expires = start
+                .duration_since(UNIX_EPOCH)?
+                .add(Duration::from_secs(5))
+                .as_secs();
+            let signature = credentials.sign(Method::GET, expires, &Url::parse(url.as_str())?, "");
+            messages.push(tungstenite::Message::try_from(signature.into_command())?);
+        }
+
+        messages.push(tungstenite::Message::try_from(Command::Subscribe(
```
luckysori: 🔧 Can we not just send immediately here?

bitmex-stream/src/lib.rs
```diff
@@ -1,31 +1,84 @@
...
+            topics.to_vec(),
+        ))?);
+
+        for message in messages {
+            let _ = connection
+                .send(message)
+                .await;
+        }
```
luckysori: 🙃 Topics into an iterator of futures could be simpler.

Alternatively if you send the authentication message separately you could simply iterate through `topics` with a `for` loop here. That would get rid of `messages`.

bitmex-stream/src/lib.rs
```diff
@@ -85,3 +138,165 @@ impl Network {
...
+            expires,
+        }
+    }
+}
+
+impl Signature {
+    fn into_command(self) -> Command {
+        Command::Authenticate(self)
+    }
+}
```
luckysori: 🙃 `impl From`?

luckysori (approved):
Nice work. I have left some suggestions. I also think we might want to extract the authentication-specific stuff into a separate private module, which we can use in `lib.rs`.

bonomat:
bors cancel

bitmex-stream/src/lib.rs
bonomat: where's the fun in that? 😅

bitmex-stream/src/lib.rs
```diff
@@ -42,6 +32,11 @@ pub fn subscribe_with_credentials<const N: usize>(
     subscribe_impl(topics, network, Some(credentials))
 }
 
+/// Connects to the BitMex websocket API, subscribes to the specified topics (comma-separated) and
+/// yields all messages.
```
luckysori: 🙃 Ah, maybe I didn't explain it clearly. I think we do want a docstring on the public API (and a slightly different one per function)! 

It's just that the second part about the ping interval is pretty irrelevant for consumers (and was duplicated), so it could just be demoted to a comment next to `subscribe_impl`.

Sorry for the confusion!

bonomat: ah, got it :)

bonomat:
f** again failed to rebase... note to myself: learn how to fork...

luckysori (approved):
Awesome, thank you!

PR #2519, opened 2022-07-22
New approach to ignore frequent spans

Restioson:
Instead of trying to delete all spans of a given name, this works by having a parent span, either `always_quiet_children` or `sometimes_quiet_children`, and then putting quietened spans inside of that. If this isn't done, then grandchildren of quietened spans simply get the quietened span's parent, or no parent, and therefore do not get ignored, as they are no longer children of the quietened span. Since the quiet_children spans are never deleted, this solves the issue.

~~xtra needs to have https://github.com/Restioson/xtra/pull/154 merged before this will work, since otherwise lookup_current may return None, which breaks this solution, as then there is no way to get the parent from the Context. Therefore, this is a draft PR until https://github.com/Restioson/xtra/pull/154 is merged.~~ Merged!


Restioson:
One thing I was wondering is if it's worth making a new, tiny crate that just defines two macros - one for `always_quiet` and one for `sometimes_quiet`, so that the user-crates don't need to copy out `info_span!(target: "quiet_children", "always_quiet_children")`, etc

klochowicz:
> One thing I was wondering is if it's worth making a new, tiny crate that just defines two macros - one for `always_quiet` and one for `sometimes_quiet`, so that the user-crates don't need to copy out `info_span!(target: "quiet_children", "always_quiet_children")`, etc

I would probably go with a free function and / or consts. I'm not sure we need a macro / crate for this - seems like it would needlessly slow down the build even further :P

xtra-libp2p-offer/src/maker.rs
```diff
@@ -59,9 +57,12 @@ impl Actor {
     async fn handle(&mut self, msg: NewOffers, ctx: &mut xtra::Context<Self>) {
         self.latest_offers = msg.0;
 
+        let quiet = tracing::info_span!(target: "quiet_children", "sometimes_quiet_children");
```
klochowicz: consider creating a factory function in `shared_bin` that just returns this, maybe named `quiet_span()` ?

then you can use it without leaking it implementation details

shared-bin/src/logger.rs
```diff
@@ -126,26 +116,26 @@ pub fn init(
...
-        ]);
-    }
-
-    let disable_spans = tracing_subscriber::filter::filter_fn(move |meta| {
-        !(*meta.level() > Level::ERROR && error_only.contains(&meta.name()))
+    let disable_spans = dynamic_filter_fn(move |meta, ctx| {
+        if meta.is_span() && !["tokio", "runtime"].contains(&meta.target()) {
+            match ctx.lookup_current() {
+                Some(parent) => {
+                    !(parent.name() == "always_quiet_children"
```
klochowicz: I suspect `always_quiet_children` could be a const in this module (important for maintainability). same goes for the other one.

klochowicz (commented):
I'm excited to see whether this solution works - interesting workaround!  :)

I've added comments mostly for maintainability sake, not for the approach itself.

shared-bin/src/logger.rs
Restioson: I don't think so, since libraries like xtra-libp2p-ping need to use it too. I think we'd need to make a new common crate for it potentially if we want to refactor it out, otherwise they'd need to depend on shared-bin too.

klochowicz: fair enough! I didn't notice that xtra-libp2p-ping needs to depend on it. in such case, I would create a small separate crate out of it indeed.

klochowicz:
Why are there 3 distinct strings and how to know which one to use (if we don't rely on a crate that provides them, it's even hard to put docs into it):
- `sometimes_quiet_children`
- `quiet_children`
- always_quiet_children` 

is there any difference between the first 2? When looking at the utility of silencing the spans, I would distinguish 2 myself:
- ones that you can override ("sometimes_quiet_children")
- ones you can't, as they are disabled permanently ("always_quiet_children")

what's the utility of `quiet_children` ?

Restioson:
> what's the utility of quiet_children ?

`quiet_children` is the log target for both, and the idea is that `shared_bin::init` will set this to `info` so that the spans are never ignored.

> if we don't rely on a crate that provides them, it's even hard to put docs into it

I agree. I am going to restructure the PR to add a helper crate so that it's more maintainable

Restioson:
I've refactored this now into a `quiet_spans` crate as you suggested :slightly_smiling_face:

quiet-spans/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/// Children of this span are ignored if `verbose_spans` is false in [`disable_noisy_spans`]
+pub fn sometimes_quiet_children() -> Span {
+    tracing::info_span!(target: TARGET, SOMETIMES_QUIET)
+}
+
+/// Layer which disables noisy spans - children of [`always_quiet_children`] and children of
+/// [`sometimes_quiet_children`] (when `verbose_spans` is false).
+pub fn disable_noisy_spans<S>(verbose_spans: bool) -> impl Layer<S>
```
klochowicz, line 24: nit: it's hard to read such API at callsites if it's a bool. Rust has expresive syntax, an enum `NoisySpanLevel  { Some, All}` or something similar would go a step further towards a readable API. 
(then some of the comment can go there)

`disable_noisy_spans(false)` might be interpreted as *not* disabling them aftor all!😝

klochowicz (approved):
awesome work! thanks for the patience in the review, I think the end result was well worth it! 🥇

quiet-spans/src/lib.rs
klochowicz: I'll open a follow-up PR to clean this up, in the meantime I think it's worth merging.

Restioson: Thanks :)

PR #2526, opened 2022-07-22
PoC: see how much we benefit from rust-cache for test runs on the CI

klochowicz:
rust-cache only caches external crates, and apparently it's most useful in small
projects, where own crates are an exception.


klochowicz:
this experiment concluded that it's still useful to have it, as the tests finished even slower here.

PR #2528, opened 2022-07-22
WIP - Link offer-id to order-id

da-kami:
I started hacking on this to see where we run into obstacles. 

One obstacle I found is, that the `order` actor in the maker will have to know the current `offer` - which is expected. @klochowicz maybe you can have a quick look and let me know if this is *not* what you had in mind. I am planning to go on with this on Monday :)

### DIFF
--- a/daemon/src/order/maker.rs
+++ b/daemon/src/order/maker.rs
@@ -112,39 +112,12 @@ impl Actor {
         };
 
         let (
-            id,
-            quantity,
-            leverage,
-            position,
-            opening_price,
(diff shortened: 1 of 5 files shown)


luckysori:
Is this still relevant or should we just look at #2292?

klochowicz:
> Is this still relevant or should we just look at #2292?

I suspect this draft was a PoC for early feedback before @da-kami rolled it out in the order protocol PR. We ended up having a walkthrough on jitsi, so there's no comments here.

PR #2533, opened 2022-07-25
Replace rollover confirmation step with configurable flag

luckysori:
This simplifies both the maker's side of the rollover protocol and the maker itself.

We were also using the confirmation step to provide certain offer parameters back to the protocol execution. Having removed the step, we now fetch them from the `maker::cfd::Actor` via the new `xtra_libp2p_rollover::GetRates` trait.

We remove two tests which no longer make sense because we now do not pause the execution of the rollover protocol to get confirmation via the HTTP API.

---

This was originally in https://github.com/itchysats/itchysats/pull/2378, but @da-kami advised to split that PR into 2. Apart from refactoring certain stuff, removing the rollover confirmation step from the maker side makes the rest of #2378 easier. Removing that step is something that we have considered for a while and some members of the team already agreed that it was a good idea [a while back](https://github.com/itchysats/architecture/pull/28#discussion_r864341856).


maker/src/cfd.rs
```diff
@@ -961,6 +886,39 @@ where
     }
 }
 
+/// Source of offer rates used for rolling over CFDs.
+#[derive(Clone)]
+pub struct RatesChannel(MessageChannel<GetOffers, Option<MakerOffers>>);
```
da-kami, line 891: Maybe we could opt to implement this directly on the `offers` protocol.
In my latest changes to the `order` protocol I also added the possibility to fetch the latest offer, see: https://github.com/itchysats/itchysats/pull/2292/commits/eaa7016d98d4c0a1e7731a0442283ebbda1cbe4c

We might want to harmonize this, but we can do that on top of this PR.

maker/src/routes.rs
```diff
@@ -276,6 +276,30 @@ pub async fn get_metrics<'r>(_auth: Authenticated) -> Result<String, HttpApiProb
     Ok(metrics)
 }
 
+#[derive(Debug, Clone, Copy, Deserialize)]
+pub struct RolloverConfig {
+    is_accepting_rollovers: bool,
+}
+
+#[rocket::post("/rollover-config", data = "<config>")]
```
da-kami: I find `/rollover/config` nicer, but not super important.

da-kami (approved):

maker/src/cfd.rs
luckysori: I like this proposal. Let's do it once both this PR and the one you reference get merged.

luckysori: See https://github.com/itchysats/itchysats/issues/2546.

maker/src/routes.rs
luckysori: Let's do it!

PR #2535, opened 2022-07-25
Tell users to include contact details in details

da-kami:
At the moment it is not obvious that we cannot contact users when they send a request via feedback fish.
There is no means to specify additional fields in the feedback fish form.

We can opt for asking users for their emails to "register" their ItchySats version, and then attach the email as metadata, but this is a separate feature that will require some more time.
The easiest way to tell the user that they have to provide contact details in order to receive a response is a tooltip.

image

### DIFF
--- a/taker-frontend/src/components/Footer.tsx
+++ b/taker-frontend/src/components/Footer.tsx
@@ -18,6 +18,7 @@ import {
     ModalOverlay,
     Switch,
     Text,
+    Tooltip,
     useColorModeValue,
     useDisclosure,
 } from "@chakra-ui/react";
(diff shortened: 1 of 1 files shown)


taker-frontend/src/components/Footer.tsx
```diff
@@ -146,7 +147,11 @@ export default function Footer(
                             leftIcon={<FaRegCommentDots />}
                             variant={"ghost"}
                         >
-                            <Text display={["none", "none", "inherit"]}>Send Feedback</Text>
+                            <Tooltip
+                                label={"To received a response please include contact details such as an email address or Telegram handle in the ticket text."}
```
luckysori:
```suggestion
                                label={"To receive a response please include contact details such as an email address or Telegram handle in the ticket text."}
```

klochowicz (approved):

bonomat (approved):
Happy to get it in but I'm not sure this is effective enough. I tried it and clicked on the `send feedback` button and did not see the tooltip as it was immediately behind the popup.

PR #2542, opened 2022-07-25
Extract vergen version into its own crate

klochowicz:
Vergen compiles whole libgit2 as its dependency, which takes a lot of time.
As this crate has no other dependencies, it can be built earlier.

### DIFF
--- a/vergen-version/Cargo.toml
+++ b/vergen-version/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "vergen-version"
+version = "0.1.0"
+edition = "2021"
+description = "Expose workspace version with git hash using vergen"
+
+[dependencies]
+
+[build-dependencies]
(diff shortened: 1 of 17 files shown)


Restioson (approved):

vergen-version/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "vergen-version"
+version = "0.1.0"
+edition = "2021"
+description = "Expose workspace version with git hash using vergen"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
bonomat:
```suggestion
```

bonomat (approved):

luckysori (approved):
Nice solution!

I think we are missing adding the new crate to `bors.toml` though.

PR #2543, opened 2022-07-25
`otlp` feature

### LINKED ISSUE #2524: Add `instrumentation` feature flag to daemon-tests
otel-tests crate requiring grpc-sys greatly increased our build times on CI (where we don't even use it!)

it can even be enabled by default in itchysats for devs convenience (we use incremental compile there, so it shouldn't matter), as long as we override and not compile stuff we don't use in the CI runs and release builds.

Restioson:
Closes #2524 by adding an `otlp` feature, which is disabled by default and prevents having to compile gRPC when OTLP instrumentation is not required. Both the `otlp` feature and `ITCHYSATS_TEST_INSTRUMENTATION` need to be enabled for instrumentation to work. If just the env var is passed, the test will panic.

### DIFF
--- a/otel-tests/src/lib.rs
+++ b/otel-tests/src/lib.rs
@@ -1,7 +1,3 @@
-use opentelemetry::sdk::trace;
-use opentelemetry::sdk::Resource;
-use opentelemetry::KeyValue;
-use opentelemetry_otlp::WithExportConfig;
 use std::sync::Once;
 use tracing_subscriber::filter::LevelFilter;
 use tracing_subscriber::layer::SubscriberExt;
@@ -13,6 +9,7 @@ use tracing_subscriber::Registry;
 #[doc(hidden)]
 pub mod __reexport {
     pub use futures;
+    #[cfg(feature = "otlp")]
     pub use opentelemetry;
     pub use tokio;
 }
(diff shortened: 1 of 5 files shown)


Restioson:
This PR took 7min on ubuntu to compile the tests, vs 15min on master.

otel-tests/src/lib.rs
```diff
@@ -22,9 +19,23 @@ pub use otel_tests_macro::otel_test;
 static INIT_OTLP_EXPORTER: Once = Once::new();
 
 pub fn init_tracing(module: &'static str) {
+    let env = std::env::var("ITCHYSATS_TEST_INSTRUMENTATION").unwrap_or_default();
```
klochowicz, line 22: do we still need the envvar? It seems that if we compile it with support, we might as well use it? (it's already opt-in)

klochowicz (approved):
thanks for tackling this!

otel-tests/src/lib.rs
Restioson: That's true. I left it in for now, as you might want to ignore the output for one run or something without recompiling everything. I can remove the env var too, though.

klochowicz: might as well keep it for now, until it bugs someone ;-) can always be done quickly in a follow-up

Restioson:
Seems like the change in compile time could be due to the caching you added @klochowicz? Regardless, I wouldn't expect any regression in compile time here, since we are compiling strictly fewer dependencies

PR #2552, opened 2022-07-26
Remove support for legacy rollover protocol

luckysori:
The original motivation for this patch was to simplify the codebase by removing any code related to the maker's manual accept/reject rollover mechanism, which was removed from the `/itchysats/rollover` protocol with https://github.com/itchysats/itchysats/pull/2533.

But doing so would mean that we would have to update the legacy rollover protocol on the maker side (not a breaking change since it only affects the maker, but a change I don't want to make). I think we can already stop maintaining this code if we intend to stop support for the legacy rollover protocol in the next release.


klochowicz:
is the expectation that we will drop all the legacy connections for the next release or just rollover?

bonomat (approved):
very nice. 

I think before we remove the old protocol we should give users on a legacy connection a chance to upgrade. We have (or had?) a protocol version which we could increment which would make a warning show up at the taker's ui. 
Should we do this first and rollout a release for 0.5.2 and leave this running for a couple of days?

luckysori:
Let's close this for now and we can reopen once we've released a version with all the new protocols and we've given takers time to upgrade.

Issue #2166, opened 2022-06-05
Position force closed because not rolled over over libp2p connection


da-kami:
Taker: mainnet, Umbrel `taker:0.4.19`

All times are UTC

It appears my taker was unable to establish a connection to the maker:

```
2022-06-03 07:51:52 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=model::cfd::Cfd cfd_version=3196 num_events=1
2022-06-03 07:51:52 DEBUG daemon::db: Applying new events to CFD order_id=6236074d-44e7-41ec-aec6-121b30b682ac aggregate=model::cfd::Cfd cfd_version=1743 num_events=0
2022-06-03 07:51:52 ERROR daemon::rollover::taker: Failed to start rollover: No connection to 12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d
2022-06-03 07:51:52 ERROR daemon::rollover::protocol: Rollover failed: No connection to 12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d
2022-06-03 07:51:52 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=model::cfd::Cfd cfd_version=3197 num_events=0
2022-06-03 07:51:52  WARN model::cfd: Rollover failed: No connection to 12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d
2022-06-03 07:51:52  INFO daemon::db: Appended event to database event=RolloverFailed order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d
2022-06-03 07:51:52 DEBUG daemon::db: Applying new events to CFD order_id=0fdfe886-52cf-4409-bd9f-ca47f3450d1d aggregate=daemon::position_metrics::Cfd cfd_version=3197 num_events=1
[...]

PR #2566, opened 2022-07-27
Taker drops connection upon failure to open substream for rollover

### LINKED ISSUE #2554: Improve `Endpoint` resilience to drop "stale" connections on the `maker`
@holzeis and me were able to reproduce a problem where the taker becomes unavailable ungracefully there are scenarios where the maker's endpoint is not notified with `ConnectionDropped`.
We were able to reproduce this by killing the `taker` process on the raspi. The `maker` did not receive a connection drop in such a case. 

This behavior results in the endpoint and protocol actors assuming that there is still a connection and the maker then fails to establish substreams when a protocol execution is triggered.

At this point we are not totally sure if this problem can be solved by configuring how the TCP connection is managed by `yamux`, or if we are not reacting to an event from the underlying stack that would notify us. This can be investigated, but it might be hard, and it is unclear if we can always rely on this.

So, it might still be better, to make the `Endpoint` resilient to "missed" connection drops, where we keep trying to open a substream but fail.

---

Some notes on solution design:

We did make an effort to overcome this problem before, by introducing an failure counter, see #2175,  but removed this feature because it had unwanted side effects (see #2207). I think the overall idea of this feature was still good, but we just need to design it in a better way.

da-kami:
Rollovers are important to keep the position alive.
If the taker fails to open a substream for rollover then something might be wrong with the connection.
To make sure we don't run into a scenario where the actor cannot open substreams the taker drops the connection in case

- `NoConnection` is reported
- we fail to open a substream over the connection

This is an emergency fallback scenario that should usually not be triggered.
It is added to avoid force closing positions due to connection problems.

---

This does not actually resolve https://github.com/itchysats/itchysats/issues/2554 but it should be good enough for now to avoid rare force-close scenarios where takers lose connection without the endpoint noticing.

### DIFF
--- a/xtra-libp2p-rollover/src/v_2_0_0/taker.rs
+++ b/xtra-libp2p-rollover/src/v_2_0_0/taker.rs
@@ -18,6 +18,7 @@ use model::Timestamp;
 use std::time::Duration;
 use tokio_extras::FutureExt;
 use xtra::Address;
+use xtra_libp2p::endpoint;
 use xtra_libp2p::Endpoint;
 use xtra_libp2p::OpenSubstream;
 use xtra_libp2p::Substream;
@@ -84,12 +85,41 @@ impl<E, O> Actor<E, O> {
                 v_2_0_0::PROTOCOL,
             ))
             .await
-            .context("Endpoint is disconnected")?
-            .context("No connection to peer")?
-            .await
-            .context("Failed to open substream")?;
+            .context("Endpoint is disconnected")?;
+
+        let substream = match substream.context("No connection to peer") {
+            Ok(substream) => substream,
+            Err(e) => {
+                self.emergency_drop_connection(peer_id, &e).await?;
+                return Err(e);
+            }
+        };
+
+        let substream = match substream.await.context("Failed to open substream") {
+            Ok(substream) => substream,
+            Err(e) => {
+                self.emergency_drop_connection(peer_id, &e).await?;
+                return Err(e);
+            }
+        };
 
         Ok(substream)
+    }
+
+    /// Emergency drop the connection
+    ///
+    /// This can be triggered by the taker in case the rollover protocol encounters an error when
+    /// opening a substream. This is only used for scenarios where the taker might have a broken
+    /// connection, but the endpoint does not detect it. To make sure that rollover can recover
+    /// the protocol can trigger an emergency re-connect in the hope that the next rollover retry
+    /// will work.
+    async fn emergency_drop_connection(&self, peer_id: PeerId, e: &anyhow::Error) -> Result<()> {
+        tracing::error!(%peer_id, "Dropping connection, because failed to open substream for rollover: {e:#}");
+        self.endpoint
+            .send(endpoint::Disconnect(peer_id.inner()))
+            .await?;
+
+        Ok(())
     }
 }
 


da-kami:
So far I only applied this to `rollover` on the taker side, because it is critical to avoid force close. For rollover I don't expect side effects, because we only evaluate and trigger the rollover in `auto_rollover` every 5 minutes, which is definitely enough to drop and re-establish the connection.

It might be a good idea to add it to the `collab-settlement` protocol for the taker as well, because that is also critical in the sense of not losing money because the user's deamon behavior blocks an action.
For `collab-settlement` the dynamic might be a bit different, because it is user-triggerd, thus I am not sure about side effects for triggering a disconnect. A user hitting the button in anger might cause repeated disconnects. That's one reason why I added it only for rollover so far ;)

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
```diff
@@ -84,13 +85,42 @@ impl<E, O> Actor<E, O> {
...
-            .context("Failed to open substream")?;
+            .context("Endpoint is disconnected")?;
+
+        let substream = match substream.context("No connection to peer") {
+            Ok(substream) => substream,
+            Err(e) => {
+                self.emergency_drop_connection(peer_id, &e).await?;
+                return Err(e);
+            }
+        };
```
luckysori, line 96: 🔧 This error means that you don't have a connection to the peer. You don't need to drop the connection!

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
```diff
@@ -84,13 +85,42 @@ impl<E, O> Actor<E, O> {
...
+
+        let substream = match substream.await.context("Failed to open substream") {
+            Ok(substream) => substream,
+            Err(e) => {
+                self.emergency_drop_connection(peer_id, &e).await?;
+                return Err(e);
+            }
+        };
```
luckysori, line 104: 🤔 There are too many reasons why this could fail that do not warrant a disconnection:
- A failed negotiation because the parties disagree on the protocols will not be solved by this.
- The listener may take too long to respond to the message, but the connection between the two parties is still completely fine. We are just adding more work to the listener to handle an unnecessary reconnect.
- The dialer themself may take too long to perform the protocol! Perhaps the async work is not being scheduled in time because there's too much load on the application.

There are some `yamux::ConnectionError`s that may indicate that we should drop the connection, but not all:

```rust
pub enum ConnectionError {
    /// An underlying I/O error occured.
    Io(std::io::Error),
    /// Decoding a Yamux message frame failed.
    Decode(FrameDecodeError),
    /// The whole range of stream IDs has been used up.
    NoMoreStreamIds,
    /// An operation fails because the connection is closed.
    Closed,
    /// Too many streams are open, so no further ones can be opened at this time.
    TooManyStreams
}
```

In conclusion, I think one data point is insufficient to drop the connection. And giving this power to any one protocol does not seem wise either.

Having looked at the `yamux::ConnectionError`s above, I would be more interested in letting the `Endpoint` decide if it should drop a particular connection depending on the error (`Closed` and `Io` maybe?).

luckysori (commented):
I don't think this is a good idea :(

klochowicz:
for some reason, my yesterday's comment didn't get published: 

are we sure that this was the reason force-closures were happening?

I'm a bit sceptical it would fail to open substream for over 24h, if at the same time we had zero problems with rolling over other things... (and afterwards, his taker interacted again without restart on his part)

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
da-kami: But we previously saw instances where we did not recover from this. I think it can happen that the endpoint does not notice that the connection is broken and then it will never notify the dialer actor to re-connect and we keep running into this problem.

da-kami:
I understand your concerns, but unfortunately I don't trust the endpoint always detecting that the connection is "broken". 

If the endpoint does not react the protocol actors will fail, potentially forever!
Rollover only happens every `2 hours`. After a failed rollover we only re-trigger after `5 minutes`. I don't see a problem disconnecting within these time constraints if we fail to roll over due to the two errors.
I understand that this change is not optimal from a boundaries / clean code perspective, but I don't see any harm in letting rollover take control here.

If we can still run into scenarios similar to what is described in https://github.com/itchysats/itchysats/issues/2166 (potentially triggered by a different source than described in the ticket) this would mean that the taker gets no feedback and the positions just gets force closed. 
Unless the user comes back to us with logs we will have to "hope" to run into such a scenario on testnet at some point so we can reproduce what is the problem... I'd rather have an emergency switch for now 😅

What do you think @bonomat ?

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
da-kami: > A failed negotiation because the parties disagree on the protocols will not be solved by this.

Yes, and it should not be. I would accept that a taker with an incompatible rollover protocol will reconnect or treat that specific error separate.

> The listener may take too long to respond to the message, but the connection between the two parties is still completely fine. We are just adding more work to the listener to handle an unnecessary reconnect.

I don't think this is a valid argument for what the user experience. The user failed to roll-over for 24h resulting in a force close. This would mean the listener was too busy to accept the substream for 24h.

> The dialer themself may take too long to perform the protocol! Perhaps the async work is not being scheduled in time because there's too much load on the application.

I don't think this is a valid argument for what the user experience. It is unlikely that the taker daemon was too busy to handle work within 24h. From our experience the taker daemon does not take many resources.

> Having looked at the yamux::ConnectionErrors above, I would be more interested in letting the Endpoint decide if it should drop a particular connection depending on the error (Closed and Io maybe?).

Thanks for having a look, I am happy to consider that. But I am not sure that would protect us from the endpoint not dropping a connection in some cases. The case where we keep failing to open a substream might not be covered.

da-kami:
> are we sure that this was the reason force-closures were happening?

If by "this" you mean the endpoint not picking up on a broken connection then yes.
I do trust the user insisting that the container was running and the internet connection worked.

> I'm a bit sceptical it would fail to open substream for over 24h, if at the same time we had zero problems with rolling over other things... (and afterwards, his taker interacted again without restart on his part)

If it is the taker's endpoint that does not properly cleanup the connection (i.e. drop it and re-establish one if the connection is broken) then this can happen. I don't think it has anything to do with the maker.

klochowicz:
> > are we sure that this was the reason force-closures were happening?
> 
> If by "this" you mean the endpoint not picking up on a broken connection then yes. I do trust the user insisting that the container was running and the internet connection worked.
> 
> > I'm a bit sceptical it would fail to open substream for over 24h, if at the same time we had zero problems with rolling over other things... (and afterwards, his taker interacted again without restart on his part)
> 
> If it is the taker's endpoint that does not properly cleanup the connection (i.e. drop it and re-establish one if the connection is broken) then this can happen. I don't think it has anything to do with the maker.

this does not prove your hypothesis though, unless we see his logs. The connection could have been working, but `auto-rollover` did not trigger rollover successfully.

da-kami:
> but `auto-rollover` did not trigger rollover successfully.

Did we touch that logic when we introduced the new protocol? If so we should check it. Otherwise I would argue that something that was not changed is unlikely to break. The logic has quite a lot of unit tests, but we can re-evaluate there is an edge case that is not covered.

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
luckysori: There is only one place where we `remove` from `controls: HashMap<PeerId, (yamux::Control, Tasks)>`, which is when we call `drop_connection` which also notifies all subscribers that the connection was dropped. It's here:

https://github.com/itchysats/itchysats/blob/fb75a75349454685fe4c8fd932e86bfe8d72ff76/xtra-libp2p/src/endpoint.rs#L257-L260

In your code you're matching on a `NoConnection` error, which can only be emitted if the `PeerId` doesn't exist in the `controls` map. And what you are then doing is sending a `Disconnect` message which will _always do nothing_ because the `PeerId` doesn't exist in the `controls` map according to the error you just got from the `Endpoint`.

klochowicz:
> > but `auto-rollover` did not trigger rollover successfully.
> 
> Did we touch that logic when we introduced the new protocol? If so we should check it. Otherwise I would argue that something that was not changed is unlikely to break. The logic has quite a lot of unit tests, but we can re-evaluate there is an edge case that is not covered.

please have a look below what else could have happened that cannot be covered by the unit tests and would look exactly the same in the logs on the maker side:
- actor died (it's not supervised, so it would not restart),
- missing DLC,
- lack of connection to the DB,
- CFD got moved to 'closed' for some reason,
- no expiry timestamp,
- user somehow started collab-settlement that never finished (this is really unlikely, because we would tell us; moreover, that protocol  would have timed out quickly)

I hope it's not any of the above, but I have seen no proof yet. First two of the list are possible, others are listed just for completeness. All I say is that we should be vigilant, and perhaps increase measures of sending feedback (either via some protocol to the maker or sending debug logs to us if something fails hard). 

if I remember correctly, we have seen an instance where we had a missing DLC error at some point. And we for sure had panics triggered by some external libs as well. I will probably add supervision to this actor anyway, as this is a good idea.

xtra-libp2p-rollover/src/v_2_0_0/taker.rs
luckysori: > > A failed negotiation because the parties disagree on the protocols will not be solved by this.
> 
> Yes, and it should not be. I would accept that a taker with an incompatible rollover protocol will reconnect or treat that specific error separate.

But with this code change you are mandating that every time they disagree on the protocol the taker reconnects, which doesn't solve anything.

> > The listener may take too long to respond to the message, but the connection between the two parties is still completely fine. We are just adding more work to the listener to handle an unnecessary reconnect.
> 
> I don't think this is a valid argument for what the user experience. The user failed to roll-over for 24h resulting in a force close. This would mean the listener was too busy to accept the substream for 24h.

The problem is that we have no idea what actually happened, because we have no logs. We just have no actual proof that the problem is that the `taker` is not dropping the connection to the `maker` correctly. And even if it were the case, this would not be the right place to fix that.

As per my original message, if you drop the connection any time the `maker` is just lagging, you're making the situation worse for the lagging `maker`, who would have to handle the disconnect and the reconnect of this peer.

> > The dialer themself may take too long to perform the protocol! Perhaps the async work is not being scheduled in time because there's too much load on the application.
> 
> I don't think this is a valid argument for what the user experience. It is unlikely that the taker daemon was too busy to handle work within 24h. From our experience the taker daemon does not take many resources.

My original message is not trying to say that the user you're referencing experienced any of this. I'm just telling you that even if you were fixing that user's specific problem, you would be making matters worse in all situations where it doesn't apply.

> > Having looked at the yamux::ConnectionErrors above, I would be more interested in letting the Endpoint decide if it should drop a particular connection depending on the error (Closed and Io maybe?).
> 
> Thanks for having a look, I am happy to consider that. But I am not sure that would protect us from the endpoint not dropping a connection in some cases. The case where we keep failing to open a substream might not be covered.

_If_ we have a bug that causes takers to remain connected to a maker with a broken connection, I'm against fixing it in the rollover actor. That's what I'm advising.

bonomat:
> this does not prove your hypothesis though, unless we see his logs. The connection could have been working, but `auto-rollover` did not trigger rollover successfully.

Afair, the taker in question did not send anything to the maker for a while which to me is an indication that something was wrong with the connection. I think this was what Daniel motivated do this change. 

Are there alternative solutions to the problem where we just don't know?

bonomat:
We had recent errors in production: 

We have rollover problems with 0.5.0 and our favorite taker. We saw the following error: `Expected Msg2 wihtin 60 seconds`. If you want to look it up, these are the order IDs and taker ID
`715963df-e11f-4e90-ab69-399528916b2d`, `b1bed33e-4e8b-4dcb-8515-77c80cc1a67e`,`38cffef4-5eb0-4b37-b45d-a0e74c719b12`, takerId:  `10ed1d0b887ac90d8d66e0fc36131cacca4cf4b9562ba34e092fe38d77f8c84e`

luckysori:
> 715963df-e11f-4e90-ab69-399528916b2d

Looking at those logs makes me _less_ convinced that we need to act rashly and merge changes like these.

Over the last 2 days each of those CFDs had 3 rollover failures out of 24 total rollovers. They all happened at the same time (2pm, 6pm and 12am). And they all succeeded 5 minutes later.

Given that they all happened at the same time, the maker was probably struggling to process all requests for some reason. We have to fix that problem because I don't think there's any inherent reason why we should fail protocols every now and then. But it does not seem critical.

bonomat:
> But it does not seem critical.

I give this a solid 5 out of 7 on the level-of-criticalness-scale. 

More seriously: I'm not advocating for the changes discussed in this PR, but I believe we have to come up with a solution. For me, these kind of errors are critical because I can't say that it will always succeed eventually if I can't say why it was failing in the first place.

da-kami:
Since nobody seems to feel an emergency fallback for rollover is a good idea I am closing this PR at this point. 
Please refer back to https://github.com/itchysats/itchysats/issues/2554 for noting further opens or open a separate ticket if you feel the problem comes from somewhere else.

PR #2567, opened 2022-07-27
Add cargo-bloat CI action to determine bloat increase in new PRs

klochowicz:
This way we can make an informed decision whether additional bloat / binary size
in a new PR is acceptable or not.

### DIFF
--- a/.github/workflows/cargo-bloat.yml
+++ b/.github/workflows/cargo-bloat.yml
@@ -19,3 +19,4 @@ jobs:
         uses: orf/cargo-bloat-action@v1
         with:
           token: ${{ secrets.ITCHY_GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


bonomat:
It looks like this is still a draft and you are testing? If so, can you use a fork when you test new CI actions please. That way it does not block other CI actions in this repository.

klochowicz:
> It looks like this is still a draft and you are testing? If so, can you use a fork when you test new CI actions please. That way it does not block other CI actions in this repository.

of course, I can do this! sorry if it caused problems, I thought that if an action gets added in the PR it does not have affect on others, as it's not pushed to the main repo.

klochowicz:
migrated to my fork until it's ready: https://github.com/klochowicz/hermes/pull/1

bonomat:
> of course, I can do this! sorry if it caused problems, 

You didn't :)

PR #2568, opened 2022-07-27
Use git_version instead of vergen

klochowicz:
This popular crate uses `git` CLI instead of compiling whole libgit.
Output version is just as good.

The only minus is that we need `git` installed in order to compile the project.


bonomat:
> The only minus is that we need git installed in order to compile the project.

I think this is acceptable 😅

vergen-version/src/lib.rs
```diff
@@ -1,8 +1,11 @@
 /// Return git semver version
```
bonomat, line 1: I think we can rename this module to `version` or similar.

klochowicz: yes, indeed! just wanted to submit as draft first to see whether this actually gives us anything!

klochowicz:
hm, for some reason the crate is not being cached locally and it wants to rebuild it for the smoke-test... this is pretty weird. 🤔

klochowicz:
> hm, for some reason the crate is not being cached locally and it wants to rebuild it for the smoke-test... this is pretty weird. 🤔

 it must be invaliding the cache between runs, because it depends on something external (`git describe` command). This is a bit unfortunate; unless such a crate is at the very top of the dependency pile, it will cause recompilation of other dependent crates; it will likely not result in any speedup :/ I need to doublecheck whether `identify` protocol sends daemon version; if it does, then this PR is likely to worsen the performance further.

klochowicz:
Closing this PR as cache invalidation makes this unfeasible as a performance improvement, we might as well stick to vergen.

PR #2569, opened 2022-07-28
Identify protocol

### LINKED ISSUE #1769: Design proposal for `/itchysats/identify` protocol

### LINKED ISSUE #2307: Provide feedback about protocol between maker/taker incompatibility based on `identify` protocol

da-kami:
Resolves #1769 
Resolves https://github.com/itchysats/itchysats/issues/2307

In the legacy connection we had bound compatibility to the connection status, not we use the `identify` protocol to ensure compatibility.

Notes: 

- Takers don't disconnect if the maker does not speak the protocol, but we display a warning in the UI. This is more graceful, because if e.g. contract setup is not compatible but rollover is existing positions can still roll over.
- Makers currently don't care about connection information. Eventually we should change this. The maker should validate the *listening* protocols of the taker as well, otherwise a maker may repeatedly fail to e.g. end offers (if offers protocol is incompatible).

More generally speaking: We can invest more time into handling if the *expected* protocols of one side are not supported on the other side.


daemon/src/listen_protocols.rs
```diff
@@ -0,0 +1,243 @@
...
+            (self.ping_v1, ping_v1.into()),
+            (self.identify_v1, identify_v1.into()),
+            (self.rollover_v1, rollover_v1.into()),
+            (self.rollover_v2, rollover_v2.into()),
+            (
+                self.collaborative_settlement_v1,
+                collaborative_settlement_v1.into(),
+            ),
+        ]
+    }
```
luckysori: 👀 These `inbound_substream_handlers` methods are the thing that is not ideal, but I didn't know how to make it better.

The problem I see is that there are no guarantees that we use all of the identifiers in `self` (or that we associate them with the correct `Address`, but that seems impossible).

Thinking about it now we could destructure `Self` and the compiler will tell us if we haven't used all of the fields. And it will complain if we add a new one! If we leave a nice comment it should be more or less obvious how the pattern should be used then.

luckysori (approved):
Thank you for rebasing this :D

daemon/src/listen_protocols.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// when receiving new inbound substreams.
+    pub fn inbound_substream_handlers<R1, R2>(
+        &self,
+        ping_v1: Address<pong::Actor>,
+        identify_v1: Address<identify::listener::Actor>,
+        rollover_v1: Address<
+            rollover::v_1_0_0::maker::Actor<command::Executor, oracle::AnnouncementsChannel, R1>,
+        >,
+        rollover_v2: Address<
+            rollover::v_2_0_0::maker::Actor<command::Executor, oracle::AnnouncementsChannel, R2>,
```
da-kami, line 100: @luckysori just to make sure this is correctly rebased: The `oracle::AnnouncementsChannel` channel for rollover `v1` and `v2` is the same, correct?

luckysori: Yep. Both protocols communicate with the maker's `oracle::Actor`, and we only have one `oracle::Actor` at a time!

da-kami: That means the bound for `O` in the `rollover::...version...::maker::actor` was introduced because we don't want to have a dependency on deamon?
I felt it was needed for being generic initially, but for both the oracle actor and the executor it is purely to avoid dependencies. 
Only for the `Rates` we have two implementations that require to be generic.

luckysori: All of the trait bounds were introduced so that the protocol could be extracted into a separate crate without depending on `daemon`. Depending on `daemon` would defeat the purpose of extracting it into a separate crate and it would create a circular dependency!

But the trait bounds are also necessary if you want to test the protocol in isolation, even if you kept the protocol inside `daemon`.

daemon/src/listen_protocols.rs
da-kami: Nice idea! I applied it in the last `fixup` commit.
Additionally I applied the same pattern when converting into the `HashSet` for more assurance there.

I re-introduces the `NR_OF_SUPPORTED_PROTOCOLS` because it is another layer of checks we can apply and does not really add complexity.
Added a few more tests as well and fixed the `difference` function.

Would be great if you could have a look at the fixups :)

daemon/src/listen_protocols.rs
```diff
@@ -81,64 +91,65 @@ impl MakerListenProtocols {
...
         >,
-        collaborative_settlement_v1: Address<collab_settlement::maker::Actor>,
-    ) -> [(&'static str, MessageChannel<NewInboundSubstream, ()>); 5]
+        collaborative_settlement_v1_handler: Address<collab_settlement::maker::Actor>,
+    ) -> [(&'static str, MessageChannel<NewInboundSubstream, ()>); Self::NR_OF_SUPPORTED_PROTOCOLS]
     where
         R1: rollover::v_1_0_0::protocol::GetRates + Send + Sync + Clone + 'static,
         R2: rollover::v_2_0_0::protocol::GetRates + Send + Sync + Clone + 'static,
     {
+        // We deconstruct to ensure that all the protocols are being passed to the array below
```
luckysori: 🔧 We could do the same for `TakerListenProtocols`.

daemon/src/listen_protocols.rs
```diff
@@ -36,13 +39,15 @@ pub fn does_maker_satisfy_taker_needs(
     maker: &HashSet<String>,
     requirements: RequiredMakerListenProtocols,
 ) -> Result<(), HashSet<String>> {
-    let missing_protocols = maker
-        .difference(&HashSet::<String>::from(requirements))
+    // difference extracts the protocols that are in self but not in other, i.e. those that are in
+    // requirements but not in maker protocols are the ones missing
+    let missing_protocols = &HashSet::<String>::from(requirements)
+        .difference(maker)
```
luckysori, line 44: 👍 Thanks for the fix. I think the comment is unnecessary once you squash this, but that's just my opinion.

daemon/src/listen_protocols.rs
```diff
@@ -234,10 +249,52 @@ mod tests {
...
+
+        assert_eq!(
+            err,
+            HashSet::from([xtra_libp2p_ping::PROTOCOL_NAME.to_string()])
         )
-        .is_ok());
+    }
+
+    #[test]
+    fn given_maker_maker_supports_more_protocols_then_required_then_ok() {
```
luckysori: 🙃 First "then" should be "than".

luckysori (approved):
Thanks for adding tests for `does_maker_satisfy_taker_needs`

taker-frontend/src/App.tsx
```diff
@@ -120,6 +121,14 @@ export const App = () => {
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
     const connectedToMakerOrUndefined = useLatestEvent<ConnectionStatus>(source, "maker_status");
+    const makerCompatibilityOrUndefined = useLatestEvent<MakerCompatibility>(source, "maker_compatibility");
+
+    let incompatible = false;
+    if (makerCompatibilityOrUndefined) {
+        incompatible = makerCompatibilityOrUndefined.unsupported_protocols !== undefined
+            && makerCompatibilityOrUndefined.unsupported_protocols.length > 0;
+    }
```
klochowicz, line 130: nit: this could have been a single expression instead of an assignment and if block 🙃

daemon/src/identify/dialer.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        if let Some(peer_info_channel) = &self.peer_info_channel {
+            if let Err(e) = peer_info_channel.send(Some(peer_info)) {
+                tracing::warn!("Failed to send identity info to notify channel: {e:#}");
+            }
+        }
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
```
klochowicz:
```suggestion
#[xtra_productivity]
```

message_impl does not exist anymore :)

daemon/src/identify.rs
```diff
@@ -0,0 +1,173 @@
+use crate::Environment;
+use std::collections::HashSet;
+
+pub mod dialer;
+pub mod listener;
+pub mod protocol;
+
+pub const PROTOCOL: &str = "/itchysats/id/1.0.0";
```
klochowicz, line 8: not relevant to this PR, but a meta-comment : it's a bit sad that we can't reconcile between `PROTOCOL ` and `PROTOCOL_NAME`  :/ mixing 2 is unnecessary 😢

daemon/src/identify.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let taker_to_maker_peer_info = || taker_receiver.borrow().clone();
+        let maker_peer_info = retry_until_some(taker_to_maker_peer_info).await;
+
+        let maker_to_taker_peer_info = || maker_receiver.borrow().clone();
+        let taker_peer_info = retry_until_some(maker_to_taker_peer_info).await;
+
+        let expected_maker_peer_info = PeerInfo {
+            wire_version: "0.3.0".to_string(),
+            daemon_version: "0.4.22".to_string(),
```
klochowicz, line 102: would be good to havea test with some of the non-released `daemon-versions`. should we strip vergen version to bare last-released-version for identify purposes?

I'll write up a ticket in a sec detailing this behaviour. I believe it can be done in a follow-up though, no need to wait with merging.

daemon/src/identify/listener.rs
```diff
@@ -0,0 +1,70 @@
...
+            daemon_version,
+            environment,
+            identity,
+            listen_addrs,
+            protocols,
+        }
+    }
+}
+
+#[xtra_productivity(message_impl = false)]
```
klochowicz:
```suggestion
#[xtra_productivity]
```

daemon/src/identify/listener.rs
```diff
@@ -0,0 +1,70 @@
...
+        );
+
+        let send_identify_msg_fut = protocol::send(stream, identify_msg);
+
+        let err_handler = move |e| async move {
+            tracing::debug!(peer_id=%peer, "Identify protocol failed upon response: {e:#}")
+        };
+
+        let this = ctx.address().expect("we are alive");
+        spawn_fallible(&this, send_identify_msg_fut, err_handler);
```
klochowicz, line 61: should this rather be send_async_next?

klochowicz (approved):
thanks for tackling this, especially with all the rebases!

I've added some comments, happy to jointly deal with them as a follow-up :)

daemon/src/identify.rs
da-kami: please link the ticket if you can :)

daemon/src/identify/listener.rs
da-kami: hard to say. From the taker's perspective yes, from the maker's perspective where we may get a lot of `InboundSubstreamRequest`s at the same time probably no.

Generally I think we should re-think the pattern of the maker contacting all the takers, I am not sure it's a good pattern in the current system.

da-kami: Happy to evaluate a follow up change for this :)

daemon/src/identify.rs
da-kami: https://github.com/itchysats/itchysats/pull/2583 :)

PR #2570, opened 2022-07-28
Add liquidation actor tests

luckysori:
After the introduction of the `itchysats/order` protocol we can also prove that the CFD includes liquidation events from the moment it is open (not the case when we first merged https://github.com/itchysats/itchysats/pull/2378).

When adding these tests we discovered and fixed a bug which was causing us to only generate _short_ liquidation events.

---

I'm gonna create an issue to add more tests based on the `maker_position`, because that determines the shape of the payout curve, which affects the liquidation intervals.

Interestingly, here (maker going short) I only check the _long_ liquidation events, because the short liquidation events are pretty useless and hard to test: we have CETs for one specific price (as opposed to an interval for long liquidation CETs). This is because of the way our payout curve is implemented.

### DIFF
--- a/daemon-tests/tests/liquidation.rs
+++ b/daemon-tests/tests/liquidation.rs
@@ -57,7 +57,7 @@ async fn given_open_cfd_when_oracle_attests_long_liquidation_price_can_liquidate
         .attestation_for_event(first_liquidation_event)
         .unwrap();
 
-    simulate_attestation!(taker, maker, order_id, attestation.clone());
+    simulate_attestation!(taker, maker, order_id, &attestation);
     wait_next_state!(order_id, maker, taker, CfdState::PendingCet);
 
     confirm!(cet, order_id, maker, taker);
@@ -113,7 +113,7 @@ async fn given_rollover_when_oracle_attests_long_liquidation_price_can_liquidate
         .attestation_for_event(first_liquidation_event)
         .unwrap();
 
-    simulate_attestation!(taker, maker, order_id, attestation.clone());
+    simulate_attestation!(taker, maker, order_id, &attestation);
     wait_next_state!(order_id, maker, taker, CfdState::PendingCet);
 
     confirm!(cet, order_id, maker, taker);
--- a/daemon-tests/tests/liquidation.rs
+++ b/daemon-tests/tests/liquidation.rs
@@ -1,0 +1,143 @@
+use daemon::projection::CfdState;
+use daemon_tests::confirm;
+use daemon_tests::expire;
+use daemon_tests::flow::next_with;
+use daemon_tests::flow::one_cfd_with_state;
+use daemon_tests::maia::OliviaData;
+use daemon_tests::mock_oracle_announcements;
+use daemon_tests::open_cfd;
+use daemon_tests::simulate_attestation;
+use daemon_tests::start_both;
+use daemon_tests::wait_next_state;
+use daemon_tests::OpenCfdArgs;
+use model::Dlc;
+use model::Price;
+use otel_tests::otel_test;
+use rust_decimal::Decimal;
+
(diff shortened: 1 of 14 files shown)


daemon-tests/tests/liquidation.rs
```diff
@@ -0,0 +1,161 @@
...
+    .await;
+
+    assert!(
+        is_attestation_price_in_any_interval_of_all_liquidation_events(
+            &taker.latest_dlc(),
+            future_attestation_price
+        )
+    );
+
+    taker.system.commit(order_id).await.unwrap();
```
da-kami: 🙃 Given that from here both tests do the same we could also extract that into a `force_close` function. This might help us when adapting the code so we don't miss a spot. Not super important given only one duplication so far :)

daemon-tests/tests/liquidation.rs
```diff
@@ -0,0 +1,161 @@
...
+use rand::thread_rng;
+use rust_decimal::Decimal;
+
+#[otel_test]
+async fn given_open_cfd_when_oracle_attests_long_liquidation_price_can_liquidate() {
+    let (mut maker, mut taker) = start_both().await;
+
+    let oracle_data = OliviaData::example_0();
+
+    // We set the initial price to a value much higher than that of the price that the mock oracle
```
da-kami, line 18: Nifty! Took me a bit to understand the comment but makes total sense :)

daemon-tests/tests/liquidation.rs
```diff
@@ -0,0 +1,161 @@
...
+    confirm!(commit transaction, order_id, maker, taker);
+    wait_next_state!(order_id, maker, taker, CfdState::OpenCommitted);
+
+    expire!(cet timelock, order_id, maker, taker);
+
+    // We should be able to use any liquidation event to close the CFD
+    let random_liquidation_event = *taker
+        .latest_dlc()
+        .liquidation_event_ids()
+        .choose(&mut thread_rng())
```
da-kami: Could we potentially check all or at least some deterministic values that represent sane boundaries? 
I would like to avoid randomness when possible in the tests to avoid flaky tests.

da-kami (approved):
Nice! Great the test helped to detect a problem :)

daemon-tests/tests/liquidation.rs
luckysori: Fair point. I think checking all of them is a no-go because it either would take too long (24 times longer and even more once we have more frequent liquidation events) or require hacky tricks like erasing/overwriting events from the test DB.

Perhaps the lesson here is that we should not try to capture everything with this actor test. I will default back to just checking the first liquidation event. And I will consider adding a unit test somewhere else that verifies that all liquidation events are correct.

daemon-tests/tests/liquidation.rs
luckysori: It's quite hard to explain :( Maybe there's a different way to write this code so that it doesn't require such a complex comment.

daemon-tests/tests/liquidation.rs
luckysori: Thanks for suggesting this, I hadn't considered it!

daemon-tests/tests/liquidation.rs
da-kami: I thought about it when I had a look, but did not have a better idea so far 😅
I think it's fine :)

da-kami (approved):
Great! Love the proptests!

Would it be worth adding a ticket to investigate the documented bug further?

luckysori:
> Would it be worth adding a ticket to investigate the documented bug further?

Absolutely. I've opened https://github.com/itchysats/itchysats/issues/2586.

Issue #2385, opened 2022-07-08
Allow to release maker seperate from taker


bonomat:
I think there is high value in being able to release taker and maker separately because this would allow us to shoot out small fixes on the maker which do not influence the taker.

bonomat:
This project is unmaintained

PR #2573, opened 2022-07-28
Fix outdated info alert

da-kami:
When testing the new alertbox that I added for incompatibility for #2569 I noticed that this alerts were actually never shown due to the state not being updated as expected.
I reproduced this on master and fixed it, will rebase #2569 onto this work. 

Additionally applied layout fix to stick the alert to the top.

Before:

After:

---

Sidenote: I am actually not totally convinced we want this feature at the moment because our users will see this message as soon as we create a release. If we use this we should make sure that we only create a `taker` release if we are sure that it will go into Umbrel, otherwise users see this forever on Umbrel without being able to update 🤪
@bonomat thoughts?

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -84,7 +84,7 @@ export const App = () => {
         if (githubVersion && daemonVersion) {
             onOutdatedWarningOpen();
         }
-    }, [githubVersion, daemonVersion]);
+    }, [githubVersion, daemonVersion, onOutdatedWarningOpen]);
 
     const [source, isConnected] = useEventSource("/api/feed");
     const walletInfo = useLatestEvent<WalletInfo>(source, "wallet");
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -59,11 +59,6 @@ export const App = () => {
     const [githubVersion, setGithubVersion] = useState<SemVer | null>();
     const [daemonVersion, setDaemonVersion] = useState<SemVer | null>();
 
-    let outdated = false;
-    if (githubVersion && daemonVersion) {
-        outdated = githubVersion > daemonVersion;
-    }
-
     useWebSocket("wss://www.bitmex.com/realtime?subscribe=instrument:.BXBT", {
         shouldReconnect: () => true,
         onMessage: (message) => {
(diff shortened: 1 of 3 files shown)


bonomat:
> If we use this we should make sure that we only create a taker release if we are sure that it will go into Umbrel, otherwise users see this forever on Umbrel without being able to update 🤪
@bonomat thoughts?

The solution to this would be 
- #2385 
- and have releases that we really want to roll out :) 

I'd be happy to go for a different approach. We just need a possibility to tell the user to upgrade somehow :)

bonomat (approved):

taker-frontend/src/App.tsx
```diff
@@ -79,6 +74,18 @@ export const App = () => {
...
 
+    const {
+        isOpen: outdatedWarningIsVisible,
+        onOpen: onOutdatedWarningOpen,
+        onClose: onOutdatedWarningClose,
+    } = useDisclosure();
+
+    useEffect(() => {
+        if (githubVersion && daemonVersion) {
+            onOutdatedWarningOpen();
```
bonomat, line 85: Does this mean it's always open? :)

bonomat (changes requested):
Please don't merge, as we will need to fix this again after https://github.com/itchysats/itchysats/pull/2574 was merged. 

I've created https://github.com/itchysats/itchysats/pull/2578 which also fixes this on top of the other PR.

da-kami:
> > If we use this we should make sure that we only create a taker release if we are sure that it will go into Umbrel, otherwise users see this forever on Umbrel without being able to update 🤪
> > @bonomat thoughts?
> 
> The solution to this would be
> 
>     * [Allow to release maker seperate from taker #2385](https://github.com/itchysats/itchysats/issues/2385)
> 
>     * and have releases that we really want to roll out :)

I agree!

> I'd be happy to go for a different approach. We just need a possibility to tell the user to upgrade somehow :)

I don't think we have if we stick to the suggestions. 
My main concern was raining awareness because it would be super weird for users to see this info box for months without being able to do anything 😬

da-kami:
Superseded by https://github.com/itchysats/itchysats/pull/2578

taker-frontend/src/App.tsx
da-kami: No, because it only reacts on state changes of `[githubVersion, daemonVersion, onOutdatedWarningOpen]`, so if the user closes it and the state does not change then it will be closed.

PR #2576, opened 2022-07-28
Persist trading pair in the Cfd

### LINKED ISSUE #2458: Make daemon contract symbol aware from maker-to-taker
Right now we assume that every CFD is a BTCUSD position. In order to support different contract symbols the first step is to make the daemon aware of the symbol i.e. make sure that both sides know which symbol a CFD is for. 

aka: Send BTC/USD symbol all the way through from Maker to Taker.

klochowicz:
Thankfully, TradingPair was part of the Offer, so this change should be fully backwards-compatible.

- remove hardcoded BtcUsd in projection,
- persist trading_pair received from the Offer in all Cfd tables,
- default to TradingPair::BtcUsd in already created CFDs.

Fixes #2458

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce the `/itchysats/identify/0.3.0` protocol, running over libp2p.
   It allows a peer to share their contact details with others on request.
   It replaces a previous similar mechanism, which ran over a bespoke network stack.
+- Breaking change: renamed `trading_pair` to `contract_symbol` in Http API.
 
 ## [0.5.2] - 2022-07-26
 
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -255,6 +255,7 @@ impl Offer {
         funding_rate: FundingRate,
         opening_fee: OpeningFee,
         leverage_choices: Vec<Leverage>,
+        contract_symbol: ContractSymbol,
     ) -> Self {
         // allowing deprecated use of field `leverage_taker` here for backwards compatibility.
         #[allow(deprecated)]
@@ -265,7 +266,7 @@ impl Offer {
             max_quantity,
             leverage_taker: Leverage::TWO,
             leverage_choices,
-            contract_symbol: ContractSymbol::BtcUsd,
+            contract_symbol,
             position_maker,
             creation_timestamp_maker: Timestamp::now(),
             settlement_interval,
@@ -291,6 +292,7 @@ impl Offer {
             self.funding_rate,
             self.opening_fee,
             self.leverage_choices.clone(),
+            self.contract_symbol,
         )
     }
 
@@ -573,6 +575,7 @@ pub struct Cfd {
     role: Role,
     opening_fee: OpeningFee,
     initial_tx_fee_rate: TxFeeRate,
+    contract_symbol: ContractSymbol,
     // dynamic (based on events)
     fee_account: FeeAccount,
 
@@ -627,6 +630,7 @@ impl Cfd {
         opening_fee: OpeningFee,
         initial_funding_rate: FundingRate,
         initial_tx_fee_rate: TxFeeRate,
+        contract_symbol: ContractSymbol,
     ) -> Self {
         let (long_leverage, short_leverage) =
             long_and_short_leverage(taker_leverage, role, position);
@@ -657,6 +661,7 @@ impl Cfd {
             initial_funding_rate,
             opening_fee,
             initial_tx_fee_rate,
+            contract_symbol,
             dlc: None,
             cet: None,
             commit_tx: None,
@@ -707,6 +712,7 @@ impl Cfd {
             offer.opening_fee,
             offer.funding_rate,
             offer.tx_fee_rate,
+            offer.contract_symbol,
         )
     }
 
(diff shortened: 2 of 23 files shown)


sqlite-db/migrations/20220728115915_add_trading_pair.sql
```diff
@@ -0,0 +1,16 @@
+-- Introduce trading_pair field for all cfd tables.
+--
+-- Default to BTCUSD for all old already opened CFDs, as this is the only
+-- trading pair active for now.
+ALTER TABLE
+    cfds
+ADD
+    COLUMN trading_pair NOT NULL DEFAULT 'BtcUsd';
```
bonomat: We are starting to introduce different terminologies 😬 
In the metrics I've used the term `symbol`. 

I've created an ADR for this so that we are all on the same page: https://github.com/itchysats/architecture/pull/35/files

bonomat: I'm aware that I was the one who introduced `trading symbol` 🙈

sqlite-db/src/lib.rs
```diff
@@ -185,8 +187,9 @@ impl Connection {
             role,
             opening_fee,
             initial_funding_rate,
-            initial_tx_fee_rate
-        ) values ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13)"#,
+            initial_tx_fee_rate,
+            trading_pair,
```
da-kami: Here is where CI currently fails because of our friend the `,` :)

sqlite-db/src/closed.rs
```diff
@@ -129,7 +129,8 @@ impl Connection {
                 fees as "fees: models::Fees",
                 expiry_timestamp,
                 lock_txid as "lock_txid: models::Txid",
-                lock_dlc_vout as "lock_dlc_vout: models::Vout"
+                lock_dlc_vout as "lock_dlc_vout: models::Vout",
+                trading_pair as "trading_pair: models::TradingPair"
```
da-kami: 🛠 Something it odd here! we are loading it but not inserting it into `closed_cfds` - same for `failed_cfds`!

da-kami (dismissed):
I think `insert` for `closed_cfds` and `failed_cfds` is missing.

sqlite-db/migrations/20220728115915_add_trading_pair.sql
klochowicz: all good, I was just following what was written in the ticket.... will update when we decide on `symbol` vs `contract_symbol`.

sqlite-db/src/closed.rs
klochowicz: added! I noted that 2 other parameters from the Cfd are missing in FailedCfd - was that on purpose???

klochowicz:
I'll update when we decide whether to go with `contract_symbol` or `symbol`.

bonomat:
@klochowicz : I'm taking over as I believe you are on leave from today.

bonomat (approved):
Approving in the name of @klochowicz

klochowicz:
thanks for adding the rename!
ACK the additions :)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Replace intermediate confirmation step in rollover protocol for the maker with a configurable flag which can be updated during runtime using the new `POST /rollover/config` endpoint.
   On startup, rollovers are configured to be accepted by default.
+- Breaking change: renamed `trading_pair` to `contract_symbol` in Http API.
```
da-kami, line 24:
```suggestion
- Renamed `trading_pair` to `contract_symbol` in Http API.
```

🙃 At some point we should revisit our discussion about what a "breaking change" is. I would not state it like this in the changelog.
For us internally this is a breaking change, but for users it should not matter because the network protocols are not affected. 
If anybody looks at the changelog this looks like this version is incompatible to the one before, but that is misleading.

model/src/cfd.rs
```diff
@@ -196,7 +196,7 @@ impl From<Origin> for Role {
 pub struct Offer {
     pub id: OfferId,
 
-    pub trading_pair: TradingPair,
+    pub contract_symbol: ContractSymbol,
```
da-kami, line 200: 🛠 I think this is not only a breaking change in the HTTP API but a breaking change on the network layer! I think we need a `#[serde(rename = "trading_pair")]` here, otherwise already rolled out `0.5.x` won't be compatible.

da-kami (dismissed):

model/src/cfd.rs
da-kami: Even though this field is not used, I think we would still run into trouble deserializing the `Offer` if the maker sends it out like this, because the field will be missing on the taker side.

CHANGELOG.md
luckysori: Hmm, you are assuming that no one is trying to run their own `maker` automation then. And that this changelog is only meant to be consumed by users of the `taker` binary.

But I agree that our compatibility story is quite confusing.

klochowicz: AFAICT; the PUT request didn't have `trading_pair` - it was only for the UI

model/src/cfd.rs
klochowicz: good point. We need to use serde rename here indeed. I'll add a fixup.

CHANGELOG.md
bonomat: For me, the changelog is for developers but  yea, eventually we should revisit this discussion :)

da-kami (approved):

PR #2580, opened 2022-07-29
Move currency selector to trading screen

da-kami:
Just wanted to give this a quick crack to see how it would look, because after staring at @bonomat 's PR I felt the redundancy of the `symbol` in the buttons (`Long ETH` `Short ETH` ...) might be an indicator that we want the selector there. I gave it a try and found it not to bad. 
Still like the side bar and would keep it. @bonomat structured the selector really nice, so we could just move it if we want :)

Advantage:
- Easier to access
- More visible (especially on mobile)

Disadvantage:
- Always present on screen, users that are only interested in one pair might not want to see it at all times

---

<details>
<summary style="font-size:14px">Screenshot</summary>
<p>
image
</p></details>

<details>
<summary style="font-size:14px">Screenshot</summary>
<p>
image
</p></details>

### DIFF
--- a/taker-frontend/package.json
+++ b/taker-frontend/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "@chakra-ui/icons": "^2.0.4",
     "@chakra-ui/react": "^2.2.4",
-    "@emotion/react": "^11",
+    "@emotion/react": "^11.8.1",
     "@emotion/styled": "^11",
     "@feedback-fish/react": "^1.2.1",
     "@react-hookz/web": "^15.0.1",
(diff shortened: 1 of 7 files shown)


klochowicz:
I don't mind the selector, although I would not put it in-between long and short. it's super annoying, as you need to travel further between checking different prices now. (if you're not interested in changing the trading pair).

bonomat:
I'm with @klochowicz : I don't like the look of this 😬 Side note: I had it above the two buttons first but that _moves_ everything unnecessarily out of focus

PR #2585, opened 2022-08-01
Improve logs around network communication

luckysori:
We have several instances were we are:

- Logging the same thing more than once (objective).
- Logging at the wrong level (subjective).

I've stated several times that I believe we should be more selective with our usage of the `ERROR` level. In my mind, if I see an error log I should immediately understand that there's something to investigate. If I see a single warning I should be able to safely ignore it. And if I see many warnings of the same kind, I should start to suspect that there's a real problem.

### DIFF
--- a/daemon/src/collab_settlement/maker.rs
+++ b/daemon/src/collab_settlement/maker.rs
@@ -200,16 +200,13 @@ impl Actor {
                 let executor = self.executor.clone();
                 move |failed| async move {
                     match failed {
-                        Failed::BeforeReceiving { source } => {
-                            emit_failed(order_id, source, &executor).await;
+                        e @ Failed::BeforeReceiving { .. } => {
+                            emit_failed(order_id, anyhow!(e), &executor).await;
                         }
-                        Failed::AfterReceiving { source, settlement } => {
-                            // TODO: proceed with the transaction when taker will be able to handle that case.
-                            tracing::trace!(
-                        ?settlement,
-                        "Failed after receiving. Ideally, we should be able to act upon this settlement"
-                    );
-                            emit_failed(order_id, source, &executor).await;
+                        e @ Failed::AfterReceiving { .. } => {
+                            // TODO: proceed with the transaction when taker will be able to handle
+                            // that case.
+                            emit_failed(order_id, anyhow!(e), &executor).await;
                         }
                     }
                 }
--- a/daemon/src/collab_settlement/maker.rs
+++ b/daemon/src/collab_settlement/maker.rs
@@ -155,7 +155,9 @@ impl Actor {
 
                     let DialerSignature { dialer_signature } = framed
                         .next()
-                        .timeout(SETTLEMENT_MSG_TIMEOUT, || tracing::debug_span!("receive dialer signature"))
+                        .timeout(SETTLEMENT_MSG_TIMEOUT, || {
+                            tracing::debug_span!("receive dialer signature")
+                        })
                         .await
                         .with_context(|| {
                             format!(
@@ -181,7 +183,9 @@ impl Actor {
                     );
 
                     framed
-                        .send(ListenerMessage::ListenerSignature(ListenerSignature { listener_signature }))
+                        .send(ListenerMessage::ListenerSignature(ListenerSignature {
+                            listener_signature,
+                        }))
                         .await
                         .map_err(|source| Failed::AfterReceiving {
                             source: anyhow!(source),
@@ -233,7 +237,7 @@ impl Actor {
                     .await
             },
             move |e| async move {
-                tracing::debug!(%order_id, "Failed to reject collaborative settlement: {e:#}")
+                tracing::warn!(%order_id, "Failed to reject collaborative settlement: {e:#}")
             },
         );
         self.protocol_tasks.insert(order_id, tasks);
--- a/daemon/src/collab_settlement/maker.rs
+++ b/daemon/src/collab_settlement/maker.rs
@@ -120,7 +120,6 @@ impl Actor {
         let (transaction, proposal) = match result {
             Ok((transaction, proposal)) => (transaction, proposal),
             Err(e) => {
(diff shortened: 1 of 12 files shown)


model/src/cfd.rs
```diff
@@ -1328,7 +1328,7 @@ impl Cfd {
     }
 
     pub fn fail_collaborative_settlement(self, error: anyhow::Error) -> CfdEvent {
-        tracing::error!(order_id=%self.id(), "Collaborative settlement failed: {:#}", error);
+        tracing::warn!(order_id=%self.id(), peer_id=?self.counterparty_peer_id, "Collaborative settlement failed: {:#}", error);
```
da-kami: 🙃 I think so far we mostly opted for keeping the log in the `protocol` rather than in `model::Cfd`. I am not sure what is better, but we should probably agree on one or the other.

For me the logs should be part of the protocol actors, because a model should not log anything for me.

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
```diff
@@ -135,11 +135,10 @@ where
                 cfd.start_rollover_maker(propose.from_commit_txid)
             })
             .await
+            .context("Rollover failed after handling taker proposal")
         {
             Ok(base_dlc_params) => base_dlc_params,
             Err(e) => {
-                tracing::warn!(%order_id, "Rollover failed after handling taker proposal: {e:#}");
```
da-kami, line 141: Same: Not sure if it would not be better to let the protocol be in charge of the log rather than the model.

da-kami (approved):
Thank you for tackling this and removing all the duplicates.
I agree that all the warning logs should be warning logs, but I think our warning load will go up significantly with them. Noting to action here yet.

Feel free to ignore my comment. I am not sure if we should default to logging in the model rather than the protocol. My intuition says the protocol is actually more appropriate, but I don't have especially good arguments for that.

model/src/cfd.rs
luckysori: But this PR doesn't add any logs to `model::cfd`. It's only modifying one instance out of 19 pre-existing log statements that were already there.

I'm not against keeping the `model` crate effect-free, but that's not something that we've actually been doing since we introduced it.

luckysori:
> I agree that all the warning logs should be warning logs, but I think our warning load will go up significantly with them. Noting to action here yet.

It's a good point that I hadn't considered. I fixed the inconsistency because I don't want us to miss them. But if they are too spammy in the current state of the codebase maybe I should drop those commits. I'm not sure.

model/src/cfd.rs
da-kami: yeah, totally. I was just wondering if we should rather remove the logs from model crate and keep them in the other crates. It's probably something that we can discuss and then add a specific ticket. Feel free to ignore it for this PR :)

Restioson: Nit on this line: do we want to have some kind of consistent style with logs and tags? Are we going with no spaces between label and value or more like `order_id = %self.id()`

daemon/src/collab_settlement/maker.rs
```diff
@@ -156,7 +155,9 @@ impl Actor {
 
                     let DialerSignature { dialer_signature } = framed
                         .next()
-                        .timeout(SETTLEMENT_MSG_TIMEOUT, || tracing::debug_span!("receive dialer signature"))
+                        .timeout(SETTLEMENT_MSG_TIMEOUT, || {
```
Restioson, line 158: Are these changes from dprint or manual?

xtra-libp2p-offer/src/maker.rs
```diff
@@ -45,7 +45,7 @@ impl Actor {
         };
 
         let err_handler =
-            move |e| async move { tracing::debug!(%peer, "Failed to send offers: {e:#}") };
+            move |e| async move { tracing::warn!(%peer, "Failed to send offers: {e:#}") };
```
da-kami, line 48: We should totally have these on `warn`. I just wanted to compare the behavior of my `mainnet` taker with the maker logs and just don't have these logs because they are currently on debug and the default log level is info 😅

daemon/src/collab_settlement/maker.rs
luckysori: It was `dprint`. 

Funnily enough it only happens when I save the files using emacs, which calls `dprint` via [`dprint-fmt`](https://github.com/luckysori/dprint-fmt).

Sorry about the diff noise!

model/src/cfd.rs
luckysori: > Nit on this line: do we want to have some kind of consistent style with logs and tags? Are we going with no spaces between label and value or more like order_id = %self.id()

Good catch. I like having spaces too, I'll fix it. But anything that cannot be automatically enforced by `rustfmt` is just too hard to keep under control :(

daemon/src/collab_settlement/maker.rs
Restioson: Huh, that's a bit weird :thinking: I wonder how it originally passed CI without being properly formatted...

luckysori: My guess is that the workspace-level call to `dprint check` is okay with the file either way.

And the [file-level call](https://github.com/luckysori/dprint-fmt/blob/88ab37467a4a7a7d7660297b462b676239508485/dprint-fmt.el#L50) to `dprint fmt -stdin <file-path>` using the `dprint-fmt` emacs package sees things differently here, which is very weird but mostly harmless.

bonomat (approved):
Thanks for the clean-ups.

luckysori:
> Feel free to ignore my comment. I am not sure if we should default to logging in the model rather than the protocol. My intuition says the protocol is actually more appropriate, but I don't have especially good arguments for that.

Let's consider your point in a separate issue/discussion because I think it's valid. Going ahead with the merge.

luckysori:
I'm gonna remake this PR, because I have no idea what's going on!

Issue #2314, opened 2022-06-28
Consider `no-password` mode for Umbrel


bonomat:
Initially posted by @nevets963 in https://github.com/getumbrel/umbrel-apps/pull/50

> Would you also consider removing the requirement for a password? In Umbrel v0.5 and onwards, we automatically password protect (including 2FA where applicable) all installed apps. If you remove the password requirement, we can improve the user experience as they then don't need a secondary password for ItchySats.

klochowicz:
I'm all for no-password mode, as long as it's an opt-in feature on mainnet ;) 

it can be useful for testing and for playing around on testnet (reducing the friction to try out itchysats)!

bonomat:
I start to disagree with this feature. The login function of ItchySats is a security feature and we should not rely on the platform we are running on doing this for us. Same as your personal computer: you still have passwords for various programs despite having a password on system level.

holzeis:
Does umbrel support openID connect? in that case we could outsource the autentication, improve the ux without loosing security.

bonomat:
With #2903 we went one step further away from this and added cookie-based authentication. The session is set to expire in 1 week, meaning, the user does not have to login that often.

PR #2599, opened 2022-08-01
Change password to `dev` in dev setup

da-kami:
Makes it easier to use the `start_all` script.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,7 @@ linker = "aarch64-linux-gnu-gcc"
 
 [alias]
 dev-maker = "run --bin maker -- --instrumentation testnet"
-dev-taker = "run --bin taker -- --instrumentation --maker localhost:9999 --maker-id 10d4ba2ac3f7a22da4009d813ff1bc3f404dfe2cc93a32bedf1512aa9951c95e --maker-peer-id 12D3KooWDjzHna3pNi1Bt1DoRfrpsBREykJKXDRDxXvhJNAdDZEk testnet" # Maker ID matches seed found in `testnet/maker_seed`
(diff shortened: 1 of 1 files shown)


da-kami:
Made me think: Should we add this commandline option to the `maker` as well?

klochowicz:
> Made me think: Should we add this commandline option to the `maker` as well? 

Yup, there's no reason why only one should have the convenience! 

Actually, for dev purposes the best would do no password mode, suggested recently. Maybe we could do that instead?

bonomat (approved):

holzeis (approved):
Tested on my machine, all fine. Maker UI did not ask for a username/password though, but that wasn't part of the change.

da-kami:
> Tested on my machine, all fine. Maker UI did not ask for a username/password though, but that wasn't part of the change.

Note: When the UI is automatically opened upon startup of the daemon we call the default browser with a link including username and password, so you won't have to enter it. But when you close and re-open the UI in the browser you may need the password (unless the credentials were saved).

da-kami:
> > Made me think: Should we add this commandline option to the `maker` as well?
> 
> Yup, there's no reason why only one should have the convenience!
> 
> Actually, for dev purposes the best would do no password mode, suggested recently. Maybe we could do that instead?

I like the idea of a no-password mode because that is what we actually want here. 
One thing that came to mind when thinking about it: Should that disable basic-auth (i.e. there is no basic auth in no-password mode) or just set the password to `""`? Do we expect the step where we have to enter the default user name? 
I think the way this was thought about in https://github.com/itchysats/itchysats/issues/2314 would be turning basic auth off. We can add to the discussion on that ticket :)

I am merging the current change, we can add further changes on top.

luckysori:
bors retry

da-kami:
bors retry

da-kami:
bors retry

PR #2600, opened 2022-08-01
Replace confusing helper text

da-kami:
We display buy/sell in the context of both long and short which is confusing.
We should generally avoid this terminology here.

--- 

Was noticed when looking at this with @holzeis 
I tried to come up with a better helper text, but am not super happy about it. If you have a better idea feel free to suggest it.

### DIFF
--- a/taker-frontend/src/components/Trade.tsx
+++ b/taker-frontend/src/components/Trade.tsx
@@ -338,7 +338,7 @@ function Quantity({ min, max, onChange, quantity, lotSize, isLong }: QuantityPro
                     </NumberInputStepper>
                 </NumberInput>
             </InputGroup>
-            <FormHelperText>How much do you want to buy or sell?</FormHelperText>
+            <FormHelperText>How many contract should your position contain?</FormHelperText>
         </FormControl>
     );
 }


luckysori (commented):
What if  we make the original message dependent on the type of position that you want to open?

taker-frontend/src/components/Trade.tsx
```diff
@@ -338,7 +338,7 @@ function Quantity({ min, max, onChange, quantity, lotSize, isLong }: QuantityPro
                     </NumberInputStepper>
                 </NumberInput>
             </InputGroup>
-            <FormHelperText>How much do you want to buy or sell?</FormHelperText>
+            <FormHelperText>How many contract should your position contain?</FormHelperText>
```
bonomat, line 341: As usual, this was inspired by our friendly friend BitMEX 😀.
Can you elaborate what was confusing about this?

image

Restioson: I'm not quite sure "how many contract should your position contain" means exactly. Firstly, should it be "contracts" (plural)? Secondly, I'm just not sure what "one contract" would mean.

da-kami:
Maybe it was just me that found it confusing.

I think this still stems from `buy/sell` vs `long/short` and if the terms are equivalent.

I find buy/sell useful in a context where I exchange something, because I i.e. buy BTC and sell USD.
In the context of a CFD I find it somewhat confusing, because I buying and selling is an implied action there, rather than actually happening; because in the end I always have BTC. 

Additionally, I find sell somewhat confusing here, because it is not immediately obvious to me that when opening a short position I am selling. (What am I selling?)
This might just be in my head. Closing this for now, if you feel it is worth changing please re-open or propose a new PR.

PR #2581, opened 2022-07-29
Produce a single binary for integration tests

klochowicz:
If we have multiple files here, they end up in different test binaries.
This means that we have to link everything separately.
And linking so much is slow (and likely single-threaded).

See: https://endler.dev/2020/rust-compile-times/#combine-all-integration-tests-in-a-single-binary

PR #2604, opened 2022-08-01
Fix #2581

Restioson:
This implements my proposed fix detailed in #2581. This should maybe get squashed into the main commit of that PR


Restioson:
Oops, wrong branch

PR #2610, opened 2022-08-02
Remove needless sleeps in `daemon-tests`

### LINKED ISSUE #2309: Remove arbitrary sleep length from actor tests
I think it would be better to do a `retry_until` logic where we only sleep a short period of time and re-check frequently if the value already changed.

Something similar to this: https://github.com/itchysats/itchysats/blob/f270f0353d84f164a3701952f4118886fa5d2fec/xtra-libp2p-ping/src/lib.rs#L116-L127

That way we don't waste test time - this test was already super long running.

_Originally posted by @da-kami in https://github.com/itchysats/itchysats/pull/2270#discussion_r908222761_

luckysori:
These sleeps always precede a call to `wait_next_state`, which already waits until the CFD reaches a particular state.

---

I've always been confused by this. If I understand it correctly, `wait_next_state` will wait until the given state is reached (even if there are other states in between!), so the sleeps should be superfluous.

### DIFF
--- a/daemon-tests/tests/connectivity.rs
+++ b/daemon-tests/tests/connectivity.rs
@@ -1,12 +1,10 @@
 use daemon::online_status::ConnectionStatus;
-use daemon_tests::flow::next;
+use daemon_tests::flow::next_with;
 use daemon_tests::Maker;
 use daemon_tests::MakerConfig;
 use daemon_tests::Taker;
 use daemon_tests::TakerConfig;
 use otel_tests::otel_test;
-use std::time::Duration;
-use tokio_extras::time::sleep;
 
 #[otel_test]
 async fn taker_notices_lack_of_maker() {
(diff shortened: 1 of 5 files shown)


daemon-tests/tests/connectivity.rs
```diff
@@ -24,30 +22,27 @@ async fn taker_notices_lack_of_maker() {
...
+    wait_next_connection_status_to_maker(&mut taker, ConnectionStatus::Offline).await;
 
     let _maker = Maker::start(&maker_config).await;
 
-    sleep(Duration::from_secs(5)).await; // wait a bit until taker notices change
+    wait_next_connection_status_to_maker(&mut taker, ConnectionStatus::Online).await;
+}
+
+/// Wait indefinitely until the `taker`'s connection status to the maker is the one `expected` by
+/// the caller.
```
klochowicz, line 37: shouldn't we add a timeout to these invocations? otherwise they will just hang in the error case now?

note: we could apply this in a clean way in nextest configuration, e.g. by giving a generous timeout of 5 minutes or so (to make sure the triple rollover tests are not failing)

da-kami (approved):
Interesting! I recently removed them as well and the tests were failing locally for me afterwards so I did not open a PR. Probably a stupid coincidence, great that we can remove them!

klochowicz (approved):
thanks for cleaning this!

da-kami:
Given recent CI test failures I begin to doubt that the sleeps were useless. I remember removing them locally and running into flaky tests.

PR #2613, opened 2022-08-02
Configure test timeout in nextest

### LINKED ISSUE #2611: add test timeout in nextest
shouldn't we add a timeout to these invocations? otherwise they will just hang in the error case now?

note: we could apply this in a clean way in nextest configuration, e.g. by giving a generous timeout of 5 minutes or so (to make sure the triple rollover tests are not failing)

_Originally posted by @klochowicz in https://github.com/itchysats/itchysats/pull/2610#discussion_r935576490_

klochowicz:
Warn about slow tests after 120s and terminate tests if they run for 5 minutes.

These values apply both when run locally and on the CI.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,6 @@
+[profile.default]
+slow-timeout = { period = "120s", terminate-after = 5 }
+
 [profile.ci]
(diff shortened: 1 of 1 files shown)


bonomat (approved):

luckysori (approved):
Nice.

PR #2614, opened 2022-08-02
Colour terminal output on the CI

klochowicz:
Spotted somewhere in nextest examples.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   CARGO_INCREMENTAL: 0
+  CARGO_TERM_COLOR: always
 
 jobs:
   formatting:


bonomat (approved):

holzeis (approved):

luckysori (approved):
LGTM but no idea why CI fails.

Restioson (approved):
CI fail is weird - seems to be a yarn cache issue?

da-kami:
As weird as it is this error is persistently happening on retries. I just retried the failed tasks twice with the same result. 
I think this is the problem: https://github.com/yarnpkg/yarn/issues/5945

luckysori:
> As weird as it is this error is persistently happening on retries. I just retried the failed tasks twice with the same result. I think this is the problem: [yarnpkg/yarn#5945](https://github.com/yarnpkg/yarn/issues/5945)

Nice find! Let's drop the second commit then.

Issue #2459, opened 2022-07-19
`itchysats/offer` to support different trading pairs


bonomat:
Update `itchysats/offer` protocol to handle multiple long and short offers (with different trading pair).

PR #2615, opened 2022-08-03
Make maker/taker HTTP api contract symbol aware

### LINKED ISSUE #2460: Upgrade maker http API
Upgrade maker http API to allow to create new offers with different trading pairs

### LINKED ISSUE #2461: Upgrade taker http API
Upgrade taker http API to allow to take new offers

### LINKED ISSUE #2462: Upgrade Maker UI to create new offers

bonomat:
Note: this is a breaking change of the HTTP Api. We will need to update auto-taker to be able to deal with this. 

Resolves #2460 
Resolves #2461 
Resolves #2462


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Add dynamic liquidation to the DLC, enabling the CFD to be unilaterally closed every hour by either party if the oracle attests to a price close to the ends of the payout curve's domain.
+  This is currently only active _after the first rollover_.
```
luckysori: 🔧 Unwanted added line that I recently removed.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- New dynamic route for maker `PUT /<contract_symbol>offer/`. This allows the creation of offers with `<contract_symbol>`.
```
luckysori: 🔧 Should be `PUT /<contract_symbol>/offer/` i.e. one `/` is missing.

daemon/src/projection.rs
```diff
@@ -83,7 +83,7 @@ pub struct Actor {
 
 pub struct Feeds {
     pub quote: watch::Receiver<Option<Quote>>,
-    pub offers: watch::Receiver<MakerOffers>,
+    pub offers: watch::Receiver<(ContractSymbol, MakerOffers)>,
```
luckysori: 🤔 `MakerOffers` already has a `ContractSymbol` inside.

daemon/src/taker_cfd.rs
```diff
@@ -86,7 +86,10 @@ impl Actor {
                 short
             });
 
-            maker_offers
+            (
+                maybe_contract_symbol.unwrap_or(ContractSymbol::BtcUsd),
```
luckysori: ❓ Wait, where's `maybe_contract_symbol` coming from?

Besides that, I would expect the contract symbol to not be optional here.

maker/src/cfd.rs
```diff
@@ -337,11 +340,11 @@ where
...
             })?;
 
-        // 1. Validate if order is still valid
+        // 1. Validate if offer is still valid
         let order_to_take = self
             .current_offers
-            .as_ref()
-            .and_then(|offers| offers.pick_offer_to_take(order_id));
+            .values()
+            .find_map(|offer| offer.pick_offer_to_take(order_id));
```
luckysori, line 347: 🤔 Maybe it's time to remove all this code related to the old connection, because I don't even want to review this logic :D

maker/src/cfd.rs
```diff
@@ -622,12 +638,14 @@ impl<O, T, W> Actor<O, T, W> {
     }
 
     async fn handle_accept_rollover(&mut self, msg: AcceptRollover) -> Result<()> {
+        let order_id = msg.order_id;
```
luckysori, line 641: 😞 I need to do something about https://github.com/itchysats/itchysats/pull/2552 so that we don't have to think about this.

maker/src/routes.rs
```diff
@@ -138,7 +138,68 @@ pub async fn put_offer_params(
...
+            ContractSymbol::BtcUsd.into(),
+        )
+        .await
+        .map_err(|e| {
+            HttpApiProblem::new(StatusCode::INTERNAL_SERVER_ERROR)
+                .title("Posting offer failed")
+                .detail(format!("{e:#}"))
+        })?;
+
+    Ok(())
```
luckysori, line 154: 🙃 Perhaps we could log something to mark this as deprecated?

maker/src/routes.rs
```diff
@@ -138,7 +138,68 @@ pub async fn put_offer_params(
...
+pub async fn put_offer_params_for_symbol(
+    symbol: Result<ContractSymbol>,
+    offer_params: Json<CfdNewOfferParamsRequest>,
+    maker: &State<Maker>,
+    _auth: Authenticated,
+) -> Result<(), HttpApiProblem> {
+    // if we use `ContractSymbol` as arg directly the error gets lost. So we need to do this:
+    let symbol = symbol.map_err(|e| {
+        HttpApiProblem::new(StatusCode::BAD_REQUEST)
+            .title("Posting offer failed")
```
luckysori: ❓ Should we have a different title since we know that the error is about the `ContractSymbol` being incorrect?

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
```diff
@@ -174,6 +174,12 @@ where
             tracing::debug_span!("next rollover message")
         }
 
+        let contract_symbol = self
```
luckysori: 📣 I think this is not how the executor is meant to be used, because there is nothing to execute here!

Luckily, we can access the contract symbol the next time we actually execute a command (a few lines below). I would call `rates.get_rates` inside that other closure.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   It allows a peer to share their contact details with others on request.
   It replaces a previous similar mechanism, which ran over a bespoke network stack.
 - Breaking change: renamed `trading_pair` to `contract_symbol` in Http API.
+- Introduce path param to SSE event route for both maker and taker. The new route is `<contract_symbol>/feed`.
```
luckysori: 🙃 I would prefer `/<contract_symbol>/feed`.

And maybe we could explain what this new feature does, in case anyone actually read this.

daemon-tests/src/flow.rs
```diff
@@ -3,18 +3,24 @@ use anyhow::Result;
...
 use model::OrderId;
+use std::collections::HashMap;
 use std::time::Duration;
 use tokio::sync::watch;
 
 /// Waiting time for the time on the watch channel before returning error
 const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) { 90 } else { 30 });
 
-/// Wait and return next non-empty maker offers
+/// Wait and return next non-empty maker offers with `symbol`
```
luckysori: 🔧 Should be `contract_symbol`.

taker/src/routes.rs
```diff
@@ -78,6 +108,23 @@ pub async fn feed(
...
+            Some(rx_offers) => rx_offers.clone(),
+        };
+        let mut rx_quote = match rx_quote {
+            None => {
+                // This should never happen as long as we instantiate `rx_quote` for each ContractSymbol variant
+                tracing::error!("No quote receiver for {symbol} found");
+                return ();
+            }
+            Some(rx_quote) => rx_quote.clone(),
+        };
```
luckysori: 🔧 I would opt to have a dedicated `rx_{btcusd,ethusd}_offers` receivers. We can avoid this annoying boilerplate and I don't think we will ever miss the hash-map's dynamism.

taker/src/routes.rs
```diff
@@ -56,9 +57,37 @@ pub struct IdentityInfo {
...
+pub enum ContractSymbol {
+    BtcUsd,
+    EthUsd,
+}
+
+impl From<ContractSymbol> for model::ContractSymbol {
+    fn from(symbol: ContractSymbol) -> Self {
+        match symbol {
+            ContractSymbol::BtcUsd => model::ContractSymbol::BtcUsd,
+            ContractSymbol::EthUsd => model::ContractSymbol::BtcUsd,
```
luckysori: 🔧 Should be `model::ContractSymbol::EthUsd`. Or `todo!()` if it doesn't exist yet, so that my brain doesn't explode.

maker/src/routes.rs
```diff
@@ -150,15 +168,17 @@ pub async fn put_offer_params(
...
 pub enum ContractSymbol {
     BtcUsd,
+    EthUsd,
 }
 
 impl From<ContractSymbol> for model::ContractSymbol {
     fn from(symbol: ContractSymbol) -> Self {
         match symbol {
             ContractSymbol::BtcUsd => model::ContractSymbol::BtcUsd,
+            ContractSymbol::EthUsd => model::ContractSymbol::BtcUsd,
```
luckysori: 🔧 Prefer a `todo!()` here.

daemon/src/taker_cfd.rs
```diff
@@ -86,20 +87,25 @@ impl Actor {
...
 
         self.current_maker_offers = takers_perspective_of_maker_offers.clone();
         tracing::trace!("new maker offers {:?}", takers_perspective_of_maker_offers);
 
+        // If there's no offer, it might as well be hard-coded to BTC
+        let symbol = self
+            .current_maker_offers
+            .clone()
+            .and_then(|offer| offer.long.map(|offer| offer.contract_symbol))
+            .unwrap_or(ContractSymbol::BtcUsd);
```
luckysori, line 101: 🤔 Surely this is wrong. We are getting an update from the maker about a particular type of offer here. We can't just assume that it's `BTCUSD` just because we don't know the contract symbol! AFAIK we need to use this information properly in order to disable certain offers.

I think we need to rework this and not just build on top of it. The `MakerOffers` type just doesn't work for what we are trying to do. I think something like this could work better:

```rust
struct MakerOfferInfo {
    pub contract_symbol: ContractSymbol,
    long: Option<OfferParams>,
    short: Option<OfferParams>,
    fee_params: Option<FeeParams>,
}

struct OfferParams {
// all the fields in `Offer` minus `contract_symbol`,
// `position_maker`, `tx_fee_rate`, and `funding_rate`
}

struct FeeParams {
    tx_fee_rate: TxFeeRate,
    funding_rate_long: FundingRate,
    funding_rate_short: FundingRate,
}

impl MakerOfferInfo {
    // based on `self.long` and `self.fee_params`
    fn long_offer(&self) -> Option<Offer>;

    // based on `self.short` and `self.fee_params`
    fn short_offer(&self) -> Option<Offer>;

    fn fee_params(&self) -> Option<FeeParams>;
}
```

This should also give us maximum flexibility to allow rollovers whilst disabling new orders if we want to. And keeping all fields orthogonal helps us avoid inconsistencies and makes this type clearer.

luckysori (commented):
Thank you for pushing this forward! I've focused my review on the backend.

I think we need to tackle some technical debt before merging this. Primarily, reworking `MakerOffers`.

daemon/src/taker_cfd.rs
bonomat: We unfortunately can't easily remove `contract_symbol` from `Offer` because that's being sent over the wire( backwards compatibility). 
I could try to get rid of `MakerOffers` and split out `FundingRate` but that's a bigger change. I'll give this a try

bonomat: No, doesn't work because we send `MakerOffers` over the wire ... :/

bonomat: Actually, the protocol does not work with multiple symbols because it assumes `LatestMakerOffers` which has an `Option<MakerOffers>`. If this inside is `None`, it would set the offers inside the taker to `None`. Hence, `LatestMakerOffers` would need a field `contract_symbol`. 

This should be the scope of https://github.com/itchysats/itchysats/issues/2459 and changing this here would kind of explode in scope. 

What do you think?

maker/src/routes.rs
bonomat: That was just not done 🙈

taker/src/routes.rs
bonomat: The advantage of the hash-map is that we should be able to just add a new contract symbol by just adding a new enum variant

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
bonomat: I could not make it work  because I think I'd need to spin up a `task` inside the other executor because `get_rates` is async .

daemon/src/projection.rs
bonomat: Unfortunately not, `MakerOffers.long` _might_ have it, but it's an option.

daemon/src/taker_cfd.rs
bonomat: looks like a mistake. Should be fixed now.

maker/src/cfd.rs
bonomat: Totally... shall we just remove it 😬

luckysori: I vote yes!

daemon/src/projection.rs
luckysori: Which is why I think we need to rework `MakerOffers` D:

daemon/src/taker_cfd.rs
luckysori: We discussed this offline and the conclusion was:

- The current state of this PR is not ideal, but we can basically merge it given that we are gonna tackle the concerns I raised with #2459.
- @bonomat is gonna try to split the `MakerOffers` into two types: one for the offer protocol (single contract symbol) and one for the rest of the application (multiple contract symbols).

bonomat: I'm afraid removing the `contract_symbol` from `Offer` is not easy. We rely on this information when placing an order: the taker places the order only using the `order id` with which we load the `offer` and create a `Cfd`. For creating the `Cfd` we need to know the `contract_symbol` which is not available there....

taker/src/routes.rs
bonomat: Done. I think it turned out nice.

bonomat:
@luckysori : I rebased and force pushed. Sorry that you have to go through this again.

daemon/src/taker_cfd.rs
luckysori: Maybe there was some miscommunication. I don't think we should remove the `ContractSymbol` from the `Offer`! My proposal was to introduce a _new_ `OfferParams` type that is `ContractSymbol`-agnostic. One can then compute an `Offer` from the proposed `MakerOfferInfo` only if all the corresponding parameters are available, by calling either `long_offer` or `short_offer`.

Does that make sense?

daemon/src/projection.rs
```diff
@@ -82,11 +82,21 @@ pub struct Actor {
...
     pub cfds: watch::Receiver<Option<Vec<Cfd>>>,
 }
 
+pub struct QuoteFeed {
+    pub btc_usd: watch::Receiver<Option<Quote>>,
+    pub eth_usd: watch::Receiver<Option<Quote>>,
+}
+
+pub struct OffersFeed {
+    pub btc_usd: watch::Receiver<MakerOffers>,
```
da-kami, line 96: ❓ What is the motivation of splitting the complete feed and not use the same feed but use different events?

daemon/src/projection.rs
```diff
@@ -1165,8 +1187,9 @@ impl Actor {
         );
     }
 
-    fn handle(&mut self, msg: Update<Option<model::MakerOffers>>) {
-        self.tx.send_order_update(msg.0);
+    fn handle(&mut self, msg: Update<(ContractSymbol, Option<model::MakerOffers>)>) {
+        self.state.update_offer(msg.0 .0, msg.0 .1.clone());
```
da-kami: ❓This seems to be a semantic change, previously we did not update the offer here. Is this on purpose? Is this actually fixing something that was broken before?

daemon/src/projection.rs
```diff
@@ -1245,14 +1268,47 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
...
+        let (long, short) = match offers {
+            None => (None, None),
+            Some(offers) => {
+                let projection_long =
+                    offers
+                        .long
+                        .and_then(|long| match TryInto::<CfdOffer>::try_into(long) {
+                            Ok(projection_long) => Some(projection_long),
+                            Err(e) => {
+                                tracing::warn!("Unable to convert long order: {e:#}");
```
da-kami:
```suggestion
                                tracing::warn!("Unable to convert long offer: {e:#}");
```

daemon/src/projection.rs
```diff
@@ -1245,14 +1268,47 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
...
+                            }
+                        });
+
+                let projection_short =
+                    offers
+                        .short
+                        .and_then(|short| match TryInto::<CfdOffer>::try_into(short) {
+                            Ok(projection_short) => Some(projection_short),
+                            Err(e) => {
+                                tracing::warn!("Unable to convert short order: {e:#}");
```
da-kami:
```suggestion
                                tracing::warn!("Unable to convert short offer: {e:#}");
```

daemon/src/taker_cfd.rs
da-kami: [snip]
> Hence, LatestMakerOffers would need a field contract_symbol.
> This should be the scope of #2459 and changing this here would kind of explode in scope.

I don't think that this change would explode in scope and until we refactor this overall it would be preferable over the `unwrap_or`.

Tbh: I was unable to fully understand the later comments with @luckysori

maker/src/cfd.rs
da-kami: If we do so please bump the network version as well, so that we "force" users to upgrade by making the complete old network stack incompatible. If we do so it should be an active decision and we should properly remove all old code - we started in some separate PRs I saw, but I feel there is no consensus on this topic yet.

I am OK to remove the old network code with the next release and force users to upgrade. But be aware that if they don't upgrade within 24h we would see force closes then. It might be better to stick to the plan and roll out another release where all the libp2p protocols are in. Then roll out another release after some time (~1 to 2 weeks) that forces the upgrade...

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
da-kami: I agree with @luckysori that this code is super confusing and I simply don't understand why it is necessary at this point. If you need information about the contract symbol, then that information is available on the `cfd`. Why can'y you just pass the `contract_symbol` out when you call `cfd.start_rollover_maker(propose.from_commit_txid)`? 
(as part of `base_dlc_params` or as separate parameter)

daemon/src/projection.rs
bonomat: I think it is cleaner to have different streams.

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
bonomat: Wrapping it in `BaseDlcParams` sounds like a different hack to me. This struct is concerned about `Parameters associated with the base DLC involved in a rollover`. I can do that but I think this goes into the wrong direction and puts things together which do not really belong together.

Even though we are _misusing_ the executor to extract something from the CFD I think this is the cleanest of all ugly solutions. 

@luckysori : proposed to create a follow-up ticket which tackles this in a different way.

daemon/src/taker_cfd.rs
bonomat: > Does that make sense?

Not yet. What's the advantage of this if we still need to keep the old struct `Offer` around?

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
bonomat: I've added one more commit which decouples the models from the network layer types. This allows us to easier to for #2459

daemon/src/projection.rs
bonomat: Good catch. That's something Mariusz introduced. The field doesn't seem to be read anywhere. So I'll just remove it again.

daemon/src/taker_cfd.rs
luckysori: I'm working under the assumption that `Offer` is a type that we can't easily modify because of DB compatibility. Furthermore, and offer is always about a particular contract symbol, so there's nothing wrong with the type.

The problem that I see is that the thing that we send around for the `/itchysats/offer` protocol, `Option<MakerOffers>` doesn't work well with multiple contract symbols:

- If we send `None`, the receiver doesn't know what we are talking about.
- If we send `MakerOffers` and the `long` and `short` offers are set `None` (in order to disable the creation of new orders), the receiver doesn't know how to use the `tx_fee_rate`, `funding_rate_long` and `funding_rate_short` fields for rollovers, because it's missing the contract symbol context.
- The `long` and `short` offer fields aren't guaranteed to be talking about the same contract symbol, which is confusing and could lead to errors. This highlights the importance of extracting the contract symbol into a common place.

That's why I propose to do something like what I said in https://github.com/itchysats/itchysats/pull/2615#discussion_r936668903. The advantages I see are:

- The taker will always know the contract symbol of the offer information they're receiving from the maker.
- The contract symbol at the top level of the struct is the only source of truth. Because the proposed `OfferParams` types don't talk about the contract symbol, they can never disagree with the top level. Consumers cannot access these inner fields and instead we provide `long_offer` and `short_offer` methods that reconstruct the `Option<Offer>` type based on what's available. The `Offer` type therefore doesn't have to change.
- If the proposed field `fee_params` is `None`, we can disable new orders and rollovers. 

We might want to move something like the `creation_timestamp` to the top level as well, if we want to be able to consider all the offer information stale.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Add dynamic liquidation to the DLC, enabling the CFD to be unilaterally closed every hour by either party if the oracle attests to a price close to the ends of the payout curve's domain.
+- Add dynamic liquidation to the DLC, enabling the CFD to be unilaterally closed every hour by either party if the oracle attests to a price close to the ends of the payout curve's domain. This is currently only active _after the first rollover_.
```
luckysori: 🔧 Appeared again after the rebase.

```suggestion
- Add dynamic liquidation to the DLC, enabling the CFD to be unilaterally closed every hour by either party if the oracle attests to a price close to the ends of the payout curve's domain.
```

daemon/src/projection.rs
```diff
@@ -1165,8 +1181,8 @@ impl Actor {
         );
     }
 
-    fn handle(&mut self, msg: Update<Option<model::MakerOffers>>) {
-        self.tx.send_order_update(msg.0);
+    fn handle(&mut self, msg: Update<(ContractSymbol, Option<model::MakerOffers>)>) {
+        self.tx.send_order_update(msg.0 .0, msg.0 .1.into());
```
luckysori, line 1185: 🔧

```suggestion
        self.tx.send_order_update(msg.0.0, msg.0.1.into());
```

maker/src/routes.rs
```diff
@@ -42,20 +42,24 @@ use uuid::Uuid;
...
-    let mut rx_offers = rx.offers.clone();
     let mut rx_wallet = rx_wallet.inner().clone();
-    let mut rx_quote = rx.quote.clone();
+    let (mut rx_offers, mut rx_quote) = match symbol {
+        ContractSymbol::BtcUsd => (rx.offers.btc_usd.clone(), rx.quote.btc_usd.clone()),
+        ContractSymbol::EthUsd => (rx.offers.eth_usd.clone(), rx.quote.eth_usd.clone()),
+    };
 
     EventStream! {
+
```
luckysori: 🙃 Rogue new line.

xtra-libp2p-offer/src/lib.rs
```diff
@@ -1,9 +1,158 @@
...
+            long: maker_offers.long.map(|offer| offer.into()),
+            short: maker_offers.short.map(|offer| offer.into()),
+            tx_fee_rate: maker_offers.tx_fee_rate,
+            funding_rate_long: maker_offers.funding_rate_long,
+            funding_rate_short: maker_offers.funding_rate_short,
+        }
+    }
+}
+
+#[allow(deprecated)] // allowing for backwards compatible conversion
```
luckysori, line 100: ❓ Why do we need this statement? What's deprecated here?

xtra-libp2p-offer/src/lib.rs
```diff
@@ -1,9 +1,158 @@
...
 pub const PROTOCOL: &str = "/itchysats/offer/1.0.0";
 
+#[derive(Clone, Serialize, Deserialize, PartialEq)]
+pub struct MakerOffers {
+    pub long: Option<Offer>,
+    pub short: Option<Offer>,
+    pub tx_fee_rate: TxFeeRate,
+    pub funding_rate_long: FundingRate,
+    pub funding_rate_short: FundingRate,
+}
```
luckysori, line 31: 👍 I like the introduction of the new type, but I would change the shape as mentioned in a recent comment.

luckysori: I think we should also change the shape of `model::MakerOffers` though. The original design has some problems as discussed.

xtra-libp2p-offer/src/lib.rs
```diff
@@ -1,9 +1,158 @@
...
+            .field("tx_fee_rate", &self.tx_fee_rate)
+            .field("funding_rate_long", &self.funding_rate_long)
+            .field("funding_rate_short", &self.funding_rate_short)
+            .finish()
+    }
+}
+
+/// A concrete order created by a maker for a taker
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct Offer {
```
luckysori, line 47: 🔧 I would opt to introduce a trimmed down version of `Offer`, like the `OfferParams` I described elsewhere.

xtra-libp2p-offer/src/taker.rs
```diff
@@ -28,7 +27,7 @@ impl Actor {
             let offers = protocol::recv(stream).await?;
             let span = tracing::debug_span!("Received new offers from maker", %peer, ?offers);
             maker_offers
-                .send(LatestMakerOffers(offers))
+                .send(LatestMakerOffers(offers.map(|offer| offer.into())))
```
luckysori: 🙃 I prefer this if it works.

```suggestion
                .send(LatestMakerOffers(offers.map(model::MakerOffers::from)))
```

luckysori (approved):
I'm happy to go ahead with this and I can tackle my suggestions separately to showcase what I mean.

daemon/src/taker_cfd.rs
bonomat: > The problem that I see is that the thing that we send around for the /itchysats/offer protocol, Option<MakerOffers> doesn't work well with multiple contract symbols:

Are you _only_ talking about the types send over the network? 
I was talking about the _internal_ structure we are using from the http api of the maker throughout `projection` and other actors. 

I think the network layer can easily be cleaned up as you propose. That's what I prepared with [`a7d2335` (#2615)](https://github.com/itchysats/itchysats/pull/2615/commits/a7d2335e9c7de27151901c125d7e33b9b11114af)  and would tackle in #2459 . The internal part is a bit more work though but once the network layer is set in stone we can shuffle things around internally easily.

daemon/src/projection.rs
bonomat: dprint is not complaining, so maybe it's expected?

maker/src/routes.rs
bonomat:
```suggestion
```

bonomat: 😈

daemon/src/projection.rs
bonomat: jap, it's dprint.

daemon/src/taker_cfd.rs
luckysori: Yeah, I was certainly focusing on these types in the context of the network layer.

luckysori:
`dprint` is having a laugh: https://github.com/itchysats/itchysats/runs/7684277964?check_suite_focus=true#step:3:27.

xtra-libp2p-offer/src/lib.rs
bonomat: For that we would need to bump the network protocol version right?

xtra-libp2p-offer/src/lib.rs
bonomat: Because of `leverage_taker` which is deprecated.

xtra-libp2p-offer/src/lib.rs
luckysori: Yep. This is part of the breaking change to the offer protocol I mentioned before.

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
da-kami: > Wrapping it in BaseDlcParams sounds like a different hack to me. This struct is concerned about Parameters associated with the base DLC involved in a rollover. I can do that but I think this goes into the wrong direction and puts things together which do not really belong together.

you can also let `model::Cfd::start_rollover_maker` return `(BaseDlcParams, ContractSymbol)` and then I don't think it would be a hack.

da-kami (approved):
I feel there is some code in this PR that will require cleanup and it's a bit hard to track what is being cleaned up in other tickets mentioned here and what is to be done in follow ups. 
Are you tracking the results of discussions somewhere @bonomat so we actually tackle the cleanups?

xtra-libp2p-rollover/src/v_2_0_0/maker.rs
luckysori: Great point! I completely missed that.

In general, I think we shouldn't access from the CFD like this, but this is way less annoying. Thank you for taking a look.

PR #2616, opened 2022-08-03
Make bitmex stream symbol aware

bonomat:
Once we have #2615 we can plug it all together.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -417,9 +417,12 @@ where
 
     #[instrument(skip(self), err)]
     pub async fn propose_settlement(&self, order_id: OrderId) -> Result<()> {
+        // TODO: we need to know which contract symbol the order is for
         let latest_quote = self
             .price_feed_actor
-            .send(xtra_bitmex_price_feed::LatestQuote)
+            .send(xtra_bitmex_price_feed::LatestQuote(
+                xtra_bitmex_price_feed::ContractSymbol::BtcUsd,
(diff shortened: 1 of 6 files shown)


daemon/src/lib.rs
```diff
@@ -417,9 +417,12 @@ where
 
     #[instrument(skip(self), err)]
     pub async fn propose_settlement(&self, order_id: OrderId) -> Result<()> {
+        // TODO: we need to know which contract symbol the order is for
```
luckysori, line 420: Ahh, this is pretty annoying!

I don't want us to have to pass it in via the HTTP API or load it from the database. Shall we move the quote-fetching to the `taker_cfd::Actor` already?

daemon/src/projection.rs
```diff
@@ -1199,8 +1199,11 @@ impl xtra::Actor for Actor {
                 loop {
                     {
                         let span = info_span!("Update projection with latest quote");
+                        // TODO: subscribe to all the quotes
```
luckysori, line 1202: 🙃 Shall we already adapt the `State` to handle more than one `Quote`?

luckysori (approved):
LGTM. I left a couple of suggestions/questions regarding things that maybe could be done already.

daemon/src/projection.rs
bonomat: That's done in #2615

bonomat:
bors retry

PR #2617, opened 2022-08-03
Use structured field in fetching attestation log

luckysori:
It makes it easier to filter by the `event_id` when consuming these logs.


bonomat (approved):
very good idea :)

PR #2618, opened 2022-08-03
Enable discrete payout generation based on quanto curve

### LINKED ISSUE #2457: Payout curve for quanto contracts
Create a new payout_curve to be able to deal with ETH/USD: ETH/USD is linear in nature, i.e. each $1 move in the price of a perpetual quanto swap is worth 0.000001 BTC for one contract (or 100 satoshis).

luckysori:
Fixes #2457.

In order to generate a `quanto::Curve` model you have to provide the same arguments needed for the original inverse payout curve, plus a `multiplier`.

The API `model::Payouts::new_quanto` could eventually be combined with `model::Payouts::new_inverse`. Passing in the `ContractSymbol` would then allow branching between inverse and quanto. The `ContractSymbol` would also imply the `multiplier` to be used in the case of quanto curves.


model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    /// The profit and loss (PNL) from the perspective of the party going short.
+    fn short(&self) -> SignedAmount {
+        self.0 * -1
+    }
+}
+
+/// Closing prices at which parties get liquidated i.e. when they get a 0 payout.
+struct LiquidationThresholds {
+    inner: u64, // TODO: Find a better name!
```
luckysori: 🤔 I'm not sure if this in-between value has a name.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+
+        let thresholds = LiquidationThresholds::new(
+            effective_initial_margin,
+            self.n_contracts,
+            self.initial_price,
+            self.multiplier,
+        )
+        .context("Could not calculate liquidation thresholds")?;
+
+        Ok(thresholds.short()..=maia_core::interval::MAX_PRICE_DEC)
```
luckysori, line 280: 👀 In `maia_core`'s documentation we talk about `MAX_PRICE_DEC` being tied to `BTCUSD`. But in reality it all depends on the maximum value that the oracle attests to.

I think we can safely assume that it will still be using 20 digits for ETHUSD. But eventually this should depend on the contract symbol.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    /// Compute the closing price under which the party going long gets liquidated.
+    fn long(&self) -> u64 {
+        self.initial_price.saturating_sub(self.inner)
+    }
+
+    /// Compute the closing price over which the party going short gets liquidated.
+    fn short(&self) -> u64 {
+        self.initial_price + self.inner
+    }
```
bonomat: Given `short` and `long` are based on constants within `LiquidationThresholds` , why don't you precompute them?

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+        // We start building the settlement intervals right after the end of the long liquidation
+        // interval
+        let mut prev_upper_bound = *long_liquidation_threshold;
+        loop {
+            let lower_bound = prev_upper_bound + 1;
+            let upper_bound = lower_bound + step;
+
+            // The settlement region of the curve ends as soon as the sub-interval we are building
+            // either touches or overlaps with the start of the short liquidation interval
+            if upper_bound >= short_liquidation_threshold - 1 {
```
bonomat, line 162: Can there be a gap between `upper_bound` and `short_liquidation_threshold`? 

Thinking out loud here: 

If `lower_bound=0`, `step=10`, then ` upper_bound=10` and `short_liquidation_threshold=15`, then we only generate an interval from `0..10`.

bonomat: Ah, got it, you are treating it as a special case below :)

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+            prev_upper_bound = upper_bound;
+        }
+
+        // We have to consider a special case if the last settlement interval is smaller than every
+        // other settlement interval
+        if prev_upper_bound + 1 < *short_liquidation_threshold {
+            let lower_bound = prev_upper_bound + 1;
+            let upper_bound = short_liquidation_threshold - 1;
+
+            let interval = lower_bound..=upper_bound;
```
bonomat, line 181: is `10..10` a valid interval? 

You could get this if: 

`prev_upper_bound=9`, `short_liquidation_threshold=11`.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+    /// Compute the initial BTC margin that the party going long has to put up.
+    fn initial_margin_long(&self) -> Result<Amount> {
+        calculate_initial_margin(
+            self.initial_price,
+            self.n_contracts,
+            self.leverage_long,
+            self.multiplier,
+        )
+        .context("Could not calculate long's initial margin")
+    }
```
bonomat, line 292: Can we have some super simple unit tests for these functions please?

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+    n_contracts: u64,
+    leverage: Leverage,
+    multiplier: Decimal,
+) -> Result<Amount> {
+    let n_contracts = Decimal::from(n_contracts);
+    let leverage = Decimal::from(leverage.get());
+    let initial_price = Decimal::from(initial_price);
+
+    let margin = (n_contracts * initial_price * multiplier) / leverage;
+    let margin = margin.round_dp_with_strategy(8, RoundingStrategy::AwayFromZero);
```
bonomat: Are you sure we want `AwayFromZero`? Would `MidpointAwayFromZero` not be more _natural_?

model/src/payouts/quanto.rs
bonomat: From the formula you used this is the `bankruptcy_price`, i.e. that's where you lose everything.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+    use super::*;
+    use crate::payouts::prop_compose::arb_fee_flow;
+    use crate::payouts::prop_compose::arb_leverage;
+    use crate::payouts::quanto;
+    use itertools::Itertools;
+    use proptest::prelude::*;
+    use rust_decimal_macros::dec;
+
+    #[test]
+    fn quanto_curve_snapshot() {
```
bonomat, line 431: Can we use our `backtesting` to verify this payout curve?

model/src/payouts.rs
```diff
@@ -106,6 +108,50 @@ impl Payouts {
         })
     }
 
+    pub fn new_quanto(
+        (position, role): (Position, Role),
+        initial_price: u64,
+        n_contracts: u64,
+        (leverage_long, leverage_short): (Leverage, Leverage),
+        n_payouts: usize,
```
bonomat, line 117: Do you think we can make `n_payouts` dynamic depending on a `step` value? I mean: What I'd like to see is payouts in e.g. `$10` increments (or ideally in `10` sats increments).

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+
+    /// Discretize the payout curve into distinct payouts.
+    pub fn discretized_payouts(&self) -> Result<Vec<Payout>, Error> {
+        let initial_margin_long = self.initial_margin_long()?;
+        let initial_margin_short = self.initial_margin_short()?;
+        let initial_margin_total = initial_margin_long + initial_margin_short;
+
+        match self.fee_offset {
+            CompleteFee::LongPaysShort(long_owes) if long_owes > initial_margin_long => {
+                return Err(Error::LongOwesTooMuch {
```
da-kami, line 116: ❓ Isn't this a liquidation scenario? Why is it treated as `Err` here? On the outside we are not treating this error specifically but just bubble it up, is that expected?

bonomat (commented):
I stopped half way but from what I can tell: excellent work 👍

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+    /// The profit and loss (PNL) from the perspective of the party going short.
+    fn short(&self) -> SignedAmount {
+        self.0 * -1
+    }
+}
+
+/// Closing prices at which parties get liquidated i.e. when they get a 0 payout.
+struct LiquidationThresholds {
+    inner: u64, // TODO: Find a better name!
+    initial_price: u64,
```
Restioson, line 379: What quantity does the u64 represent here? Which payout bucket the initial price corresponds to?

luckysori: Are you asking about `initial_price`? If so, the units are whole USD, which is why we use a `u64`. 

The reason why I think using whole dollars makes sense is because the oracle only attests to the nearest dollar! So the extra precision seems unnecessary to me.

Restioson: That makes sense - I just wasn't sure what unit of measurement the u64 referred to

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+        let step = Decimal::from(short_liquidation_threshold - long_liquidation_threshold)
+            / Decimal::from(self.n_payouts - 2);
+        let step = step.to_u64().context("Could not convert step to u64")?;
+
+        // We start building the settlement intervals right after the end of the long liquidation
+        // interval
+        let mut prev_upper_bound = *long_liquidation_threshold;
+        loop {
+            let lower_bound = prev_upper_bound + 1;
+            let upper_bound = lower_bound + step;
```
da-kami, line 158: ❓ Does it make sense to first increment the `lower_bound` and then create the `upper_bound` based on the already incremented `lower_bound`?

This would mean the interval will almost always result in the "special case" that the last interval is smaller!

I created [gsheet](https://docs.google.com/spreadsheets/d/1A59TL0rcFmBet_y-2N5s9tXngf_Ifrx652pcgCy09Rk/edit?usp=sharing) to simulate the behavior.

I would have expected that we say

`let upper_bound = lower_bound + step - 1;` in the loop, see the gsheet for details. Happy to discuss it if something is unclear.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,609 @@
...
+                .payout_at_interval(interval.clone())
+                .with_context(|| format!("Could not calculate payout at interval {interval:?}"))?;
+            payouts.push(payout);
+
+            prev_upper_bound = upper_bound;
+        }
+
+        // We have to consider a special case if the last settlement interval is smaller than every
+        // other settlement interval
+        if prev_upper_bound + 1 < *short_liquidation_threshold {
```
da-kami, line 177: Given my comment above I am not 100% sure this condition holds.It would be great to have tests that prove sanity.

I propose to have a few tests that use representative values to showcase the behaviour and on top of that maybe a proptest.

da-kami (dismissed):
I tried to independently verify that the behavior is sane and am not sure. 
I think the way the intervals are created in `discretized_payouts` might not have the expected behavior. Could you add sanity tests for `discretized_payouts` to prove sanity?

model/src/payouts/quanto.rs
luckysori: These errors guard against trying to construct a payout curve with parameters that make no sense: why are you trying to create a payout curve if the effective margin (initial margin minus fees) of one of the parties is less than or equal to zero?

> On the outside we are not treating this error specifically but just bubble it up, is that expected?

There's nothing to distinguish for the caller. Ultimately this prevents someone from opening or rolling over a CFD which should not exist or should be closed instead.

da-kami: Ah, because this *only* concerns the initial margin! Then it makes sense! 
The initial margin certainly has to be more than we owe.

model/src/payouts/quanto.rs
luckysori: The base point to start computing what I've called the "settlement payouts" is the end of the long liquidation interval. Because our payout price intervals are _inclusive_, I have to set the first lower bound of these to `long_liquidation_threshold + 1` to avoid an overlap.

> This would mean the interval will almost always result in the "special case" that the last interval is smaller!

I don't see how you got to this conclusion. Can you explain your reasoning?

> I would have expected that we say
>
> let upper_bound = lower_bound + step - 1; in the loop

What I see here is that you propose to make the `step` smaller by 1, but I don't understand why.

Keep in mind that when we calculate the `step` we already take into account the fact that we are starting from `long_liquidation_threshold + 1`:

```
step = (short_liquidation_threshold - long_liquidation_threshold) / (n_payouts - 2)
     = ((short_liquidation_threshold + 1) - (long_liquidation_threshold - 1 ) / (n_payouts - 2) // accounting for inclusive intervals
```

> I created [gsheet](https://docs.google.com/spreadsheets/d/1A59TL0rcFmBet_y-2N5s9tXngf_Ifrx652pcgCy09Rk/edit?usp=sharing) to simulate the behavior.

I can't quite parse what's happening in that one :(

Btw, with your suggestion the last settlement interval is still smaller than the others _and_ we produce an extra CET (I tested it using the snapshot test). It is inevitable that we will more often than not have a smaller settlement interval, because the overall settlement interval will usually not be divisible by the step.

model/src/payouts/quanto.rs
luckysori: > Given my comment above I am not 100% sure this condition holds.

But this is not an invariant. It just handles the special case of the last interval being smaller than the rest.

That is, if the overall settlement interval is not divisible by the step, there will be some leftover (the remainder of the integer division) which will be the last interval.

> I propose to have a few tests that use representative values to showcase the behaviour and on top of that maybe a proptest.

Can you describe the tests that you are looking for that are not already part of the PR? I am sure we can test other properties and perhaps add some unit tests as @bonomat suggested elsewhere, but I'm not sure what you are asking for yet.

model/src/payouts/quanto.rs
da-kami: > The base point to start computing what I've called the "settlement payouts" is the end of the long liquidation interval. Because our payout price intervals are _inclusive_, I have to set the first lower bound of these to `long_liquidation_threshold + 1` to avoid an overlap.
> 
> > This would mean the interval will almost always result in the "special case" that the last interval is smaller!
> 
> I don't see how you got to this conclusion. Can you explain your reasoning?

I think the comment stating that it is a special case 

```
        // We have to consider a special case if the last settlement interval is smaller than every
        // other settlement interval
```

lead me to believe that this would not be something that can happen.

[snip]
> Btw, with your suggestion the last settlement interval is still smaller than the others _and_ we produce an extra CET (I tested it using the snapshot test). It is inevitable that we will more often than not have a smaller settlement interval, because the overall settlement interval will usually not be divisible by the step.

True, my suggestion is incorrect, because it is based on the assumption of `n_payouts + 1` then 🙈
Yes, you are right, there is likely always some remainder, and that is just one of the intervals.
Sorry, I hope what I said was not too confusing.

Question on top, is it "fair" that the shorter interval is the last interval before the `short` liquidation? Should that interval not be the one closest to the original price (i.e. in the middle of the distribution) to be fair? (not super important)

model/src/payouts/quanto.rs
da-kami: I think what @bonomat suggests is in line with what I had in mind, smaller unit tests for e.g. `discretized_payouts`. 

Sorry, I only saw his review after submitting mine (we reviewed in parallel).

model/src/payouts/quanto.rs
luckysori: Yep, single-valued intervals are valid and we have to support them.

Normally they will only occur for the special case in the settlement interval. But it could be that we model every price value as a separate CET with sufficient resolution!

model/src/payouts/quanto.rs
luckysori: I definitely wasn't sure! I'm not too fussed, so `MidpointAwayFromZero` it is :D

model/src/payouts/quanto.rs
luckysori: Thank you! I knew there would be a name.

model/src/payouts/quanto.rs
luckysori: I see your point. But each of them is only called once, separately. So it's a minuscule bit more efficient to compute them on demand.

The only reason why I introduced the type was so that they could share the code between them. Two free functions calling to a common free function would be valid, but then I would have to return the intermediate values in a tuple, which is kind of ugly.

model/src/payouts/quanto.rs
luckysori: Nothing stops us from doing so, the return type is the same if you call `model::Payouts::new_quanto`.

Do you want me to create a ticket?

model/src/payouts.rs
luckysori: We absolutely can replace `n_payouts` with a `max_interval_size` argument. I just wanted to keep it in line with the other API.

Mind you that I think the reason why we use `n_payouts` is so that we can more or less consistently control the number of CETs per CFD.

model/src/payouts/quanto.rs
luckysori: > Question on top, is it "fair" that the shorter interval is the last interval before the short liquidation? Should that interval not be the one closest to the original price (i.e. in the middle of the distribution) to be fair? (not super important)

It's important to highlight that no matter where the short settlement interval lies, the liquidation intervals remain intact. So in that sense it makes no difference.

Any slightly shorter settlement interval just means that a particular region has a bit more resolution in terms of deciding the payouts. The fairest approach would be to randomise the location of the shorter settlement interval, but the complexity cost far outweighs the benefits in my opinion.

da-kami (approved):
Went over it again and don't have anything to add :)
Great work

model/src/payouts.rs
bonomat: > Mind you that I think the reason why we use n_payouts is so that we can more or less consistently control the number of CETs per CFD.

Right. I think the number of CETs are not that important as long as they don't explode. Meaning, we should figure out what good step sizes are. 

The advantage I see is that we have always the same payout interval no matter what amounts we trade.

model/src/payouts/quanto.rs
bonomat: Yes please. I think that would be good.

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,691 @@
...
+
+        Ok(Self {
+            bankruptcy_price: inner,
+            initial_price,
+        })
+    }
+
+    /// Compute the closing price under which the party going long gets liquidated.
+    fn long(&self) -> u64 {
+        self.initial_price.saturating_sub(self.bankruptcy_price) // TODO: Not sure if this is right
```
bonomat: What's your concern here?

model/src/payouts/quanto.rs
```diff
@@ -0,0 +1,691 @@
...
+}
+
+#[cfg(test)]
+mod unit_tests {
+    use rust_decimal_macros::dec;
+
+    use super::*;
+
+    #[test]
+    fn initial_margin_snapshot() {
```
bonomat, line 622: Thanks 👍

bonomat (approved):
The only concern I have is that BitMEX calculator seems to come up with different values for long and short while we come up with the same. 


Margin if you are short is `0.0726`
image

Margin if you are long is `0.0101`
image


I don't get it why :/ 

I would say we move forward with PR and see to get answers on this afterwards.

bonomat:
> The only concern I have is that BitMEX calculator seems to come up with different values for long and short while we come up with the same.
> 
> Margin if you are short is `0.0726` image
> 
> Margin if you are long is `0.0101` image
> 
> I don't get it why :/
> 
> I would say we move forward with PR and see to get answers on this afterwards.

You know what, I think the calculator is bollocks 😅... When I open a position the margin requriements are the same for long and short

image

image

model/src/payouts/quanto.rs
luckysori: Ooop, should have removed that TODO.

I just noticed that the long liquidation threshold is usually at a 0 price, which seemed odd. But I checked with the BitMEX calculator and the behaviour was consistent.

model/src/payouts.rs
luckysori: I've created https://github.com/itchysats/itchysats/issues/2632.

model/src/payouts/quanto.rs
luckysori: See https://github.com/itchysats/backtesting/issues/70.

Issue #2631, opened 2022-08-04
Taker cannot rollover one position but can rollover another one


bonomat:
taker version `0.5.0`
network: `mainnet`
peerid: `12D3KooWQKVeZWeD5WBKEjN4pDVjmd4VwyNrdm2Ey6bJWGW7FEw7`
takerid: `921d68aa0b584901d038b6460d3155464febe2dbad74c77a9eae91be60a4c771`
orderid: `2c6b3182-f648-4f96-87c2-60a698501e33`

I noticed that I cannot collaborative settle a position. In the UI I only get 

<img width="353" alt="image" src="https://user-images.githubusercontent.com/224613/182962930-6349db11-5111-47b4-99a1-42414356a20b.png">

In the console I see: 
```
2022-08-04 22:20:58 DEBUG http: Handled request method=POST path=/api/cfd/2c6b3182-f648-4f96-87c2-60a698501e33/settle status=500 Internal Server Error
2022-08-04 22:23:52 DEBUG POST /cfd/<id>/<action>{id=2c6b3182-f648-4f96-87c2-60a698501e33 action="settle"}:propose_settlement{order_id=OrderId(2c6b3182-f648-4f96-87c2-60a698501e33)}:xtra_actor_request{actor_type=daemon::taker_cfd::Actor<daemon::oracle::Actor, daemon::wallet::Actor<bdk::blockchain::electrum::ElectrumBlockchain, sled::tree::Tree>> message_type=daemon::taker_cfd::ProposeSettlement}:xtra_message_handler:Handle ProposeSettlement: daemon::taker_cfd: Proposing settlement of contract order_id=2c6b3182-f648-4f96-87c2-60a698501e33 proposal_closing_price=22560.5 bid=22560 ask=22560.5 quote_timestamp=2022-08-04T22:23:00Z
2022-08-04 22:23:52 ERROR POST /cfd/<id>/<action>{id=2c6b3182-f648-4f96-87c2-60a698501e33 action="settle"}:propose_settlement{order_id=OrderId(2c6b3182-f648-4f96-87c2-60a698501e33)}: daemon: error=could not start closing position
2022-08-04 22:23:52 ERROR POST /cfd/<id>/<action>{id=2c6b3182-f648-4f96-87c2-60a698501e33 action="settle"}: taker::routes: error=500 Internal Server Error - settle failed - could not start closing position: Failed to execute command on CFD: Collaborative close without DLC
[...]

PR #2636, opened 2022-08-05
Remove database cache

bonomat:
We initially had the cache because we had such a big db. Ever since we are pruning old event data this is not the case anymore. Hence, the cache should not be needed anymore.

This is an attempt at fixing #2631. But we have no way of verifying that it does unless the error never happens again.

### DIFF
--- a/sqlite-db/Cargo.toml
+++ b/sqlite-db/Cargo.toml
@@ -9,7 +9,6 @@ edition = "2021"
 anyhow = "1"
 async-stream = "0.3"
 bdk = "0.19.0"
-chashmap-async = "0.1"
 futures = { version = "0.3", default-features = false }
 hex = "0.4"
 libp2p-core = { version = "0.33", default-features = false }
--- a/sqlite-db/src/lib.rs
+++ b/sqlite-db/src/lib.rs
@@ -3,7 +3,6 @@ mod sqlx_ext; // Must come first because it is a macro.
 use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
-use chashmap_async::CHashMap;
 use futures::future::BoxFuture;
 use futures::FutureExt;
 use futures::Stream;
@@ -29,11 +28,8 @@ use sqlx::Connection as _;
 use sqlx::Sqlite;
 use sqlx::SqlitePool;
 use sqlx::Transaction;
-use std::any::Any;
-use std::any::TypeId;
 use std::path::PathBuf;
 use std::str::FromStr;
-use std::sync::Arc;
 use time::Duration;
 
 pub use closed::*;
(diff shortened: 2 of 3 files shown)


sqlite-db/Cargo.toml
```diff
@@ -9,7 +9,6 @@ edition = "2021"
 anyhow = "1"
 async-stream = "0.3"
 bdk = "0.19.0"
-chashmap-async = "0.1"
```
luckysori, line 12: 👍 Thanks!

luckysori (approved):

da-kami:
After our discussion on #2631 I had some more thoughts, and I think before we merge this we should verify the assumption that loading the events on every load is not *significantly* longer than using the cache. 

Even though it is true that we only have one `Dlc` at all times now, loading that one `Dlc` is not unexpensive neither, and it might be worth caching this data. By removing the cache we *always* load the complete `Cfd`, including the `Dlc`. Given that we load all `Cfd`s quite often this might still be quite expensive.

I'd love to see the cache removed, but I think it'd be worth to get some benchmarks on this. Maybe @Restioson can help with that :)

Restioson:
What could be interesting is to run a little A/B test :) Half the time, randomly, send the request to the cache. The other half of the time, just get it from the directory. Then, we can try and compare the span metrics for the different spans and see how they differ. This would best reflect real world usage. Otherwise, we could write some small test cases and run them on the testnet database and try to get a cached CFD and get one from the DB directly, and measure the difference that way.

sqlite-db/src/lib.rs
```diff
@@ -277,39 +269,20 @@ impl Connection {
         let mut conn = self.inner.acquire().await?;
         let mut db_tx = conn.begin().await?;
 
-        let cache_key = (TypeId::of::<C>(), id);
-        let aggregate = std::any::type_name::<C>();
-
-        let cfd = match self.aggregate_cache.remove(&cache_key).await {
```
Restioson, line 283: Even if the cache remains, I think that using a concurrent hashmap should be unnecessary. Something DashMap should probably suffice, as the CFD guard is never very long, so the .awaits probably add unnecessary overhead.

This might sound surprising, given that acquiring a lock is generally considered blocking. However, when the critical section is so short as it is here, an async lock does more harm than good. See the docs of [`tokio::sync::Mutex`](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html) for more on this concept.

Restioson:
I ended up running a simple test of loading all the open CFDs from the testnet maker's sqlite database to figure out what the performance difference was like. Then, once I ascertained that there was a large discrepancy, I instrumented various parts of the process (extracted into #2684) and ran it under Jaeger locally to find the bottlenecks. I have pushed a branch based on #2684 containing the experiment to [Restioson/bench-cache-vs-from-db](https://github.com/Restioson/itchysats/tree/bench-cache-vs-from-db).

From database:
image

From cache:
image

So, cache is significantly faster. 

Jumping into one of the loaded-from-DB spans, we see that loading the rollover data is the bottleneck here
image

Indeed, this is what makes one of the loads take 300ms

image

I am not sure if this is possible to "fix" (maybe someone else can explain what would be required?), so maybe it is best to retain some kind of caching strategy for now.

luckysori:
Thanks for testing this, @Restioson!

I agree that it would be too much of a performance hit to remove the cache.

I'm gonna close this PR for the time being.

PR #2637, opened 2022-08-05
Overwrite short liquidation upper bound with max attestation price

### LINKED ISSUE #2571: Short liquidation events when maker goes short are unusable
### Context

### Problem

### What to do?

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2571.

### DIFF
--- a/model/src/payouts.rs
+++ b/model/src/payouts.rs
@@ -77,7 +77,7 @@ impl Payouts {
         n_payouts: usize,
         fee: CompleteFee,
     ) -> Result<Self> {
-        let payouts = payouts::inverse::calculate(
+        let mut payouts = payouts::inverse::calculate(
             price,
             quantity,
             long_leverage,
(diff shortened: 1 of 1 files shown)


model/src/payouts.rs
```diff
@@ -86,6 +86,15 @@ impl Payouts {
             fee,
         )?;
 
+        // Overwrite the short liquidation upper bound with the maximum price that Olivia can attest
+        // to
+        {
+            let n_payouts = payouts.len() - 1;
```
da-kami, line 92: ❓Why not use the `n_payouts` parameter passed into the function here? I find it confusing that we re-define this.

model/src/payouts.rs
```diff
@@ -86,6 +86,15 @@ impl Payouts {
...
+            let short_liquidation = payouts.get_mut(n_payouts).expect("several payouts");
+            short_liquidation.range =
+                *short_liquidation.range.start()..=maia_core::interval::MAX_PRICE_DEC;
```
da-kami, line 95: Given that the maximum price is `2 ^ 20 - 1 = 1,048,575` this should be fine for a while, but for the next bull market we might want to support a higher maximum in `maia` 🤪

da-kami (approved):

holzeis (approved):

Restioson (approved):

model/src/payouts.rs
luckysori: I didn't even think about it haha.

I prefer this because otherwise we rely on knowing what the implementation of `payouts::inverse::calculate` does to ensure that we choose the very last payout (which is crucial!).

PR #2638, opened 2022-08-05
Ignore failures to accept/reject rollovers on maker HTTP API

luckysori:
These are only relevant for the legacy rollover protocol, which is no longer being developed.

But we still emit them for the new rollover protocol, because the `projection::Actor` creates `RolloverAccept` and `RolloverReject`
actions for CFDs that are _not_ legacy too.

Therefore downgrading them to trace and not returning a HTTP API Problem should do the job until we actually remove the legacy rollover protocol in a future release.


da-kami (approved):
After removing the old connection there will be a loooot to clean up 😅

luckysori:
> After removing the old connection there will be a loooot to clean up 😅

And it will feel goooood.

PR #2639, opened 2022-08-05
Drop support for all legacy protocols

luckysori:
I'm doing this now because changing the new offer protocol was incredibly annoying without tackling this first.

It deals with most of the first section of https://github.com/itchysats/itchysats/issues/2664.


Restioson (commented):
I'm not 100% sure of what the implications are for breakage for existing users (e.g how long has this been deprecated, anyway?) but the simplification is definitely nice!

luckysori:
> I'm not 100% sure of what the implications are for breakage for existing users (e.g how long has this been deprecated, anyway?) but the simplification is definitely nice!

Fair question! We've been allowing protocols over the old network stack for a very long time. We released 0.5.4 partly to [inform takers](https://github.com/getumbrel/umbrel-apps/pull/104/files#diff-1c58e2de022efd21a21ddedaf772bd5136920ab09c98b5a452f1fdbfa4c279e0R61-R63) that our maker will no longer support these protocols in the next release.

Thinking about it, that's pretty silly because any taker on 0.5.4 by definition doesn't use any legacy protocols haha. In any case, our intention is to drop support for the next release, so this in line with that goal and the https://github.com/itchysats/itchysats/issues/2664.

maker/src/actor_system.rs
```diff
@@ -77,6 +75,7 @@ pub struct ActorSystem<O: 'static, W: 'static> {
             cfd::RatesChannel,
         >,
     >,
+    _oracle_actor: Address<O>,
```
da-kami, line 74: ❓Is this now necessary because we don't keep a reference to this actor in `cfd_actor` anymore?

da-kami (approved):
This is such a feels good!
The tests are passing without much changes to the tests, that's nice.

Let's get this on `testnet`!

da-kami:
sorry for the confusion, I had to re-check that this is needed for getting the API change PR in. It is, so I am merging this and then the API change. Will see to re-review the protocol change to `offer/2.0.0` later and potentially merge that one as well.

maker/src/actor_system.rs
Restioson: I never fully understood the underlying purpose of these references. I get that they are to keep the actor alive, but why must it be done in this way? Couldn't it be done "for free" by keeping strong references in whatever part of the application provides the input messages to these actors - would someone mind explaining this?

PR #2649, opened 2022-08-05
Harmonize fields for better logging

da-kami:
This helps when debugging issues and is good for a cleaner code base in general

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1325,7 +1325,9 @@ impl Cfd {
                 price: settlement.price,
             })
         } else {
-            self.fail_collaborative_settlement(anyhow!("Cannot complete collaborative settlement"))
+            self.fail_collaborative_settlement(anyhow!(
+                "CFD not eligible for collaborative settlement anymore"
+            ))
         }
     }
 
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1257,7 +1257,7 @@ impl Cfd {
             )
         }
 
-        tracing::info!(order_id = %self.id, "Contract setup was completed");
+        tracing::info!(order_id = %self.id, peer_id=?self.counterparty_peer_id, "Contract setup was completed");
 
         Ok(self.event(EventKind::ContractSetupCompleted { dlc: Some(dlc) }))
     }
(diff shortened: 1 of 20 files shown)


xtra-libp2p/tests/basic.rs
```diff
@@ -360,11 +360,11 @@ impl Actor for EndpointSubscriberStats {
 #[xtra_productivity]
 impl EndpointSubscriberStats {
     async fn handle(&mut self, msg: endpoint::ConnectionEstablished) {
-        self.connected_peers.insert(msg.peer);
+        self.connected_peers.insert(msg.peer_id);
```
luckysori, line 363: 😆 I laugh that you aligned the tests as well.

model/src/cfd.rs
```diff
@@ -1325,7 +1325,9 @@ impl Cfd {
                 price: settlement.price,
             })
         } else {
-            self.fail_collaborative_settlement(anyhow!("Cannot complete collaborative settlement"))
+            self.fail_collaborative_settlement(anyhow!(
+                "CFD not eligible for collaborative settlement anymore"
+            ))
```
luckysori, line 1330: 🙃 It would be better if we returned an error from `can_settle_collaboratively` and used it here. 

Otherwise what you have to do is know what `can_settle_collaboratively` does in order to write this error message, which is harder and can get out of sync.

luckysori (approved):
Amazing stuff. I think this is actually really important and it's sort of insane that we haven't done this before. Thank you.

xtra-libp2p/tests/basic.rs
luckysori: I meant to say "love", but it's funnier this way.

model/src/cfd.rs
da-kami: I implemented this, but will open in separate PR because I should get more reviews given that I am flipping some logic :)

da-kami: https://github.com/itchysats/itchysats/pull/2652

da-kami:
bors retry

da-kami:
bors failed on `payouts::quanto::api_tests::payout_intervals_have_no_gaps` which will be solved by https://github.com/itchysats/itchysats/pull/2650

PR #2650, opened 2022-08-05
Handle overlapping liquidation intervals for quanto curve

luckysori:
This is an edge case that most likely only appears in proptests. We already saw this in CI: https://github.com/itchysats/itchysats/runs/7685221192?check_suite_focus=true. But it's hard to avoid, so we either need to handle it or explicitly
`reject` the errors like we do with `{Long,Short}OwesTooMuch`.

I chose to handle it because it's less code. I chose to increase the lower bound of the short liquidation threshold rather than to
decrease the upper bound of the long liquidation threshold to avoid possibly getting a -1 as a result, which would panic.

### DIFF
--- a/model/src/payouts/quanto.rs
+++ b/model/src/payouts/quanto.rs
@@ -130,9 +130,16 @@ impl Curve {
         let long_liquidation_interval = self
             .long_liquidation_interval()
             .context("Could not calculate long liquidation interval")?;
-        let short_liquidation_interval = self
+        let mut short_liquidation_interval = self
             .short_liquidation_interval()
             .context("Could not calculate short liquidation interval")?;
+
+        // Under very specific conditions the liquidation intervals can overlap. To avoid this
+        // situation we shift the short liquidation interval by 1
+        if long_liquidation_interval.end() == short_liquidation_interval.start() {
+            short_liquidation_interval =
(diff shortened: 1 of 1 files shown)


da-kami (approved):
I think a scenario where the liquidation intervals overlap actually does not make sense for our usecase and it would be fair to just reject it rather than doing the shift. 

If it makes it easier to roll with this solution feel free to go ahead, I just feel it might be hard to understand this complexity when looking at it again after some time.

luckysori:
> I think a scenario where the liquidation intervals overlap actually does not make sense for our usecase and it would be fair to just reject it rather than doing the shift.
> 
> If it makes it easier to roll with this solution feel free to go ahead, I just feel it might be hard to understand this complexity when looking at it again after some time.

The problem is that the overlap unfortunately can happen with the code I wrote, so I have to do something about it if we want to keep the proptests (I think we definitely do).

I am also unhappy about the complexity, but the alternative is more complicated for me:

- Checking if the overlap occurs (we already do this in this PR) and explicitly modelling a `LiquidationIntervalsOverlap` error variant to be returned instead of tweaking the intervals.
- In the proptests, handling the new error variant to emit a `proptest::TestCaseError::Reject` so that the proptests don't fail when we encounter this.

luckysori:
bors r+

Going ahead with the merge, but we can reconsider this and change it afterwards.

PR #2651, opened 2022-08-05
Insert complete `RolloverCompleted` events only

luckysori:
We don't need to deal with the entirety of `EventKind` when we insert rollover-specific information into the DB.


sqlite-db/src/rollover/insert.rs
```diff
@@ -19,61 +17,38 @@ use sqlx::Transaction;
 pub async fn insert(
```
da-kami: 🙃Since this is not just `insert`ing data but also deleting, should we rename it `replace_rollover`. Not sure it makes it better, but since I noticed I thought I mention it.

da-kami (approved):

sqlite-db/src/rollover/insert.rs
luckysori: I've pushed a few commits in this direction.

da-kami (approved):
Nice, thanks for addressing the feedback! I'm happy to merge like this.

sqlite-db/src/rollover/overwrite.rs
```diff
@@ -16,64 +15,47 @@ use sqlx::Connection as SqlxConnection;
...
-        _ => {
-            tracing::error!("Invalid event type. Use `append_event` function instead")
+    for (event_id, cets) in dlc.cets {
+        for cet in cets {
+            insert_cet(&mut inner_transaction, event_id.into(), order_id, cet).await?;
         }
     }
 
+    // Commit the transaction to either write all or rollback
+    inner_transaction.commit().await?;
```
holzeis, line 57: @luckysori you mention here that it should either commit or rollback, however I don't see an error handling here that would rollback. Is this somewhere hidden in the rust magic? furthermore, do we need a bit more error handling here - otherwise a failed transaction will remain failed. Do we store the events - so we can keep a status or something?

sqlite-db/src/rollover/overwrite.rs
```diff
@@ -16,64 +15,47 @@ use sqlx::Connection as SqlxConnection;
...
 use sqlx::Transaction;
 
-pub async fn insert(
+mod delete;
+
+/// Overwrite a CFD's latest rollover data.
+///
+/// After a successful rollover, we can forget about the previous `Dlc`, `FundingFee` and
+/// `CompleteFee`.
+pub async fn overwrite(
```
holzeis, line 24: Do we need to think about mutual exclusion here?

luckysori: Can you elaborate? Mind you, this PR is only refactoring things.

sqlite-db/src/rollover/overwrite.rs
luckysori: I didn't write this comment, I just moved it around 😛 Tbh, I think the comment is a bit confusing/unnecessary.

> you mention here that it should either commit or rollback, however I don't see an error handling here that would rollback. Is this somewhere hidden in the rust magic?

The library we are using [handles the rollback for us](https://docs.rs/sqlx/latest/sqlx/struct.Transaction.html).

> furthermore, do we need a bit more error handling here - otherwise a failed transaction will remain failed. Do we store the events - so we can keep a status or something?

If we fail, we roll back and bubble up an error. Since the `RolloverCompleted` event doesn't get inserted, the state of the CFD won't move forward for this party. Assuming that it did succeed for the other party, they would eventually need to catch up via the "rollover retry" feature that we introduced recently.

Basically, failing to overwrite this state is not handled very well. I've opened https://github.com/itchysats/itchysats/issues/2662 to talk about it.

sqlite-db/src/rollover/overwrite.rs
luckysori: This was discussed offline and I came to the conclusion that we are already implementing mutual exclusion when writing to the database because we only have one call to `append_event` in production:

https://github.com/itchysats/itchysats/blob/1305340f1fbb8567ed72f594bbf03013cbae93da/daemon/src/process_manager.rs#L74

I'm gonna take this conversation to a separate ticket because it will likely get lost here.

luckysori: Here it is: https://github.com/itchysats/itchysats/issues/2663.

PR #2652, opened 2022-08-05
Capture logic for no collab settlement as error

da-kami:


model/src/cfd.rs
```diff
@@ -796,11 +809,24 @@ impl Cfd {
         Ok(())
     }
 
-    fn can_settle_collaboratively(&self) -> bool {
-        !self.is_closed()
-            && !self.commit_finality
-            && !self.is_attested()
-            && !self.is_in_force_close()
+    fn can_settle_collaboratively(&self) -> Result<(), CannotSettleCollaborativeley> {
```
luckysori: 🙃 Since it doesn't return a boolean anymore should we call it `check_collaborative_settlement` or something?

model/src/cfd.rs
```diff
@@ -361,6 +361,19 @@ pub enum NoRolloverReason {
     Closed,
 }
 
+/// Reasons why we cannot collab close a CFD
+#[derive(thiserror::Error, Debug, PartialEq, Clone, Copy)]
+pub enum CannotSettleCollaborativeley {
+    #[error("The CFD was already force closed")]
+    OngoingForceClose,
+    #[error("The CFD is already Committed")]
```
luckysori: 🙃 🙃 Weird that "Committed" is capitalised.

model/src/cfd.rs
```diff
@@ -361,6 +361,19 @@ pub enum NoRolloverReason {
     Closed,
 }
 
+/// Reasons why we cannot collab close a CFD
+#[derive(thiserror::Error, Debug, PartialEq, Clone, Copy)]
+pub enum CannotSettleCollaborativeley {
```
luckysori: 🔧 Should be `CannotSettleCollaboratively`.

luckysori (approved):
Thanks! A couple of nitty things and a typo.

model/src/cfd.rs
da-kami: hahaha, I fixed it three times before opening the PR already 🙈

model/src/cfd.rs
da-kami: I'd keep it for now, given that we also have `can_rollover` with the same pattern, and not other `can_xxx` pattern for similar logic.

PR #2661, opened 2022-08-08
Changelog fixes

da-kami:
I noticed that the protocol name for the `identify` protocol was wrong. I fixed that manually in the release page on https://github.com/itchysats/itchysats/releases/tag/0.5.4 and then updated the changelog. While doing so I noticed that some the recent merges actually wrongly updated the changelog, so I fixed that in another commit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce the `/itchysats/order/1.0.0` protocol, running over libp2p.
   It is used to place an order and immediately set up the CFD on chain.
   It replaces the previous mechanism, which ran over a bespoke network stack.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Introduce the `/itchysats/order/1.0.0` protocol, running over libp2p.
   It is used to place an order and immediately set up the CFD on chain.
   It replaces the previous mechanism, which ran over a bespoke network stack.
-- Introduce the `/itchysats/identify/0.3.0` protocol, running over libp2p.
+- Introduce the `/itchysats/id/1.0.0` protocol, running over libp2p.
```
luckysori, line 31: 👍 Oops, my bad. I think it did have that name at some point :O

luckysori (approved):
Thanks for fixing this!

PR #2671, opened 2022-08-10
Roll back if `rollover::overwrite` fails when appending events to the DB

### LINKED ISSUE #2662: What happens if `RolloverCompleted` cannot be overwritten?
Now that we segregate the storage of the DLC data from the `RolloverCompleted` event, we can run into a situation where we succeed in appending to the `events` table, but fail to overwrite the DLC data stored in the `rollover_completed_event_data` table.

https://github.com/itchysats/itchysats/blob/1305340f1fbb8567ed72f594bbf03013cbae93da/sqlite-db/src/lib.rs#L227-L290

Is this a problem?

luckysori:
Previously we were allowing the following scenario:

- `RolloverCompleted` event is appended to the `events` table.
- `RolloverCompleted` event data fails to be appended to `rollover_completed_event_data` for whatever reason.

That scenario would leave us in a situation where the state between the two tables would be inconsistent.

To fix it, we ensure that the rollback logic doesn't just cover `rollover::overwrite::overwrite`, but rather the entirety of `append_event`.

Fixes https://github.com/itchysats/itchysats/issues/2662.

### DIFF
--- a/sqlite-db/src/failed.rs
+++ b/sqlite-db/src/failed.rs
@@ -33,10 +33,8 @@ use model::FundingFee;
 use model::OrderId;
 use model::Timestamp;
 use models::FailedKind;
-use sqlx::pool::PoolConnection;
-use sqlx::Connection as _;
-use sqlx::Sqlite;
-use sqlx::Transaction;
+use sqlx::Acquire;
+use sqlx::SqliteConnection;
 
 /// A trait for building an aggregate based on a `FailedCfd`.
 pub trait FailedCfdAggregate: CfdAggregate {
@@ -159,7 +157,7 @@ impl Connection {
 }
 
 async fn insert_failed_cfd(
-    conn: &mut Transaction<'_, Sqlite>,
+    conn: &mut SqliteConnection,
     cfd: Cfd,
     event_log: &EventLog,
 ) -> Result<()> {
@@ -257,7 +255,7 @@ async fn insert_failed_cfd(
 }
 
 async fn insert_event_log(
-    conn: &mut Transaction<'_, Sqlite>,
+    conn: &mut SqliteConnection,
     id: OrderId,
     event_log: EventLog,
 ) -> Result<()> {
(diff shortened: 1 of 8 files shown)


luckysori:
@holzeis: I'm gonna start using https://www.conventionalcommits.org/en/v1.0.0/ in this repo at least.

The only downside I see is that I get less space to write the commit title, but otherwise I really like the idea.

da-kami (approved):
Nice! really like the refactor to avoid the need of the transaction. 
Happy to see this go in once the compile issues are resolved!

sqlite-db/src/failed.rs
```diff
@@ -256,11 +255,12 @@ async fn insert_failed_cfd(
...
 }
 
-async fn insert_event_log(
-    conn: &mut Transaction<'_, Sqlite>,
-    id: OrderId,
-    event_log: EventLog,
-) -> Result<()> {
+async fn insert_event_log<'a, A>(conn: A, id: OrderId, event_log: EventLog) -> Result<()>
+where
+    A: Acquire<'a, Database = Sqlite>,
```
Restioson: This refactor is nice! I don't really know exactly how this all works internally, but one thing I would note that in future, for some functions, `Transaction` might be preferable over `impl Acquire`, if the given operation should *always* be part of a bigger transaction, i.e if it could leave the db in an inconsistent state if a subsequent operation fails, if the function is designed only to be a helper function and always a constitutent of a larger operation.

I have no idea if this is ever true in this case, though :sweat_smile:

Restioson (approved):

sqlite-db/src/failed.rs
luckysori: You raise an interesting point. On the one hand, by asking for a `Transaction` you inform the caller that they should be using one and therefore thinking about committing or rolling back. On the other hand, these functions don't need to interface with the `Transaction` API at all: we really don't want to call `commit` inside one of these functions (unless they are the very last thing that needs to be done)!

Given that we don't actually need to (or should) use the functionality provided by the trait inside these functions, I would lean towards passing in something that implements `Execute` (in this case we have to settle for `Acquire`, because of library limitation).

luckysori:
bors r+

I've fixed the build and opted for using the `SqliteConnection` type instead of `Acquire`, which leads to even less code. I'm merging to speed things up, but I'm asking for reviews to raise awareness about this change. In practice I think it's just the same but easier. 

The `Acquire` docs already kind of recommend this approach, but I missed this: https://docs.rs/sqlx/latest/sqlx/trait.Acquire.html.

PR #2679, opened 2022-08-11
Remove contract symbol from feed api

### LINKED ISSUE #2673: Remove contract symbol from feed api
Currently the feed api exposes individual endpoints per contract symbols by which offers and quotes can be put into context. However all other data events are replicated like wallet, maker_status, maker_identity and cfds.

Following the original design of the feed api the more sensible approach would have been to keep one endpoint which provides all feeds and make the offers and quote feeds contract symbol aware to allow the subscriber to deal with it accordingly.

https://github.com/itchysats/itchysats/blob/64f07eb160c428999258bdb7c66a4cade682c6b0/daemon/src/projection.rs#L90-L98

At a later stage we could even think about adding a query parameter to the feed to allow for filtering specific event types. e.g. only send quotes for ethusd, etc. I think that would give us far more flexibility as such a filter could also implement something like versioning into the query language. In my opinon this kind of design will also help us eventually If we intent toopen up the daemon api for third party applications - as integration will be simplified.

holzeis:
Proposes to remove the symbol from the feed api - allowing the consumer to filter for the events interested on the client side. This reduces a lot of complexity in the implementation as well as in the consumption of the API. 

This PR does not yet include the event filtering through a query parameter yet, but should be added in following PRs if we decide to go with that approach.

- [x] Upon connecting to the feed api the latest sent events will be resent, resulting into an issue that either btcusd or ethusd will be sent as the last sent could be either. This results into a delay for btcusd or ethusd to get the latest quote, offer.

Resolves #2673

### DIFF
--- a/daemon-tests/src/flow.rs
+++ b/daemon-tests/src/flow.rs
@@ -3,7 +3,6 @@ use anyhow::Result;
 use daemon::projection::Cfd;
 use daemon::projection::CfdState;
 use daemon::projection::MakerOffers;
-use daemon::projection::OffersFeed;
 use model::ContractSymbol;
 use model::OrderId;
 use std::time::Duration;
@@ -14,21 +13,23 @@ const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) {
 
 /// Wait and return next non-empty maker offers with `contract_symbol`
 pub async fn next_maker_offers(
-    rx_a: &mut OffersFeed,
-    rx_b: &mut OffersFeed,
+    rx_a: &mut watch::Receiver<MakerOffers>,
+    rx_b: &mut watch::Receiver<MakerOffers>,
     contract_symbol: &ContractSymbol,
 ) -> Result<(MakerOffers, MakerOffers)> {
-    let (mut rx_a, mut rx_b) = match contract_symbol {
-        ContractSymbol::BtcUsd => (rx_a.btc_usd.clone(), rx_b.btc_usd.clone()),
-        ContractSymbol::EthUsd => (rx_a.eth_usd.clone(), rx_b.eth_usd.clone()),
-    };
+    let mut rx_a = rx_a.clone();
+    let mut rx_b = rx_b.clone();
 
-    let non_empty_offer = |offer: MakerOffers| {
-        if offer.long.is_some() && offer.short.is_some() {
-            Some(offer)
-        } else {
-            None
+    let non_empty_offer = |offers: MakerOffers| match &offers {
+        MakerOffers {
+            short: Some(short),
+            long: Some(long),
+        } if &short.contract_symbol == contract_symbol
+            && &long.contract_symbol == contract_symbol =>
+        {
+            Some(offers)
         }
+        _ => None,
     };
 
     let wait_until_a = next_with(&mut rx_a, non_empty_offer);
@@ -39,15 +40,8 @@ pub async fn next_maker_offers(
     Ok((a?, b?))
 }
 
-pub async fn is_next_offers_none(
-    rx: &mut OffersFeed,
-    contract_symbol: &ContractSymbol,
-) -> Result<bool> {
-    let mut rx = match contract_symbol {
-        ContractSymbol::BtcUsd => rx.btc_usd.clone(),
-        ContractSymbol::EthUsd => rx.eth_usd.clone(),
-    };
-    let maker_offers = next(&mut rx).await?;
+pub async fn is_next_offers_none(rx: &mut watch::Receiver<MakerOffers>) -> Result<bool> {
+    let maker_offers = next(rx).await?;
     Ok(maker_offers.long.is_none() && maker_offers.short.is_none())
 }
 
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -82,19 +82,9 @@ pub struct Actor {
 }
 
 pub struct Feeds {
-    pub quote: QuoteFeed,
-    pub offers: OffersFeed,
+    pub quote: watch::Receiver<Option<Quote>>,
+    pub offers: watch::Receiver<MakerOffers>,
     pub cfds: watch::Receiver<Option<Vec<Cfd>>>,
-}
-
-pub struct QuoteFeed {
-    pub btc_usd: watch::Receiver<Option<Quote>>,
-    pub eth_usd: watch::Receiver<Option<Quote>>,
-}
-
-pub struct OffersFeed {
-    pub btc_usd: watch::Receiver<MakerOffers>,
-    pub eth_usd: watch::Receiver<MakerOffers>,
 }
 
 impl Actor {
@@ -108,43 +98,26 @@ impl Actor {
     ) -> (Self, Feeds) {
         let (tx_cfds, rx_cfds) = watch::channel(None);
 
-        let (tx_order_btc_usd, rx_order_btc_usd) = watch::channel(MakerOffers {
+        let (tx_offer, rx_offer) = watch::channel(MakerOffers {
             long: None,
             short: None,
         });
-        let (tx_quote_btc_usd, rx_quote_btc_usd) = watch::channel(None);
-        let (tx_order_eth_usd, rx_order_eth_usd) = watch::channel(MakerOffers {
-            long: None,
-            short: None,
-        });
-        let (tx_quote_eth_usd, rx_quote_eth_usd) = watch::channel(None);
+        let (tx_quote, rx_quote) = watch::channel(None);
 
         let actor = Self {
             db,
             tx: Tx {
                 cfds: tx_cfds,
-                order: OrderSenders {
-                    btc_usd: tx_order_btc_usd,
-                    eth_usd: tx_order_eth_usd,
-                },
-                quote: QuoteSenders {
-                    btc_usd: tx_quote_btc_usd,
-                    eth_usd: tx_quote_eth_usd,
-                },
+                offer: tx_offer,
+                quote: tx_quote,
             },
             state: State::new(network),
             price_feed,
         };
         let feeds = Feeds {
             cfds: rx_cfds,
-            offers: OffersFeed {
-                btc_usd: rx_order_btc_usd,
-                eth_usd: rx_order_eth_usd,
-            },
-            quote: QuoteFeed {
-                btc_usd: rx_quote_btc_usd,
-                eth_usd: rx_quote_eth_usd,
-            },
+            offers: rx_offer,
+            quote: rx_quote,
         };
 
         (actor, feeds)
@@ -788,18 +761,8 @@ impl Cfd {
 /// Internal struct to keep all the senders around in one place
 struct Tx {
     cfds: watch::Sender<Option<Vec<Cfd>>>,
-    pub order: OrderSenders,
-    pub quote: QuoteSenders,
-}
-
-pub struct OrderSenders {
-    pub btc_usd: watch::Sender<MakerOffers>,
-    pub eth_usd: watch::Sender<MakerOffers>,
-}
-
-pub struct QuoteSenders {
-    pub btc_usd: watch::Sender<Option<Quote>>,
-    pub eth_usd: watch::Sender<Option<Quote>>,
+    pub offer: watch::Sender<MakerOffers>,
+    pub quote: watch::Sender<Option<Quote>>,
 }
 
 impl Tx {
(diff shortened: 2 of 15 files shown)


daemon-tests/src/flow.rs
```diff
@@ -39,15 +40,9 @@ pub async fn next_maker_offers(
...
-pub async fn is_next_offers_none(
-    rx: &mut OffersFeed,
-    contract_symbol: &ContractSymbol,
-) -> Result<bool> {
-    let mut rx = match contract_symbol {
-        ContractSymbol::BtcUsd => rx.btc_usd.clone(),
-        ContractSymbol::EthUsd => rx.eth_usd.clone(),
-    };
-    let maker_offers = next(&mut rx).await?;
+// TODO: what is this function even doing? only called from the tests and the response is ignored.
```
holzeis: ❓ @luckysori @da-kami do you know what this function is used for? It is only used by the tests, but the result is alwas ignored. I would like to remove it, but am not sure if this code is hidding some kind of magic :)

daemon-tests/src/flow.rs
```diff
@@ -14,17 +13,19 @@ const NEXT_WAIT_TIME: Duration = Duration::from_secs(if cfg!(debug_assertions) {
...
-    let (mut rx_a, mut rx_b) = match contract_symbol {
-        ContractSymbol::BtcUsd => (rx_a.btc_usd.clone(), rx_b.btc_usd.clone()),
-        ContractSymbol::EthUsd => (rx_a.eth_usd.clone(), rx_b.eth_usd.clone()),
-    };
+    let mut rx_a = rx_a.clone();
+    let mut rx_b = rx_b.clone();
 
     let non_empty_offer = |offer: MakerOffers| {
-        if offer.long.is_some() && offer.short.is_some() {
+        if offer.long.is_some()
```
da-kami: 🛠️ I think we should avoid `unwrap`s in the codebase .

Maybe try n `match` for the `Some` case with an additional `if`.
Or maybe an `if let Some(offer) = offer.long` 

---

Is it intentional that we always send out long and short together? (I guess yes)

daemon-tests/src/flow.rs
da-kami: This function ensures that (initially) the value consumed from the feed is `None`.
This is used for sanity in the tests. I think we should keep it, but am open to discussing better solutions.

da-kami (commented):
Thanks, great work! I think it's definitely better to let the caller decide and identify the sent data both the symbol.

I'll have another look once the unwrap is gone.

daemon-tests/src/flow.rs
holzeis: Happy to change, but i figured I can safely unwrap here since I've checked before for is_some()

----

Well I think it is intentional at least that's how it has been.

holzeis: fixed in https://github.com/itchysats/itchysats/pull/2679/commits/8089dac7d577f12f3bb2688b9fefc586a4b64e27

daemon-tests/src/flow.rs
holzeis: well as mentioned the response if true or false is no where used. So if we really want to ensure that we should add an assert.

maker-frontend/src/MakerApp.tsx
```diff
@@ -63,14 +63,23 @@ export default function App() {
...
     let [leverages, setLeverages] = useState(["1", "2", "3"]);
 
     const cfdsOrUndefined = useLatestEvent<Cfd[]>(source, "cfds", intoCfd);
     let cfds = cfdsOrUndefined ? cfdsOrUndefined! : [];
-    const makerLongOrder = useLatestEvent<MakerOffer>(source, "long_offer");
-    const makerShortOrder = useLatestEvent<MakerOffer>(source, "short_offer");
+    const makerLongOrder = useLatestEvent<MakerOffer>(
+        source,
+        "long_offer",
+        undefined,
```
luckysori, line 75: ❓ Why this `undefined` here but not below?

luckysori: It looks like you were meaning to use the `filter` optional argument below, so the `undefined` is missing there.

maker-frontend/src/components/Hooks.tsx
```diff
@@ -5,6 +5,7 @@ export default function useLatestEvent<T,>(
     source: EventSource,
     event_name: string,
     mapping: (key: string, value: any) => any = (key, value) => value,
+    filter?: (event: Event) => boolean,
```
luckysori, line 8: ❓ What if we don't introduce this filter and rely on the caller to pass in a source with already filtered events?

Alternatively we could also move the filter to the `event_name` itself, with names such as `btc_usd_short_offer`.

daemon/src/taker_cfd.rs
```diff
@@ -93,19 +92,11 @@ impl Actor {
...
-            .unwrap_or(ContractSymbol::BtcUsd);
-
         if let Err(e) = self
             .projection_actor
-            .send(projection::Update((
-                symbol,
+            .send(projection::Update(
                 takers_perspective_of_maker_offers.clone(),
-            )))
+            ))
```
luckysori, line 99: 👍 I think these simplifications of `taker_cfd` and `projection` are in line with what I've done in https://github.com/itchysats/itchysats/pull/2680. 

We'll have to adapt some code to merge them together, but I'm now sending lists of pure `model::Offer`s around, which is a step further than what you've done here.

maker-frontend/src/MakerApp.tsx
```diff
@@ -63,14 +63,23 @@ export default function App() {
...
+        (data: any) => data && data.contract_symbol == "BtcUsd",
+    );
+    const makerShortOrder = useLatestEvent<MakerOffer>(
+        source,
+        "short_offer",
+        (data: any) => data && data.contract_symbol == "BtcUsd",
```
luckysori: ❌ If we hard-code this to `BtcUsd` we will not filter the correct offers when we flip the switch. I think you need to use the variable `symbol` on line 63.

daemon-tests/src/flow.rs
luckysori: I agree that it wasn't working as intended but I think what we should do is simply bubble up an error if the condition is not met.

That means the return type would be `Result<()>` and the caller can just unwrap since we're in a test, and ignore the `Ok(())`.

I would vote against `assert`ing because that should really only be used to verify whatever the test is actually testing, rather than to check if there's  a bug in the implementation of the test.

daemon-tests/src/flow.rs
```diff
@@ -21,15 +21,25 @@ pub async fn next_maker_offers(
...
+        match &offer.short {
+            Some(short) => {
+                if contract_symbol != &short.contract_symbol {
+                    return None;
+                }
+            }
+            None => return None,
+        }
+
+        Some(offer)
```
luckysori: 🙃 It's a nice idea to use pattern matching here, but you can even go a bit further. Let me show you something that I consider more Rust-idiomatic:

```rust
let non_empty_offers = |offers: MakerOffers| match &offers {
    MakerOffers {
        long: Some(long),
        short: Some(short),
    } if &long.contract_symbol == contract_symbol
        && &short.contract_symbol == contract_symbol =>
    {
        Some(offers)
    }

    _ => None,
};
```

There are a couple things going on here that might be interesting:

- Pattern matching against a custom struct, which can be really powerful.
- Match [guards](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html).
- Avoiding explicit returns. I think idiomatic Rust uses returns sparingly. I'd say this team strives to use them mostly in early returns or not at all. The reason being that it's easier to read a function when you can expect it to return from only one or two predictable places.

luckysori (commented):
Looks good! 

I think there's one thing that seems wrong. The rest is less important.

bonomat:
I've added a comment to the ticket which should have been on this PR - sorry for that, only saw this PR afterwards. Can you please check it and see if this works as expected?

holzeis:
> I've added a comment to the ticket which should have been on this PR - sorry for that, only saw this PR afterwards. Can you please check it and see if this works as expected?

With the current implementation its the responsibility of the client side to filter events. I've introduced an optional filter closure that will skip all not symbol related offers.

https://github.com/itchysats/itchysats/blob/8089dac7d577f12f3bb2688b9fefc586a4b64e27/taker-frontend/src/useLatestEvent.ts#L3-L28

This will only update the state if the event is actually the symbol we are interested in.

https://github.com/itchysats/itchysats/blob/8089dac7d577f12f3bb2688b9fefc586a4b64e27/taker-frontend/src/App.tsx#L125-L130

Eventually I suggest we filter events using a query parameter on the feed api so that the events don't have to be filtered on the client side. (saving bandwidth)

maker-frontend/src/MakerApp.tsx
holzeis: good catch, I've forgot the undefined below.

bonomat:
> ...
> 

I believe it will still update the whole application because a new sse event will be sent from the server if either of the two contract symbols has changed. 
The offer which is displayed might not change in value but a re-render will be issued. 
I'm not sure react is able to detect that it should not reredender as the values are set to the same the same again. 

I think this can be easily tested: 

Create a log message in App.tax and see how often this is printed when a new offer of either type was created.

maker-frontend/src/components/Hooks.tsx
holzeis: I haven't found a way to filter events on the event source - happy to change if you know how. Regarding the alternative, I am not so sure about it. I can think of argument for and against it. I guess my key concern here is that this requires our backend to make assumptions about how events are consumed by the client. It might bring benefits for our case, but it might be cumbersome for others. Our events are currently simple changed events - adding what has changed into the event could get very complex very quickly (because where do you draw the line?) Instead I'd suggest we introduce server side filtering using e.g. a query parameter that could also resolve the need for client side filtering.

maker-frontend/src/MakerApp.tsx
holzeis: It was hardcoded before so I did not change that as it wasn't part of the issue. Happy to change though.

luckysori: AFAIK it wasn't hard-coded: we had a dynamic source based on the value of `symbol`, which changed every time `setSymbol` was called.

The way I see it, with this changes when we change the symbol to `ETHUSD` we wills till display `BTCUSD`.

holzeis: you are right.. must have changed it myself and thought it has been like that before 🙈

daemon-tests/src/flow.rs
holzeis: thank you for that rust lesson 😃

luckysori (approved):
LGTM

luckysori:
> > ...
> 
> I believe it will still update the whole application because a new sse event will be sent from the server if either of the two contract symbols has changed. The offer which is displayed might not change in value but a re-render will be issued. I'm not sure react is able to detect that it should not reredender as the values are set to the same the same again.
> 
> I think this can be easily tested:
> 
> Create a log message in App.tax and see how often this is printed when a new offer of either type was created.

The taker UI only re-renders when an offer matching the page we're viewing comes in. So that seems to be working!

taker/src/routes.rs
```diff
@@ -57,37 +56,9 @@ pub struct IdentityInfo {
     pub(crate) taker_peer_id: String,
 }
 
-#[derive(Debug, Copy, Clone, strum_macros::Display)]
-pub enum ContractSymbol {
```
da-kami, line 61: 🙃 I think having a dedicated struct for the API here is actually a good thing, because we can decide the serialization can be guaranteed in a better way. (e.g. not breaking the API with renaming a field)

da-kami (approved):

daemon/src/projection.rs
```diff
@@ -108,43 +98,26 @@ impl Actor {
     ) -> (Self, Feeds) {
         let (tx_cfds, rx_cfds) = watch::channel(None);
 
-        let (tx_order_btc_usd, rx_order_btc_usd) = watch::channel(MakerOffers {
-            long: None,
-            short: None,
-        });
-        let (tx_quote_btc_usd, rx_quote_btc_usd) = watch::channel(None);
-        let (tx_order_eth_usd, rx_order_eth_usd) = watch::channel(MakerOffers {
+        let (tx_order, rx_order) = watch::channel(MakerOffers {
```
da-kami: Please rename to `offer` here, feel free to do in follow up

taker/src/routes.rs
luckysori: I agree on principle, but this was removed because it (or anything like it) is not used at all after these changes.

da-kami: My concern was that we expose `model::ContractSymbol` as part of `projection::CfdOffers`, but that was already the case before.
Please ignore this for now.

da-kami (approved):

PR #2680, opened 2022-08-11
Add `/itchysats/offer/2.0.0`

### LINKED ISSUE #2459: `itchysats/offer` to support different trading pairs

### LINKED ISSUE #2670: Ensure network protocol compatibility to taker `0.5.0` and `0.5.3`
It seems the API changes for support `ETHUSD` broke backwards compatibility in an unexpected way:

```
2022-08-09 14:12:29	
{"timestamp":"2022-08-09T14:12:29.904465676Z","level":"DEBUG","fields":{"message":"Failed to process maker offers: Failed to decode maker offers.: JSON error: missing field `leverage` at line 1 column 391: missing field `leverage` at line 1 column 391","peer":"12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1"},"target":"xtra_libp2p_offer::taker","span":{"err":"Failed to decode maker offers.","name":"fallible task handle_error"}}
	
2022-08-09 14:11:29	
{"timestamp":"2022-08-09T14:11:29.926697913Z","level":"DEBUG","fields":{"message":"Failed to process maker offers: Failed to decode maker offers.: JSON error: missing field `leverage` at line 1 column 391: missing field `leverage` at line 1 column 391","peer":"12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1"},"target":"xtra_libp2p_offer::taker","span":{"err":"Failed to decode maker offers.","name":"fallible task handle_error"}}
	
2022-08-09 14:10:29	
{"timestamp":"2022-08-09T14:10:29.86286764Z","level":"DEBUG","fields":{"message":"Failed to process maker offers: Failed to decode maker offers.: JSON error: unknown variant `EthUsd`, expected `BtcUsd` at line 1 column 76: unknown variant `EthUsd`, expected `BtcUsd` at line 1 column 76","peer":"12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1"},"target":"xtra_libp2p_offer::taker","span":{"err":"Failed to decode maker offers.","name":"fallible task handle_error"}}
	
2022-08-09 14:09:29	
{"timestamp":"2022-08-09T14:09:29.833906787Z","level":"DEBUG","fields":{"message":"Failed to process maker offers: Failed to decode maker offers.: JSON error: unknown variant `EthUsd`, expected `BtcUsd` at line 1 column 76: unknown variant `EthUsd`, expected `BtcUsd` at line 1 column 76","peer":"12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1"},"target":"xtra_libp2p_offer::taker","span":{"err":"Failed to decode maker offers.","name":"fallible task handle_error"}}
	
2022-08-09 14:08:59	
{"timestamp":"2022-08-09T14:08:59.833733485Z","level":"DEBUG","fields":{"message":"Failed to process maker offers: Failed to decode maker offers.: JSON error: missing field `leverage` at line 1 column 390: missing field `leverage` at line 1 column 390","peer":"12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1"},"target":"xtra_libp2p_offer::taker","span":{"err":"Failed to decode maker offers.","name":"fallible task handle_error"}}
```

Here's the [query](https://itchysats.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-itchysats-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bcontainer%3D~%5C%22autotaker_release%7Ctaker_release%5C%22,setup%3D%5C%22testnet%5C%22%7D%20!%3D%5C%22TRACE%5C%22%20%7C~%5C%22Failed%20to%20process%7CFailed%20to%20decode%5C%22%22%7D%5D,%22range%22:%7B%22from%22:%221660054084709%22,%22to%22:%221660054377004%22%7D%7D) to reprocude.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2459.
Fixes #2670.

The maker supports versions 1.0.0 and 2.0.0. We keep a copy of the old version of the protocol, but remove a lot of the code that is not actually necessary (the taker side and the tests) and adapt what's left so that it is easier to integrate it in `maker` together with `/itchysats/offer/2.0.0`.

Because the maker is the dialer in the offer protocol, it means that currently all their takers will get doubled offers (one per version) and will ignore the ones that correspond to the protocol they don't support. To prevent us from generating false positive warnings we filter out the `yamux::NegotiationError::Failed` error.

Additionally:

- Rework the way the offers are modelled in different components across the entire codebase.
- The `projection::Actor` now takes a `Role` parameter, so that we can do away with the awkward `Origin` type in the new offer protocol.
- Simplify `model::Offer`, which is not actually used anywhere where backwards-compatibility matters.
- Calculate `oracle_event_id` based on `settlement_interval` when constructing `Offer`. As opposed to passing two arguments that could be out of sync, like we used to do.

---

TODOs:

- [x] Update changelog.
- [x] Update commit message to mention that the new version of the offer protocol inherently supports new contract symbols.
- [x] Consider splitting into more patches.
- [x] Check if we have some unused dependencies after these patches.


daemon/src/projection.rs
```diff
@@ -1158,8 +1208,24 @@ impl Actor {
...
-    fn handle(&mut self, msg: Update<(ContractSymbol, Option<model::MakerOffers>)>) {
-        self.tx.send_order_update(msg.0 .0, msg.0 .1.into());
+    fn handle(&mut self, msg: Update<Vec<model::Offer>>) {
+        let new_offers = msg
+            .0
+            .into_iter()
+            .filter_map(|offer| match CfdOffer::new(offer, self.role) {
+                Ok(offer) => Some(offer),
+                Err(e) => {
+                    tracing::warn!("Failed to map new offer: {e:#}");
```
Restioson: I think "failed to map" is probably a bit vague

xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -0,0 +1,139 @@
...
+                .connected_peers
+                .extend(connection_stats.connected_peers),
+            Err(e) => {
+                tracing::error!(
+                    "Unable to receive connection stats from the endpoint upon startup: {e:#}"
+                );
+
+                // This code path should not be hit, but in case we run into an error this sleep
+                // prevents a continuous endless loop of restarts.
+                tokio_extras::time::sleep(Duration::from_secs(2)).await;
```
Restioson, line 131: Is this referring to the actor restarts? If so, couldn't this be better served by a supervisor restart policy with timeout?

luckysori: I only moved this code around, but unfortunately the diff isn't very good :(

I don't fully understand why we need this, but what you suggest sounds like a great idea!!

da-kami:
Is it on purpose that this is based on a really old branch? 😅

image

daemon/src/order/maker.rs
```diff
@@ -54,7 +52,7 @@ pub struct Actor {
     n_payouts: usize,
     decision_senders: HashMap<OrderId, oneshot::Sender<protocol::Decision>>,
     db: sqlite_db::Connection,
-    latest_offers: MessageChannel<xtra_libp2p_offer::maker::GetLatestOffers, Option<MakerOffers>>,
+    latest_offers: MessageChannel<xtra_libp2p_offer::maker::GetLatestOffers, Vec<model::Offer>>,
```
da-kami, line 55: ❓Why do we use a `Vec` here and not a `Hashmap` that indexes per `symbol` as well? Or hide the whole abstraction behind a struct that hides how this is implemented?

daemon/src/projection.rs
```diff
@@ -828,26 +829,70 @@ impl Tx {
...
-                self.order.btc_usd.send(offer).unwrap_or_default();
-            }
-            ContractSymbol::EthUsd => {
-                self.order.eth_usd.send(offer).unwrap_or_default();
-            }
-        }
+    fn send_offer_update(&self, offers: &Offers) -> Result<()> {
+        let OffersByContractSymbol { btc_usd, eth_usd } = split_offers_by_contract_symbol(offers);
+
+        self.order.btc_usd.send(btc_usd)?;
```
da-kami: 🙃 `order` should actually be renamed to `offer` - this might be because this is based on an old branch?

daemon/src/projection.rs
```diff
@@ -828,26 +829,70 @@ impl Tx {
...
+
+        Ok(())
     }
 }
 
 /// Internal struct to keep state in one place
 struct State {
     network: Network,
     quote: Option<xtra_bitmex_price_feed::Quote>,
+    offers: Offers,
```
da-kami: ❓ Keeping the state for the "current / latest offers" up to date in multiple places in the codebase can be error prone. We already have the `latest_offers` in multiple places.  
What is the motivation for this (avoid one more message between actors...?)? 
I do think it makes sense to push it and store it, but I wonder if this could lead to unwanted side-effects where the state is not up to date in some cases.

daemon/src/projection.rs
da-kami: ❓ Why do we need two channels for different offers per symbol? With https://github.com/itchysats/itchysats/pull/2679 I think one channel should be enough or am I missing something?

xtra-libp2p-offer/src/deprecated.rs
```diff
@@ -0,0 +1,4 @@
+pub mod maker;
+pub mod protocol;
+
+pub const PROTOCOL: &str = "/itchysats/offer/1.0.0";
```
da-kami, line 4: ❓ Why is this called `deprecated`, when for `rollover` we added versions, i.e. `v_1_0_0` and `v_2_0_0` - I think it would help the code base to have some consistency here.

da-kami (commented):
I am not sure about some of these changes, I think it would help to do some design work before implementing a refactoring for this protocol. Was this re-design planned or did it evolve while writing the code? 

Happy to talk about it together.

daemon/src/projection.rs
luckysori: Nope this is the same in master AFAIK.

luckysori: The PR you reference hasn't been merged yet. We'll have to adapt this once that happens.

xtra-libp2p-offer/src/deprecated.rs
luckysori: Because it's much clearer (I introduced the original horrible terminology). I've adapted the rollover one in a separate PR.

daemon/src/order/maker.rs
luckysori: I wanted to keep the offer actor as simple as possible. I didn't see the need to use a `HashMap` because a simple list works, but we can consider it if there is a good reason for it.

luckysori:
> Is it on purpose that this is based on a really old branch? 😅

It's based on #2639. Which is as old as this PR.

luckysori:
> I am not sure about some of these changes, I think it would help to do some design work before implementing a refactoring for this protocol.

Sure, we can spend time further designing the new version of the protocol (we already spent some time with @holzeis, who helped me steer the PR in the current direction). If you are referring to the protocol, because all your review comments touched on implementation details.

> Was this re-design planned or did it evolve while writing the code? 

Originally we thought we would be able to reuse version 1. But at some point before working on this PR we realised that it wouldn't work. And it would be a bad idea to build on the dubious design of `MakerOffers` anyway.

Because the design of every actor that touches the offers was affected by `MakerOffers`, it is no surprise that the internals of those actors had to change (improve imo).

xtra-libp2p-offer/src/deprecated.rs
luckysori: Correction: I forgot to actually open the PR 5 days ago! [Here it is](https://github.com/itchysats/itchysats/pull/2706).

daemon/src/projection.rs
luckysori: > What is the motivation for this (avoid one more message between actors...?)?

No messages have changed. The `projection::Actor` still hears about the most recent offers via the `Update` message. What has had to change is the fact that we no longer necessarily send _all_ the current offers from maker to taker every time we run the `/itchysats/offer`. It's perfectly acceptable for the maker to only tell us about ETHUSD offers if that's the only thing that has changed.

With that in mind, I thought that it would be useful to keep a holistic view of the offers in the `projection::Actor` and update it every time we receive an `Update` message. Thinking about it now I am not totally sure that it is necessary. I wonder if it's always sufficient to relay the offers that have changed via the feed. What do you think? Keep in mind that this is different to what we used to do. Before we would always receive _all_ the offers (all BTCUSD) and pump that into the feed.

> I do think it makes sense to push it and store it, but I wonder if this could lead to unwanted side-effects where the state is not up to date in some cases.

With the current functionality in this PR what we are doing is:

1. Receive an update with the new offers.
2. Combine our state with the new offers.
3. Pushing all offers to the feeds.

I think maybe we just need:

1. Receive an update with the new offers.
2. Relay the new offers no matter what.

It's probably bad to send events for offers that haven't actually changed! I'm not 100% sure about how we're supposed to use SSE though, so maybe you have more thoughts on this.

daemon/src/order/maker.rs
luckysori: Taking as feedback to wrap Vec<model::Offer> around a offer protocol newtype.

daemon/src/projection.rs
da-kami: I think I missed that one when cleaning up 😅 - we can do it in a follow up :)

xtra-libp2p-offer/src/current/maker.rs
luckysori: I've created https://github.com/itchysats/itchysats/issues/2717 to track this!

daemon/src/projection.rs
```diff
@@ -1263,46 +1240,12 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
...
-                                tracing::warn!("Unable to convert short offer: {e:#}");
-                                None
-                            }
-                        });
-                (projection_long, projection_short)
-            }
-        };
-        MakerOffers { long, short }
-    }
+    pub btcusd_long: Option<CfdOffer>,
```
da-kami, line 1243: 🤔 This is a bit annoying, but on the `projection` this is acceptable to flatten out instead of using e.g. a `HashMap` as representation.

xtra-libp2p-offer/src/current/protocol.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    tracing::info!(?offers, "Received offers");
+
+    MESSAGES_RECEIVED.inc();
+
+    Ok(offers)
+}
+
+#[derive(Clone, Serialize, Deserialize, PartialEq, Debug)]
+pub(crate) struct Offers(Vec<Offer>);
```
da-kami, line 50: 🤔 After another look I am not sure I agree with this model.

This model *implies* that the maker sends sane data.
A maker could send multiple offers for the same `contract_symbol` and `position` with this API.
Can't we be more explicit here?

This is what popped into my head:

* long and short are fixed
* the contract_symbold should scale, i.e. we might have more in the future

This leads me to a model like this:

```rust
pub struct MakerOffers {
   long: HashMap<ContractSymbol, Offer>,
   short: HashMap<ContractSymbol, Offer>
}
```

Additionally I would add a constructor that makes it explicit that one always has to set them for both symbols.
Alternatively we could also go for explicit setters, but I think the constructor it better to ensure we always handle the `Option`.

And getters for the respective one. 

```rust
impl MakerOffers {
    pub new(
        btcusd_long: Option<Offer>, 
        btcusd_short:Option<Offer>, 
        ethusd_long:Option<Offer>, 
        ethusd_short: Option<Offer>
    ) -> Self {
        if let Some(btcusd_long) = btcusd_long {
            self.long.insert(ContractSymbol::BtcUsd, btcusd_long);
        }

        ...
    }
}
```

This will scale as good over the wire, but is more explicit about the data. We can roll with this simpler model of `Vec` if you think it's not worth it, but I think it would be better to be explicit.

---

It is then questionable of the `Offer` should have a position; but I would be OK with that redundancy. The API to set long/short within `Offers` can ensure that long is always set to `long` internally.

---

If you find the `HashMap` too cumbersome we can still model the internals using a `Vec` but I think we should still restrict what can be set in that `Vec`. At the moment the code allows too arbitrary things, because we map from one `Vec` into another, I think at least the model should ensure sanity by e.g. a constructor that restricts the combinations to what is possible.

daemon/src/projection.rs
da-kami: I'd say it's better to tackle the state of the `Offer` throughout the application (and multiple actors) in a separate follow-up. You can decide if you want to keep it in the projection as you implemented. 
Tbh I don't think it will make a big difference, given that offers are regularly published anyway, resulting in new `OfferId`s which in turn will require an update.

daemon/src/order/maker.rs
da-kami: Yeah, that way we can make the `impl` of `model::Offers` explicit to the combinations that can be set at the moment and the internal representation does not matter.

For the `Offers` sent over the wire we can consider a model that uses a `Hashmap` - or our `model::Offers` can ensure that the `protocol::Offers` is sane when converting into it. We can opt to keep a `Vec` internally for both, but then then the accessors should ensure that one can only set sane combinations and no duplicates.

daemon-tests/tests/main/offer.rs
```diff
@@ -35,10 +35,10 @@ async fn taker_receives_offer_from_maker_on_publication() {
 }
 
 fn assert_eq_offers(published: MakerOffers, received: MakerOffers) {
-    assert_eq_offer(published.btcusd_long.clone(), received.btcusd_long.clone());
-    assert_eq_offer(published.btcusd_long.clone(), received.btcusd_long.clone());
-    assert_eq_offer(published.btcusd_long.clone(), received.btcusd_long.clone());
     assert_eq_offer(published.btcusd_long, received.btcusd_long);
+    assert_eq_offer(published.btcusd_short, received.btcusd_short);
```
luckysori, line 39: Haha thanks!

xtra-libp2p-offer/src/current/protocol.rs
luckysori: > 🤔 After another look I am not sure I agree with this model.
> 
> This model _implies_ that the maker sends sane data. A maker could send multiple offers for the same `contract_symbol` and `position` with this API. Can't we be more explicit here?
> 
> This is what popped into my head:
> 
>     * long and short are fixed
> 
>     * the contract_symbold should scale, i.e. we might have more in the future
> 
> 
> This leads me to a model like this:
> 
> ```rust
> pub struct MakerOffers {
>    long: HashMap<ContractSymbol, Offer>,
>    short: HashMap<ContractSymbol, Offer>
> }
> ```

I had something similar to this at some point, but I decided against it because you don't actually get that much for the price of making the message more complicated. The taker still has to make sense of the offers they just received together with the offers they already knew about. It's not any more complicated to iterate through a simple list applying the same logic. In the end the taker will always have to:

1. Combine offers received and offers stored.
2. Only keep one copy per `(ContracSymbol, Position)`, always choosing the newest one.

> It is then questionable of the `Offer` should have a position; but I would be OK with that redundancy. The API to set long/short within `Offers` can ensure that long is always set to `long` internally.

I would only opt for something like what you suggest if the `Offer` type sent over the wire omits both the position and the contract symbol. Otherwise the taker would still have to check that offers are correctly categorised on reception. And we would have to maybe ignore offers that are incorrectly categorised?

> Additionally I would add a constructor that makes it explicit that one always has to set them for both symbols. 

I don't see the need for this. Particularly not at the protocol level. Why is it important to limit what offers can be sent when? I expect the maker to want to send offers for only one contract symbol most of the time.

> If you find the `HashMap` too cumbersome we can still model the internals using a `Vec` but I think we should still restrict what can be set in that `Vec`. At the moment the code allows too arbitrary things, because we map from one `Vec` into another, I think at least the model should ensure sanity by e.g. a constructor that restricts the combinations to what is possible.

I am not fully seeing your concern here. Whether the maker sends 4 new offers for all the supported combinations or, for example, 20 copies of the same offer, the taker will still have to process them in the same way: keep the newest per `(ContractSymbol, Position)` pair.

daemon/src/projection.rs
luckysori: It's not what I wanted to do originally, but it is the simplest thing I could come up with for the time being. Luckily we can change and break this as much as we want, even after releasing.

xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -0,0 +1,139 @@
...
+impl Actor {
+    async fn handle_connection_established(
+        &mut self,
+        msg: endpoint::ConnectionEstablished,
+        ctx: &mut xtra::Context<Self>,
+    ) {
+        tracing::trace!("Adding newly established connection: {:?}", msg.peer_id);
+        self.connected_peers.insert(msg.peer_id);
+        self.send_offers(msg.peer_id, ctx).await;
+    }
```
luckysori, line 93: I've noticed that this is not exactly what we want. Here the maker is only sharing with the taker the very last offers sent over the wire.

I think we want to be able to tell them about all the relevant offers that they have missed. To do so we would need to introduce more complex state in the maker actor i.e. a `HashMap<(ContractSymbol, Position), Offer>`. I spent some time trying to work on this, which also motivated me to model the wire message similarly (as you suggested above, @da-kami).

It gets quite complicated and I don't want to spend more time on this until we have ETHUSD almost working. To move forward I would just create a ticket and add it to the milestone. When we get to tackle it we can decide if we want to do some more remodelling depending on how much time we have.

And I've stashed my changes locally in case we want to revisit them specifically.

xtra-libp2p-offer/src/current/protocol.rs
da-kami: > I am not fully seeing your concern here. Whether the maker sends 4 new offers for all the supported combinations or, for example, 20 copies of the same offer, the taker will still have to process them in the same way: keep the newest per (ContractSymbol, Position) pair.

As long as the `maker_creation_timestamp` is sane and allow sorting we will be able to pick a "newest", otherwise it will get tricky. The taker will have an error-path to deal with that is: I received an offer with different values for the same symbol + position + timestamp - which one shall the taker pick?
 
But fair enough, this is somewhat fabricated.  I think you might be right that the cost of modelling it more complicated it not worth it. This model will scale and that's what's important.

xtra-libp2p-offer/src/current/maker.rs
da-kami: As long as *all* the latest know offers (i.e. all symbol + position combinations supported) are being sent I think that is sufficient. Currently `takers` don't really care about anything else.

daemon/src/projection.rs
```diff
@@ -789,15 +786,20 @@ impl Tx {
         let _ = self.quote.send(quote.map(|q| q.into()));
     }
 
-    fn send_offer_update(&self, offer: MakerOffers) {
-        self.offer.send(offer).unwrap_or_default();
+    fn send_offer_update(&self, offers: MakerOffers) -> Result<()> {
+        tracing::info!(?offers, "Sending maker offers to feed");
```
da-kami: ❓Did you add this log here because it is the only point in the application where `maker` and `taker` behave the same when handling "new offers"?

daemon/src/projection.rs
```diff
@@ -1112,8 +1142,24 @@ impl Actor {
...
+            .filter_map(|offer| match CfdOffer::new(offer, self.role) {
+                Ok(offer) => Some(offer),
+                Err(e) => {
+                    tracing::warn!("Failed to build CfdOffer from model::Offer: {e:#}");
+                    None
+                }
+            })
+            .collect_vec();
+
+        self.state.update_offers(new_offers);
```
da-kami, line 1156: Just to re-iterate: I don't think updating the `offers` as part of the state here is particularly useful. Feel free to just assemble the `MakerOffers` locally here and then send out an update. My understanding is that we always send out a complete update anyway, so I am not sure keeping `offers` as part of the state has any advantage.

maker/src/routes.rs
```diff
@@ -60,8 +60,10 @@ pub async fn maker_feed(
         yield wallet_info.to_sse_event();
 
         let offers = rx_offers.borrow().clone();
-        yield Event::json(&offers.long).event("long_offer");
-        yield Event::json(&offers.short).event("short_offer");
+        yield Event::json(&offers.btcusd_long).event("btcusd_long_offer");
```
da-kami, line 63: 🤔 Previously I did not notice that you opted for splitting this on the `event` level now. That's what I had suggested originally, but we discarded it because it was somewhat not scalable. 
I still think it is the cleanest solution when it comes to having less decision in the UI so I am not against this change. Still wondering what motivated you to change this :)

xtra-libp2p-offer/src/current/maker.rs
da-kami: I think this is actually outdated and should be removed because there is already a more global strategy in place. But I think it is a good idea to tackle it outside of this PR. Thanks for adding the ticket!

da-kami (approved):
Thanks for all the great work. I'll unblock this :)

daemon/src/projection.rs
luckysori: I think I was debugging and this slipped through. I will make sure I remove any such log statements.

maker/src/routes.rs
luckysori: It's the only way that I could keep the UI behaviour without introducing bigger changes. Eventually I think we want something different, but this is good enough now imo.

daemon/src/projection.rs
luckysori: If we don't compose the state here we will only send the latest offers (these will reference a single currency pair at a time at the moment) to the feed.

I wanted to keep the pre-existing feed behaviour: on every update we send everything we know, even if it's redundant. It's not efficient (and I think we should not do this eventually), but I didn't want to introduce any unexpected changes.

da-kami: Right, because the maker does not send the offers alltogeher, but sends them by symbol - even thought the new protocol would allow sending them all together (but it's questionable if that is something we would want).
Ok, then let's keep the state for now and tackle this in follow up(s).

daemon/src/projection.rs
da-kami: I think it's actually good to introduce a log statement that logs the offers - I don't think we have one since the new protocol was introduced. So feel free to keep one somewhere :)

PR #2681, opened 2022-08-12
Drop connection to peer on yamux connection error

### LINKED ISSUE #2554: Improve `Endpoint` resilience to drop "stale" connections on the `maker`

luckysori:
Fix #2554.

We already categorised certain errors emitted by yamux as `BadConnection` errors. It therefore makes sense to drop the
connection to a peer if we get one of those errors when trying to open a substream to them.

---

We don't necessarily have to close #2554 if we merge this. But this might be good enough!

### DIFF
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -530,10 +530,10 @@ impl Endpoint {
     async fn handle(
         &mut self,
         msg: OpenSubstream<Single>,
-        _: &mut Context<Self>,
+        ctx: &mut Context<Self>,
     ) -> Result<Pin<Box<dyn futures::Future<Output = Result<Substream, Error>> + Send>>, Error>
     {
-        let peer = msg.peer_id;
+        let peer_id = msg.peer_id;
         let protocols = msg.protocols;
 
         debug_assert!(
@@ -541,15 +541,29 @@ impl Endpoint {
             "Type-system enforces that we only try to negotiate one protocol"
         );
 
-        let (control, _) = self.controls.get(&peer).ok_or(Error::NoConnection(peer))?;
+        let (control, _) = self
+            .controls
+            .get(&peer_id)
+            .ok_or(Error::NoConnection(peer_id))?;
 
+        let this = ctx.address().expect("self to be alive");
         let fut = {
             let connection_timeout = self.connection_timeout;
             let control = control.clone();
             async move {
-                let (protocol, stream) =
-                    Self::open_substream(control, peer, protocols.clone(), connection_timeout)
-                        .await?;
+                let res =
+                    Self::open_substream(control, peer_id, protocols.clone(), connection_timeout)
+                        .await;
+
+                if let Err(Error::BadConnection(e)) = &res {
+                    tracing::debug!(
+                        %peer_id,
+                        "Disconnecting peer due to yamux connection error when opening substream: {e}"
+                    );
+                    let _ = this.send_async_next(Disconnect(peer_id)).await;
+                }
+
+                let (protocol, stream) = res?;
 
                 debug_assert!(
                     protocol == protocols[0],


xtra-libp2p/src/endpoint.rs
```diff
@@ -530,26 +530,40 @@ impl Endpoint {
...
+                let res =
+                    Self::open_substream(control, peer_id, protocols.clone(), connection_timeout)
+                        .await;
+
+                if let Err(Error::BadConnection(e)) = &res {
+                    tracing::debug!(
+                        %peer_id,
+                        "Disconnecting peer due to yamux connection error when opening substream: {e}"
+                    );
+                    let _ = this.send_async_safe(Disconnect(peer_id)).await;
```
Restioson: Should this not be a send_async_next (I forget the exact name)? Alternatively, could it just be `self.handle(Disconnect(peer_id), ctx).await`?

luckysori: I like your first suggestion, primarily because then we don't have to ignore the result.

Do you think it is important that we give this message priority though? I suppose the earlier we get it processed the quicker we'll stop using the faulty connection. But it didn't strike me as something that needed to be done immediately.

Restioson: I think it makes sense to give priority in case the next queued message was a Ping message or similar - in that case we will skip the message sending to the disconnected peer

da-kami (approved):
I think this is a valuable change, let's see how it behaves on testnet. 
For testing we should start multiple takers to ensure we are not dropping the connection repeatedly and kill the `maker`.

I had a look at `yamux::ConnectionError` which is what is mapped to `BadConnection` and I think all instances of this error should result in dropping the connection:

```
pub enum ConnectionError {
    /// An underlying I/O error occured.
    Io(std::io::Error),
    /// Decoding a Yamux message frame failed.
    Decode(FrameDecodeError),
    /// The whole range of stream IDs has been used up.
    NoMoreStreamIds,
    /// An operation fails because the connection is closed.
    Closed,
    /// Too many streams are open, so no further ones can be opened at this time.
    TooManyStreams,
}
```

As far as I could see unsupported protocols are not mapped into this error, which is what kept us from dropping the connection before. I think `BadConnection` is thus save to be dropped!

xtra-libp2p/src/endpoint.rs
```diff
@@ -530,26 +530,40 @@ impl Endpoint {
...
-                    Self::open_substream(control, peer, protocols.clone(), connection_timeout)
-                        .await?;
+                let res =
+                    Self::open_substream(control, peer_id, protocols.clone(), connection_timeout)
+                        .await;
+
+                if let Err(Error::BadConnection(e)) = &res {
+                    tracing::debug!(
+                        %peer_id,
+                        "Disconnecting peer due to yamux connection error when opening substream: {e}"
```
Restioson, line 561: I missed this in the initial review, but I wonder if it could be useful to add this as an `err` field (or maybe some derivative of it, like the error kind) to make it searchable in traces? This would not only apply here but in only places.

luckysori: Is your general suggestion to move our typical printed errors in logs to the `tracing` fields? I think it's a good idea. But will we lose the ability to include causes using `anyhow` (with `{e:#}`)?

PR #2684, opened 2022-08-12
Instrument load_open_cfd

Restioson:
This instruments load_open_cfd and its helper functions. It is helpful to find out where time is being spent in loading CFDs (it turns out mostly in loading rollover data). Related to #2636

### DIFF
--- a/sqlite-db/src/lib.rs
+++ b/sqlite-db/src/lib.rs
@@ -33,6 +33,9 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use std::sync::Arc;
 use time::Duration;
+use tracing::instrument;
+use tracing::Instrument;
+use tracing::Span;
 
 pub use closed::*;
 pub use failed::*;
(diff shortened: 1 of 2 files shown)


Restioson:
I've just realised that this is unfinished as it may create strange parent spans from the helper functions being called from uninstrumented database functions. Could someone mark it as a draft? It may be better to have a go at (at least minimally, covering all top-level functions) instrumenting the entire database module before merging this. 

This is likely to be a little more controversial than just instrumenting `load_open_cfd` and its helpers, and may increase the trace disk utilisation, so care and possibly a discussion is required here.

da-kami:
> This instruments load_open_cfd and its helper functions. It is helpful to find out where time is being spent in loading CFDs (it turns out mostly in loading rollover data). Related to #2636

I would say that it is expected that most of the time is needed for loading rollover data, because that is where the biggest chunk of data is located (i.e. the CETs). I am happy to add this, but I am not sure how much value it actually adds. 
I see value in favor of removing the rollover data handling from the event handling; I think we should handle the `Dlc` separate from applying the events, but that is a separate discussion :)

klochowicz:
@Restioson perhaps we could convert this one into a draft? if it's not finished?

Restioson:
> @Restioson perhaps we could convert this one into a draft? if it's not finished?

We can do that, yea. Otherwise, it could be closed and revisited the next time we want to look a bit deeper into CFD loading

klochowicz:
closing as it's not planned to be merged anytime soon according to @Restioson

PR #2695, opened 2022-08-15
Differentiate between contract setup and rollover when monitoring transactions

### LINKED ISSUE #1248: Do not look for `LockTransaction` again after rollover
Our event-sourcing system does not seem to suffer because of the `LockConfirmed` event being emitted twice, but it's still unnecessary. We should not just start monitoring all transactions after rollover.

luckysori:
Fix https://github.com/itchysats/itchysats/issues/1248.

Anyone using the `monitor::Actor` now has a bit more flexibility in defining what transactions they want to monitor. This means that we can distinguish between contract setup and rollover completions.

These changes lead to the actor no longer monitoring certain transactions unnecessarily, such as the lock transaction after a
successful rollover.

### DIFF
--- a/daemon-tests/src/mocks/monitor.rs
+++ b/daemon-tests/src/mocks/monitor.rs
@@ -35,7 +35,9 @@ impl xtra::Actor for MonitorActor {
 impl MonitorActor {
     async fn handle(&mut self, _: monitor::Sync) {}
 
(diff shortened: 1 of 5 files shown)


luckysori:
Oops, I didn't mean to `bors r+` this. I thought it was a Dependabot PR 🤦

luckysori:
Please review and I'll address the comments in a follow-up PR.

PR #2703, opened 2022-08-16
Log connection getting dropped without `ConnectionDropped` event

da-kami:
This is not a problem, but it indicates that the system is unaware of a broken connection.
If this happens a lot it can lead to issues.

### DIFF
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -354,7 +354,10 @@ impl Endpoint {
             },
         );
 
-        self.controls.insert(peer_id, (control, tasks));
+        if self.controls.insert(peer_id, (control, tasks)).is_some() {
+            tracing::warn!(%peer_id, "Missed drop event, replacing old connection")
+        }
+
         self.notify_connection_established(peer_id).await;
     }
 


luckysori (approved):

xtra-libp2p/src/endpoint.rs
```diff
@@ -354,7 +354,10 @@ impl Endpoint {
             },
         );
 
-        self.controls.insert(peer_id, (control, tasks));
+        if self.controls.insert(peer_id, (control, tasks)).is_some() {
+            tracing::debug!(%peer_id, "Missed drop event, replacing old connection")
```
Restioson: Should this be a warning given this generally should not happen, but could if the connection is replaced very fast?

da-kami: We *know* that it can happen in certain scenarios where the taker daemon is exited non-gracefully; so to some extent it is expected at the moment. Potentially we will be able to catch these cases differently, e.g. with @luckysori 's https://github.com/itchysats/itchysats/pull/2681

At this point I am not sure it should be a warning, but I agree that it *should* not happen, and if we see it often it would indicate that there is a problem. I'll change it to `warn`!

PR #2704, opened 2022-08-16
Check the electrum RPC url's network is the same as what was configured

### LINKED ISSUE #816: Check electrum network against argument
We should check the connected electrum being the same network as the application. Otherwise weird things might happen (transactions will never be found). 

Note: maybe bdk does that already?

Restioson:
This is done by comparing the hashes of the respective genesis blocks, as I couldn't find any other methods within bdk or electrum-client to do this.

Fixes #816.

### DIFF
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -1,17 +1,24 @@
 use anyhow::bail;
+use anyhow::ensure;
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
+use bdk::bitcoin::blockdata::constants;
+use bdk::bitcoin::hashes::Hash;
 use bdk::bitcoin::util::bip32::ExtendedPrivKey;
 use bdk::bitcoin::util::psbt::PartiallySignedTransaction;
 use bdk::bitcoin::Address;
 use bdk::bitcoin::Amount;
+use bdk::bitcoin::BlockHash;
+use bdk::bitcoin::Network;
 use bdk::bitcoin::OutPoint;
 use bdk::bitcoin::PublicKey;
 use bdk::bitcoin::Txid;
 use bdk::blockchain::Blockchain;
 use bdk::blockchain::ElectrumBlockchain;
 use bdk::database::BatchDatabase;
+use bdk::electrum_client;
+use bdk::electrum_client::ElectrumApi;
 use bdk::sled;
 use bdk::wallet::tx_builder::TxOrdering;
 use bdk::wallet::AddressIndex;
@@ -102,8 +109,13 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
         db_path: PathBuf,
         wallet_name: String,
     ) -> Result<(xtra::Address<Self>, watch::Receiver<Option<WalletInfo>>)> {
-        let client = bdk::electrum_client::Client::new(electrum_rpc_url)
+        let client = electrum_client::Client::new(electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
+
+        ensure!(
+            seed_and_rpc_on_same_network(&client, ext_priv_key.network)?,
+            "Wallet seed and Electrum RPC client on different networks."
+        );
 
         // Create a database (using default sled type) to store wallet data
         let db = sled::open(db_path)?;
@@ -442,6 +454,18 @@ impl LockedUtxos {
             .skip_while(|(locked_at, _)| now >= *locked_at + self.time_to_lock)
             .collect();
     }
+}
+
+/// Compare the hash of the genesis block of the electrum RPC endpoint to the expected network's
+/// genesis block hash. If they differ, the electrum RPC is not for the network that we expect.
+fn seed_and_rpc_on_same_network(rpc: &electrum_client::Client, network: Network) -> Result<bool> {
+    let network_hash = constants::genesis_block(network).block_hash();
+    let mut hash = rpc.server_features()?.genesis_hash;
+    hash.reverse(); // Sha256d hashes are displayed backwards
+    let rpc_hash = BlockHash::from_slice(&hash)
+        .context("Invalid genesis block hash returned by electrum RPC")?;
+
+    Ok(network_hash == rpc_hash)
 }
 
 #[cfg(test)]


daemon/src/wallet.rs
```diff
@@ -109,6 +117,26 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
         let client = bdk::electrum_client::Client::new(electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
 
+        let network_hash = constants::genesis_block(ext_priv_key.network).block_hash();
+        // This needs to be converted to a BlockHash first, as Sha256d hashes are displayed backward
```
Restioson: This is *weird!* Nothing we can do about it though - maybe someone else knows the history better, but judging by the comments in `BlockHash::from_byte_iter`, this is just a bitcoin oddity.

daemon/src/wallet.rs
```diff
@@ -1,3 +1,6 @@
+use crate::bitcoin::blockdata::constants;
+use crate::bitcoin::hashes::hex::FromHex;
+use crate::bitcoin::BlockHash;
```
luckysori: 🔧 I think we usually import `bitcoin` things via `bdk::bitcoin`, as you can see below.

daemon/src/wallet.rs
```diff
@@ -99,6 +103,10 @@ pub struct Actor<B, DB> {
 #[error("The transaction is already in the blockchain")]
 pub struct TransactionAlreadyInBlockchain;
 
+#[derive(thiserror::Error, Debug, Clone, Copy)]
+#[error("The electrum rpc endpoint's network is not the same as the configured network")]
+pub struct WrongRpcNetwork;
```
luckysori: Given that we are not modelling the others errors explicitly and that we are not matching against this anywhere, we would normally just add context via `anyhow::Context` or use `anyhow::bail`.

daemon/src/wallet.rs
```diff
@@ -109,6 +117,26 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
...
+        .context("Invalid genesis block hash returned by electrum RPC")?;
+
+        if rpc_hash != network_hash {
+            tracing::error!(
+                %network_hash,
+                %rpc_hash,
+                "Electrum RPC and network genesis hashes do not match (wrong RPC url for network?)",
+            );
+            return Err(anyhow::anyhow!(WrongRpcNetwork));
+        }
```
luckysori: 🔧 Given where `wallet::Actor::spawn` is called (in `main.rs` for `taker` and `maker`), we should just return an error. The log is redundant, because we'll exit with a similar error message anyway!

 I would opt for something like:

```suggestion
        ensure!(rpc_hash == seed_hash, "Electrum RPC and network genesis hashes do not match");
```

daemon/src/wallet.rs
```diff
@@ -109,6 +117,26 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
         let client = bdk::electrum_client::Client::new(electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
 
+        let network_hash = constants::genesis_block(ext_priv_key.network).block_hash();
```
luckysori: 🙃 Could we capture computing the whole predicate in a helper? It would make the method more readable if we produce a condition like `are_seed_and_rpc_on_same_network` that we can `ensure!` on.

daemon/src/wallet.rs
```diff
@@ -109,6 +117,26 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
...
+        // This needs to be converted to a BlockHash first, as Sha256d hashes are displayed backward
+        let rpc_hash = BlockHash::from_byte_iter(
+            client
+                .server_features()?
+                .genesis_hash
+                .iter()
+                .map(|c| Ok(*c)),
+        )
+        .context("Invalid genesis block hash returned by electrum RPC")?;
```
luckysori: ❓ Doesn't this work:

```suggestion
        let rpc_hash = BlockHash::from_slice(&client.server_features()?.genesis_hash)?;
```

luckysori (commented):
Thank you for looking into this. 

It's kind of strange that we don't get the `Network` directly from `electrum_client`, but you're right. It's a clever idea to compare the genesis blocks!

I'm suggesting a few things that could make this a bit simpler, but the approach LGTM.

daemon/src/wallet.rs
Restioson: Thanks, I didn't know about this method!

daemon/src/wallet.rs
Restioson: The log currently prints a bit more info than the error, giving also the hashes themselves. Maybe this is not worth printing, though?

daemon/src/wallet.rs
Restioson: This turns out to not reverse the bytes for us, so we must first do this ourselves. This is fewer lines but it may be slightly less clear. It's also nicer that we do not need to wrap everything in Ok

Restioson:
I've addressed the review comments and rebased on master

daemon/src/wallet.rs
luckysori: Maybe actually we should just log something more obvious and user-friendly like:

```
Wallet seed and Electrum RPC client on different networks.
```

A user should have an easier time figuring out why their application doesn't work then.

daemon/src/wallet.rs
```diff
@@ -95,16 +102,33 @@ pub struct Actor<B, DB> {
     sender: watch::Sender<Option<WalletInfo>>,
 }
 
+/// Compare the hash of the genesis block of the electrum RPC endpoint to the expected network's
+/// genesis block hash. If they differ, the electrum RPC is not for the network that we expect.
+fn seed_and_rpc_on_same_network(rpc: &electrum_client::Client, network: Network) -> Result<bool> {
```
luckysori: 🔧 So that the arguments match the name of the function:

```suggestion
fn seed_and_rpc_on_same_network(rpc: &electrum_client::Client, ext_priv_key: ExtendedPrivKey) -> Result<bool> {
```

luckysori: 🙃 Could we move this towards the bottom of the file? I think it's slightly easier to read files if helper code is hidden away down below 😛

luckysori (approved):
Thank you! LGTM.

daemon/src/wallet.rs
Restioson: Done!

daemon/src/wallet.rs
Restioson: Sure

da-kami (approved):

Restioson:
I think this macOS test failure is spurious

PR #2705, opened 2022-08-16
Remove unused error struct

Restioson:
This isn't used anywhere in the workspace, so it should be safe to remove. It was not found by a lint as it is a public type in a library crate.


da-kami (approved):

luckysori (approved):
Good catch!

PR #2706, opened 2022-08-16
Rename rollover crates

luckysori:
The `v_{1,2}_0_0` suffixes are noisy and too verbose. And using the word "deprecated" correctly flags the older version as one that will be phased out soon.

We leverage `pub use current::*` so that we don't have to add `current` as part of any imports.


Restioson (commented):
If a version 3 of rollover happens, would support for v1 be removed?

luckysori:
> If a version 3 of rollover happens, would support for v1 be removed?

Yep, that's why I think this is better: I think we are only ever going to support two versions of any given protocol at the same time, according to [ADR019](https://github.com/itchysats/architecture/blob/main/decision-records/019-maker-compatible-with-last-two-takers.md).

maker/src/actor_system.rs
```diff
@@ -272,8 +268,8 @@ where
                 pong_address.clone(),
                 identify_listener_actor,
                 order,
-                rollover_v_1_0_0_addr.clone(),
-                rollover_v_2_0_0_addr.clone(),
+                rollover_deprecated_addr.clone(),
```
da-kami: 🙃 maybe better to flip these so we always have `rollover_addr` first and the deprecated on second.

da-kami (approved):
I initially liked naming by version because it was explicit. But this is more intuitive when integrating the protocols, and the fact that we will unlikely support more than two version at the moment justifies this change. Let's do it :)

da-kami (approved):

PR #2719, opened 2022-08-18
Extend actor tests to test `ethusd`

### LINKED ISSUE #2699: Write happy path actor test using ETHUSD offer

da-kami:
Extend the internals to be able to configure tests per contract symbol.

Note that this results in quite a lot more actor tests, but as far as I could see the test runtime did not go up significantly.

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -250,7 +250,10 @@ pub struct OpenCfdArgs {
 
 impl OpenCfdArgs {
     fn offer_params(&self) -> OfferParams {
-        OfferParamsBuilder::new().price(self.initial_price).build()
+        OfferParamsBuilder::new()
+            .contract_symbol(self.contract_symbol)
+            .price(self.initial_price)
+            .build()
     }
 
     pub fn fee_calculator(&self) -> FeeCalculator {
@@ -287,12 +290,12 @@ impl Default for OpenCfdArgs {
 /// This allows callers to use it as a starting point for their test.
 pub async fn open_cfd(taker: &mut Taker, maker: &mut Maker, args: OpenCfdArgs) -> OrderId {
     let offer_params = args.offer_params();
-    let contract_symbol = offer_params.contract_symbol;
     let OpenCfdArgs {
         oracle_data,
         position_maker,
         quantity,
         taker_leverage,
+        contract_symbol,
         ..
     } = args;
 
@@ -301,16 +304,14 @@ pub async fn open_cfd(taker: &mut Taker, maker: &mut Maker, args: OpenCfdArgs) -
     tracing::debug!("Sending {offer_params:?}");
     maker.set_offer_params(offer_params).await;
 
-    let (_, received) = next_maker_offers(
-        maker.offers_feed(),
-        taker.offers_feed(),
-        &ContractSymbol::BtcUsd,
-    )
-    .await
-    .unwrap();
+    let (_, received) =
+        next_maker_offers(maker.offers_feed(), taker.offers_feed(), &contract_symbol)
+            .await
+            .unwrap();
 
     tracing::debug!("Received from maker {received:?}");
 
+    // FIXME: At some point we will need different announcement for different symbols
     mock_oracle_announcements(maker, taker, oracle_data.announcements()).await;
 
     let offer_to_take = match (contract_symbol, position_maker) {
@@ -1054,6 +1055,12 @@ impl OfferParamsBuilder {
 
     pub fn leverage_choices(mut self, choices: Vec<Leverage>) -> Self {
         self.0.leverage_choices = choices;
+
+        self
+    }
+
+    pub fn contract_symbol(mut self, contract_symbol: ContractSymbol) -> Self {
+        self.0.contract_symbol = contract_symbol;
 
         self
     }
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -515,6 +515,13 @@ impl FeeCalculator {
 
         (maker_fee_account.balance(), taker_fee_account.balance())
     }
+
+    pub fn complete_fee_for_expired_settlement_event(&self) -> (SignedAmount, SignedAmount) {
+        // The expected to be charged for 24h because we only charge one full term
+        // This is due to the rollover falling back to charging one full term if the event is
+        // already past expiry.
+        self.complete_fee_for_rollover_hours(24)
+    }
 }
 
 fn oracle_pk() -> XOnlyPublicKey {
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -26,6 +26,7 @@ use daemon::seed::RandomSeed;
 use daemon::seed::Seed;
 use daemon::Environment;
 use daemon::N_PAYOUTS;
+use maia::olivia::btc_example_0;
 use maia::OliviaData;
 use maker::cfd::OfferParams;
 use model::libp2p::PeerId;
@@ -276,7 +277,7 @@ impl Default for OpenCfdArgs {
             initial_price: Price::new(dummy_price()).unwrap(),
             quantity: Contracts::new(100),
             taker_leverage: Leverage::TWO,
-            oracle_data: OliviaData::example_0(),
+            oracle_data: btc_example_0(),
         }
     }
 }
--- a/daemon-tests/src/maia.rs
+++ b/daemon-tests/src/maia.rs
@@ -8,6 +8,7 @@ use time::ext::NumericalDuration;
 
 pub mod olivia {
     use super::btc;
+    use super::eth;
     use super::OliviaData;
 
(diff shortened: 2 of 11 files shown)


da-kami:
There are still some TODOs in the code that don't *have* to be actioned, but get higher confidence in the code actually behaving as expected it might be better. 
Generally my confidence with adding these tests is not super high yet, because we are not testing with "ETH/USD" prices (or values in general) yet. 

@klochowicz it would be great if you could take this PR work over :)

daemon-tests/src/lib.rs
```diff
@@ -272,7 +273,7 @@ impl Default for OpenCfdArgs {
             initial_price: Price::new(dummy_price()).unwrap(),
             quantity: Contracts::new(100),
             taker_leverage: Leverage::TWO,
-            oracle_data: OliviaData::example_0(),
+            oracle_data: btc_example_0(),
```
luckysori, line 283: 🙃 The name `btc_example_0` could apply to so many things.

daemon-tests/src/maia.rs
```diff
@@ -27,6 +27,16 @@ pub mod olivia {
             &btc::ATTESTATIONS_1,
         )
     }
+
+    // TODO: Use real-world ETH oracle data.
```
luckysori, line 32: It should be exactly the same as BTCUSD, but with a different event ID. For example:

```
/x/BitMEX/BETH/2022-08-26T01:08:00.price?n=20
```

This means we don't have to wait for olivia to support ETHUSD. We can just pull the values from any random BTCUSD event.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -2,22 +2,35 @@ use daemon::projection::CfdState;
...
+}
+
+#[otel_test]
+async fn collaboratively_close_an_open_eth_usd_cfd_maker_going_short() {
+    collaboratively_close_an_open_cfd(Position::Short, ContractSymbol::EthUsd).await;
+}
+
+#[otel_test]
+async fn collaboratively_close_an_open_eth_usd_cfd_maker_going_long() {
+    collaboratively_close_an_open_cfd(Position::Long, ContractSymbol::EthUsd).await;
```
luckysori, line 33: ❓ Do we really need long and short versions of these tests? 

I do think there's probably value in having one test per contract symbol because the quote will come from a different place. Although the test won't do anything to check if the values are correct at the end. So even that is debatable.

daemon-tests/tests/main/non_collaborative_settlement.rs
```diff
@@ -2,24 +2,36 @@ use daemon_tests::open_cfd;
...
-async fn force_close_open_cfd(position_maker: Position) {
+#[otel_test]
+async fn force_close_an_open_eth_usd_cfd_maker_going_short() {
+    force_close_open_cfd(Position::Short, ContractSymbol::EthUsd).await;
+}
+
+#[otel_test]
+async fn force_close_an_open_eth_usd_cfd_maker_going_long() {
+    force_close_open_cfd(Position::Long, ContractSymbol::EthUsd).await;
+}
```
luckysori, line 27: 🤔 Also not sure if we need two of these.

daemon-tests/tests/main/offer.rs
```diff
@@ -9,7 +9,18 @@ use model::Leverage;
...
+async fn taker_receives_btc_usd_offer_from_maker_on_publication() {
+    taker_receives_offer_from_maker_on_publication(ContractSymbol::BtcUsd).await;
+}
+
+#[otel_test]
+async fn taker_receives_eth_usd_offer_from_maker_on_publication() {
+    taker_receives_offer_from_maker_on_publication(ContractSymbol::EthUsd).await;
+}
+
+// TODO: Potentially create a test where we send out both and ensure we receive as expected
```
luckysori: 👍 I think we should do that as part of this PR.

daemon-tests/tests/main/order.rs
```diff
@@ -189,6 +198,9 @@ async fn taker_places_order_for_same_offer_twice_results_in_two_cfds() {
     );
 }
 
+// TODO: Potentially create a test where we take one ethusd and one btc usd order to ensure both can
+// be set up and we receive both on the cfd feed
+
 /// To be used for the first contract setup
 async fn first_contract_setup(maker: &mut Maker, taker: &mut Taker, order_id: OrderId) {
```
luckysori: 🔧 What is a "first" contract setup? Every contract setup is the first and only one. Can we rename this?

daemon-tests/tests/main/order.rs
```diff
@@ -189,6 +198,9 @@ async fn taker_places_order_for_same_offer_twice_results_in_two_cfds() {
     );
 }
 
+// TODO: Potentially create a test where we take one ethusd and one btc usd order to ensure both can
+// be set up and we receive both on the cfd feed
```
luckysori: 👎 I don't really see the value of this. It seems like an arbitrary combination that shouldn't tell us anything interesting about the system. I would remove this TODO.

daemon-tests/tests/main/rollover.rs
```diff
@@ -36,10 +38,27 @@ async fn rollover_an_open_cfd_maker_going_short() {
     .await;
 }
 
+#[otel_test]
+async fn rollover_an_open_eth_usd_cfd_maker_going_short() {
+    let (mut maker, mut taker, order_id, fee_calculator) =
+        prepare_rollover(Position::Short, ContractSymbol::EthUsd, eth_example_0()).await;
+
+    // We charge 24 hours for the rollover because that is the fallback strategy if the timestamp of
+    // the settlement-event is already expired
```
luckysori: 🙃 I hate this comment so much :D 

What if we create a method `complete_fee_for_expired_settlement_event` that includes this and why we want to use it (because our test oracle events are hard-coded to be in the past)?

That way we can remove these noisy comments from several places.

luckysori (approved):
Thanks!

daemon-tests/src/lib.rs
klochowicz: I thought I improved it over `example_0` 😂 happy to change it over to any other suggestion :)

daemon-tests/src/maia.rs
klochowicz: thanks, will use the suggested ID! I wasn't sure what was the format as I wasn't following olivia closely.

klochowicz: I'll stick to the current BTC events for now, and will update with some real values for ETHUSD when olivia supports it (along with different quotes).

daemon-tests/tests/main/order.rs
klochowicz: for some reason Daniel has created 2 functions - one for the "first" and then for "additional" contract setup. I'll investigate whether we need both and remove one if not needed.

I'm surprised by this too.

daemon-tests/tests/main/order.rs
klochowicz: removed :-)

daemon-tests/tests/main/order.rs
klochowicz: it turns out that one was enough. incorporated it in a commit in this PR.

daemon-tests/tests/main/collaborative_settlement.rs
klochowicz: they don't add anything more, and it's trivial to test it. At least for now I'd keep them whilst we're getting more confidence.

daemon-tests/src/maia.rs
luckysori: Can already do this, e.g. https://outcome.observer/h00.ooo/x/BitMEX/BETH/2022-08-29T05:00:00.price?n=20

daemon-tests/tests/main/offer.rs
```diff
@@ -34,15 +39,25 @@ async fn taker_receives_offer_from_maker_on_publication(contract_symbol: Contrac
                 .build(),
         )
         .await;
+}
 
+/// Verify that an offer with a given contract symbol gets published by the
+/// maker an is received by the taker
+async fn assert_offer(maker: &mut Maker, taker: &mut Taker, symbol: ContractSymbol) {
+    publish_offer(maker, symbol).await;
```
luckysori: ❓ Isn't it a bit weird to act and assert in the same test function? I would expect the `publish` bit to happen outside. When reading the tests I was quite confused.

luckysori (approved):
Thanks!

daemon-tests/tests/main/offer.rs
klochowicz: I'll rename the function, as indeed `assert_` name was confusing.

daemon-tests/src/maia.rs
klochowicz: thanks, I'll post a small follow-up today that uses them! in the meantime, I'll bors this so we can have these running on the CI already.

PR #2726, opened 2022-08-20
Log issuing all events in one place

da-kami:
When checking for the `ManualCommit` in the logs I noticed that there are still quite a few events where we don't have the`peer_id` available. For overall better logs I suggest to...

Stop logging events in the model all over the place and created on dedicated way to log all events.
This helps with following the history of a cfd and enhance the logs with further information in a structured way (e.g. log the `peer_id` for all events issued).

Given that we want to log most events on `info`, but some on `warn` and `error` we distinguish by event kind.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1396,14 +1396,12 @@ impl Cfd {
         };
 
         match event {
-            EventKind::ContractSetupFailed => {
-                tracing::error!(order_id = %self.id, peer_id=?self.counterparty_peer_id, message);
-            }
-            EventKind::OfferRejected
-            | EventKind::RolloverRejected
-            | EventKind::CollaborativeSettlementRejected
+            EventKind::ContractSetupFailed
             | EventKind::RolloverFailed
             | EventKind::CollaborativeSettlementFailed
+            | EventKind::OfferRejected
+            | EventKind::RolloverRejected
+            | EventKind::CollaborativeSettlementRejected
             | EventKind::RefundConfirmed
             | EventKind::RevokeConfirmed => {
                 tracing::warn!(order_id = %self.id, peer_id=?self.counterparty_peer_id, message);
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1402,6 +1402,7 @@ impl Cfd {
             | EventKind::OfferRejected
             | EventKind::RolloverRejected
             | EventKind::CollaborativeSettlementRejected
+            | EventKind::CetConfirmed
             | EventKind::RefundConfirmed
             | EventKind::RevokeConfirmed => {
                 tracing::warn!(order_id = %self.id, peer_id=?self.counterparty_peer_id, message);
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1197,8 +1197,6 @@ impl Cfd {
             )
         }
 
-        tracing::info!(order_id = %self.id, peer_id=?self.counterparty_peer_id, "Contract setup was completed");
-
         Ok(self.event(EventKind::ContractSetupCompleted { dlc: Some(dlc) }))
     }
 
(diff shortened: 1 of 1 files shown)


model/src/cfd.rs
```diff
@@ -1403,10 +1379,45 @@ impl Cfd {
         }))
     }
 
+    fn event_with_reason(&self, event: EventKind, reason: anyhow::Error) -> CfdEvent {
```
bonomat: I find the name `reason` weird. Can we call it `error` ? Same for `log_cfd_event`

da-kami: We can - we used `reason` in rejection cases and `error` in failed cases. I generally don't like the `reason` terminology. we can opt to rename this throughout, but have to be careful not to break database backwards compatibility (I don't think we save the reason, so it should be fine, but this should be double-checked).

model/src/cfd.rs
```diff
@@ -1403,10 +1379,45 @@ impl Cfd {
...
+        self.log_cfd_event(&event, Some(reason));
+        self.issue_event_internal(event)
+    }
+
     fn event(&self, event: EventKind) -> CfdEvent {
+        self.log_cfd_event(&event, None);
+        self.issue_event_internal(event)
+    }
+
+    fn issue_event_internal(&self, event: EventKind) -> CfdEvent {
```
klochowicz: not sure if this one is worth it; after all, it allows users of this API to bypass your mandatory logging introduced in this PR 🙃
I'd consider inlining these 2 calls.

model/src/cfd.rs
klochowicz: throwing bunch of other names that could work:
- `detail` 
- `context` (a bit overloaded in the codebase)

Note: I would be hesitant to remove "reason" wholesale, I think it has its merit in other places - e.g. as "reason for rejecting an offer / settlement / something else" it is perfectly valid. I agree with @bonomat that in this context it is a bit awkward, because we abstracted away from these particular events here.

model/src/cfd.rs
```diff
@@ -1403,10 +1379,45 @@ impl Cfd {
...
         CfdEvent::new(self.id, event)
     }
 
+    fn log_cfd_event(&self, event: &EventKind, reason: Option<anyhow::Error>) {
```
klochowicz: especially here `reason` could be renamed to something more substantial, e.g. `detail` or `context`

klochowicz (approved):
nice logging improvement!

model/src/cfd.rs
```diff
@@ -1403,10 +1379,45 @@ impl Cfd {
...
+        let message = match reason {
+            None => format!("Issuing event: {event}"),
+            Some(reason) => format!("Issuing event: {event} with reason: {reason:#}"),
+        };
+
+        match event {
+            EventKind::ContractSetupFailed
+            | EventKind::RolloverFailed
+            | EventKind::CollaborativeSettlementFailed => {
```
luckysori: 🤔 You are silently promoting these two from warning to error. Is that what we want? I'm not necessarily against it, but the change shouldn't be buried in a refactoring commit.

da-kami: I added a separate commit for this

klochowicz: I'm not sure we should promote these to errors right now. We append failures here when the other party rejected, and they're not error cases.

da-kami: What is an `error` for you, and how would you define when to use it?

I do agree that the `...Failed` events are strictly not critical errors if they don't happen repeatedly, but at the same time if we run into a failure when executing one of the protocols I feel we should get triggered to look into it. Maybe a warning is sufficient.
We treat `ContractSetupFailed` as `error` - would you downgrade that to be a warning then?

da-kami: I harmonized the cases to all be on `warn` now. I do think this is more correct, because strictly speaking non of these events happening is critical on it's own. 
At some point we said that `ContractSetup` is the one that should not fail, because it could lead to scenarios where one party publishes the `lock` transaction and the other one does not. Not sure if that is enough of a reason to log `ContractSetupFailed` on `error`. We generally might want to work on our alerts and configure to receive alerts on warnings - or a specific amount of warnings of a certain type within a time interval.

model/src/cfd.rs
```diff
@@ -1396,14 +1396,12 @@ impl Cfd {
         };
 
         match event {
```
da-kami, line 1398: Arguably we could also say that force close should be a warning, because it is not a happy path scenario. What do you think @klochowicz @bonomat ?

bonomat: Yes, maybe a good idea. I think it won't hurt to make it stand out a bit more than happy path.

holzeis (approved):

model/src/cfd.rs
klochowicz: I like this idea! we try to be alerted to it on grafana, this will make it stand out more indeed.

klochowicz (approved):

PR #2728, opened 2022-08-22
Show outdated warning if daemon is outdated

### LINKED ISSUE #2716: New version available disclaimer is not showing up
Currently, when running an outdated version the disclaimer that a newer one is available is not displayed in the UI.
This was tested with `0.5.4` on Umbrel when a newer version `0.5.5` was already.

Previously we prepared a fix in https://github.com/itchysats/itchysats/pull/2573/files which was superseded by another PR.
The superseding fix should include the PR, but the solution looked a bit different.
We should reproduce this locally for testing and then fix it again.

bonomat:
We need to manually open the disclosure if outdated.

Resolves #2716

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -208,7 +208,14 @@ export const App = () => {
     const {
         isOpen: incompatibleWarningIsVisible,
         onClose: onCloseIncompatibleWarning,
+        onOpen: onOpenIncompatibleWarning,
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thanks! I can assume you've tested it locally.

bonomat:
> Thanks! I can assume you've tested it locally.

I did indeed :) 

bors r+

PR #2730, opened 2022-08-22
Introduce Lot Size

### LINKED ISSUE #2660: Let maker define `lot_size` in offer
We hard-coded it here for simplicity:

https://github.com/itchysats/itchysats/blob/7a17415dfcc7177381fd7521bd693a1a514517ab/daemon/src/projection.rs#L1382

With the introduction of ETHUSD we want this to be configurable by the maker.

bonomat:
Resolves #2660


model/src/lib.rs
```diff
@@ -62,15 +62,14 @@ pub enum Error {
     NegativePrice,
 }
 
-/// Represents "quantity" or "contract size" in Cfd terms
```
klochowicz, line 65: is this comment not helpful anymore?

model/src/lib.rs
```diff
@@ -192,44 +191,44 @@ impl fmt::Display for Leverage {
     }
 }
 
-// add impl's to do algebra with Usd, Leverage, and ExhangeRate as required
-impl Mul<Leverage> for Usd {
-    type Output = Usd;
+// add impl's to do algebra with Quantity, Leverage, and ExhangeRate as required
```
klochowicz:
```suggestion
```

this was a vague TODO without even stating that it was a TODO. I suggest getting rid of it, as of course people will add code that's needed when it's needed :-)

klochowicz (approved):
nice one!

model/src/lib.rs
```diff
@@ -62,15 +62,14 @@ pub enum Error {
     NegativePrice,
 }
 
-/// Represents "quantity" or "contract size" in Cfd terms
 #[derive(Debug, Copy, Clone, PartialEq, PartialOrd, Serialize, Deserialize)]
-pub struct Usd(Decimal);
+pub struct Contracts(Decimal);
```
luckysori, line 67: 🤔 Do we want to model contracts using `Decimal` instead of non-negative integers, like `u64`?

I would assume that fractional contracts make no sense.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `/itchysats/offer/2.0.0` protocol, which supports any contract symbol.
 - New dynamic route for maker `PUT /<contract_symbol>/offer/`.
   This allows the creation of offers with `<contract_symbol>`.
+- Introducing a new field `lot_size`: Lot Size (in number of contracts) is the minimum trading unit of a contract. Order quantity need to be a multiple of this.
```
luckysori: 🙃 It would be good to mention that this field has been added to the _offer_.

model/src/lib.rs
bonomat: I opted for decimal for simpler calculations. Might be a bit more work to change to `u64`. What do you think?

luckysori: I think it's important for our core types to be correct, so I would opt for `u64`.

I would only convert to `Decimal` whenever we need to calculate things. We already have to do this for types backed by `Amount` or `SignedAmount`.

In any case, we don't have to tackle this now. Perhaps we can create an issue if you agree with my assessment.

bonomat: I'll create a ticket. I think we should then use checked math for contract arithmethic because of overflows etc.

bonomat: https://github.com/itchysats/itchysats/issues/2735

PR #2736, opened 2022-08-24
Use ETHUSD quotes

### LINKED ISSUE #2718: Use the correct quote depending on the contract symbol
Every time we fetch the latest quote from the `xtra_bitmex_price_feed::Actor` using the `LatestQuote` message we specify `ContractSymbol::BtcUsd`. We can no longer hard-code it because we want to support other contract symbols i.e. ETHUSD.

- For the taker this is important, because it is the price used in the settlement proposal.
- For the maker this is (less) important, because it is only used to pre-fill the short and long prices of the offer to be created via the UI.

klochowicz:

### DIFF
--- a/daemon-tests/tests/main/collaborative_settlement.rs
+++ b/daemon-tests/tests/main/collaborative_settlement.rs
@@ -1,8 +1,8 @@
 use daemon::projection::CfdState;
 use daemon_tests::confirm;
-use daemon_tests::dummy_quote;
 use daemon_tests::flow::next_with;
 use daemon_tests::flow::one_cfd_with_state;
+use daemon_tests::mock_quotes;
 use daemon_tests::open_cfd;
 use daemon_tests::start_both;
 use daemon_tests::wait_next_state;
@@ -24,12 +24,7 @@ async fn collaboratively_close_an_open_cfd_maker_going_long() {
 async fn maker_rejects_collab_settlement_after_commit_finality() {
     let (mut maker, mut taker) = start_both().await;
     let order_id = open_cfd(&mut taker, &mut maker, OpenCfdArgs::default()).await;
-
-    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    let mut quote_receiver = taker.quote_feed().clone();
-    next_with(&mut quote_receiver, |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
-
+    mock_quotes(&mut maker, &mut taker).await;
     taker.system.propose_settlement(order_id).await.unwrap();
 
     wait_next_state!(
@@ -50,11 +45,7 @@ async fn maker_rejects_collab_settlement_after_commit_finality() {
 async fn maker_accepts_collab_settlement_after_commit_finality() {
     let (mut maker, mut taker) = start_both().await;
     let order_id = open_cfd(&mut taker, &mut maker, OpenCfdArgs::default()).await;
-
-    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    let mut quote_receiver = taker.quote_feed().clone();
-    next_with(&mut quote_receiver, |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
+    mock_quotes(&mut maker, &mut taker).await;
 
     taker.system.propose_settlement(order_id).await.unwrap();
 
@@ -83,10 +74,7 @@ async fn collaboratively_close_an_open_cfd(position_maker: Position) {
         },
     )
     .await;
-    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    let mut quote_receiver = taker.quote_feed().clone();
-    next_with(&mut quote_receiver, |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
+    mock_quotes(&mut maker, &mut taker).await;
 
     taker.system.propose_settlement(order_id).await.unwrap();
 
--- a/daemon/src/command.rs
+++ b/daemon/src/command.rs
@@ -35,6 +35,20 @@ impl Executor {
         }
     }
 
+    /// A way of querying values of a CFD, without issuing any commands or
+    /// mutating its state.
+    ///
+    /// Can be useful when preparing a command.
+    pub async fn query<T>(&self, id: OrderId, query: impl Fn(&Cfd) -> Result<T>) -> Result<T> {
+        let cfd = self
+            .db
+            .load_open_cfd(id, ())
+            .await
+            .context("Failed to load CFD")?;
+
+        query(&cfd).context("Failed to execute command on CFD")
+    }
+
     pub async fn execute<T: ExtractEventFromTuple>(
         &self,
         id: OrderId,
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -106,8 +106,10 @@ where
         + Handler<wallet::Withdraw, Return = Result<Txid>>
         + Handler<wallet::Sync, Return = ()>
         + Actor<Stop = ()>,
-    P: Handler<xtra_bitmex_price_feed::LatestQuote, Return = Option<xtra_bitmex_price_feed::Quote>>
-        + Actor<Stop = xtra_bitmex_price_feed::Error>,
+    P: Handler<
+            xtra_bitmex_price_feed::GetLatestQuotes,
+            Return = xtra_bitmex_price_feed::LatestQuotes,
+        > + Actor<Stop = xtra_bitmex_price_feed::Error>,
 {
     #[instrument(
         name = "Create TakerActorSystem",
@@ -392,14 +394,17 @@ where
 
     #[instrument(skip(self), err)]
     pub async fn propose_settlement(&self, order_id: OrderId) -> Result<()> {
-        // TODO: we need to know which contract symbol the order is for
-        let latest_quote = self
+        let contract_symbol = self
+            .executor
+            .query(order_id, |cfd| Ok(cfd.contract_symbol()))
+            .await?;
+
+        let latest_quote = *self
             .price_feed_actor
-            .send(xtra_bitmex_price_feed::LatestQuote(
-                xtra_bitmex_price_feed::ContractSymbol::BtcUsd,
-            ))
+            .send(xtra_bitmex_price_feed::GetLatestQuotes)
             .await
             .context("Price feed not available")?
+            .get(&into_price_feed_symbol(contract_symbol))
             .context("No quote available")?;
 
         let quote_timestamp = latest_quote
@@ -468,6 +473,13 @@ impl Environment {
             "docker" => Environment::Docker,
             _ => Environment::Unknown,
         }
+    }
+}
+
+fn into_price_feed_symbol(symbol: model::ContractSymbol) -> xtra_bitmex_price_feed::ContractSymbol {
+    match symbol {
+        model::ContractSymbol::BtcUsd => xtra_bitmex_price_feed::ContractSymbol::BtcUsd,
+        model::ContractSymbol::EthUsd => xtra_bitmex_price_feed::ContractSymbol::EthUsd,
     }
 }
 
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -57,6 +57,7 @@ use tokio::sync::watch;
 use tracing::info_span;
 use tracing::Instrument;
 use xtra::prelude::MessageChannel;
+use xtra_bitmex_price_feed::GetLatestQuotes;
 use xtra_productivity::xtra_productivity;
 use xtras::SendAsyncNext;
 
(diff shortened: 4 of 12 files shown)


klochowicz:
converting to draft until I resolve the test failures that crept in whilst I was refactoring :/

daemon/src/command.rs
```diff
@@ -35,6 +35,20 @@ impl Executor {
...
+    pub async fn query<T>(&self, id: OrderId, query: impl Fn(&Cfd) -> Result<T>) -> Result<T> {
+        let cfd = self
+            .db
+            .load_open_cfd(id, ())
+            .await
+            .context("Failed to load CFD")?;
+
+        query(&cfd).context("Failed to execute command on CFD")
+    }
+
```
klochowicz, line 51: I would like to bring this little thing to everyone's attention. I have added this method to be able to retrieve values more easily, when some additional data is needed. "query" or "peek" into the Cfd, without issuing a command.

Feedback welcome :-)

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -24,12 +24,7 @@ async fn collaboratively_close_an_open_cfd_maker_going_long() {
 async fn maker_rejects_collab_settlement_after_commit_finality() {
     let (mut maker, mut taker) = start_both().await;
     let order_id = open_cfd(&mut taker, &mut maker, OpenCfdArgs::default()).await;
-
-    taker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    maker.mocks.mock_latest_quote(Some(dummy_quote())).await;
-    let mut quote_receiver = taker.quote_feed().clone();
-    next_with(&mut quote_receiver, |q| q).await.unwrap(); // if quote is available on feed, it propagated through the system
```
bonomat, line 31: is this not needed anymore?

taker/src/routes.rs
```diff
@@ -134,7 +135,8 @@ pub async fn feed(
                 }
                 Ok(()) = rx_quote.changed() => {
                     let quote = rx_quote.borrow().clone();
-                    yield quote.to_sse_event();
+                    yield Event::json(&quote.get(&model::ContractSymbol::BtcUsd)).event("btcusd_quote");
```
bonomat: Looks like these events are not consumed by the taker-ui. Maybe we don't have to emit them then?

xtra-bitmex-price-feed/src/lib.rs
```diff
@@ -131,10 +135,16 @@ pub enum Error {
 #[derive(Debug)]
 struct NewQuoteReceived(Quote);
 
-/// Request the latest quote from the price feed.
+/// Request the latest quote for a given contract symbol from the price feed.
 #[derive(Debug, Clone, Copy)]
 pub struct LatestQuote(pub ContractSymbol);
```
bonomat: Is this still being used?

maker-frontend/src/MakerApp.tsx
```diff
@@ -98,7 +98,20 @@ export default function App() {
...
+
+    let [priceInfo, setPriceInfo] = useState<PriceInfo | null>(null);
+
+    useEffect(() => {
+        if (symbol === Symbol.btcusd) {
+            setPriceInfo(btcPriceInfo);
+        } else if (symbol === Symbol.ethusd) {
+            setPriceInfo(ethPriceInfo);
+        } else {
+            setPriceInfo(null);
```
bonomat, line 116: Should we add an error log here? Could be hard to find out if no quote is shown and nothing is logged. I would not bother with an alert or so because we are the only one using this UI.

maker-frontend/src/MakerApp.tsx
```diff
@@ -98,7 +98,20 @@ export default function App() {
...
+        }
+    }, [btcPriceInfo, ethPriceInfo, symbol]);
```
bonomat, line 118: How often does this get executed? 
Given we have 1 event for `ethusd` and 1 event for  1 for `btcusd` I'm afraid that this gets fired too often.  

Again: we are the only one using this UI so I don't think this is a blocker but this is not ideal as this will slow down the frontend.

bonomat (commented):
Just some preliminary comments. Will have another look tomorrow if I have time :)

daemon-tests/tests/main/collaborative_settlement.rs
klochowicz: I've hidden this boilerplate inside `mock_quotes()` method - it's easier to read now.

taker/src/routes.rs
klochowicz: I don't mind disabling them altogether; I'll probably do that explicitly in a separate commit on top, so that it can be reverted if we ever change our minds about this.

xtra-bitmex-price-feed/src/lib.rs
klochowicz: nope, after a while of having them both I realised it was not needed...

maker-frontend/src/MakerApp.tsx
klochowicz: in theory, it should be executed on quote subscription interval + when user changes the active contract symbol.

as there are 2 feeds updated on the same time (maybe it would be actually better to send the hashmap in the feed?), it fires 2x in short succession.

at times it seems to go a bit more often (I've seen new log line after 10s sometimes), but it's not too often IMHO.

bonomat: > as there are 2 feeds updated on the same time (maybe it would be actually better to send the hashmap in the feed?), it fires 2x in short succession.

yes, that's what I was thinking.

daemon/src/projection.rs
```diff
@@ -1238,6 +1233,24 @@ impl From<xtra_bitmex_price_feed::Quote> for Quote {
     }
 }
 
+pub type LatestQuotes = HashMap<ContractSymbol, Quote>;
+
+/// Converts between ContractSymbol types
+fn as_contract_symbol(symbol: &xtra_bitmex_price_feed::ContractSymbol) -> ContractSymbol {
```
bonomat, line 1233: Why not using `From<xtra_bitmex_price_feed::ContractSymbol> for ContractSymbol` ?

bonomat (approved):
LGTM

daemon/src/lib.rs
```diff
@@ -471,6 +476,13 @@ impl Environment {
     }
 }
 
+fn into_price_feed_symbol(symbol: model::ContractSymbol) -> xtra_bitmex_price_feed::ContractSymbol {
```
luckysori, line 479: 🔧 `From` implementation seems more appropriate.

xtra-bitmex-price-feed/src/lib.rs
```diff
@@ -13,17 +13,18 @@ use xtra_productivity::xtra_productivity;
...
 pub struct Actor {
-    latest_quote: HashMap<ContractSymbol, Quote>,
+    latest_quotes: LatestQuotes,
 
     /// Contains the reason we are stopping.
     stop_reason: Option<Error>,
     network: Network,
 }
 
+/// Subscribes to BitMEX and retrieves latest quotes for BTCUSD and ETHUSD.
```
luckysori: 🙃 I would expect this to go on the `struct`.

daemon/src/projection.rs
```diff
@@ -567,7 +564,7 @@ impl Cfd {
         self
     }
 
-    pub fn with_current_quote(self, latest_quote: Option<xtra_bitmex_price_feed::Quote>) -> Self {
+    pub fn with_current_quote(self, latest_quotes: Option<&LatestQuotes>) -> Self {
```
luckysori, line 567: 🔧 We can remove the `Option`, right? And empty `HashMap` will have the same meaning as `None`.

daemon/src/lib.rs
```diff
@@ -392,14 +394,17 @@ where
...
+            .executor
+            .query(order_id, |cfd| Ok(cfd.contract_symbol()))
+            .await?;
+
+        let latest_quote = *self
             .price_feed_actor
-            .send(xtra_bitmex_price_feed::LatestQuote(
-                xtra_bitmex_price_feed::ContractSymbol::BtcUsd,
-            ))
+            .send(xtra_bitmex_price_feed::GetLatestQuotes)
```
luckysori, line 404: 🔧 Since you've introduced the `GetLatestQuotes` message, we could just pass that map through to the `collab_settlement` protocol and use it here:

https://github.com/itchysats/itchysats/blob/a7b51027fe2b955280e3275321017d6dec41587a/daemon/src/collab_settlement/taker.rs#L53-L59

You'd have access to the CFD's contract symbol in the closure and would be able to pick the correct quote from the map before calling `start_collab_settlement_taker`.

This would save us from introducing the `query` API on `executor`, which is a bit suspicious.

daemon/src/command.rs
```diff
@@ -35,6 +35,20 @@ impl Executor {
...
+    /// Can be useful when preparing a command.
+    pub async fn query<T>(&self, id: OrderId, query: impl Fn(&Cfd) -> Result<T>) -> Result<T> {
+        let cfd = self
+            .db
+            .load_open_cfd(id, ())
+            .await
+            .context("Failed to load CFD")?;
+
+        query(&cfd).context("Failed to execute command on CFD")
+    }
```
luckysori, line 50: 🤔 Doesn't seem like it would belong in the `Executor`.

My knowledge of event sourcing is limited, but I do think we have messed something up in the model when we have to query the aggregate for data that doesn't change throughout its lifetime. I hope this is improved if/when we split the order from the CFD.

In any case, I'm not blocking the PR on this. But I do think it sends the wrong message to consumers of `Executor` and I would prefer to do something ad-hoc like what I described in another comment.

luckysori (approved):
LGTM, but I'm not sold on the new API on `Executor`.

daemon/src/projection.rs
klochowicz: I wanted to avoid yet another `ContractSymbol` type (we have 4 already!). Currently one comes from `model` and one from `xtra_bitmex_price_feed`, and both crates don't know about each other, making it impossible to place `From` trait.

daemon/src/lib.rs
klochowicz: as spoken above, I didn't want to introduce an unnecessary dependency on the individual crates, nor create one more type just for this purpose (although if you'd like me to create `projection::ContractSymbol`, I can do that).

daemon/src/lib.rs
klochowicz: thanks for the suggestion, I'll give it a shot!

daemon/src/command.rs
klochowicz: A couple of points why I don't see it as a hack:
- one needs to know *which* command to issue, or with what parameter to use it
- I tried to propose an alternative to the previous hack, with which we have gone to great lengths (piggybacking some data as part of result tuple that often doesn't change throughout its lifetime either, e.g. https://github.com/itchysats/itchysats/blob/master/model/src/cfd.rs#L780 )
The way I see it is that `query` or `peek` API to have a glance at values of CFD 
The latter ("peek") is common in functional programming, so one can decide whether they should consume/remap an iterator or not.

I totally agree that avoiding such change when we're under time pressure is a good idea, I'll try to pass the HashMap in! Might retry this when we run into this problem again at a more convenient time :-)

daemon/src/projection.rs
klochowicz: Technically yes, I opted against it because an empty `HashMap` has a semantically different meaning in this context IMHO:
1) we don't want to use quotes (passing `None`)
2) there were no quotes - we wanted to use quotes, but there were none.

xtra-bitmex-price-feed/src/lib.rs
klochowicz: lol, indeed! :)

daemon/src/lib.rs
klochowicz: it seems it can be done, but will move around a few things (we used to check invariants before we proposed settlement, now we'll need to check them *inside* taker_cfd, after it receives the proposal. Funnily enough, inside that actor we have access to the DB so I will be able to just fish out the contract symbol then from the hashmap. 

The awkward thing is that the `ProposeSettlement` will get a HashMap of latest quotes, instead of a concrete *proposal*, which is a particular bid and ask price at a given time.

I'll show you how it looks like and we can decide then. in the end, I don't think it will be cleaner :/

daemon/src/command.rs
luckysori: > one needs to know which command to issue, or with what parameter to use it

I agree that we need to read things from the CFD. I just think it doesn't belong in the "command executor" module. Clearly this wasn't designed for reading from the CFD. 

Perhaps we could simply rebrand it or create a different wrapper around the db connection that let's us query the CFD. I'm not sure if it has any larger implications, because we are basically already doing this with `ExtractEventFromTuple`.

daemon/src/lib.rs
luckysori: I had a look too before proposing this and I thought it was weird to check invariants in a `TakerActorSystem` method anyway.

> The awkward thing is that the ProposeSettlement will get a HashMap of latest quotes, instead of a concrete proposal, which is a particular bid and ask price at a given time.

I thought we could move the feed to the `taker_cfd::Actor`. It's not used for anything else in the `TakerActorSystem`. That way you wouldn't have to include the `HashMap` in the message.

daemon/src/command.rs
klochowicz: fair enough, I see how it can be perceived as a hack!

I fully support an idea of some "query" wrapper for the DB; in most cases we actually needlessly pass in the whole DB, which can be written. I will look into this as a follow-up today to address the valuable feedback given here.

taker/src/routes.rs
klochowicz: removed quote SSE events from the taker side altogether.

daemon/src/lib.rs
luckysori: Not a huge deal, this is fine. If these methods are only used once though, we could just inline them.

daemon/src/projection.rs
luckysori: Yep, you are right! We do actually use case 1.

PR #2742, opened 2022-08-24
Use dynamic contract symbol in `model` crate

### LINKED ISSUE #2657: Allow margin and liquidation price to be calculated for other contract symbols
Currently, margin and liquidation price are calculated under the assumption that the CFD's contract symbol is BTCUSD. With the introduction of ETHUSD, we need to rectify this.

### LINKED ISSUE #2658: Allow PNL to be calculated for other contract symbols
Currently, PNL is calculated under the assumption that the CFD's contract symbol is BTCUSD. With the introduction of ETHUSD, we need to rectify this.

luckysori:
Fix #2657.
Fix #2658.

Also:

- Calculate the payout based on the contract symbol for the `projection::Actor` (c2b103c509f2ee2139085343285ede6eeb50075a).
- Calculate the funding fee based on the contract symbol (acd3eb7880356b97d837a22c42d5a3e8786222cc).
- Use a different payout curve to calculate discretised payouts depending on the contract symbol (989e128f9b8e5207f765079d0faa9f73254b23e2).


daemon/src/projection.rs
```diff
@@ -1323,11 +1351,26 @@ impl CfdOffer {
...
+                        offer.max_quantity,
+                        *leverage,
+                        offer.contract_symbol,
+                    ),
+                    Position::Short => calculate_short_liquidation_price(
+                        offer.price,
+                        offer.max_quantity,
+                        *leverage,
+                        offer.contract_symbol,
+                    ),
```
luckysori, line 1386: ❓ What should we do about this?

The liquidation price for ETHUSD is calculated based on the number of contracts, which is dynamic in the offer.

- Is the formula that we are using to calculate the quanto bankruptcy price correct?
- If so, is it okay to take an arbitrary value here since this is only informative?

model/src/cfd.rs
```diff
@@ -1797,6 +1808,27 @@ pub fn market_closing_price(bid: Price, ask: Price, role: Role, position: Positi
...
+    quantity: Contracts,
+    leverage: Leverage,
+) -> Amount {
+    match contract_symbol {
+        ContractSymbol::BtcUsd => inverse::calculate_margin(price, quantity, leverage),
+        ContractSymbol::EthUsd => quanto::calculate_initial_margin(
+            price.to_u64(),
+            quantity.to_u64(),
+            leverage,
+            dec!(0.000001),
```
bonomat: Should we introduce a constant for this somewhere?

model/src/payout_curve/quanto.rs
```diff
@@ -332,19 +325,36 @@ pub fn calculate_initial_margin(
...
+    initial_margin: Amount,
+    fee_offset: SignedAmount,
+    pnl: SignedAmount,
+) -> Result<Amount> {
+    let initial_margin = initial_margin
+        .to_signed()
+        .context("Could not convert initial margin to bitcoin::SignedAmount")?;
+
+    let payout = initial_margin + fee_offset + pnl;
+    let payout = payout.to_unsigned().unwrap_or(Amount::ZERO);
```
bonomat, line 337: Do we need a check that `payout>dust`?

daemon/src/projection.rs
bonomat: I guess we are slowly reaching the boundries of calculating everything in the daemon... 

Can we use this value as a `unit` value for `1` contract and finish the calculation in the frontend?

model/src/cfd.rs
luckysori: Done here: https://github.com/itchysats/itchysats/pull/2742/commits/0541e0c74a82f18814d40a2ca37c3eb3df9b549f.

model/src/payout_curve/quanto.rs
luckysori: This is already handled in `maia`:

https://github.com/comit-network/maia/blob/fc6b78b98407b10b55f8cfd152062ad77f98cd9f/maia/src/protocol/transactions.rs#L606-L623

daemon/src/projection.rs
luckysori: I don't think it would work well because:

- It only applies to ETHUSD. The value returned for BTCUSD will always be correct so it shouldn't be modified based on the number of contracts.
- The equation we have for ETHUSD is `liquidation_price = initial_price +/- (initial_margin / (multiplier * n_contracts))`. We would have to return all the components separately in order to compute it in the UI. It seems pretty complicated :(

bonomat: This sucks... Imho using `offer.max_quantity` here is wrong as it will show a very high (or low if going short) liquidation price. 

I guess we need to compute the whole liquidation price in frontend then

luckysori: It's interesting how the behaviour is completely different between inverse and quanto payout curves. Does this match your expectations? I've not gone back to verify the quanto formulas, but I did research it when I first implemented it.

In any case, I would prefer to document this bug in a separate issue and add it to the current milestone and iteration.

bonomat: Imho the formulas are correct if that is what you are asking. 
 
- either we decide to compute things in the daemon: then we need to add another http api which allows us to pass in the n_contracts the user wants and let the daemon compute those values
- or we compute these values purely in the frontend

I wanted to go for the second option when we ran into a similar problem with the inverse contracts. @da-kami convinced me otherwise and we made use of a nominal value and just _finalized_ it in the frontend. 

Happy to move on and create a bug/issue which captures this.

model/src/cfd.rs
```diff
@@ -1895,27 +1898,34 @@ pub fn calculate_profit_at_price(
...
-                false => short_margin - amount_changed - fee_account.balance(),
+            let short_margin = inverse::calculate_margin(opening_price, quantity, short_leverage);
+
+            let payout = {
+                let long_margin = inverse::calculate_margin(opening_price, quantity, long_leverage)
+                    .to_signed()
+                    .context("Unable to compute long margin")?;
+                let short_margin = short_margin
+                    .to_signed()
+                    .context("Unable to compute long margin")?;
```
klochowicz: I suspect a copy-paste error?

klochowicz:
for some reason I can't respond to @bonomat 's comment above:
https://github.com/itchysats/itchysats/pull/2742#discussion_r954512019

I believe we already have a precedent for that (calculating for a unit value - that's the `margin_per_lot`!)

klochowicz:
out of curiosity - what part killed the projection actor when run locally? were any of the the invariants too strict?

klochowicz (commented):
it's a biggie. only half-way through, I'll start with the rest when I'm fresh in the morning!

luckysori:
> out of curiosity - what part killed the projection actor when run locally? were any of the the invariants too strict?

It's the `Price` not being able to represent the value 0. I get that it's useful so that we avoid division-by-zero errors in the model, but we need to be able to represent 0 prices because the quanto long liquidation price can be zero.

This lead me to having to write this [hacky bit](https://github.com/itchysats/itchysats/pull/2742/files#diff-f078e5601c18fcb29f773d04d16902354c8f66a57b41aedaea122e11a6f8fd3dR1939-R1946).

I think it would be good to introduce a different, less opinionated type in `projection` instead of reusing `model::Price`.

model/src/cfd.rs
luckysori: I think that was already in `master`! But it's fixed in a latter commit anyway :D

daemon/src/projection.rs
luckysori: I've created https://github.com/itchysats/itchysats/issues/2754.

klochowicz:
> > out of curiosity - what part killed the projection actor when run locally? were any of the the invariants too strict?
> 
> It's the `Price` not being able to represent the value 0. I get that it's useful so that we avoid division-by-zero errors in the model, but we need to be able to represent 0 prices because the quanto long liquidation price can be zero.
> 
> This lead me to having to write this [hacky bit](https://github.com/itchysats/itchysats/pull/2742/files#diff-f078e5601c18fcb29f773d04d16902354c8f66a57b41aedaea122e11a6f8fd3dR1939-R1946).
> 
> I think it would be good to introduce a different, less opinionated type in `projection` instead of reusing `model::Price`.

good idea! this way we could have sound model type and a convenient thing for the UI (which of course *can* represent `0` :) ). happy for this to also be done as a follow-up, not to block the big feature addition.

daemon-tests/src/lib.rs
```diff
@@ -382,16 +385,16 @@ pub struct FeeCalculator {
     /// Funding fee for the first 24h calculated when opening a Cfd
     initial_funding_fee: FundingFee,
 
-    /// The maker's position for the Cfd
```
klochowicz, line 387: nit: did you mean to remove this comment?

model/src/cfd.rs
```diff
@@ -2752,64 +2829,69 @@ mod tests {
...
+            Price::new(dec!(1_500_000)).unwrap(),
+            Price::new(dec!(15_000_000)).unwrap(),
+        ] {
+            let long_margin =
+                calculate_margin(contract_symbol, initial_price, quantity, long_leverage);
+            let long_payout = calculate_payout_at_price(
+                contract_symbol,
                 initial_price,
-                price,
+                closing_price,
```
klochowicz, line 2873: nice one! `closing_price` makes it more clear!

model/src/cfd.rs
```diff
@@ -2752,64 +2829,69 @@ mod tests {
...
                 quantity,
-                leverage,
-                counterpart_leverage,
-                taker_long,
+                long_leverage,
+                short_leverage,
```
klochowicz, line 2876: likewise, standardising this to long and short seems like a good idea.

klochowicz (approved):
thank you for splitting it up in many detailed commits, it was a bit easier to follow this way. 

I'm keen to get it in so we can test it on testnet.

klochowicz:
rebased and going ahead with the merge, to gain valuable test data from testnet.

PR #2744, opened 2022-08-25
Ensure maker shares all offers with newly connected taker

### LINKED ISSUE #2722: Ensure maker shares all current offers when taker connects
After #2680 merges, a newly connected taker will only hear about the very last offers that the maker sent over the wire (for one contract symbol only). It's not critical because soon enough the maker will send more offers, but the UX can suffer.

The maker offer actor should instead keep track of all the current offers (latest ones, no duplicates), so that they can immediately share this information with any new taker.

_Originally posted by @luckysori in https://github.com/itchysats/itchysats/pull/2680#discussion_r949760237_

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2722.

We introduced a regression with e4f910f35070c47672c8e37e224b51e3b532e833 and this patch solves it.

We add a test which failed before introducing the changes.

### DIFF
--- a/xtra-libp2p-offer/src/current/maker.rs
+++ b/xtra-libp2p-offer/src/current/maker.rs
@@ -1,6 +1,9 @@
 use crate::current::protocol;
 use crate::current::PROTOCOL;
 use async_trait::async_trait;
+use model::ContractSymbol;
+use model::Position;
+use std::collections::HashMap;
 use std::collections::HashSet;
 use std::time::Duration;
 use tokio_extras::spawn_fallible;
@@ -15,7 +18,7 @@ use xtra_productivity::xtra_productivity;
 pub struct Actor {
     endpoint: xtra::Address<Endpoint>,
     connected_peers: HashSet<PeerId>,
-    latest_offers: Vec<model::Offer>,
+    current_offers: Offers,
 }
 
 impl Actor {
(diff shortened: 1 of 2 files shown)


xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -111,6 +119,22 @@ impl NewOffers {
 #[derive(Clone, Copy)]
 pub struct GetLatestOffers;
 
+#[derive(Clone, Default)]
+struct Offers(HashMap<(ContractSymbol, Position), model::Offer>);
```
bonomat, line 123: What am I missing that you chose a hashmap here if you always use `to_vec` before consuming it?

bonomat (approved):
Looks good. Just questioning the choice of the `HashMap`

xtra-libp2p-offer/src/current/maker.rs
luckysori: The point of the map is to ensure that we only have one offer per `contract_symbol`-`position_maker` pair. That is, when we get new ones we always replace the old ones. This is done on line 129.

If we stored it as a list we would send unnecessary and conflicting data to the newly connected takers. They would be capable of dealing with it the way we implemented the taker side. But it's not much work for the maker to do this.

Also if we just kept appending to a list the size of the message for newly connected takers would keep growing towards infinity over time 😆

bonomat: Fair enough :) 
I guess you could achieve the same thing with a vec: finding the offer in it and replacing it if found or adding it to the vec.

luckysori: Absolutely. I'm just assuming that inserting into a map is faster than finding and replacing into a vector.

klochowicz (approved):
LGTM! 👍

PR #2745, opened 2022-08-25
Supervise `projection::Actor`s

luckysori:
If this kind of actor dies and is never restarted, we are royally screwed.

---

Motivated by encountering a bug with https://github.com/itchysats/itchysats/pull/2742 which completely killed the projection actor.


bonomat (approved):
I guess this was long overdue. Thanks

klochowicz (approved):
thanks for tackling this!

PR #2746, opened 2022-08-25
Currency selector bug in taker-ui

### LINKED ISSUE #2729: Switching between contracts in taker UI has no effect
I think the PR: #2679 broke the dropdown in the taker UI. 

Switching between contract symbols will not update the rate. I believe this is because we do not receive the necessary event and do not reconnect to the stream to retrieve old offers. 

The taker UI will only update if a new offer was received. 

I tried to explain this in this comment: https://github.com/itchysats/itchysats/issues/2673#issuecomment-1212732598

Available offers: 

image

image

image

bonomat:
By always subscribing to all offer events we guarantee that we do not miss an offer.

resolves #2729

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -56,6 +56,11 @@ export enum Symbol {
     ethusd = "ethusd",
 }
 
+const BTC_USD_LONG_EVENT = "btcusd_long_offer";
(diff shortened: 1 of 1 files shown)


luckysori (approved):
LGTM, thank you!

luckysori:
bors retry

PR #2749, opened 2022-08-25
Merge back 0.5.5 patch release

klochowicz:
Without this, we don't get the tag on master, and the latest HEAD indicates that we're running an outdated version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Drop connection to peer if we get a `yamux::ConnectionError` when opening a substream to them.
   Dropping the connection should lead to a reconnect on the `taker`, which should fix problems where the `taker` has a broken connection to the `maker` for a long time.
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


bonomat (approved):
0 changes?

luckysori (commented):
I understand the need to merge back the tag, but I don't get exactly how this works. 

Release 0.5.5 was a patch release on top of 0.5.4, where a few commits were cherry-picked from master. They are still in master, but in a different order. How can this tag go back into master at all?

klochowicz:
> I understand the need to merge back the tag, but I don't get exactly how this works.
> 
> Release 0.5.5 was a patch release on top of 0.5.4, where a few commits were cherry-picked from master. They are still in master, but in a different order. How can this tag go back into master at all?

the idea is that this makes the tag reachable from `master` branch, and I can see this fixes both my magit view as the annoying popup where latest HEAD was "not the latest version" :P 
I'm not a git expert by any means though, so I can't comment more. Maybe there is a better way.

klochowicz:
> 0 changes?

yes, because all of the changes happened to already be included on master.

PR #2750, opened 2022-08-26
Model BETH Olivia events

### LINKED ISSUE #2659: Fetch ETHUSD oracle data from Olivia
This includes announcements and attestations. Currently we assume that this data always corresponds to BTCUSD, so we might need to add a `ContractSymbol` to certain types.

Until Olivia actually starts supporting these events we won't be able to use/test these changes. For the time being we can assume that the event IDs will be of the form:

```
/x/BitMEX/BETH/2022-08-02T00:00:00.price?n=20
```

This would conform to BitMEX's symbol for this index: https://www.bitmex.com/app/index/.BETH.

luckysori:
Fixes #2659.

---

We temporarily "disable" the feature with dab80b9aec412baed827467cbabe0ac7ab31aed4 until Olivia supports the new events.

### DIFF
--- a/model/src/olivia.rs
+++ b/model/src/olivia.rs
@@ -84,7 +84,7 @@ impl fmt::Display for IndexPrice {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let s = match self {
             IndexPrice::Bxbt => "BXBT",
-            IndexPrice::Beth => "BETH",
+            IndexPrice::Beth => "BXBT", // TODO: Change back to "BETH" once Olivia supports it
         };
 
         s.fmt(f)
--- a/model/src/olivia.rs
+++ b/model/src/olivia.rs
@@ -1,3 +1,4 @@
+use anyhow::bail;
 use anyhow::ensure;
 use anyhow::Context;
 use anyhow::Result;
@@ -19,6 +20,8 @@ use time::OffsetDateTime;
 use time::PrimitiveDateTime;
 use time::Time;
 use url::Url;
+
+use crate::ContractSymbol;
 
 pub const EVENT_TIME_FORMAT: &[FormatItem] =
     format_description!("[year]-[month]-[day]T[hour]:[minute]:[second]");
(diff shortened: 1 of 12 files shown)


bonomat (approved):
Very nice

model/src/olivia.rs
```diff
@@ -104,13 +155,25 @@ impl BitMexPriceEventId {
     pub fn digits(&self) -> usize {
         self.digits
     }
+
+    pub fn index_price(&self) -> IndexPrice {
+        self.index
+    }
+
+    pub fn contract_symbol(&self) -> ContractSymbol {
```
klochowicz, line 163: nit: `as_contract_symbol`?

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

sqlite-db/src/models.rs
```diff
@@ -749,18 +762,68 @@ impl From<model::olivia::BitMexPriceEventId> for BitMexPriceEventId {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(match s {
+            "BXBT" => Self::Bxbt,
+            "BETH" => Self::Beth,
+            _ => bail!("Index price {s} not supported"),
+        })
+    }
+}
+
```
klochowicz, line 806: I think you can ged rid of these manual impls with:
`#[strum(serialize_all = "UPPERCASE")]` 

(used elsewhere).

klochowicz (approved):
thanks, we're getting closer :)

model/src/olivia.rs
luckysori: It's not a conversion on `self` though. Otherwise I would agree!

sqlite-db/src/models.rs
luckysori: Let's do it with https://github.com/itchysats/itchysats/issues/2751, because it's surprisingly annoying to do this given that we temporarily _serialize_ to the same string for both.

I'll add it to the issue.

PR #2752, opened 2022-08-26
Make position metrics contract symbol aware

### LINKED ISSUE #2464: Update internal metrics to deal with new contract symbols

bonomat:
resolves #2464

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -458,6 +458,7 @@ impl From<Decimal> for Percent {
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, EnumIter, Display)]
+#[strum(serialize_all = "UPPERCASE")]
 pub enum ContractSymbol {
     BtcUsd,
     EthUsd,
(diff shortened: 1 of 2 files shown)


model/src/lib.rs
```diff
@@ -458,6 +458,7 @@ impl From<Decimal> for Percent {
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Hash, EnumIter, Display)]
+#[strum(serialize_all = "UPPERCASE")]
```
luckysori, line 461: 🔧 I know it's a bit annoying but it would be good to have a dedicated type (or simply some specific serialisation logic) in `position_metric` instead of this.

I can see my future self changing this and breaking the metrics by mistake.

luckysori (approved):

klochowicz (approved):
👍

model/src/lib.rs
klochowicz: ay, indeed, model shouldn't change because of projection needs :p

a `ContractSymbol` in projection would also allow me to write `From` trait in my latest PR, so it's not the first time it might have been handy.

klochowicz: added a new small issue, I'll put in on my plate for today.

PR #2768, opened 2022-08-29
Compute liquidation price in ui

### LINKED ISSUE #2754: Fix displayed liquidation price for ETHUSD offer
Due to the way the liquidation price is calculated for quanto payout curves, after we merge #2742 we will display incorrect liquidation prices for ETHUSD offers.

Check the thread linked below for full context and proposed solutions.

_Originally posted by @luckysori in https://github.com/itchysats/itchysats/pull/2742#discussion_r953713156_

bonomat:
Without back and forth it is impossible to compute the liquidation price for quanto contracts in the backend. Hence, we compute it in the frontend.

resolves #2754

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -1340,9 +1340,6 @@ pub struct CfdOffer {
 #[derive(Debug, Clone, Copy, PartialEq, Serialize)]
 pub struct LeverageDetails {
     pub leverage: Leverage,
-    /// Own liquidation price according to position and leverage
-    #[serde(with = "round_to_two_dp")]
-    pub liquidation_price: Price,
     /// Margin per lot from the perspective of the role
     ///
     /// Since this is a calculated value that we need in the UI this value is based on the
@@ -1371,20 +1368,6 @@ impl CfdOffer {
             .leverage_choices
             .iter()
             .map(|leverage| {
-                let liquidation_price = match own_position {
-                    Position::Long => calculate_long_liquidation_price(
-                        offer.price,
-                        offer.max_quantity,
-                        *leverage,
-                        offer.contract_symbol,
-                    ),
-                    Position::Short => calculate_short_liquidation_price(
-                        offer.price,
-                        offer.max_quantity,
-                        *leverage,
-                        offer.contract_symbol,
-                    ),
-                };
                 // Margin per lot price is dependent on one's own leverage
                 let margin_per_lot = calculate_margin(
                     offer.contract_symbol,
@@ -1415,7 +1398,6 @@ impl CfdOffer {
 
                 Ok(LeverageDetails {
                     leverage: *leverage,
-                    liquidation_price,
                     margin_per_lot,
                     initial_funding_fee_per_lot,
                 })
--- a/taker-frontend/src/components/LiquidationPrice.ts
+++ b/taker-frontend/src/components/LiquidationPrice.ts
@@ -14,3 +14,16 @@ export function liquidationPriceQuantoContracts(leverage: number, quantity: numb
         return price + quantoLiquidationPrice(leverage, quantity, price);
     }
 }
+
+// returns the liquidation price for inverse contracts e.g. BTCUSD rounded to 0
+export function liquidationPriceInverseContracts(leverage: number, price: number, isLong: boolean) {
+    if (isLong) {
+        return Math.round(price * leverage / (leverage + 1));
+    } else {
+        if (leverage === 1) {
+            // With a leverage of 1 there is theoretically no liquidation
+            return 21_000_000;
+        }
+        return Math.round(price * leverage / (leverage - 1));
+    }
+}
--- a/taker-frontend/src/components/LiquidationPrice.ts
+++ b/taker-frontend/src/components/LiquidationPrice.ts
@@ -1,0 +1,16 @@
+function quantoLiquidationPrice(leverage: number, quantity: number, price: number) {
+    return (price * (1 / leverage));
+}
+
+// returns the liquidation price for quanto contracts e.g. ETHUSD
+export function liquidationPriceQuantoContracts(leverage: number, quantity: number, price: number, isLong: boolean) {
+    if (isLong) {
+        const liquidationPrice = price - quantoLiquidationPrice(leverage, quantity, price);
+        if (liquidationPrice <= 0) {
+            return 1;
+        }
+        return liquidationPrice;
+    } else {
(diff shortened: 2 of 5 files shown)


taker-frontend/src/components/LiquidationPrice.ts
```diff
@@ -14,3 +14,16 @@ export function liquidationPriceQuantoContracts(leverage: number, quantity: numb
...
 }
+
+// returns the liquidation price for inverse contracts e.g. BTCUSD rounded to 0
+export function liquidationPriceInverseContracts(leverage: number, price: number, isLong: boolean) {
+    if (isLong) {
+        return Math.round(price * leverage / (leverage + 1));
+    } else {
+        if (leverage === 1) {
+            // With a leverage of 1 there is theoretically no liquidation
+            return 21_000_000;
```
luckysori, line 25: ❓ This confused me in the `daemon` too. Why this number? The price can be `2^20-1` according to Olivia.

taker-frontend/src/components/LiquidationPrice.ts
```diff
@@ -0,0 +1,16 @@
+function quantoLiquidationPrice(leverage: number, quantity: number, price: number) {
+    return (price * (1 / leverage));
+}
+
+// returns the liquidation price for quanto contracts e.g. ETHUSD
+export function liquidationPriceQuantoContracts(leverage: number, quantity: number, price: number, isLong: boolean) {
+    if (isLong) {
+        const liquidationPrice = price - quantoLiquidationPrice(leverage, quantity, price);
+        if (liquidationPrice <= 0) {
+            return 1;
```
luckysori, line 10: 🔧 I think this should be zero. It's only not allowed to be zero in the daemon because of how we model `Price`.

taker-frontend/src/components/LiquidationPrice.ts
```diff
@@ -0,0 +1,16 @@
+function quantoLiquidationPrice(leverage: number, quantity: number, price: number) {
```
luckysori, line 1: 🔧 Can we rename these functions? In `daemon` I added the suffix `shift`, because it is by how much the initial price has to shift before you get liquidated.

taker-frontend/src/components/LiquidationPrice.ts
```diff
@@ -0,0 +1,16 @@
...
+    return (price * (1 / leverage));
```
luckysori, line 2: 📣 Wait a minute, this is a different formula! And it's correct. In the `daemon` I had:

```
bankruptcy_price_shift = initial_margin / (multiplier * n_contracts)
```

but this is equal to

```
bankruptcy_price_shift = ((n_contracts * initial_price * multiplier) / leverage) / (multiplier * n_contracts)
                       = initial_price / leverage
```

which means we don't need the number of contracts to calculate the quanto liquidation price!!! So we can calculate it using the same components we use for the inverse one. Meaning we can compute all this in the `daemon`.

Sorry for the extra work. I had a suspicion that my equation wasn't "right", but we didn't see it at the time.

luckysori (dismissed):

taker-frontend/src/components/LiquidationPrice.ts
bonomat: It's an easter egg 😅

taker-frontend/src/components/LiquidationPrice.ts
bonomat: haha... I could have noticed that 😅 
How boring is that if we can compute everything in the daemon...

bonomat: actually, we still need the leverage from the ui for that, which mean we would need to compute a _unit_ value and finish it in the frontend.

luckysori: > actually, we still need the leverage from the ui for that, which mean we would need to compute a _unit_ value and finish it in the frontend.

We already did that for inverse CFDs, right? We could also pre-compute all possible values in the daemon based on the valid leverages 😛

bonomat: Oh... this was never working correctly ... we always computed it for leverage 2 but did not change it for leverage 1 or 3 😅

bonomat: I'm obviously biased here, but given we have to do some computations in the frontend anyways, I would be in favor of going the approach of this PR, i.e. compute the liquidation price fully in the frontend.

luckysori: I'm not a huge fan if it can be avoided, because it's duplicated logic and it's written in TS haha. I won't block the PR nevertheless.

da-kami: Our conclusion in the past was: Don't do complex calculations in the UI, especially avoid divisions (because that might lead to loss in JS). We usually went for using the `lot` and `leverage` as multiplier, but we avoided divisions. 
If we can apply this rule here as well I think it will be better than moving more complex calculations into the UI now. 
Having duplicated calculations in rust and TS is not encouraged because it can lead to problems where different components of the system disagree on state.

da-kami: I don't think this is a critical part, if we have loss with calculating the liquidation price there should be not any disagreement between the components and for the user I don't think it will matter much. But nonetheless I don't think it's a good idea to keep changing the paradigms we introduce without recording a decision because it makes the code harder to reason about. If we decide to merge this please add a ticket so we can clean this up (i.e. move it into the daemon) or record a decision afterwards.

daemon/src/projection.rs
```diff
@@ -1371,20 +1368,6 @@ impl CfdOffer {
             .leverage_choices
             .iter()
             .map(|leverage| {
-                let liquidation_price = match own_position {
```
da-kami, line 1374: 🤔 This removes information from the http API, but only the taker user interface was adapted - the maker UI is not using these values, or we just don't care about updating the deprecated UI?

Will this require changes for other consumers of the API?

da-kami (approved):
As far as I can see the changes proposed in the PR are correct and are moving us forward, but if this gets merged like this please make sure to create a ticket to move the calculation into the daemon / record a decision on how we treat calculations in the UI.
Please also make sure that we are not depending on the information that was removed from the API in other consumers (I think we don't, but I am not 100% sure).

bonomat:
Thanks for the feedback @da-kami  & @luckysori . This was actually easier than expected. I'll push a pr in a minute.

PR #2771, opened 2022-08-29
Bump Rust to 1.63

klochowicz:


sqlite-db/src/closed.rs
```diff
@@ -1036,11 +1036,7 @@ mod tests {
         db.move_to_closed_cfds().await.unwrap();
 
         let load_from_open = db.load_open_cfd::<DummyAggregate>(order_id, ()).await;
-        let load_from_events = {
-            let res = load_cfd_events(&mut *conn, order_id, 0).await.unwrap();
-
-            res
-        };
+        let load_from_events = { load_cfd_events(&mut *conn, order_id, 0).await.unwrap() };
```
luckysori: ❓ Do we even need the braces?

luckysori (approved):
LGTM.

sqlite-db/src/closed.rs
klochowicz: good point - we don't! :) removed from all instances (this code has been copy-pasted around, so it's a wide-spread pattern).

PR #2772, opened 2022-08-29
Show correct liqudation price in ui

### LINKED ISSUE #2754: Fix displayed liquidation price for ETHUSD offer

bonomat:
resolves #2754

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1917,7 +1917,6 @@ pub fn calculate_profit(payout: Amount, margin: Amount) -> (SignedAmount, Percen
 /// Compute the liquidation price for the party going long.
 pub fn calculate_long_liquidation_price(
     initial_price: Price,
-    quantity: Contracts,
     leverage: Leverage,
     contract_symbol: ContractSymbol,
 ) -> Price {
@@ -1927,18 +1926,8 @@ pub fn calculate_long_liquidation_price(
         }
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
-            let n_contracts = quantity.to_u64();
-            let multiplier = ETHUSD_MULTIPLIER;
 
-            let initial_margin =
-                quanto::calculate_initial_margin(initial_price, n_contracts, leverage, multiplier);
-
-            let liquidation_price = quanto::bankruptcy_price_long(
-                initial_margin,
-                n_contracts,
-                initial_price,
-                multiplier,
-            );
+            let liquidation_price = quanto::bankruptcy_price_long(initial_price, leverage);
 
             // The `model::Price` type does not allow non-positive values, but the quanto long
             // liquidation price can easily be 0. We avoid this problem by defaulting to 1
@@ -1955,7 +1944,6 @@ pub fn calculate_long_liquidation_price(
 /// Compute the liquidation price for the party going short.
 pub fn calculate_short_liquidation_price(
     initial_price: Price,
-    quantity: Contracts,
     leverage: Leverage,
     contract_symbol: ContractSymbol,
 ) -> Price {
@@ -1965,18 +1953,8 @@ pub fn calculate_short_liquidation_price(
         }
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
-            let n_contracts = quantity.to_u64();
-            let multiplier = ETHUSD_MULTIPLIER;
 
-            let initial_margin =
-                quanto::calculate_initial_margin(initial_price, n_contracts, leverage, multiplier);
-
-            let liquidation_price = quanto::bankruptcy_price_short(
-                initial_margin,
-                n_contracts,
-                initial_price,
-                multiplier,
-            );
+            let liquidation_price = quanto::bankruptcy_price_short(initial_price, leverage);
 
             Price::new(Decimal::from(liquidation_price))
                 .expect("liquidation price to fit into Price")
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1927,7 +1927,13 @@ pub fn calculate_long_liquidation_price(
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
 
-            let liquidation_price = quanto::bankruptcy_price_long(initial_price, leverage);
+            // TODO: this liquidation price is potentially not in-line with the _actual_
+            // liquidation price as computed in the payout curve. The problem is we do not have the
+            // fee account available at this place and always compute the liquidation
+            // price based on the initial margin.
+            // Same for short liquidation price a few lines below.
+            let liquidation_price =
+                quanto::bankruptcy_price_long(initial_price, leverage.as_decimal());
 
             // The `model::Price` type does not allow non-positive values, but the quanto long
             // liquidation price can easily be 0. We avoid this problem by defaulting to 1
@@ -1954,7 +1960,8 @@ pub fn calculate_short_liquidation_price(
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
 
-            let liquidation_price = quanto::bankruptcy_price_short(initial_price, leverage);
+            let liquidation_price =
+                quanto::bankruptcy_price_short(initial_price, leverage.as_decimal());
 
             Price::new(Decimal::from(liquidation_price))
                 .expect("liquidation price to fit into Price")
--- a/model/src/payout_curve/quanto.rs
+++ b/model/src/payout_curve/quanto.rs
@@ -621,6 +621,7 @@ mod api_tests {
 #[cfg(test)]
 mod unit_tests {
     use super::*;
+    use crate::payout_curve::ETHUSD_MULTIPLIER;
     use rust_decimal_macros::dec;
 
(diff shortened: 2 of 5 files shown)


model/src/payout_curve/quanto.rs
```diff
@@ -621,6 +621,7 @@ mod api_tests {
 #[cfg(test)]
 mod unit_tests {
     use super::*;
+    use crate::payout_curve::ETHUSD_MULTIPLIER;
```
luckysori: 🔧 I introduced an arbitrary multiplier just below. Shouldn't need to import the `ETHUSD` one.

luckysori: I think it shouldn't matter what constant you use to check for regressions!

model/src/payout_curve/quanto.rs
```diff
@@ -231,45 +230,15 @@ impl Curve {
...
-            .to_unsigned()
-            .context("Could not convert long's effective initial margin to bitcoin::Amount")?;
-
-        let bankruptcy_price = bankruptcy_price_long(
-            effective_initial_margin,
-            self.n_contracts,
-            self.initial_price,
-            self.multiplier,
-        );
+        let bankruptcy_price = bankruptcy_price_long(self.initial_price, self.leverage_long);
```
luckysori: 🤔 I think this will lead to the liquidation price not changing as we roll over and accumulate fees. 

As I mentioned in private, I think this works for the offer because there are minimal (perhaps zero) fees at that stage and it's just an informative figure.

But as we roll over I think we need to take into account how the effective margins change. Otherwise the liquidation price will remain the same. Maybe that's what we want, but that's not how we implemented this at first, so we at least need to think about this change.

da-kami: As far as I can remember this is an old problem that boils down to the margin changing because the interest (i.e. fees) is deducted from the funds locked up (which is a constraint that cannot easily be changed in our solution). The liquidation price calculated changing with the effective margin is not a feature we actively perused so far but I suppose it would be mathematically more correct. I wonder if this is something that a user would expect in behavior compared to other trading solutions out there.

model/src/cfd.rs
```diff
@@ -1965,18 +1953,8 @@ pub fn calculate_short_liquidation_price(
...
-            let initial_margin =
-                quanto::calculate_initial_margin(initial_price, n_contracts, leverage, multiplier);
 
-            let liquidation_price = quanto::bankruptcy_price_short(
-                initial_margin,
-                n_contracts,
-                initial_price,
-                multiplier,
-            );
+            let liquidation_price = quanto::bankruptcy_price_short(initial_price, leverage);
```
da-kami: 🙃 nit: maybe we can harmonize the names and order of parameters of these functions, given that `inverse::calculate_short_liquidation_price(leverage, initial_price)` is named different and the parameters are passed flipped.

da-kami (approved):

model/src/payout_curve/quanto.rs
luckysori: My intuition would be that it's more correct. It would be interesting to consider how this behaviour compares to the inverse curve as well.

I don't know what's best, but if we decide to change it, could you document this change and the implications in the commit message, @bonomat?

bonomat: I think it just clicked, let me see if I can explain it. 

The formula I used is correct but @luckysori is right, we will need to take the margin into account when doing rollover as the new bankruptcy price will differ.

The initial margin was calculated with 
`margin=(n_contracts * initial_price * Multiplier) / leverage`

but over time, the margin is getting reduced (or increased) through the fee account. 

What we have to do is calculate the _effective leverage_, i.e. the inverse of the above formula: 

`effective_leverage=(n_contracts * initial_price * Multiplier) / margin`

So basically, if we take my simplified formula of calculating the bankruptcy price of

`=EntryPrice - (EntryPrice * ((1/leverage) + InitialMargin))` 

we would need to swap `leverage` with the `effective leverage` calculated using the reduced (or increased) margin when rolling over and should get an accurate bankruptcy price when rolling over.

Does this make sense? I've played around here: 

https://docs.google.com/spreadsheets/d/1Hm9BKz1EKymGpNyto0SNShQ-CdZPC9EJqWwklvEtQ-w/edit#gid=0


I think that's actually what @luckysori was doing with his formulas but I did not get it.

bonomat: The same applies for inverse contracts as well.

bonomat: Implemented in [`f2599c1` (#2772)](https://github.com/itchysats/itchysats/pull/2772/commits/f2599c163bf6761aeca49e84953692625ffe7575)

model/src/cfd.rs
```diff
@@ -1927,7 +1927,13 @@ pub fn calculate_long_liquidation_price(
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
 
-            let liquidation_price = quanto::bankruptcy_price_long(initial_price, leverage);
+            // TODO: this liquidation price is potentially not in-line with the _actual_
+            // liquidation price as computed in the payout curve. The problem is we do not have the
+            // fee account available at this place and always compute the liquidation
+            // price based on the initial margin.
+            // Same for short liquidation price a few lines below.
```
luckysori, line 1934: 🔧 I would move this to the docstring and not mark it as a TODO.

model/src/payout_curve/quanto.rs
```diff
@@ -369,8 +407,12 @@ pub fn bankruptcy_price_short(initial_price: u64, leverage: Leverage) -> u64 {
...
 /// This is an absolute value. How to apply it in order to calculate the bankruptcy price will
 /// depend on the party's position.
-fn bankruptcy_price_shift(initial_price: u64, leverage: Leverage) -> u64 {
-    initial_price / leverage.0 as u64
+/// Note: the resulting bankruptcy price is rounded to 0 decimal points.
+fn bankruptcy_price_shift(initial_price: u64, leverage: Decimal) -> u64 {
+    (Decimal::from(initial_price) / leverage)
+        .round_dp_with_strategy(0, RoundingStrategy::ToZero)
+        .to_u64()
+        .expect("Bankruptcy price to fit into u64")
```
luckysori, line 415: 🙃 I think it's clearer if we convert everything to `Decimal` first, then have the formula in a separate line, and finally convert the result into the return type. That's what I've tried to do in this file.

model/src/payout_curve/quanto.rs
```diff
@@ -369,8 +407,12 @@ pub fn bankruptcy_price_short(initial_price: u64, leverage: Leverage) -> u64 {
 ///
 /// This is an absolute value. How to apply it in order to calculate the bankruptcy price will
 /// depend on the party's position.
-fn bankruptcy_price_shift(initial_price: u64, leverage: Leverage) -> u64 {
-    initial_price / leverage.0 as u64
+/// Note: the resulting bankruptcy price is rounded to 0 decimal points.
```
luckysori, line 410: 🔧 I think this line is unnecessary because the return type doesn't support decimal points.

model/src/payout_curve/quanto.rs
```diff
@@ -230,19 +231,56 @@ impl Curve {
...
+    /// formula to compute the initial margin in the first place: `margin=(n_contracts *
+    /// initial_price * multiplier) / leverage` --> `effective_leverage=(n_contracts *
+    /// initial_price * multiplier) / margin`
+    fn effective_leverage(&self, margin: Amount) -> Result<Decimal> {
+        let effective_leverage =
+            (Decimal::from(self.n_contracts) * Decimal::from(self.initial_price) * self.multiplier)
+                / Decimal::from_f64(margin.as_btc())
+                    .context("Could not convert initial margin to decimal")?;
+        Ok(effective_leverage)
+    }
```
luckysori, line 282: 🔧 This is just a generic leverage formula. It just so happens that you pass in the effective margin wherever you call it, but that's circumstantial.

I think we should have this method should therefore be a private free function.

model/src/payout_curve/quanto.rs
```diff
@@ -230,19 +231,56 @@ impl Curve {
...
+    /// margin (reduced or increased due to fee account). The formula below was derived from the
+    /// formula to compute the initial margin in the first place: `margin=(n_contracts *
+    /// initial_price * multiplier) / leverage` --> `effective_leverage=(n_contracts *
+    /// initial_price * multiplier) / margin`
+    fn effective_leverage(&self, margin: Amount) -> Result<Decimal> {
+        let effective_leverage =
+            (Decimal::from(self.n_contracts) * Decimal::from(self.initial_price) * self.multiplier)
+                / Decimal::from_f64(margin.as_btc())
+                    .context("Could not convert initial margin to decimal")?;
+        Ok(effective_leverage)
```
luckysori, line 281: 🙃 If you put the formula on a separate line from all the conversions you probably don't need to add it to the documentation.

luckysori (approved):
It looks like it works, but I think we can refactor things slightly.

PR #2777, opened 2022-08-29
feat(daemon-tests): Use real-world values in ETHUSD tests

klochowicz:
Use a real-world Olivia event and ETHUSD price (1500USD).

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -248,10 +248,25 @@ pub struct OpenCfdArgs {
     pub oracle_data: OliviaData,
 }
 
+fn initial_price_for(symbol: ContractSymbol) -> Price {
+    Price::new(match symbol {
+        ContractSymbol::BtcUsd => dummy_btc_price(),
+        ContractSymbol::EthUsd => dummy_eth_price(),
+    })
+    .unwrap()
+}
+
+/// Different contract symbols can have different lot sizes
+fn lot_size_for(symbol: ContractSymbol) -> LotSize {
+    match symbol {
+        ContractSymbol::BtcUsd => LotSize::new(100),
+        ContractSymbol::EthUsd => LotSize::new(1),
+    }
+}
+
 impl OpenCfdArgs {
     fn offer_params(&self) -> OfferParams {
-        OfferParamsBuilder::new()
-            .contract_symbol(self.contract_symbol)
+        OfferParamsBuilder::new(self.contract_symbol)
             .price(self.initial_price)
             .build()
     }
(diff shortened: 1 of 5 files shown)


daemon-tests/src/lib.rs
```diff
@@ -276,8 +289,8 @@ impl Default for OpenCfdArgs {
     fn default() -> Self {
         Self {
             contract_symbol: ContractSymbol::BtcUsd,
+            initial_price: initial_price_for(ContractSymbol::BtcUsd),
```
luckysori: 🙃 Could extract the `ContractSymbol` into a variable.

daemon-tests/src/lib.rs
```diff
@@ -1048,8 +1060,8 @@ impl OfferParamsBuilder {
             funding_rate_short: FundingRate::new(dec!(0.00024)).unwrap(),
             opening_fee: OpeningFee::new(Amount::from_sat(2)),
             leverage_choices: vec![Leverage::TWO],
-            contract_symbol: ContractSymbol::BtcUsd,
-            lot_size: LotSize::new(100),
+            contract_symbol: symbol,
```
luckysori, line 1066: 🙃 A little simpler.

```suggestion
            contract_symbol,
```

daemon-tests/tests/main/order.rs
```diff
@@ -23,12 +23,11 @@ async fn taker_places_order_and_maker_rejects() {
 
     ensure_null_next_offers(taker.offers_feed()).await.unwrap();
 
+    let symbol = ContractSymbol::BtcUsd;
     maker
-        .set_offer_params(OfferParamsBuilder::new().build())
+        .set_offer_params(OfferParamsBuilder::new(symbol).build())
```
luckysori, line 28: 🙃 It's a bit confusing to use different names for this.

```suggestion
        .set_offer_params(OfferParamsBuilder::new(contract_symbol).build())
```

luckysori, line 28: ❓ Why can't we use a default value here? This test doesn't care about the contract symbol, so it's just noise.

luckysori: We could just default to `BtcUsd` in the implementation of `Default`.

luckysori (approved):
LGTM, just some minor suggestions!

daemon-tests/src/lib.rs
klochowicz: indeed!

daemon-tests/tests/main/order.rs
klochowicz: IMHO it should just be `symbol` all the way 😄

klochowicz: we need to accurately mock the oracle announcements, and I find it much less surprising to set it up consistently for all the variables. 

Otherwise, we're relying on the intimate knowledge that `Default` is setting up BtcUsd, because one has to explicitly configure the announcements anyway.

luckysori: Fair, I forgot that we had to configure the mock below.

holzeis (approved):

klochowicz:
@luckysori thanks for looking into why the test was failing, I've force-pushed a fix (the `OfferParamsBuilder` uses `initial_price` now).

Issue #2769, opened 2022-08-29
Rollover failed: Expect event to exist in msg


bonomat:
Just saw this locally.

```
2022-08-29 06:28:11  INFO xtra_libp2p_rollover::current::taker: Rollover proposal got accepted order_id=3712a56e-9e0d-4662-a9e3-3f47f3f5af6a
2022-08-29 06:28:13  WARN fallible task handle_error{err=Expect event to exist in msg}: model::cfd: Rollover failed: Expect event to exist in msg order_id=3712a56e-9e0d-4662-a9e3-3f47f3f5af6a peer_id=Some(12D3KooWEsK2X8Tp24XtyWh7DM65VfwXtNH2cmfs2JsWmkmwKbV1)
2022-08-29 06:28:13  INFO fallible task handle_error{err=Expect event to exist in msg}:xtra_actor_request{actor_type=daemon::process_manager::Actor message_type=daemon::process_manager::Event}:xtra_message_handler:Handle Event: sqlite_db: Appended event to database event=RolloverFailed order_id=3712a56e-9e0d-4662-a9e3-3f47f3f5af6a
```

luckysori:
This is caused by one party talking about `BETH` events and the other about `BXBT`. I'm not convinced this is caused by the weirdness we introduced by temporarily serialising `BETH` events to `BXBT`. So I think this is probably a bug.

luckysori:
Also, I was able to intermittently reproduce this locally during contract setup. Perhaps that's a good way of trying to reproduce and debug this bug.

PR #2778, opened 2022-08-29
Serialize oracle events according to index price

luckysori:
Could be related to #2769.

In a484929f7c0d1b57697fac3ebbe3febb48d9b09c, I correctly adapted the `FromStr` implementation of `sqlite_db::models::BitMexPriceEventId`, but not the `Display` implementation!

This could lead to bugs where `BETH` events saved to the database would be loaded as `BXBT` events.


klochowicz (approved):
well spotted!

da-kami (approved):

Issue #2773, opened 2022-08-29
`Short` and `Long` button hidden when version disclaimer is displayed


da-kami:
![image](https://user-images.githubusercontent.com/5557790/187164690-24c7f6cf-3f67-4dde-86e9-263c45eab805.png)

This is not critical but not the best UX :)

PR #2779, opened 2022-08-29
fix: hidden short and long button

holzeis:
Added a conditional margin top in case a disclaimer is shown

* incompatible maker version
* outdated taker version

issue2773

It's not the nicest solution but I guess it fits the current design 😄.

### DIFF
--- a/taker-frontend/src/components/MainPageLayout.tsx
+++ b/taker-frontend/src/components/MainPageLayout.tsx
@@ -61,7 +61,10 @@ export function MainPageLayout(
                                 textAlign="center"
                                 bg={useColorModeValue(BG_LIGHT, BG_DARK)}
                                 maxWidth={VIEWPORT_WIDTH_PX}
-                                marginTop={`${HEADER_HEIGHT}px`}
+                                marginTop={(HEADER_HEIGHT
+                                    + (outdatedWarningIsVisible || incompatibleWarningIsVisible
+                                        ? HEADER_HEIGHT
+                                        : 0)) + "px"}
                                 minHeight={`calc(100vh - ${FOOTER_HEIGHT}px - ${HEADER_HEIGHT}px)`}
                                 width={"100%"}
                             >


holzeis:
resolves #2773

taker-frontend/src/components/MainPageLayout.tsx
```diff
@@ -61,7 +61,8 @@ export function MainPageLayout(
                                 textAlign="center"
                                 bg={useColorModeValue(BG_LIGHT, BG_DARK)}
                                 maxWidth={VIEWPORT_WIDTH_PX}
-                                marginTop={`${HEADER_HEIGHT}px`}
+                                marginTop={(HEADER_HEIGHT
+                                    + (outdatedWarningIsVisible || incompatibleWarningIsVisible ? 50 : 0)) + "px"}
```
klochowicz: we can avoid the magic number, otherwise if we change the header height const this would get out of sync 

```suggestion
                                    + (outdatedWarningIsVisible || incompatibleWarningIsVisible ? HEADER_HEIGHT : 0)) + "px"}
```

holzeis: well, its actually not the HEADER_HEIGHT, but incidentally the height is the same.. its the space that the disclaimer takes and we do not have a constant for that (at least I haven't found one).

How about creating a new constant DISCLAIMER_HEIGHT and using that one?

klochowicz: https://github.com/itchysats/itchysats/blob/real-world-ethusd-values/taker-frontend/src/components/NavBar.tsx#L355 - it seems it is the same one?

holzeis: You are right, still its not the same height due to the padding 4.. but anyhow.. I'll change it.

bonomat (approved):
Very nice. thanks

PR #2782, opened 2022-08-30
fix (taker-ui): get months obviously starts with 0

bonomat:
but getDate starts with 1 🤦‍♂️

noticed because a position created yesterday had an expiry last month:

image

fixed: 

image


luckysori (approved):
Nice catch! Haha.

PR #2764, opened 2022-08-28
fix: long/short offers were flipped

bonomat:
resolves #2760

How did this ever work? 😅

PR #2784, opened 2022-08-30
fix (maker/taker-ui): short/long offers were flipped

### LINKED ISSUE #2774: Pressing `Long` for BTCUSD opens a `Short` position
image

I just manually opened a position by pressing `Long`, but that resulted in a `Short` position in the UI. I did it again to verify the behavior and same happened. I think something is flipped somewhere (at least for `BTCUSD`).

bonomat:
When the maker is long and provides a long price, then it's an offer to go short from the taker's perspective with the long price provided by the maker being the short price for the taker🤯.

Men, we really need to change this... everytime I'm looking at this it breaks my brain 😅

resolves #2774

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -1099,22 +1099,22 @@ impl State {
                     contract_symbol: ContractSymbol::BtcUsd,
                     position_maker: Position::Long,
                     ..
-                } => self.offers.btcusd_long = Some(new_offer),
+                } => self.offers.btcusd_short = Some(new_offer),
                 CfdOffer {
                     contract_symbol: ContractSymbol::BtcUsd,
                     position_maker: Position::Short,
                     ..
-                } => self.offers.btcusd_short = Some(new_offer),
+                } => self.offers.btcusd_long = Some(new_offer),
                 CfdOffer {
                     contract_symbol: ContractSymbol::EthUsd,
                     position_maker: Position::Long,
                     ..
-                } => self.offers.ethusd_long = Some(new_offer),
+                } => self.offers.ethusd_short = Some(new_offer),
                 CfdOffer {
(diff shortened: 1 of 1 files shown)


daemon/src/projection.rs
```diff
@@ -1099,22 +1099,22 @@ impl State {
                     contract_symbol: ContractSymbol::BtcUsd,
                     position_maker: Position::Long,
                     ..
-                } => self.offers.btcusd_long = Some(new_offer),
+                } => self.offers.btcusd_short = Some(new_offer),
```
holzeis, line 1102: 🔧 I think an explanatory comment would be helpful to explain why we switch here from long to short (and vice versa), as you mentioned it's not obvious from just looking at the code..

holzeis (approved):

da-kami:
Note that https://github.com/itchysats/itchysats/pull/2764/files was always flipped like this. We are moving this flip into the projection now. We have to be careful that this is then correct for the consumers. 
Prior to https://github.com/itchysats/itchysats/pull/2764/files the projection API always exposed the position of the *maker* and the taker UI flipped this. 
I am happy to change this to the API always exposing one's "own" position, but we have to ensure that this is correctly consumed by all consumers. I think it is, because only the taker UI needed the flip and the own position should represent that correctly.

bonomat:
Thanks for your comment @da-kami : this really makes me want to change the whole model: https://github.com/itchysats/itchysats/issues/2786

As a reference: https://github.com/itchysats/itchysats/pull/2764/ was done because the values were flipped in the UI.  Might have been a wrong fix to the same problem we are facing here.

bonomat:
closing this as this problem was proably introduced in https://github.com/itchysats/itchysats/pull/2764

PR #2798, opened 2022-08-30
Restore `offer` protocol compatibility with `taker:0.5.5`

### LINKED ISSUE #2791: Order protocol incompatible between `maker:master` and `taker:0.5.5`
Reproduced locally with this error:

```
Failed to receive order from taker: Unable to decode order: JSON error: missing field `contract_symbol` at line 1 column 467: missing field `contract_symbol` at line 1 column 467
```

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2791.


xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -75,9 +75,11 @@ impl Actor {
         let quiet = quiet_spans::sometimes_quiet_children();
         for peer_id in self.connected_peers.iter().copied() {
             self.send_offers(peer_id, msg.0.clone(), ctx)
-                .instrument(quiet.in_scope(|| {
```
holzeis, line 78: ❓ I just starting to read through the tracing library and this might be a silly question, but why do we instrument here explicitly instead of simple annotating the send offers method?

holzeis (approved):

daemon/src/order/maker.rs
```diff
@@ -134,13 +134,13 @@ impl Actor {
             }
         };
 
-        let (order_id, offer, quantity, leverage) = match order {
+        let (order_id, offer_id, quantity, leverage) = match order {
```
da-kami, line 137: 🤔 Sorry, I cannot follow why this file was changes in this way.

I had a look at https://github.com/itchysats/itchysats/commit/e4f910f35070c47672c8e37e224b51e3b532e833#diff-4826e36fe738978b5d804d1cd212d9060a660e636c4c4a9552621a23ce260843 and there are changes to `daemon/src/order/maker.rs` but not the changes you are "reverting" here.

da-kami (commented):
Sorry, I am a bit confused by the commit message.

The linked commit https://github.com/itchysats/itchysats/commit/e4f910f35070c47672c8e37e224b51e3b532e833 does not include any changes to `daemon/src/order/protocol.rs` and the changes to `daemon/src/order/maker.rs` are also different.
Did you want to link a different commit?

da-kami:
> Sorry, I am a bit confused by the commit message.
> 
> The linked commit [e4f910f](https://github.com/itchysats/itchysats/commit/e4f910f35070c47672c8e37e224b51e3b532e833) does not include any changes to `daemon/src/order/protocol.rs` and the changes to `daemon/src/order/maker.rs` are also different. Did you want to link a different commit?

Probably this one? https://github.com/itchysats/itchysats/commit/a484929f7c0d1b57697fac3ebbe3febb48d9b09c#diff-148a9d02b2823290bac9a7b07f81a66660db3e040ca1fc2b40ba4f84081b90c0

luckysori:
> Sorry, I am a bit confused by the commit message.
> 
> 
> 
> The linked commit https://github.com/itchysats/itchysats/commit/e4f910f35070c47672c8e37e224b51e3b532e833 does not include any changes to `daemon/src/order/protocol.rs` and the changes to `daemon/src/order/maker.rs` are also different.
> 
> Did you want to link a different commit?

Nope, I linked the right commit.

Quoting my commit:

> This fixes an unintended breaking change introduced by
e4f910f, which modified the
`model::Offer` type in several ways. At the time we assumed that we
were no longer sending said type over the wire, but that was wrong.

Patch e4f910f modified the `model::Offer` (its (de)serialisation) meaning that it unintentionally introduced breaking changes to the `order` protocol, which used it as part of the wire message before this PR.

da-kami:
I tested this with running the `maker` locally and a `0.5.5` taker container against it and it fixes the problem.
We do run into a different problem, but I recorded that in a separate ticket: https://github.com/itchysats/itchysats/issues/2801

da-kami (approved):

klochowicz (approved):
thanks!

da-kami:
bors retry

xtra-libp2p-offer/src/current/maker.rs
luckysori: I'm not sure either! @Restioson, what do you think?

Restioson: The main reason is because we want the span to be a child of `quiet` so that it is silenced. That being said, we might be able to achieve this through instrumenting send_offers and then using `.instrument(quiet)`. I'll open a PR to do this as I agree that it will be slightly cleaner

Restioson: See #2851

Issue #2760, opened 2022-08-27
ETHUSD: Long and short liquidation price are swapped


bonomat:
In addition to https://github.com/itchysats/itchysats/issues/2754 we should check how we compute the liquidation prices. Right now they are swapped: 

Long position shows short liquidation price and vice versa: 

<img width="677" alt="image" src="https://user-images.githubusercontent.com/224613/187026281-5a5f48b9-4b6b-4ff9-b39d-e5108a21ca3f.png">

<img width="580" alt="image" src="https://user-images.githubusercontent.com/224613/187026289-a6eb8129-01b4-46e2-9803-98e476857792.png">

da-kami:
https://github.com/itchysats/itchysats/issues/2774 should be resolved with https://github.com/itchysats/itchysats/pull/2790 
The flipped liquidation price is yet to be investigated, hence I am reopening this ticket!

da-kami:
Note: Further assertions in the actor tests - https://github.com/itchysats/itchysats/pull/2799 - currently point to this problem coming from the UI.
Once https://github.com/itchysats/itchysats/pull/2799 is finished (see https://github.com/itchysats/itchysats/issues/2802) I will have another look at this.

bonomat:
So, the problem is again pointing us to fix #2786

The issue is that we create an offer from the perspective with liquidation price of the maker.
In the frontend, we think we just flip it around but this is wrong because the liquidation price for the maker != the liquidation price for the taker.

In values expressed (other values omitted): 

These are the offers as received in the frontend: 

```
[
  {
    "contract_symbol": "EthUsd",
    "position": {
      "key": "Long"
    },
    "price": "9850.00",
    "leverage_details": [
      {
        "leverage": 2,
        "liquidation_price": "4925",
      },
      {
        "leverage": 1,
        "liquidation_price": "1",
      },
      {
        "leverage": 3,
        "liquidation_price": "6567",
      }
    ],
  },
  {
    "contract_symbol": "EthUsd",
    "position": {
      "key": "Short"
    },
    "price": "11165.00",
    "leverage_details": [
      {
        "leverage": 2,
        "liquidation_price": "16747",
      },
      {
        "leverage": 1,
        "liquidation_price": "22330",
      },
      {
        "leverage": 3,
        "liquidation_price": "14886",
      }
    ],
  }]
```

It says: 
- maker long: 
 - price: `9850`
 - liquidation price `4925`, ..
- maker short:
 - price: `11165`
[...]

PR #2799, opened 2022-08-30
feat(daemon-tests): Assert more values exposed from projection

klochowicz:
Assert on:
- closing price,
- liquidation price
- position (maker and taker should have opposite positions)

Assert for closed CFDs for all possibilities of collab settlement tests (they
test both directions and both contract symbols).

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -248,7 +248,7 @@ pub struct OpenCfdArgs {
     pub oracle_data: OliviaData,
 }
 
-fn initial_price_for(symbol: ContractSymbol) -> Price {
+pub fn initial_price_for(symbol: ContractSymbol) -> Price {
     Price::new(match symbol {
         ContractSymbol::BtcUsd => dummy_btc_price(),
         ContractSymbol::EthUsd => dummy_eth_price(),
@@ -1099,11 +1099,11 @@ fn dummy_funding_fee() -> FundingFee {
     .unwrap()
 }
 
-fn dummy_btc_price() -> Decimal {
+pub fn dummy_btc_price() -> Decimal {
     dec!(50_000)
 }
 
-fn dummy_eth_price() -> Decimal {
+pub fn dummy_eth_price() -> Decimal {
     dec!(1_500)
 }
 
@@ -1122,13 +1122,12 @@ pub async fn mock_oracle_announcements(
         .await;
 }
 
-pub async fn mock_quotes(maker: &mut Maker, taker: &mut Taker) {
+pub async fn mock_quotes(maker: &mut Maker, taker: &mut Taker, contract_symbol: ContractSymbol) {
     taker.mocks.mock_latest_quotes().await;
     maker.mocks.mock_latest_quotes().await;
     let mut quote_receiver = taker.quote_feed().clone();
 
-    let non_empty_quote =
-        |quotes: projection::LatestQuotes| quotes.get(&ContractSymbol::BtcUsd).copied();
+    let non_empty_quote = |quotes: projection::LatestQuotes| quotes.get(&contract_symbol).copied();
 
     next_with(&mut quote_receiver, non_empty_quote)
         .await
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -1133,3 +1133,27 @@ pub async fn mock_quotes(maker: &mut Maker, taker: &mut Taker, contract_symbol: 
         .await
         .unwrap(); // if quote is available on feed, it propagated through the system
 }
+
+pub fn expected_taker_liquidation_price(symbol: ContractSymbol, taker_position: Position) -> Price {
+    let expected_liquidation_price = match (symbol, taker_position) {
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(33_333.333333333333333333333333),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(100_000),
+        // quanto linear payout curve
+        (ContractSymbol::EthUsd, Position::Long) => dec!(750),
+        (ContractSymbol::EthUsd, Position::Short) => dec!(2_250),
+    };
+    Price::new(expected_liquidation_price).unwrap()
+}
+
+pub fn expected_maker_liquidation_price(symbol: ContractSymbol, maker_position: Position) -> Price {
+    let expected_liquidation_price = match (symbol, maker_position) {
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(25_000),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(21_000_000), // INFINITE
+        // quanto linear payout curve
+        (ContractSymbol::EthUsd, Position::Long) => dec!(1),
+        (ContractSymbol::EthUsd, Position::Short) => dec!(3_000),
+    };
+    Price::new(expected_liquidation_price).unwrap()
+}
--- a/daemon-tests/tests/main/collaborative_settlement.rs
+++ b/daemon-tests/tests/main/collaborative_settlement.rs
@@ -2,6 +2,8 @@ use daemon::projection::CfdState;
 use daemon_tests::confirm;
 use daemon_tests::dummy_btc_price;
 use daemon_tests::dummy_eth_price;
+use daemon_tests::expected_maker_liquidation_price;
+use daemon_tests::expected_taker_liquidation_price;
 use daemon_tests::flow::next_with;
 use daemon_tests::flow::one_cfd_with_state;
 use daemon_tests::initial_price_for;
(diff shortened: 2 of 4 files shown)


klochowicz:
thank you @da-kami for finding another sneaky hard-coded part in tests ❤️

klochowicz:
@da-kami I added one more commit that adds some checks to the offer. This is likely where the problem can be pinned down.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+fn assert_maker_liquidation_price(maker: &mut Maker) {
+    assert_eq!(dummy_btc_price(), dec!(50_000), "precondition");
+    assert_eq!(dummy_eth_price(), dec!(1_500), "precondition");
+
+    let maker_position = maker.first_cfd().position;
+    let symbol = maker.first_cfd().contract_symbol;
+    let expected_liquidation_price = match (symbol, maker_position) {
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(25_000),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(21_000_000), // INFINITE
```
luckysori: ❓ Don't we have a constant for the arbitrary infinite value we chose? It might be better to use that if possible.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+        initial_price_for(maker.first_cfd().contract_symbol)
+    );
+
+    assert_maker_liquidation_price(maker);
+    assert_taker_liquidation_price(taker);
+}
+
+fn assert_maker_liquidation_price(maker: &mut Maker) {
+    assert_eq!(dummy_btc_price(), dec!(50_000), "precondition");
+    assert_eq!(dummy_eth_price(), dec!(1_500), "precondition");
```
luckysori, line 155: 👍

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+
+fn assert_taker_liquidation_price(taker: &mut Taker) {
+    assert_eq!(dummy_btc_price(), dec!(50_000), "precondition");
+    assert_eq!(dummy_eth_price(), dec!(1_500), "precondition");
+
+    let taker_position = taker.first_cfd().position;
+    let symbol = taker.first_cfd().contract_symbol;
+    let expected_liquidation_price = match (symbol, taker_position) {
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(33333.333333333333333333333333),
```
luckysori: 🙃 Should be `33_333.333...` to be consistent.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+    assert_eq!(
+        maker.first_cfd().contract_symbol,
+        taker.first_cfd().contract_symbol
+    );
+    assert_ne!(maker.first_cfd().position, taker.first_cfd().position);
+
+    assert_eq!(
+        maker.first_cfd().closing_price.unwrap(),
+        initial_price_for(maker.first_cfd().contract_symbol)
+    );
```
luckysori: ❓ Why is this true? How come the _closing_ price is expected to be equal to the _initial_ price.

Even if it holds true, it's highly unintuitive to me.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -44,7 +44,7 @@ async fn collaboratively_close_an_open_eth_usd_cfd_maker_going_long() {
 async fn maker_rejects_collab_settlement_after_commit_finality() {
     let (mut maker, mut taker) = start_both().await;
     let order_id = open_cfd(&mut taker, &mut maker, OpenCfdArgs::default()).await;
-    mock_quotes(&mut maker, &mut taker).await;
+    mock_quotes(&mut maker, &mut taker, ContractSymbol::BtcUsd).await;
```
luckysori: 🔧 I think we should extract this from the `OpenCfdArgs` generated on line 46. Otherwise these values can diverge.

luckysori: The same applies on line 68.

daemon-tests/src/lib.rs
```diff
@@ -1133,3 +1133,27 @@ pub async fn mock_quotes(maker: &mut Maker, taker: &mut Taker, contract_symbol:
...
+    let expected_liquidation_price = match (symbol, maker_position) {
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(25_000),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(21_000_000), // INFINITE
+        // quanto linear payout curve
+        (ContractSymbol::EthUsd, Position::Long) => dec!(1),
+        (ContractSymbol::EthUsd, Position::Short) => dec!(3_000),
+    };
+    Price::new(expected_liquidation_price).unwrap()
+}
```
luckysori: 🙃 I suppose this refactor could be in the first commit.

luckysori (approved):
Great stuff!

Somehow I was expecting us to add similar assertions to contract setup (and even rollover). Is that not possible?

Also, I think it would be good to deal with the `OpenCfdArgs`  properly before merging this. We seem to be misusing it given the `Default` implementation that was originally introduced.

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(25_000),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(21_000_000), // INFINITE
+        // quanto linear payout curve
+        (ContractSymbol::EthUsd, Position::Long) => dec!(1),
+        (ContractSymbol::EthUsd, Position::Short) => dec!(3_000),
+    };
+    assert_eq!(
+        maker.first_cfd().liquidation_price,
+        Price::new(expected_liquidation_price).unwrap()
+    );
```
bonomat, line 158: There is your bug: the problem is that the offer is from the point of view of the maker but in the frontend we need to know the liquidation price for the taker. See https://github.com/itchysats/itchysats/issues/2760#issuecomment-1232383910

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -113,4 +120,66 @@ async fn collaboratively_close_an_open_cfd(
...
+        // inverse payout curve
+        (ContractSymbol::BtcUsd, Position::Long) => dec!(25_000),
+        (ContractSymbol::BtcUsd, Position::Short) => dec!(21_000_000), // INFINITE
+        // quanto linear payout curve
+        (ContractSymbol::EthUsd, Position::Long) => dec!(1),
+        (ContractSymbol::EthUsd, Position::Short) => dec!(3_000),
+    };
+    assert_eq!(
+        maker.first_cfd().liquidation_price,
+        Price::new(expected_liquidation_price).unwrap()
```
bonomat: We might also want to assert the `funding_rate`.

bonomat (approved):
LGTM

daemon-tests/tests/main/collaborative_settlement.rs
klochowicz: so far we use a fairly primitive mock for quotes - we use the same dummy price for both bid and ask prices, both when you create offers, and the "current quote" when you try to settle them. then, taking the average of that, we still get the same price...

I will add a comment to make this more obvious.

daemon-tests/src/lib.rs
klochowicz: it only became reusable in last commit (when I used it in offers too).

daemon-tests/tests/main/collaborative_settlement.rs
klochowicz: it's not as easy, as we don't have the funding rate exposed in the Cfd (it's in the offer), and the fee account is not public.

daemon-tests/tests/main/collaborative_settlement.rs
klochowicz: I've exposed it and used here.

daemon-tests/src/lib.rs
```diff
@@ -1122,15 +1122,38 @@ pub async fn mock_oracle_announcements(
...
```
luckysori: 🤔 It's quite unexpected to find these in a common place for all tests. These are functions where we return predefined expected values that will only ever apply if tests have been set up in a very particular way.

As much as I can understand why you'd want to reuse this, I think these should just be hard-coded in the individual tests.

PR #2800, opened 2022-08-30
fix: copy and paste error

holzeis:
While going through the tests I found these replicated match clauses. This looks like a copy and paste error to me. I think one clause should be short respectively.


klochowicz (approved):
thanks, this indeed seems like a copy-paste error!

PR #2807, opened 2022-08-31
fix (model): funding rate > 1

### LINKED ISSUE #2806: Actor panic!
Experienced locally

```
thread 'rocket-worker-thread' panicked at 'Self conversion from one funding rate to the other should work: Funding rate can't be higher than 100%', sqlite-db/src/models.rs:436:14
2022-08-31 01:42:23  INFO xtras::supervisor: Actor panicked actor=daemon::projection::Actor reason=unknown restart=true
```

bonomat:
There is no reason why a funding rate can't be greater than >100%

resolves #2806

### DIFF
--- a/model/src/lib.rs
+++ b/model/src/lib.rs
@@ -492,11 +492,6 @@ pub struct FundingRate(Decimal);
 
 impl FundingRate {
     pub fn new(rate: Decimal) -> Result<Self> {
-        ensure!(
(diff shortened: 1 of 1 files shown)


luckysori (approved):
😄

bonomat:
Note on the bug: this was a mean one because when deserializing the funding rate through the http API this constructor is not called, meaning, it will only fail when you load your position again from the database.

PR #2808, opened 2022-08-31
Refactors after #2772

luckysori:
This is me tackling my own suggestions from this old review: https://github.com/itchysats/itchysats/pull/2772#pullrequestreview-1089983538.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -1860,11 +1860,6 @@ pub fn calculate_long_liquidation_price(
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
 
-            // TODO: this liquidation price is potentially not in-line with the _actual_
-            // liquidation price as computed in the payout curve. The problem is we do not have the
-            // fee account available at this place and always compute the liquidation
-            // price based on the initial margin.
-            // Same for short liquidation price a few lines below.
             let liquidation_price =
                 quanto::bankruptcy_price_long(initial_price, leverage.as_decimal());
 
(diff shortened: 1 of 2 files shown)


model/src/cfd.rs
```diff
@@ -1848,6 +1848,9 @@ pub fn calculate_profit(payout: Amount, margin: Amount) -> (SignedAmount, Percen
 }
 
 /// Compute the liquidation price for the party going long.
+///
+/// The formulas used do not take into account the fees that can flow between the two parties. This
```
bonomat: 🙃 Not sure if the first sentence adds much value. I'd rather formulate in a way saying _what it does_, e.g. the second sentence.

model/src/cfd.rs
```diff
@@ -1860,11 +1863,6 @@ pub fn calculate_long_liquidation_price(
         ContractSymbol::EthUsd => {
             let initial_price = initial_price.to_u64();
 
-            // TODO: this liquidation price is potentially not in-line with the _actual_
```
bonomat, line 1863: Did you create a ticket for that?

bonomat (approved):
Looks good. Thanks for taking over.

model/src/cfd.rs
luckysori: Good reminder, I'll do it.

luckysori: Here it is: https://github.com/itchysats/itchysats/issues/2809.

model/src/cfd.rs
luckysori: Actually, the added paragraph is kind of wrong because it assumes that the leverage passed cannot be the effective leverage, which is not true. We don't (cannot) enforce that the `leverage` argument is the initial leverage!

Issue #2786, opened 2022-08-30
Offer fields are ambiguous


bonomat:
Personally, I find our offers model a bit confusing (see https://github.com/itchysats/itchysats/pull/2784) 

When creating an offer, the maker needs to set two prices: 
https://github.com/itchysats/itchysats/blob/1bb5dd1e8bd1d45a5e159311c18c5ab099e847a5/maker/src/routes.rs#L147-L148

However, the prices here are not for the taker but actually the maker, i.e. if he says he wants to have a price of 
- `$100` for going long
- `$110` for going short

the `going long` and `going short` part is actually related to him and not to the taker who will eventually consume this. 
The taker needs to flip it before hand: 

https://github.com/itchysats/itchysats/blob/d1564dacfe6f421cd1feed1a14d95f2534375461/daemon/src/projection.rs#L1097-L1118

Imho, the maker should create an offer which is ready to be consumed by the taker, i.e. if he says
- `$110` for going long
- `$100` for going short

it means this is what the taker needs to pay when `going long` or `going short` respectively.

bonomat:
This project is unmaintained

PR #2810, opened 2022-08-31
fix: liquidation details are flipped

### LINKED ISSUE #2760: ETHUSD: Long and short liquidation price are swapped

bonomat:
The problem is that the offer is from the perspective of the maker but in the frontend we need to show it from the perspective of the taker, hence we need to take the leverage details of the short offer and put it into the one from the long offer and vice versa.

resolves #2760

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -135,10 +135,10 @@ export const App = () => {
 
     const identityOrUndefined = useLatestEvent<IdentityInfo>(source, "identity");
 
-    const shortOffer = makerOfferToTakerOffer(makerLong);
-    const longOffer = makerOfferToTakerOffer(makerShort);
+    const shortOffer = makerOfferToTakerOffer(makerLong, makerShort?.leverage_details);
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Seems fine.

Is the `projection::CfdOffer` supposed to be from the perspective of the maker though? It's not clearly documented anywhere as far as I can tell.

Because if it is meant to be from the perspective of the `daemon`'s owner we should flip in `projection` instead.

bonomat:
> Is the projection::CfdOffer supposed to be from the perspective of the maker though? It's not clearly documented anywhere as far as I can tell.

That's what @da-kami said in https://github.com/itchysats/itchysats/pull/2784#issuecomment-1231183640. I find it confusing and propose to fix it with #2786

bonomat:
only the role was flipped: https://github.com/itchysats/itchysats/pull/2816

PR #2812, opened 2022-08-31
Use actual liquidation price for open CFDs

### LINKED ISSUE #2809: Liquidation price in `projection` does not match actual liquidation price
This is the case for both inverse and quanto payout curves, but the reasons are different:

- For the inverse payout curve, we use a different formula altogether. It's probably unfeasible to reuse any of the code inside of the payout curve implementation unless we rewrite it. Perhaps we should derive the liquidation prices from the the discretised payouts we compute during contract setup and rollover.
- For the quanto payout curve, we use the same formula, but with different values. In the payout curve implementation we correctly use the _effective_ leverage (which varies as the `FeeAccount` changes). In `projection` we only ever use the _initial_ leverage. This could perhaps also be solved by using the values from the discretised payouts. Alternatively, we could calculate the effective leverage before calling the `calculate_{long, short}_liquidation_price` APIs.

luckysori:
I propose to do something like this to deal with https://github.com/itchysats/itchysats/issues/2809. Eventually we may just want to store this information directly, particularly for closed and failed CFDs.

---

We can deduce the liquidation prices for open CFDs by looking at the stored CETs, as soon as we have them (after we apply `{ContractSetup, Rollover}Completed` events.

This is only possible for open CFDs, because they are the only ones that have access to this information (unlike offers, closed CFDs and failed CFDs). Having correct liquidation prices also matters a lot more for open CFDs, as the user can actually use this information to make decisions.

### DIFF
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -151,7 +151,7 @@ pub struct Cfd {
     pub contract_symbol: ContractSymbol,
     pub position: Position,
     #[serde(with = "round_to_two_dp")]
-    pub liquidation_price: Price,
+    pub liquidation_price: Decimal,
 
     #[serde(with = "round_to_two_dp")]
     pub quantity: Contracts,
@@ -1324,7 +1324,7 @@ pub struct LeverageDetails {
     pub leverage: Leverage,
     /// Own liquidation price according to position and leverage
     #[serde(with = "round_to_two_dp")]
-    pub liquidation_price: Price,
+    pub liquidation_price: Decimal,
     /// Margin per lot from the perspective of the role
     ///
     /// Since this is a calculated value that we need in the UI this value is based on the
--- a/daemon/src/projection.rs
+++ b/daemon/src/projection.rs
@@ -350,6 +350,8 @@ impl Cfd {
             counterparty_leverage,
         );
 
+        // This value will be updated as we apply `{ContractSetup,Rollover}Completed` events to the
+        // `Cfd`
         let liquidation_price = match position {
             Position::Long => {
                 calculate_long_liquidation_price(initial_price, our_leverage, contract_symbol)
@@ -428,6 +430,12 @@ impl Cfd {
             }
             ContractSetupCompleted { dlc } => {
                 self.expiry_timestamp = dlc.as_ref().map(|dlc| dlc.settlement_event_id.timestamp());
+
+                if let Some(dlc) = &dlc {
+                    self.liquidation_price =
+                        Decimal::from(dlc.liquidation_price(self.role, self.position));
+                }
+
                 self.aggregated.latest_dlc = dlc;
 
                 self.aggregated.state = CfdState::PendingOpen;
@@ -444,6 +452,12 @@ impl Cfd {
                 complete_fee,
             } => {
                 self.expiry_timestamp = dlc.as_ref().map(|dlc| dlc.settlement_event_id.timestamp());
+
+                if let Some(dlc) = &dlc {
+                    self.liquidation_price =
+                        Decimal::from(dlc.liquidation_price(self.role, self.position));
+                }
+
                 self.aggregated.latest_dlc = dlc;
 
                 self.aggregated.fee_account = match complete_fee {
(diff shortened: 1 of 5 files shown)


daemon/src/projection.rs
```diff
@@ -444,6 +458,18 @@ impl Cfd {
                 complete_fee,
             } => {
                 self.expiry_timestamp = dlc.as_ref().map(|dlc| dlc.settlement_event_id.timestamp());
+
+                if let Some(dlc) = &dlc {
+                    let liquidation_price = match (self.role, self.position) {
+                        (Role::Maker, Position::Long) => *dlc.maker_liquidation_interval().end(),
```
bonomat: Can you explain how `.end` , `.start` work in combination with the function used inside of `maker_liquidation_interval`: `.min_by(|x, y|` 
I'm not sure I get it. 

My understanding is that `.min_by(|x, y|` returns you the smallest value in the list. This will be a range, e.g. `0..10`, i.e. the values where `maker` receives `0`. For getting the `long` liquidation price you can then do `end()`, i.e. you get `10` but for his `short` liquidation price you would need to get `.max_by..start` no?

luckysori: Here's my explanation with inline comments:

```rust
// cfd.rs

pub fn maker_liquidation_interval(&self) -> RangeInclusive<u64> {
    // Find CETs at settlement event. We use the settlement event
    // because all CFDs are guaranteed to have a settlement event, but
    // they might not have any liquidation events
    let (_, settlement_cets) = self
        .cets
        .iter()
        .find(|(id, _)| id == &&self.settlement_event_id)
        .expect("to have settlement CETs");

    // We search for the settlement CET that gives the maker the
    // least. That one must be the one that liquidates the maker and
    // will allow us to see the maker's liquidation price interval
    let liquidation_cet = settlement_cets
        .iter()
        .min_by(|x, y| x.maker_amount.cmp(&y.maker_amount))
        .expect("to have cets");

    liquidation_cet.range.clone()
}

// projection.rs

// Let's consider the maker side only, because the taker is equivalent
let liquidation_price = match (self.role, self.position) {
    // Long liquidation intervals go from 0-`end`, therefore `end`
    (Role::Maker, Position::Long) => *dlc.maker_liquidation_interval().end(),
    // Long liquidation intervals go from `start`-`max`, therefore `start`
    (Role::Maker, Position::Short) => *dlc.maker_liquidation_interval().start(),
    // ...
};
```

And here I'll respond to your message:

> My understanding is that .min_by(|x, y| returns you the smallest value in the list.

Continuing with the maker side: when we use `min_by` we are looking for the `Cet` that has the lowest `maker_amount`. That is, the CET that gives the the maker the lowest payout for a given oracle event. It's important to distinguish that from the CET that has the lowest _price_ in an interval!

> This will be a range, e.g. 0..10, i.e. the values where maker receives 0. 

We do return a range, but we don't know for sure that the payout will be zero, which is why we use `min_by`.

> For getting the long liquidation price you can then do end(), i.e. you get 10

I think we agree up to this point.

> but for his short liquidation price you would need to get .max_by..start no?

Nope, because we are ordering the CETs by the _payout amount_. The method `min_by` simply identifies the CET that gives the maker the least amount of money. And then if we know that the maker is going short we know that the (short) liquidation price interval will be of the form `start..=max`, where `start` will be the liquidation price.

luckysori: I will try to make it clearer, because it can be confusing.

luckysori: What do you think about https://github.com/itchysats/itchysats/pull/2812/commits/ded16b5038d5a887f0aff0d60dd7e54a0b64d570 now?

bonomat: > Nope, because we are ordering the CETs by the payout amount. The method min_by simply identifies the CET that gives the maker the least amount of money. And then if we know that the maker is going short we know that the (short) liquidation price interval will be of the form start..=max, where start will be the liquidation price.

Ah, that makes sense. https://github.com/itchysats/itchysats/commit/ded16b5038d5a887f0aff0d60dd7e54a0b64d570 is very clean. Thanks

bonomat (approved):
Nice, I think with this we will also get _dynamic_ liquidation prices as the payout curve changes with every rollover.

daemon/src/projection.rs
```diff
@@ -151,7 +151,7 @@ pub struct Cfd {
     pub contract_symbol: ContractSymbol,
     pub position: Position,
     #[serde(with = "round_to_two_dp")]
-    pub liquidation_price: Price,
+    pub liquidation_price: Decimal,
```
klochowicz, line 154: It's a good step to get away from `model::Price` here as it had an arbitrary drawback that doesn't apply to presentation (lack of `0`), but I don't understand why not introduce `projection::Price` instead? stronger typing allows better type safety and flexibility for rounding etc.

klochowicz (approved):

daemon/src/projection.rs
luckysori: I can add a dedicated type, but it didn't seem very useful since it currently won't be used to uphold any invariants or go through any layers.

luckysori: I started doing it and having `projection::Price` and `model::Price` is not great. We can't easily get rid of `model::Price` in, for example, `projection::Cfd` because it's used in calculations as we `apply` events.

I could just use the `projection::Price` for liquidation prices, but I'm not sure that's worth it. Feel free to give it a go yourself if you know how you'd like to improve the situation here.

klochowicz:
@luckysori it seems that it's not our typical rollover test failure, but an assertion of liquidation price:
https://github.com/itchysats/itchysats/blob/master/daemon-tests/tests/main/collaborative_settlement.rs#L171

PR #2814, opened 2022-08-31
Do not overwrite inverse payout curve liquidation bound

### LINKED ISSUE #2571: Short liquidation events when maker goes short are unusable

luckysori:
Unfixes https://github.com/itchysats/itchysats/issues/2571 😛 

---

This reverts 2298059164d39ea744b708e6f54dcee46ab96037. Whilst this fixed https://github.com/itchysats/itchysats/issues/2571, it also meant that takers without this fix would not be able to perform contract setup and rollover with a maker with this fix.

We will very likely introduce an equivalent fix in the future, but it cannot be safely introduced without changing `/itchysats/order/1.0.0` or introducing a new version altogether.

### DIFF
--- a/model/src/payout_curve.rs
+++ b/model/src/payout_curve.rs
@@ -109,7 +109,7 @@ impl Payouts {
         n_payouts: usize,
         fee: CompleteFee,
     ) -> Result<Self> {
-        let mut payouts = payout_curve::inverse::calculate(
+        let payouts = payout_curve::inverse::calculate(
             price,
             quantity,
             leverage_long,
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Oh. Should we version `payout_curve` according to the protocol?

luckysori:
> Oh. Should we version `payout_curve` according to the protocol?

It's not just the `payout_curve` though, the entirety of `model` could be versioned. We've already had to introduce things such as https://github.com/itchysats/itchysats/blob/8c84d463d1139af3fd05d00c98df136b49997368/model/src/cfd/rollover_deprecated.rs#L22

I think this goes hand in hand with my idea of extracting protocols into crates, where each version of the crate depends on a different version of `model` (and `payout_curve`). But this introduces complexity in the form of not being able to use the same `model` types across `daemon` if you have multiple versions of the `model` crate.

Basically, it's tricky.

klochowicz (approved):

PR #2834, opened 2022-08-31
Forbid sending empty offers via `/itchysats/offer/1.0.0`

### LINKED ISSUE #2817: Compatibility to `0.5.4` and `0.5.5` - Offers are frequently getting "reset"
Behavior: The taker UI frequently does not show that we have an offer.

This might be related to sending `ETHUSD` offers to the taker or how the taker UI receives it over the feed.

luckysori:
Fix https://github.com/itchysats/itchysats/issues/2817 (definitely).

By leveraging the `nonempty::NonEmpty` type we can use the type system to ensure that we will never send empty offers to takers via `/itchysats/offer/1.0.0`.

The problem with sending a `MakerOffers` with `None` values for `long` and/or `short` is that takers will interpret this as an instruction to delete their local offer state. That is something that we no longer want to trigger from the maker's perspective.

In particular, this patch fixes a bug where any `taker:<=0.5.5` was deleting its local offer state every time the maker was trying to send ETHUSD offers via the `/itchysats/offer/2.0.0` protocol. This was precisely because we were filtering out all ETHUSD offers derived from the `OfferParams`, resulting in an empty list of offers.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2101,6 +2101,7 @@ dependencies = [
  "maia 0.2.0",
  "maia-core",
  "model",
+ "nonempty",
  "prometheus",
  "quiet-spans",
  "rocket",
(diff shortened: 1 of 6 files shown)


da-kami (approved):
Tested locally with master on this branch and `0.5.5` container. Works!

klochowicz (approved):
wow. really elegant solution 👏

da-kami:
Disabling the macos tests is about to merge, will rebase and bors again then

Issue #2555, opened 2022-07-26
[Epic] Introduce `order` abstraction and re-design the `cfd` lifecycle


da-kami:
With #2292 with introduce the abstraction of an order, but the `Cfd` lifecycle within the daemon is still odd, because we mix order and the `Cfd` together.

I opted to keep this weirdness for now and create this ticket to action a proper `order` abstraction.
Part of this model was already discussed in https://github.com/itchysats/itchysats/issues/1770.

The motivation of this ticket is to design a better model for the lifecycle of a `Cfd`. This will help with creating better boundaries for the APIs, the application flow, the data and the eventually the user experience.

## Recap, `Offer` and `Order` as introduced in #2292

An offer is created by the maker and sent to all connected takers.
An offer has a dedicated id, the `OfferId`.
One offer can be taken be taken multiple times by one/multiple takers.

In order to taker an offer a taker creates an `order`. 
The order is identified by a unique id, the `OrderId`.
The taker proposes the `OrderId` when sending the order to the maker.

## Vision for `Order` and `Cfd`:

As long as the order is not filled, we should treat the order as order.
The order is filled once "contract setup is done", only then we should create the `Cfd`. As "contract setup is done" is not clearly defined, I would opt for defining it as: Once we have a `Dlc` we create the `Cfd` and attach the `Dlc`.

This means, that the `Cfd` lifecylce starts with the negotiated `Dlc`, not with the state on the blockchain. We could also opt for starting the `Cfd` lifecycle once the transaction is in mempool, because only then we can trigger actions to close the position.
The challenge here is where to store the data and how to present it to the user.

Both maker and taker should store orders in their daemon database.
The accept / reject logic should be for the order, not the `Cfd`.
The rollover tables are already separate. 
When rolling over we could opt for creating an `order` as well, that is then filled (i.e. new Dlc upding the Cfd) or not. We can also opt for doing that in a separate iteration.
If we do so, the `Cfd` is either `open` or `closed`, there is not other state.

The user interface *should* display orders and open positions separately. In an intermediary step we can opt for keeping the current display and map the `order`  model to the `projection::Cfd` model. This might be messy though.

bonomat:
> An offer is created by the maker and sent to all connected takers.
> An offer has a dedicated id, the `OfferId`.
> One offer can be taken be taken multiple times by one/multiple takers.

One note on this: 
[...]

PR #2835, opened 2022-08-31
Stop emitting errors if we cannot auto-rollover an empty CFD

luckysori:
By empty CFD we mean a CFD without any events associated with it. This can happen if the CFD is inserted into the database, but placing the order actually fails.

These CFDs are currently not cleaned up from the database, so their presence causes error log spamming when failing to auto-rollover on the taker side. Since these error logs are false positives (there's nothing dangerous about not being able to rollover an empty CFD), we opt to suppress them.

The strategy followed is to emit a different kind of error as a reason for not being able to roll over a CFD: a `CannotRollover::NoEvents` error. Since the `auto_rollover::Actor` only logs errors about `CannotRollover::NoDlc`, we no longer see error logs associated with not being to auto-rollover empty CFDs.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -278,6 +278,8 @@ pub enum CannotRollover {
     InCollaborativeSettlement,
     #[error("Cannot roll over when CFD is already closed")]
     Closed,
+    #[error("Cannot rollover CFD without events")]
+    NoEvents,
 }
 
 /// Reasons why we cannot collab close a CFD
@@ -730,6 +732,10 @@ impl Cfd {
 
         if self.is_in_force_close() {
             return Err(CannotRollover::Committed);
+        }
+
+        if self.version == 0 {
+            return Err(CannotRollover::NoEvents);
         }
 
         // Rollover and collaborative settlement are mutually exclusive, if we are currently


da-kami (approved):
Nice! that should at least stop the spam until we action https://github.com/itchysats/itchysats/issues/2555

da-kami:
Unfortunately it does not look like this was solved. The fix was already deployed on `testnet`, but we still see the [logs](
https://itchysats.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22code%22,%22expr%22:%22%7Bcontainer%3D~%5C%22taker_master%7Cmaker%7Ctaker_release%7Ctaker_release_minus_one%5C%22,setup%3D%5C%22testnet%5C%22%7D%20%7C%3D%5C%22ERROR%5C%22%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221661954121866%22,%22to%22:%221661954443510%22%7D%7D)

I think the problem might be that the `version` of these `cfd`s is actually not `0`. They might have a `ContractSetupStarted` event, but nothing else. Would have to verify.

PR #2838, opened 2022-08-31
Run CI tests in release mode

klochowicz:
It takes a bit longer to compile, but the execution should be much faster,
bringing in an overall improvement.


klochowicz:
This was an attempt to test whether the benefit of faster execution of tests would outweigh the cost of compiling tests in release mode. conclusion: it does not bring any benefits.

Issue #2824, opened 2022-08-31
Deal with panics in unsupervised parts of code


klochowicz:
A panic can lend the system unusable until the app is restarted, with the most troubling part being the system *appearing* to work, with no indication that it's severely broken (other than a panic printed in the logs). 

As all our current platforms supervise the app itself (via docker/systemd), exiting the app would start the app anew and thus would be my preferred choice.

We could also make the user do this after showing some alert, but this would lead to worse UX.

@da-kami raised a good point that we should ensure that DB will not end up in corrupted state.

klochowicz:
I have a changeset locally that just terminates the program, but that can lead to a corrupted state. I tried to achieve a graceful rocket shutdown, but failed to satisfy the borrow checker 😢 seems like this is not actionable at the moment.

da-kami:
Closing because not actionable given the current framework constraints.
I feel keeping this open does not have any value if we can't action is 😅

klochowicz:
@da-kami thank you for closing, I was meant to do the same. I'll just change it to "Not planned" instead of "Completed"

Issue #2602, opened 2022-08-01
Investigate frequent protocols failures


da-kami:
I noticed this in my taker's logs on Umbrel:

```
2022-08-01 12:11:31 DEBUG fallible task handle_error{err=The stream has terminated.}: xtra_libp2p_offer::taker: Failed to process maker offers: The stream has terminated. peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:17:08 DEBUG fallible task handle_error{err=unexpected end of file}: xtra_libp2p_ping::pong: Inbound ping protocol failed: unexpected end of file peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:17:46 DEBUG fallible task handle_error{err=The stream has terminated.}: xtra_libp2p_offer::taker: Failed to process maker offers: The stream has terminated. peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:23:39 DEBUG fallible task handle_error{err=unexpected end of file}: xtra_libp2p_ping::pong: Inbound ping protocol failed: unexpected end of file peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:25:39 DEBUG fallible task handle_error{err=The stream has terminated.}: xtra_libp2p_offer::taker: Failed to process maker offers: The stream has terminated. peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:27:42 DEBUG fallible task handle_error{err=The stream has terminated.}: xtra_libp2p_offer::taker: Failed to process maker offers: The stream has terminated. peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
2022-08-01 12:29:42 DEBUG fallible task handle_error{err=The stream has terminated.}: xtra_libp2p_offer::taker: Failed to process maker offers: The stream has terminated. peer=12D3KooWP3BN6bq9jPy8cP7Grj1QyUBfr7U6BeQFgMwfTTu12wuY
[...]

PR #2839, opened 2022-09-01
Supervise `Endpoint`s

### LINKED ISSUE #2821: Supervise `Endpoint`s
Whilst there might not be any obvious `Error`s that we should exit with and handle, we should handle panics using a supervisor.

luckysori:
Fix #2821.

### DIFF
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -321,7 +321,6 @@ where
                             ],
                             vec![
                                 dialer_actor.clone().into(),
-                                ping_actor.clone().into(),
                                 online_status_actor.clone().into(),
                                 identify_dialer_actor.clone().into(),
                             ],
--- a/daemon/src/lib.rs
+++ b/daemon/src/lib.rs
@@ -296,34 +296,45 @@ where
             Supervisor::new(move || ping::Actor::new(endpoint_addr.clone(), PING_INTERVAL));
         tasks.add(supervisor.run_log_summary());
 
-        let endpoint = Endpoint::new(
-            Box::new(TokioTcpConfig::new),
-            identity.libp2p,
-            ENDPOINT_CONNECTION_TIMEOUT,
-            TAKER_LISTEN_PROTOCOLS.inbound_substream_handlers(
-                pong_address.clone(),
-                identify_listener_actor,
-                libp2p_offer_addr,
-            ),
-            endpoint::Subscribers::new(
-                vec![
-                    online_status_actor.clone().into(),
-                    ping_actor.clone().into(),
-                    identify_dialer_actor.clone().into(),
-                ],
-                vec![
-                    dialer_actor.into(),
-                    ping_actor.into(),
-                    online_status_actor.clone().into(),
-                    identify_dialer_actor.clone().into(),
-                ],
-                vec![],
-                vec![],
-            ),
+        let endpoint_supervisor = Supervisor::new_with_context(
+            {
+                let key_pair = identity.libp2p;
+                let pong_address = pong_address.clone();
+                let online_status_actor = online_status_actor.clone();
+                let identify_dialer_actor = identify_dialer_actor.clone();
+
+                move || {
+                    Endpoint::new(
+                        Box::new(TokioTcpConfig::new),
+                        key_pair.clone(),
+                        ENDPOINT_CONNECTION_TIMEOUT,
+                        TAKER_LISTEN_PROTOCOLS.inbound_substream_handlers(
+                            pong_address.clone(),
+                            identify_listener_actor.clone(),
+                            libp2p_offer_addr.clone(),
+                        ),
+                        endpoint::Subscribers::new(
+                            vec![
+                                online_status_actor.clone().into(),
+                                ping_actor.clone().into(),
+                                identify_dialer_actor.clone().into(),
+                            ],
+                            vec![
+                                dialer_actor.clone().into(),
+                                ping_actor.clone().into(),
+                                online_status_actor.clone().into(),
+                                identify_dialer_actor.clone().into(),
+                            ],
+                            vec![],
+                            vec![],
+                        ),
+                    )
+                }
+            },
+            endpoint_context,
         );
 
-        tasks.add(endpoint_context.run(endpoint));
-
+        tasks.add(endpoint_supervisor.run_log_summary());
         tasks.add(dialer_supervisor.run_log_summary());
         tasks.add(offers_supervisor.run_log_summary());
         tasks.add(identify_listener_supervisor.run_log_summary());
--- a/xtra-libp2p-offer/src/current/maker.rs
+++ b/xtra-libp2p-offer/src/current/maker.rs
@@ -14,6 +14,7 @@ use xtra_libp2p::Endpoint;
 use xtra_libp2p::GetConnectionStats;
 use xtra_libp2p::OpenSubstream;
 use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncNext;
 
 pub struct Actor {
     endpoint: xtra::Address<Endpoint>,
@@ -50,20 +51,34 @@ impl Actor {
             anyhow::Ok(())
         };
 
-        let err_handler = move |e: anyhow::Error| async move {
-            if let Some(xtra_libp2p::Error::NegotiationFailed(
-                xtra_libp2p::NegotiationError::Failed,
-            )) = e.downcast_ref::<xtra_libp2p::Error>()
-            {
-                // It's normal to disagree on the protocols now that we broadcast on both versions
-                // to _all_ our peers
-            } else {
-                tracing::warn!(%peer_id, "Failed to send offers: {e:#}")
+        let this = ctx.address().expect("self to be alive");
+        let err_handler = {
+            let this = this.clone();
+            move |e: anyhow::Error| async move {
+                match e.downcast_ref::<xtra_libp2p::Error>() {
+                    Some(xtra_libp2p::Error::NegotiationFailed(
+                        xtra_libp2p::NegotiationError::Failed,
+                    )) => {
+                        // It's normal to disagree on the protocols now that we broadcast on both
+                        // versions to _all_ our peers
+                    }
+                    Some(xtra_libp2p::Error::NoConnection(peer_id)) => {
+                        this.send_async_next(NoConnection(*peer_id)).await;
+                    }
+                    _ => {
+                        tracing::warn!(%peer_id, "Failed to send offers: {e:#}")
+                    }
+                }
             }
         };
 
-        let this = ctx.address().expect("self to be alive");
         spawn_fallible(&this, task.instrument(span), err_handler);
+    }
+
+    fn remove_peer(&mut self, peer_id: PeerId) {
+        if self.connected_peers.remove(&peer_id) {
+            tracing::trace!(%peer_id, "Removed dropped connection");
+        }
     }
 }
 
@@ -102,9 +117,8 @@ impl Actor {
             .await;
     }
 
-    async fn handle_connection_dropped(&mut self, msg: endpoint::ConnectionDropped) {
-        tracing::trace!("Remove dropped connection: {:?}", msg.peer_id);
-        self.connected_peers.remove(&msg.peer_id);
+    async fn handle_no_connection(&mut self, msg: NoConnection) {
+        self.remove_peer(msg.0);
     }
 }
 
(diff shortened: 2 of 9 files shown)


klochowicz (commented):
I am not sure whether this is a good idea - all of the Endpoint users do not expect the endpoint to be restarted as they keep count of the connections it notifies about during runtime.
 if it gets restarted, we will catch the panic, but we will go straight into undefined behaviour territory. I'm sceptical whether we actually need the additional complexity that would supporting this properly bring with it.

Personally, the best way to deal with this would be: #2824 - there are places in the code which, when causing an error, cause such ripples, it would be easier to just start again instead of trying to restare the balance.

note: the only panic that happened ever from the Endpoint so far was locally, when I tried invalid config values.

luckysori:
> I am not sure whether this is a good idea - all of the Endpoint users do not expect the endpoint to be restarted as they keep count of the connections it notifies about during runtime. 

I've looked at the places where we get `endpoint::ConnectionEstablished` notifications and in all cases it would be perfectly normal to get said notification:

- In `identify::dialer::Actor`, we would perform the protocols again. The same thing would happen if we restarted the whole application.
- In `online_status::Actor`, we would just report that we reconnected to the maker. The same thing would happen if we restarted the whole application.
- In `offer::{deprecated, current}::Actor`, we would insert the connected peers back into the map (no duplicates) and would send them the latest offers. Effectively the same thing would happen if we restarted the whole application.
- In `ping::Actor`, we would insert the connected peers back into the map (no duplicates). Nothing would happen.

> if it gets restarted, we will catch the panic, but we will go straight into undefined behaviour territory. I'm sceptical whether we actually need the additional complexity that would supporting this properly bring with it.

The patches proposed are quite simple and certainly the behaviour described above is not undefined.
 
> Personally, the best way to deal with this would be: #2824 - there are places in the code which, when causing an error, cause such ripples, it would be easier to just start again instead of trying to restare the balance.

There might be situations where restarting would be warranted (unsure), but the `Endpoint` panicking doesn't seem like one of those to me.

da-kami:
I would have expected the endpoint to be supervised and am in favor of supervising it.
For me this PR is independent of https://github.com/itchysats/itchysats/issues/2824 which would add an additional / different layer of resilience.

Thinking about the concerns that @klochowicz pointed out:
I think one problem with restarting the endpoing would be that `offer` and `ping` do not refresh their complete connected peer state upon the endpoing restarting because they only fetch the complete connected-peer-state upon startup (and in case the endpoint is restarted they would not get restarted themselves as far as my understanding goes).
So, we may see quite a lot of cases where the offer and ping actors can't connect to a registered peer after a restart of the endpoint. This problem is not critical, because the peers are expected to re-connect (which results in a `ConnectionEstablished` event that will be handled correctly), or the connection should be eventually dropped (resulting in a `ConnectionDropped` event) because it will be regarded as "broken" (not sure our code handles that correctly for all cases in the endpoint, that is related to #2602 )

That said: I don't think supervising the endpoint should lead to undefined behavior, but there might be side effects we don't see yet.

da-kami (approved):

klochowicz:
"side effects we don't see yet" is what I think of when I talk about undefined behaviour.
Even if the current users use its API in a way that should not break, we can't have this guarantee moving forward. I see this as needless fragility that will bite us in the future.

I see the reasons for supervising Endpoint, and I believe we could come up with a way to achieve no new side-effects. How about:
- consider adding a new `EndpointRestarted` message that invalidates the counts on all the observed connections / listeners
- clear the tracked connections maps after a restart 

This way the users know that:
1) Endpoint can be restarted,
2) after a restart, one has to clear the count.

thoughts? @luckysori @da-kami

luckysori:
> "side effects we don't see yet" is what I think of when I talk about undefined behaviour. Even if the current users use its API in a way that should not break, we can't have this guarantee moving forward. I see this as needless fragility that will bite us in the future.
> 
> I see the reasons for supervising Endpoint, and I believe we could come up with a way to achieve no new side-effects. How about:
> 
>     * consider adding a new `EndpointRestarted` message that invalidates the counts on all the observed connections / listeners
> 
>     * clear the tracked connections maps after a restart
> 
> 
> This way the users know that:
> 
>     1. Endpoint can be restarted,
> 
>     2. after a restart, one has to clear the count.
> 
> 
> thoughts? @luckysori @da-kami

I think it's simpler if we just remove the peers from the map if we get an `endpoint::Error::NoConnection`. It's less communication between actors and it applies to more situations than just `Endpoint` restarts.

klochowicz:
> I think it's simpler if we just remove the peers from the map if we get an `endpoint::Error::NoConnection`. It's less communication between actors and it applies to more situations than just `Endpoint` restarts.

- isn't `NoConnection` only for particular peer id? This would rule it out from being useful for this purpose.
- there would be hardly any more communication, a panic in Endpoint should be a rare occurrence

luckysori:
>     * isn't `NoConnection` only for particular peer id? This would rule it out from being useful for this purpose.

When we try to do something with the stale information we have (open a substream), the `Endpoint` tells us that we don't actually have a connection. I think we should already be acting on this information even if we don't supervise the `Endpoint`.

>     * there would be hardly any more communication, a panic in Endpoint should be a rare occurrence

I'm not referring to the message volume. It's increasing the complexity of `Endpoint` and its consumers, particularly if we already have a mechanism in place to achieve the same goal.

daemon/src/lib.rs
```diff
@@ -296,34 +296,45 @@ where
...
-        let endpoint = Endpoint::new(
-            Box::new(TokioTcpConfig::new),
-            identity.libp2p,
-            ENDPOINT_CONNECTION_TIMEOUT,
-            TAKER_LISTEN_PROTOCOLS.inbound_substream_handlers(
-                pong_address.clone(),
-                identify_listener_actor,
-                libp2p_offer_addr,
-            ),
-            endpoint::Subscribers::new(
```
klochowicz, line 308: consider adding a new EndpointRestarted message that invalidates the counts on all the observed connections / listeners

IMHO all the subscribers should be interested in that message, as this break the guarantees given earlier by the endpoint.

klochowicz (changes requested):

daemon/src/lib.rs
luckysori: My proposal is https://github.com/itchysats/itchysats/pull/2839/commits/0d1dd1bdb5cc12e38bfcf51dfc4d9cd43a58adf8.

Restioson:
> "side effects we don't see yet" is what I think of when I talk about undefined behaviour.

Just to clarify, this is not 'undefined behaviour' in the unsoundness sense, but rather just unpredictable (but well-defined at a Rust-safety-contract level) behaviour?

klochowicz:
> > "side effects we don't see yet" is what I think of when I talk about undefined behaviour.
> 
> Just to clarify, this is not 'undefined behaviour' in the unsoundness sense, but rather just unpredictable (but well-defined at a Rust-safety-contract level) behaviour?

yes. the worst case, because no tooling can save us from this one :p

Restioson:
Kinda, although if you hit real UB you can get time travel and `if true` to fail

xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -102,9 +117,8 @@ impl Actor {
             .await;
     }
 
-    async fn handle_connection_dropped(&mut self, msg: endpoint::ConnectionDropped) {
-        tracing::trace!("Remove dropped connection: {:?}", msg.peer_id);
-        self.connected_peers.remove(&msg.peer_id);
+    async fn handle_no_connection(&mut self, msg: NoConnection) {
```
da-kami, line 120: 🤔 I am not sure I would replace one solution with the other. 
I think if the endpoint already knows that the connection was dropped we should also react to it in the protocols that want to know. 

I would add what you proposed as a resilience fallback on top of the previous solution. Otherwise it is expected to see the warning above, and I don't think that is expected behavior.

Note: Same for all other protocols where this was changed.

luckysori: > I think if the endpoint already knows that the connection was dropped we should also react to it in the protocols that want to know.

I couldn't justify to myself the need for the connection-dropped notification in any of the places where I removed the handlers. It's not like anything dangerous can happen with this stale information. As soon as the actor tries to use the information they will get a `NoConnection` error which is functionally the same as receiving an eager notification.

For the places where I didn't remove the handling of connection-dropped notifications, the reasons are pretty specific:

- I wanted to remove it for the `identify::dialer::Actor`, but we currently have that questionable need to report the number of live connections from there. If we didn't have that I think that actor wouldn't need to know about dropped connections at all.
- The `online_status::Actor` is the only place where I think the subscription is a good idea. Although to be fair it is only used by the taker to keep track of one specific connection (the maker's), even though we will hear about all dropped connections (probably only the maker's, but still). One could argue that we might just want to add a simple query mechanism to the `Endpoint` to actively check if the connection is still alive and remove the dropped-connection notifications altogether!

> Otherwise it is expected to see the warning above, and I don't think that is expected behavior.

Nope, the warning can be (and is) skipped if we don't have a connection to the peer.

da-kami: I am not sure I like this solution, because we now can have different state concerning a peer being connected in different actors. 

This might be resilient for these specific actors, but I think the original idea was that the endpoint is in charge of managing connected peers. So far we always got notified about connected peers from the endpoint, i.e. we passed the state of the endpoint on to subscribers. Now these actors change this state independent of the state in the endpoint. I am not saying this is wrong, but I am not sure this is what we want.

da-kami (approved):
I still don't see a reason why we should not supervise endpoints, so I still approve. 
I added another comment concerning the state of connected peers now being changed independent of the state in the endpoint. I am not sure this is good, but I think it is more resilient for each actor. We should be able to rely on the specific error here.

xtra-libp2p-offer/src/current/maker.rs
luckysori: > So far we always got notified about connected peers from the endpoint, i.e. we passed the state of the endpoint on to subscribers. Now these actors change this state independent of the state in the endpoint.

But it is not independent of the state in the endpoint! It is the endpoint that tells them about an actor no longer being connected, but only when they ask.

If we want to eagerly synchronise these states then we can reintroduce the notifications, but I don't see the need. In fact, I would like to eventually remove the feature altogether. But that might be too much.

da-kami: > It is the endpoint that tells them about an actor no longer being connected, but only when they ask.

I am not sure I can follow. Until this PR we subscribed to state changes of the endpoint, so the endpoint did notify subscribed protocol actors about peer disconnection. 

> If we want to eagerly synchronise these states then we can reintroduce the notifications, but I don't see the need. 

Fair enough, I think the solution you came up with works, it might just not be as efficient as it could be, because we might still try to open a protocol substream with a peer even if the endpoint already knows that the peer idisconneced. Acceptable I suppose.

> In fact, I would like to eventually remove the feature altogether. But that might be too much.

You mean the feature of protocol actors subscribing to state changes of connect/disconnect in the endpoint? Initially we did not have this feature, but that lead to having to query the endpoint about peer information for all protocols where we need all peers. In the long run it might just be more efficient (and resilient) to change these protocols to be polled by the `taker` instead of the `maker` pushing the information. Then, we also don't need `ping` from `maker` to `taker` anymore.

klochowicz:
I'm sorry, but I don't think the concerns I had before were addressed. The notification feature is still being broken as a part of this PR and we can't guarantee that any new actor will deal with its new idiosyncrasies. The fact that the existing actors were adjusted to deal with it being broken is not enough in my opinion; it can still lead to problems in the feature.

To me, there are two ways forward:
- Amending the peer/listener notifications feature with the additional message that tells the subscribers to reset their count (as outlined originally in my review)
- Completely removing the feature (as you hinted earlier).

@luckysori I don't mind completely removing the feature if you can show that it's not needed - but this should be done *before* we intentionally cripple peer/listener notifications.

I'm happy to pair up on finishing this. Whilst I believe that supervising Endpoint is a good thing, but at the same time I don't find it critical as it can only panic if passed an improper config (something we never do).

klochowicz:
> I still don't see a reason why we should not supervise endpoints, so I still approve.

I was never against supervising endpoints, I am just against having half-working subscriber feature (in the sense that we can't really rely on the cumulative state to be true).

luckysori:
No consensus was reached on the proposed solution.

It is not a critical feature, so we can safely close this PR and we do not need to action https://github.com/itchysats/itchysats/issues/2821 any time soon.

PR #2841, opened 2022-09-01
Reintroduce actor address in `Supervisor`

luckysori:
So that the supervised actor stays alive as long as the `Supervisor` stays alive.


luckysori:
I've decided against this because:

- It goes against the new design of `Supervisor`. It's not impossible to fix the tests, but I think the idea is that we should stop supervising if no one cares about the supervised actors (no `Address`es left).
- The first point leads me to think that we shouldn't use actors for long running tasks that don't receive external messages. Perhaps there are certain situations where an actor is useful if you handle several internal messages and no external messages. But in general it is a bad sign to have to `drop` the address or store it somewhere as `_<name_of_actor>` to never use it.

PR #2844, opened 2022-09-01
Check for DLC presence in one spot

luckysori:
Actually fixes what https://github.com/itchysats/itchysats/pull/2835 set out to fix.

Patch e39a0134950d92931a8534d7d8c3f4380d908e0c was a step in the right direction in terms of suppressing false positives about not being able to auto-rollover CFDs without a DLC, it wasn't enough.

It was insufficient because we were checking for DLC presence in more than one spot in the same method. By checking in only one spot we can simplify the method and also ensure that we check if the CFD is empty _before_ we check for DLC presence.

### DIFF
--- a/model/src/cfd.rs
+++ b/model/src/cfd.rs
@@ -652,12 +652,6 @@ impl Cfd {
         )
     }
 
-    fn expiry_timestamp(&self) -> Option<OffsetDateTime> {
-        self.dlc
-            .as_ref()
-            .map(|dlc| dlc.settlement_event_id.timestamp())
-    }
-
     fn margin(&self) -> Amount {
         match self.position {
             Position::Long => calculate_margin(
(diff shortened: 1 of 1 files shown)


luckysori:
Tested locally with DB with empty CFDs.

model/src/cfd.rs
```diff
@@ -704,16 +698,19 @@ impl Cfd {
         &self,
         now: OffsetDateTime,
     ) -> Result<(Txid, BitMexPriceEventId), CannotRollover> {
-        let expiry_timestamp = self.expiry_timestamp().ok_or(CannotRollover::NoDlc)?;
+        self.can_rollover()?;
+
+        let dlc = match &self.dlc {
```
da-kami: ❓Any particular reason why you prefer the `match` over the `ok_or`? I think it should be equivalent.

da-kami (approved):

model/src/cfd.rs
```diff
@@ -704,16 +698,19 @@ impl Cfd {
...
+            Some(dlc) => dlc,
+            None => return Err(CannotRollover::NoDlc),
+        };
```
klochowicz: perhaps neater:

```suggestion
        let dlc = self.dlc.as_ref().ok_or(CannotRollover::NoDlc)?;
```

klochowicz (approved):

model/src/cfd.rs
luckysori: Nope, I just forgot!

model/src/cfd.rs
luckysori: Yep, for sure!

da-kami (approved):

PR #2845, opened 2022-09-01
Fix counting already connected peers

da-kami:
In some cases we are not notified that a connection was dropped by the underlying network stack.
Thus, it can happen that we just replace a `peer_info` in the hashmap without removing it first.
Incrementing the gauge on every insert can thus lead to counting connections for one peer multiple times.
We now ensure that we only increment if the `peer_info` was not recorded already.

### DIFF
--- a/daemon/src/identify/dialer.rs
+++ b/daemon/src/identify/dialer.rs
@@ -76,16 +76,18 @@ impl Actor {
         let daemon_version = peer_info.daemon_version.clone();
         let environment = peer_info.environment;
 
-        NUM_LIBP2P_CONNECTIONS_GAUGE
-            .with(&HashMap::from([
-                (WIRE_VERSION_LABEL, wire_version.as_str()),
-                (DAEMON_VERSION_LABEL, daemon_version.as_str()),
-                (ENVIRONMENT_LABEL, environment.to_string().as_str()),
-            ]))
-            .inc();
-
         tracing::info!(%peer_id, %daemon_version, %environment, %wire_version, "New identify message received");
-        self.peer_infos.insert(peer_id, peer_info.clone());
+        if self.peer_infos.insert(peer_id, peer_info.clone()).is_none() {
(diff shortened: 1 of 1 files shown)


daemon/src/identify/dialer.rs
```diff
@@ -76,13 +76,17 @@ impl Actor {
...
-        NUM_LIBP2P_CONNECTIONS_GAUGE
-            .with(&HashMap::from([
-                (WIRE_VERSION_LABEL, wire_version.as_str()),
-                (DAEMON_VERSION_LABEL, daemon_version.as_str()),
-                (ENVIRONMENT_LABEL, environment.to_string().as_str()),
-            ]))
-            .inc();
+        // Only increment if we don't know the peer info already because sometimes we are not
+        // notified about ConnectionDropped
+        if self.peer_infos.get(&peer_id).is_none() {
```
bonomat: 🙃 You could do the same thing a few lines below when inserting into the map: 
 
From rust doc: 
> Inserts a key-value pair into the map.
If the map did not have this key present, None is returned.
If the map did have this key present, the value is updated, and the old value is returned. 

e.g. 
```
if let None = self.peer_infos.insert(peer_id, peer_info.clone()) { 
... 
}
```

bonomat (approved):

luckysori (approved):
Nice!

daemon/src/identify/dialer.rs
klochowicz: maybe even `if (...).is_none` {
} 

no need to bind it with `let`

klochowicz (approved):

daemon/src/identify/dialer.rs
da-kami: I like the proposal, it's cleaner. Done!

da-kami: > 

true, could just go for `is_none` here

da-kami: clippy agrees as well 🤪

PR #2851, opened 2022-09-01
Instrument send_offers with attr instead of at use

Restioson:
This will have one extra span in ConnectionEstablished.

Originally suggested by @holzeis in #2798


holzeis (approved):
LGTM 👍

klochowicz (approved):
👍

Issue #2822, opened 2022-08-31
Extract otel-tests and otel-tests-macros into a separate repo


klochowicz:
These crates have little to do with the project, and can be moved out of it. This will shave a bit of compilation time (rust-cache works only for external crates).

Incidentally, this is a blocker for partitioning our tests on CI: #2741 
(cargo nextest has problems with parsing tests from these crates).

luckysori:
I'm not sure we can easily do this. These crates know about other crates in the workspace:

https://github.com/itchysats/itchysats/blob/9926f9d201dccc6a70dfeaf323dfd09842cdf7cd/otel-tests/src/lib.rs#L62-L75

klochowicz:
> I'm not sure we can easily do this. These crates know about other crates in the workspace:
> 
> https://github.com/itchysats/itchysats/blob/9926f9d201dccc6a70dfeaf323dfd09842cdf7cd/otel-tests/src/lib.rs#L62-L75

whoops. indeed, that might be fragile to do so :(

Restioson:
There might be some way to provide this info to the macro in a different way, but I'm not 100% sure what is best. It would depend if the goal is to just make it independent from the other crates, or configurable enough to be useful for other people.

klochowicz:
> There might be some way to provide this info to the macro in a different way, but I'm not 100% sure what is best. It would depend if the goal is to just make it independent from the other crates, or configurable enough to be useful for other people.

I was thinking maybe a optional toml config file with `directives` list?

edit: I added #2854 as an attempt of dealing with this awkwardness.

klochowicz:
this issue has been addressed by @Restioson.

PR #2854, opened 2022-09-01
Supply a config file with otel-tests tracing directives

klochowicz:
Removes the need for the crate to know about other crates that are set further
down the dependency chain and allows for extracting this generic crate out of
the itchysats workspace.

New `otel-tests.toml` file needs to be located in the directory of the crate
using `otel-tests` macro for tests.

In case the file is not present or it is malformed, an otel-test will fail to compile.

This paves the way to #2822

### DIFF
--- a/otel-tests/src/lib.rs
+++ b/otel-tests/src/lib.rs
@@ -6,6 +6,11 @@ use tracing_subscriber::EnvFilter;
 use tracing_subscriber::Layer;
 use tracing_subscriber::Registry;
 
+#[derive(serde::Deserialize)]
+struct Config {
+    directives: Vec<String>,
+}
+
 #[doc(hidden)]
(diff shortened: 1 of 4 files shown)


luckysori (approved):

otel-tests/src/lib.rs
```diff
@@ -59,20 +64,25 @@ pub fn init_tracing(module: &'static str) {
             None
         };
 
-        let filter = EnvFilter::from_default_env()
+        let global_directive = LevelFilter::WARN.into();
```
holzeis, line 67: ❓ why don't we extract that directive also to the otel-tests.toml?

holzeis (approved):

Restioson (approved):

PR #2857, opened 2022-09-02
Do not dial peers that allegedly don't listen

### LINKED ISSUE #2763: Do not send offers via wrong `offer` protocol to peers
Since https://github.com/itchysats/itchysats/pull/2680 we are sending offers via two different protocols to _all_ takers. One of those protocols will always fail during the negotiation phase. I documented this behaviour here: e4f910f35070c47672c8e37e224b51e3b532e833.

As noted in that commit message, on the maker side we can suppress these warnings because we can choose to ignore that specific error when opening the substream. On the taker side it is emitted in a place where we handle all possible incoming substreams, so it's unavoidable.

To not send needless offers to takers we need to evolve the `Endpoint` to be able to decide during runtime what protocol to dial to which takers, depending on their supported listen protocols.

---

If you look at the [Grafana logs](https://itchysats.grafana.net/explore?orgId=1&left=%7B%22datasource%22:%22grafanacloud-itchysats-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22editorMode%22:%22builder%22,%22expr%22:%22%7Bsetup%3D%5C%22testnet%5C%22%7D%20%7C%3D%20%60negotiation%20failed%60%22,%22queryType%22:%22range%22%7D%5D,%22range%22:%7B%22from%22:%221660809217202%22,%22to%22:%221661734356186%22%7D%7D) you can see that the frequency for this particular log message increased just after https://github.com/itchysats/itchysats/pull/2680 merged.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2763.

### DIFF
--- a/xtra-libp2p-offer/src/current/maker.rs
+++ b/xtra-libp2p-offer/src/current/maker.rs
@@ -52,14 +52,14 @@ impl Actor {
         };
 
         let err_handler = move |e: anyhow::Error| async move {
-            if let Some(xtra_libp2p::Error::NegotiationFailed(
-                xtra_libp2p::NegotiationError::Failed,
-            )) = e.downcast_ref::<xtra_libp2p::Error>()
-            {
-                // It's normal to disagree on the protocols now that we broadcast on both versions
-                // to _all_ our peers
-            } else {
-                tracing::warn!(%peer_id, "Failed to send offers: {e:#}")
+            match e.downcast_ref::<xtra_libp2p::Error>() {
+                Some(xtra_libp2p::Error::ProtocolNotSupportedByPeer) => {
+                    // Some peers may not support this protocol as listeners
+                }
+                Some(xtra_libp2p::Error::NegotiationFailed(_)) => {
+                    tracing::debug!(%peer_id, "Failed to send offers: {e:#}")
+                }
+                _ => tracing::warn!(%peer_id, "Failed to send offers: {e:#}"),
             }
         };
 
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -62,6 +62,7 @@ pub struct Endpoint {
     inflight_connections: HashSet<PeerId>,
     connection_timeout: Duration,
     subscribers: Subscribers,
+    peer_listen_protocols: HashMap<PeerId, HashSet<String>>,
 }
 
 /// Open a substream to the provided peer.
@@ -153,6 +154,12 @@ pub struct NewInboundSubstream {
     pub stream: Substream,
 }
 
+/// Message used to tell the [`Endpoint`] about the listen protocols that a peer supports.
+pub struct RegisterListenProtocols {
+    pub peer_id: PeerId,
+    pub listen_protocols: HashSet<String>,
+}
+
 #[derive(Debug, Error)]
 pub enum Error {
     #[error("No connection to {0}")]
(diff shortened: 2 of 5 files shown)


xtra-libp2p-offer/src/current/maker.rs
```diff
@@ -52,14 +52,14 @@ impl Actor {
...
-                xtra_libp2p::NegotiationError::Failed,
-            )) = e.downcast_ref::<xtra_libp2p::Error>()
-            {
-                // It's normal to disagree on the protocols now that we broadcast on both versions
-                // to _all_ our peers
-            } else {
-                tracing::warn!(%peer_id, "Failed to send offers: {e:#}")
+            match e.downcast_ref::<xtra_libp2p::Error>() {
+                Some(xtra_libp2p::Error::ProtocolNotSupportedByPeer) => {
+                    // Some peers may not support this protocol as listeners
```
holzeis, line 57: 🤔 shouldn't we add a log message here?

luckysori: I don't think logging will add much other than noise.

The current expected behaviour is that we will attempt to open a substream but be stopped by the `Endpoint` if it knows that the peer doesn't listen. It's not noteworthy because these `offer` actors learn about _all_ connected peers from the `Endpoint`, so we expect some of those peers to be incompatible (because we run two versions of the same protocol side by side on the maker).

What we could do is remove the incompatible peer from the internal state of the `offer::Actor`, so that we don't keep spamming the `Endpoint` if we know it's not gonna work.

holzeis: that sounds reasonable.. 🙂

luckysori: But let's add the extra functionality with #2839, because we already introduce an internal message for the same purpose with 0d1dd1bdb5cc12e38bfcf51dfc4d9cd43a58adf8.

da-kami (approved):
It would be great if you could add a contained protocol test for this! I think we should be able to quite easily spin up a setup where we can test this behavior.

xtra-libp2p/src/endpoint.rs
```diff
@@ -250,10 +259,37 @@ impl Endpoint {
...
+            .collect::<Vec<_>>();
+
+        if !protocols
+            .iter()
+            .any(|protocol| listen_protocols.contains(protocol))
+        {
+            return Err(Error::ProtocolNotSupportedByPeer);
         }
+
+        Ok(())
```
klochowicz: 🙃 imho a bit more readable version (and a one-liner ;D) 

```suggestion
        protocols
            .iter()
            .any(|protocol| listen_protocols.contains(protocol))
            .then(|| ())
            .ok_or(Error::ProtocolNotSupportedByPeer)
```

klochowicz (approved):

xtra-libp2p/src/endpoint.rs
luckysori: I had this in mind at first but I decided against it because `.then(|| ())` is not so obvious. But given that you like it I will write it like this!

luckysori:
> It would be great if you could add a contained protocol test for this! I think we should be able to quite easily spin up a setup where we can test this behavior.

I think I did try for a bit and it was annoying. But I'll try again!

EDIT: Thanks for your suggestion! I was approaching the tests from the wrong angle before, trying to add them to `identify`. They clearly belong in `xtra_libp2p`. I've added them to the first patch.

PR #2859, opened 2022-09-02
Consider rollover fees stale after 1 minute

### LINKED ISSUE #2531: Do not allow maker rollovers if offer is not up-to-date
We currently assume that the offer is up-to-date if the maker automation is running and able to accept rollovers. This logic is flawed because the automation could easily be running and at the same time failing to update the offer.

In any case, the situation gets slightly worse after the maker does not need rollover confirmation via its HTTP API. This will happen once #2378 is merged.

Therefore we need to introduce a mechanism to ensure that rollovers only happen if the offer is up-to-date.

_Originally posted by @luckysori in https://github.com/itchysats/itchysats/pull/2378#discussion_r918489749_

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/2531.

This prevents the maker from agreeing to rollover if it doesn't have up-to-date fees.

### DIFF
--- a/maker/src/cfd.rs
+++ b/maker/src/cfd.rs
@@ -1,4 +1,5 @@
 use crate::metrics::time_to_first_position;
+use anyhow::bail;
 use anyhow::Context;
 use anyhow::Result;
 use async_trait::async_trait;
@@ -19,9 +20,12 @@ use model::TxFeeRate;
 use nonempty::NonEmpty;
 use std::collections::HashMap;
 use time::Duration;
+use time::OffsetDateTime;
 use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 use xtras::SendAsyncSafe;
+
+const ROLLOVER_PARAMS_TTL: Duration = Duration::minutes(5);
 
 #[derive(Clone)]
 pub struct NewOffers {
@@ -59,7 +63,7 @@ pub struct TakerDisconnected {
 }
 
 #[derive(Clone, Copy)]
-pub struct GetRolloverParams;
+pub struct GetRolloverParams(ContractSymbol);
 
 #[derive(Clone, Debug)]
 pub struct OfferParams {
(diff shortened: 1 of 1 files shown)


maker/src/cfd.rs
```diff
@@ -19,10 +20,13 @@ use model::TxFeeRate;
 use nonempty::NonEmpty;
 use std::collections::HashMap;
 use time::Duration;
+use time::OffsetDateTime;
 use xtra::prelude::MessageChannel;
 use xtra_productivity::xtra_productivity;
 use xtras::SendAsyncSafe;
 
+const ROLLOVER_PARAMS_TTL: Duration = Duration::MINUTE;
```
bonomat: Wouldn't it be better that this is passed in through the HTTP API?

bonomat: Also, the funding rate might have a way longer expiry, something like <8h.

bonomat (approved):
LGTM

maker/src/cfd.rs
luckysori: > Wouldn't it be better that this is passed in through the HTTP API?

It would give us more flexibility, but the value seems limited unless we anticipate ever wanting to change it. Do you think it's worth it to add it to the `OfferParams`?

> Also, the funding rate might have a way longer expiry, something like <8h.

Good point. The value I chose was extremely arbitrary.

da-kami: I agree with @bonomat that it would be better to just set this through the API as part of the update. For me this is not only about "it's more flexible" but also a better design, because the data model itself tells you that this has an expiry and it is not hidden inside.

If you decide to keep it hardcoded: I think one minute is quite little given that there is sometimes latency in the system and we don't expect the offer to be refreshed faster than one-minute cycles. 
I think this might lead to failing more rollovers than wanted.
I would allow at least two minutes if we hardcode it.

da-kami (approved):
Thanks for tackling this. Please consider either changing the value to `2` minutes or changing the API.

maker/src/cfd.rs
luckysori: @da-kami, @bonomat: At first I was going to go with your suggestions, but I don't want to rush to make hasty changes to a HTTP API that I think already has some problems. I've opened a discussion about this topic: https://github.com/itchysats/itchysats/discussions/2886.

I've also opened a [ticket](https://github.com/itchysats/itchysats/issues/2887) to make this time-to-live configurable.

Finally, I'm gonna set the constant to 5 minutes. It's true that the value should live for 8 hours, but the updates to the offer parameters are so frequent that using that value would lead to problems e.g. setting it to 8 hours a few minutes before the value actually changes and then never hearing back from the API because of a problem, leading to allowing rollovers with stale data for an extended period of time.

2 minutes is also fine, but I don't think we have to cut it so close.

PR #2862, opened 2022-09-02
Set maximum price of inverse payout curve to Olivia's max attestation price

### LINKED ISSUE #2571: Short liquidation events when maker goes short are unusable

luckysori:
Fix https://github.com/itchysats/itchysats/issues/2571.

This is a second attempt after the simple https://github.com/itchysats/itchysats/pull/2637 broke every protocol under the sun and we had to revert it.

The diff is large because we have to introduce new versions of several protocols to be able to introduce this small breaking change into the inverse payout curve implementation. This can be a good thing though, as pointed out by @bonomat, as it gives us an opportunity to introduce other breaking changes for these protocols before the next release.


model/src/payout_curve.rs
```diff
@@ -76,8 +76,9 @@ impl Payouts {
         (leverage_long, leverage_short): (Leverage, Leverage),
         n_payouts: usize,
         fee: CompleteFee,
+        inverse_max_price_config: InverseMaxPrice,
```
bonomat, line 121: What do you think of not exposing this but creating a new function something like `new_inverse_deprecated` which defaults to `InverseMaxPrice::DoubleOfInitial`. That way we don't need to expose the constant outside of this module and the caller does not need to know about it.

xtra-libp2p-rollover/src/next/taker.rs
```diff
@@ -0,0 +1,368 @@
...
+            },
+            {
+                let executor = self.executor.clone();
+                move |e| async move {
+                    emit_failed(order_id, e, &executor).await;
+                }
+            },
+        );
+    }
+}
```
bonomat: I did not check the logic in here. I guess this is all copy&paste?

daemon/src/collab_settlement/current.rs
```diff
@@ -0,0 +1,5 @@
+pub mod maker;
+pub mod protocol;
+pub mod taker;
+
+pub const PROTOCOL: &str = "/itchysats/collab-settlement/2.0.0";
```
bonomat, line 5: Since you introduce 0.2.0 we could go for a breaking change here and implement: https://github.com/itchysats/itchysats/issues/2924 :)

model/src/payout_curve.rs
```diff
@@ -86,6 +87,13 @@ impl Payouts {
             fee,
         )?;
 
+        if let InverseMaxPrice::OliviaMax = inverse_max_price_config {
+            let n_payouts = payouts.len() - 1;
+            let short_liquidation = payouts.get_mut(n_payouts).expect("several payouts");
+            short_liquidation.range =
+                *short_liquidation.range.start()..=maia_core::interval::MAX_PRICE_DEC;
```
bonomat, line 136: You did not have to adapt a test for this?

bonomat (commented):
Added some initial comments.  Looking good so far.

model/src/payout_curve.rs
luckysori: Yep, that's a better idea! Let me give it a go.

xtra-libp2p-rollover/src/next/taker.rs
luckysori: It should be a straight copy of the original `current`, apart from using `MaxInversePrice::OliviaMax`.

It is a limitation of our approach to backwards compatibility (creating multiple copies with some differences): we can't easily tell what the differences are.

daemon/src/collab_settlement/current.rs
luckysori: Sure, we should prioritise that ticket before we release something with these changes.

model/src/payout_curve.rs
luckysori: Well, I didn't modify the inverse payout curve implementation. That's where most tests are located.

The `proptest` below still works after these changes because it doesn't check against hard-coded values.

model/src/payout_curve.rs
luckysori: I did it everywhere except for a few places in `model` itself. 

And I marked `InverseMaxPrice` and `new_inverse` as `pub(crate)` to ensure that they're not used elsewhere.

bonomat (approved):
Good work!

PR #2899, opened 2022-09-06
Add the capability to block peers through a configurable blocklist

### LINKED ISSUE #2527: Configurable `peer_id` blacklist for the maker
We see one peer-id spamming the maker, which is annoying. 

There are some solutions to deal with that that are less restrictive than a blacklist (i.e. find out what peer it is, change log levels, ...), but in the long run a feature to ignore certain peer-ids might actually come in handy for the maker. 

It would be good if we could store this information in the database and create an endpoint to add/delete blacklisted peers.

Restioson:
Fixes #2527. I have not currently included documentation on this feature as I was not 100% on where to put it. I am happy to write some docs for this once someone can tell me where to write it :sweat_smile: It currently is just a simple plaintext list with each peer id on a new line, stored in `[data_dir]/blocked_peers`.

This does not address IP blacklisting. Peers which do not specify their peer ID in their multiaddr will only be blocked after the libp2p connection is upgraded, which could result in `DEBUG libp2p_core::upgrade::apply: Successfully applied negotiated protocol` messages being logged. I am not sure if this is relevant as it could just be disabled on mainnet.

Currently, blocked peers are logged with their peer id each time they are blocked. This is logged to the blocked_peers target at INFO, so this can be easily disabled. 

## TODO
- [x] TOML format
- [x] Test
- [x] Should Arc, &'static, or clone be used?
- [x] Document (where?)
- [x] Should the taker get an error?

### DIFF
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -53,6 +53,7 @@ use model::TxFeeRate;
 use model::SETTLEMENT_INTERVAL;
 use rust_decimal::Decimal;
 use rust_decimal_macros::dec;
+use std::collections::HashSet;
 use std::net::IpAddr;
 use std::net::Ipv4Addr;
 use std::net::SocketAddr;
@@ -557,12 +558,13 @@ pub async fn start_both() -> (Maker, Taker) {
     (maker, taker)
 }
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub struct MakerConfig {
     oracle_pk: XOnlyPublicKey,
     seed: RandomSeed,
     n_payouts: usize,
     libp2p_port: u16,
+    blocked_peers: HashSet<xtra_libp2p::libp2p::PeerId>,
 }
 
 impl Default for MakerConfig {
@@ -572,6 +574,7 @@ impl Default for MakerConfig {
             seed: RandomSeed::default(),
             n_payouts: N_PAYOUTS,
             libp2p_port: portpicker::pick_unused_port().expect("to be able to find a free port"),
+            blocked_peers: HashSet::new(),
         }
     }
 }
@@ -705,6 +708,7 @@ impl Maker {
             projection_actor,
             identities.clone(),
             endpoint_listen.clone(),
+            config.blocked_peers.clone(), // TODO(restioson): test
         )
         .unwrap();
 
--- a/daemon-tests/src/lib.rs
+++ b/daemon-tests/src/lib.rs
@@ -708,7 +708,7 @@ impl Maker {
             projection_actor,
             identities.clone(),
             endpoint_listen.clone(),
-            config.blocked_peers.clone(), // TODO(restioson): test
+            config.blocked_peers.clone(),
         )
         .unwrap();
 
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -18,12 +18,16 @@ use model::olivia;
 use model::Role;
 use model::SETTLEMENT_INTERVAL;
 use rocket_cookie_auth::users::Users;
+use serde::Deserialize;
 use shared_bin::catchers::default_catchers;
 use shared_bin::cli::Withdraw;
 use shared_bin::fairings;
 use shared_bin::logger;
+use std::collections::HashSet;
 use std::net::SocketAddr;
+use std::path::Path;
 use tokio_extras::Tasks;
+use xtra_libp2p::libp2p::PeerId;
 use xtras::supervisor::always_restart;
 use xtras::supervisor::Supervisor;
 
@@ -117,6 +121,19 @@ async fn main() -> Result<()> {
     let db =
         sqlite_db::connect(data_dir.join("maker.sqlite"), opts.ignore_migration_errors).await?;
 
+    let blocked_peers_path = data_dir.join("blocked_peers.toml");
+    let blocked_peers = load_blocked_peers(&blocked_peers_path)
+        .await
+        .unwrap_or_else(|err| {
+            tracing::error!(
+                ?blocked_peers_path,
+                %err,
+                "Error loading blocked peers list; ignoring and allowing all connections",
+            );
+            HashSet::default()
+        });
+
+    // Create actors
     let endpoint_listen =
         daemon::libp2p_utils::create_listen_tcp_multiaddr(&p2p_socket.ip(), p2p_socket.port())
             .expect("to parse properly");
@@ -161,6 +178,7 @@ async fn main() -> Result<()> {
         projection_actor.clone(),
         identities,
         endpoint_listen,
+        blocked_peers,
     )?;
 
     if let Some(password) = opts.password {
@@ -240,4 +258,19 @@ impl rocket_cookie_auth::Database for RocketAuthDbConnection {
         self.inner.clone().update_password(password).await?;
         Ok(())
     }
+}
+
+/// Convenience type to load the blocked peer list from toml
+#[derive(Deserialize)]
+struct BlockedPeers {
+    blocked: HashSet<PeerId>,
+}
+
+async fn load_blocked_peers(blocked_peers_path: &Path) -> Result<HashSet<PeerId>> {
+    anyhow::ensure!(
+        blocked_peers_path.try_exists()?,
+        "No blocked peers file found in {blocked_peers_path:?}",
+    );
+    let raw = tokio::fs::read_to_string(blocked_peers_path).await?;
+    Ok(toml::from_str::<BlockedPeers>(&raw)?.blocked)
 }
--- a/xtra-libp2p/examples/hello_world_listener.rs
+++ b/xtra-libp2p/examples/hello_world_listener.rs
@@ -7,6 +7,8 @@ use futures::StreamExt;
 use libp2p_core::identity::Keypair;
 use libp2p_core::Multiaddr;
 use libp2p_tcp::TokioTcpConfig;
+use std::collections::HashSet;
+use std::sync::Arc;
 use std::time::Duration;
 use tracing::Level;
 use xtra::prelude::*;
@@ -56,6 +58,7 @@ async fn main() -> Result<()> {
         Duration::from_secs(30),
         [("/hello-world/1.0.0", hello_world_addr.clone().into())],
         Subscribers::default(),
+        Arc::new(HashSet::default()),
     )
     .create(None)
     .spawn_global();
--- a/xtra-libp2p/src/endpoint.rs
+++ b/xtra-libp2p/src/endpoint.rs
@@ -25,6 +25,7 @@ use std::collections::HashMap;
 use std::collections::HashSet;
 use std::marker::PhantomData;
 use std::pin::Pin;
+use std::sync::Arc;
 use std::time::Duration;
 use thiserror::Error;
 use tokio_extras::Tasks;
@@ -60,6 +61,7 @@ pub struct Endpoint {
     inbound_substream_channels: HashMap<&'static str, MessageChannel<NewInboundSubstream, ()>>,
     listen_addresses: HashSet<Multiaddr>,
     inflight_connections: HashSet<PeerId>,
+    blocked_peers: Arc<HashSet<PeerId>>,
     connection_timeout: Duration,
     subscribers: Subscribers,
     peer_listen_protocols: HashMap<PeerId, HashSet<String>>,
@@ -224,6 +226,7 @@ impl Endpoint {
         connection_timeout: Duration,
         inbound_substream_handlers: [(&'static str, MessageChannel<NewInboundSubstream, ()>); N],
         subscribers: Subscribers,
+        blocked_peers: Arc<HashSet<PeerId>>,
     ) -> Self
     where
         T: Transport + Send + Sync + 'static,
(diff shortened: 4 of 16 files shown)


daemon-tests/src/lib.rs
```diff
@@ -721,6 +724,7 @@ impl Maker {
             projection_actor,
             identities.clone(),
             endpoint_listen.clone(),
+            config.blocked_peers.clone(), // TODO(restioson): test
```
klochowicz: do you want to keep this TODO or address it?

xtra-libp2p/src/endpoint.rs
```diff
@@ -215,6 +217,7 @@ impl Endpoint {
         connection_timeout: Duration,
         inbound_substream_handlers: [(&'static str, MessageChannel<NewInboundSubstream, ()>); N],
         subscribers: Subscribers,
+        blocked_peers: Arc<HashSet<PeerId>>,
```
klochowicz, line 229: 🔧  I think we don't need `Arc` here. it's immutable data in this changeset, and compiles just fine without it.

maker/src/main.rs
```diff
@@ -123,6 +127,20 @@ async fn main() -> Result<()> {
...
+        Err(error) => {
+            tracing::error!(
+                ?blocked_peers_path,
+                %error,
+                "Error loading blocked peers list; ignoring and allowing all connections",
+            );
+
+            HashSet::default()
+        }
+    };
```
klochowicz: maybe we just just use TOML instead of rolling out our own format? 
it should be super easy to derive from that. As a bonus, it could be easily expanded to different variables later on, and it is lintable with dprint (you can see how to do it in `otel-tests`)

maker/src/main.rs
```diff
@@ -210,3 +229,29 @@ async fn main() -> Result<()> {
...
+                        %error, %line, "Error parsing line in blocked peers list; skipping",
+                    );
+                    None
+                }
+            })
+            .collect();
+        Ok(blocked)
+    } else {
+        tracing::info!(
+            ?blocked_peers_path,
```
klochowicz: this function would go away if we parsed using toml lib

xtra-libp2p/examples/hello_world_listener.rs
```diff
@@ -56,6 +56,7 @@ async fn main() -> Result<()> {
         Duration::from_secs(30),
         [("/hello-world/1.0.0", hello_world_addr.clone().into())],
         Subscribers::default(),
+        Default::default(),
```
klochowicz: nit(subjective): I find it a bit clearer when used with `HashSet::default()` (tells something about the type when reading the code)

xtra-libp2p/src/endpoint.rs
```diff
@@ -483,6 +488,19 @@ impl Endpoint {
                                 remote_addr,
                                 ..
                             }) => {
+                                match PeerId::try_from_multiaddr(&remote_addr) {
```
klochowicz, line 526: can we add test to test this functionality? it should be awesome if we had a test that shows blacklisting feature inside `xtra-libp2p` crate.

klochowicz (commented):
great start on tackling a much needed functionality :)

xtra-libp2p/src/endpoint.rs
```diff
@@ -483,6 +488,19 @@ impl Endpoint {
...
+                                    Some(peer_id) if blocked_peers.contains(&peer_id) => {
+                                        tracing::info!(
```
holzeis, line 528: 🤔 Shouldn't we also respond with an error to the taker? I guess simply skipping here - will produce a timeout on the taker side.

Restioson: I think the intended use is to block peers that do not speak the itchysats protocol. In this case, a message wouldn't be of much use. However, maybe there are other cases?

xtra-libp2p/src/endpoint.rs
Restioson: That would involve cloning it everywhere, which is wasteful RAM-wise. I thought about passing around an &'static HashSet using Box::leak, which is currently OK, but this could lead to memory leaks if life cycles change.

maker/src/main.rs
Restioson: While it is technically rolling out own format, it's just lines of peer ids, so this doesn't add any maintenance burden and isn't confusing imo. It might be beneficial if we add e.g comments later.

maker/src/main.rs
Restioson: We still need to handle the no-file case so I don't think so, but the happy path would be a little shorter

xtra-libp2p/examples/hello_world_listener.rs
Restioson: It would be Arc::new(HashSet::default()) but I can change it to this top

xtra-libp2p/src/endpoint.rs
Restioson: Good point! I intended to do this initially but then I forgot to

daemon-tests/src/lib.rs
Restioson: Relevant to the comment down below about testing, so I'll keep it until I add the test

maker/src/main.rs
Restioson: Done now :)

xtra-libp2p/src/endpoint.rs
holzeis: Isn't that scenario already handled here? 
https://github.com/itchysats/itchysats/blob/466a4efbf2cf35436fcb396be71e83f6b3fb818c/xtra-libp2p/src/endpoint.rs#L341-L343

I figured blocking peer ids would be used to block potentially malicious actors or otherwise strange behaving peers.

Restioson: That is true, but if it gets that far, does it not already entail a lot of logs that could potentially clog stuff up? Unfortunately #2527 is a bit vague on what "spamming" means. @da-kami would you be able to clarify this?

xtra-libp2p/src/endpoint.rs
Restioson: Is it ok to leave it as Arc for now?

xtra-libp2p/src/endpoint.rs
Restioson: Done in latest commit

xtra-libp2p/src/endpoint.rs
da-kami: > That is true, but if it gets that far, does it not already entail a lot of logs that could potentially clog stuff up? Unfortunately #2527 is a bit vague on what "spamming" means. @da-kami would you be able to clarify this?

At the time of writing the ticket the blacklist was meant for peers that connect but don't support *any* of our protocols. These are peers we can't do anything with - spam was meant as in log spam on the maker side, because we see those peers connecting, but then can't really perform any protocol with them. 

The idea was to be able to add `peer-id`s to a manually managed blacklist. The assumption is, that such peers actually don't run itchysats but try to connect from a different application. We obviously have to be careful not to add `peer-id`s to that list that are wrongly configured and later change their configuration to work - those would be blocked then.

That said: I don't think we can "respond" to peers when they try to establish a connection. Sending a response would require a protocol, but this is at a lower level. Unless I'm missing something we can only close the connection and this is what was planned for the ticket.

Restioson: Thanks for the clarification! In that case, can this be left as is?

xtra-libp2p/src/endpoint.rs
klochowicz: I guess `Arc` is much better than the other proposed option, there's no point in going into leaking memory just save a bit of memory.

let's keep `Arc` 👍 

Indeed, passing an `Arc` there is a nifty trick that might save some RAM. perhaps we should consider using it in more places...

xtra-libp2p/src/endpoint.rs
klochowicz: I would say 'yes'.

xtra-libp2p/tests/basic.rs
```diff
@@ -57,6 +59,37 @@ async fn hello_world() {
     assert_eq!(string, "Hello Bob!");
 }
 
+#[tokio::test]
+async fn blocked_peers_cannot_connect() {
+    let alice_hello_world_handler = HelloWorld::default().create(None).spawn_global();
+    let (alice, bob, _) = alice_and_blockable_bob(
```
klochowicz: it's probably not worth making an abstraction for a once-off test (this function reads really fishy), especially when one of the parameter is a bool 😢

IMHO much better would be just doing calling `make_node` and `make_node_with_blocks` here. 

Then it actually reads easier and we're getting rid of all the weirdness.

xtra-libp2p/tests/util.rs
```diff
@@ -26,6 +26,13 @@ pub struct Node {
 
 pub fn make_node<const N: usize>(
     substream_handlers: [(&'static str, MessageChannel<NewInboundSubstream, ()>); N],
+) -> Node {
+    make_node_with_blocks(substream_handlers, Arc::new(HashSet::new()))
+}
+
+pub fn make_node_with_blocks<const N: usize>(
```
klochowicz: `make_node_with_blocklist` ?

maker/src/main.rs
```diff
@@ -205,3 +225,22 @@ async fn main() -> Result<()> {
 
     Ok(())
 }
+
+/// Convenient type to load from toml
```
klochowicz: nit(spelling): I think you meant "Convenience type"?

maker/src/main.rs
```diff
@@ -205,3 +225,22 @@ async fn main() -> Result<()> {
...
+    if blocked_peers_path.try_exists()? {
+        let raw = tokio::fs::read_to_string(blocked_peers_path).await?;
+        Ok(toml::from_str::<BlockedPeers>(&raw)?.blocked)
+    } else {
+        tracing::info!(
+            ?blocked_peers_path,
+            "No blocked peers list found; ignoring and allowing all connections",
+        );
+        Ok(Default::default())
+    }
```
klochowicz: there's no need for special handling here; when you call the function you also just allow all peers in case of any error

```rust
async fn load_blocked_peers(blocked_peers_path: &Path) -> Result<HashSet<PeerId>> {
    anyhow::ensure!(
        blocked_peers_path.try_exists(),
        "No blocked peers file found in {blocked_peers_path}",
    );
    let raw = tokio::fs::read_to_string(blocked_peers_path).await?;
    Ok(toml::from_str::<BlockedPeers>(&raw)?.blocked)
}
```

maker/src/main.rs
```diff
@@ -123,6 +127,21 @@ async fn main() -> Result<()> {
...
+        Err(error) => {
+            tracing::error!(
+                ?blocked_peers_path,
+                %error,
+                "Error loading blocked peers list; ignoring and allowing all connections",
+            );
+
+            HashSet::default()
+        }
+    };
```
klochowicz: nit: perhaps a bit more idiomatic (feel free to push back if it's too subjective)

```rust
    let blocked_peers = load_blocked_peers(&blocked_peers_path).unwrap_or_else(|error| {
        tracing::error!(
            ?blocked_peers_path,
            %error,
            "Error loading blocked peers list; ignoring and allowing all connections",
        );
        HashSet::default()
    });
```

klochowicz (approved):
sorry for the wait! looks good, just some minor things :-)

xtra-libp2p/src/endpoint.rs
da-kami: Thanks for the discussion, I would have asked as well :D

xtra-libp2p/src/endpoint.rs
```diff
@@ -530,6 +548,16 @@ impl Endpoint {
                                                     None => format!("Failed to connect with multi-address: {remote_addr}"),
                                                 }
                                             })?;
+
+                                        if blocked_peers.contains(&peer_id) {
+                                            tracing::info!(
+                                                target: "blocked_peers",
+                                                peer_id = %peer_id, // Weird but required
```
da-kami, line 555: ❓ When using a target the syntax for fields changes? `%peer_id` does not work?

da-kami (approved):
TLDR; Looks good, nice work!

---

When reading through the code I thought about using a `config.toml` instead, also thought about using routes instead to define blocked peers in the database. But I came to the conclusion that this solution is good and if we want to change should only if we find a good reason :D

xtra-libp2p/tests/util.rs
Restioson: Done!

xtra-libp2p/src/endpoint.rs
Restioson: Yea, just `%peer_id` tries to use the `Value` impl, which doesn't exist, instead of the `Display` impl. I think this is a bug.

maker/src/main.rs
Restioson: Good catch, thanks!

klochowicz (approved):
Great work; thank you for addressing my comments. 

as for the last leftover question (documenting) I think that adding a changelog entry with a short description should be enough.

If you could rebase and squash, it would be awesome (I can't do that as it's on a fork 😭)

PR #2903, opened 2022-09-07
Replace basic-auth with the much better cookie-auth.

bonomat:
Replace basic-auth with the much better cookie-auth. 

The default password for the pre-existing user is `weareallsatoshi`. You can overwrite it using the `--password` argument. If you start with the default password you are asked to change it at first login. 

Have a look, give it a try and let me know what you think. 

https://user-images.githubusercontent.com/224613/188820762-e5cd96dc-019a-4d10-a54e-6caf03875a3b.mov


da-kami:
> The default password for the pre-existing user is weareallsatoshi

On Umbrel the user gets the password displayed in the app store and we set it with `--password` using the `APP_PASSWORD` envvar at the moment.
What is the expectation on rolling this feature out to Umbrel users?
Are we able to set this to the password we use on Umbrel for users on Umbrel?

rocket-cookie-auth/src/user.rs
```diff
@@ -0,0 +1,76 @@
...
+use rocket::Request;
+use serde::Deserialize;
+use serde::Serialize;
+
+/// A request guard that can be included in handler definitions to enforce authentication.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone, Hash, PartialOrd, Ord)]
+pub struct User {
+    pub id: u32,
+    #[serde(skip_serializing)]
+    pub(crate) password: String,
```
Restioson: Do we want to include a hash scheme version in case of some future issue so that old passwords can be noted to be updated to the new storage scheme?

bonomat:
> > The default password for the pre-existing user is weareallsatoshi
> 
> 
> 
> On Umbrel the user gets the password displayed in the app store and we set it with `--password` using the `APP_PASSWORD` envvar at the moment.
> 
> What is the expectation on rolling this feature out to Umbrel users?
> 
> Are we able to set this to the password we use on Umbrel for users on Umbrel?

Did you read the PR description? 
You can still overwrite it using : --password
If none is set it will default to 'weareallsatoshi' and needs to be changed in first login

rocket-cookie-auth/src/user.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.password = password;
+        Ok(())
+    }
+}
+
+pub fn create_password(password: &str) -> Result<String> {
+    let password = password.as_bytes();
+    let salt = rand_string(10);
+    let config = argon2::Config::default();
+    let hash = argon2::hash_encoded(password, salt.as_bytes(), &config)?;
```
Restioson, line 55: This should maybe be run inside a tokio `spawn_blocking` or `block_in_place` task

da-kami:
> > > The default password for the pre-existing user is weareallsatoshi
> > 
> > 
> > On Umbrel the user gets the password displayed in the app store and we set it with `--password` using the `APP_PASSWORD` envvar at the moment.
> > What is the expectation on rolling this feature out to Umbrel users?
> > Are we able to set this to the password we use on Umbrel for users on Umbrel?
> 
> Did you read the PR description? You can still overwrite it using : --password If none is set it will default to 'weareallsatoshi' and needs to be changed in first login

OK, so the expectation for Umbrel is, that we always overwrite it?
I did read the PR description and am aware that one can overwrite it, I was wondering if you had in mind that Umbrel users can also change it. I suppose that is not planned then and on Umbrel it will just be the `APP_PASSWORD`, correct?

rocket-cookie-auth/src/forms.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+#[allow(dead_code)]
+fn has_uppercase(password: &str) -> Result<(), ValidationError> {
+    for c in password.chars() {
+        if c.is_uppercase() {
+            return Ok(());
+        }
+    }
+    Err(ValidationError::new(
+        "The password must include least one uppercase character.\n",
```
da-kami:
```suggestion
        "The password must include at least one uppercase character.\n",
```

rocket-cookie-auth/src/forms.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+#[allow(dead_code)]
+fn has_lowercase(password: &str) -> Result<(), ValidationError> {
+    for c in password.chars() {
+        if c.is_lowercase() {
+            return Ok(());
+        }
+    }
+    Err(ValidationError::new(
+        "The password must include least one uppercase character.\n",
```
da-kami:
```suggestion
        "The password must include at least one uppercase character.\n",
```

da-kami:
❓ Did you have a look at https://docs.rs/rocket_auth/latest/rocket_auth/ - any reason why you did not use this existing crate?

taker/src/main.rs
```diff
@@ -232,20 +257,18 @@ async fn main() -> Result<()> {
...
     let bitcoin_network = network.bitcoin_network();
-    let (ext_priv_key, identities, web_password) = match opts.umbrel_seed {
+    let (ext_priv_key, identities) = match opts.umbrel_seed {
         Some(seed_bytes) => {
             let seed = UmbrelSeed::from(seed_bytes);
             let ext_priv_key = seed.derive_extended_priv_key(bitcoin_network)?;
             let identities = seed.derive_identities();
-            let web_password = opts.password.unwrap_or_else(|| seed.derive_auth_password());
-            (ext_priv_key, identities, web_password)
+            (ext_priv_key, identities)
```
da-kami, line 269: 🤔 Given that we remove `let web_password = opts.password.unwrap_or_else(|| seed.derive_auth_password());` is the password passed in compatible for existing installs?

da-kami (commented):
I had a first look.
I definitely like the idea of improving user authentication, thanks for investing time into it!
I added a few more comments and will see to do another more thorough review once addressed.

bonomat:
> OK, so the expectation for Umbrel is, that we always overwrite it?
> I did read the PR description and am aware that one can overwrite it, I was wondering if you had in mind that Umbrel users can also change it. I suppose that is not planned then and on Umbrel it will just be the `APP_PASSWORD`, correct?

Sorry, I misunderstood the question then. The idea was to keep the same behavior as we have right now: hardcode in docker-compose `--password=APP_PASSWORD`.

bonomat:
> ❓ Did you have a look at [docs.rs/rocket_auth/latest/rocket_auth](https://docs.rs/rocket_auth/latest/rocket_auth/) - any reason why you did not use this existing crate?

I did, and used it as a template 😬 The proposed library did not work out of the box because: 
- it hardcodes a user which requires a username/email and password: there is no reason why we would want that. We could always hardcode a dummy email though. Signup and login forms are bound to this. 
- the user cannot not be extended have an _expiry_ or similar, i.e. when not passing in `--password` we need to randomly generate a password for the user. This is bad because the user needs read it from the logs. With the proposed approach in this PR a user who does not pass in `--password` will get a default password `weareallsatoshi` and needs to change it on the first login.

rocket-cookie-auth/src/user.rs
bonomat: Interesting idea. How would that look like?

bonomat: It is actually, aragon the library is taking care of this. The `password` might look like this: `$argon2i$v=19$m=65536,t=3,p=1$YOtX2//7NoD/owm8RZ8llw==$fPn4sPgkFAuBJo3M3UzcGss3dJysxLJdPdvojRF20ZE=` which will include all the bits and pieces needed. 

Checkout this [stackoverflow answer](https://security.stackexchange.com/questions/222744/which-part-of-this-encoded-argon2-hash-is-the-salt)

rocket-cookie-auth/src/user.rs
bonomat: Can you explain why?

taker/src/main.rs
bonomat: Do you mean the password  we pass in using `--password` or the one generated from the seed? 
The former one is compatible, the latter one is obsolete.

bonomat:
Rebased

taker/src/main.rs
da-kami: I missed the part where we update the password in the db whenever it is passed with the parameter `--password` (line 388). To recap: If a password is provide we save it in the db. If no password provided we set a default on first startup an then ask the user to change it. That makes sense.

taker/src/routes.rs
```diff
@@ -64,8 +67,12 @@ pub async fn feed(
     rx_maker_status: &State<watch::Receiver<ConnectionStatus>>,
     rx_maker_identity: &State<watch::Receiver<Option<identify::PeerInfo>>>,
     identity_info: &State<IdentityInfo>,
-    _auth: Authenticated,
+    user: User,
 ) -> EventStream![] {
+    if user.expired {
+        // TODO fail this call. But how?
```
da-kami: 🤔 What should the behavior of this fail be?

I don't think we want to blow up the `EventStream`, but we would like to report back to the user that there is a problem. Given that the daemon is pushing information to the consumer (the UI), to solve this we would have to introduce an error stream. Or we change constraints around the API; https://github.com/itchysats/itchysats/issues/2906 might help but would not solve this problem completely.

rocket-cookie-auth/src/session.rs
```diff
@@ -0,0 +1,113 @@
...
+            .get(&id)
+            .context("Could not acquire lock on session store")?;
+        Ok(Some(key.secret.clone()))
+    }
+
+    fn clear_all(&self) {
+        self.clear();
+    }
+
+    fn clear_expired(&self) {
```
da-kami: 🤔 This function is never called - given it's functionality I would have expected that it *should* be called in a case where the session is expired. Is there something missing here?

rocket-cookie-auth/src/session.rs
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    fn get(&self, id: u32) -> Result<Option<String>> {
+        let key = self
+            .get(&id)
+            .context("Could not acquire lock on session store")?;
+        Ok(Some(key.secret.clone()))
+    }
+
+    fn clear_all(&self) {
```
da-kami: This function is not used.

da-kami (approved):
I did not find anything that would be blocking - I'd love to see this go in!
More eyes on will be better :)

Please have a look at the comments concerning the session expiry - I don't think this is critical, but I wonder why we have trait functions that are not used.

taker/src/routes.rs
da-kami: ❓ The current behavior would be that the user still receives information even if the cookie is already expired, correct?
Given that the user then cannot trigger any actions (because those endpoints should be guarded) that is acceptable.

rocket-cookie-auth/src/user.rs
Restioson: The process of encoding a hash is a CPU bound task, so it should not be run as-is on the runtime thread. I've used argon2 for a similar purpose before (though for validating session authentication tokens), and verifying them can take up to 11ms. See this example:
image

Therefore, to avoid blocking the runtime thread, it is best to use `spawn_blocking` or `block_in_place` here.

taker/src/routes.rs
```diff
@@ -64,8 +67,12 @@ pub async fn feed(
...
+        tracing::error!("User should not have access if expired password");
```
holzeis: expired password? wouldn't expired session make more sense?

rocket-cookie-auth/src/user.rs
Restioson: Neat! I assume in future if we want to change from argon2 to another hash (for whatever future reason), that encoded form could also include some marker of how it was encoded.

taker/src/routes.rs
holzeis: I guess we should implement a timeout on the frontend which is logging the user automatically off if the session expired. Also I would implement some kind of refresh session upon the users activity.

Generally I would also not go for an auth cookie thats valid for 7 days (I know its different here because if we run frontend and backend in the home network - however we are only assuming that the deployment is like that)

holzeis (approved):
Definitely an improvement in comparison to basic auth. However, why did you go for cookies instead of JWT?

I guess secure cookies is not an option? https://api.rocket.rs/v0.4/rocket/http/enum.Cookies.html#example-1

There are still quite some todos in the code. I did not point each and everyone out, but are your going to work on them still in this pull request?

taker/src/main.rs
bonomat: Yes, exactly :)

taker/src/routes.rs
bonomat: > 🤔 What should the behavior of this fail be?

I think that this function returns mission critical information and should require authentication. With http best practices in mind, ideally this function returns `HTTP::unauthorized` or `forbidden` but I was not able to return this http error type here.

taker/src/routes.rs
bonomat: That's different. The idea of the expired password is to force the user to change the password. This is used for the default password.

bonomat:
> Definitely an improvement in comparison to basic auth. However, why did you go for cookies instead of JWT?
> 

No particular reason besides that I did not know about JWT 🙈.

> I guess secure cookies is not an option? [api.rocket.rs/v0.4/rocket/http/enum.Cookies.html#example-1](https://api.rocket.rs/v0.4/rocket/http/enum.Cookies.html#example-1)

Do you mean `private cookies`? That's what I'm using: https://github.com/itchysats/itchysats/blob/cabad66a14c700435777379293a05e27f53721d8/rocket-cookie-auth/src/auth.rs#L73

> 
> There are still quite some todos in the code. I did not point each and everyone out, but are your going to work on them still in this pull request?

Most of them I'll clean up. [This one](https://github.com/itchysats/itchysats/pull/2903#discussion_r969303995) I did not know how to solve and if no one has an immediate idea I would create a followup ticket. Maybe someone smarter than me has an idea.

taker/src/routes.rs
bonomat: > Generally I would also not go for an auth cookie thats valid for 7 days (I know its different here because if we run frontend and backend in the home network - however we are only assuming that the deployment is like that)

@holzeis : What expiry would you go for? I'm happy to change it.

rocket-cookie-auth/src/user.rs
bonomat: I failed using these function due to do some lifetime issues. :/

Restioson: What were the issues? Might need to clone or move in the password

taker/src/routes.rs
holzeis: @bonomat typically you go for a few minutes to reduce the risk of accepting a leaked token / cookie. To keep the user session alive you can use silent authentications or refresh tokens..

Typically I would go for an auth2.0 authorization code flow with PKCE in that case. That would also allow us to connect to other identity providers if needed in the future. (we could even allow the user to use a social login)

holzeis:
> No particular reason besides that I did not know about JWT 🙈.

Have a look at auth2.0 and openid connect. 😉 

> Do you mean `private cookies`? That's what I'm using:

No I am referring towards letting the cookie only be transported via https connections.

model/src/lib.rs
```diff
@@ -1042,6 +1042,14 @@ pub struct Lock {
     pub dlc_vout: Vout,
 }
 
+pub struct User {
```
luckysori: 🔧  Should this really go in the `model` crate? Login details should be specific to an application.

bonomat:
> Have a look at auth2.0 and openid connect. 😉

I think this goes into the wrong direction: we have a single user system which we want to protect with a password and don't want users to sign-up and create accounts. Does this explanation make sense ? 😅

> No I am referring towards letting the cookie only be transported via https connections.

Right, that would make sense. It's just hard to achieve with our current setup because then each itchysats user would need to have (or create) (self-signed) certificates. I guess this is best to be solved by the platform you run on, e.g. Umbrel and RaspiBlitz are offering a nginx proxy which takes care of HTTPS.

taker/src/routes.rs
bonomat: > @bonomat typically you go for a few minutes to reduce the risk of accepting a leaked token / cookie. To keep the user session alive you can use silent authentications or refresh tokens..

Interesting, I'll create a follow-up ticket on this once this is merged.

sqlite-db/src/user.rs
```diff
@@ -0,0 +1,45 @@
+use crate::models::User;
+use crate::Connection;
+use anyhow::Result;
+
+impl Connection {
+    pub async fn load_user(self) -> Result<Option<model::User>> {
+        let mut conn = self.inner.acquire().await?;
+        // we only want to have max 1 user, hence, we hardcode its ID to 1
```
luckysori: 🔧 Would be better to use a constant for this. Then the comment doesn't need to be duplicated or inline.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ members = [
   "sqlite-db",
   "quiet-spans",
   "vergen-version",
+    "rocket-cookie-auth",
```
luckysori: 🔧 Wrong format.

rocket-cookie-auth/.gitignore
```diff
@@ -0,0 +1 @@
+example-http-cookie-auth.sql
```
luckysori: 🙃 Missing newline.

rocket-cookie-auth/src/user.rs
```diff
@@ -0,0 +1,90 @@
...
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone, Hash, PartialOrd, Ord)]
+pub struct User {
+    pub id: u32,
+    #[serde(skip_serializing)]
+    pub(crate) password: String,
+    pub auth_key: String,
+    pub expired: bool,
+}
+
+impl From<model::User> for User {
```
luckysori: 🔧 Bringing in the whole of `model` just for this `From` implementation is not great.

rocket-cookie-auth/examples/example.rs
```diff
@@ -0,0 +1,74 @@
+use anyhow::Result;
```
luckysori: ❓ How much of this code from `rocket-auth` is actually needed? 

As an example, I'd normally be very happy to add examples, but here we only want to modify the original library for a very particular purpose. I would try to keep this new crate as small as possible.

rocket-cookie-auth/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#[macro_use]
```
luckysori, line 1: 🙃 It might be useful to document where this crate originally comes from, in case we find a bug or need to modify it in the future.

rocket-cookie-auth/src/users.rs
```diff
@@ -0,0 +1,90 @@
+use crate::auth::rand_string;
+use crate::error::Error;
+use crate::forms::Login;
+use crate::session::Session;
+use crate::session::SessionManager;
+use crate::user::verify_password;
+use crate::user::User;
+use anyhow::Context;
+use anyhow::Result;
+use sqlite_db::Connection;
```
luckysori: 🙃 Similar to the `model` crate, it would be great if we didn't need to depend on `sqlite-db` here. A trait could be used instead of the concrete type `sqlite_db::Connection`.

rocket-cookie-auth/templates/base.html.tera
```diff
@@ -0,0 +1,34 @@
+<html lang="en">
```
luckysori, line 1: ❓ Are these files needed?

shared-bin/src/fairings.rs
```diff
@@ -33,21 +33,8 @@ pub fn log_requests() -> impl Fairing {
...
-                    tracing::warn!("Username and password not configured correctly");
-                    return;
-                }
-            };
-
             let http_endpoint = format!(
-                "http://{}:{}@{}:{}",
-                username,
-                password,
+                "http://{}:{}",
```
luckysori, line 40: ❓ How do these changes affect the maker? This is code used for both binaries, right?

luckysori: ❓ Could we just replace `basicauth` with `cookie-auth` for the maker too? It's annoying to maintain two different ways of doing the same thing if one is preferred.

taker/src/main.rs
```diff
@@ -51,6 +53,29 @@ const TESTNET_MAKER: &str = "testnet.itchysats.network:10000";
...
+pub struct Password(String);
+
+impl From<[u8; 32]> for Password {
+    fn from(bytes: [u8; 32]) -> Self {
+        Self(hex::encode(bytes))
+    }
+}
+
+impl std::str::FromStr for Password {
+    type Err = Void;
```
luckysori: ❓ Does `std::convert::Infallible` not work? https://doc.rust-lang.org/std/convert/enum.Infallible.html

sqlite-db/migrations/20220904225056_add-user-db.sql
```diff
@@ -0,0 +1,15 @@
+-- Add migration script here
+CREATE TABLE IF NOT EXISTS frontend_users (
```
luckysori: 🔧 This table name is confusing to me. Why does it not apply to other users? Surely if you use the taker CLI the login details will be stored here too. If so, I propose something like `login_details` or `login_details_taker`.

taker/src/routes.rs
```diff
@@ -64,8 +68,12 @@ pub async fn feed(
     rx_maker_status: &State<watch::Receiver<ConnectionStatus>>,
     rx_maker_identity: &State<watch::Receiver<Option<identify::PeerInfo>>>,
     identity_info: &State<IdentityInfo>,
-    _auth: Authenticated,
+    user: User,
 ) -> EventStream![] {
+    if user.expired {
```
luckysori: ❓ What does it mean for the user to be expired? Originally, I thought it was just a way of telling the user that they need to update their password the first time they use the app. If that's the case, why should that block them from getting new events? Having an unsafe, default password doesn't seem to have anything to do with being able to use the app.

Or are there other ways in which this field could be `true`?

taker/src/routes.rs
```diff
@@ -159,12 +167,17 @@ pub struct CfdOrderRequest {
...
 pub async fn post_order_request(
     cfd_order_request: Json<CfdOrderRequest>,
     taker: &State<Taker>,
-    _auth: Authenticated,
+    user: User,
 ) -> Result<(), HttpApiProblem> {
+    if user.expired {
+        return Err(HttpApiProblem::new(StatusCode::FORBIDDEN)
+            .title("Password expired")
+            .detail("Please change your password to continue"));
```
luckysori: 🤔 I'm not sure I follow the logic. Why should an expired password stop you from using the app? A wrong password, sure. But, as a user, I've come to expect to be able to use accounts even if they're nagging me to change my password for whatever reason.

Also, having to add all these expiry checks on every endpoint is not great. If for some reason we want to stop the user from using the app because their password has expired, shouldn't we just _not_ mount any of these routes?

luckysori (commented):
I have a few suggestions and questions and one thing that I don't understand: why do we want to block the app until the user has chosen a new password? That's not very friendly UX and it currently adds annoying code to the taker routes.

Cargo.toml
luckysori: Resolved in a latter commit.

model/src/lib.rs
bonomat: Fair enough. Initially I didn't have it but added it because it felt more _complete_.

rocket-cookie-auth/examples/example.rs
bonomat: I don't understand the question. :)

rocket-cookie-auth/src/users.rs
bonomat: Do you propose to turn the dependency around: 
- have a trait (e.g. `LoadUser`) in `rocket-cookie-auth` and in `sqlite_db` we depend on this and implement the trait for `Connection`?

rocket-cookie-auth/templates/base.html.tera
bonomat: Yes, they are needed for the example.

shared-bin/src/fairings.rs
bonomat: > ❓ How do these changes affect the maker? This is code used for both binaries, right?

Good point!

> ❓ Could we just replace basicauth with cookie-auth for the maker too? It's annoying to maintain two different ways of doing the same thing if one is preferred.

Yes, I'll do this after the changeset is mostly approved as the rest is copy&paste for the maker.

taker/src/main.rs
bonomat: Haven't considered it but seems to work.

sqlite-db/migrations/20220904225056_add-user-db.sql
bonomat: Our cli does not have any auth, but if it would have, we probably could reuse the same user... happy to rename it.

taker/src/routes.rs
bonomat: I had to come up with an approach to force the user to change the password on first login. The variable `expired` should indicate that the password is _expired_ and needs to be changed. 

> Having an unsafe, default password doesn't seem to have anything to do with being able to use the app.

I think it does 🤷‍♂️ but I'm happy to remove those checks if you think this is better UX.

taker/src/routes.rs
bonomat: > 🤔 I'm not sure I follow the logic. Why should an expired password stop you from using the app? A wrong password, sure. But, as a user, I've come to expect to be able to use accounts even if they're nagging me to change my password for whatever reason.

As a user yes, but as the app I want to protect you until you have a secure password.

> Also, having to add all these expiry checks on every endpoint is not great. If for some reason we want to stop the user from using the app because their password has expired, shouldn't we just not mount any of these routes?

I'm not sure how we could achieve this in a more general way.

rocket-cookie-auth/src/lib.rs
bonomat: I'll add a readme

model/src/lib.rs
bonomat: Removed again.!

bonomat:
@luckysori : I changed the logic of `expired` slightly and called it `first_login` indicating that the user logged in for the first time. I've also removed the logic in the routes and do only client side check, i.e. the user can login with the _expired_ password but on client side is forced to change the password.

rocket-cookie-auth/examples/example.rs
luckysori: I was wondering if we could trim down the new crate any further.

For instance, I would just remove the example (since it's not meaningfully different from the original) and all the other files related to the example. It's not like we're gonna use this library anywhere else.

rocket-cookie-auth/src/users.rs
luckysori: Yes, although I would also not let `sqlite-db` depend on this new crate! I think the implementation of such a trait should live in `taker` (wherever the new crate is used).

For that you could just wrap the `sqlite_db::Connection`.

taker/src/routes.rs
```diff
@@ -267,12 +271,12 @@ pub struct WithdrawRequest {
...
 
 #[rocket::post("/withdraw", data = "<withdraw_request>")]
-#[instrument(name = "POST /withdraw", skip(taker, _auth), err)]
+#[instrument(name = "POST /withdraw", skip(taker, _user), err)]
 pub async fn post_withdraw_request(
     withdraw_request: Json<WithdrawRequest>,
     taker: &State<Taker>,
     network: &State<Network>,
-    _auth: Authenticated,
+    _user: User,
```
luckysori, line 279: 🔧 I think you can remove `_user: User` from all these routes where it isn't used.

luckysori (approved):
Thank you for addressing my main concern that affected the UX.

The rest can be tackled later/never 😛

rocket-cookie-auth/src/users.rs
bonomat: I'll give this a try

taker/src/routes.rs
bonomat: No, it's needed because this is the guard. The field `user` will only be available if the user is logged in.

luckysori: Oh, interesting. That's not obvious at all. How does that work?

bonomat: To be precise, it's a [`RequestGuard`](https://rocket.rs/v0.4/guide/requests/#request-guards). We implement `FromRequest` for this type here: https://github.com/itchysats/itchysats/blob/82b32f1d6388881bd471243411ee8a3810682070/rocket-cookie-auth/src/user.rs#L23 

It's similar to before with `Authenticate` which implemented `FromRequest` here: https://github.com/itchysats/itchysats/blob/ea3170f2da4fcbbdf9b6152d84cd1ab4bcc3b6d6/rocket-basicauth/src/lib.rs#L81

maker/src/lib.rs
```diff
@@ -14,6 +15,29 @@ pub mod cfd;
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(Self(s.to_owned()))
+    }
+}
+
+impl std::fmt::Display for Password {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
```
klochowicz, line 39: no need for this duplication - you can easily move it to some shared location, e.g. `shared-bin/src/auth.rs`

maker/src/main.rs
```diff
@@ -201,3 +212,31 @@ async fn main() -> Result<()> {
...
+            auth_key: rocket_cookie_auth::NO_AUTH_KEY_SET.to_string(),
+            first_login: user.first_login,
+        }))
+    }
+
+    async fn update_password(&self, password: String) -> Result<()> {
+        self.inner.clone().update_password(password).await?;
+        Ok(())
+    }
+}
```
klochowicz, line 242: this one could be moved to `shared_bin` as well to avoid maintenance burden

klochowicz (approved):
nice one!

Restioson:
Can anyone verify whether the testnet metrics still work after merging this PR? Many seem to be blank for me, but I'm not sure if this is unrelated. 

image

image

holzeis:
same here.. I guess the endpoint is now authenticated, hence the missing data.

PR #2909, opened 2022-09-08
chore: Remove remnants of the legacy network connection in CLI args

klochowicz:
We used to derive libp2p port everywhere to be "legacy+1". Since legacy
connection is fully retired, we can directly specify libp2p ports now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY $BINARY_PATH /usr/bin/binary
 
 VOLUME /data
 
-# HTTP Port, legacy P2P Port and libp2p P2P Port
-EXPOSE 8000 9999 10000
+# HTTP Port and libp2p P2P Port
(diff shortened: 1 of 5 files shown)


holzeis:
🤔 Do we need to adapt the exposed ports in the Dockerfile as well?

https://github.com/itchysats/itchysats/blob/4b67344760c76349c1f2d729f30bf20b3631fabc/Dockerfile#L17

klochowicz:
> 🤔 Do we need to adapt the exposed ports in the Dockerfile as well?
> 
> https://github.com/itchysats/itchysats/blob/4b67344760c76349c1f2d729f30bf20b3631fabc/Dockerfile#L17

thanks for pointing it out, I totally missed the Dockerfile in the crowded repo! fix should be there in a moment :)

bonomat (approved):

PR #2922, opened 2022-09-09
test: for malicious taker trying to settle unrelated order

### LINKED ISSUE #2891: Test that a `peer-id` cannot act on a `Cfd` that does not belong to said `peer-id`
We introduced `verify_peer_id()` but it was never tested.

https://github.com/itchysats/itchysats/blob/master/model/src/cfd.rs#L1446

holzeis:
resolves #2891

I spent some time trying to find the best was to implement that test and I ended up with replicating the state from the takers database to mallorys database. That allowed me to use the existing methods to propose the settlement without changing too much. Another approach I was considering was directly to send the `Settle` event via the libp2p connection - however I didn't find a lean way of doing that.

One drawback of the current implementation is that I had to make the Dlc property on the Cfd public as otherwise I wouldn't have been able to replicate the state from the taker to mallory. side note, its quite odd that the dlc is stored in the events and upon loading mapped to the cfd.

Lastly, it was quite challenging to make well defined assertions about the expected state. One challenge I've noticed is that the changed method on the cfd_feed might get overwritten by newer events - that makes the test potentially flaky. Since all events are stored to the database a better approach could be loading them from the database to ensure nothing has been missed.

Potentially, we could come around some visibility issues (e.g. having to make dlc public on the cfd) by making the connection accessible to the tests. Allowing the test to query a certain state specifically from the database for assertions.

### DIFF
--- a/daemon-tests/tests/main/collaborative_settlement.rs
+++ b/daemon-tests/tests/main/collaborative_settlement.rs
@@ -14,8 +14,12 @@ use daemon_tests::wait_next_state;
 use daemon_tests::Maker;
 use daemon_tests::OpenCfdArgs;
 use daemon_tests::Taker;
+use daemon_tests::TakerConfig;
+use model::Cfd;
+use model::CfdEvent;
 use model::ContractSymbol;
 use model::Contracts;
+use model::EventKind;
 use model::Leverage;
 use model::Position;
 use otel_tests::otel_test;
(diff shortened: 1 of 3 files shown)


daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -87,6 +91,54 @@ async fn maker_accepts_collab_settlement_after_commit_finality() {
...
+        &TakerConfig::default(),
+        maker.identity,
+        maker.connect_addr.clone(),
+    )
+    .await;
+
+    let cfd = taker.db.load_open_cfd::<Cfd>(order_id, ()).await.expect("");
+
+    // copy cfd from alice to mallory to bypass load_open_cfd on the taker side.
+    mallory.db.insert_cfd(&cfd).await.expect("");
```
da-kami:
```suggestion
    mallory.db.insert_cfd(&cfd).await.unwrap();
```

da-kami (commented):
I had a discussion about some of the issues addressed by this test with @holzeis offline, here is a summary from my side:

I think it's great that we come up with more tests for edge cases! It is important to get more assurance on the behavior of the system.
However, I am not sure our current actor test framework is designed to depict tests as proposed by this PR. Our actor tests are quite long running, and this test tests something very specific. 

It might be better to change this test to be on the protocol-actor level rather than creating the complete `ActorSystem`. It is sufficient to assert that the protocol actors handle this scenario correctly. 
We currently don't have tests involving `Cfd` state on the protocol level, but that can be introduced. The hard part is initializing the `Executor` with a database that contains proper test data. I describe one vision for that in https://github.com/itchysats/itchysats/issues/2928

Given that this specific test does not assert on the `Cfd` values we could opt for just using a dummy `Cfd` similar to what we do in the `model::Cfd` tests:

https://github.com/itchysats/itchysats/blob/d7385544e334a82731755aa7b0802640df2bb644/model/src/cfd.rs#L4076-L4352

However, it is currently hard to create a `dummy` `Cfd` outside of the `model` crate. I think a crate that provides dedicated test data would be useful to achieve this - or we should re-think splitting off the model crate and protocol crates because it makes it hard to extend for test purposes.

klochowicz:
Zooming out, it seems that in order to try such an attack mallory would need to have access to the taker's DB... this means that a compromised taker would likely be in greater trouble rather than settling a Cfd for them...

I agree with @da-kami - it seems to be not quite the right abstraction to test such case (although it's positively reinforcing that such an attack would be too cumbersome to be carried!).  At the actor level tests we try to pretty much do black-box testing, without diving into internals.

model/src/cfd.rs
```diff
@@ -2935,6 +2935,22 @@ mod tests {
...
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn given_correct_counterparty_id_then_verify_succeed() {
+        let cfd = Cfd::dummy_maker_short();
+        let peer_id = cfd.counterparty_peer_id.unwrap();
+        let result = cfd.verify_counterparty_peer_id(&peer_id);
+        assert!(result.is_ok());
+    }
```
luckysori, line 2952: 👍 We can definitely keep the unit tests!

daemon-tests/tests/main/collaborative_settlement.rs
```diff
@@ -87,6 +91,54 @@ async fn maker_accepts_collab_settlement_after_commit_finality() {
     wait_next_state!(order_id, maker, taker, CfdState::OpenCommitted);
 }
 
+#[otel_test]
+async fn malicious_taker_tries_to_settle_unrelated_order() {
```
luckysori, line 95: It's not that bad, but it's true that it's not the nicest test.

I think it might be a good idea to write a test for the `collab_settlement` actors instead, but I haven't given it much thought.

luckysori (commented):
I see why we don't want to merge this particular test, but I'm sure it was a good exercise to try to write it at all.

holzeis:
Closing this PR as this is not going to be merged.

PR #2923, opened 2022-09-09
Improve randomness of ports in tests

klochowicz:
We're picking random ports first, and testing whether they are both free no UDP
and TCP instead of just relying on the OS to give us one.

### DIFF
--- a/daemon-tests/Cargo.toml
+++ b/daemon-tests/Cargo.toml
@@ -14,6 +14,7 @@ mockall = "0.11"
 mockall_derive = "0.11"
 model = { path = "../model" }
 otel-tests = { version = "0.1", default-features = false }
+portpicker = "0.1.1"
 quiet-spans = { path = "../quiet-spans" }
 rand = "0.6"
 rust_decimal = "1.26"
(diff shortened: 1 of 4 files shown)


daemon-tests/Cargo.toml
```diff
@@ -14,6 +14,7 @@ mockall = "0.11"
 mockall_derive = "0.11"
 model = { path = "../model" }
 otel-tests = { version = "0.1", default-features = false }
+portpicker = "0.1.1"
```
da-kami, line 17: ❓ The assumption is that `portpicker` is more reliable with picking a port than `TcpListener`?

I guess we can only test this by merging it and observing the behavior, or do you think we can actually test this?

da-kami (approved):
Thanks for driving this forward!

daemon-tests/Cargo.toml
klochowicz: as described in the commit message, the mechanism for choosing the port is different. `TcpListener` just tries to bind to a local socket, which will give back a free port from the OS (but could lead to a race condition, if 2 tests ask at the same time).
`portpicker` starts by randomly choosing ports and checking whether they're free, only falling back to asking the OS to provide a port if it can't guess it :)

klochowicz:
> Thanks for driving this forward!

no worries, thank you for the nudge 🙏

PR #2935, opened 2022-09-13
Bump maia to latest HEAD and bdk to 0.21

### LINKED ISSUE #2917: Starting maker or taker deamon with testnet seeds fails with ChecksumMismatch on bdk 0.21
The bdk bump to version 0.21 broke the testnet maker and taker seed files. Deleting them and letting itchysats recreate them works just fine, though.

```
$ cargo run --bin maker -- --password dev testnet
    Finished dev [unoptimized + debuginfo] target(s) in 0.51s
     Running `target/debug/maker --password dev testnet`
2022-09-09 07:39:26  INFO shared_bin::logger: Initialized logger
2022-09-09 07:39:26  INFO maker: Running version: 0.6.1-83-gac3c438
2022-09-09 07:39:26  INFO maker: CFDs created with this release will settle after 24 hours
Error: ChecksumMismatch
```

bonomat:
Includes latest BDK and drops backwards compatibility.

BDK changed balance API in 0.21 and now you need to explicitly specify which
balance one you want (I opted for all spendable utxos).
It also includes a breaking change for the wallet derivation paths:
From now on keys for testnet/regtest are derived from the path `m/84'/1'/0` while mainnet they are derived from `m/84'/0'/0`. Before that testnet/regtest keys were derived from `m/84'/0'/0` as well.

The second commit includes a fix for #2917

With bdk>=0.20 the wallets are derived from a different derivation paths for testnet/regtest. This may lead to checksum errors if you had a db pre initialized. In order to not ask the user to delete their db we derive the wallet name from the key.

Resolves #2917

Note: this is a breaking change for our testnet setup. Once merged we should withdraw all funds from the old versions or simply fund them again using faucets. 

Note2: I opted for the breaking change because I think the testnet setup is not critical. No funds are in danger on mainnet as the derivation path on mainnet has not changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Breaking change: with bdk>0.20 key derivations will take into account the network, i.e. for mainnet it uses `m/84'/0'/0'` and for testnet/regtest it uses `m/84'/1'/0'`. This was caused by https://github.com/bitcoindevkit/bdk/pull/585. If you want to withdraw your testnet funds make sure to use the withdraw functionality prior upgrading to a different wallet or return them to the faucet. Existing mainnet wallets are not affected.
+
 ## [0.6.1] - 2022-09-01
 
 ### Fixed
--- a/daemon/src/wallet.rs
+++ b/daemon/src/wallet.rs
@@ -1,3 +1,4 @@
+use crate::bitcoin::secp256k1::Secp256k1;
 use anyhow::bail;
 use anyhow::ensure;
 use anyhow::Context;
@@ -21,6 +22,7 @@ use bdk::electrum_client;
 use bdk::electrum_client::ElectrumApi;
 use bdk::sled;
 use bdk::wallet::tx_builder::TxOrdering;
+use bdk::wallet::wallet_name_from_descriptor;
 use bdk::wallet::AddressIndex;
 use bdk::FeeRate;
 use bdk::KeychainKind;
@@ -107,7 +109,6 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
         electrum_rpc_url: &str,
         ext_priv_key: ExtendedPrivKey,
         db_path: PathBuf,
-        wallet_name: String,
     ) -> Result<(xtra::Address<Self>, watch::Receiver<Option<WalletInfo>>)> {
         let client = electrum_client::Client::new(electrum_rpc_url)
             .context("Failed to initialize Electrum RPC client")?;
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An issue where the taker logs were spammed with unnecessary error message on auto-rollover.
 - An issue where metrics on connected peers where not reported correctly.
 
+### Changed
+
+- Breaking change: with bdk>0.20 key derivations will take into account the network, i.e. for mainnet it uses `m/84'/0'/0'` and for testnet/regtest it uses `m/84'/1'/0'`. This was caused by https://github.com/bitcoindevkit/bdk/pull/585.
+
```
klochowicz: maybe we can expand in lay terms that users with testnet wallets should move their funds off ItchySats wallet before upgrading or return them to the faucet.

daemon/src/wallet.rs
```diff
@@ -117,6 +118,13 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
             "Wallet seed and Electrum RPC client on different networks."
         );
 
+        let wallet_name = wallet_name_from_descriptor(
```
klochowicz, line 121: 🤔 do we lose the ability to run maker and taker in the same directory, or is the result of this function a unique string? previously used `maker-wallet` was covering this by being hard-coded

bonomat: As long as we don't use the same key (which we shouldn't anyway) they can run in the same directory.

holzeis (approved):
LGTM 👍

luckysori (approved):
Nice! Two minor git-related things:

- Since you follow conventional commits for the second commit, can you do the same for the first one? It's more important for the first one because you can then say something like `feat!: Bump bdk` and it's immediately obvious that it includes a breaking change.
- Should we just squash the two commits? The first one has a known bug that is fixed in the second one. It's fine either way, because it is nice to have separate commit messages. Your call.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- Breaking change: with bdk>0.20 key derivations will take into account the network, i.e. for mainnet it uses `m/84'/0'/0'` and for testnet/regtest it uses `m/84'/1'/0'`. This was caused by https://github.com/bitcoindevkit/bdk/pull/585. If you want to withdraw your testnet funds make sure to use the withdraw functionality prior upgrading to a different wallet or return them to the faucet.
```
klochowicz: nit: so that people are not worried about upgrading ;D

```suggestion
- Breaking change: with bdk>0.20 key derivations will take into account the network, i.e. for mainnet it uses `m/84'/0'/0'` and for testnet/regtest it uses `m/84'/1'/0'`. This was caused by https://github.com/bitcoindevkit/bdk/pull/585. If you want to withdraw your testnet funds make sure to use the withdraw functionality prior upgrading to a different wallet or return them to the faucet. Existing mainnet wallets are not affected.
```

klochowicz (approved):
LGTM!

bonomat:
> Should we just squash the two commits? The first one has a known bug that is fixed in the second one. It's fine either way, because it is nice to have separate commit messages. Your call.

I'll leave it. I think it's nice to have these two commits separately and it is not a _real_ bug (:

PR #2936, opened 2022-09-13
fix: added contract symbol and position to selection context

### LINKED ISSUE #2905: Contract symbol selection is lost after clicking the trade or wallet button
Image

holzeis:
Introduces a selection context which stores the state of selection
 - contract symbol
 - position (long / short) per contract symbol

resolves #2905

### DIFF
--- a/taker-frontend/src/App.tsx
+++ b/taker-frontend/src/App.tsx
@@ -4,7 +4,7 @@ import isBetween from "dayjs/plugin/isBetween";
 import relativeTime from "dayjs/plugin/relativeTime";
 import utc from "dayjs/plugin/utc";
 import * as React from "react";
-import { useEffect, useState } from "react";
+import { createContext, useContext, useEffect, useState } from "react";
 import { Route, Routes, useLocation, useNavigate } from "react-router-dom";
 import useWebSocket from "react-use-websocket";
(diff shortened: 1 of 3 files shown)


taker-frontend/src/App.tsx
```diff
@@ -62,24 +62,30 @@ const BTC_USD_SHORT_EVENT = "btcusd_short_offer";
...
+export interface Selection {
+    symbol: Symbol;
+    position: Map<Symbol, string>;
+}
+
+const defaultSelection: Selection = {
+    position: new Map().set(Symbol.btcusd, "long").set(Symbol.ethusd, "long"),
+    symbol: Symbol.btcusd,
+};
+export const SelectionContext = createContext(defaultSelection);
```
bonomat, line 74: What do you think of giving this a better name? E.g. `SelectedSymbolContext`? Same goes for the interface `Selection` and the variable names.

bonomat (approved):
Nice, looks good.

taker-frontend/src/App.tsx
holzeis: It's not just the symbol but also the position 😉.

bonomat: Still, `selection` does not say much :)

PR #2943, opened 2022-09-14
chore: Go back to macos-latest on CI

klochowicz:
We switched to  macos-12, as it was a bit more reliable at the time in
successfully running tests. This is not the case anymore (it got so
unreliable we had to turn it off in bors). Moreover, we started hitting concurrency
limits for the agent pool on macos-12 jobs; hopefully macos-latest has more
pools as it's the officially endorsed one.


klochowicz:
closing this as we can't build on macos-11 for some reason, and it can't be easily overcome.

PR #2944, opened 2022-09-14
Replace infinitely growing `Tasks` from some actors 

luckysori:
I think this might be the main cause of the unbounded memory growth we see in production. We somehow forgot to replace these.

Originally I thought we would want to do the same for `Endpoint`, but I noticed two things:

- We want to be able to explicitly drop these tasks when a peer disconnects.
- We only insert into the `controls: HashMap<PeerId, (yamux::Control, Tasks)>` map _once per peer_, so this one cannot grow out of control.

We also use `Tasks` in the actor system constructors, but we only insert a handful of times as the actor systems is started up.

Finally, we use `Tasks` in some tests, but that obviously doesn't affect production.


klochowicz (approved):
nice find!

da-kami (approved):
Removing `AddressMap` should ensure that we are not using it anymore 😃
Thanks!

Restioson (approved):
I just skimmed through this but looks good to me!

PR #2955, opened 2022-09-15
feat: Extend Environment enum to new platforms

klochowicz:
Extend Environment enums for the platforms we're about to integrate on.

### DIFF
--- a/daemon/src/identify/protocol.rs
+++ b/daemon/src/identify/protocol.rs
@@ -110,6 +110,9 @@ where
 pub enum Environment {
     Umbrel,
     RaspiBlitz,
+    Citadel,
+    Start9,
+    MyNode,
     Docker,
     Binary,
     Test,
@@ -122,6 +125,9 @@ impl From<crate::Environment> for Environment {
         match environment {
             crate::Environment::Umbrel => Environment::Umbrel,
             crate::Environment::RaspiBlitz => Environment::RaspiBlitz,
+            crate::Environment::Citadel => Environment::Citadel,
+            crate::Environment::Start9 => Environment::Start9,
+            crate::Environment::MyNode => Environment::MyNode,
             crate::Environment::Docker => Environment::Docker,
             crate::Environment::Binary => Environment::Binary,
             crate::Environment::Test => Environment::Test,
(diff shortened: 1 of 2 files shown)


daemon/src/identify/protocol.rs
```diff
@@ -122,6 +125,9 @@ impl From<crate::Environment> for Environment {
         match environment {
             crate::Environment::Umbrel => Environment::Umbrel,
             crate::Environment::RaspiBlitz => Environment::RaspiBlitz,
+            crate::Environment::Citadel => Environment::Citadel,
```
bonomat, line 128: Proposal: change it to a String field which can be passed in. That way it is _dynamic_ and we don't have to extend this enum and release a new version.

klochowicz: I like this, as I never understood why we went with an enum in the first place. I will re-work this PR.

klochowicz: turns out that this will break the Identify protocol. I don't think it's worth bumping the protocol just for this reason :/

da-kami: This should not break anything, because the taker will always receive an `Unknown` on the environment from the maker. It breaks it for the maker only and that is acceptable.

da-kami (approved):
I think it can go on like this and we can decide to make it dynamic afterwards. 
This should be rolled out soon given that we started rolling out on other platforms, and the `taker` has to set this in order to make sense. Until we roll this out + update the environment settings on each platform will not be reported correctly.

daemon/src/identify/protocol.rs
klochowicz: perhaps I should have been more clear: 
- this changeset does not break anything,
- I tried to re-work this changeset into @bonomat 's suggestion and it will break, because suddenly we will pass a `String` instead of `Option<Enum>` in the identify protocol.

klochowicz:
> I think it can go on like this and we can decide to make it dynamic afterwards. This should be rolled out soon given that we started rolling out on other platforms, and the `taker` has to set this in order to make sense. Until we roll this out + update the environment settings on each platform will not be reported correctly.

I will create a refactor ticket to just pass it dynamically, on the next opportunity to bump the protocol.

klochowicz:
Going ahead with  merging this as @bonomat suggestion has been captured in a dedicated ticket

bors r+

daemon/src/identify/protocol.rs
bonomat: > I tried to re-work this changeset into @bonomat 's suggestion and it will break, because suddenly we will pass a String instead of Option<Enum> in the identify protocol.

Are you sure this would have been a breaking change? 
When de/serialize it `Option<String>` should have the same effect as `Option<Enum>`.

Issue #2889, opened 2022-09-06
Wrap ItchySats into electron


klochowicz:
Wrap our application in an [electron](https://www.electronjs.org/) application which runs our ItchySats daemon as an [integrated binary](https://stackoverflow.com/questions/33152533/bundling-precompiled-binary-into-electron-app).

In theory this should give us a working solution without needing to change anything in our project setup.

klochowicz:
one thing to consider is that we should really have some preferences pane where the user can tweak values normally input via CLI... and maybe initial screen if no config has been provided.

PR #2956, opened 2022-09-15
feat: Add headless mode for the binaries

### LINKED ISSUE #2544: Allow starting maker and taker in headless mode (ie. without GUI)
[...]
I agree with @bonomat that this behavior can be annoying and I think a flag to disable automatically opening the UI would be useful for test setups for maker and taker.
[...]

_Originally posted by @da-kami in https://github.com/itchysats/itchysats/pull/2495#pullrequestreview-1049803667_

klochowicz:
By default, the UI is opened at startup in system webbrowser. For users who do
not wish this to happen, a new parameter called `--headless` got introduced.

### DIFF
--- a/maker/src/lib.rs
+++ b/maker/src/lib.rs
@@ -50,6 +50,10 @@ pub struct Opts {
     #[clap(long)]
     pub verbose_spans: bool,
 
+    /// If enabled, browser UI is not automatically launched at startup.
+    #[clap(long)]
+    pub headless: bool,
+
     /// OTEL collector endpoint address
     ///
     /// If not specified it defaults to the local collector endpoint.
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -191,7 +191,7 @@ async fn main() -> Result<()> {
         .register("/", default_catchers())
         .attach(fairings::log_launch())
         .attach(fairings::log_requests())
-        .attach(fairings::ui_browser_launch())
+        .attach(fairings::ui_browser_launch(!opts.headless))
         .launch()
         .await?;
 
--- a/shared-bin/src/fairings.rs
+++ b/shared-bin/src/fairings.rs
@@ -30,7 +30,10 @@ pub fn log_requests() -> impl Fairing {
 }
 
 /// Attach this fairing to enable loading the UI in the system default browser
-pub fn ui_browser_launch() -> impl Fairing {
+///
+/// Passing `true` opens browser at launch, passing `false` logs the link to
+/// open it manually.
+pub fn ui_browser_launch(open_browser: bool) -> impl Fairing {
     AdHoc::on_liftoff("ui browser launch", move |rocket| {
         Box::pin(async move {
             let (username, password) = match (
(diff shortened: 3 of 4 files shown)


maker/src/lib.rs
```diff
@@ -50,6 +50,10 @@ pub struct Opts {
     #[clap(long)]
     pub verbose_spans: bool,
 
+    /// If enabled, browser UI is not automatically launched at startup.
+    #[clap(long)]
+    pub headless: bool,
```
bonomat, line 55: I propose to do it the otherway around. Pass in `--browser` to open the default browser. Our default platforms are servers and there it does not make sense to open a browser automatically.

bonomat (approved):

maker/src/lib.rs
klochowicz: I went with this as this is the conventional flag name used for this purposes, even browsers themselves have a flag with thhis nome.  Moveover, on a server this will not open a browser anyway, so there's no need to specify the flag.

As far as I remember, the feature of opening the browser was meant for people running the binary and who had less experience with CLI.

bonomat: > As far as I remember, the feature of opening the browser was meant for people running the binary and who had less experience with CLI.

We never had anyone doing so and our efforts in https://github.com/itchysats/itchysats/issues/2889 would make this feature obsolete.

luckysori: I would also prefer it if we didn't open a browser window by default. Maybe just call the flag `disable_headless` or `auto_open_browser`. Even if the name isn't as intuitive.

klochowicz: again, this is the standard terminology, you can see it in many programs (even likes of Chrome). There's even a wikipedia entry: https://en.wikipedia.org/wiki/Headless_software 

I don't see a reason why we should not enable GUI for casual users starting the binary. I have provided this feature for devs who start things multiple times, as I've heard that some of the devs would prefer to be able to turn it off. this PR or ticket are not about changing the default UX.

luckysori: > again, this is the standard terminology, you can see it in many programs (even likes of Chrome). There's even a wikipedia entry: https://en.wikipedia.org/wiki/Headless_software

I don't think it's a convincing argument to enable behaviour by default just because there is well-known terminology around it. The suggestion to rename the flag (and go against said terminology) is so that we don't have to enable the behaviour by default, not because I think the name is better.

> I don't see a reason why we should not enable GUI for casual users starting the binary. I have provided this feature for devs who start things multiple times, as I've heard that some of the devs would prefer to be able to turn it off. this PR or ticket are not about changing the default UX.

Sure, I can see why someone could benefit from this. We can keep the behaviour.

klochowicz: I'm sorry, I misunderstood you; I thought we were just talking about the name.

The reason to keep the behaviour is mostly "principle of least surprise" - typically if you open something on desktop, then you want to start the UI if there is one.

bonomat: > The reason to keep the behaviour is mostly "principle of least surprise" - typically if you open something on desktop, then you want to start the UI if there is one.

I agree with this statement but we haven't seen any users using it on desktop.

klochowicz:
I'm going ahead and merging this, as I consider this an improvement (at least gives out an option to not start the UI for ppl who don't want it). in any case, changing the defaults will be easier with this in if we ever decide to.

PR #2960, opened 2022-09-15
chore: Rename umbrel seed to app seed

klochowicz:
The mechanism of passing external app seed is not specific to Umbrel, it is used
in other distributions as well. It is even called APP_SEED on Umbrel itself.

Note: this will require rolling out on umbrel with the next ItchySats version bump.

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -10,9 +10,9 @@ use daemon::libp2p_utils::create_connect_tcp_multiaddr;
 use daemon::monitor;
 use daemon::oracle;
 use daemon::projection;
+use daemon::seed::AppSeed;
 use daemon::seed::RandomSeed;
 use daemon::seed::Seed;
-use daemon::seed::UmbrelSeed;
 use daemon::wallet;
 use daemon::wallet::TAKER_WALLET_ID;
 use daemon::Environment;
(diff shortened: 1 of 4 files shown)


taker/src/main.rs
```diff
@@ -126,10 +126,10 @@ struct Opts {
     #[clap(subcommand)]
     network: Option<Network>,
 
-    #[clap(short, long, parse(try_from_str = parse_umbrel_seed))]
-    umbrel_seed: Option<[u8; 32]>,
+    #[clap(short, long, parse(try_from_str = parse_app_seed))]
```
bonomat, line 129: 📯
imho this asks for a changelog update

taker/src/main.rs
```diff
@@ -126,10 +126,10 @@ struct Opts {
...
+    app_seed: Option<[u8; 32]>,
```
bonomat, line 130: Is the term `app` redundant?

bonomat (approved):

taker/src/main.rs
klochowicz: thank you for reminder, it is indeed easier to update it as-we-go.

taker/src/main.rs
klochowicz: not really, we have more than one seed used in our app (and this is the name used in some of these distros)

bonomat: If this is true, then `app` is not precise enough. What other seeds do we have?

klochowicz: we have our own `taker_seed` and `maker_seed` files. We don't want people to confuse that. 

I'm happy to take any other suggestion; I just used the same name that's available as an env var on Umbrel and Citadel (and that's what it's referred to).

bonomat: That's the same thing... given you are in the `taker` binary it might be enough to call `seed`, because `taker` + `seed` = `taker_seed`, which is the same naming we use in the filesystem :) 

Eitherway, no harm done. 🙃

PR #2967, opened 2022-09-16
Wrap itchysats into electron

### LINKED ISSUE #2889: Wrap ItchySats into electron

holzeis:
This PR exposes the taker daemon as library to be used by the taker-electron project, which embeds the taker into a NodeJS library which can be used by electron.

The following topics are not addressed by this PR.

- Publishing itchysats to the apple AppStore
- Adapting the build pipeline to automatically wrap the latest itchysats taker daemon into electron and make it downloadable from github. (However, electron-forge has been integrated - thus you can build the electron app based on your host platform - currently `yarn make` will build the architectures x64 and arm64.
- Enhancing the about section.
- Testing electron package build on other platforms like windows or linux.
- Electrons auto update feature
- Piping itchysats taker daemon logs to a file (stdout is not accessible on the packaged electron app)

resolves #2889


bonomat:
BlandGloomyAntelopegroundsquirrel-max-1mb

taker-electron/src/index.html
```diff
@@ -0,0 +1,75 @@
...
+        });
+    </script>
+</head>
+<body>
+    <div class="content">
+        <img src="logo.svg" class="logo" />
+        <p id="message">Starting Itchy Sats ...</p>
+    </div>
+</body>
+</html>
```
Restioson: nit: no newline here

taker-electron/src/lib.rs
```diff
@@ -0,0 +1,58 @@
+use neon::prelude::*;
+use once_cell::sync::OnceCell;
+use taker::Opts;
+use tokio::runtime::Runtime;
+
+// Return a global tokio runtime or create one if it doesn't exist.
+// Throws a JavaScript exception if the `Runtime` fails to create.
+fn runtime<'a, C: Context<'a>>(cx: &mut C) -> NeonResult<&'static Runtime> {
```
Restioson, line 8: Am I correct in saying that this is only called once, on line 17 in the `wrap` function? Also, does tokio not already have its own way of handling what the global runtime is? I see that this function is defined in a neon example [here](https://github.com/neon-bindings/examples/blob/855eb2d52d650afa489b6b4724095c5fdfd5c9ad/examples/tokio-fetch/src/lib.rs#L14), but I am not sure why it's done this way.

taker-electron/src/index.html
holzeis: you mean you want a newline here? why? 😅

taker-electron/src/lib.rs
holzeis: The way I've understand it, It is the rust way of implementing a singleton pattern. Have a look 
[here](https://github.com/neon-bindings/examples/tree/855eb2d52d650afa489b6b4724095c5fdfd5c9ad/examples/tokio-fetch#executor) for more details on the boilerplate.

taker-electron/src/index.html
```diff
@@ -0,0 +1,96 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Itchy Sats</title>
```
bonomat: `Itchy Sats` vs `Itchysats` vs `ItchySats` vs `ItChYsAtS` 🤪 ... Let's roll with `ItchySats` please :)

taker-electron/src/index.html
bonomat: Because of reasons :) https://stackoverflow.com/a/729795

taker/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+                Some(PeerId::from_str(TESTNET_MAKER_PEER_ID).expect("maker peer id"));
+            opts.network = Some(Network::Testnet {
+                electrum: "ssl://blockstream.info:993".to_string(),
+                withdraw: None,
+            });
+        }
+
+        opts.data_dir = Some(PathBuf::from(data_dir));
+        if !password.is_empty() {
+            opts.password = Some(Password::from_str(&password).expect("password"));
```
bonomat: 🙃
I think this deserve a better error message then `password` :)

taker/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+        let mut opts = Opts::default();
+
+        // signet, regtest are not implemented, since there is no maker available.
+        if network == "testnet" {
+            opts.maker = Some(TESTNET_MAKER.to_string());
+            opts.maker_id = Some(parse_x25519_pubkey(TESTNET_MAKER_ID).expect("public key"));
+            opts.maker_peer_id =
+                Some(PeerId::from_str(TESTNET_MAKER_PEER_ID).expect("maker peer id"));
+            opts.network = Some(Network::Testnet {
+                electrum: "ssl://blockstream.info:993".to_string(),
```
bonomat: Don't we have (had?) constants for this?

taker/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+        Opts::parse()
+    }
+
+    pub fn new(network: String, password: String, data_dir: String) -> Opts {
+        // mainnet is the default setting nothing to do if network is mainnet.
+        let mut opts = Opts::default();
+
+        // signet, regtest are not implemented, since there is no maker available.
+        if network == "testnet" {
+            opts.maker = Some(TESTNET_MAKER.to_string());
```
bonomat: Setting the maker and its ids is not necessary as you could the defaults in `maker()` by only setting the network to `testnet`.

https://github.com/itchysats/itchysats/blob/36b3987150c68209a11f6dfc0666d1c76fae7913/taker/src/lib.rs#L207-L216

bonomat (approved):
Very cool. The minor comments I've added can also be addressed in a follow-up PR if you want this to get merged for Demo :)

.github/workflows/ci.yml
```diff
@@ -28,7 +28,9 @@ jobs:
         run: rustup show
       - id: cargo-metadata
         run: |
-          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages | map(.name)')
+          # replacing itchysats with taker-electron: dirty fix, as neon and electron require the package name for processing. neon requires the package name to equal the crate name - 
+          # and electron uses the package name as name of the application.
+          WORKSPACE_MEMBERS=$(cargo metadata --format-version=1 --no-deps | jq -c '.packages |.[] | select(.name == "itchysats").name |= "taker-electron" | .name' | jq -s | jq -c '.')
```
luckysori, line 33: ❓ Any chance we could handle the new cases in a separate line somehow?

Also, I don't fully understand what's going on from the comment. You've modified the workspace `Cargo.toml` to add the new crate. What else needs to go through CI?

taker-electron/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "itchysats"
```
luckysori, line 2: ❓ Is this the correct name? A comment might be useful here too if there is a reason why we have to use this name.

taker-electron/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+once_cell = "1"
+taker = { path = "../taker" }
+tokio = { version = "1", features = ["rt-multi-thread", "macros", "sync", "net", "tracing"] }
+
+[dependencies.neon]
+version = "0.10.1"
+default-features = false
+features = ["channel-api", "napi-6", "promise-api"]
```
luckysori: 🙃 I'm not necessarily against this, but AFAIK we have consistently declared one dependency per line, rather than using dedicated separate sections.

taker-electron/README.md
```diff
@@ -0,0 +1,47 @@
+# Itchy Sats Electron
```
luckysori: 🔧 

```suggestion
# ItchySats Electron
```

taker-electron/src/lib.rs
luckysori: I am surprised that this works the way you want it to if you initialise the `RUNTIME` _inside_ of the function. Won't it just create a new one every time?

taker-electron/src/lib.rs
```diff
@@ -0,0 +1,58 @@
+use neon::prelude::*;
+use once_cell::sync::OnceCell;
+use taker::Opts;
+use tokio::runtime::Runtime;
+
+// Return a global tokio runtime or create one if it doesn't exist.
+// Throws a JavaScript exception if the `Runtime` fails to create.
```
luckysori, line 7: 🔧 You probably want these comments to be docstrings?

```suggestion
/// Return a global tokio runtime or create one if it doesn't exist.
/// Throws a JavaScript exception if the `Runtime` fails to create.
```

taker-electron/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+fn runtime<'a, C: Context<'a>>(cx: &mut C) -> NeonResult<&'static Runtime> {
+    static RUNTIME: OnceCell<Runtime> = OnceCell::new();
+
+    RUNTIME.get_or_try_init(|| Runtime::new().or_else(|err| cx.throw_error(err.to_string())))
+}
+
+// Starts the itchysats taker daemon.
+// returns a `Promise`and executes asynchronously on the `tokio` thread pool
+pub fn wrap(mut cx: FunctionContext) -> JsResult<JsPromise> {
```
luckysori: ❓ Why call it `wrap`? 🌯

taker-electron/src/main.js
```diff
@@ -0,0 +1,88 @@
+// Modules to control application life and create native browser window
```
luckysori: ❓ Could we use TypeScript instead? We usually work with it.

taker/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+    #[clap(short, long)]
+    pub wallet_xprv: Option<ExtendedPrivKey>,
+}
+
+impl Opts {
+    pub fn read() -> Opts {
+        Opts::parse()
+    }
+
+    pub fn new(network: String, password: String, data_dir: String) -> Opts {
```
luckysori: 🔧 This is only used for `electron`, so I would not try to make it generic at all, and even call it `new_electron` or something like that.

taker/src/lib.rs
```diff
@@ -0,0 +1,539 @@
...
+            headless: true,
+            service_name: "".to_string(),
+            log_level: LevelFilter::DEBUG,
+            password: None,
+            network: None,
+            app_seed: None,
+            wallet_xprv: None,
+        }
+    }
+}
```
luckysori: 🔧 I think it would be best if we didn't use `Default` for this. These are just things that make sense for `electron`. Let's just move these values into the constructor.

taker/src/lib.rs
luckysori: 🔧 I think it would be good to parse the `network` and `password` at the top and convert them into the types we want or fail early.

luckysori (approved):
Excellent stuff. I tried to clean up the `taker/src/lib.rs` changes with a fixup commit. Let me know what you think. It addresses some of the comments that I left in my review.

taker-electron/src/lib.rs
holzeis: That should be handled by OnceCell, shouldn't it?

https://docs.rs/tokio/1.5.0/tokio/sync/struct.OnceCell.html

taker-electron/src/lib.rs
holzeis: 😂 because its wrapping the taker daemon into a nodejs module 😉 any better suggestions?

taker-electron/src/lib.rs
luckysori: But in the example linked the `OnceCell` is declared globally.

I think this code is creating a new runtime every time, because the `static` is dropped as soon as we exit the `runtime` function.

taker-electron/src/lib.rs
luckysori: Hmm, I would call it something like `run` or `start`, since it's starting the taker.

taker-electron/Cargo.toml
holzeis: well, as mentioned on the ci.yaml this name is needs to match the package name - otherwise the neon build target will fail. However the package name is also used by electron-forge to build the app name. I didn't find another way of naming the app differently for electron-forge, or specify the crate-name independently from the package name for neon. We could fork the cargo-cp-artifact script and fix it, but thats seems to be a bit of an overkill to me.

taker/src/lib.rs
luckysori: Having looked at this in more detail, they are reasonable defaults. But we only use them for `electron` and I think we can avoid mutation by doing something like what I propose in https://github.com/itchysats/itchysats/pull/2967/commits/e0a05cc79fabd20aca7fc1d3dce17f91cf600dd8.

.github/workflows/ci.yml
holzeis: > Any chance we could handle the new cases in a separate line somehow?
I guess we could remove the itchysats crate from the cargo metadata result and add the taker-electron crate afterwards, but I don't really see the value in that.

The issue here is that the script assumes that the crate name and the folder name are equal. In this case its not - thus without that fix the clippy job would fail.

taker-electron/Cargo.toml
klochowicz: interesting idea, but for consistency's sake I would vote for sticking to the usual declarations (unless we change them all :P )

taker/src/lib.rs
holzeis: since we have now an acceptable default + change password, I removed the password option all together.

taker/src/lib.rs
holzeis: we could use them for any environment where we are using the nodejs module. e.g. taker-android would use it the same way.

taker/src/lib.rs
holzeis: well today its only used for electron, but basically this is a way of initializing the opts without the use of stdin. This is needed when we are using the taker as library, e.g. for electron, but it would be the same for android, iOS, etc. Hence I wouldn't call it new_electron - if we need to make it concrete, we could call it new_neon - as this is used by the neon-binding wrapper - which will then produce a nodejs module that is used by electron.

taker/src/lib.rs
holzeis: yes, but they weren't exposed here - @luckysori made a nice commit suggestion, which is fixing that.

taker-electron/src/main.js
holzeis: yes we can 👍

taker-electron/src/lib.rs
Restioson: > I think this code is creating a new runtime every time, because the static is dropped as soon as we exit the runtime function.

I do not think putting the static inside of the function has any effect. This is purely a namespace/privacy change. See the example in [the Rust reference](https://doc.rust-lang.org/reference/items/static-items.html).

Restioson: > The way I've understand it, It is the rust way of implementing a singleton pattern. Have a look
[here](https://github.com/neon-bindings/examples/tree/855eb2d52d650afa489b6b4724095c5fdfd5c9ad/examples/tokio-fetch#executor) for more details on the boilerplate.

I agree that it is the standard way of making a singleton, but I don't think that one should be required here. If `start` is not called multiple times, then there is no point in making the reactor global, given that it's only used in `start` itself - it could be constructed at the beginning of the function. Correct me if I am wrong on this as I am not familiar with neon!

taker-electron/style.css
```diff
@@ -0,0 +1,41 @@
...
+    100% {
+        -webkit-transform: scale(1);
+        transform: scale(1);
+    }
+}
+
+#message {
+    text-align: center;
+    color: white;
+}
```
Restioson: Another newline nit :sweat_smile:

holzeis: 🙈 oh boy

taker-electron/src/lib.rs
holzeis: @Restioson do you have a suggestion how to implement it differently?

holzeis:
bors retry

PR #2973, opened 2022-09-19
fix: add secret key to rocket config figment

### LINKED ISSUE #2972: Troubles with authentication when running taker on `master`
I'm running into troubles with both latest Docker image and binary.

- docker master errors with:
`Error: insecure secret key config`

docker fails both with and without providing password explicitly, even with a fresh config.

- `dev-taker` (and `cargo run --bin taker`) just show that we're not authenticated on the feed, displaying annoying information that the daemon is not running; nothing seems to be working.
excerpt from the logs:
```
2022-09-19 08:29:12 DEBUG http: Handled request method=GET path=/api/version status=200 OK
2022-09-19 08:29:12 DEBUG http: Handled request method=GET path=/api/feed status=401 Unauthorized
```

were there any necessary steps that I'm missing?

holzeis:
resolves #2972

For some reason this works when starting the taker with cargo run, but does not work in docker. I guess if we release the binary the debug profile is disabled and a secret key has to be provided.

    /// The default profile: "debug" on `debug`, "release" on `release`.
    #[cfg(debug_assertions)]
    pub const DEFAULT_PROFILE: Profile = Self::DEBUG_PROFILE;

https://github.com/SergioBenitez/Rocket/blob/6778089c129ee15dfd524413f067c9c572226159/core/lib/src/config/config.rs#L429-L431

see also 
* https://api.rocket.rs/master/rocket/config/struct.SecretKey.html

### DIFF
--- a/taker/src/main.rs
+++ b/taker/src/main.rs
@@ -308,7 +308,8 @@ async fn main() -> Result<()> {
     let figment = rocket::Config::figment()
         .merge(("address", opts.http_address.ip()))
         .merge(("port", opts.http_address.port()))
-        .merge(("cli_colors", false));
+        .merge(("cli_colors", false))
+        .merge(("secret_key", RandomSeed::default().seed()));
 
     let db = sqlite_db::connect(data_dir.join("taker.sqlite"), true).await?;
 
--- a/maker/src/main.rs
+++ b/maker/src/main.rs
@@ -108,7 +108,8 @@ async fn main() -> Result<()> {
     let figment = rocket::Config::figment()
         .merge(("address", opts.http_address.ip()))
         .merge(("port", opts.http_address.port()))
-        .merge(("cli_colors", false));
+        .merge(("cli_colors", false))
+        .merge(("secret_key", RandomSeed::default().seed()));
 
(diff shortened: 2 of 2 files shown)


taker/src/main.rs
```diff
@@ -308,7 +308,10 @@ async fn main() -> Result<()> {
     let figment = rocket::Config::figment()
         .merge(("address", opts.http_address.ip()))
         .merge(("port", opts.http_address.port()))
-        .merge(("cli_colors", false));
+        .merge(("cli_colors", false))
+        .merge(("secret_key", RandomSeed::default().seed()));
+    // in case we are implementing a session store - this seed should not be automatically generated
```
bonomat: 🙃
I would prefer the log before the line

bonomat (approved):
Good catch. Thanks. 

🔧
We will need to do this also for the maker binary.

taker/src/main.rs
bonomat: Maybe this comment is not even needed given we have the ticket https://github.com/itchysats/itchysats/issues/2978 already?

bonomat:
bors retry

PR #2986, opened 2022-09-20
chore: Move build.rs frontend prep to relevant crates

klochowicz:
Daemon crate should not be concerned about frontend directories; this is the job
of the binaries.


klochowicz:
tiny change, hopefully this will allow for better caching of the `daemon` crate by sccache.

Issue #897, opened 2021-12-14
Secure sensitive data with password


thomaseizinger:
Currently, we are storing the seedfile unencrypted on disk. The seedfile is used to derive the internal wallet's descriptor. With https://github.com/itchysats/itchysats/issues/865, that will change to a passphrase but we will have to store that on disk as well.

We should introduce a password-derived encryption scheme that we can use to encrypt this information on disk.

Challenges:

- Delay p2p network communication until unlocked (cannot derive identity key until we have the seed)
- Delay wallet initialization until unlocked (cannot derive descriptor until we have the seed)

Related issues: https://github.com/itchysats/itchysats/discussions/753, https://github.com/itchysats/itchysats/issues/865
Relation section in vision document: https://github.com/itchysats/itchysats/blob/master/docs/VISION.md#always-available-ui

klochowicz:
I believe this should be addressed along with  #2967 .
For the full node perspectives, we could assume that they are isolated network nodes and there's some degree of protection there. For Electron, the data sits in `Application Support` folder and could be accessed easier.

bonomat:
Does this make sense for users who run with `--password` (e.g. Raspiblitz, Cloud9, etc)? The problem is that you could start with a different password and then we won't be able to open `taker_seed` anymore. 

Which means, on platforms where we don't have control over how the password is generated we couldn't make use of this.

klochowicz:
it will be tricky, but if we want to support  #1065 on any of these platforms, we should figure out something, because an unencrypted `taker_seed` would get saved to the Downloads folder...

bonomat:
This project is unmaintained

PR #2988, opened 2022-09-20
Allow exporting seed in taker UI

### LINKED ISSUE #1065: Make it easier to back up the seed-file
After our discussion in https://github.com/itchysats/itchysats/issues/865 and #1016 we dropped the feature to backup and re-initialize the ItchySats (wallet) seed.

Allow users of the binary to backup their seed in textual form (so they don't have to backup a complete file, the generated seed-file is still binary), so they can save it in e.g. a password safe.

klochowicz:
TODO: 
- [x] Restoring from the seed (will be covered by #2989 )
- [x]  Encrypting the saved seed (or at least being clear that it needs to be handled safely) (will be covered by #897 )
- [x] A toast message about the outcome of the request (success or fail)
- ~~Add tests~~

### DIFF
--- a/taker/src/routes.rs
+++ b/taker/src/routes.rs
@@ -1,4 +1,5 @@
-#![allow(clippy::let_unit_value)] // see: https://github.com/SergioBenitez/Rocket/issues/2211
+#![allow(clippy::let_unit_value)]
+// see: https://github.com/SergioBenitez/Rocket/issues/2211
 use daemon::bdk;
 use daemon::bdk::bitcoin::Amount;
 use daemon::bdk::bitcoin::Network;
@@ -10,6 +11,7 @@ use daemon::oracle;
 use daemon::projection;
 use daemon::projection::CfdAction;
 use daemon::projection::FeedReceivers;
+use daemon::seed::ThreadSafeSeed;
 use daemon::wallet;
 use daemon::TakerActorSystem;
 use http_api_problem::HttpApiProblem;
@@ -33,6 +35,8 @@ use rocket_cookie_auth::auth::Auth;
 use rocket_cookie_auth::forms::ChangePassword;
 use rocket_cookie_auth::forms::Login;
 use rocket_cookie_auth::user::User;
+use rocket_download_response::mime;
+use rocket_download_response::DownloadResponsePro;
 use rust_embed::RustEmbed;
 use rust_embed_rocket::EmbeddedFileExt;
 use serde::Deserialize;
@@ -40,6 +44,7 @@ use serde::Serialize;
 use shared_bin::ToSseEvent;
 use std::borrow::Cow;
 use std::path::PathBuf;
+use std::sync::Arc;
 use tokio::select;
 use tokio::sync::watch;
 use tracing::instrument;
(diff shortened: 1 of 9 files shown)


taker/src/routes.rs
```diff
@@ -336,6 +340,15 @@ pub async fn get_version() -> Json<HealthCheck> {
     })
 }
 
+#[rocket::get("/seed_backup")]
+#[instrument(name = "GET /seed_backup", skip_all)]
+pub async fn get_seed_backup(
+    seed: &State<Arc<ThreadSafeSeed>>,
```
bonomat, line 349: Don't forget to add the `user` guard, otherwise this API does not need authentication.

holzeis:
Added a toast about the successful outcome. file-saver will not produce any errors as it is based on the native html 5 download href, hence no error message in case of no success.

holzeis:
Opted against adding regression tests, as there is not much logic to test anyway.

bonomat (approved):
Haven't tested it but looks good to me. Can you squash the commits please?

klochowicz:
I'm not sure whether this should go in as-is (hence the draft status), especially without prior to testing this. I have kept it as draft as I noticed on Friday that the saved file is not the same binary size as the original one; and worked on a test that would ensure we backup the right thing.

 I suspect RocketResponderPro is adding headers / screwing up encoding - might be easier to go with serde and then deserialise into a blob to ensure the file is as it's supposed to be. 

The other way (perhaps simpler altogether) that I came up with during the weekend (but haven't tried) would be just serving `taker_seed` as a `NamedFile` in Rocket.  Of course then we can't backup from `--app-seed`, but that's not a requirement.

klochowicz:
> Opted against adding regression tests, as there is not much logic to test anyway.

As this is one of the paths where the user might lose money if it doesn't work properly, I would certainly test this to prove that it works.
However, manual testing should be enough for this PR, don't worry about it here - I already started working on the test, so I will submit it separately for review 😊

holzeis:
> I'm not sure whether this should go in as-is (hence the draft status), especially without prior to testing this. I have kept it as draft as I noticed on Friday that the saved file is not the same binary size as the original one; and worked on a test that would ensure we backup the right thing.

The downloaded seed file works properly.. I've tested it manually.

1. startup itchysats
2. backup seed
3. stop itchysats
4. replace seed file with downloaded one
5. startup itchysats

Everything is working ;)

>> Opted against adding regression tests, as there is not much logic to test anyway.

>As this is one of the paths where the user might lose money if it doesn't work properly, I would certainly test this to prove that it works.

I think the regressions tests are better covered including the restore feature.

> The other way (perhaps simpler altogether) that I came up with during the weekend (but haven't tried) would be just serving taker_seed as a NamedFile in Rocket. Of course then we can't backup from --app-seed, but that's not a requirement.

I guess that would also work, for the app-seed argument I was thinking if we could store that also to a taker_seed file if we get it, then we have the same behavior. wdyt?

klochowicz:
> The downloaded seed file works properly.. I've tested it manually.
> 
> 1. startup itchysats
> 2. backup seed
> 3. stop itchysats
> 4. replace seed file with downloaded one
> 5. startup itchysats
> 
> Everything is working ;)

awesome! 🤘

klochowicz:
> I guess that would also work, for the app-seed argument I was thinking if we could store that also to a taker_seed file if we get it, then we have the same behavior. wdyt?

that would also work, yes! however it might be a too niche feature to even bother. :-)

PR #2993, opened 2022-09-20
feat(wallet-history): show transaction history

bonomat:
I've chosen to go for a custom type which allows us to add a convenience field for linking to mempool. In the long run we could add a link to a CFD there if the transaction was funding/closing a CFD.

Looks like this: 

image

### DIFF
--- a/shared-bin/src/to_sse_event.rs
+++ b/shared-bin/src/to_sse_event.rs
@@ -1,4 +1,7 @@
 use daemon::bdk::bitcoin::Amount;
+use daemon::bdk::bitcoin::Network;
+use daemon::bdk::bitcoin::Txid;
+use daemon::bdk::BlockTime;
 use daemon::identify;
 use daemon::listen_protocols::does_maker_satisfy_taker_needs;
 use daemon::listen_protocols::REQUIRED_MAKER_LISTEN_PROTOCOLS;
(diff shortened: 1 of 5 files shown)


shared-bin/src/to_sse_event.rs
```diff
@@ -25,14 +28,54 @@ pub struct WalletInfo {
...
+}
+
+impl From<(Network, &daemon::bdk::TransactionDetails)> for TransactionDetails {
+    fn from((network, tx): (Network, &daemon::bdk::TransactionDetails)) -> Self {
+        let txid = tx.txid;
+        let link = match network {
+            Network::Bitcoin => format!("https://mempool.space/tx/{txid}"),
+            Network::Testnet => format!("https://mempool.space/testnet/tx/{txid}"),
+            Network::Signet => format!("https://mempool.space/signet/tx/{txid}"),
+            Network::Regtest => txid.to_string(),
```
luckysori: ❓ Perhaps an optional link would be better?

luckysori (approved):
Very cool!

taker-frontend/src/types.ts
```diff
@@ -10,6 +10,20 @@ export interface WalletInfo {
     balance: number;
     address: string;
     last_updated_at: number;
+    transactions: Transaction[];
+}
+
+export interface Transaction {
```
holzeis, line 16: Isn't the convention for naming type declarations something like *.d.ts?

holzeis (approved):
Nice 👍

taker-frontend/src/types.ts
bonomat: I honestly don't know.

PR #2999, opened 2022-09-21
Extend electron configuration to allow logging into a file

### LINKED ISSUE #2982: Extend electron configuration to allow logging into a file
Using itchysats wrapped e.g. in electron will not have a terminal accessible to view the logs. Hence having the option to pipe the logs to a file would be helpful for supporting these kind of users.

holzeis:
Introduces an opt for writing logs to a file. If flag is set, logs will be written to the data dir. The file name will be concatenated from the service name + itchysats.log

### Examples

```
cargo run --bin taker -- --data-dir /itchysats/ --log-to-file testnet
```
You can find the logs here.
/itchysats/testnet/taker.log

```
cargo run --bin maker -- --data-dir /itchysats/ --log-to-file mainnet
```
You can find the logs here.
- /itchysats/mainnet/maker.log

resolves #2982

### DIFF
--- a/shared-bin/src/logger.rs
+++ b/shared-bin/src/logger.rs
@@ -7,7 +7,9 @@ use opentelemetry::sdk::Resource;
 use opentelemetry::KeyValue;
 use opentelemetry_otlp::WithExportConfig;
 use time::macros::format_description;
+use tracing_appender::non_blocking::WorkerGuard;
 use tracing_subscriber::filter::Directive;
+use tracing_subscriber::fmt;
 use tracing_subscriber::fmt::time::UtcTime;
 use tracing_subscriber::EnvFilter;
 use tracing_subscriber::Layer;
@@ -32,9 +34,11 @@ pub fn init(
     verbose_spans: bool,
     service_name: &str,
     collector_endpoint: &str,
-) -> Result<()> {
+    log_to_file: bool,
+    data_dir: &str,
+) -> Result<Option<WorkerGuard>> {
     if level == LevelFilter::OFF {
-        return Ok(());
+        return Ok(None);
     }
 
     let is_terminal = atty::is(atty::Stream::Stderr);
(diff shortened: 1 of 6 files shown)


shared-bin/src/logger.rs
```diff
@@ -116,18 +120,31 @@ pub fn init(
         None
     };
 
+    let (file_log, guard) = if log_to_file {
+        let file_appender =
+            tracing_appender::rolling::never(data_dir, format!("{service_name}.log"));
```
bonomat, line 125: Is there a way to say what the max size of this file can be? I'm afraid that we will flood the disk with logs if this is unbound.

holzeis: We could roll the log file daily.

bonomat: Would that delete the old file automatically?

holzeis: unfortunately not.. there is a github project which would allow to create a new log file after a certain size has exceeded, but it does not have a lot of attention https://github.com/Axcient/rolling-file-rs (last commit 2 years ago and just a few stars)

tracing appender is from the tokio team but only supports a limited set of rollover scenarios 
- minutely
- hourly
- daily
- never
https://github.com/tokio-rs/tracing/tree/master/tracing-appender

We could go for daily log files, but I feel like that will produce a lot of small files on the long run, which is also not very nice. Can we estimate how much logs will be produced as to get a feeling if never would be a viable option? respectively can we limit the log level so that the file will not grow unmanageable?

da-kami: We used https://github.com/tokio-rs/tracing/tree/master/tracing-appender in the XMR-BTC project with a `never` appender:

https://github.com/comit-network/xmr-btc-swap/blob/9e85637419594f69d05e0d03767d66b07001a0ae/swap/src/cli/tracing.rs#L19

We should be able to define a writer for `tracing::appender`, the [repo readme](https://github.com/tokio-rs/tracing/tree/master/tracing-appender) states:

| Using any type implementing [std::io::Write](https://doc.rust-lang.org/std/io/trait.Write.html) in a non-blocking fashion.

When looking around I found https://github.com/BourgondAries/file-rotate which I think offers what we want and seems to be still under active development. As far as I could see it implements `std::io::Write` we we should be able to use it with `tracing-appender`. Haven't given it a try though.

You could give it a "quick try" and timebox it so we don't end up with a rabbit hole.

holzeis: Doesn't have many stars either, but I will give it a try.

edit: created a ticket https://github.com/itchysats/itchysats/issues/3057 for that evaluation and might fix with a follow up PR.

PR #3002, opened 2022-09-23
Allow unauthenticated access to  `/api/metrics` endpoint

klochowicz:
Prometheus does not allow cookie-based auth. Leave it unauthenticated until we
support different type of auth for prometheus.


bonomat (approved):

Restioson:
Thanks! We should make sure to implement proper auth before this gets to mainnet if we can, just in case something unforseen is sensitive, and also for defence in depth.

PR #3017, opened 2022-09-23
Allow basic auth on the /metrics endpoint

### LINKED ISSUE #3001: Allow for basic or bearer auth for prometheus /api/metrics endpoint
Since #2903, Prometheus can no longer authenticate to scrape the /api/metrics endpoint. Prometheus does not support adding cookies to scraping endpoints, so we cannot provide a cookie for Prometheus to use. I am marking this as a 'bug' as this currently means that metrics don't work in versions deployed from master (e.g testnet setup).

Possible solutions: 
- Allow for basic auth to this endpoint again. I have taken a look locally at this but I realise it may be unwanted/controversial. I believe that the motivation for #2903 was mostly user experience, so in this case it shouldn't matter as this only touches how the automation works, but perhaps there are other considerations here. 
- Modify rocket-cookie-auth to be able to take a session token in some format through an `Authorization: Bearer` header. We would also need to have a way to generate a bearer token for Prometheus to use, and ensure that it would always be up to date.

Restioson:
I've moved the rocket-basicauth/lib.rs into rocket-cookie-auth as they now share some state. I also renamed `Authenticated` to `BasicAuthGuard` so we can be explicit as to what can be authenticated by basic auth. Fixes #3001.

A few unresolved questions
- [ ] Should I try and combine the two error enums? Basic auth has different errors than cookie does
- [ ] How should the code be organised? Maybe `basic` could be a submodule of `auth`?
- [ ] Should any other endpoints be updated to use basic auth?


holzeis:
Looks good 👍 , but I guess the naming could be improved now.. we could rename it to a rocket-auth in general and also name the cookie guard explicitly, thus the library allow the dev to choose how to protect the endpoint. you already do that of course.. I am just referring to the naming 😉 

Regarding the enums I suggest to combine them and make one implementation out of it, instead of making basic a submodule.

holzeis (approved):

bonomat (commented):
Can we keep the module `rocket-basic-auth` and use it in combination with `rocket-cookie-auth`? 

Given we can define the guards per route we should be able to use both.

Restioson:
> Can we keep the module rocket-basic-auth and use it in combination with rocket-cookie-auth?

The basic auth module needs to use the same state as cookie auth does in order to keep up to date with changing passwords, so if it is a separate crate it will still need to depend on `rocket-cookie-auth`. Therefore, I'm not so sure it's worth it.

> Given we can define the guards per route we should be able to use both.

Currently that's how it's done for metrics - instead of User, basic auth has a separate guard.

Restioson:
I've fixed the CI failures here

Restioson:
Rebased on master too :)

crates/taker/src/routes.rs
```diff
@@ -302,11 +303,9 @@ pub async fn post_withdraw_request(
     Ok(projection::to_mempool_url(txid, *network.inner()))
 }
 
-// TODO: Use non-cookie auth for /metrics endpoint as Prometheus does not
-// support cookie-auth (for now, leave unauthenticated)
 #[rocket::get("/metrics")]
 #[instrument(name = "GET /metrics", skip_all, err)]
-pub async fn get_metrics<'r>() -> Result<String, HttpApiProblem> {
+pub async fn get_metrics<'r>(_basic_auth: BasicAuthGuard) -> Result<String, HttpApiProblem> {
```
bonomat, line 308: Is it possible to have either `basic_auth` or `cookie_auth` ? That way we don' thave to login twice if we ever call the endpoint manually.

bonomat (approved):
Very nice. 

> Should I try and combine the two error enums? Basic auth has different errors than cookie does

I think the separation is fine.

> How should the code be organised? Maybe basic could be a submodule of auth?

I think where it is right now (`crates/rocket-cookie-auth/src/basic.rs`) is fine.

> Should any other endpoints be updated to use basic auth?

No I think that's enough.

crates/rocket-cookie-auth/src/lib.rs
```diff
@@ -5,6 +5,7 @@ use anyhow::Result;
 use rocket::async_trait;
 
 pub mod auth;
+pub mod basic;
```
luckysori, line 8: 🙃 We could rename the crate to `rocket-auth` and have modules for `basic` and `cookie`-based authentication.

```suggestion
pub mod cookie;
pub mod basic;
```

luckysori (approved):
LGTM.

bonomat:

PR #3021, opened 2022-09-26
Timeout for place order in case maker does not respond

da-kami:
If the maker does not respond within `1` minutes the taker treats the placed order as failure and does not proceed with contract setup. This prevents us from positions that remain in `PendingSetup` in case the maker never responds.

---

This was the expected behavior before, but when we introduced the new protocol the timeout was removed. I don't think the taker should wait indefinitely for a response here. 

Configured to `60` seconds at the moment because we expect the maker to take a decision within `60` seconds. A decision to take an offer should be fast, if a decision takes longer than one minute the taker might not want to proceed.

### DIFF
--- a/daemon/src/order/current/taker.rs
+++ b/daemon/src/order/current/taker.rs
@@ -32,10 +32,15 @@ use model::Leverage;
 use model::Offer;
 use model::OrderId;
 use model::Role;
+use std::time::Duration;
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Thanks!

holzeis (approved):

PR #3026, opened 2022-09-26
Automatically bump version to release version

da-kami:
With removing `vergen` we have to bump the version to the release version when releasing. By automatically bumping the version we have less manual steps in our release pipeline.

### DIFF
--- a/.github/workflows/draft-new-release.yml
+++ b/.github/workflows/draft-new-release.yml
@@ -34,13 +34,22 @@ jobs:
         with:
           version: ${{ github.event.inputs.version }}
 
+      - name: Bump version in Cargo.toml
+        uses: thomaseizinger/set-crate-version@1.0.0
+        with:
+          version: ${{ github.event.inputs.version }}
+          manifest: daemon/Cargo.toml
+
+      - name: Update Cargo.lock
+        run: cargo update --workspace
+
       - name: Commit changelog and manifest files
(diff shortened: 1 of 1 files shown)


da-kami:
Looks like our release process knew the future all along 🤪

https://github.com/itchysats/itchysats/blob/46a1fe05c936860e927da75a54859ce60f46132a/.github/workflows/draft-new-release.yml#L56

luckysori (approved):
Thanks!

.github/workflows/draft-new-release.yml
```diff
@@ -34,13 +34,22 @@ jobs:
         with:
           version: ${{ github.event.inputs.version }}
 
+      - name: Bump version in Cargo.toml
+        uses: thomaseizinger/set-crate-version@1.0.0
+        with:
+          version: ${{ github.event.inputs.version }}
```
bonomat, line 40: This will not commit it right?

bonomat (approved):

.github/workflows/draft-new-release.yml
da-kami: This modifies the version in the file, adding and committing the changes is in the lines below. 

Note: I did not specifically check the code of the action but configured it according to the xmr-btc project where we set it up like this.

holzeis (approved):

PR #3029, opened 2022-09-27
Separate monitor logic from transaction data 

### LINKED ISSUE #3028: Commit published but no CET published
We have two positions on mainnet which were manual committed by us but the CETs were not published: 

- 5a334531-dcc6-4f8a-b2ec-331af1ee1f56
	- commit tx: https://mempool.space/tx/957e0a9c808e6a36b76897a30a864ad3ccc439bacc8a3be1f980259199222914

- 7d346ced-b26c-4d65-9740-7a3296595e29
	- commit tx: https://mempool.space/tx/50d26b17343c2559b72c482115bf103dcbd34374dc84e3397143badbde2c22ae

This is bad because this also means that the taker neither didn't publish the CET (he might have been offline though)

Excerpt of the logs: 

```
[
  {
    "timestamp": "2022-09-23T03:03:00.257824635Z",
    "level": "INFO",
    "fields": {
      "message": "Appended event to database",
      "event": "CommitConfirmed",
      "order_id": "7d346ced-b26c-4d65-9740-7a3296595e29"
    },
    "target": "sqlite_db",
    "span": {
      "name": "Handle Event"
    }
  },
  {
    "timestamp": "2022-09-23T03:03:00.254404027Z",
    "level": "INFO",
    "fields": {
      "order_id": "7d346ced-b26c-4d65-9740-7a3296595e29",
      "peer_id": "Some(12D3KooWRG3cBytxhAiHBZYeXCb8nDRWLLfDdy9Yn31MDr9WVdeR)",
      "message": "Issuing event: CommitConfirmed"
    },
    "target": "model::cfd",
    "span": {
      "name": "Sync monitor"
    }
  }
  Show
  context,
  {
    "timestamp": "2022-09-23T03:01:24.588098307Z",
    "level": "INFO",
    "fields": {
      "message": "Appended event to database",
      "event": "OracleAttestedPriorCetTimelock",
      "order_id": "7d346ced-b26c-4d65-9740-7a3296595e29"
    },
    "target": "sqlite_db",
    "span": {
      "name": "Handle Event"
    }
  },
  {
    "timestamp": "2022-09-23T03:01:24.582554882Z",
    "level": "INFO",
    "fields": {
      "order_id": "7d346ced-b26c-4d65-9740-7a3296595e29",
      "peer_id": "Some(12D3KooWRG3cBytxhAiHBZYeXCb8nDRWLLfDdy9Yn31MDr9WVdeR)",
      "message": "Issuing event: OracleAttestedPriorCetTimelock"
    },
    "target": "model::cfd",
    "span": {
      "name": "Handle NewAttestationFetched"
    }
  },
  {
    "timestamp": "2022-09-23T03:00:05.493433357Z",
    "level": "INFO",
    "fields": {
      "order_id": "5a334531-dcc6-4f8a-b2ec-331af1ee1f56",
      "peer_id": "Some(12D3KooWRG3cBytxhAiHBZYeXCb8nDRWLLfDdy9Yn31MDr9WVdeR)",
      "message": "Issuing event: ManualCommit"
    },
    "target": "model::cfd",
    "span": {
      "action": "\"commit\"",
      "order_id": "5a334531-dcc6-4f8a-b2ec-331af1ee1f56",
      "name": "POST /cfd/<order_id>/<action>"
    }
  },
  {
    "timestamp": "2022-09-23T03:00:05.429035339Z",
    "level": "DEBUG",
    "fields": {
      "message": "Handled request",
      "method": "POST",
      "path": "/api/cfd/7d346ced-b26c-4d65-9740-7a3296595e29/commit",
      "status": "200 OK"
    },
    "target": "http"
  },
  {
    "timestamp": "2022-09-23T03:00:05.428494617Z",
    "level": "INFO",
    "fields": {
      "message": "Appended event to database",
      "event": "ManualCommit",
      "order_id": "7d346ced-b26c-4d65-9740-7a3296595e29"
    },
    "target": "sqlite_db",
    "span": {
      "name": "Handle Event"
    }
  }]
```

We 
1. manual committed
2. fetched the attestation
3. issued event `OracleAttestedPriorCetTimelock`
4. issued event `CommitConfirmed`

then nothing

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/3028.

When we worked on https://github.com/itchysats/itchysats/commit/b4225d5d35e27caef749d5c8ebf4e0f28c55d649 we inadvertently introduced a bug[1]. The gist of it is that, after a restart following a commit transaction being confirmed, we were not monitoring for the CET and refund timelocks anymore. This means that such a CFD would be stuck.

To fix it we have separated the transaction data that becomes available as events are processed from the logic of whether a
transaction (or a timelock) should be monitored or not. The data should always be remembered, and only the flags that control the monitoring logic should change as new events are processed.

[1]: https://github.com/itchysats/itchysats/issues/3028.

### DIFF
--- a/daemon/src/monitor.rs
+++ b/daemon/src/monitor.rs
@@ -117,19 +117,29 @@ pub struct Actor {
 struct Cfd {
     id: OrderId,
 
-    monitor_lock_finality: Option<Lock>,
-    monitor_commit_finality: Option<Commit>,
+    lock: Option<Lock>,
+    monitor_lock_finality: bool,
+
+    collaborative_settlement: Option<(Txid, Script)>,
+    monitor_collaborative_settlement_finality: bool,
+
+    commit: Option<Commit>,
+    monitor_commit_finality: bool,
     monitor_cet_timelock: bool,
     monitor_refund_timelock: bool,
-    monitor_refund_finality: Option<Refund>,
+
+    cet: Option<(Txid, Script)>,
+    monitor_cet_finality: bool,
+
+    refund: Option<Refund>,
+    monitor_refund_finality: bool,
+
     monitor_revoked_commit_transactions: Vec<RevokedCommit>,
-    monitor_collaborative_settlement_finality: Option<(Txid, Script)>,
-    monitor_cet_finality: Option<(Txid, Script)>,
 
     // Rebroadcast transactions upon startup
-    lock_tx: Option<Transaction>,
-    cet: Option<Transaction>,
-    commit_tx: Option<Transaction>,
+    broadcast_lock: Option<Transaction>,
+    broadcast_cet: Option<Transaction>,
+    broadcast_commit: Option<Transaction>,
 
     version: u32,
 }
@@ -140,17 +150,22 @@ impl sqlite_db::CfdAggregate for Cfd {
     fn new(_: Self::CtorArgs, cfd: sqlite_db::Cfd) -> Self {
         Self {
             id: cfd.id,
-            monitor_lock_finality: None,
-            monitor_commit_finality: None,
+            lock: None,
+            monitor_lock_finality: false,
+            collaborative_settlement: None,
+            monitor_collaborative_settlement_finality: false,
+            commit: None,
+            monitor_commit_finality: false,
             monitor_cet_timelock: false,
             monitor_refund_timelock: false,
-            monitor_refund_finality: None,
-            monitor_revoked_commit_transactions: Vec::new(),
-            monitor_collaborative_settlement_finality: None,
-            monitor_cet_finality: None,
-            lock_tx: None,
             cet: None,
-            commit_tx: None,
+            monitor_cet_finality: false,
+            refund: None,
+            monitor_refund_finality: false,
+            monitor_revoked_commit_transactions: Vec::new(),
+            broadcast_lock: None,
+            broadcast_cet: None,
+            broadcast_commit: None,
             version: 0,
         }
     }
@@ -186,16 +201,16 @@ impl Cfd {
                 } = TransactionsAfterContractSetup::new(&dlc);
 
                 Self {
-                    monitor_lock_finality: Some(lock),
-                    monitor_commit_finality: Some(commit),
+                    lock: Some(lock),
+                    monitor_lock_finality: true,
+                    commit: Some(commit),
+                    monitor_commit_finality: true,
                     monitor_cet_timelock: true,
                     monitor_refund_timelock: true,
-                    monitor_refund_finality: Some(refund),
+                    refund: Some(refund),
+                    monitor_refund_finality: true,
                     monitor_revoked_commit_transactions: Vec::new(),
-                    monitor_collaborative_settlement_finality: None,
-                    lock_tx: Some(dlc.lock.0),
-                    cet: None,
-                    commit_tx: None,
+                    broadcast_lock: Some(dlc.lock.0),
                     ..self
                 }
             }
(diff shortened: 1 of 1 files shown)


bonomat (approved):
Tested locally ☑️

da-kami:
Beyond the scope of this PR: 
I think it would be great if we could have better test coverage for the monitoring logic so we avoid breaking when we refactor. 
I think we already have the means to test the logic aside from the actual monitoring task, it would be great if we could have tests that purely assert on the logic of what is to be monitored when certain events happen.
Do you think it's worth adding a ticket for this?

daemon/src/monitor.rs
```diff
@@ -117,19 +117,29 @@ pub struct Actor {
...
-    monitor_collaborative_settlement_finality: Option<(Txid, Script)>,
-    monitor_cet_finality: Option<(Txid, Script)>,
 
     // Rebroadcast transactions upon startup
-    lock_tx: Option<Transaction>,
-    cet: Option<Transaction>,
-    commit_tx: Option<Transaction>,
+    broadcast_lock: Option<Transaction>,
+    broadcast_cet: Option<Transaction>,
+    broadcast_commit: Option<Transaction>,
```
da-kami, line 142: ❓ Why is there no `broadcast_refund`?

da-kami, line 142: 🤔 Unless I missed something `broadcast_commit` is never set to `Some`. Is that on purpose or is there something missing here?

daemon/src/monitor.rs
```diff
@@ -207,86 +222,70 @@ impl Cfd {
...
-            CetTimelockExpiredPostOracleAttestation { cet, .. } => Self {
-                cet: Some(cet.clone()),
-                monitor_cet_finality: cet_txid_and_script(cet),
+            CetTimelockExpiredPostOracleAttestation { cet, .. }
+            | OracleAttestedPostCetTimelock { cet, .. } => Self {
+                broadcast_cet: Some(cet.clone()),
+                cet: cet_txid_and_script(cet),
+                monitor_cet_finality: true,
                 monitor_cet_timelock: false,
                 ..self
```
da-kami, line 287: 🤔 Given that we apply the events in order and some events set `broadcast_...` transactions, shouldn't *all* other events (that don't set `broadcast_...` to `Some`) set the value to `None`? 
Otherwise we might pass on `Some` values unintended.

daemon/src/monitor.rs
luckysori: There was never a `refund: Option<Transaction>` as you can see on the left side of this diff.

I think we want to avoid situations where refund is published automatically after a restart. 

---

Aside from this PR:

To be honest, I'm not sure refund should be published automatically during runtime either. It's rare enough and disrupting enough (doesn't really enforce the bet) that I would argue it should always be triggered manually by either party if the oracle never does its thing. You could theoretically even let the parties settle collaboratively instead.

luckysori: Good catch! But looking at the diff we never actually set it to `Some` before either! Which means if we ever found ourselves in a situation where a restart occurred after a `ManualCommit` was emitted but the commit transaction publication triggered by the `process_manager` _failed_, we would have been stuck.

I'll add another commit to fix this.

daemon/src/monitor.rs
luckysori: I don't think so.

As an example, if we have a `ManualCommit` event which sets the `broadcast_commit` transaction to `Some` and we get an `OracleAttestedPriorCetTimelock`, we must still broadcast the commit transaction.

I think particularly with the rebroadcasting mechanism we don't need to be that careful anyway. If we fail to rebroadcast because it's already there or because something else ended up happening nothing bad can happen.

daemon/src/monitor.rs
luckysori: Done with https://github.com/itchysats/itchysats/pull/3029/commits/63f0a89c801326e5fe3f045efb32d17e3a3c6d95.

luckysori:
> Beyond the scope of this PR: I think it would be great if we could have better test coverage for the monitoring logic so we avoid breaking when we refactor. I think we already have the means to test the logic aside from the actual monitoring task, it would be great if we could have tests that purely assert on the logic of what is to be monitored when certain events happen. Do you think it's worth adding a ticket for this?

Absolutely, @bonomat and I talked about this offline too! I'll create a ticket to test the logic in `monitor` in isolation.

See https://github.com/itchysats/itchysats/issues/3035.

luckysori:
bors r+

Any further comments I can address in a separate PR.

PR #3031, opened 2022-09-27
feature(environment): support string

### LINKED ISSUE #2969: Change `Environment` to a String field
Proposal: change it to a String field which can be passed in. That way it is _dynamic_ and we don't have to extend this enum and release a new version.

_Originally posted by @bonomat in https://github.com/itchysats/itchysats/pull/2955#discussion_r971447151

### LINKED ISSUE #3008: Set electron environment to the using platform
Currently we only set the environment to electron if the electron app version of itchysats is used. However, electron itself could be used various platforms. We intend to publish a windows and a mac version - hence we would like to know whether itchysats electron is used under mac or windows.

Two additional environment variables should be added - windows (win32 and win64) and macOs (darwin). If electron.platfom does not match either a fallback to electron should be used.

bonomat:
We can run on all kinds of environment, hence Environment should just be a wrapper around string. However, for backwards compatibility with <=0.6.x we need to support certain formats: `Docker`, `Umbrel`, `RaspiBlitz`, `Unknown`. For all other we format the string to lowercase.

Resolves #3008
Resolves #2969

Let me know what you think.

### DIFF
--- a/crates/daemon-tests/src/lib.rs
+++ b/crates/daemon-tests/src/lib.rs
@@ -906,7 +906,7 @@ impl Taker {
             projection_actor,
             maker_identity,
             maker_multiaddr.clone(),
-            Environment::Test,
+            Environment::new("test"),
         )
         .unwrap();
 
--- a/crates/daemon/src/identify.rs
+++ b/crates/daemon/src/identify.rs
@@ -65,14 +65,14 @@ mod tests {
     async fn both_parties_request_identify_info_on_connection_established() {
         let (maker_peer_id, maker_endpoint, maker_receiver) = create_endpoint_with_identify(
             "0.4.22".to_string(),
-            Environment::Unknown,
+            Environment::unknown(),
             Keypair::generate_ed25519().public(),
             HashSet::new(),
             HashSet::from(["some_maker_protocol".to_string()]),
         );
         let (_, taker_endpoint, taker_receiver) = create_endpoint_with_identify(
             "0.4.22".to_string(),
-            Environment::Umbrel,
+            Environment::new("umbrel"),
             Keypair::generate_ed25519().public(),
             HashSet::new(),
             HashSet::from(["some_taker_protocol".to_string()]),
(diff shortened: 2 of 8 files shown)


holzeis (approved):

daemon/src/lib.rs
```diff
@@ -447,32 +447,34 @@ where
...
+
+    pub fn unknown() -> Environment {
+        Self("Unknown".to_string())
+    }
+
+    pub fn as_string(&self) -> String {
+        self.0.clone()
+    }
+
+    fn parse_known_variances(string: &str) -> String {
```
luckysori: 🙃 
```suggestion
    fn parse_known_variants(string: &str) -> String {
```

daemon/src/lib.rs
```diff
@@ -447,32 +447,34 @@ where
...
+        match string.to_lowercase().as_str() {
+            "docker" => "Docker".to_string(),
+            "umbrel" => "Umbrel".to_string(),
+            "raspiblitz" => "RaspiBlitz".to_string(),
```
luckysori: ❓ It's not obvious why we need this. What part of the system needs the environment variable to be not-lowercase for these cases?

luckysori: Okay, I think the point is that we need to format certain variants differently, because old versions expect something other than lowercase.

Is this even worth it though? Only the `maker` is consuming this information to gather metrics. I would prefer it if we just "broke" this compatibility for `taker`s, since they don't actually do anything with this field.

luckysori (approved):
LGTM, although I wouldn't bother with backwards-compatibility in this case.

daemon/src/lib.rs
bonomat: If we break it we would need to update the identify protocol  and I didn't want to bother with this :D

luckysori: But I don't think that's true. The taker would just get an `Unknown` variant. And the taker doesn't even care about this information!

bonomat: Unfortunately not, he will get an error saying: can't parse XYZ, expected variants are `RaspiBlitz`, `Umbrel,`, etc...

luckysori: But I just tested removing this code and I get this on the taker:

```
2022-09-29 01:15:44  INFO xtra_actor_request{actor_type=daemon::identify::dialer::Actor message_type=daemon::identify::dialer::IdentifyMsgReceived}:xtra_message_handler:Handle IdentifyMsgReceived: daemon::identify::dialer: New identify message received peer_id=12D3KooWDjzHna3pNi1Bt1DoRfrpsBREykJKXDRDxXvhJNAdDZEk daemon_version=0.6.1 environment=Unknown wire_version=0.3.0
```

and this on the maker:

```
xtra_actor_request{actor_type=daemon::identify::dialer::Actor message_type=daemon::identify::dialer::IdentifyMsgReceived}:xtra_message_handler:Handle IdentifyMsgReceived: daemon::identify::dialer: New identify message received peer_id=12D3KooWBswfzW4GGePjyRGeZ9NHGNTCV4hXofcznpYEUbTKbsTg daemon_version=0.6.2 environment=Binary wire_version=0.3.0
```

bonomat: Mhm, thanks for trying. I'll look into this again.

bonomat: Got it, the only one we need to support is `Unknown`.

PR #3038, opened 2022-09-28
feat(ci): Only run relevant jobs when there were changes to that file

### LINKED ISSUE #2937: Run ci workflow only if file have changed
Currently the whole CI workflow is executed regardless of the change. e.g. if we only change something in the frontend the daemon tests and clippy jobs are also run. We could save a lot of execution time if we'd execute jobs only if the corresponding folder has changed - https://github.com/dorny/paths-filter

klochowicz:
Apart from the main advertised thing (skipping CI jobs when nothing changed), it also removes the maintenance burden of manually updating various files when adding new crate.

includes new meta CI job that tells bors when everything passed.


klochowicz:
we can notice that only the backend jobs have been build now (first commit moved things into `crates` dir)

luckysori (approved):
Still not thrilled about e53334878e506c17bc58aa31acf8ad94bfb268e1, but I like the other changes 😃

klochowicz:
> Still not thrilled about [e533348](https://github.com/itchysats/itchysats/commit/e53334878e506c17bc58aa31acf8ad94bfb268e1), but I like the other changes 😃

honestly, I am most thrilled about potentially stopping bors timeouts (another PR just timed out!) ;D

.github/workflows/ci.yml
```diff
@@ -18,6 +18,27 @@ jobs:
...
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            backend:
+              - 'crates/**'
+              - 'Cargo.toml'
+              - 'Cargo.lock'
+            frontend:
+              - 'taker-frontend/**'
```
bonomat, line 39: you are missing `taker-electron`

klochowicz: fixed, thanks for spotting!

klochowicz: w8 a second, `frontend` job does not build or check electron stuff, so the check should not really trigger it. 

I propose to address it by adding a separate check and job for electron CI stuff (as it was not covered here anyway, so it's not a regression)

edit: I think that you meant that electron should be rebuilt when taker-frontend changes, because we embed it.

klochowicz: I've opened #3050

bonomat: Oh, we actually have 0 checks in taker-electron. So there is nothing to do here :D

bonomat (approved):
I guess I'll survive the project restructuring. If not, I know where to complain 😈

.github/workflows/ci.yml
```diff
@@ -18,6 +18,27 @@ jobs:
       - uses: actions/checkout@v3
       - uses: dprint/check@v2.1
 
+  # job to run change detection
+  changes:
+    runs-on: ubuntu-latest
+    # Set job outputs to values from filter step
+    outputs:
+      backend: ${{ steps.filter.outputs.backend }}
```
holzeis, line 26: 🔧 Why only separate between frontend and backend.. we could even be more specific.

e.g. why build the maker-frontend if only the taker-frontend has changed? etc.

I would create an change output for every sub project we have. It might be cumbersome because of dependencies, but I think that's a complexity worth maintaining.

holzeis (commented):
LGTM 👍 but I think it doesn't go far enough! This change will still build project unnecessarily.

.github/workflows/ci.yml
klochowicz: Happy to accept further patches! 😁 it will be covered by #3050 (that will cover electron changes, and as a prerequisite for that will have to split maker and taker frontends.)

klochowicz:
bors retry

klochowicz:
thanks bors, I knew persistence was the key here. 

for others: I had to tweak bors to make the `CI` summary job "pass" so long as there were no *failures*.

PR #3054, opened 2022-09-29
Log protocols of identified peer

luckysori:
This information is useful if we want to determine if a peer should be blocked.


holzeis (approved):

klochowicz (approved):
👍

PR #3059, opened 2022-09-29
chore: Put shared routes inside shared-bin

klochowicz:
A lot of the routes are exactly the same for both binaries.
There is no reason for such duplication.


holzeis (approved):
LGTM 👍

bonomat (commented):
I'm not convinced this is a good idea. It's just a coincident that the two binaries are sharing the same routes but it's not an architectual decision. 
Differently said, if want that taker-routes and maker-routes always evolve in the same way, then this is the right move, if not, I would not go for this as it just increases code churn unnecessarily. I think the latter one is the case.

klochowicz:
> I'm not convinced this is a good idea. It's just a coincident that the two binaries are sharing the same routes but it's not an architectual decision.
> Differently said, if want that taker-routes and maker-routes always evolve in the same way, then this is the right move, if not, I would not go for this as it just increases code churn unnecessarily. I think the latter one is the case.

we already have separate routes that are maker or taker only (and there's no intention to unify any of them). All I did was to remove the ones that will *not* be different to avoid maintenance burden. 
 
Case in point:
if someone comes along and fixes the authentication on `/metrics` endpoint, they would need to know to fix it in two places, not one. That's unnecessarily risky.

It also is way more clear which routes are maker or taker specific when you read the code, which is a nice bonus (but not essential). 

How come this increases code churn? The other day I had to put the same TODO in two separate places because of code duplication; I find this exactly unnecessary code churn (and fixing this in two places would be such unnecessary churn too).

klochowicz:
@bonomat given my last response/clarification, are you against this going in?
after all, we retain exactly the same flexibility with maker and taker specific routes, and any of the newly shared ones can be extracted to be unique as well. We only prevent doing this *by accident* now; making it impossible to make a mistake and making it clear when we actually desire that behaviour to be unique.

luckysori (commented):
I think this change is okay (although I wouldn't do it), so I won't block it. It does go against what we set out to do with the `shared-bin` crate. According to its readme:

```markdown
This crate should disappear over time so please refrain from adding code here.
```

klochowicz:
> I think this change is okay (although I wouldn't do it), so I won't block it. It does go against what we set out to do with the `shared-bin` crate. According to its readme:
> 
> ```gfm
> This crate should disappear over time so please refrain from adding code here.
> ```

thanks for bringing this up! I strongly believe that it that part of readme should be changed. there are heaps of things that were pointlessly duplicated:
- loggers
- catchers
- fairings
- CLI args

This is just another one of these things. 

I believe the original idea was to hope that we could speed up the build, but was not backed with any measurements and this is not where our build slows down. I will create a separate discussion or PR to make that intent clear.

bonomat:
> @bonomat given my last response/clarification, are you against this going in? after all, we retain exactly the same flexibility with maker and taker specific routes, and any of the newly shared ones can be extracted to be unique as well. We only prevent doing this _by accident_ now; making it impossible to make a mistake and making it clear when we actually desire that behaviour to be unique.

No, It can go in if others think this is a good change. I only wanted to point out that we have high code churn which is a red flag.

da-kami (commented):
Tbh I don't remember what the intent of `shared-bin` was and why the comment not to add code there was added. I think it is legit to have a shared lib for avoiding code duplication. On the other hand I don't think the amount of duplication that we are talking about here justifies a shared lib. I don't really mind much about merging this, but would have probably waited until the annoyance counter is higher. @klochowicz if your annoyance counter is already high on this feel free to merge :)

klochowicz:
bors retry

klochowicz:
bors retry

PR #3064, opened 2022-09-29
feat(electron): pick random port

### LINKED ISSUE #2992: Pick random port for electron app
As itchysats will run on the local machine we might get into conflict with another service running on the same port. By randomely choosing an open port this could be prevented.

**Side note** Would choosing an random open port on startup also allow for the app to run multiple times? I guess there will be at least some issues with the database.

holzeis:
resolves #2992

Using [get-port](https://github.com/sindresorhus/get-port) would have been nicer, but the library is built ESM only and can not be used from our CommonJS project. Hence I followed the title of the ticket and simply picked a random port between 10000 and 65535.

### DIFF
--- a/taker-electron/src/main/main.ts
+++ b/taker-electron/src/main/main.ts
@@ -10,6 +10,7 @@
  */
 import { app, BrowserWindow, net, shell } from "electron";
 import log from "electron-log";
+import nodenet from "node:net";
 import path from "path";
 import MenuBuilder from "./menu";
 import { resolveHtmlPath } from "./util";
@@ -17,6 +18,7 @@ import { resolveHtmlPath } from "./util";
 const { itchysats } = require("../../index.node");
 
 let mainWindow: BrowserWindow | null = null;
+let port = 8000;
 
 if (process.env.NODE_ENV === "production") {
     const sourceMapSupport = require("source-map-support");
@@ -43,8 +45,8 @@ const installExtensions = async () => {
 };
 
 const alive = (timeout: number) => {
-    log.info("Calling url");
-    const request = net.request("http://127.0.0.1:8000");
+    log.info(`Probing if ItchySats is alive at http://127.0.0.1:${port}`);
+    const request = net.request(`http://127.0.0.1:${port}`);
     request.on("response", () => {
         if (!mainWindow) {
             log.error("Main window not defined. Terminating");
@@ -53,7 +55,7 @@ const alive = (timeout: number) => {
         log.log("ItchySats is available!");
         log.debug("Loading ItchySats into browser window.");
         mainWindow
-            .loadURL("http://127.0.0.1:8000")
+            .loadURL(`http://127.0.0.1:${port}`)
             .then(() => {
                 log.info("Successfully loaded ItchySats!");
             })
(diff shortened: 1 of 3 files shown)


taker-electron/src/main/main.ts
```diff
@@ -132,34 +133,34 @@ app.on("window-all-closed", () => {
...
+
+    process.env.ITCHYSATS_ENV = "electron";
+
+    const dataDir = app.isPackaged ? app.getPath("userData") : app.getAppPath();
+    const network = app.isPackaged ? "mainnet" : "testnet";
+
+    const minPort = 10_000;
+    const maxPort = 65_535;
+    // pick and try random port
+    port = Math.floor(Math.random() * (maxPort - minPort + 1) + minPort);
```
klochowicz: I think you should also check whether the picked port is actually free and available to use

holzeis: Like written in the PR - I wanted to use get-port for that, but failed due to a CommonJS restriction (or my lack of understanding). I did not find any other comparable project in community support. I could implement a custom solution for that, though. But I am not sure if its worth it.

edit: I've implemented a custom solution to check if the tested port is available.

taker-electron/src/main/main.ts
```diff
@@ -132,34 +134,64 @@ app.on("window-all-closed", () => {
     }
 });
 
-app
-    .whenReady()
-    .then(async () => {
-        await createWindow();
+// retry checking random port by the given amount of max retries.
+const retry = (maxRetries: number, fn: (port: number) => Promise<number>, port: number): Promise<number> => {
```
bonomat, line 138: 🙃
this could have a better name than just `retry`

bonomat, line 138: Ignore me

bonomat (approved):
Tested locally and works great. The dis/advantage is that you don't know the port anymore so you can't open it in the browser as well :)

taker-electron/src/main/main.ts
```diff
@@ -132,34 +134,64 @@ app.on("window-all-closed", () => {
...
+        server.on("error", reject);
+        log.debug(`Trying port: ${port}`);
+        server.listen({ port, host: "127.0.0.1" }, () => {
+            const { port } = <any> server.address();
+            server.close(() => {
+                log.debug(`Found open port: ${port}!`);
+                resolve(port);
+            });
+        });
+    });
```
klochowicz, line 166: nit: it would be good to extract this into a function to hide the implementation details

taker-electron/src/main/main.ts
```diff
@@ -132,34 +134,64 @@ app.on("window-all-closed", () => {
     }
 });
```
klochowicz, line 135: is the `alive` closure still used anywhere, maybe it can be removed? I don't think it does anything sensible (also: `timeout` is not used there, and `port` should be captured as parameter if it was to be used).

taker-electron/src/main/main.ts
```diff
@@ -10,13 +10,15 @@
...
 import log from "electron-log";
+import nodenet from "node:net";
 import path from "path";
 import MenuBuilder from "./menu";
 import { resolveHtmlPath } from "./util";
 // eslint-disable-next-line import/no-unresolved
 const { itchysats } = require("../../index.node");
 
 let mainWindow: BrowserWindow | null = null;
+let port = 8000;
```
klochowicz, line 21: I think we can remove this assignment as it's confusing? (it changes before first use)
```suggestion
```

klochowicz (approved):
some nits, happy to postpone the tidying to after the release so it's not a blocker.

bonomat:
Going ahead and merge this so that we have it in the next release. 
@holzeis : please address the comments in a follow-up PR. Thank you :) 

bors r+

holzeis:
> Tested locally and works great. The dis/advantage is that you don't know the port anymore so you can't open it in the browser as well :)

The default port electron will try is 7113. if that does not work - it will start to try random ports. So in the most cases you should know the port. But if it's really taken you can still check the logs ;)

taker-electron/src/main/main.ts
holzeis: Yes alive is still used.. It tests periodically if the itchysats service is available.

PR #3065, opened 2022-09-29
Move libp2p_connections_total to endpoint; add a new metric for connections by env/version

### LINKED ISSUE #2843: Move `NUM_LIBP2P_CONNECTIONS_GAUGE` to to the endpoint and create separate mertic for environment/version
At the moment we always increment upon receiving an `identify` message.

We should move this gauge into the `endpoint`, it is weird that it is in the responsibility of `identify` to record this metric. But at the moment the metric depends on additional information (wire version, daemon version, environment) that is not known to the endpoint but only received in the identify message. We can create a separate metric for recording the environment/version.

Restioson:
Fixes #2843.

I cannot yet verify that this works, since I can't seem to get prometheus to return anything for a query of just `libp2p_connections_total`, even on master, even although other metrics work.

### DIFF
--- a/crates/daemon/src/identify/dialer.rs
+++ b/crates/daemon/src/identify/dialer.rs
@@ -22,7 +22,7 @@ pub struct Actor {
 
 impl Actor {
     pub fn new(endpoint: Address<Endpoint>) -> Self {
-        NUM_LIBP2P_CONNECTIONS_GAUGE.reset();
+        LIBP2P_PEER_INFORMATION.reset();
 
         Self {
             endpoint,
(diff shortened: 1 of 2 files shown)


luckysori (approved):
LGTM!

> I cannot yet verify that this works, since I can't seem to get prometheus to return anything for a query of just libp2p_connections_total, even on master, even although other metrics work.

I've never tried to consume metrics locally :(

bonomat:
> Fixes #2843.
> 
> I cannot yet verify that this works, since I can't seem to get prometheus to return anything for a query of just `libp2p_connections_total`, even on master, even although other metrics work.

You can just call http://localhost:8001/api/metrics locally and see if this metric changes as you expect (i.e. if a taker connects and disconnects). No need to run prometheus :)

Restioson:
> You can just call http://localhost:8001/api/metrics locally and see if this metric changes as you expect (i.e. if a taker connects and disconnects). No need to run prometheus :)

It's unfortunately absent from there too

bonomat:
> It's unfortunately absent from there too

If it is missing, it means it was never set. Try to set it to 0 in the constructor (if we have something like this), it should atleast appear then.

bonomat (approved):
Seems to be working as expected: I started maker and taker locally and get this: 

image


taker terminates and it jumps down to 0 again

image

Restioson:
Ah ok, thanks for testing it! I guess in that case it works :)

PR #3066, opened 2022-09-30
feat: Auto-fill generated password on first login

klochowicz:
Before the mandatory change of password, user is not interested in manually
re-typing the old password.

Use localStorage key to check whether user has logged in before at all.

### DIFF
--- a/taker-frontend/src/authentication/useAuth.tsx
+++ b/taker-frontend/src/authentication/useAuth.tsx
@@ -47,6 +47,7 @@ export function AuthProvider({
             .then(({ authenticated, first_login }: Authenticated) => {
                 setAuthenticated(authenticated);
                 setFirstLogin(first_login);
+                localStorage.setItem("oldUser", "true");
             })
             .catch((_error) => {
                 setAuthenticated(false);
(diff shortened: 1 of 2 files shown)


taker-frontend/src/authentication/useAuth.tsx
```diff
@@ -47,6 +47,7 @@ export function AuthProvider({
             .then(({ authenticated, first_login }: Authenticated) => {
                 setAuthenticated(authenticated);
                 setFirstLogin(first_login);
+                localStorage.setItem("oldUser", "true");
```
bonomat, line 50: This won't work, because we always get here even if the user never logged in. Which means, we always set it to true and never show the password.

bonomat: Why didn't you go for a simple tooltip?

bonomat:
superseded by #3067

taker-frontend/src/authentication/useAuth.tsx
klochowicz: because :
1) making the user type the default password manually is an unnecessary hurdle, it's "making the user think"
2) you don't want to display the tooltip / footer whenever the user is using the non-default password; because then it is just uselessly spamming the user

PR #3069, opened 2022-09-30
feat: add electron tray icon

### LINKED ISSUE #3009: Run electron in the background visible only in the tray menu
https://www.electronjs.org/docs/latest/tutorial/tray

bonomat:
For now, it has only two functions: (i) show the app, (ii) close the app.

resolves #3009 

Tested on a mac, maybe someone can test it on Windows? 

Demo:

https://user-images.githubusercontent.com/224613/193192579-6fc2ac9b-8a6d-4afa-9393-a6f135174a02.mov


klochowicz (approved):
nice touch!

taker-electron/src/main/main.ts
```diff
@@ -31,6 +32,28 @@ if (isDebug) {
     require("electron-debug")();
 }
 
+let tray: Tray | undefined = undefined;
```
holzeis: 🔧 I personally keep global variables at the top.. like the mainWindow.

taker-electron/src/main/main.ts
```diff
@@ -31,6 +32,28 @@ if (isDebug) {
...
+function createTray() {
```
holzeis: 🔧 Staying consistent with the other functions, I suggest changing that line to the following.

`const createTray = () => {`

taker-electron/src/main/main.ts
```diff
@@ -74,6 +97,16 @@ const createWindow = async () => {
         await installExtensions();
     }
 
+    if (!tray) { // if tray hasn't been created already.
```
holzeis: 🔧 I think the app should either be in the tray or dock / taskbar.. both is a bit much. At least, when it is in tray then it should not be in my dock / taskbar. We could do that by using `app.dock.hide()` for mac and `win.setSkipTaskbar(true);` for windows.

holzeis (approved):
Minor remark addressing UX. IMHO should be fixed before we release, but doesn't have to block this PR. 👍

taker-electron/src/main/main.ts
bonomat: done

taker-electron/src/main/main.ts
bonomat: I like it. I kept the tray icon though even if the window is there.

bonomat:
Still needs some testing on Windows please. @luckysori ?

luckysori:
> Still needs some testing on Windows please. @luckysori ?

I want to test it, but running `yarn start` results in an endless loop of the app refreshing and I can't interact with it :(

bonomat:
> > Still needs some testing on Windows please. @luckysori ?
> 
> I want to test it, but running `yarn start` results in an endless loop of the app refreshing and I can't interact with it :(

can you do `yarn package`?

bonomat:
@luckysori : I've released it on my fork, can you try this binary? https://github.com/bonomat/hermes/releases/download/v0.9.0/ItchySats-Setup-0.9.0.exe

luckysori:
@bonomat:

> can you do `yarn package`?

Sorry, I missed this! I'll give it a go.

UPDATE: It fails and I think it would actually build for Linux since I'm running it from within WSL.

> I've released it on my fork, can you try this binary?

It doesn't work on Windows :( It just shuts down the app.

bonomat:
> It doesn't work on Windows :( It just shuts down the app.

@klochowicz / @holzeis / @da-kami : what do you think ? It seems to be only working on macOS, shall we still merge it?

klochowicz:
> @bonomat:
> 
> > can you do `yarn package`?
> 
> Sorry, I missed this! I'll give it a go.
> 
> UPDATE: It fails and I think it would actually build for Linux since I'm running it from within WSL.
> 
> > I've released it on my fork, can you try this binary?
> 
> It doesn't work on Windows :( It just shuts down the app.

what is your `platform` variable? isn't this the case that you have a different one than `win32` when you boot it from WSL?

crates/taker-electron/src/main/main.ts
```diff
@@ -127,9 +166,11 @@ const createWindow = async () => {
  */
 
 app.on("window-all-closed", () => {
-    // Respect the OSX convention of having the application in memory even
-    // after all windows have been closed
-    if (process.platform !== "darwin") {
+    if (process.platform === "win32" && mainWindow) {
+        mainWindow.setSkipTaskbar(true);
+    } else if (process.platform === "darwin") {
+        app.dock.hide();
```
klochowicz, line 130: is this one needed? if all windows are closed, hide should have no effect?

klochowicz (approved):
to me this is still a step in the right direction; I vote for adding it.

crates/taker-electron/src/main/main.ts
bonomat: This will hide the `dock` icon when closing the window, i.e. the icon in the app dock normally at the lower end of your screen.

luckysori:
> what is your `platform` variable? isn't this the case that you have a different one than `win32` when you boot it from WSL?

It should be different from WSL, yep. That's what I remember from testing it a few weeks ago. Because now I just can't run `yarn start` successfully.

crates/taker-electron/src/main/main.ts
klochowicz: are you sure we want this behaviour? it's unexpected, in the sense that other apps don't behave like that when their window is hidden on macos.
There's typically another switch in the settings that allows the user to remove the app from the dock.

bonomat: I think it's good practice ? 🤷‍♂️
It was actually feedback from @holzeis [here](https://github.com/itchysats/itchysats/pull/3069#discussion_r984554809).

klochowicz: interesting; from the UX perspective you typically want to go for "principle of least surprise", meaning that it's best to stick to established conventions and that the app should behave the same way as all the other apps on the system.

I tested manually some apps, and they do not disappear from dock on being hidden.
The closest example is probably Telegram, which has exactly the same options as us in the tray - and still does not remove the app from the dock.

taker-electron/src/main/main.ts
klochowicz: at least on macOS, this is the expected behaviour and that's what the apps do. I would recommend to follow established UX conventions rather than making things more clever (but unexpected).
can't say about windows, as I haven't used that one in ages.
PS on macOS, there is typically a switch that removes app from the dock altogether, relying only on the tray icon. We might want to implement that someday.

da-kami (approved):
Happy to see this go in even if it does not work on Windows. We might want to open a ticket to tackle Windows explicitly before adding more functionality to the tray :)

PR #3076, opened 2022-09-30
Cleanup endpoint logic

holzeis:
Follow up PR to the review remarks on https://github.com/itchysats/itchysats/pull/3064

### DIFF
--- a/crates/taker-electron/src/main/endpoint.ts
+++ b/crates/taker-electron/src/main/endpoint.ts
@@ -1,0 +1,62 @@
+import { net } from "electron";
+import log from "electron-log";
+import { AddressInfo } from "net";
+import nodenet from "node:net";
+
+const DEFAULT_PORT = 8000;
+const MAX_RETRIES = 5;
+const TIMEOUT = 500; // in miliseconds
+const MIN_PORT = 10_000;
+const MAX_PORT = 65_535;
+const HOST = "127.0.0.1";
+
+export class Endpoint {
+    public url: string;
+
+    constructor(public port: number) {
+        this.url = `http://${HOST}:${port}`;
+    }
+
+    wait(onReady: (endpoint: string) => void) {
+        const alive = (timeout: number) => {
+            log.info(`Probing if ItchySats is alive at ${this.url}`);
+            const request = net.request(this.url);
+            request.on("response", () => onReady(this.url));
+            request.on("error", (error) => {
+                log.warn("Could not connect to ItchySats.");
+                log.error(error);
+                setTimeout(alive, timeout, timeout + timeout);
+            });
+            request.end();
+        };
+
+        setTimeout(alive, TIMEOUT, TIMEOUT);
+    }
+
+    static create(maxRetries = MAX_RETRIES, port = DEFAULT_PORT): Promise<Endpoint> {
+        // retry checking random port by the given amount of max retries.
+        return new Promise<Endpoint>((resolve, reject) => {
+            const server = nodenet.createServer();
+            server.unref();
+            server.on("error", (err) => {
+                log.info(`Port: ${port} is not available retrying another random port. Retries: ${maxRetries}`);
+                if (maxRetries <= 0) {
+                    log.error(`Reached max amount of retries. quitting.`);
+                    return reject(err);
+                }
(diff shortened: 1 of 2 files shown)


taker-electron/src/main/endpoint.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    wait(onReady: (endpoint: string) => void) {
+        const alive = (timeout: number) => {
+            log.info(`Probing if ItchySats is alive at ${this.url}`);
+            const request = net.request(this.url);
+            request.on("response", () => onReady(this.url));
+            request.on("error", (error) => {
+                log.warn("Could not connect to ItchySats.");
+                log.error(error);
+                setTimeout(alive, timeout, timeout + timeout);
```
da-kami: 🤔 I find the usage of `setTimeout` a bit confusing here. In this line you set it with three parameters, one being the double timeout. At the end of the function (line 27) you set it with four parameters. It's not obvious for me why or what this code achieves. Is there a more intuitive way to do this?

taker-electron/src/main/endpoint.ts
```diff
@@ -0,0 +1,69 @@
...
+                    const { port } = <any> server.address();
+                    server.close(() => {
+                        log.debug(`Found open port: ${port}!`);
+
+                        resolve(new Endpoint(port));
+                    });
+                });
+            });
+        };
+        return retry(5, checkAvailablePort, DEFAULT_PORT);
```
da-kami: 🙃 It might be worth extracting the nr of retries, timeout secs into constants so the code is more readable.

taker-electron/src/main/endpoint.ts
```diff
@@ -0,0 +1,69 @@
...
+            const minPort = 10_000;
+            const maxPort = 65_535;
+
+            return fn(port).catch(function(err) {
+                log.info(`Port: ${port} is not available retrying another random port. Retries: ${maxRetries}`);
+                if (maxRetries <= 0) {
+                    log.error(`Reached max amount of retries. quitting.`);
+                    throw err;
+                }
+                return retry(maxRetries - 1, fn, Math.floor(Math.random() * (maxPort - minPort + 1) + minPort));
```
da-kami: ❓ Here we pass the function to be called and the value to be used in. Why this level of complexity? Are we planning to make the function that checks the available port to be configurable? 
A simple function call to `checkAvailablePort` should be good enough here or am I missing something?

da-kami (approved):
Did you look for existing libraries that could handle picking a port for us? 
Would be great to arrange the code in a way so we can test the boundaries of the port-picking logic with a unit test.

taker-electron/src/main/endpoint.ts
```diff
@@ -0,0 +1,69 @@
...
+            request.on("response", () => onReady(this.url));
+            request.on("error", (error) => {
+                log.warn("Could not connect to ItchySats.");
+                log.error(error);
+                setTimeout(alive, timeout, timeout + timeout);
+            });
+            request.end();
+        };
+
+        setTimeout(alive, 500, 500, 500);
```
klochowicz: it also would be good to extract the (arbitrary?) `500` to a const - is this in miliseconds?

klochowicz (approved):

taker-electron/src/main/endpoint.ts
holzeis: yes this is in miliseconds.

taker-electron/src/main/endpoint.ts
holzeis: actually the 4 parameters was one too many.. I've fixed that. The first timeout is the amount of time that should be waited until alive will be called. the second timeout represents how much the the next alive check should wait. This way we a debouncing the alive checks and prevent endless http requests. e.g. starting with 500ms the interval would look like this. 1. 500, 2. 1000, 3. 2000, 4. 4000, 5. 8000, ...

Its a recursive function, I guess that makes it a bit more confusing.

taker-electron/src/main/endpoint.ts
holzeis: fair enough, I simplified the code. 🙂

PR #3077, opened 2022-09-30
Remove action to add electron packages to `0.7.0`

da-kami:
`electron-builder publish` is unable to add artifacts to an existing release, but creates a new release. Trying to add to an existing release will resuly in the built artifacts being skipped when trying to upload them:

```
skipped publishing  file=ItchySats-0.7.0-arm64.dmg reason=existing release published more than 2 hours ago tag=v0.7.0 version=0.7.0 date=Fri Sep 30 2022 05:55:48 GMT+0000 (Coordinated Universal Time)
```

We overcame that by using `upload-release-asset@v1.0.2` to upload the release binaries, but that action requires the uload URL of the release, which is only available when creating a release. As a quick-fix we just used the hardcoded upload URL for `0.7.0` to attach the electron artifacts, but we cannot easily adapt this action to be re-usable. Thus, we have to re-use it and fix the release process.


da-kami:
closing this for now because of #3078

PR #3085, opened 2022-10-02
Taker-electron eslint and ci

### LINKED ISSUE #3050: `taker-electron` typescript should have a CI job
`taker-electron` typescript project should be built by CI as a separate CI job. 

The idea is to monitor for its changes separately (`taker-frontend` dir and `taker-electron` dir), and only run it after `taker-frontend`.

Additionally, its dependencies should be checked by dependabot.

blocked by #3038

bonomat:


crates/taker-electron/src/main/main.ts
```diff
@@ -14,6 +14,7 @@ import nodenet from "node:net";
 import path from "path";
 import MenuBuilder from "./menu";
 import { resolveHtmlPath } from "./util";
+import {AddressInfo} from "net";
```
luckysori: 🔧 Missing spaces.

luckysori (approved):

bonomat:
@klochowicz : can you have a look please: I think it should run the `electron` ci job but it doesn't.

klochowicz:
> @klochowicz : can you have a look please: I think it should run the `electron` ci job but it doesn't.

pushed a small commit that should fix it; feel free to squash it if you deem it not worthy being a separate one :) 

confirmed: it works (as seen on the CI below)

klochowicz (approved):
concept ACK; happy for it to go in as long as it passes the CI (the newly-enabled job still fails)

klochowicz:
it still fails on `yarn test`... we can either add it  to the `taker-electron`, or skip that step for that package altogether?

bonomat:
> it still fails on `yarn test`... we can either add it to the `taker-electron`, or skip that step for that package altogether?

There are not tests in `taker-electron` 🙈

PR #3094, opened 2022-10-04
docs: added dev dependencies

holzeis:
I've added the two requirements that need to be installed that are not obvious. however, we could also think about adding rust as requirement to be installed in the version we are using.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,11 @@ To start the local dev-environment all the components can be started at once by 
 ./start_all.sh
 ```
 
-Note: Before first run, you need to run `cd maker-frontend; yarn install; cd../taker-frontend; yarn install` command to ensure that all dependencies get
-installed.
+Note: Before first run, you need to install all dependencies
+
+- Protocol buffers compiler e.g. `brew install protobuf` // see https://docs.rs/prost-build/latest/prost_build/#sourcing-protoc for more details.
+- CMake e.g. `brew install cmake`
+- Maker and taker frontend `cd maker-frontend; yarn install; cd../taker-frontend; yarn install`
 
 The script combines the logs from all binaries inside a single terminal so it
 might not be ideal for all cases, but it is convenient for quick regression testing.


README.md
```diff
@@ -67,8 +67,11 @@ To start the local dev-environment all the components can be started at once by
 ./start_all.sh
 ```
 
-Note: Before first run, you need to run `cd maker-frontend; yarn install; cd../taker-frontend; yarn install` command to ensure that all dependencies get
-installed.
+Note: Before first run, you need to install all dependencies
+
+- `cd maker-frontend; yarn install; cd../taker-frontend; yarn install`
+- `brew install protobuf` // see https://docs.rs/prost-build/latest/prost_build/#sourcing-protoc for more details.
+- `brew install cmake` // adapt accordingly to your os.
```
klochowicz: this is macOC specific; I doubt that most of the github audience is.

maybe provide macOS and linux variants? (or just mention that `protobuf` and `cmake` need to be installed?)

holzeis: interesting.. what OS do you think the github audience uses the most? :) I feel like there majority runs on macOS, but I might be living in a bubble 😅.

holzeis: since I am not developing on linux, I am not sure of the linux (or windows for that matter) variants - so I am only going to mention it and add the brew command as reference.

klochowicz: > interesting.. what OS do you think the github audience uses the most? :) I feel like there majority runs on macOS, but I might be living in a bubble 😅.

then you certainly live in a bubble 😂 I would have said "Linux", but latest SO dev survey proves us both wrong.  Apparently most devs use Windows - source:  https://insights.stackoverflow.com/survey/2021#section-most-popular-technologies-operating-system
not that surprising, considering github is owned by Microsoft and WSL got better over the years.

klochowicz: I'd probably say it's enough to just mention that `yarn, `rust` `protoc` and `cmake` are dev dependencies. a dev should know how to get them :p

klochowicz (approved):
step in the right direction, so might as well go in. I might generalise it later for other platforms.

PR #3105, opened 2022-10-05
Update dependencies on all frontends

klochowicz:
1. Update yarn.lock on maker-frontend (has been out of dependabot for a long time). I've built and tested creating an offer, still works.
2. Add `maker-frontend` and `taker-electron` to dependabot monitoring

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,3 +24,15 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+
+  - package-ecosystem: "npm"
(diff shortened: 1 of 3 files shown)


bonomat (approved):
Even more dependabot spam :/

klochowicz:
> Even more dependabot spam :/

we can consider alternatives! 

> Even more dependabot spam :/

I would probably not have dared to do so if we still needed to build & test the backend everytime when a JS dependency upgrades 😬

bonomat:
I think we need to turn this down... 

image

PR #3127, opened 2022-10-06
chore: dependabot is spam

bonomat:
We don't have to stay up-2-date for all dependencies at all time. Dependabot will give us security alerts no matter what open-pull-requests-limit says, hence, I changed it to 1 update per week.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,15 +24,3 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
-
-  - package-ecosystem: "npm"
-    directory: "/maker-frontend"
-    schedule:
-      interval: "daily"
-    open-pull-requests-limit: 10
-
-  - package-ecosystem: "npm"
-    directory: "/crates/taker-electron"
-    schedule:
-      interval: "daily"
-    open-pull-requests-limit: 10


klochowicz:
this looks like a knee-jerk reaction to the "spam" caused by the fact that we didn't update dependencies in maker-frontend for ages in the first place. maybe we could wait and see how annoying it is when the baseline is up-to-date?

IMHO a better solution would be apply auto-merging - if the CI passes, merge things automatically (at least for the frontend dependencies this would solve a lot of pointless churn and remove the manual step from the process altogether).

klochowicz:
alternatively, let's just revert enabling `maker-frontend` if we don't need it worthy of regular updates, as this was the immediate cause of this PR. 
I find it much easier solution.

bonomat:
> IMHO a better solution would be apply auto-merging - if the CI passes, merge things automatically (at least for the frontend dependencies this would solve a lot of pointless churn and remove the manual step from the process altogether).

For me this gets a simple no. In the danger of making this a religious discussion now: automatically updating dependencies for the sake of having them up2date has not much value to me. 

I think there are 3 reasons why we want a new version of a dependency: 

1. it fixes a security issue: those we will still get even if we set `open-pull-requests-limit = 0`
2. if fixes a bug
3. it adds an additional feature/improvement which we are keen on using

in any other case, I see dependencies updates as spam, they block our CI and unnecessarily take away our time to work on something that is more important.

klochowicz:
> > IMHO a better solution would be apply auto-merging - if the CI passes, merge things automatically (at least for the frontend dependencies this would solve a lot of pointless churn and remove the manual step from the process altogether).
> 
> For me this gets a simple no. In the danger of making this a religious discussion now: automatically updating dependencies for the sake of having them up2date has not much value to me.

would you mind to elaborate so I could better understand your reasons?
Do you consider it harmful or introducing more risk or just "not worth it"? (configuring the latter is 30 mins once) 
> 
> I think there are 3 reasons why we want a new version of a dependency:
> 
> 1. it fixes a security issue: those we will still get even if we set `open-pull-requests-limit = 0`

this deals only with disclosed CVEs in the wild - do you trust that all security holes get reported, or a lot of them are being patched straight by the devs if they spot them first?
Moreover, *patch* releases contain only bugfixes, so by definition auto-merging them would apply to 1)

> 2. if fixes a bug

not having them raised by dependabot in the first place makes it hard to notice whether a bug was in the software. We risk spending our time chasing a bug only to discover it was from a dependency and it got fixed in the meantime. 

> 3. it adds an additional feature/improvement which we are keen on using
again, we don't know about the features / improvements if we're not notified about them; dependabot tells us about that.
> 
> in any other case, I see dependencies updates as spam, they block our CI and unnecessarily take away our time to work on something that is more important.

the frontend updates only block CI  for few minutes thanks to the recent changes, so I consider this a strawman argument (until last week this would have clogged our CI).
Also, automatically updating minor and patch versions will actually save our time in the long run; it is easier to keep software tidy and up-to-date.

bonomat:
> would you mind to elaborate so I could better understand your reasons? Do you consider it harmful or introducing more risk or just "not worth it"? (configuring the latter is 30 mins once)

I tried with the 3 reasons below.

> 
> > I think there are 3 reasons why we want a new version of a dependency:
> > 
> > 1. it fixes a security issue: those we will still get even if we set `open-pull-requests-limit = 0`
> 
> this deals only with disclosed CVEs in the wild - do you trust that all security holes get reported, or a lot of them are being patched straight by the devs if they spot them first? Moreover, _patch_ releases contain only bugfixes, so by definition auto-merging them would apply to 1)
> 

No, but I also don't trust devs to adhere to semantic versioning nor that they introduce new bugs in later versions.

> > 2. if fixes a bug
> 
> not having them raised by dependabot in the first place makes it hard to notice whether a bug was in the software. We risk spending our time chasing a bug only to discover it was from a dependency and it got fixed in the meantime.

A version we have been using for a while is less likely to have 0-day bugs than a version which was randomly updated without us even noticing. I remember at least one case where I debugged for a while before I found that a version update broke our frontend. So your argument of `chasing a bug only to discover it was from a dependency` goes both ways and I would argue it's in my favor. 

> 
> > 3. it adds an additional feature/improvement which we are keen on using
>    again, we don't know about the features / improvements if we're not notified about them; dependabot tells us about that.

If we don't know about it and we don't need it, is there any loss?

> > 
> > in any other case, I see dependencies updates as spam, they block our CI and unnecessarily take away our time to work on something that is more important.
> 
> the frontend updates only block CI for few minutes thanks to the recent changes, so I consider this a strawman argument (until last week this would have clogged our CI). 

In german we have this saying: `small livestock also makes crap` (thanks google 😅) .. the small PRs you are talking about add up and are not to be neglected: it runs once on CI (and maybe fust) but then each of us has to update the dependencies locally which adds up. 

> Also, automatically updating minor and patch versions will actually save our time in the long run; it is easier to keep software tidy and up-to-date.

I disagree with this: if a dependency update does not add any value (no bug fix, no security fix, no new feature) it is wasted time and we probably should not need to upgrade it at all.

klochowicz:
thank you for elaborating. 
All in all, I respect your decision and will stop caring about our dependencies until explicitly tasked with such a thing. 

I agree with your point that there is a non-zero risk of breaking some functionality with updates (and we have examples of this happening with chakra version update). However, IMHO the proper solution is to write tests that cover this so nothing that we care about breaks - we were hit by this because we didn't seem that UI tests were a good use of our time.

> A version we have been using for a while is less likely to have 0-day bugs than a version which was randomly updated without us even noticing.

I agree to this in terms of bugs in software behaviour (after all, we tested it longer); however the opposite is true in regards to security exploits.

bonomat:
> thank you for elaborating. All in all, I respect your decision and will stop caring about our dependencies until explicitly tasked with such a thing.

After re-reading what I wrote I realized that some wordings were unnecessary harsh. Please excuse, it was not my intention to shut down the discussion. 


> I agree with your point that there is a non-zero risk of breaking some functionality with updates (and we have examples of this happening with chakra version update). However, IMHO the proper solution is to write tests that cover this so nothing that we care about breaks - we were hit by this because we didn't seem that UI tests were a good use of our time.
> 

I completely agree, ideally we would have tests to detect everything. The problem I was talking about was a coloring issue where an upgrade broke our tooltips due to a black background color - I'm not sure if it is possible or wanted to have tests on this level. 

> > A version we have been using for a while is less likely to have 0-day bugs than a version which was randomly updated without us even noticing.
> 
> I agree to this in terms of bugs in software behaviour (after all, we tested it longer); however the opposite is true in regards to security exploits.

Fair point. I guess the ultimate solution would be to have the least amount of external dependencies.

klochowicz:
> > thank you for elaborating. All in all, I respect your decision and will stop caring about our dependencies until explicitly tasked with such a thing.
> 
> After re-reading what I wrote I realized that some wordings were unnecessary harsh. Please excuse, it was not my intention to shut down the discussion.

all good 👍

> I completely agree, ideally we would have tests to detect everything. The problem I was talking about was a coloring issue where an upgrade broke our tooltips due to a black background color - I'm not sure if it is possible or wanted to have tests on this level.
> 

I agree that it would be hard to test on that level and most probably not wanted; it's not a critical path, no user would stop using our product due to tooltips being improperly rendered.

> Fair point. I guess the ultimate solution would be to have the least amount of external dependencies.

been there, done that; you end up needing to build all the features yourself, massively slowing down your dev velocity. I'd take this minor annoyance of dependabot any day over that ;-)

bonomat:
@klochowicz : I've started a discussion around this topic here https://github.com/itchysats/itchysats/discussions/3152. 

I would like to silence dependabot though. Would you like to see any changes here or can we merge this until we come up with a different solution in https://github.com/itchysats/itchysats/discussions/3152?

klochowicz:
> @klochowicz : I've started a discussion around this topic here #3152.
> 
> I would like to silence dependabot though. Would you like to see any changes here or can we merge this until we come up with a different solution in #3152?

I was worried by this PR being a step in the wrong direction, so I outlined 3 distinct alternatives to it earlier.  I'll quickly recap:
- removing `maker-frontend` from dependabot, as this was the direct cause of this PR (and we don't care about it as much, it's not used in the wild),
- waiting a bit until the baseline of newly enabled `taker-electron` and `maker-frontend`  stabilises - I have volunteered to deal with the flurry of the bors PRs, but was instructed to not waste time on this;
- auto-merging patch (and perhaps minor) versions if all tests pass, so that we don't need to see uninteresting events, whilst reducing the amount of things we're looking at.

I also outlined that these frontend changes now are a breeze to bors due to the job selectivity introduced recently. 

I don't have anything more to add really. However, I am curious to see what the rest of the team thinks and will accept the new status quo if we do not wish to care about updating the dependencies as much anymore.

bonomat:
> waiting a bit until the baseline of newly enabled taker-electron and maker-frontend stabilises - I have volunteered to deal with the flurry of the bors PRs, but was instructed to not waste time on this;

I've updated the PR with removing these two for now.  
 
Let's discuss the rest in https://github.com/itchysats/itchysats/discussions/3152 please.

klochowicz (approved):
👍 
happy to continue working towards a sustainable solution in the discussion

PR #3130, opened 2022-10-06
Fix build release binary workflow

klochowicz:
`protoc` is not bundled anymore; we fixed other CI jobs but missed this one.

This prevents us from creating Docker images of latest master.

the job was failing for a while now: https://github.com/itchysats/itchysats/actions/workflows/docker-push-master.yml


bonomat (approved):
thanks for fixing this. Let's merge this rightaway.

PR #3136, opened 2022-10-07
fix(oracle): Use a timeout for all GET requests from Olivia

klochowicz:
Globally configured request timeout avoids needing to worry about timeouts
whenever we use `client.get()`.
In this example, timeout was applied when fetching pending attestations, but was
missing for announcements.


luckysori (approved):
LGTM!

PR #3159, opened 2022-10-10
fix: taker-electron window width was too small

bonomat:
Before

image

After

image


bonomat:
We could probably also fix it in the frontend but I thought this is easier 😬

holzeis (approved):

klochowicz (approved):

PR #3167, opened 2022-10-10
chore: Add convenience script that builds latest versions of the frontends

klochowicz:
Refer in the docs to the convenience script

### DIFF
--- a/build_frontends.sh
+++ b/build_frontends.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+echo "Install latest dependencies and build both maker and taker frontends"
+
+cd maker-frontend
+yarn install
+yarn build
+cd ..
+cd taker-frontend
+yarn install
+yarn build
+cd ..
+echo "Maker and taker frontends built successfully"
--- a/README.md
+++ b/README.md
@@ -98,7 +98,6 @@ Note: Before first run, you need to install all dependencies
 
 - Protocol buffers compiler e.g. `brew install protobuf` // see https://docs.rs/prost-build/latest/prost_build/#sourcing-protoc for more details.
 - CMake e.g. `brew install cmake`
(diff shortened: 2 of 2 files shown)


klochowicz:
after getting bitten too many times by the changing frontend, I realised that I really needed some simple way to ensure that both frontends are up to date when I'm testing the app. Happy to just keep it in my personal `gitignore`, but I reckon this might be of use to other devs too.

build_frontends.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cd maker-frontend
+yarn install
+yarn build
+cd ..
+cd taker-frontend
+yarn install
+yarn build
+cd ..
+echo "Maker and taker frontends built successfully"
```
holzeis, line 15: nit: This line may lie... 😄 If you really want to print a success message you should look at the result of the commands above 😉

holzeis (approved):

holzeis:
🤔 How about a version check on the UI and taker daemon? if these don't match we could show a warning on the frontend - similar on how we do it when the daemon versions do not match. But that might be an overkil..

build_frontends.sh
klochowicz: sorry, how can this lie? I'm using `set -e`, which will stop this script on error

https://www.delftstack.com/howto/linux/bash-set-e/

klochowicz:
> 🤔 How about a version check on the UI and taker daemon? if these don't match we could show a warning on the frontend - similar on how we do it when the daemon versions do not match. But that might be an overkil..

I was toying with augmenting `build.rs` for taker and maker, and recompiling the frontend then. Unfortunately `yarn build` takes 10s even if nothing changes, so that's just more annoying than it's worth.... :|

build_frontends.sh
holzeis: Ups, I've overlooked that..  🙈

PR #3169, opened 2022-10-11
fix: pass in gh token

bonomat:
It looks like we need to pass in the github token as otherwise it is not available

untested 🙈

### DIFF
--- a/.github/actions/build-release-binary/action.yml
+++ b/.github/actions/build-release-binary/action.yml
@@ -4,6 +4,9 @@ inputs:
   target:
     description: "The Rust target to build for."
     required: true
+  gh_token:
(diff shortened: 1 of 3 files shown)


.github/actions/build-release-binary/action.yml
```diff
@@ -4,6 +4,9 @@ inputs:
   target:
     description: "The Rust target to build for."
     required: true
+  gh_token:
```
klochowicz, line 7: thanks for the example, I have learned how to pass new  args to actions today!

klochowicz (approved):
cool, hopefully this fixes the workflow!

da-kami (approved):

PR #3173, opened 2022-10-12
feat(taker): import seed

### LINKED ISSUE #2989: Restore from seed file
Follow-up ticket from #1065: We should also be able to restore from a seed file.

holzeis:
- allows to import a seed from a file.
- only supports a 256 bytes seed file - umbrel seeds are actively rejected.
- splits the identity from the wallet seed for taker and maker.
- extended the wallet info event for a managed_wallet flag, indicating if the seed was externally provided or generated by ItchySats.
- creates an backup file of the old seed file before overwriting it with the imported one.

import-seed

resolves #2989 

see also https://github.com/itchysats/itchysats/discussions/3048

### DIFF
--- a/crates/bdk-ext/src/lib.rs
+++ b/crates/bdk-ext/src/lib.rs
@@ -6,6 +6,7 @@ use bdk::bitcoin::secp256k1::SecretKey;
 use bdk::bitcoin::util::bip32::ExtendedPrivKey;
 use bdk::bitcoin::Amount;
 use bdk::bitcoin::Network;
+use bdk::database::BatchDatabase;
 use bdk::BlockTime;
 use rand::CryptoRng;
 use rand::RngCore;
@@ -17,6 +18,20 @@ pub fn new_test_wallet(
     utxo_amount: Amount,
     num_utxos: u8,
 ) -> Result<bdk::Wallet<bdk::database::MemoryDatabase>> {
+    new_test_wallet_from_database(
+        rng,
+        utxo_amount,
+        num_utxos,
+        bdk::database::MemoryDatabase::new(),
+    )
+}
+
+pub fn new_test_wallet_from_database<DB: BatchDatabase>(
+    rng: &mut (impl RngCore + CryptoRng),
+    utxo_amount: Amount,
+    num_utxos: u8,
+    mut database: DB,
+) -> Result<bdk::Wallet<DB>> {
     use bdk::populate_test_db;
     use bdk::testutils;
 
@@ -25,8 +40,6 @@ pub fn new_test_wallet(
 
     let key = ExtendedPrivKey::new_master(Network::Regtest, &seed)?;
     let descriptors = testutils!(@descriptors (&format!("wpkh({key}/*)")));
-
-    let mut database = bdk::database::MemoryDatabase::new();
 
     for index in 0..num_utxos {
         populate_test_db!(
--- a/crates/daemon-tests/tests/main/import_seed.rs
+++ b/crates/daemon-tests/tests/main/import_seed.rs
@@ -1,0 +1,42 @@
+use daemon::bdk::bitcoin;
+use daemon::seed;
+use daemon::seed::RandomSeed;
+use daemon::seed::Seed;
+use daemon_tests::open_cfd;
+use daemon_tests::start_both;
+use daemon_tests::OpenCfdArgs;
+use otel_tests::otel_test;
+use rand::distributions::Alphanumeric;
+use rand::thread_rng;
+use rand::Rng;
+use std::env;
+use std::path::Path;
+
+#[otel_test]
+async fn fail_to_import_seed_with_open_cfds() {
+    let (mut maker, mut taker) = start_both().await;
+
+    let cfd_args = OpenCfdArgs::default();
+    open_cfd(&mut taker, &mut maker, cfd_args.clone()).await;
+
+    let random_dir: String = thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(7)
+        .map(char::from)
+        .collect();
+    let data_dir = env::temp_dir().join(Path::new(&random_dir));
+    tokio::fs::create_dir_all(&data_dir)
+        .await
+        .expect("failed to create random temp folder");
+
+    taker
+        .system
+        .import_seed(
+            RandomSeed::default().seed(),
+            data_dir,
+            bitcoin::Network::Testnet,
+            seed::TAKER_WALLET_SEED_FILE,
+        )
+        .await
+        .expect_err("import seed should be rejected as open CFDs exist");
+}
--- a/crates/daemon/src/lib.rs
+++ b/crates/daemon/src/lib.rs
@@ -1,6 +1,7 @@
 #![cfg_attr(not(test), warn(clippy::unwrap_used))]
 
 use crate::bitcoin::util::psbt::PartiallySignedTransaction;
+use crate::bitcoin::Network;
 use crate::bitcoin::Txid;
 use crate::listen_protocols::TAKER_LISTEN_PROTOCOLS;
 use anyhow::bail;
@@ -31,6 +32,7 @@ use ping_pong::ping;
 use ping_pong::pong;
 use seed::Identities;
 use std::collections::HashSet;
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 use time::ext::NumericalDuration;
@@ -78,7 +80,7 @@ pub const N_PAYOUTS: usize = 200;
 
 pub struct TakerActorSystem<O, W, P> {
     pub cfd_actor: Address<taker_cfd::Actor>,
-    wallet_actor: Address<W>,
+    pub wallet_actor: Address<W>,
     _oracle_actor: Address<O>,
     pub auto_rollover_actor: Address<auto_rollover::Actor>,
     pub price_feed_actor: Address<P>,
@@ -93,6 +95,8 @@ pub struct TakerActorSystem<O, W, P> {
     pub identify_info_feed_receiver: watch::Receiver<Option<PeerInfo>>,
 
     _tasks: Tasks,
+
+    db: sqlite_db::Connection,
 }
 
 impl<O, W, P> TakerActorSystem<O, W, P>
@@ -105,6 +109,7 @@ where
     W: Handler<wallet::BuildPartyParams, Return = Result<maia_core::PartyParams>>
         + Handler<wallet::Sign, Return = Result<PartiallySignedTransaction>>
         + Handler<wallet::Withdraw, Return = Result<Txid>>
+        + Handler<wallet::ImportSeed, Return = Result<bdk::wallet::AddressInfo>>
         + Handler<wallet::Sync, Return = ()>
         + Actor<Stop = ()>,
     P: Handler<
@@ -333,7 +338,7 @@ where
         let close_cfds_actor = archive_closed_cfds::Actor::new(db.clone())
             .create(None)
             .spawn(&mut tasks);
-        let archive_failed_cfds_actor = archive_failed_cfds::Actor::new(db)
+        let archive_failed_cfds_actor = archive_failed_cfds::Actor::new(db.clone())
             .create(None)
             .spawn(&mut tasks);
 
@@ -354,6 +359,7 @@ where
             _online_status_actor: online_status_actor,
             _pong_actor: pong_address,
             _identify_dialer_actor: identify_dialer_actor,
+            db,
         })
     }
 
@@ -443,6 +449,36 @@ where
     #[instrument(skip(self), err)]
     pub async fn sync_wallet(&self) -> Result<()> {
         self.wallet_actor.send(wallet::Sync).await?;
+        Ok(())
+    }
+
+    #[instrument(skip(self, seed), err)]
+    pub async fn import_seed(
+        &self,
+        seed: Vec<u8>,
+        data_dir: PathBuf,
+        network: Network,
+        name: &str,
+    ) -> Result<()> {
+        let open_cfd_ids = self.db.load_open_cfd_ids().await?;
+
+        if !open_cfd_ids.is_empty() {
+            use anyhow::bail;
+            bail!("Cannot accept imported seed since open CFDs exist");
+        }
+
+        // recreate the wallet within the wallet actor
+        self.wallet_actor
+            .send(wallet::ImportSeed {
+                seed,
+                path: data_dir,
+                network,
+                name: name.to_string(),
+            })
+            .await??;
+
+        self.wallet_actor.send(wallet::Sync).await?;
+
         Ok(())
     }
 }
--- a/crates/daemon/src/seed.rs
+++ b/crates/daemon/src/seed.rs
@@ -14,6 +14,15 @@ use std::fmt;
 use std::fmt::Debug;
 use std::path::Path;
 
+pub const TAKER_WALLET_SEED_FILE: &str = "taker_seed";
+pub const TAKER_IDENTITY_SEED_FILE: &str = "taker_id_seed";
+
+pub const MAKER_WALLET_SEED_FILE: &str = "maker_seed";
+pub const MAKER_IDENTITY_SEED_FILE: &str = "maker_id_seed";
+
+pub const RANDOM_SEED_SIZE: usize = 256;
+pub const APP_SEED_SIZE: usize = 32;
+
 /// Struct containing keys for both legacy and libp2p connections.
 ///
 /// It is located here as all the information is derived from the seed.
@@ -32,6 +41,8 @@ impl Identities {
 
 pub trait Seed {
     fn seed(&self) -> Vec<u8>;
+
+    fn is_managed(&self) -> bool;
 
     fn derive_extended_priv_key(&self, network: Network) -> Result<ExtendedPrivKey> {
         let mut ext_priv_key_seed = [0u8; 64];
@@ -82,11 +93,14 @@ pub trait Seed {
 }
 
 #[derive(Copy, Clone)]
-pub struct RandomSeed([u8; 256]);
+pub struct RandomSeed([u8; RANDOM_SEED_SIZE]);
 
 impl Seed for RandomSeed {
     fn seed(&self) -> Vec<u8> {
         self.0.to_vec()
+    }
+    fn is_managed(&self) -> bool {
+        true
     }
 }
 
@@ -133,9 +147,15 @@ impl RandomSeed {
     }
 }
 
+impl From<[u8; RANDOM_SEED_SIZE]> for RandomSeed {
+    fn from(bytes: [u8; RANDOM_SEED_SIZE]) -> Self {
+        Self(bytes)
+    }
+}
+
 impl Default for RandomSeed {
     fn default() -> Self {
-        let mut seed = [0u8; 256];
+        let mut seed = [0u8; RANDOM_SEED_SIZE];
         rand::thread_rng().fill(&mut seed);
 
         Self(seed)
@@ -143,16 +163,19 @@ impl Default for RandomSeed {
 }
 
 #[derive(Copy, Clone)]
-pub struct AppSeed([u8; 32]);
+pub struct AppSeed([u8; APP_SEED_SIZE]);
 
 impl Seed for AppSeed {
     fn seed(&self) -> Vec<u8> {
         self.0.to_vec()
     }
+    fn is_managed(&self) -> bool {
+        false
+    }
 }
 
-impl From<[u8; 32]> for AppSeed {
-    fn from(bytes: [u8; 32]) -> Self {
+impl From<[u8; APP_SEED_SIZE]> for AppSeed {
+    fn from(bytes: [u8; APP_SEED_SIZE]) -> Self {
         Self(bytes)
     }
 }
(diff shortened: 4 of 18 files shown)


crates/daemon/src/seed.rs
```diff
@@ -30,6 +30,12 @@ impl Identities {
     }
 }
 
+pub const TAKER_WALLET_SEED_FILE: &str = "taker_seed";
+pub const TAKER_IDENTITY_SEED_FILE: &str = "taker_id_seed";
```
bonomat: This is unused.

crates/daemon/src/wallet.rs
```diff
@@ -512,10 +579,42 @@ mod tests {
...
         }
     }
 
+    impl Actor<ElectrumBlockchain, Tree> {
+        pub async fn spawn_test(
+            data_dir: PathBuf,
+        ) -> Result<(xtra::Address<Self>, watch::Receiver<Option<WalletInfo>>)> {
+            // generate empty export seed
+            let mut export_seed = [0u8; 256];
+            thread_rng().fill(&mut export_seed);
```
bonomat: Given this test will run now regularly we should check if this seed indeed does not hold any Bitcoin 🤑

crates/daemon/src/wallet.rs
```diff
@@ -512,10 +579,42 @@ mod tests {
                     time_to_lock,
                 },
                 blockchain_client: (),
+                db: None,
             })
         }
     }
 
+    impl Actor<ElectrumBlockchain, Tree> {
+        pub async fn spawn_test(
```
bonomat: I'm not sure this is a good idea: this will run an online test against testnet using public infrastructure. 
Besides that it can have random side effects if the server is down, it might slow down our test executions. 

Do you think we can use our [test wallet](https://github.com/itchysats/itchysats/blob/master/crates/bdk-ext/src/lib.rs#L14) for this?

taker-frontend/src/components/Wallet.tsx
```diff
@@ -228,17 +279,119 @@ export default function Wallet(
...
-                            onClick={exportSeed}
-                        >
-                            Backup Seed
-                        </Button>
                     </HStack>
                 </VStack>
 
-                <Divider marginTop={2} marginBottom={2} />
+                {environment !== "umbrel" && (
+                    <Box>
```
bonomat, line 284: Would be nice to get some screenshots in the PR description, my braincompiler is not showing me that nicely :)

bonomat (commented):
Had a quick look and added some comments. Will need to have a nother look though :)

crates/daemon/src/seed.rs
holzeis: it's not.. it is used at couple of places.

- https://github.com/itchysats/itchysats/pull/3173/files#diff-12b21198b101f645ffc5a64ed4a61f034cd0b0bd04075a83f05a83e3ec680193R355
- https://github.com/itchysats/itchysats/pull/3173/files#diff-cad97fd20b55ed7bc39ba6ad93b50ab25691651c1cb5141ba03470d35fe59c71R478
- https://github.com/itchysats/itchysats/pull/3173/files#diff-12b21198b101f645ffc5a64ed4a61f034cd0b0bd04075a83f05a83e3ec680193R361

taker-frontend/src/components/Wallet.tsx
holzeis: I've added a gif to the PR description.

crates/daemon/src/wallet.rs
holzeis: well.. we could do that, but I have the feeling we would very quickly end up with testing only test code. however, I can give it a try.

holzeis: @bonomat have a look at 01f62ef334e28a966bc975e724bba05c1beca2a8. It uses the test wallet now, and does not sync with electrum for balance, but I guess checking on the restored address is sufficient.

crates/maker/src/main.rs
```diff
@@ -104,7 +105,16 @@ async fn main() -> Result<()> {
         return Ok(());
     }
 
-    let identities = seed.derive_identities();
+    let identity_seed_file = &data_dir.join(seed::MAKER_IDENTITY_SEED_FILE);
+    if !identity_seed_file.exists() {
+        tracing::info!("Copying wallet seed file for identity seed file.");
```
luckysori: 🙃 Our convention seems to be to not add periods at the end of logs.

crates/bdk-ext/src/lib.rs
```diff
@@ -6,16 +6,18 @@ use bdk::bitcoin::secp256k1::SecretKey;
...
 
 pub mod keypair;
 
-pub fn new_test_wallet(
+pub fn new_test_wallet<DB: BatchDatabase>(
     rng: &mut (impl RngCore + CryptoRng),
     utxo_amount: Amount,
     num_utxos: u8,
-) -> Result<bdk::Wallet<bdk::database::MemoryDatabase>> {
+    mut database: DB,
```
luckysori, line 33: 🙃 I think adding a new method like `new_test_wallet_from_database` would be a bit cleaner. The new argument is effectively irrelevant in most scenarios, because we usually just want an empty `MemoryDatabase`.

crates/daemon-tests/tests/main/import_seed.rs
```diff
@@ -0,0 +1,40 @@
...
+use std::path::Path;
+
+#[otel_test]
+async fn fail_to_import_seed_with_open_cfds() {
+    let (mut maker, mut taker) = start_both().await;
+
+    let cfd_args = OpenCfdArgs::default();
+    open_cfd(&mut taker, &mut maker, cfd_args.clone()).await;
+
+    let random_folder: String = thread_rng()
```
luckysori: 😛 

```suggestion
    let random_dir: String = thread_rng()
```

crates/daemon/src/lib.rs
```diff
@@ -445,14 +453,46 @@ where
...
+        network: Network,
+    ) -> Result<()> {
+        let open_cfd_ids = self.db.load_open_cfd_ids().await?;
+
+        if !open_cfd_ids.is_empty() {
+            tracing::warn!("Rejecting imported seed as open CFDs exist.");
+            // in case of open cfds we do not accept an imported seed.
+            return Err(anyhow!(
+                "Can not accept imported seed since open CFDs exist."
+            ));
```
luckysori: 🔧 Logging and bubbling up at the same time seems unnecessary. I would just bubble up the error and let the caller decide what to do (probably log it).

The idiomatic way to do this with `anyhow` would be:

```suggestion
            use anyhow::bail;
            
            bail!("Cannot accept imported seed since open CFDs exist");
```

I would also omit the comment since it's saying basically the same as the message and the log.

crates/daemon/src/lib.rs
```diff
@@ -445,14 +453,46 @@ where
...
+        // recreate the wallet within the wallet actor
+        self.wallet_actor
+            .send(wallet::ImportSeed {
+                seed,
+                seed_path: data_dir.join(seed::TAKER_WALLET_SEED_FILE),
+                network,
+            })
+            .await??;
+
+        // sync imported wallet
```
luckysori: 🙃 I think the line below is self-explanatory.

crates/daemon/src/seed.rs
```diff
@@ -30,6 +30,12 @@ impl Identities {
...
+
+pub const MAKER_WALLET_SEED_FILE: &str = "maker_seed";
+pub const MAKER_IDENTITY_SEED_FILE: &str = "maker_id_seed";
```
luckysori: 🙃 We usually put constants at the top of modules, just below `use` statements and `mod` declarations.

crates/daemon/src/lib.rs
```diff
@@ -445,14 +453,46 @@ where
...
+    }
 }
 
 /// A struct defining our environment
 ///
 /// We can run on all kinds of environment, hence this is just a wrapper around string.
 /// However, for backwards compatibility with <=0.6.x we need to support to support
 /// `Unknown`. For all other we format the string to lowercase.
-#[derive(Debug, Clone, Display, PartialEq, Eq)]
+#[derive(Debug, Clone, Display, PartialEq, Eq, Serialize)]
```
luckysori: ❓ Why do we need to add this?

crates/daemon/src/wallet.rs
```diff
@@ -158,10 +154,61 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
...
+impl<B> Actor<B, Tree>
+where
+    Self: xtra::Actor,
+{
+    pub fn import_seed(&mut self, msg: ImportSeed) -> Result<AddressInfo> {
+        let seed = msg.seed;
+        let import_seed: [u8; 256] = seed
+            .clone()
+            .try_into()
+            .map_err(|_| anyhow!("seed must be 256 bytes long"))?;
```
luckysori: 🔧 This is more idiomatic.

```suggestion
            // use anyhow::Context;
            .context("seed must be 256 bytes long")?;
```

crates/daemon/src/wallet.rs
```diff
@@ -158,10 +154,61 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
...
+
+        self.wallet
+            .get_address(AddressIndex::LastUnused)
+            .map_err(|_| anyhow!("Could not get address"))
+    }
 }
 
 impl<DB> Actor<ElectrumBlockchain, DB>
 where
+    Self: xtra::Actor,
```
luckysori: 🔧 Not needed!

crates/daemon/src/wallet.rs
```diff
@@ -398,7 +451,7 @@ trait BuildLockTx {
     ) -> Result<PartiallySignedTransaction>;
 }
 
-impl<D> BuildLockTx for bdk::Wallet<D>
+impl<D> BuildLockTx for Wallet<D>
```
luckysori: 🙃 Not your change, but to be consistent with other type parameters.

```suggestion
impl<DB> BuildLockTx for Wallet<DB>
```

crates/daemon/src/wallet.rs
```diff
@@ -158,10 +154,61 @@ impl Actor<ElectrumBlockchain, sled::Tree> {
...
+        let ext_priv_key = import_seed.derive_extended_priv_key(msg.network)?;
+
+        // reuse existing database as the file has already been opened.
+        let db = self.db.clone().expect("database should be existing.");
+
+        // recreate and update wallet
+        self.wallet = Actor::build_wallet(ext_priv_key, db)?;
+
+        // write imported seed file to disk.
+        tokio::fs::write(msg.seed_path.as_path(), seed).await?;
```
luckysori: 🤔 Aren't we just overwriting the original seed file here? That seems very dangerous to me.

crates/daemon/src/wallet.rs
```diff
@@ -487,20 +540,46 @@ fn seed_and_rpc_on_same_network(rpc: &electrum_client::Client, network: Network)
...
+        244, 99, 105, 36, 227, 118, 221, 245, 104, 33, 3, 91, 176, 176, 157, 42, 201, 92, 99, 104,
+        235, 84, 237, 229, 131, 204, 227, 15, 223, 109, 8, 47, 155, 132, 236, 108, 106, 59, 185,
+        70, 20, 247, 133, 192, 164, 93, 65, 141, 16, 18, 105, 251, 23, 168, 95, 145, 252, 101, 27,
+        218, 97, 26, 20, 71, 102, 221, 182, 146, 148, 11, 117, 159, 67, 132, 223, 245, 141, 174,
+        157, 226, 67, 5, 61, 218, 98, 131, 8, 170, 233, 140, 139, 198, 24, 107, 60, 147, 236, 194,
+        234, 227, 58, 54, 61, 245, 94, 209, 12, 28, 131, 156, 37, 169, 32, 38, 38, 212, 2, 26, 204,
+        231, 154, 194, 38, 142, 30, 71, 33, 100, 164, 169, 168, 34, 6, 110, 126, 22, 120, 191, 168,
+        219, 64, 70, 172, 224, 218, 142, 135, 179, 79, 140, 174, 157, 60, 43, 141, 218, 178, 45,
+        30, 184, 79, 100, 43, 97, 228, 35, 122, 60, 143, 239, 109, 104, 56, 64, 3, 131,
+    ]; // holds 0.001 btc on testnet
```
luckysori: 🤔 Surely this isn't relevant? We should not rely on a testnet address having coins in it for automated testing.

crates/daemon/src/wallet.rs
```diff
@@ -487,20 +540,46 @@ fn seed_and_rpc_on_same_network(rpc: &electrum_client::Client, network: Network)
...
     use rand::thread_rng;
+    use rand::Rng;
     use std::collections::HashSet;
+    use std::env;
+    use std::path::Path;
     use tokio_extras::Tasks;
 
-    impl Actor<(), bdk::database::MemoryDatabase> {
-        pub fn new_offline(
+    const IMPORT_SEED: [u8; 256] = [
```
luckysori: 🙃 Since it's only used in one test I would just inline it.

crates/daemon/src/wallet.rs
```diff
@@ -512,6 +591,7 @@ mod tests {
                     time_to_lock,
                 },
                 blockchain_client: (),
+                db: None,
```
luckysori, line 595: ❓ Why is this field an `Option`? I would expect to always be able to provide the database.

crates/daemon/src/wallet.rs
```diff
@@ -526,9 +606,112 @@ mod tests {
...
+
+        let mut tasks = Tasks::default();
+
+        let database = sled::open(data_dir.clone()).expect("could not open database");
+        let database = database
+            .open_tree("wallet name")
+            .expect("could not open tree");
+
+        // create wallet with only one UTXO which will be locked for a
+        // long time after being used
```
luckysori, line 638: 🔧 This comment seems to be irrelevant for this test.

luckysori: Probably better if you set up the test with an empty wallet, since it's not relevant for what you are testing.

crates/daemon/src/wallet.rs
luckysori: ❓ Why can't it be a random seed with the right length?

crates/taker/src/routes.rs
```diff
@@ -94,6 +98,8 @@ pub async fn feed(
 
         yield Event::json(&identity).event("identity");
 
+        yield Event::json(&environment).event("environment");
```
luckysori: 🤔 This seems weird. Is there no alternative to this?

It's also questionable to rely on the environment variable to determine whether importing a seed is possible or not. Perhaps a better approach would be to selectively mount the route depending on the platform we are targeting? This would mean we would have a slightly different binary for `Umbrel`.

crates/taker/src/routes.rs
```diff
@@ -305,20 +311,78 @@ pub async fn put_sync_wallet(taker: &State<Taker>, _user: User) -> Result<(), Ht
...
+) -> Result<(), HttpApiProblem> {
+    if environment.inner().as_string() == "umbrel" {
+        return Err(HttpApiProblem::new(StatusCode::BAD_REQUEST)
+            .title("Could not import seed file")
+            .detail("Cannot import seed file for umbrel environment"));
+    }
+
+    // fetch seed from upload stream as bytes. only fetch max 256 bytes as the seed file shouldn't
+    // be bigger than that.
+    let seed = seed.open(256.bytes()).into_bytes().await.map_err(|e| {
```
luckysori: 🔧 It would be nice if the `256` number came from a constant in the appropriate module. That way you can get rid of the comment too by adding documentation onto the constant.

luckysori (commented):
Nice work!

I am mostly unsure about:

1. Importing leading to overwriting the current wallet seed.
2. The way we determine the environment.

crates/daemon/src/lib.rs
luckysori: I understand after completing my review. It would be good to remove this if we end up not using it if we modify the current solution.

crates/daemon/src/wallet.rs
holzeis: @luckysori I am asserting on the address. To ensure that from the same import seed we will always derive the same address. Before, it was also used to check the balance (it owns 0.001 btc)

crates/daemon/src/wallet.rs
holzeis: @luckysori well, I don't like it either, but I didn't find a better way of doing it with the existing tests. Since the tests mainly use a MemoryDatabase, I do not have a Db available there. Do you have a suggestion?

crates/daemon/src/wallet.rs
holzeis: did not add that intentional.

crates/taker/src/routes.rs
holzeis: Having to build different binaries just for that seems to be an overkill. I think one alternative to the environment variable, could be to use the app-seed flag. If set it means ItchySats is not managing the seed and hence will not support exporting or importing. It's implicitly what the umbrel environment is saying, but I guess that way its clearer.

I will adapt.

holzeis:
@luckysori 

> I am mostly unsure about: 

>     1. Importing leading to overwriting the current wallet seed.

I've committed a change https://github.com/itchysats/itchysats/pull/3173/commits/84a510c745d1ec9d72bc49791456de8ba582f936 which will automatically copy the existing seed file into a backup file with the timestamp in its name. e.g. `taker_seed.1666082503.back`

>     2. The way we determine the environment.

Alternatively to the environment variable, I could make use of the app seed argument, meaning if the seed is provided from extern we are not managing it in terms of importing or exporting. This will then be reflect with a simple boolean.

e.g. something like that.
```
let managed_seed = match opts.app_seed {
    Some(_) => false,
    None => true,
};
```

I will wait for your feedback before implementing that path though, as I am not sure if this is a much better approach than the current one.

crates/daemon/src/wallet.rs
luckysori: Do we need to assert on the address? Maybe it's enough to just check that it changes as that implies that a new wallet has been loaded.

crates/daemon/src/wallet.rs
luckysori: Hmm, I didn't come up with anything after some limited experimenting. I might come back to this later myself.

luckysori:
> Alternatively to the environment variable, I could make use of the app seed argument, meaning if the seed is provided from extern we are not managing it in terms of importing or exporting. This will then be reflect with a simple boolean.
> 
> e.g. something like that.
> 
> ```
> let managed_seed = match opts.app_seed {
>     Some(_) => false,
>     None => true,
> };
> ```
> 
> I will wait for your feedback before implementing that path though, as I am not sure if this is a much better approach than the current one.

I think it is better to rely on the presence of the app seed argument, because that is the reason why importing (and perhaps even exporting?) a seed shouldn't be allowed. On the other hand, the `environment` variable was introduced for metrics.

What I also care about is _disabling_ the route based on this information, rather than exposing an event via SSE that the client has to interpret. It removes an event that doesn't really belong there imo and it also means that the API is unavailable even if you don't go through the UI.

crates/daemon/src/wallet.rs
holzeis: what benefits would we get by that - i feel like the test would be less effective without that assertion.

holzeis:
> What I also care about is _disabling_ the route based on this information, rather than exposing an event via SSE that the client has to interpret. It removes an event that doesn't really belong there imo and it also means that the API is unavailable even if you don't go through the UI.

That's actually a very good point. I will adapt that - however i still need to transfer the information about the disabled import / export feature. I will implement it using the app seed argument. Thanks for the feedback!

holzeis:
@luckysori have a look at https://github.com/itchysats/itchysats/pull/3173/commits/a98bdb51d9549261ae90ed96668b9b4625f15694.

I've introduced a managed_wallet flag and added it to the WalletInfo event. Additionally the export / import routes will only be added in case the wallet is managed. I like it much better now, wdyt?

crates/daemon/src/wallet.rs
luckysori: I suppose it is valuable to check that importing a wallet is deterministic.

I would then suggest to move the expected seed into the specific test (I believe it is only used once). It's a bit noisy otherwise.

crates/daemon/src/lib.rs
holzeis: removed it in favor of https://github.com/itchysats/itchysats/pull/3173/commits/a98bdb51d9549261ae90ed96668b9b4625f15694

crates/daemon/src/wallet.rs
holzeis: @bonomat what do you have in mind exactly?

luckysori:
> I've introduced a managed_wallet flag and added it to the WalletInfo event. Additionally the export / import routes will only be added in case the wallet is managed. I like it much better now, wdyt?

Nice work! I like it much better too. It makes sense for it to go with the `WalletInfo` event!

luckysori (approved):
Thanks for addressing my comments!

holzeis:
> Thanks for addressing my comments!

Thanks for taking the time and give me comments 🙂 👍

PR #3182, opened 2022-10-13
A bunch of small optimisations

klochowicz:
See commit details.

### DIFF
--- a/crates/model/src/lib.rs
+++ b/crates/model/src/lib.rs
@@ -4,7 +4,6 @@ use anyhow::Context;
 use anyhow::Result;
 use bdk::bitcoin::Address;
 use bdk::bitcoin::Amount;
-use bdk::bitcoin::Denomination;
 use bdk::bitcoin::Network;
 use bdk::bitcoin::SignedAmount;
 use bdk::bitcoin::Txid;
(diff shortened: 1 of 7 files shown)


crates/model/src/lib.rs
```diff
@@ -240,8 +239,7 @@ impl Div<Price> for Contracts {
     fn div(self, rhs: Price) -> Self::Output {
         let mut btc = self.0 / rhs.0;
         btc.rescale(8);
-        Amount::from_str_in(&btc.to_string(), Denomination::Bitcoin)
-            .expect("Error computing BTC amount")
+        Amount::from_btc(btc.to_f64().expect("to fit in f64")).expect("Parse the BTC amount")
```
bonomat, line 242: This does the same thing underthehood.

bonomat (approved):

crates/model/src/lib.rs
klochowicz: not quite - we avoid using `Decimal`s `Display` implementation, which is AFAICT not as straightforward as f64 (regular float).

klochowicz: but good point that `Amount::from_btc` creates a `String`. I have overlooked that.

PR #3190, opened 2022-10-17
Fix the memory growth

da-kami:
This is an attempt to fix the memory growth caused by `rust-electrum-client`'s `batch_script_get_history` implementation. We chunk up the scripts to be checked into batches and call `script_get_history` for the batches in parallel in separate worker threads.

The results are communicated back using a channel, and are processed as they come; we don't have to rely on any order here.

---

The implementation should be complete and compiles, but there are still a few rough edges ;)
@klochowicz please run this with profiling so we can see if it has the expected behavior.

TODO:

- [x] Fix tests: The tests have to be adapted to cover the new behavior. 
- [x] Evaluate if the buffer size for the channel should be in relation to the batch size

### DIFF
--- a/crates/daemon/src/monitor.rs
+++ b/crates/daemon/src/monitor.rs
@@ -30,6 +30,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use tracing::Instrument;
 use xtra_productivity::xtra_productivity;
+use xtras::SendAsyncNext;
 use xtras::SendInterval;
 
 const LOCK_FINALITY_CONFIRMATIONS: u32 = 1;
@@ -453,7 +454,9 @@ impl Actor {
     }
 
     #[tracing::instrument("Sync monitor", skip_all, err)]
-    async fn sync(&mut self) -> Result<()> {
+    async fn start_sync(
+        &mut self,
+    ) -> Result<(BlockHeight, tokio::sync::mpsc::Receiver<Vec<GetHistoryRes>>)> {
         // Fetch the latest block for storing the height.
         // We do not act on this subscription after this call, as we cannot rely on
         // subscription push notifications because eventually the Electrum server will
@@ -470,12 +473,7 @@ impl Actor {
 
         tracing::trace!("Updating status of {num_transactions} transactions",);
 
-        // TODO: Evaluate buffer; if we do batches of 25 and allow buffer of 100 here, does that
-        // mean we can at most process 4 batches at the time?  What happens if the buffer is
-        // exhausted? (I would expect the next messages will be queued) We use tokio::sync::
-        // mpsc::channel because we need the receiver to be in an async context because we do async
-        // work upon receiving script updates
-        let (sender, mut receiver) = tokio::sync::mpsc::channel(100);
+        let (tx_script_updates, rx_script_updates) = tokio::sync::mpsc::channel(BATCH_SIZE * 4);
 
         let scripts = self
             .state
@@ -487,21 +485,22 @@ impl Actor {
 
         let client = self.client.clone();
 
-        let batch_futures = batches.into_iter().map(move |batch| {
+        for batch in batches {
             tokio::task::spawn_blocking({
                 let client = client.clone();
-                let sender = sender.clone();
+                let tx_script_updates = tx_script_updates.clone();
                 move || {
                     for script in batch {
-                        // TODO: Evaluate if we should notify the receivers about errors, I feel
-                        //  logging is good enough here because the receiver won't know how to deal
-                        //  with an error.
                         match client.script_get_history(&script) {
-                            Ok(resp) => {
+                            Ok(script_history_response) => {
                                 // We use blocking_send to stay within a sync context here
                                 // One should not use async code in a spawn_blocking block
-                                if let Err(e) = sender.blocking_send(resp) {
-                                    tracing::error!("Error during script history batching: {e:#}")
+                                if let Err(e) =
+                                    tx_script_updates.blocking_send(script_history_response)
+                                {
+                                    tracing::error!(
+                                        "Error when processing script_get_history response: {e:#}"
+                                    )
                                 }
                             }
                             Err(e) => {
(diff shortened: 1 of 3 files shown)


crates/daemon/src/monitor.rs
```diff
@@ -466,25 +470,76 @@ impl Actor {
...
+
+        let client = self.client.clone();
+
+        let batch_futures = batches.into_iter().map(move |batch| {
+            tokio::task::spawn_blocking({
+                let client = client.clone();
+                let sender = sender.clone();
+                move || {
+                    for script in batch {
+                        // TODO: Evaluate if we should notify the receivers about errors, I feel
```
bonomat: Agree, as long as we don't intend to deal with the errors there is no need in bubbling it up

crates/daemon/src/monitor.rs
```diff
@@ -466,25 +470,76 @@ impl Actor {
...
-                            height: response.height,
-                            tx_hash: response.tx_hash,
-                        })
-                        .collect()
-                })
-                .collect(),
-        );
+        // TODO: Evaluate buffer; if we do batches of 25 and allow buffer of 100 here, does that
+        // mean we can at most process 4 batches at the time?  What happens if the buffer is
+        // exhausted? (I would expect the next messages will be queued) We use tokio::sync::
```
bonomat: From the docs: 

> The channel will buffer up to the provided number of messages. Once the buffer is full, attempts to send new messages will wait until a message is received from the channel. The provided buffer capacity must be at least 1.

so you are right with your assumption.

crates/daemon/src/monitor.rs
```diff
@@ -466,25 +470,76 @@ impl Actor {
...
+            .state
+            .monitoring_scripts()
+            .cloned()
+            .collect::<Vec<Script>>();
+
+        let batches = scripts.chunks(BATCH_SIZE).map(|batch| batch.to_owned());
+
+        let client = self.client.clone();
+
+        let batch_futures = batches.into_iter().map(move |batch| {
```
bonomat: 🔧
I think you will need to wait between the batches, otherwise you just fire one after the other because you immediately spawn them away.

bonomat (commented):
We should also change how often we call `Sync`.

crates/daemon/src/monitor.rs
klochowicz: It would be interesting to see whether we actually need to wait here; we're not concerned about the results coming out of order, as long as the results will come back (depends on the behaviour of Electrum?)

bonomat: If we don't wait here, we don't have to use batches because they get all spawned away. Or am I misreading this?

crates/daemon/src/monitor.rs
klochowicz: there are some errors that we might be interested in, e.g. if we're exceeding the rate, we could slow down the requests.

crates/daemon/src/monitor.rs
klochowicz: well, the difference is still that we're partitioning them in different worker threads (the `spawn_blocking`)

bonomat: So we will spawn batches of 25 in parallel and execute all (4x) in parallel and wait until at least 1 is finished before we spawn another 25?

klochowicz: pretty much, but the 4x figure depends on the limit of the worker threads. AFAICT, this limit is by default 512, so we pretty much fire all the batches at the same time, and then reach the electrum API one at a time from all these threads.
tokio has separate threads, separate from the runtime that are used here.

bonomat: gotcha, thanks for the explanation. :)

da-kami: Maybe I misunderstand what you are saying.

The idea of this implementation is to chunk up the requests in packages of 25 and then process all these packages (I called them batches) in parallel.
Maybe the term `batch` is not used correctly here, because we are not processing what is in the batch in parallel, but we process all the chunks in parallel. 
I am happy to rename what is called `batch` at the moment. 
Do you think the implementation needs changes?

---

Note that in fixup https://github.com/itchysats/itchysats/pull/3190/commits/c009237fd04b7258b066e4d867323d77fd5d6785 we stop joining all the `spawn_blocking` join handles, because that is not necessary and just adds complexity.

crates/daemon/src/monitor.rs
da-kami: > there are some errors that we might be interested in, e.g. if we're exceeding the rate, we could slow down the requests.

Is it fair to say that we can add such a feature later, once we know how this performs on testnet? 
I find it hard to design this without knowing how the solution will perform under load.

We might also want to change the `BATCH_SIZE` - 25 is a somewhat arbitrary number. At the moment we process batches of 25 requests in parallel, for `7000` subscriptions this means `280` worker threads in parallel. When running the test that simulates this behavior this did not cause problems on our local machines, but there was CPU load. We will have to see how this performs on the testnet machine.

crates/daemon/src/monitor.rs
```diff
@@ -466,66 +485,46 @@ impl Actor {
...
+                    for script in batch {
+                        match client.script_get_history(&script) {
+                            Ok(script_history_response) => {
+                                // We use blocking_send to stay within a sync context here
+                                // One should not use async code in a spawn_blocking block
+                                if let Err(e) =
+                                    tx_script_updates.blocking_send(script_history_response)
+                                {
+                                    tracing::error!(
+                                        "Error when processing script_get_history response: {e:#}"
```
klochowicz, line 514: 🛠 this one gets fired off all the time when I tested it with response "channel closed", and we never stop.

I don't think just logging the errors here is an option....

crates/sqlite-db/src/lib.rs
```diff
@@ -706,6 +706,8 @@ async fn load_cfd_events(
             if let Some((dlc, funding_fee, complete_fee)) =
                 rollover::load(&mut *conn, *cfd_row_id, *event_row_id).await?
             {
+                let dlc_id = dlc.commit.0.txid();
+                tracing::warn!("Attaching DLC {dlc_id} from event {event_row_id} to CFD {id}");
```
bonomat, line 710: Why is this a `warn`?

da-kami:
Closing this in favor of: https://github.com/itchysats/itchysats/pull/3197

crates/sqlite-db/src/lib.rs
da-kami: This is a weird rebasing mistake, not sure how that sneaked in but it is removed in the PRs that replace this one :)

crates/daemon/src/monitor.rs
da-kami: I was unable to directly reproduce the "channel closed" issue, but I found a problem in the implementation where the sender and the receiver were not properly tied together, because the sender was not moved correctly. 
I think that this issue caused this behavior. In the PRs replacing this one I also added tests to ensure that the underlying batch functionality works, but we yet have to test the complete production code again. 

I also thought about adding a test that spins up the complete monitor Actor, but that is a bit too much for now...

PR #3196, opened 2022-10-18
Remember rollover state if maker fails to send revocation sk 

### LINKED ISSUE #3019: Keep taker's failed rollover data if maker could have succeeded
I'm [rewriting the punishment transaction issue](https://github.com/itchysats/itchysats/issues/907) and I've just had a realisation.

We should never just let the taker fail a rollover at the last step and forget about everything that has already been negotiated. We need to remember all the information, because it is the _only_ way that we can safely non-collaboratively close the CFD! Otherwise we can be held hostage by the maker:

- They can publish the new commit transaction, but they don't have to (to their own detriment, but still).
- We must not publish the old commit transaction, because they can punish us if we do so.

In order to keep the CFD alive, I think we are already doing the correct thing: allowing takers to catch up if they fall behind.
But we need to be able to model this in-between state more explicitly if we want to be able to close the CFD without the maker's permission. That means we cannot just pretend that nothing happened and not update the DLC.

Looking at the code, the `Dlc` struct must be updated as usual, except for the new entry to

https://github.com/itchysats/itchysats/blob/4961726fd5106d8c77dfa0047460be26e66634ed/model/src/cfd.rs#L1990

which should reflect the fact that there's missing information (the `revocation_sk_theirs`).

The new design should allow us to:

- Catch up by using the _second to last_ commit TXID in the list of revoked commit transactions.
- Unilaterally close the CFD by publishing the _current_ commit transaction.

The only unfortunate and unavoidable power that the maker has over a taker that is lagging behind is the ability to commit to the _second to last_ commit transaction without punishment. But this was already the case.

luckysori:
Fixes https://github.com/itchysats/itchysats/issues/3019.

This ensures that the taker can _always_ safely non-collaboratively settle the CFD.

Before these changes, the taker could fall behind in the state of the DLC if the maker failed to send their revocation sk in the last step of the rollover protocol. The taker would throw away everything exchanged during that instance of the rollover protocol. This would mean that the maker would know how to punish the publication of the commit transaction the taker considered current.

With these changes, the taker remembers the incomplete rollover details, updating the state of the DLC so that they can safely publish their latest commit transaction if they want to. The taker also notes that the maker did not share their revocation sk. This allows the taker to still insist on rolling over from the previous, yet-to-be-revoked-by-the-maker commit transaction, like they have been doing since the introduction of the rollover retry mechanism with patch https://github.com/itchysats/itchysats/commit/e0e643e44d9fa5f03422f4430ffc860d86ef2de6.

TODOs:

- [ ] Add a test.
- [ ] Mention it in the changelog.

### DIFF
--- a/crates/sqlite-db/src/rollover/load.rs
+++ b/crates/sqlite-db/src/rollover/load.rs
@@ -145,7 +145,9 @@ async fn load_revoked_commit_transactions(
             revocation_sk_ours: row
                 .revocation_sk_ours
                 .map(|revocation_sk_ours| revocation_sk_ours.into()),
-            revocation_sk_theirs: row.revocation_sk_theirs.into(),
+            revocation_sk_theirs: row
+                .revocation_sk_theirs
+                .map(|revocation_sk_theirs| revocation_sk_theirs.into()),
             publication_pk_theirs: row.publication_pk_theirs.into(),
             script_pubkey: Script::from_hex(row.script_pubkey.as_str())?,
             txid: row.txid.into(),
--- a/crates/sqlite-db/src/rollover/overwrite.rs
+++ b/crates/sqlite-db/src/rollover/overwrite.rs
@@ -169,7 +169,7 @@ async fn insert_revoked_commit_transaction(
     revoked: RevokedCommit,
 ) -> Result<()> {
     let revoked_tx_script_pubkey = revoked.script_pubkey.to_hex();
-    let revocation_sk_theirs = models::SecretKey::from(revoked.revocation_sk_theirs);
+    let revocation_sk_theirs = revoked.revocation_sk_theirs.map(models::SecretKey::from);
     let revocation_sk_ours = revoked.revocation_sk_ours.map(models::SecretKey::from);
     let publication_pk_theirs = models::PublicKey::from(revoked.publication_pk_theirs);
     let encsig_ours = models::AdaptorSignature::from(revoked.encsig_ours);
--- a/crates/daemon-tests/src/lib.rs
+++ b/crates/daemon-tests/src/lib.rs
@@ -14,6 +14,7 @@ use daemon::bdk::bitcoin::Network;
 use daemon::bdk::bitcoin::SignedAmount;
 use daemon::bdk::bitcoin::Txid;
 use daemon::libp2p_utils::create_connect_multiaddr;
+use daemon::maia_core::secp256k1_zkp::SecretKey;
 use daemon::maia_core::secp256k1_zkp::XOnlyPublicKey;
 use daemon::online_status::ConnectionStatus;
 use daemon::oracle::Attestation;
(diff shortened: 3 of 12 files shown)


crates/sqlite-db/src/rollover/load.rs
```diff
@@ -145,7 +145,9 @@ async fn load_revoked_commit_transactions(
             revocation_sk_ours: row
                 .revocation_sk_ours
                 .map(|revocation_sk_ours| revocation_sk_ours.into()),
-            revocation_sk_theirs: row.revocation_sk_theirs.into(),
+            revocation_sk_theirs: row
+                .revocation_sk_theirs
+                .map(|revocation_sk_theirs| revocation_sk_theirs.into()),
```
holzeis, line 150: ❓ What does that do?

crates/sqlite-db/src/rollover/overwrite.rs
```diff
@@ -169,7 +169,7 @@ async fn insert_revoked_commit_transaction(
     revoked: RevokedCommit,
 ) -> Result<()> {
     let revoked_tx_script_pubkey = revoked.script_pubkey.to_hex();
-    let revocation_sk_theirs = models::SecretKey::from(revoked.revocation_sk_theirs);
+    let revocation_sk_theirs = revoked.revocation_sk_theirs.map(models::SecretKey::from);
```
holzeis, line 172: ❓ Is there logically a difference or is this just syntax sugar?

holzeis (approved):
👍 LGTM

Do I read it correctly that this fix will ensure that the DLC gets updated even if the maker or taker fails to send the revocation sk?

crates/sqlite-db/src/rollover/load.rs
luckysori: Before, the `revocation_sk_theirs` was declared as `NOT NULL`, so we would always get a `sqlite_db::models::SecretKey` when loading from the database. We would then convert it into a `secp256k1::SecretKey` with the `.into()`.

With this PR, the `revocation_sk_theirs` is declared as `NULL`. This means that we load an `Option<sqlite_db::models::SecretKey>` from the database. We now need to convert it into an `Option<secp256k1::SecretKey`. To do so we map the contents of the `Option` with the appropriate `Into` (auto-generated by implementing `From`) implementation.

You can even write:

```rust
.map(secp256k1::SecretKey::from)
```

crates/sqlite-db/src/rollover/overwrite.rs
luckysori: Same answer as above, but for writing to the database. So we are mapping from `secp256k1::SecretKey` to `sqlite_db::models::SecretKey`.

luckysori:
@holzeis:

> Do I read it correctly that this fix will ensure that the DLC gets updated even if the maker or taker fails to send the revocation sk?

Almost. The DLC will get updated if:

- rollover succeeds; or
- rollover fails _for the taker_ whilst waiting for the maker's revocation key i.e. at the last step.

These changes don't affect the maker.

The reason why it is important to update the DLC if the maker fails to send their revocation key is because the taker has already revoked the previous commit transaction! If they don't remember the new one, they cannot publish any commit transaction safely! So they can be held hostage by the maker.

luckysori:
Not working on this atm. We can reopen it if we wanna focus on this work in the future.

PR #3197, opened 2022-10-18
Self-implemented batching for script_get_history

da-kami:
This is an attempt to fix the memory growth caused by `rust-electrum-client`'s `batch_script_get_history` implementation.
We chunk up the scripts to be checked into batches and call `script_get_history` for the batches in parallel in separate worker threads.
The results are communicated back using a channel and accumulated before processing.

Includes tests that simulate the behavior. The tests rely on a mainnet Electrum instance and are ignored on CI.

### DIFF
--- a/crates/daemon/src/monitor.rs
+++ b/crates/daemon/src/monitor.rs
@@ -11,6 +11,7 @@ use bdk::bitcoin::Txid;
 use bdk::descriptor::Descriptor;
 use bdk::electrum_client;
 use bdk::electrum_client::ElectrumApi;
+use bdk::electrum_client::GetHistoryRes;
 use bdk::miniscript::DescriptorTrait;
 use btsieve::ScriptStatus;
 use btsieve::State;
@@ -24,6 +25,7 @@ use model::CET_TIMELOCK;
 use serde_json::Value;
 use sqlite_db;
 use std::collections::HashMap;
+use std::sync::Arc;
 use std::time::Duration;
 use tracing::Instrument;
 use xtra_productivity::xtra_productivity;
@@ -34,6 +36,7 @@ const CLOSE_FINALITY_CONFIRMATIONS: u32 = 3;
 const COMMIT_FINALITY_CONFIRMATIONS: u32 = 1;
 const CET_FINALITY_CONFIRMATIONS: u32 = 3;
 const REFUND_FINALITY_CONFIRMATIONS: u32 = 3;
+const BATCH_SIZE: usize = 25;
 
 /// Electrum client timeout in seconds
 ///
@@ -114,7 +117,7 @@ pub struct Sync;
 //  -> Might as well just send out all events independent of sending to the cfd actor.
 pub struct Actor {
     executor: command::Executor,
-    client: bdk::electrum_client::Client,
+    client: Arc<bdk::electrum_client::Client>,
     state: State<Event>,
     db: sqlite_db::Connection,
 }
@@ -352,7 +355,7 @@ impl Actor {
             .into();
 
         Ok(Self {
-            client,
+            client: Arc::new(client),
             executor,
             state: State::new(latest_block),
             db,
@@ -478,10 +481,13 @@ impl Actor {
 
         tracing::trace!("Updating status of {num_transactions} transactions",);
 
-        let histories = self
-            .client
-            .batch_script_get_history(self.state.monitoring_scripts())
-            .context("Failed to get script histories")?;
+        let scripts = self
+            .state
+            .monitoring_scripts()
+            .cloned()
+            .collect::<Vec<Script>>();
+
+        let histories = batch_script_get_history(self.client.clone(), scripts).await;
 
         let mut ready_events = self.state.update(
             latest_block_height,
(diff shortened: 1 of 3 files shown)


da-kami:
This PR was created to keep the changeset smaller and at the same time be able to write some tests.
I ported the tests that we wrote for the `RawClient` of `rust-electrum-client` over to our project and adapted them. 

@klochowicz I think this is what you meant earlier. If this runs into the same problem when running it we will have a smaller changeset to work on.

crates/daemon/src/monitor.rs
```diff
@@ -478,10 +481,14 @@ impl Actor {
...
-            .batch_script_get_history(self.state.monitoring_scripts())
-            .context("Failed to get script histories")?;
+        let scripts = self.state.monitoring_scripts().cloned().collect();
+
+        let mut rx_script_history = batch_script_get_history(self.client.clone(), scripts);
+
+        let mut histories = Vec::new();
+        while let Some(script_history) = rx_script_history.recv().await {
+            histories.push(script_history)
+        }
```
klochowicz: yeah, this is pretty much what I was doing last night!

some things for consideration:
- I would have put that inside the `batch_script_get_history` to have it encapsulated (otherwise, we're pointlessly leaking the internal implementation with the receiver (we can just expose the stream if we don't want to collect to the vector)

crates/daemon/src/monitor.rs
```diff
@@ -478,10 +481,14 @@ impl Actor {
...
 
-        let histories = self
-            .client
-            .batch_script_get_history(self.state.monitoring_scripts())
-            .context("Failed to get script histories")?;
+        let scripts = self.state.monitoring_scripts().cloned().collect();
+
+        let mut rx_script_history = batch_script_get_history(self.client.clone(), scripts);
+
+        let mut histories = Vec::new();
```
klochowicz: we can use `with_capacity(scripts.len())` here to avoid unnecessary reallocations of this vector as it grows - we know exactly the max size of the results we're gathering.

crates/daemon/src/monitor.rs
```diff
@@ -478,10 +481,14 @@ impl Actor {
...
+        while let Some(script_history) = rx_script_history.recv().await {
+            histories.push(script_history)
+        }
```
klochowicz, line 491: you could also add a poor man's check whether all the scripts got a result by comparing `anyhow::ensure!(scripts.len() == histories.len()` (or in a less dramatic way, by just logging instead of throwing an error.

crates/daemon/src/monitor.rs
```diff
@@ -1023,3 +1030,139 @@ static TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<prometheus::IntCounterV
         )
         .unwrap()
     });
+
+fn batch_script_get_history(
```
klochowicz: nit: I feel like the `batch_size` could just be the argument here

klochowicz (commented):
cool! I'll give this a spin to see. I wrote some comments, none of which should impact my results much (maybe except defining the `Vec` capacity upfront to avoid many reallocations).

will post the detailed review after getting the results.

crates/daemon/src/monitor.rs
```diff
@@ -1023,3 +1030,139 @@ static TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<prometheus::IntCounterV
...
+    client: Arc<electrum_client::Client>,
+    scripts: Vec<Script>,
+) -> tokio::sync::mpsc::Receiver<Vec<GetHistoryRes>> {
+    let (tx_script_updates, rx_script_updates) = tokio::sync::mpsc::channel(BATCH_SIZE * 4);
+
+    let batches = scripts.chunks(BATCH_SIZE).map(|batch| batch.to_owned());
+
+    // It's important to move here so the sender gets dropped and the receiver finishes correctly
```
luckysori, line 1042: 🙃 Given that you extracted this into a function, I think it should work with a `for` loop, since the original copy of the `sender` will be dropped as soon as you return from this function.

This is fine anyway, but the comment might not be entirely true anymore.

crates/daemon/src/monitor.rs
```diff
@@ -1023,3 +1030,139 @@ static TRANSACTION_BROADCAST_COUNTER: conquer_once::Lazy<prometheus::IntCounterV
...
+        // Some testnet address (to simulate a script where we get an empty response)
+        let script = bitcoin::Address::from_str("2MsQEtPJ6JJZszMYrD6udjUyTDFLczWQrv9")
+            .unwrap()
+            .script_pubkey();
+        let response_length_to_assert = 0;
+
+        test_runner(vec![script], response_length_to_assert).await;
+    }
+
+    async fn test_runner(scripts: Vec<Script>, assert_response_len: usize) {
```
luckysori: 🔧 Should probably be called `test_batch_script_get_history` or something like that.

luckysori (approved):
LGTM!

crates/daemon/src/monitor.rs
da-kami: I did this in https://github.com/itchysats/itchysats/pull/3197/commits/f8532aeff0a7a1a316a8b03110a9539d36ef387d
Given the context of this PR it makes sense to do this.

Note that the granularity and assertions of the tests changed. 
On the long run I would push for #3198 which will require processing on the outside, but we can discuss these changes on #3198

crates/daemon/src/monitor.rs
da-kami: Also done in https://github.com/itchysats/itchysats/pull/3197/commits/f8532aeff0a7a1a316a8b03110a9539d36ef387d

da-kami:
@klochowicz @luckysori  How should we move forward with testing this for production? 

Test benchmarks when running the test `given_many_scripts_then_batch_script_get_history_works` with `7000` scripts in release mode on my machine:

branch:

* `monitor-batching-again-refactor`
    * constant at ~54 MB
    * total execution time  `198.594171s`
* `monitor-batching-again`
   * constant at ~37 MB
   * total execution time `106.01s`

Internet speed was good, roughly `30mbit`

Note: It somehow slowed down at times, initially it was fast but got slower over time, towards the end faster again. I am sure what this is related to - it might just be the speed of my internet connection.

Looks like my optimization is not really an optimization. 

Is that good enough for now?

klochowicz (approved):
thanks for addressing this!

PR #3198, opened 2022-10-18
Process script updates one by one

da-kami:
Previously we assembled the complete history of all batches and then processed all updates.
With this change we now process them one by one, update the state and only accumulate the events that are ready.

The state update is cheap and can be done as part of processing the request.
Handling the CFD update could be done outside of sync, but we could regard that update as part of sync.
Only once the CFD is updated the sync is actually complete, because only then the state is communicated in the system, so it might be better to keep that within this scope.

### DIFF
--- a/crates/daemon/src/monitor.rs
+++ b/crates/daemon/src/monitor.rs
@@ -487,24 +487,23 @@ impl Actor {
             .cloned()
             .collect::<Vec<Script>>();
 
-        let histories = batch_script_get_history(self.client.clone(), scripts).await;
+        let mut rx_script_updates = batch_script_get_history(self.client.clone(), scripts).await;
 
-        let mut ready_events = self.state.update(
-            latest_block_height,
-            histories
+        let mut all_ready_events = Vec::new();
+        while let Some(script_history) = rx_script_updates.recv().await {
+            let script_status = script_history
                 .into_iter()
-                .map(|list| {
-                    list.into_iter()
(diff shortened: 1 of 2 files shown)


crates/daemon/src/monitor.rs
```diff
@@ -1012,9 +957,76 @@ struct ReinitMonitoring {
...
+                            height: history.height,
+                            tx_hash: history.tx_hash,
+                        })
+                        .collect();
+
+                    let mut ready_events = self.state.update(latest_block_height, script_status);
+
+                    if ready_events.is_empty() {
+                        return;
+                    }
```
luckysori: ❓ Is this necessary?

luckysori (approved):

da-kami:
@luckysori I did another iteration on this, because I was not convinced anymore that processing the CFD update should not be in scope of sync. I don't think we have to trigger another sync while we are still updating the CFDs with the result of the previous sync.

I found a simpler solution for optimizing the memory consumption further, while not moving the CFD update outside, I think it's good enough to aim for that. The changeset is smaller.

da-kami:
Looks like this might not actually an optimization, see: https://github.com/itchysats/itchysats/pull/3197#issuecomment-1282219467

da-kami:
Closing this for now, because we don't see an actual performance improvement by this code.

bonomat:
If you don't need the branch anymore, can you delete it please?

PR #3202, opened 2022-10-19
Add histogram to track monitor sync execution time

da-kami:
Track how the monitoring sync is performing over time.

### DIFF
--- a/crates/daemon/src/monitor.rs
+++ b/crates/daemon/src/monitor.rs
@@ -27,6 +27,7 @@ use sqlite_db;
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
+use std::time::Instant;
 use tracing::Instrument;
 use xtra_productivity::xtra_productivity;
 use xtras::SendInterval;
@@ -465,6 +466,8 @@ impl Actor {
 
     #[tracing::instrument("Sync monitor", skip_all, err)]
     async fn sync(&mut self) -> Result<()> {
+        let start_time = Instant::now();
+
         // Fetch the latest block for storing the height.
         // We do not act on this subscription after this call, as we cannot rely on
         // subscription push notifications because eventually the Electrum server will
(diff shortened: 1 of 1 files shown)


crates/daemon/src/monitor.rs
```diff
@@ -465,6 +465,8 @@ impl Actor {
 
     #[tracing::instrument("Sync monitor", skip_all, err)]
     async fn sync(&mut self) -> Result<()> {
+        let start_time = std::time::SystemTime::now();
```
bonomat: 🙃
You could also use 
```
let now = Instant::now();
...
now.elapsed().as_secs()
```

crates/daemon/src/monitor.rs
```diff
@@ -1075,6 +1086,20 @@ async fn batch_script_get_history(
     histories
 }
 
+static SYNC_DURATION_HISTOGRAM: conquer_once::Lazy<prometheus::Histogram> =
+    conquer_once::Lazy::new(|| {
+        prometheus::register_histogram!(
+            "monitor_sync_duration_seconds",
+            "The duration of one sync run of the monitor.",
+            vec![
+                10.0, 20.0, 30.0, 40.0, 50.0, 60.0, 70.0, 80.0, 90.0, 100.0, 110.0, 120.0, 130.0,
```
bonomat: Is 10 seconds the shortest you would expect?

bonomat (approved):
LGTM

crates/daemon/src/monitor.rs
da-kami: Fair questions - I think if we go under it would mean there are not many (long running) open positions. 
Given the current data we will likely not go under 10, but I'm happy to add a bucket for 5 and 2 as well.

crates/daemon/src/monitor.rs
da-kami: Neat! No error path for this one :)

PR #3222, opened 2022-10-24
Change the monitor sync interval from `20` to `120` secs

da-kami:
We talked about this, but never actually did it. 
Might be worth to see if that makes a difference.


klochowicz (commented):
won't this have a negative effect on the UX, when we detect state changes much less often?

bonomat (approved):

luckysori (commented):
I'm not against trying this, although I don't see it helping.

I think this will result in:

1. 2 minutes of intense work.
2. 2 minutes of rest.
3. 2 minutes of intense work.
4. 2 minutes of rest.
5. Etc.

Do we want those long breaks in between?

holzeis (approved):

da-kami:
See https://github.com/itchysats/itchysats/pull/3250

PR #3244, opened 2022-11-09
Feature: market close order

### LINKED ISSUE #2924: Feature: market close order
Right now the taker subscribes to the bitmex price feed and uses the bid/ask when proposing a settlement to the maker.

However, closing a position equals opening a counter position. For example, if you are **long** `100 BTCUSD` contracts, closing this position would mean you go `100 BTCUSD` contracts **short**.
This means, when collaborative settling, we should use the latest offer from the counterparty to close the position. Given the maker can still reject, it does not make much sense to send over his own offer price. Instead, the taker could just say, I want to settle and the maker accepts. This behavior equals a market close of your position.

bonomat:
The taker will _market_  close his position. 

Resolves #2924 

Note: this patch should go against 0.7.0 and can be released as a patch release from there.


daemon/src/taker_cfd.rs
```diff
@@ -73,24 +70,44 @@ impl Actor {
...
 
         let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
 
-        let proposal_closing_price = market_closing_price(bid, ask, Role::Taker, cfd.position());
+        let offer = self
+            .latest_offers
+            .get(&(cfd.contract_symbol(), cfd.position().counter_position()))
+            .context("Cannot propose settlement without price")?;
+
+        if !offer.is_safe_to_take(OffsetDateTime::now_utc()) {
```
da-kami, line 101: Given the frequent protocol failures the user might see some annoyance of "outdated offer" here when triggering collab settlement. This is unrelated to this fix, nothing to change here, but I wanted to point it out :)

klochowicz:
@bonomat would you mind creating a branch that branches from `release/0.7.0` and point this PR against that? It would be easier to review on github 🙏

daemon/src/taker_cfd.rs
```diff
@@ -73,24 +70,44 @@ impl Actor {
...
-            order_id,
-            bid,
-            ask,
-            quote_timestamp,
-        } = msg;
+        let ProposeSettlement { order_id } = msg;
 
         let cfd = self.db.load_open_cfd::<Cfd>(order_id, ()).await?;
 
-        let proposal_closing_price = market_closing_price(bid, ask, Role::Taker, cfd.position());
```
da-kami, line 91: 🛠️ We should remove the function `market_closing_price` and also use the maker's price in the projection - otherwise what we show in the UI does not match what is actually used!

bonomat:
> @bonomat would you mind creating a branch that branches from `release/0.7.0` and point this PR against that? It would be easier to review on github 🙏

done :)

daemon/src/taker_cfd.rs
bonomat: The timeout is currently set to 10 minutes . That's plenty of time to hopefully get an offer :D

da-kami: But that would mean that the only thing that protects us from using an outdated price is the automation's logic of "what is acceptable" as price. Is this fixing the problem of using an outdated price? Is the problem not that the price in the automation is potentially not up to date?

bonomat: That should be fine. The taker simply proposes the latest price what he got from the maker in here. The logic whether to accept or decline this price is outside of this scope. 
We simply spare the roundtrip and potentially protect the user by sending a settlement request with an outdated price.

da-kami (approved):
I think we can actually remove the `model::market_closing_price` - I think the now replaced usage in `projection` should have been the last usage of that function or did I miss something?

bonomat:

PR #3250, opened 2022-11-14
Revert "Self-implemented batching for `script_get_history`"

### LINKED ISSUE #3208: Monitor sync stops indefinitely
Follow up of https://github.com/itchysats/itchysats/pull/3197 and https://github.com/itchysats/meta/issues/89

We can [observe on Grafana](https://itchysats.grafana.net/goto/3z7aOrS4z?orgId=1) that when starting the `testnet` setup with an old version of the `mainnet` db eventually the `Sync` gets stuck - no new `Sync` is triggered in the monitor anymore.

There is no `Error` or `panic` recorded in the logs. Electrum should be reachable because the wallet is still syncing.

I think the problem is likely the `sync` function being stuck, and no new `Sync` being triggered (because the current one never finishes). I am not exactly sure what causes this yet.

da-kami:
This reverts commit c020c985

fixes https://github.com/itchysats/itchysats/issues/3208

---

If we revert this our testnet setup should be stable again, at the moment testnet is basically broken because we fail to sync due to #3208  and it remains unclear why that happens.

We don't want to invest more time into this "fix" for now, because the memory problem was eventually "overcome" in a different way.

Note: I kept the metric to measuring how long the monitoring sync takes because it might be useful in the future.

Consequences: 

Closed https://github.com/itchysats/itchysats/pull/3222 because we don't know about side effects and we don't have to change this for now. If you think we should still merge https://github.com/itchysats/itchysats/pull/3222 please re-open.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1054,7 +1054,6 @@ dependencies = [
  "tokio-extras",
  "tokio-util",
  "tracing",
- "tracing-subscriber",
  "uuid 1.1.2",
  "x25519-dalek",
  "xtra",
--- a/crates/daemon/Cargo.toml
+++ b/crates/daemon/Cargo.toml
@@ -60,4 +60,3 @@ xtras = { path = "../xtras" }
 [dev-dependencies]
 serde_test = "1"
 time = { version = "0.3.15", features = ["std"] }
-tracing-subscriber = { version = "0.3", features = ["env-filter"] }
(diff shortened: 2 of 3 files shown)


klochowicz (approved):
👍

bonomat (approved):

luckysori (approved):

da-kami:
bors retry

Note: we keep running into this problem on CI lately:

```
   Downloaded ndk-sys v0.4.0
error: failed to verify the checksum of `ndk-sys v0.4.0`
Error: Process completed with exit code 102.
```

@klochowicz @luckysori any idea what could cause this and if we can do something about this? e.g. update dependencies?

luckysori:
> any idea what could cause this and if we can do something about this? e.g. update dependencies?

Nope, but what fixed it?

PR #3255, opened 2022-11-15
Remove dependencies from bitmex stream lib

holzeis:
Adding the itchysats bitmex-stream as dependency to 10101 resulted into the known dependency conflict of the normalization-unicode between url and bip39. I adapted the bitmex stream lib and refactored the code to not being dependent on the url crate anymore. This allows us to use it together with bip39 in 10101.

We might not want to merged that PR, but rather keep it until BIP39 upgrades their dependencies and we don't need that change anymore.

### DIFF
--- a/crates/bitmex-stream/src/lib.rs
+++ b/crates/bitmex-stream/src/lib.rs
@@ -83,7 +83,7 @@ fn subscribe_impl<const N: usize>(
 
         loop {
             tokio::select! {
-                _ = tokio_extras::time::sleep_silent(Duration::from_secs(5)) => {
+                _ = sleep(Duration::from_secs(5)) => {
                     let span = tracing::trace_span!("Ping BitMex");
                     span.in_scope(|| tracing::trace!("No message from BitMex in the last 5 seconds, pinging"));
                     let _ = connection
@@ -124,6 +124,11 @@ fn subscribe_impl<const N: usize>(
     };
 
     stream.boxed()
+}
+
+pub async fn sleep(duration: Duration) {
+    #[allow(clippy::disallowed_methods)]
+    tokio::time::sleep(duration).await
 }
 
 #[derive(Debug, Copy, Clone)]
--- a/crates/bitmex-stream/src/lib.rs
+++ b/crates/bitmex-stream/src/lib.rs
@@ -13,7 +13,6 @@ use std::time::SystemTime;
 use std::time::UNIX_EPOCH;
 use tokio_tungstenite::tungstenite;
 use tracing::Instrument;
-use url::Url;
 
 pub use anyhow::Error;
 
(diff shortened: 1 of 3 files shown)


crates/bitmex-stream/src/lib.rs
```diff
@@ -185,19 +184,9 @@ impl Credentials {
         }
     }
 
-    fn sign(&self, method: Method, expires: u64, url: &Url, body: &str) -> Signature {
+    fn sign(&self, method: Method, expires: u64, url: &str, body: &str) -> Signature {
```
bonomat, line 187: Given `url` should now only be the path you could call it `path`.

klochowicz:
I found a simpler alternative which we might want to consider: https://github.com/itchysats/itchysats/pull/3256 (just relaxing patch version requirements is enough). Tested locally with your project, can push a fixup to your PR if it gets merged.

crates/bitmex-stream/src/lib.rs
```diff
@@ -68,7 +67,7 @@ fn subscribe_impl<const N: usize>(
                 .duration_since(UNIX_EPOCH)?
                 .add(Duration::from_secs(5))
                 .as_secs();
-            let signature = credentials.sign(Method::GET, expires, &Url::parse(url.as_str())?, "");
+            let signature = credentials.sign(Method::GET, expires, &url, "");
```
bonomat, line 70: Does this still work? Before we passed in a `Url` and called `.path()` on it, returning only the `path` of the url, now you pass in the whole `url`.

bonomat: I tested it for you, it does not 😅
```
2022-11-15T23:36:36.300115Z  INFO authenticated_api: {"status":401,"error":"Signature not valid.","meta":{},"request":{"op":"authKeyExpires","args":["**********--oa9",1668555400,"XXXXXXX"]}}
```

bonomat (changes requested):
This breaks signing unfortunately.

klochowicz:
I suspect this will not be needed if we go with #3256

crates/bitmex-stream/src/lib.rs
holzeis: see comment on this commit 8f935e9dfc704b12eb6c98abd873db30dd39aece. I changed the api asking the caller to provide the path + query parameters without the host. The tests are working.

holzeis:
Closing in favor of #3256

holzeis:
> I found a simpler alternative which we might want to consider: #3256 (just relaxing patch version requirements is enough). Tested locally with your project, can push a fixup to your PR if it gets merged.

Thanks @klochowicz thats indeed the bettter solution.

